PR #1127, opened 2018-01-31
Rule S3353: Unchanged local variables should be 'const'

### LINKED ISSUE #1065: Rule S3353: Unchanged local variables should be "const"
[RSPEC-3353](https://jira.sonarsource.com/browse/RSPEC-3353)

Marking a variable that is unchanged after initialization const is an indication to future maintainers that "no this isn't updated, and it's not supposed to be". const should be used in these situations in the interests of code clarity.

michalb-sonar:
Fix #1065

### DIFF
--- a/sonaranalyzer-dotnet/rspec/cs/S3353_c#.json
+++ b/sonaranalyzer-dotnet/rspec/cs/S3353_c#.json
@@ -1,0 +1,15 @@
+{
+  "title": "Unchanged local variables should be \"const\"",
+  "type": "CODE_SMELL",
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "2min"
+  },
+  "tags": [
+    
+  ],
+  "defaultSeverity": "Critical",
+  "ruleSpecification": "RSPEC-3353",
+  "sqKey": "S3353"
+}
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
@@ -41,9 +41,7 @@ namespace SonarAnalyzer.Rules.CSharp
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
 
-        public SyntaxToken paramSymbol { get; private set; }
-
-        private enum DeclarationType { Unknown, Value, Reference, String };
+        private enum DeclarationType { CannotBeConst, Value, Reference, String };
 
         protected override void Initialize(SonarAnalysisContext context)
         {
@@ -56,7 +54,7 @@ namespace SonarAnalyzer.Rules.CSharp
                     }
 
                     var declaredType = FindDeclarationType(localDeclaration, c.SemanticModel);
-                    if (declaredType == DeclarationType.Unknown)
+                    if (declaredType == DeclarationType.CannotBeConst)
                     {
                         return;
                     }
@@ -77,16 +75,15 @@ namespace SonarAnalyzer.Rules.CSharp
             SemanticModel semanticModel)
         {
             var declaredTypeSyntax = localDeclaration.Declaration?.Type;
-
             if (declaredTypeSyntax == null)
             {
-                return DeclarationType.Unknown;
+                return DeclarationType.CannotBeConst;
             }
 
             var declaredType = semanticModel.GetTypeInfo(declaredTypeSyntax).Type;
             if (declaredType == null)
             {
-                return DeclarationType.Unknown;
+                return DeclarationType.CannotBeConst;
             }
 
             if (declaredType.Is(KnownType.System_String))
@@ -94,11 +91,16 @@ namespace SonarAnalyzer.Rules.CSharp
                 return DeclarationType.String;
             }
 
+            if (declaredType.OriginalDefinition?.DerivesFrom(KnownType.System_Nullable_T) ?? false)
+            {
+                // Defining nullable as const raises error CS0283.
+                return DeclarationType.CannotBeConst;
+            }
+
             return declaredType.IsValueType
                 ? DeclarationType.Value
                 : DeclarationType.Reference;
         }
-
 
         private static bool IsInitializedWithCompatibleConstant(VariableDeclaratorSyntax variableDeclarator,
             SemanticModel semanticModel, DeclarationType declarationType)
@@ -115,7 +117,6 @@ namespace SonarAnalyzer.Rules.CSharp
             }
 
             var constantValue = constantValueContainer.Value;
-
             if (constantValue is string)
             {
                 return declarationType == DeclarationType.String;
@@ -133,7 +134,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private static bool HasMutableUsagesInMethod(VariableDeclaratorSyntax parameter, ISymbol parameterSymbol,
             SemanticModel semanticModel)
         {
-            var methodSyntax = parameter.Parent?.AncestorsAndSelf()?.FirstOrDefault(IsMethodLike);
+            var methodSyntax = parameter?.Ancestors()?.FirstOrDefault(IsMethodLike);
             if (methodSyntax == null)
             {
                 return false;
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
@@ -1,0 +1,189 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2018 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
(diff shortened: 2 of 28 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
```diff
@@ -0,0 +1,189 @@
...
+            {
+                var symbol = semanticModel.GetSymbolInfo(id).Symbol;
+                return Equals(parameterSymbol, symbol);
+            }
+        }
+
+        private static bool IsMutatingUse(IdentifierNameSyntax id)
+        {
+            var parent = id.Parent;
+            if (parent is AssignmentExpressionSyntax assignmentExpression &&
```
valhristov: Does this cover `x += 5` and the variations / * % - | &, etc.? Also shifting operators too...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
```diff
@@ -0,0 +1,189 @@
...
+                Equals(id, assignmentExpression?.Left))
+            {
+                return true;
+            }
+
+            if (parent is PostfixUnaryExpressionSyntax ||
```
valhristov: if this `if` is before the first `if`, you will be able to merge the two `if`s

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
```diff
@@ -0,0 +1,189 @@
...
+        internal const string DiagnosticId = "S3353";
+        private const string MessageFormat = "Add the 'const' modifier to '{0}'.";
+
+        private static readonly DiagnosticDescriptor rule =
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+        public SyntaxToken paramSymbol { get; private set; }
+
+        private enum DeclarationType { Unknown, Value, Reference, String };
```
valhristov: Why do we care about Unknown and String? Could we just do it with "CanBeConst" and "CannotBeConst"? (except that it sounds worse)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
```diff
@@ -0,0 +1,189 @@
...
+    public sealed class UnchangedLocalVariablesShouldBeConst : SonarDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S3353";
+        private const string MessageFormat = "Add the 'const' modifier to '{0}'.";
+
+        private static readonly DiagnosticDescriptor rule =
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+        public SyntaxToken paramSymbol { get; private set; }
```
valhristov: This should be deleted

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
michalb-sonar: Not without the braces, which would be quite confusing (parent is either one type or the other).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
michalb-sonar: Yes. I'll add more tests.

sonaranalyzer-dotnet/rspec/cs/S3353_c#.json
```diff
@@ -0,0 +1,15 @@
...
+  "type": "CODE_SMELL",
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "2min"
+  },
+  "tags": [
+    
+  ],
+  "defaultSeverity": "Critical",
```
Evangelink: It might be worth reconsidering the default severity: `Critical` for this looks overkill.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
Evangelink: Unused property.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
michalb-sonar: That was my first attempt.
The const value types do not accept nulls, const reference type only accept nulls. And string is special as it accepts literal strings (or simple manipulation of it!) and nulls. Simplifying that to "CanBeConst" results in few FPs.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
```diff
@@ -0,0 +1,189 @@
...
+            }
+
+            return declarationType == DeclarationType.Reference ||
+                   declarationType == DeclarationType.String;
+        }
+
+        private static bool HasMutableUsagesInMethod(VariableDeclaratorSyntax parameter, ISymbol parameterSymbol,
+            SemanticModel semanticModel)
+        {
+            var methodSyntax = parameter.Parent?.AncestorsAndSelf()?.FirstOrDefault(IsMethodLike);
```
Evangelink: Why `Parent?.AncestorsAndSelf()` instead of `Ancestors()`?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnchangedLocalVariablesShouldBeConst.cs
```diff
@@ -0,0 +1,261 @@
...
+            string str8 = String.Empty;
+
+            object obj1 = null; // Noncompliant
+            object obj2 = new object();
+            object obj3 = TestUtils.GetObj();
+            object obj4 = 1;
+
+            IEnumerable<string> enumerable = null; // Noncompliant
+
+            //dynamic value = 42; // Compliant. Const only works with null.
```
Evangelink: Why are these tests commented out?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnchangedLocalVariablesShouldBeConst.cs
```diff
@@ -0,0 +1,261 @@
...
+        {
+            x = 42;
+        }
+
+        public static void RefParam(ref int x)
+        {
+            x = 42;
+        }
+
+        public static void NormalCall(int x) { }
```
duncanp-sonar: Why not set _x = 42_ in this test case too?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnchangedLocalVariablesShouldBeConst.cs
```diff
@@ -0,0 +1,261 @@
...
+            System.Random random = 1;
+        }
+
+        // invalid code
+        public void (int arg)
+        {
+            int intVar = 1;
+        }
+    }
+}
```
duncanp-sonar: Are there any test cases using arrays?

The following examples shouldn't trigger the rule as they can't be made _const_:
_int[] x = { 1, 2, 3 };
int[] y = new int[] { 1, 2, 3 };
Exception[] z = { };_

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
michalb-sonar: That would be too simple ;)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnchangedLocalVariablesShouldBeConst.cs
michalb-sonar: Great catch! 😱

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
Evangelink: I like simple :)

PR #1128, opened 2018-01-31
Fix S1607: Do not raise when reason is specified in the Ignore attrib…

### LINKED ISSUE #1124: Fix S1607: Should be applied for test classes as well
### Description
Since tests can be ignored by putting the [Ignore] attribute on the test class, I think it would make sense that the rule as also checking those cases as well.
Now it's just checking if the [Ignore] is put on the test method level.

[RSPEC-1607](https://jira.sonarsource.com/browse/RSPEC-1607)

### Repro steps

Put [Ignore] next to the [TestClass] attribute and all the tests are ignored but the rule is not triggered.

### Expected behavior
Rule should be triggered for the test class as well.

### Additional improvement to be considered
- Another rule which would fail if the Ignore attribute is put on the class level would also be satisfactory.

- In our team people are putting the comment for ignoring the test as a parameter of the Ignore attribute. Maybe the rule could also take that comment as a valid fix.
For example:
[Ignore("Test is ignored because....")]

### LINKED ISSUE #1126: Fix S1607: Should not raise issues for [Ignore] attribute with a parameter
The latest version of MSTest adds a string parameter to the Ignore attribute to specify a reason. S1607 should not raise issues when the parameter is specified.

Requested in #1124

valhristov:
…ute, extend support for test classes

Fix #1124 
Fix #1126

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
@@ -55,6 +55,11 @@ namespace SonarAnalyzer.Rules.CSharp
         {
             context.RegisterSyntaxNodeActionInNonGenerated(c =>
             {
+                if (!c.IsTest())
+                {
+                    return;
+                }
+
                 var attribute = (AttributeSyntax)c.Node;
                 if (HasReasonPhrase(attribute) ||
                     HasTrailingComment(attribute) ||
@@ -69,16 +74,24 @@ namespace SonarAnalyzer.Rules.CSharp
                     return;
                 }
 
-                var testMethodOrClass = c.SemanticModel.GetDeclaredSymbol(attributeTarget);
+                var attributes = GetAllAttributes(attributeTarget, c.SemanticModel);
 
-                if (testMethodOrClass != null &&
-                    testMethodOrClass.GetAttributes().Any(IsTestOrTestClassAttribute) &&
-                    !testMethodOrClass.GetAttributes().Any(IsWorkItemAttribute))
+                if (attributes.Any(IsTestOrTestClassAttribute) &&
+                    !attributes.Any(IsWorkItemAttribute))
                 {
                     c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, attribute.GetLocation()));
                 }
             },
             SyntaxKind.Attribute);
+        }
+
+        private IEnumerable<AttributeData> GetAllAttributes(SyntaxNode syntaxNode, SemanticModel semanticModel)
+        {
+            var testMethodOrClass = semanticModel.GetDeclaredSymbol(syntaxNode);
+
+            return testMethodOrClass == null
+                ? Enumerable.Empty<AttributeData>()
+                : testMethodOrClass.GetAttributes();
         }
 
         private bool HasReasonPhrase(AttributeSyntax ignoreAttributeSyntax) =>
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
@@ -23,6 +23,7 @@ using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
@@ -43,6 +44,7 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static readonly ISet<KnownType> TrackedTestMethodAttributes = ImmutableHashSet.Create(
             KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_TestMethodAttribute,
+            KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_TestClassAttribute,
             KnownType.NUnit_Framework_TestAttribute,
             KnownType.NUnit_Framework_TestCaseAttribute,
             KnownType.NUnit_Framework_TestCaseSourceAttribute,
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldNotBeIgnored.cs
```diff
@@ -13,18 +13,50 @@ public void Foo1()
         }
 
         [TestMethod]
-        [Ignore] // Reenable when something
+        [Ignore] // Ignored because reasons...
```
michalb-sonar: 🤣

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
```diff
@@ -51,46 +53,56 @@ public sealed class TestMethodShouldNotBeIgnored : SonarDiagnosticAnalyzer
...
+                var testMethodOrClass = c.SemanticModel.GetDeclaredSymbol(attributeTarget);
+
+                if (testMethodOrClass != null &&
+                    testMethodOrClass.GetAttributes().Any(IsTestOrTestClassAttribute) &&
+                    !testMethodOrClass.GetAttributes().Any(IsWorkItemAttribute))
+                {
+                    c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, attribute.GetLocation()));
+                }
+            },
+            SyntaxKind.Attribute);
```
valhristov: Converted to AttributeSyntax action to support test class and test method with the same code

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
```diff
@@ -51,46 +53,56 @@ public sealed class TestMethodShouldNotBeIgnored : SonarDiagnosticAnalyzer
 
         protected override void Initialize(SonarAnalysisContext context)
         {
-            context.RegisterSyntaxNodeActionInNonGenerated(
-                c =>
-                {
-                    if (!c.IsTest())
```
valhristov: Should I include this check?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
```diff
@@ -51,46 +53,56 @@ public sealed class TestMethodShouldNotBeIgnored : SonarDiagnosticAnalyzer
...
+            },
+            SyntaxKind.Attribute);
+        }
+
+        private bool HasReasonPhrase(AttributeSyntax ignoreAttributeSyntax) =>
+            ignoreAttributeSyntax.ArgumentList?.Arguments.Count > 0; // Any ctor argument counts are reason phrase
+
+        private static bool HasTrailingComment(SyntaxNode ignoreAttributeSyntax) =>
+            ignoreAttributeSyntax.Parent.GetTrailingTrivia()
+                .Any(trivia => trivia.IsKind(SyntaxKind.SingleLineCommentTrivia));
```
Evangelink: I know this was like this but maybe we should also handle multiline comments `/* Ignored because something */`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
```diff
@@ -51,46 +53,56 @@ public sealed class TestMethodShouldNotBeIgnored : SonarDiagnosticAnalyzer
...
+                var attributeTarget = attribute.Parent?.Parent;
+                if (attributeTarget == null)
+                {
+                    return;
+                }
+
+                var testMethodOrClass = c.SemanticModel.GetDeclaredSymbol(attributeTarget);
+
+                if (testMethodOrClass != null &&
+                    testMethodOrClass.GetAttributes().Any(IsTestOrTestClassAttribute) &&
```
Evangelink: Isn't it better to avoid querying `GetAttributes()` twice?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldNotBeIgnored.cs
Evangelink: This was changed because the text was confusing for @valhristov

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
michalb-sonar: Why did you remove that check? RSPEC still says test code.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
valhristov: @Evangelink I don't know, it returns ImmutableArray<T> which is value type and I cannot say if it is the same instance; I suppose it is returning a new instance because `GetAttributes()` is a method...

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldNotBeIgnored.cs
```diff
@@ -1,142 +1,174 @@
...
+        [TestMethod]
+        [Ignore]
+        [WorkItem(1234)]
+        public void Foo5()
+        {
+        }
+    }
+
+    [Ignore, TestClass]
+//   ^^^^^^
```
duncanp-sonar: Should this line have _Noncompliant_ on it/

Evangelink: Good catch

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldNotBeIgnored.cs
```diff
@@ -1,142 +1,174 @@
...
+    class MsTestClass4
+    {
+    }
+}
+
+namespace Tests.Diagnostics2
+{
+    using NUnit.Framework;
+
+    class NUnitClass
```
Evangelink: Is there no way to end-up in the same situation with NUnit? The ignore attribute at  the class level.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldNotBeIgnored.cs
valhristov: If you have `// ^^^` the `// Noncompliant` comment is not required; you could also include the `{{Message}}` check.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldNotBeIgnored.cs
valhristov: NUnit's Ignore attribute requires message argument, same as XUnit

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldNotBeIgnored.cs
valhristov: @duncanp-sonar @Evangelink let me know if you think I should add the Noncompliant comment

duncanp-sonar: Thanks. I didn't know whether it was required or not.
I think the issues are easier to see if the Noncompliant comment is there, and it seems to be normal practice to include it.

PR #1129, opened 2018-02-01
Fix S4220 : Rule should not throw StackOverflowException

kvpt:
Hi,

We encountered a StackOverflow in one of our projects during the build when using the sonar c# analyser.

After a divide and conquer strategy, I was able to isolate the code and deduce the faulty rule.

It's a very specific combination of criteria :
- An encapsuled EventHandler without event keyword
- A nested conditionnal access and invoke of the handler
In this case the method GetEventSymbol get called in infinite loop.

Here, the reproduction case and the fix.

There may be better way to handle this case, I don't know well the roselyn api.

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/EventHandlerDelegateShouldHaveProperArguments.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/EventHandlerDelegateShouldHaveProperArguments.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2018 SonarSource SA
  * mailto: contact AT sonarsource DOT com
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/EventHandlerDelegateShouldHaveProperArgumentsTest.cs
```diff
@@ -34,5 +34,15 @@ public void EventHandlerDelegateShouldHaveProperArguments()
             Verifier.VerifyAnalyzer(@"TestCases\EventHandlerDelegateShouldHaveProperArguments.cs",
                 new EventHandlerDelegateShouldHaveProperArguments());
         }
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void EventHandlerDelegateShouldHaveProperArgumentsShoudNotStackOverflow()
+        {
+            Verifier.VerifyNoExceptionThrown(@"TestCases\EventHandlerDelegateShouldHaveProperArguments.StackOverflow.Fixed.cs", new []
```
michalb-sonar: We use `.Fixed.cs` convention for code fixes.

Please remove this test method and move the test to `EventHandlerDelegateShouldHaveProperArguments.cs`. Preferably to a separate class/namespace with a comment on what it is testing.

kvpt:
I made the modifications.
I merged the test case and removed the test since the existing test is also affected by the problem.

michalb-sonar:
Thank you for a quick change @kvpt. We appreciate your contribution!

PR #1130, opened 2018-02-01
[DO NOT MERGE] Fix S4220 : Rule should not throw StackOverflowException

michalb-sonar:


michalb-sonar:
Merged in https://github.com/SonarSource/sonar-csharp/pull/1129

PR #1134, opened 2018-02-01
Fix S3343: False positive when no parameter without caller info is sp…

### LINKED ISSUE #1132: Fix S3343: False positive when no parameter without caller info is specified
The following should not raise issues:
```
public void Method8([CallerMemberName]string callerMemberName = null,
    [CallerFilePath]string callerFilePath = null, [CallerLineNumber]int callerLineNumber = 0) { }
```

https://peach.sonarsource.com/project/issues?id=nlog&issues=AWFI2naqjhnAOMVnQVmw&open=AWFI2naqjhnAOMVnQVmw
https://peach.sonarsource.com/project/issues?id=nlog&issues=AWFI2naqjhnAOMVnQVmv&open=AWFI2naqjhnAOMVnQVmv

valhristov:
Fix #1132

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallerInformationParametersShouldBeLast.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallerInformationParametersShouldBeLast.cs
@@ -60,25 +60,30 @@ namespace SonarAnalyzer.Rules.CSharp
                         return;
                     }
 
-                    ParameterSyntax parameterWithoutCallerInfoAttribute = null;
+                    ParameterSyntax noCallerInfoParameter = null;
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallerInformationParametersShouldBeLast.cs
```diff
@@ -60,26 +60,31 @@ protected override void Initialize(SonarAnalysisContext context)
                         return;
                     }
 
-                    ParameterSyntax parameterWithoutCallerInfoAttribute = null;
+                    ParameterSyntax noCallerInfoParameter = null;
```
michalb-sonar: I think I liked the old name better.

PR #1135, opened 2018-02-01
Fix S3925: Should not raise on explicit implementation of ISerializable

### LINKED ISSUE #1076: Fix S3925: Should not raise on explicit implementation of ISerializable
[S3925](https://jira.sonarsource.com/browse/RSPEC-3925) raises issues on `protected virtual GetObjectData`, but it should not. See the example here:
https://msdn.microsoft.com/en-us/library/ms182342.aspx

The bug was reported here (example):
https://groups.google.com/forum/#!topic/sonarlint/_EtvqIgl0e8

valhristov:
Fix #1076 

Not perfect solution, but to make it right I have to rewrite half of the rule. Basically disabled the check for public/virtual GetObjectData on explicitly implemented interfaces. There is no check if the explicit implementation calls the virtual method, or if there is a virtual method at all.

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ImplementISerializableCorrectly.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ImplementISerializableCorrectly.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2018 SonarSource SA
  * mailto: contact AT sonarsource DOT com
@@ -21,6 +21,7 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Runtime.Serialization;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -142,9 +143,9 @@ namespace SonarAnalyzer.Rules.CSharp
                 yield break;
             }
 
-            var isPublicVirtual = getObjectData.DeclaredAccessibility == Accessibility.Public &&
-                (getObjectData.IsVirtual || getObjectData.IsOverride);
-            if (typeSymbol.IsSealed || isPublicVirtual)
+            if (typeSymbol.IsSealed ||
+                IsPublicVirtual(getObjectData) ||
+                IsExplicitImplementation(getObjectData))
             {
                 yield break;
             }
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ImplementISerializableCorrectly.cs
```diff
@@ -231,23 +229,24 @@ private static bool IsCallingBaseConstructor(IMethodSymbol constructorSymbol)
...
+        private static bool IsOrImplementsISerializable(ITypeSymbol typeSymbol) =>
+            typeSymbol.Is(KnownType.System_Runtime_Serialization_ISerializable) ||
+            typeSymbol.Implements(KnownType.System_Runtime_Serialization_ISerializable);
+
+        private static bool IsPublicVirtual(IMethodSymbol methodSymbol) =>
+            methodSymbol.DeclaredAccessibility == Accessibility.Public &&
+            (methodSymbol.IsVirtual || methodSymbol.IsOverride);
+
+        private static bool IsExplicitImplementation(IMethodSymbol methodSymbol) =>
+            methodSymbol.Name == "System.Runtime.Serialization.ISerializable.GetObjectData";
```
michalb-sonar: This is a bit odd. Would that work?
`methodSymbol.ExplicitInterfaceImplementations.Any(IsGetObjectData)`

I'd also rename the method to `IsExplicitImplementationOfGetObject` or something similar.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ImplementISerializableCorrectly.cs
```diff
@@ -95,17 +95,27 @@ protected Serializable_NoAttribute_1(SerializationInfo info, StreamingContext co
 
     [Serializable]
     public class Serializable_ExplicitImplementation : ISerializable
-//               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Noncompliant
     {
         public Serializable_ExplicitImplementation()
         { /*do something*/ }
         protected Serializable_ExplicitImplementation(SerializationInfo info, StreamingContext context)
         { /*do something*/ }
         void ISerializable.GetObjectData(SerializationInfo info, StreamingContext context)
```
michalb-sonar: Please add a comment stating this is a FN, as the virtual method is not called from implicit implementation.

valhristov: done

PR #1136, opened 2018-02-01
Update S3440: 'Variables should not be checked before assignment' should not raise on properties

### LINKED ISSUE #1000: Update S3440: "Variables should not be checked before assignment" should not raise on properties
[RSPEC-3440](https://jira.sonarsource.com/browse/RSPEC-3440)

There's no point in checking a variable against the value you're about to assign it. Save the cycles and lines of code, and simply perform the assignment.

This rule should not raise issues on properties because they could have undesired side effects and the check might be on purpose.

michalb-sonar:
Fix #1000

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantConditionalAroundAssignment.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantConditionalAroundAssignment.cs
@@ -19,6 +19,7 @@
  */
 
 using System.Collections.Immutable;
+using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -48,7 +49,8 @@ namespace SonarAnalyzer.Rules.CSharp
                 {
                     var ifStatement = (IfStatementSyntax)c.Node;
                     if (ifStatement.Else != null ||
-                        ifStatement.Parent is ElseClauseSyntax)
+                        ifStatement.Parent is ElseClauseSyntax ||
+                        ifStatement.Ancestors().OfType<PropertyDeclarationSyntax>().Any())
                     {
                         return;
                     }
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantConditionalAroundAssignment.cs
```diff
@@ -48,7 +49,8 @@ protected sealed override void Initialize(SonarAnalysisContext context)
                 {
                     var ifStatement = (IfStatementSyntax)c.Node;
                     if (ifStatement.Else != null ||
-                        ifStatement.Parent is ElseClauseSyntax)
+                        ifStatement.Parent is ElseClauseSyntax ||
+                        ifStatement.Ancestors().OfType<PropertyDeclarationSyntax>().Any())
```
Evangelink: I am not sure if there is any difference but since Roslyn provides a `FirstAncestorOrSelf<T>` it might be better to use it rather than `Ancestors().OfType<PropertyDeclarationSyntax>().Any()`.

michalb-sonar: That is interesting. They boil down to the same thing, but `OfType()` does two casts (is and then C-style cast), which I presume is slower.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantConditionalAroundAssignment.cs
```diff
@@ -48,7 +49,8 @@ protected sealed override void Initialize(SonarAnalysisContext context)
                 {
                     var ifStatement = (IfStatementSyntax)c.Node;
                     if (ifStatement.Else != null ||
-                        ifStatement.Parent is ElseClauseSyntax)
+                        ifStatement.Parent is ElseClauseSyntax ||
+                        ifStatement.FirstAncestorOrSelf<PropertyDeclarationSyntax>() is PropertyDeclarationSyntax)
```
Evangelink: I am going to let you review the line of code you wrote :D

Evangelink: ... here

Evangelink: Actually I don't know if it is wrong or not but this looks really weird...

michalb-sonar: That should work fine. I'll double-check on peach just in case.

PR #1139, opened 2018-02-02
Fix S3459: False Positive with private classes that could be passed a…

### LINKED ISSUE #242: Fix S3459: False Positive with interop methods
S3459 is thrown on ullTotalPhys even though the Windows API function GlobalMemoryStatusEx sets it before use.

```csharp
public static class Memory
{
    [return: MarshalAs(UnmanagedType.Bool)]
    [DllImport("kernel32.dll", CharSet = CharSet.Auto, SetLastError = true)]
    static extern bool GlobalMemoryStatusEx([In, Out] MEMORYSTATUSEX lpBuffer);

    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Auto)]
    private class MEMORYSTATUSEX
    {
        public uint dwLength;
        public uint dwMemoryLoad;
        public ulong ullTotalPhys;
        public ulong ullAvailPhys;
        public ulong ullTotalPageFile;
        public ulong ullAvailPageFile;
        public ulong ullTotalVirtual;
        public ulong ullAvailVirtual;
        public ulong ullAvailExtendedVirtual;
        public MEMORYSTATUSEX()
        {
            this.dwLength = (uint)Marshal.SizeOf(typeof(MEMORYSTATUSEX));
        }
    }

    public static ulong TotalPhysicalMemory
    {
        get
        {
            MEMORYSTATUSEX memStatus = new MEMORYSTATUSEX();
            if (GlobalMemoryStatusEx(memStatus))
            {
                return memStatus.ullTotalPhys;
            }
            else
            {
                return 0;
            }
        }
    }
}
```

valhristov:
…s arguments to extern methods

Fix #242 

Private classes could be passed to extern methods without the need a reference in the consuming method.

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
@@ -58,6 +58,7 @@ namespace SonarAnalyzer.Rules.CSharp
                 {
                     var namedType = (INamedTypeSymbol)c.Symbol;
                     if (!namedType.IsClassOrStruct() ||
+                        namedType.GetAttributes().Any(IsStructLayoutAttribute) ||
                         namedType.ContainingType != null)
                     {
                         return;
@@ -244,6 +245,9 @@ namespace SonarAnalyzer.Rules.CSharp
 
             return false;
         }
+
+        private static bool IsStructLayoutAttribute(AttributeData attribute) =>
+            attribute.AttributeClass.Is(KnownType.System_Runtime_InteropServices_StructLayoutAttribute);
 
         private static readonly ISet<SyntaxKind> PreOrPostfixOpSyntaxKinds = ImmutableHashSet.Create(
             SyntaxKind.PostDecrementExpression,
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -58,6 +58,7 @@ protected sealed override void Initialize(SonarAnalysisContext context)
                 {
                     var namedType = (INamedTypeSymbol)c.Symbol;
                     if (!namedType.IsClassOrStruct() ||
+                        namedType.GetAttributes().Any(IsStructLayoutAttribute) ||
```
Evangelink: I really like this approach ;)

duncanp-sonar: +1.
Could perhaps have a very short comment explaining why e.g. _// ignore interop types - see #xxx_

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NotAssignedPrivateMember.cs
```diff
@@ -93,4 +94,14 @@ void M()
...
+
+    // https://github.com/SonarSource/sonar-csharp/issues/242
+    public class MyClass
+    {
+        [StructLayout(LayoutKind.Sequential)]
+        private class InteropMethodArgument
+        {
+            public uint number; // Compliant, we don't raise on members of classes with StructLayout attribute
+        }
+    }
```
duncanp-sonar: Real corner case - should report on subclasses of _InteropMethodArgument_ since the attribute isn't inherited.
Not sure it's worth testing for though.

valhristov: We will raise in my opinion, but I am not going to add a test case :)

PR #1140, opened 2018-02-02
Fix S1643: check if expression is concatenation

### LINKED ISSUE #1138: Fix S1643: check if expression is concatenation
### Description

The following code triggers a S1643 `Strings should not be concatenated using '+' in a loop` error:

```csharp
string fileContent = null;
foreach (var item in items)
  if (item.requiresFile)
     fileContent = fileContent ?? FileSystem.ReadAllText(fn);
```

I think the root cause of this issue is that you never check that the binary expression (in this case `x ?? y`) uses the string concatenation operator `+` or not.

I am not 100% sure, but I think a check is missing after this:
https://github.com/SonarSource/sonar-csharp/blob/b8a7415a5cee73685924cf92b7ee3ed7c3bbdb16/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/StringConcatenationInLoopBase.cs#L64

### Repro steps

Try to analize this code:
```csharp
string fileContent = null;
foreach (var item in new[]{ false, true })
  if (item)
     fileContent = fileContent ?? FileSystem.ReadAllText(fn);
```

### Expected behavior

SonarQube does not warn about S1643.

### Actual behavior

SonarQube DOES warn about S1643.

### Known workarounds

No workaround is known at the moment.

### Related information

* SonarC# Version: `SonarQube Scanner for MSBuild 4.0.2`
* Visual Studio Version: `2017`
* If running through the Scanner for MSBuild, its version: `15.5.180.51428`

michalb-sonar:
Fix #1138

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringConcatenationInLoop.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringConcatenationInLoop.cs
@@ -60,6 +60,10 @@ namespace SonarAnalyzer.Rules.CSharp
             SyntaxKind.ForStatement,
             SyntaxKind.ForEachStatement);
 
+        private static readonly ISet<SyntaxKind> AddOperators = ImmutableHashSet.Create(
+            SyntaxKind.PlusToken,
+            SyntaxKind.PlusEqualsToken);
+
         private static readonly ImmutableArray<SyntaxKind> simpleAssignmentKinds =
             ImmutableArray.Create(SyntaxKind.SimpleAssignmentExpression);
 
@@ -71,5 +75,9 @@ namespace SonarAnalyzer.Rules.CSharp
         protected override ImmutableArray<SyntaxKind> CompoundAssignmentKinds => compoundAssignmentKinds;
 
         protected sealed override Helpers.GeneratedCodeRecognizer GeneratedCodeRecognizer => Helpers.CSharp.GeneratedCodeRecognizer.Instance;
+
+        protected override bool IsAddExpression(BinaryExpressionSyntax rightExpression) =>
+            rightExpression != null &&
+            rightExpression.OperatorToken.IsAnyKind(AddOperators);
     }
 }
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringConcatenationInLoop.cs
```diff
@@ -71,5 +75,9 @@ public sealed class StringConcatenationInLoop
         protected override ImmutableArray<SyntaxKind> CompoundAssignmentKinds => compoundAssignmentKinds;
 
         protected sealed override Helpers.GeneratedCodeRecognizer GeneratedCodeRecognizer => Helpers.CSharp.GeneratedCodeRecognizer.Instance;
+
+        protected override bool IsAddExpression(BinaryExpressionSyntax rightExpression) =>
+            rightExpression != null &&
+            rightExpression.OperatorToken.IsAnyKind(AddOperators);
```
Evangelink: Can `OperatorToken` be null? If so is `IsAnyKind` handling null?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StringConcatenationInLoop.cs
```diff
@@ -21,6 +21,11 @@ public StringConcatenationInLoop()
             }
             s += "a";
 
+            while (true)
+            {
+                // See https://github.com/SonarSource/sonar-csharp/issues/1138
+                s = s ?? "b";
```
Evangelink: Please try to fix Vb.Net too and add a test case.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringConcatenationInLoop.cs
michalb-sonar: Cannot be, it's a struct.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StringConcatenationInLoop.cs
michalb-sonar: There is no null-coalescing operator in VB.NET. The equivalent (`If`) doesn't cause a problem.

PR #1158, opened 2018-02-08
Update to Roslyn 1.3.2 and prevent installation of NuGet on VS2015 < U3

### LINKED ISSUE #1155: Move analyzers to Roslyn 1.3.2 and drop compatibility with VS 2015 < Update 3
Moving to Roslyn 1.3.2 will allow to analyze more syntax and fix some parsing bugs while remaining compatible with Visual Studio 2015 Update 3.

This change leads to the drop of support for Visual Studio 2015 RTM, Update 1 and Update 2.

Evangelink:
Fix #1155


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
```diff
@@ -9,8 +9,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="1.0.0" />
-    <PackageReference Include="System.Collections.Immutable" Version="1.1.36" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="1.3.2" />
+    <PackageReference Include="System.Collections.Immutable" Version="1.2.0" />
```
Evangelink: @duncanp-sonar Installing the version _1.3.2_ forces the installation of _1.2.0_ of `System.Collections.Immutable` yet looking at the available dlls in VS2015 I can see that the version is _1.1.37_. When trying to downgrade I end up with a compilation error from NUGET stating I cannot downgrade dependencies. Do you know how I can fix this issue?

Not fixing it causes failures when using the NuGet or within SLVS because the version doesn't exist.

sonaranalyzer-dotnet/SonarAnalyzer.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.27130.2010
+VisualStudioVersion = 15.0.27130.2024
```
Evangelink: VS auto change.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/packages.SonarAnalyzer.VisualBasic.config
```diff
@@ -1,11 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
Evangelink: Unused file that wasn't removed before.

PR #1173, opened 2018-02-13
Allows to prevent the registration of a diagnostic on the context

### LINKED ISSUE #1146: Disabling the analyzer should not be done with the register but outside
We should test that it is possible to put the condition testing whether the rule is enabled or not outside of the register action and not inside. This will allow to improve performance by avoid a test for every callback.

Evangelink:
Fix #1146

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/DiagnosticDescriptorBuilder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/DiagnosticDescriptorBuilder.cs
@@ -104,13 +104,8 @@ namespace SonarAnalyzer.Helpers
             throw new NotSupportedException($"Not supported severity");
         }
 
-        private static DiagnosticSeverity ToDiagnosticSeverity(this Severity severity)
-        {
-            return severity.ToDiagnosticSeverity(IdeVisibility.Visible);
-        }
-
         private static DiagnosticSeverity ToDiagnosticSeverity(this Severity severity,
-            IdeVisibility ideVisibility)
+            IdeVisibility ideVisibility = IdeVisibility.Visible)
         {
             switch (severity)
             {
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
@@ -21,7 +21,6 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.Diagnostics;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
@@ -33,7 +32,7 @@ namespace SonarAnalyzer.Helpers
     /// <see cref="AnalysisContext"/> to allow for specialized control over the analyzer.
     /// Here is the list of fine-grained changes we are doing:
     /// - Avoid duplicated issues when the analyzer NuGet (SonarAnalyzer) and the VSIX (SonarLint) are installed simultaneously.
-    /// - Allow a specific kind of ruleset for SonarLint (enable/disable a rule).
+    /// - Allow a specific kind of rule-set for SonarLint (enable/disable a rule).
     /// - Prevent reporting an issue when it was suppressed on SonarQube.
     /// </summary>
     public class SonarAnalysisContext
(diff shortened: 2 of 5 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/DiagnosticDescriptorBuilder.cs
```diff
@@ -104,13 +104,8 @@ private static Severity ParseSeverity(string severity)
             throw new NotSupportedException($"Not supported severity");
         }
 
-        private static DiagnosticSeverity ToDiagnosticSeverity(this Severity severity)
```
Evangelink: I'll move this change to a separate commit.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -145,57 +152,57 @@ internal void RegisterCodeBlockStartAction<TLanguageKindEnum>(Action<CodeBlockSt
...
-                    }
-                }, symbolKinds);
+                        if (!IsAnalysisDisabled(c.GetSyntaxTree()))
+                        {
+                            action(c);
+                        }
+                    }, symbolKinds));
 
-        internal static bool IsAnalysisDisabled(SyntaxTree tree, IEnumerable<DiagnosticDescriptor> supportedDiagnostics = null)
+        private void RegisterContextAction(Action action)
```
Evangelink: Please suggest a better name!

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -145,57 +152,57 @@ internal void RegisterCodeBlockStartAction<TLanguageKindEnum>(Action<CodeBlockSt
...
-                Debug.Assert(ShouldAnalysisBeDisabled == null, "Not expecting SonarLint to set both the old and the new" +
-                    "delegates.");
+                action();
+            }
+        }
 
-                // If any of the diagnostic is enabled then allow to run the rule action BUT filter out at the time when the
-                // issue is being reported.
-                return !ShouldExecuteRuleFunc(new AnalysisRunContext(tree, supportedDiagnostics));
+        internal static bool IsAnalysisDisabled(SyntaxTree tree)
```
Evangelink: I only reverted the change but I could probably change the method to be positive.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
duncanp-sonar: I can't think of one!

duncanp-sonar: It would be worth adding a comment somewhere to say why we have such a complicated registration/suppression mechanism. The following is a pretty lengthy explanation, but it will help us with future maintenance:

For performance reasons, we don't want to register callbacks for rules that should not be run. However, we don't necessarily have enough information to make the decision up front, so the check is broken into two parts:
1) pre-registration: at this point we only know the rules that are being registered. If we can decide from the diagnostic descriptors that a rule should not be run then we won't register the callback with the analysis context.
2) post-registration: we are now being called back by the analysis context, so we now have access to the syntax tree (and project) that are being analysed, and can use that information to decide whether to execute the logic of the rule.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -145,57 +111,49 @@ internal void RegisterCodeBlockStartAction<TLanguageKindEnum>(Action<CodeBlockSt
...
+                    });
             }
-
-            // ... but for compatibility purposes we need to keep handling the old-fashioned way
-            return tree != null &&
-                ShouldAnalysisBeDisabled != null &&
-                ShouldAnalysisBeDisabled(tree);
         }
+
+        internal static bool IsRegisteredActionEnabled(SyntaxTree tree) =>
```
Evangelink: I have negated the condition, could you guys please check I did it correctly :)

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -145,57 +111,49 @@ internal void RegisterCodeBlockStartAction<TLanguageKindEnum>(Action<CodeBlockSt
...
+            ShouldExecuteRegisteredAction == null ||
+            tree == null ||
```
Evangelink: I am not sure why we want to execute the action when the syntax tree is null. I wouldn't have done that but it is there and I am not sure of the impact.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -86,56 +88,20 @@ internal SonarAnalysisContext(AnalysisContext context, IEnumerable<DiagnosticDes
...
         internal void RegisterCodeBlockAction(Action<CodeBlockAnalysisContext> action) =>
-            context.RegisterCodeBlockAction(
-                c =>
-                {
-                    if (!IsAnalysisDisabled(c.GetSyntaxTree(), supportedDiagnostics))
-                    {
-                        action(c);
-                    }
-                });
+            RegisterContextAction(context.RegisterCodeBlockAction, action, c => c.GetSyntaxTree());
```
duncanp-sonar: Nice - much neater.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
duncanp-sonar: If the tree is ever null then it means there's a bug somewhere, either in our code or in Roslyn. I guess our main concern at that point is just coding defensively so we don't throw.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
duncanp-sonar: My brain hurts.

duncanp-sonar: ...but it looks ok to me.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
Evangelink: But with this code if the tree is null then we DO call the action while for me we shouldn't.

duncanp-sonar: As long as we don't throw I think it's ok to do either.

PR #1179, opened 2018-02-16
SonarC# should compare paths using the OS case-sensitivity

### LINKED ISSUE #1174: SonarC# should compare paths using the OS case-sensitivity
### Description

SonarC# doesn't handle correctly path comparison (it currently is case sensitive even for windows).

Relates to #212 
Relates to [SONARMSBRU-375](https://jira.sonarsource.com/browse/SONARMSBRU-375)

### Repro steps

Create a new `Console Application`. Add a folder (through VS) called `foo` to the project and move the `Program.cs` class into it. Open the file explorer and rename `foo` into `FOO`.

Run the analysis. The file is indexed as `FOO\Program.cs` but contains no metrics nor issues.

### Expected behavior

Metrics and issues are displayed.

### Actual behavior

No metric nor issues.

### Known workarounds

Rename the folder.

Evangelink:
Fix #1174

### DIFF
--- a/its/projects/CasingApp/CasingApp.sln
+++ b/its/projects/CasingApp/CasingApp.sln
@@ -1,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.27130.2024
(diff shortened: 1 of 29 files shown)


Evangelink:
@henryju I'll let you see if you handle it on the common part.

PR #1191, opened 2018-02-22
Fix S3168: Rule should not be too strict on first parameter name

### LINKED ISSUE #1161: Fix S3168: Rule should not be too strict on first parameter name
### Description

Hi, 
Thank you for the great job you done still doing.. 

I have little issue with analyzer S3168, it reports false positive about events handlers has async to change it to returns Task instead of void. it's impossible to change it as you know.

### Repro steps

1. Create WinFroms project
2. Add a button to the form
3. Double click on the button.
4. Add a new method that return Task.
5. Call that method returns task from the button's event handler. and you will get the problem.

### Expected behavior

It should not report the event handler, while the analyzer is really really useful with other methods.

### Actual behavior

It reports on the event handlers that uses async.

### Known workarounds

No workaround known.

### Related information

* SonarC# Version: 6.8.1.4648
* Visual Studio Version: 15.5.5

Thank you

Evangelink:
Fix #1161

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/KnownMethods.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/KnownMethods.cs
@@ -245,6 +245,16 @@ namespace SonarAnalyzer.Helpers
                 methodSymbol.ContainingType.ConstructedFrom.Is(KnownType.System_Collections_Generic_List_T);
         }
 
+        public static bool IsProbablyEventHandler(this IMethodSymbol methodSymbol)
+        {
+            return methodSymbol.ReturnsVoid &&
+                methodSymbol.Parameters.Length == 2 &&
+                methodSymbol.Parameters[0].Type.Is(KnownType.System_Object) &&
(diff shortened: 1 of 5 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/KnownMethods.cs
```diff
@@ -245,6 +245,16 @@ public static bool IsListAddRange(this IMethodSymbol methodSymbol)
...
         }
 
+        public static bool IsProbablyEventHandler(this IMethodSymbol methodSymbol)
+        {
+            return methodSymbol.ReturnsVoid &&
+                methodSymbol.Parameters.Length == 2 &&
+                methodSymbol.Parameters[0].Type.Is(KnownType.System_Object) &&
+                // Cannot check that the 2nd parameters is or derives from EventArgs as UWP uses other kind of event args
+                // (e.g. ILeavingBackgroundEventArgs)
+                methodSymbol.Parameters[1].Type.ToString().EndsWith("EventArgs", StringComparison.Ordinal);
```
duncanp-sonar: I guess we could check that the current class ends with _EventArgs_ or one of its parent classes does.
Is it worth it?

PR #1199, opened 2018-03-01
Rule S4261: Methods should be named according to their synchronicity

### LINKED ISSUE #780: Rule S4261: Methods should be named according to their synchronicity
implement [RSPEC-4261](https://jira.sonarsource.com/browse/RSPEC-4261)

Evangelink:
FIx #780

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNameAccordingToSynchronicity.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNameAccordingToSynchronicity.cs
@@ -1,0 +1,90 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2018 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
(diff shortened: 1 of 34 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNameAccordingToSynchronicity.cs
```diff
@@ -0,0 +1,90 @@
...
+                    var methodSymbol = c.SemanticModel.GetDeclaredSymbol(methodDeclaration);
+                    if (methodSymbol == null ||
+                        methodSymbol.GetInterfaceMember() != null ||
+                        methodSymbol.GetOverriddenMember() != null)
+                    {
+                        return;
+                    }
+
+                    var isTaskReturnType = methodSymbol.ReturnType.DerivesFromAny(TaskTypes);
+                    var hasAsyncSuffix = "async".Equals(methodDeclaration.Identifier.ValueText.SplitCamelCaseToWords().LastOrDefault());
```
michalb-sonar: Why not `async ==`?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodShouldBeNameAccordingToSynchronicity.cs
```diff
@@ -0,0 +1,101 @@
+﻿using System;
```
michalb-sonar: Please add a test for local functions. Fine if they are not covered by the rule (don't think they are now), but good to have it documented.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNameAccordingToSynchronicity.cs
michalb-sonar: Why not `async ==`?

Evangelink:
Stupid analysis that doesn't support branches...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNameAccordingToSynchronicity.cs
Evangelink: Good question! I will use `==`.

Evangelink: @valhristov sadly the `SplitCamelCaseToWords` is changing the case (in the wrong way `ToLowerInvariant`)

michalb-sonar: If it was case-insensitive, we may get (granted, unlikely) FPs on things like `GigaSync`.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodShouldBeNameAccordingToSynchronicity.cs
Evangelink: They are not and they can't be (this will require C# 7 for the analyzer).

PR #1200, opened 2018-03-01
Rule S109: Magic numbers should not be used

### LINKED ISSUE #996: Rule S109: Magic numbers should not be used
### Description

A magic number is a number that comes out of nowhere, and is directly used in a statement. Magic numbers are often used, for instance to limit the number of iterations of a loops, to test the value of a property, etc.
Using magic numbers may seem obvious and straightforward when you're writing a piece of code, but they are much less obvious and straightforward at debugging time.
That is why magic numbers must be demystified by first being assigned to clearly named variables before being used.
-1, 0 and 1 are not considered magic numbers.

See [RSPEC-109](https://jira.sonarsource.com/browse/RSPEC-109)

Evangelink:
Fix #996

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
@@ -31,6 +31,7 @@ namespace Tests.Diagnostics
 
         public void Foo(int value = 42)
         {
+            var x = -1 < 1;
         }
     }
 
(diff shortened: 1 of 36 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        public int GetValue()
+        {
+            return 13; // Noncompliant
+        }
+    }
+}
+
+#pragma warning disable 1998 //async method lacks an await
```
duncanp-sonar: Additional test cases:
* -1 and 1 are not considered magic numbers
* indexers
* array declaration

PR #1201, opened 2018-03-01
Rule S1048: Destructors should not throw exceptions

### LINKED ISSUE #1131: Rule S1048: Destructors should not throw exceptions
[RSPEC-1048](https://jira.sonarsource.com/browse/RSPEC-1048)

If Finalize or an override of Finalize throws an exception, and the runtime is not hosted by an application that overrides the default policy, the runtime terminates the process and no active try/finally blocks or finalizers are executed. This behavior ensures process integrity if the finalizer cannot free or destroy resources.

valhristov:
Fix #1131


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotThrowFromDestructors.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+            }
+
+            try
+            {
+                Foo();
+            }
+            catch (Exception)
+            {
+                throw; // Compliant, it was already thrown, we are not supposed to swallow all exceptions
```
valhristov: Not 100% sure about this, would love to hear comments.

duncanp-sonar: Interesting. I agree with your comment in the test - _rethrow_ is not the same as _throw_. It's more like the destructor has decided not to squash those exceptions it was expecting.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotThrowFromDestructors.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(c =>
+                {
+                    var throwStatement = (ThrowStatementSyntax)c.Node;
+                    if (throwStatement.Expression != null && // Do not raise for rethrown exceptions
+                        throwStatement.FirstAncestorOrSelf<BaseMethodDeclarationSyntax>() is DestructorDeclarationSyntax)
```
Evangelink: Why don't you use `throwStatement.FirstAncestorOrSelf<DestructorDeclarationSyntax>() != null` instead?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotThrowFromDestructors.cs
Evangelink: I am fine with both approaches (raising and not raising).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotThrowFromDestructors.cs
```diff
@@ -0,0 +1,57 @@
...
+        private const string MessageFormat = "Remove this 'throw' statement.";
+
+        private static readonly DiagnosticDescriptor rule =
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(c =>
```
Evangelink: [cosmetic][minor] I prefer the `c =>` on the next line :)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotThrowFromDestructors.cs
valhristov: Personal preference. Also there are fewer iterations this way (when we reach method declaration we stop, otherwise we continue to the root).

PR #1202, opened 2018-03-01
Rule S1264: A 'while' loop should be used instead of a 'for' loop

### LINKED ISSUE #1172: Rule S1264: A "while" loop should be used instead of a "for" loop
Implements [RSPEC-1264](https://jira.sonarsource.com/browse/RSPEC-1264)

Evangelink:
Fix #1172


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UseWhileLoopInstead.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
+                {
+                    var forStatement = (ForStatementSyntax)c.Node;
+
+                    if (forStatement.Declaration == null &&
+                        forStatement.Incrementors.Count == 0)
```
michalb-sonar: It looks a odd that empty declaration is null and empty incrementor is a non-null empty collection...
Could `Incrementors` be null?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UseWhileLoopInstead.cs
```diff
@@ -0,0 +1,36 @@
...
+            var x = 5;
+            for (; x < 10;) // Noncompliant {{Replace this 'for' loop with a 'while' loop.}}
+//          ^^^
+            {
+            }
+
+            for (; ; ) // Noncompliant
+            {
+
+            }
```
duncanp-sonar: Other compliant cases:
* just the declaration is null
* just the incrementor is null

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UseWhileLoopInstead.cs
Evangelink: Haven't found a way but as I told you we should go defensive I am going to handle the case anyway.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UseWhileLoopInstead.cs
```diff
@@ -0,0 +1,46 @@
...
+            {
+            }
+
+            var x = 5;
+            for (; x < 10;) // Noncompliant {{Replace this 'for' loop with a 'while' loop.}}
+//          ^^^
+            {
+            }
+
+            for (; ; ) // Noncompliant
```
valhristov: Not sure if `while(true) { }` is much better than `for (;;) { }`... IMO this syntax is a matter of personal preference and could be controversial, e.g. I would probably not raise here.

Evangelink: That's the rule example.

PR #1208, opened 2018-03-02
Fix S2187: Rule should not raise False Positives for class inheriting…

### LINKED ISSUE #1196: Fix S2187: Rule should not raise False Positives for class inheriting class with tests
### Description

We are getting S2187 warnings for unit test classes that have no test directly but inherit tests from there base class.

### Repro steps

Our units tests usually follow the class inheritance of the classes that are being tested.
We then end up with some classes that have no tests of their own because the base class covers all functionality.

  [TestFixture]
  public class TestA
  {
    [Test]
    public void Test1()
    {
      // test something
    }
  }

  [TestFixture]
  public class TestB : TestA
  {
    // no test here but base class test is executed
  }

### Expected behavior

No warning.

### Actual behavior

Warning that class has no tests.

### Known workarounds

Add an additional test to the class.

### Related information

* SonarC# Version 6.8.2.4717
* Visual Studio Version 15.6.0 Preview 6.0

Evangelink:
… class with tests

Fix #1196

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestClassShouldHaveTestMethod.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestClassShouldHaveTestMethod.cs
@@ -41,17 +41,21 @@ namespace SonarAnalyzer.Rules.CSharp
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
 
-        private static readonly ISet<KnownType> HandledTestClassAttributes = ImmutableHashSet.Create(
+        private static readonly ISet<KnownType> HandledTestClassAttributes = new HashSet<KnownType>
+        {
             KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_TestClassAttribute,
-            KnownType.NUnit_Framework_TestFixtureAttribute);
+            KnownType.NUnit_Framework_TestFixtureAttribute
+        };
 
-        private static readonly ISet<KnownType> HandledTestMethodAttributes = ImmutableHashSet.Create(
+        private static readonly ISet<KnownType> HandledTestMethodAttributes = new HashSet<KnownType>
+        {
             KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_TestMethodAttribute,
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestClassShouldHaveTestMethod.cs
```diff
@@ -1,86 +1,101 @@
...
+    public sealed class TestClassShouldHaveTestMethod : SonarDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S2187";
+        private const string MessageFormat = "Add some tests to this class.";
+
+        private static readonly DiagnosticDescriptor rule =
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+        private static readonly ISet<KnownType> HandledTestClassAttributes = new HashSet<KnownType>
```
Evangelink: Changed to `HashSet` for consistency with other classes and for performance reasons.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestClassShouldHaveTestMethod.cs
```diff
@@ -1,86 +1,101 @@
...
+        {
+            KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_TestClassAttribute,
+            KnownType.NUnit_Framework_TestFixtureAttribute
+        };
+
+        private static readonly ISet<KnownType> HandledTestMethodAttributes = new HashSet<KnownType>
```
Evangelink: Changed to `HashSet` for consistency with other classes and for performance reasons.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestClassShouldHaveTestMethod_MSTest.cs
```diff
@@ -0,0 +1,70 @@
...
+        public void Foo(int i) { }
+    }
+
+    [TestClass]
+    public abstract class MyCommonCode1
+    {
+    }
+
+    [TestClass]
+    public class MySubCommonCode1 : MyCommonCode1 // Noncompliant
```
Evangelink: We could decide to ignore this case but for now I am forcing the base class to have tests.

PR #1209, opened 2018-03-02
Rule S1151: "switch case" clauses should not have too many lines of code

### LINKED ISSUE #1170: Rule S1151: "switch case" clauses should not have too many lines of code
Implements [RSPEC-1151](https://jira.sonarsource.com/browse/RSPEC-1151)

Evangelink:
Fix #1170

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SwitchSectionShouldNotHaveTooManyStatements.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SwitchSectionShouldNotHaveTooManyStatements.cs
@@ -42,7 +42,7 @@ namespace SonarAnalyzer.Rules.CSharp
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
 
-        private const int ThresholdDefaultValue = 5;
+        private const int ThresholdDefaultValue = 8;
         [RuleParameter("max", PropertyType.Integer, "Maximum number of lines of code.", ThresholdDefaultValue)]
         public int Threshold { get; set; } = ThresholdDefaultValue;
 
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SwitchSectionShouldNotHaveTooManyStatements.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SwitchSectionShouldNotHaveTooManyStatements.cs
@@ -1,0 +1,77 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2018 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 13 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SwitchSectionShouldNotHaveTooManyStatements_DefaultValue.cs
```diff
@@ -0,0 +1,61 @@
...
+                                Console.WriteLine();
+                                Console.WriteLine();
+                            }
+                        }
+                    }
+                case 5: // Noncompliant
+                    var x = Foo(42);
+                    break;
+
+                    string Foo(int value)
```
Evangelink: Not sure what to do with local functions. Shall we ignore them as we recommend to use a method? Shall I skip the function declaration (e.g. `string Foo(int value)`) from the count?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SwitchSectionShouldNotHaveTooManyStatements.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        protected override void Initialize(ParameterLoadingAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
+                {
+                    var switchSection = (SwitchSectionSyntax)c.Node;
+
+                    if (switchSection.IsMissing ||
+                        switchSection.Labels.Count <= 0)
```
valhristov: How could this be less than 0?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SwitchSectionShouldNotHaveTooManyStatements_DefaultValue.cs
```diff
@@ -0,0 +1,61 @@
...
+                case 1: // Noncompliant {{Reduce this 'switch/case' number of lines from 6 to at most 5, for example by extracting code into a method.}}
+//              ^^^^^^^
+                    Console.WriteLine("1");
+                    Console.WriteLine("2");
+                    Console.WriteLine("3");
+                    Console.WriteLine("4");
+                    Console.WriteLine("5");
+                    break;
+                case 2:
+                    {
```
valhristov: Do we count blocks as one line? If yes, this test should be negative.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SwitchSectionShouldNotHaveTooManyStatements_DefaultValue.cs
valhristov: The rule is about readability - blocks with too many lines are difficult to read, we should not ignore locals

valhristov:
But I guess you need to run the QA before merging :)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SwitchSectionShouldNotHaveTooManyStatements.cs
Evangelink: This shouldn't be... Just playing defensive as usual.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SwitchSectionShouldNotHaveTooManyStatements_DefaultValue.cs
Evangelink: The statements collection here will return only one statement, the block but as you said this is about readability so I thought I should expand blocks hence this count. Shall I change this?

PR #1227, opened 2018-03-12
Fix S3900: Rule should not raise an issue when compared with IsNullOr…

### LINKED ISSUE #693: Fix S3900: Rule should not raise an issue when compared with IsNullOrWhitespace/IsNullOrEmpty
### Description

The rule gives a false positive, indicating that a string variable hasn't been tested for null befor its use. BUT the variable is tested, just not with an `x == null` but with a `string.IsNullOrWhiteSpace()` or with a `string.IsNullOrEmpty()`. 

### Repro steps
```
public void a(string b)
{
     if (string.IsNullOrWhiteSpace(b))  return;
     var c = b.Split('_');
     Console.WriteLine(c.Count());
}
```
### Expected behavior

It souldn't fire the rule because the string b is being checked for null before it's use.

### Actual behavior

False positive is reported on line `var c = b.Split('_');`, marking b as being used without previos check.

### Known workarounds

First check for null explicitly as in `if (b == null) return;` and THEN check for IsNullOrWhiteSpace, which effectively beats the whole purpose of the nice string function IsNullOrWhiteSpace

### Related information

* SonarC# Version: SonarLint for Visual Studio 3.4.0.1732
* Visual Studio Version : 15.2 (26430.16)

### LINKED ISSUE #826: Fix S3900: Arguments of public methods should be validated against null should take guarding methods into account
A lot developers use Guarding to do their parameter checking.

The implementation that was once introduced by Microsoft itself (CA1062) supported this quite nicely by code that could look like the following:

``` C#
/// <summary>Supplies input parameter guarding.</summary>
public static class Guard
{
    /// <summary>Guards the parameter if not null, otherwise throws an argument (null) exception.</summary>
    /// <typeparam name="T">The type to guard; cannot be a structure.</typeparam>
    /// <param name="parameter">The parameter to guard.</param>
    /// <param name="paramName">The name of the parameter.</param>
    [DebuggerStepThrough]
    public static T NotNull<T>([ValidatedNotNull]T parameter, string paramName) where T : class
    {
        if (ReferenceEquals(parameter, null))
        {
            throw new ArgumentNullException(paramName);
        }
        return parameter;
    }
}

/// <summary>Marks the NotNull argument as being validated for not being null, to satisfy the static code analysis.</summary>
/// <remarks>
/// Notice that it does not matter what this attribute does, as long as
/// it is named ValidatedNotNullAttribute.
/// </remarks>
[AttributeUsage(AttributeTargets.Parameter)]
sealed class ValidatedNotNullAttribute : Attribute { }
```
So, if the parameter of the guarding method is decorated with an attribute with the name 'ValidatedNotNullAttribute' it is assumed to have done so.

Could this be added?

Also see https://esmithy.net/2011/03/15/suppressing-ca1062/

valhristov:
…Whitespace/IsNullOrEmpty

Fix #693 
Fix #826

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/InvocationVisitor.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/InvocationVisitor.cs
@@ -20,6 +20,7 @@
 
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -71,6 +72,12 @@ namespace SonarAnalyzer.SymbolicExecution
             if (IsStringNullCheckMethod(methodSymbol))
             {
                 return HandleStringNullCheckMethod();
+            }
+
+            if (methodSymbol != null &&
+                ValidatesNotNull(methodSymbol, out var validatedArgumentIndex))
+            {
+                return HandleNullValidationMethod(validatedArgumentIndex, invocationArgsCount);
             }
 
             if (invocation.IsNameof(semanticModel))
@@ -148,6 +155,13 @@ namespace SonarAnalyzer.SymbolicExecution
             return SetConstraintOnValueEquals(equals, newProgramState);
         }
 
+        private ProgramState HandleNullValidationMethod(int validatedArgumentIndex, int invocationArgsCount) =>
+            programState
+                .PopValues(invocationArgsCount - validatedArgumentIndex - 1)
+                .PopValue(out var guardedArgumentValue)
+                .PopValues(validatedArgumentIndex)
+                .SetConstraint(guardedArgumentValue, ObjectConstraint.NotNull);
+
         private static readonly ISet<string> IsNullMethodNames = new HashSet<string>
         {
             nameof(string.IsNullOrEmpty),
@@ -183,6 +197,19 @@ namespace SonarAnalyzer.SymbolicExecution
                 methodSymbol.ContainingType.Is(KnownType.System_Object) &&
                 methodSymbol.IsStatic &&
                 methodSymbol.Name == EqualsLiteral;
+        }
+
+        private static bool ValidatesNotNull(IMethodSymbol methodSymbol, out int validatedArgumentIndex)
+        {
+            validatedArgumentIndex = methodSymbol.Parameters.IndexOf(IsValidatedParameter);
+
+            return validatedArgumentIndex >= 0;
+
+            bool IsValidatedParameter(IParameterSymbol parameterSymbol) =>
+                parameterSymbol.GetAttributes().Any(IsValidatedNotNullAttribute);
+
+            bool IsValidatedNotNullAttribute(AttributeData attribute) =>
+                "ValidatedNotNullAttribute".Equals(attribute.AttributeClass?.Name);
         }
 
         internal static ProgramState SetConstraintOnValueEquals(ValueEqualsSymbolicValue equals,
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/InvocationVisitor.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/InvocationVisitor.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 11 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/InvocationVisitor.cs
```diff
@@ -148,9 +148,11 @@ private ProgramState HandleInstanceEqualsCall()
             return SetConstraintOnValueEquals(equals, newProgramState);
         }
 
-        private static readonly ImmutableHashSet<string> IsNullMethodNames = ImmutableHashSet.Create(
+        private static readonly ISet<string> IsNullMethodNames = new HashSet<string>
```
valhristov: Standard perf improvement, consistent with the rest of the code

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PublicMethodArgumentsShouldBeCheckedForNull.cs
```diff
@@ -64,5 +64,26 @@ public void CompliantCases(bool b, object o1, object o2, object o3, object o4, E
 
             Program.staticField.ToString(); // Compliant
         }
+
+        public void MoreCompliantCases(string s1, string s2)
+        {
+            if (string.IsNullOrEmpty(s1))
```
duncanp-sonar: Are there any tests for e.g. _!string.IsNullOrEmpty(s1)_ ?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/InvocationVisitor.cs
```diff
@@ -148,9 +155,18 @@ private ProgramState HandleInstanceEqualsCall()
             return SetConstraintOnValueEquals(equals, newProgramState);
         }
 
-        private static readonly ImmutableHashSet<string> IsNullMethodNames = ImmutableHashSet.Create(
+        private ProgramState HandleNullValidationMethod(int validatedArgumentIndex, int invocationArgsCount) =>
+            programState
+                .PopValues(invocationArgsCount - validatedArgumentIndex - 1)
+                .PopValue(out var guardedArgumentValue)
+                .PopValues(validatedArgumentIndex)
+                .SetConstraint(guardedArgumentValue, ObjectConstraint.NotNull);
```
valhristov: Set NotNull on the argument that has ValidatedNotNull attribute, ignore the rest

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/EnumerableExtensions.cs
```diff
@@ -100,5 +100,11 @@ public static IEnumerable<T> WhereNotNull<T>(this IEnumerable<T> enumerable)
                 }
             }
         }
+
+        public static int IndexOf<T>(this IEnumerable<T> enumerable, Func<T, bool> predicate) =>
+            enumerable
+                .Select((item, index) => new { item, index })
+                .FirstOrDefault(x => predicate(x.item))
+                ?.index ?? -1;
```
valhristov: I don't like this ugly code in InvocationVisitor, hence the extension method

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PublicMethodArgumentsShouldBeCheckedForNull.cs
valhristov: It should not be required, as the `!` is not part of the invocation. The negation is handled in a separate node of the exploded graph.

valhristov:
In general the invalid syntax tests should not be required as the graph is walked instruction by instruction and if something is null, it should have been handled already.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PublicMethodArgumentsShouldBeCheckedForNull.cs
Evangelink: I think it is still useful to add a test to be sure this is covered. If not it's probably not needed to cover it right now.

michalb-sonar:
I added `if (string.IsNullOrEmpty()) { }` to line 782 in `ConditionEvaluatesToConstant.cs` test and got the following error:
`Message: Issue with message '' not expected on line 781`

Not sure if we care much to fix it, but at least documenting this in test would be good.

valhristov:
@michalb-sonar are you sure you didn't mess with the secondary locations? The rule reports secondary location when there is unreachable code and usually "Issue with message '' not expected" happens when a secondary location is not configured correctly. Adding invalid code works in a separate test case...

michalb-sonar:
@valhristov quite possible possible, I only had a very quick look - probably missed that test with invalid code.

PR #1228, opened 2018-03-14
Do you plan to release an open source VB.NET Plugin?

karoldeland:
Releasing all the VB.NET stuff in the sonaranalyzer-dotnet projects makes it easy to create a VB.NET plugin. I made a sample here. Do you plan to open source a VB.NET plugin?

Thanks!


valhristov:
Thank you for the PR, it is impressive work. Obviously we cannot accept it, but indeed we do have this idea of opening more language plugins, including VB.NET. It is in relatively early review stage and I cannot say if and when it will happen, so stay tuned.

Actually, the VB.NET analyzers have always been open source and free, because you could easily use them with SonarLint without a license, it is just the SonarQube integration that requires a license.

PR #1229, opened 2018-03-14
Fix S3264: FP when using add/remove syntax

### LINKED ISSUE #1123: Fix S3264: Rule should not raise when 'BeginInvoke' or 'EndInvoke' are called
### Description

The issue `S3264: Remove this unused event or invoke it.` is raised when using `BeginInvoke` and/or `EndInvoke` while it shouldn't.

See the [Google Group Thread](https://groups.google.com/d/topic/sonarlint/PZgSeyXjLCo/discussion)

### Related information

* SonarC# Version: SonarAnalyzer.CSharp 6.7.1.4347
* Visual Studio Version: VS 2017
* SonarLint for Visual Studio 2017 3.8.1.2823

### LINKED ISSUE #1219: Fix S3264: FP when using add/remove syntax
### Description

Rule raises FP when using the add/remove syntax of an `event`.

### Repro steps

```
public class Foo
{
    private EventHandler someEvent;
    public event EventHandler SomeEvent // Noncompliant - should be compliant
    {
        add
        {
            someEvent += value;
        }
        remove
        {
            someEvent -= value;
        }
    }

    public DoRaise()
    {
        someEvent?.Invoke(this, EventArgs.Empty);
    }
}
```

### Expected behavior

Issue not raised because the event is raised using the private handler.

### Actual behavior

Issue is raised because we don't use the `event` `Invoke` whereas it is impossible to do so.

### Known workarounds

None.

### Related information

* SonarC# Version 7.0
* Visual Studio Version 2015/2017

valhristov:
Removed the check for EventDeclarationSyntax as it cannot be invoked. One can only invoke EventFieldDeclarationSyntax but checking whether the event declaration is using the _correct_ field is a different rule IMO. S3237 will report if the value of the accessor is not used (e.g. if you don't assign an event field).

Refactored the implementation a bit to make it more functional and somewhat easier to understand.

Fix #1219 
Fix #1123

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UninvokedEventDeclaration.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UninvokedEventDeclaration.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
@@ -161,8 +162,9 @@ namespace SonarAnalyzer.Rules.CSharp
                             SemanticModel = container.SemanticModel,
                             Symbol = container.SemanticModel.GetSymbolInfo(node).Symbol as IMethodSymbol
                         }))
-                 .Where(tuple => tuple.Symbol != null &&
-                    tuple.Symbol.MethodKind == MethodKind.DelegateInvoke);
+                 .Where(tuple =>
+                    tuple.Symbol != null &&
+                    IsDelegateInvocation(tuple.Symbol));
 
             var invokedEventSymbols = delegateInvocations
                 .Select(tuple =>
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UninvokedEventDeclaration.cs
```diff
@@ -60,41 +65,41 @@ private void RaiseOnUninvokedEventDeclaration(SymbolAnalysisContext context)
 
             var removableDeclarationCollector = new RemovableDeclarationCollector(namedType, context.Compilation);
 
-            var removableEvents = removableDeclarationCollector.GetRemovableDeclarations(
-                ImmutableHashSet.Create(SyntaxKind.EventDeclaration), maxAccessibility);
-            var removableEventFields = removableDeclarationCollector.GetRemovableFieldLikeDeclarations(
-                ImmutableHashSet.Create(SyntaxKind.EventFieldDeclaration), maxAccessibility);
```
valhristov: We found that ImmutableHashSet is a bit slower than HashSet before, hence I switched to HashSet.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UninvokedEventDeclaration.cs
```diff
@@ -93,6 +93,25 @@ public void M()
...
+
+        public event EventHandler PublicEvent
+        {
+            add { privateEvent += value; }
+            remove { privateEvent -= value; }
+        }
+
+        protected void OnEvent1()
+        {
+            PublicEvent += (s, o) => { }; // Need to use the event to be able to trigger issues
```
valhristov: IF the event declaration is not used we don't report, hence we need this line.

Evangelink: If that's not used do you know if we have another rule triggering (like unused member or something like that)? I actually think we should always report because this rule is about dead code so not subscribing + not raising = issue.

WDYT?

valhristov:
@Evangelink, the unused member will raise issue when the event is not used.

PR #1232, opened 2018-03-14
Fix S2971: Do not trigger on ToList().AsEnumerable()

### LINKED ISSUE #1149: S2971 should not be triggered on ToList().AsEnumerable()
### Description

`var myEnumerable = someEnumerable.ToList().AsEnumerable();` should not trigger S2971 as the AsEnumerable() here is used to cast the result to `IEnumerable<T>`

### Repro steps

Write some code where `ToList()` or `ToArray()` is followed by `.AsEnumerable()`

### Expected behavior

S2971 is not riased.

### Actual behavior

S2971 is raised.

### Known workarounds

None.

### Related information

* SonarAnalyzer.CSharp 6.7.1.4347 NuGet Package
* Visual Studio 2017

valhristov:
Fix #1149

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CollectionQuerySimplification.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CollectionQuerySimplification.cs
@@ -46,23 +46,36 @@ namespace SonarAnalyzer.Rules.CSharp
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
 
-        private static readonly ISet<string> MethodNamesWithPredicate = ImmutableHashSet.Create(
+        private static readonly ISet<string> MethodNamesWithPredicate = new HashSet<string>
+        {
             "Any", "LongCount", "Count",
             "First", "FirstOrDefault", "Last", "LastOrDefault",
-            "Single", "SingleOrDefault");
+            "Single", "SingleOrDefault",
+        };
 
-        private static readonly ISet<string> MethodNamesForTypeCheckingWithSelect = ImmutableHashSet.Create(
+        private static readonly ISet<string> MethodNamesForTypeCheckingWithSelect = new HashSet<string>
+        {
             "Any", "LongCount", "Count",
             "First", "FirstOrDefault", "Last", "LastOrDefault",
-            "Single", "SingleOrDefault", "SkipWhile", "TakeWhile");
+            "Single", "SingleOrDefault", "SkipWhile", "TakeWhile",
+        };
 
-        private static readonly ISet<string> MethodNamesToCollection = ImmutableHashSet.Create(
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CollectionQuerySimplification.cs
```diff
@@ -131,6 +144,11 @@ private static void CheckToCollectionCalls(SyntaxNodeAnalysisContext context)
 
         private static bool MethodExistsOnIEnumerable(IMethodSymbol methodSymbol, SemanticModel semanticModel)
         {
+            if (IgnoredMethodNames.Contains(methodSymbol.Name))
```
valhristov: That's the actual change

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CollectionQuerySimplification.cs
```diff
@@ -46,23 +46,36 @@ public class CollectionQuerySimplification : SonarDiagnosticAnalyzer
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
 
-        private static readonly ISet<string> MethodNamesWithPredicate = ImmutableHashSet.Create(
+        private static readonly ISet<string> MethodNamesWithPredicate = new HashSet<string>
```
valhristov: Converting ImmutableHasSet to HashSet (several times)

Evangelink: I think it is worth doing a separate commit changing all the `ImmutableHashSet` instances to `HashSet` to avoid doing this so often.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CollectionQuerySimplification.cs
```diff
@@ -38,6 +38,8 @@ public CollectionQuerySimplification(List<object> coll)
 
             x = Enumerable.Select(coll, element => element as object).Any(element => element != null); //Noncompliant
             x = Enumerable.Any(Enumerable.Select(coll, element => element as object), element => element != null); //Noncompliant
+
+            coll.ToList().AsEnumerable(); // Compliant, we ignore AsEnumerable() as it is somewhat cleaner way to cast to IEnumerable<T> and has no side effects
```
Evangelink: Please add more tests with `AsEnumerable` in different locations.

valhristov: It should not be needed - the rule raises when ToList() is called before other Linq method.

PR #1235, opened 2018-03-15
Rule S4457: Parameter validation in 'async'/'await' methods should be…

### LINKED ISSUE #1164: Rule S4457: Parameter validation in "async"/"await" methods should be wrapped
[RSPEC-4457](https://jira.sonarsource.com/browse/RSPEC-4457)

Because of the way async/await methods are rewritten by the compiler, any exceptions thrown during the parameters check will happen only when the task is observed. That could happen far away from the source of the buggy code or never happen for fire-and-forget tasks.

Evangelink:
… wrapped

Fix #1164


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterValidationInAsyncShouldBeWrapped.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Tests.Diagnostics
+{
+    public class InvalidCases
+    {
+        public static async Task SkipLinesAsync(this TextReader reader, int linesToSkip) // Noncompliant {{Split this method into two, one handling parameters check and the other handling the iterator.}}
```
michalb-sonar: It would be useful to point to where the exceptions are thrown (in secondary locations).

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterValidationInAsyncShouldBeWrapped.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System;
```
michalb-sonar: Usual annoying comment: please add tests for local functions.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParameterValidationInAsyncShouldBeWrapped.cs
```diff
@@ -0,0 +1,72 @@
...
+                {
+                    var awaitExpression = (AwaitExpressionSyntax)c.Node;
+
+                    var methodDeclaration = awaitExpression.FirstAncestorOrSelf<MethodDeclarationSyntax>();
+                    if (methodDeclaration == null)
+                    {
+                        return;
+                    }
+
+                    var hasAnyArgumentException = methodDeclaration.DescendantNodes()
```
michalb-sonar: We could check if the exceptions are actualy thrown. Judging by ITs, maybe it's not necessary though. Your call.

Evangelink: I won't bother with this check for now and will wait for peach results.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterValidationInAsyncShouldBeWrapped.cs
Evangelink: I have added a test but I am not sure what should be the behavior.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterValidationInAsyncShouldBeWrapped.cs
```diff
@@ -25,19 +27,36 @@ public class InvalidCases
...
             if (sender == null)
             {
-                throw new ArgumentNullException(nameof(sender));
+                throw new ArgumentNullException(nameof(sender)); // Secondary
             }
 
             await Task.Yield();
         }
+
+        public static Task SkipLinesAsync(this TextReader reader, int linesToSkip) // Noncompliant - Not sure what should be the expected result (i.e. do local functions suffer from this behavior?)
```
michalb-sonar: That is nice. Perhaps one more test when the exceptions ale only thrown in local functions and not in the method?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterValidationInAsyncShouldBeWrapped.cs
michalb-sonar: That seems like a nice pattern if outer method is sync (and validates args) and inner function is async - does exactly what we want and to the outside world still looks like one method.

To support that, we'd need to check if async local function does any argument checking and is inside sync method. Up to you if you want to handle that (could be in a separate ticket).

At minimum, let's document (in test) the behaviour of that case. Pseudo-code:
```
public void Foo(int x)
{
    async localFoo(int y)
    {
        if (y == 0) throw new ArgumentNullException("y");
    }
}
```

PR #1238, opened 2018-03-15
Fix S2933: issue should not be raised for types marked with [Serializ…

### LINKED ISSUE #1210: Fix S2933: issue should not be raised for types marked with [Serializable]
### Description

Rule S2933 checks for fields that can be made readonly. However, for types to be serializable, fields to be serialized must not be readonly. So this produces false positives for fields in types marked [Serializable] if the field is not marked with [NonSerialized].

Remark: rule S3925 should IMHO also warn a user for readonly fields, even if the class is just marked [Serializable] and does not implement ISerializable.

### Repro steps

Create a class marked as [Serializable] similar to the following:

```csharp
[Serializable]
class Person
{
    int _birthYear;  // compliant
    [NonSerialized]
    int _notSerializedValue;  // Noncompliant
    Person(int birthYear, int notSerializedValue)
    {
        _birthYear = birthYear;
        _notSerializedValue = notSerializedValue;
    }
}
```

### Expected behavior

Rule 2933 produces a warning for the "_notSerializedValue", but no warning for "_birthYear" as this field must not be readonly for deserialization to work.

### Actual behavior

Rule 2933 produces a warning for both fields "_birthYear" and "_notSerializedValue".

### Known workarounds

None.

### Related information

* SonarC# Version = 6.7.1
* Visual Studio Version = VS 2017

valhristov:
…able]

Fix #1210

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldBeReadonly.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldBeReadonly.cs
@@ -171,4 +171,25 @@ namespace Tests.Diagnostics
             var x = teststring;
         }
     }
+
+    [Serializable]
+    public class SerializableClass
+    {
+        private int field; // Compliant, containing class is marked with [Serializable]
+
+        public SerializableClass()
+        {
+            field = 5;
+        }
+    }
+
+    public class DerivedFromSerializable : SerializableClass
+    {
+        private int otherField; // Noncompliant, Serializable attribute is not inherited
+
+        public DerivedFromSerializable()
+        {
+            otherField = 5;
+        }
+    }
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldBeReadonly.cs
```diff
@@ -171,4 +171,15 @@ public void Do()
...
+    [Serializable]
+    public class SerializableClass
+    {
+        private int field; // Compliant, containing class is marked with [Serializable]
+
+        public SerializableClass()
+        {
+            field = 5;
+        }
+    }
```
duncanp-sonar: Corner case that would be trivial to add a test for: [Serializable] is not inherited so the rule should raise for child classes.

PR #1240, opened 2018-03-15
Fix S1168: "Return empty collection" should not raise when the method…

### LINKED ISSUE #1037: Fix S1168: "Return empty collection" should not raise when the method return type is XmlNode
### Description

<pre><code>
        public static XmlNode TryGetSingleChildByName(XmlNode xmlNode, string elementName)
        {
            IList<XmlNode> elements = GetChildNodesMatchingName(xmlNode, elementName);
            switch (elements.Count)
            {
                case 0:
                    {
                        return null;
                    }
                case 1:
                    {
                        return elements[0];
                    }

                default:
                    {
                        throw new InvalidOperationException();
                    }
            }
        }
</code></pre>

triggers

> warning S1168: Return an empty collection instead of null

### Repro steps

Either open a class containing the above code snippet in Visual Studio or run an analysis on one such file.

### Expected behavior

S1168 should not be triggered as the return type of the method is not a collection.

### Actual behavior

S1168 is triggered.

### Known workarounds

N/A

### Related information

* SonarC# 6.6 (build 3969)
* Visual Studio 2017 (15.5.2) (also triggered if having https://www.sonarlint.org/visualstudio/ installed, version 3.8.1.2823 )
* Scanner for MSBuild: 4.0.2.892

valhristov:
… return type is XmlNode

Fix #1037

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReturnEmptyCollectionInsteadOfNull.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReturnEmptyCollectionInsteadOfNull.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2018 SonarSource SA
  * mailto: contact AT sonarsource DOT com
@@ -45,6 +45,11 @@ namespace SonarAnalyzer.Rules.CSharp
         {
             KnownType.System_Collections_IEnumerable,
             KnownType.System_Array
+        };
+
+        private static readonly ISet<KnownType> IgnoredTypes = new HashSet<KnownType>
+        {
+            KnownType.System_Xml_XmlNode
         };
 
         protected override void Initialize(SonarAnalysisContext context)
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReturnEmptyCollectionInsteadOfNull.cs
```diff
@@ -103,7 +108,8 @@ private static bool IsReturningCollection(SyntaxNodeAnalysisContext context)
 
             return methodSymbol != null &&
                 !methodSymbol.ReturnType.Is(KnownType.System_String) &&
-                methodSymbol.ReturnType.DerivesOrImplementsAny(CollectionTypes);
+                methodSymbol.ReturnType.DerivesOrImplementsAny(CollectionTypes) &&
+                !methodSymbol.ReturnType.DerivesOrImplementsAny(IgnoredTypes);
```
Evangelink: [minor] I would use only `DerivesFromAny` to avoid the extra check if implements when it doesn't derive as for now the collection contains only a class.

PR #1247, opened 2018-03-19
Rule S4456: Parameter validation in yielding methods should be wrapped

### LINKED ISSUE #1165: Rule S4456: Parameter validation in yielding methods should be wrapped
[RSPEC-4456](https://jira.sonarsource.com/browse/RSPEC-4456)

Because of the way yield methods are rewritten by the compiler (they become lazily evaluated state machines) any exceptions thrown during the parameters check will happen only when the collection is iterated over. That could happen far away from the source of the buggy code.

Evangelink:
Fix #1165


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParameterValidationInMethodShouldBeWrapped.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.Common;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    public abstract class ParameterValidationInMethodShouldBeWrapped<TRegisterExpressionSyntax> : SonarDiagnosticAnalyzer
```
valhristov: We usually create a non-generic class to contain the constants. It is not a big deal of you don't do it, but for consistency I would recommend changing the code.

PR #1249, opened 2018-03-20
Rule S3400: Methods should not return constants

### LINKED ISSUE #1072: Rule S3400: Methods should not return constants
[RSPEC-3400](https://jira.sonarsource.com/browse/RSPEC-3400)

There's no point in forcing the overhead of a method call for a method that always returns the same constant value. Even worse, the fact that a method call must be made will likely mislead developers who call the method thinking that something more is done. Declare a constant instead.

Evangelink:
Fix #1072


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodShouldNotOnlyReturnConstant.cs
```diff
@@ -0,0 +1,79 @@
+﻿namespace Tests.Diagnostics
+{
+    public class Program
+    {
+        public int GetInt() // Noncompliant {{Remove this method and declare a constant for this value.}}
+//                 ^^^^^^
+        {
+            return 12;
```
valhristov: Does it raise when you do `return 3 * 2;`?

Evangelink: It doesn't raise.

PR #1251, opened 2018-03-20
Rule S4211: Members should not have conflicting transparency annotations

### LINKED ISSUE #1057: Rule S4211: Members should not have conflicting transparency annotations
[RSPEC-4211](https://jira.sonarsource.com/browse/RSPEC-4211)

Transparency attributes, SecurityCriticalAttribute and SecuritySafeCriticalAttribute are used to identify code that performs security-critical operations. The second one indicates that it is safe to call this code from transparent, while the first one does not. Since the transparency attributes of code elements with larger scope take precedence over transparency attributes of code elements that are contained in the first element a class, for instance, with a SecurityCriticalAttribute can not contain a method with a SecuritySafeCriticalAttribute.

Evangelink:
Fix #1057


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MemberShouldNotHaveConflictingTransparencyAttributes.cs
```diff
@@ -0,0 +1,124 @@
...
+                                // All parts declaring the 'SecuritySafeCriticalAttribute' are incorrect since the assembly
+                                // itself is marked as 'SecurityCritical'.
+                                foreach (var item in nodesWithSecuritySafeCritical)
+                                {
+                                    cac.ReportDiagnosticWhenActive(Diagnostic.Create(rule, item.Value.GetLocation(),
+                                        additionalLocations: new[] { assemblySecurityLocation }));
+                                }
+                            }
+                            else
+                            {
```
valhristov: This block should not be difficult to simplify.

PR #1252, opened 2018-03-21
Rule S1192: String literals should not be duplicated

### LINKED ISSUE #1171: Rule S1192: String literals should not be duplicated
Implements [RSPEC-1192](https://jira.sonarsource.com/browse/RSPEC-1192)

Evangelink:
Fix #1171


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StringLiteralShouldNotBeDuplicated.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public void Validate(object foobar)
+        {
+            if (o == null)
+            {
+                throw new ArgumentNullException("foobar"); // Compliant - matches one of the parameter name
+            }
+
+            Do("foobar"); // Compliant - matches one of the parameter name
+        }
```
duncanp-sonar: Additional test cases:
* case-sensitivity i.e. show that strings comparisons are case-sensitive
* repeating strings below the threshold doesn't raise an issue
* local functions ;-)

PR #1253, opened 2018-03-22
Rule S4462: Calls to 'async' methods should not be blocking

### LINKED ISSUE #1159: Rule S4462: Calls to 'async' methods should not be blocking
### Description

Code base has uses of `.Result` when it should be using `await` to [prevent deadlocks](https://msdn.microsoft.com/en-us/magazine/jj991977.aspx).

I can't find a rule anywhere that can pick this up. Is it something that fits within the scope of this project, or should it be a custom rule?

### Repro steps

1. Analyse code with async calls using `.Result`

2. Review analysis

### Expected behavior

Bug or code smell found where it is recommended to use `await` instead of `.Result`.

### Actual behavior

No issue found.

### Known workarounds

Writing a custom rule may be a workaround.

### Related information

* SonarC# Version: unknown (I don't have access)
* Visual Studio Version: 2017
* If running through the Scanner for MSBuild, its version: 3.0.2

Evangelink:
Fix #1159

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
@@ -1,0 +1,154 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2018 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarAnalyzer.Common;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public sealed class CallToAsyncMethodShouldNotBeBlocking : SonarDiagnosticAnalyzer
+    {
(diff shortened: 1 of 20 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -0,0 +1,154 @@
...
+                !DeadlockPreventingConstructs.ContainsKey(memberAccess.Name.Identifier.ValueText))
+            {
+                return false;
+            }
+
+            var memberAccessSymbol = semanticModel.GetSymbolInfo(memberAccess).Symbol?.ContainingType?.ConstructedFrom;
+            return memberAccessSymbol.Is(DeadlockPreventingConstructs[memberAccess.Name.Identifier.ValueText]);
+        }
+
+        private static string[] GetMessageArguments(string memberAccessName)
```
valhristov: This could probably become a `Dictionary<string, string[]>`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -0,0 +1,154 @@
...
+                case "GetResult":
+                    return new[] { "Task.GetAwaiter.GetResult", "await" };
+                case "WaitAny":
+                    return new[] { "Task.WaitAny", "await Task.WhenAny" };
+                case "WaitAll":
+                    return new[] { "Task.WaitAll", "await Task.WhenAll" };
+                case "Sleep":
+                    return new[] { "Thread.Sleep", "await Task.Delay" };
+                default:
+                    return new string[0];
```
valhristov: I think the `default` clause will throw if executed (string.Format expects two arguments, we provide only one); if it is not needed perhaps we should remove it, or explicitly throw...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -0,0 +1,154 @@
...
+                {
+                    if (c.IsTest())
+                    {
+                        return;
+                    }
+
+                    var simpleMemberAccess = (MemberAccessExpressionSyntax)c.Node;
+                    var memberAccessNameName = simpleMemberAccess.Name?.Identifier.ValueText;
+
+                    if (memberAccessNameName == null ||
```
valhristov: The logic in this and the next if statement is difficult to follow. I think it could/should be simplified.

dibley1973:
Hi, Please can anyone confirm if this rule is now implemented, as **RSPEC-4462** looks to be still unresolved. [RSPEC-4462 Calls to "async" methods should not be blocking](https://jira.sonarsource.com/browse/RSPEC-4462)

Although above we can see that the chaneg was merged to `master`...

>  Evangelink merged commit [3e3cca5](https://github.com/SonarSource/sonar-dotnet/commit/3e3cca5f9657be62e09c7afeeaaa2e9ffe688bc6) into master on Mar 27, 2018

andrei-epure-sonarsource:
it's part of the analyzer since 7.1, @dibley1973 - https://rules.sonarsource.com/csharp/RSPEC-4462 . However, it's not part of the default SonarWay quality profile, so you need to manually activate it from SonarQube

the RSPEC is the rule specification, and normally we don't close specifications. As long as a specification is still relevant, it is active

dibley1973:
@andrei-epure-sonarsource - Ahhh. Thank you for quick the clarification. :)

Issue #817, opened 2017-10-10
Update test framework: refactor the Verifier to allow assertions on Diagnostic.Properties


Evangelink:
This refactoring would allow us to write some specific tests about the actual content of the `properties` bag provided by Roslyn.

Edit:
This refers to [Diagnostic.Properties](docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.diagnostic.properties?view=roslyn-dotnet#Microsoft_CodeAnalysis_Diagnostic_Properties). 
We have a few rules which are using these properties and could benefit from this. 
E.g. https://github.com/SonarSource/sonar-dotnet/blob/734df7968037387faade9353c7aedd1a77c010a0/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GetHashCodeMutable.cs#L89

andrei-epure-sonarsource:
Not clear what this is about, so we are closing it.
@Evangelink - do you remember exactly what this was about?

Evangelink:
Some rules are storing extra information in the `properties` bag, for example the sub-messages related to secondary locations. With the test framework in place, you can test the location and the message reported but not these info so you could introduce a regression without noticing.

The idea of this ticket was to allow testing these extra data.

costin-zaharia-sonarsource:
So this refers to https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.diagnostic.properties?view=roslyn-dotnet#Microsoft_CodeAnalysis_Diagnostic_Properties. Indeed there are a few rules which are using this. E.g. https://github.com/SonarSource/sonar-dotnet/blob/734df7968037387faade9353c7aedd1a77c010a0/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GetHashCodeMutable.cs#L89

pavel-mikula-sonarsource:
We won't prioritize this. The properties are mostly for secondary locations (that are asserted separately), or codefix data that would demonstrate themselves in the codefix UT.

PR #1266, opened 2018-03-26
Fix S2228: Should not raise any issue when in a console application

### LINKED ISSUE #1223: Fix S2228: Should not raise any issue when in a console application
See https://github.com/SonarSource/sonarlint-visualstudio/issues/582

Evangelink:
Fix #1223


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConsoleLogging.cs
```diff
@@ -48,7 +48,8 @@ protected sealed override void Initialize(SonarAnalysisContext context)
             context.RegisterSyntaxNodeActionInNonGenerated(
                 c =>
                 {
-                    if (c.IsTest())
+                    if (c.IsTest() ||
+                        c.Compilation.Options.OutputKind == OutputKind.ConsoleApplication)
```
duncanp-sonar: We should refactor the _Verifier_ at some point so we can easily write unit tests for this kind of thing (either by changing it so we can pass in _CompilationOptions_ or perhaps a _Compilation_).

Evangelink: Yes there is a lot of work todo on the Verifier (see #817)

PR #1268, opened 2018-03-26
Rule S4524: switch 'default' case should be first or last

### LINKED ISSUE #1263: Adjust the behavior of S131 to not raise an issue when "default" is not at the correct position
Implements [RSPEC-131](https://jira.sonarsource.com/browse/RSPEC-131)

[RSPEC-4524 ](https://jira.sonarsource.com/browse/RSPEC-4524) is taking care to raise an issue if the "default" is not at the correct position.

### LINKED ISSUE #1267: Rule S4524: Switch 'default' case should be first or last
### Description

`switch` can contain a `default` clause for various reason: to handle unexpected values, to show all the cases where properly considered.
For readability purpose, to help a developer to quickly find the default behavior of a `switch` statement, it is recommended to put the `default` clause at the beginning or the end of the `switch` statement. This rule raised an issue if the `default` statement is not the first or the last one of the switch's cases.

### Related information

Relates to #1263

Evangelink:
Fix #1263 and fix #1267

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DefaultSectionShouldBeFirstOrLast.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DefaultSectionShouldBeFirstOrLast.cs
@@ -43,6 +43,39 @@ namespace Tests.Diagnostics
                 case 42:
                     break;
             }
+
+            switch (a)
+            {
+                default:
+                    break;
+            }
+
+            switch (a)
+            {
+                case 0:
+                    break;
+                case 1:
+                default:
+                    break;
+            }
+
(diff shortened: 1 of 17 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DefaultSectionShouldBeFirstOrLast.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            switch (a)
+            {
+            }
+
+            switch (a)
+            {
+                case 42:
+                    break;
+            }
```
duncanp-sonar: For completeness, how about a switch statement that only has a _default_ section?

PR #1273, opened 2018-03-27
Rules should be enabled by default depending on whether they are part…

### LINKED ISSUE #1272: Rules should be enabled by default depending on whether they are part of SonarWay
### Description

Let's revert to the previous state where rules default enable state was based on whether the rule is part of the SonarWay profile or not.

Evangelink:
… of SonarWay

Fix #1272


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
```diff
@@ -125,17 +125,6 @@ public void SonarDiagnosticAnalyzer_IsUsedInAllRules()
             }
         }
 
-        [TestMethod]
```
duncanp-sonar: Could we have a unit test to check that only SonarWay rules are activated by default? (I guess the utility rules might be a special case)

PR #1275, opened 2018-03-27
 Parameterized rules should be disabled by default

### LINKED ISSUE #1274: Parameterized rules should be disabled by default
### Description

Some of the parameterized rules are part of SonarWay but as the parameter configuration is not easy it is better to disable them by default.

Evangelink:
Fix #1274


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
```diff
@@ -125,17 +125,6 @@ public void SonarDiagnosticAnalyzer_IsUsedInAllRules()
             }
         }
 
-        [TestMethod]
```
duncanp-sonar: Could we have a _OnlyRulesInSonarWarAreActivatedByDefault_ test? (I guess the utility rules might be a special case)

PR #1280, opened 2018-03-27
Refactor utility classes

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/DiagnosticDescriptorBuilder.cs
```diff
@@ -27,6 +27,18 @@ namespace SonarAnalyzer.Helpers
 {
     public static class DiagnosticDescriptorBuilder
     {
+        internal const string SonarWayTag = "SonarWay";
+
+        public static DiagnosticDescriptor GetUtilityDescriptor(string diagnosticId, string title) =>
```
Evangelink: Extracted from all utility classes

PR #1281, opened 2018-03-28
Refactoring symbolic execution analysis to run once per mehod

valhristov:
Each existing SE rule was converted to a factory and analyzer (state) classes. The existing SE checks are not changed. There is a new diagnostic analyzer - SymbolicExecutionAnalyzer, which uses the registered factories to create analyzers for the corresponding rule when a SE analysis is requested.

Currently the factories are registered (hardcoded) in the SymbolicExecutionAnalyzer class' constructor. This could be changed in the future to reverse the dependency, but it will require more extensive testing, because we will introduce dependencies between diagnostic analyzers, which are generally meant to be independent. The hypothesis seems feasible, but it is not known whether it will work or not.

In general this will reduce the execution time of the its from ~600s to <500s (on my machine).

Best viewed without whitespace:
https://github.com/SonarSource/sonar-csharp/pull/1281/files?w=1

### DIFF
--- a/sonaranalyzer-dotnet/its/expected/Nancy/Nancy-{34576216-0DCA-4B0F-A0DC-9075E75A676F}-S3900.json
+++ b/sonaranalyzer-dotnet/its/expected/Nancy/Nancy-{34576216-0DCA-4B0F-A0DC-9075E75A676F}-S3900.json
@@ -1848,32 +1848,6 @@
 },
 {
 "id":  "S3900",
-"message":  "Refactor this method to add validation of parameter 'modelType' before using it.",
-"location":  {
-"uri":  "sources\Nancy\src\Nancy\ModelBinding\DefaultBinder.cs",
-"region":  {
-"startLine":  76,
-"startColumn":  35,
-"endLine":  76,
-"endColumn":  44
-}
-}
-},
-{
-"id":  "S3900",
-"message":  "Refactor this method to add validation of parameter 'modelType' before using it.",
-"location":  {
-"uri":  "sources\Nancy\src\Nancy\ModelBinding\DefaultBinder.cs",
-"region":  {
-"startLine":  81,
-"startColumn":  25,
-"endLine":  81,
-"endColumn":  34
-}
-}
-},
-{
-"id":  "S3900",
 "message":  "Refactor this method to add validation of parameter 'bodyStream' before using it.",
 "location":  {
 "uri":  "sources\Nancy\src\Nancy\ModelBinding\DefaultBodyDeserializers\JsonBodyDeserializer.cs",
@@ -2727,19 +2701,6 @@
 "startColumn":  39,
 "endLine":  46,
 "endColumn":  46
-}
-}
-},
-{
-"id":  "S3900",
-"message":  "Refactor this method to add validation of parameter 'path' before using it.",
-"location":  {
-"uri":  "sources\Nancy\src\Nancy\Routing\DefaultRouteSegmentExtractor.cs",
-"region":  {
-"startLine":  20,
-"startColumn":  41,
-"endLine":  20,
-"endColumn":  45
 }
 }
 },
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
@@ -18,7 +18,6 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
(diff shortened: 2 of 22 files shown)


sonaranalyzer-dotnet/its/expected/Nancy/Nancy-{34576216-0DCA-4B0F-A0DC-9075E75A676F}-S3900.json
```diff
@@ -1848,32 +1848,6 @@
 },
 {
 "id":  "S3900",
-"message":  "Refactor this method to add validation of parameter 'modelType' before using it.",
```
valhristov: Since all SE checks run simultaneously we could generate much bigger exploded graphs, thus creating some FNs in large methods. We could investigate the further increase of the EG limits in the future.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
```diff
@@ -79,203 +74,198 @@ public class ConditionEvaluatesToConstant : SonarDiagnosticAnalyzer
...
             DiagnosticDescriptorBuilder.GetDescriptor(S2589DiagnosticId, S2589MessageFormat, RspecStrings.ResourceManager);
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(s2583, s2589);
+        public ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(s2583, s2589);
 
-        protected sealed override void Initialize(SonarAnalysisContext context)
-        {
-            context.RegisterExplodedGraphBasedAnalysis((e, c) => CheckForRedundantConditions(e, c));
-        }
+        ISymbolicExecutionAnalyzer ISymbolicExecutionAnalyzerFactory.Create(CSharpExplodedGraph explodedGraph) =>
```
valhristov: The interface is internal because the CSEG is internal. I did not wanted to make the whole class internal, hence the explicit implementation

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Verifier.cs
```diff
@@ -120,27 +120,33 @@ internal static class Verifier
...
         }
 
         public static void VerifyAnalyzer(string path, SonarDiagnosticAnalyzer diagnosticAnalyzer,
             ParseOptions options = null, params MetadataReference[] additionalReferences)
         {
-            VerifyAnalyzer(new[] { path }, diagnosticAnalyzer, null, options, additionalReferences);
+            VerifyAnalyzer(new[] { path }, new[] { diagnosticAnalyzer }, null, options, additionalReferences);
+        }
+
+        public static void VerifyAnalyzer(string path, IEnumerable<SonarDiagnosticAnalyzer> diagnosticAnalyzers,
```
valhristov: Adding overload that allows you to run two analyzers in one test. This was needed for the InvalidCast rule which was split into two.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/AbstractExplodedGraph.cs
```diff
@@ -146,19 +146,16 @@ public void Walk()
             OnExplorationEnded();
         }
 
-        internal void AddExplodedGraphCheck<T>(T check)
+        public T GetOrAddCheck<T>(Func<T> factory)
```
valhristov: Since the SE analyzer registers states that subscribe for check events, adding a check does not replace an existing one, but reuses it.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecutionAnalyzer.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            try
+            {
+                explodedGraph.Walk();
+            }
+            finally
+            {
+                explodedGraph.ExplorationEnded -= OnExplorationEnded;
+
+                analyzerInstances.ForEach(analyzer => analyzer.Dispose());
```
valhristov: Unsubscribe from events.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecutionAnalyzer.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        protected sealed override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterExplodedGraphBasedAnalysis(Analyze);
+        }
+
+        private void Analyze(CSharpExplodedGraph explodedGraph, SyntaxNodeAnalysisContext context)
+        {
+            var analyzerInstances = analyzerFactories
+                .Where(factory => factory.IsEnabled(context))
```
valhristov: The analyzer factories could be disabled for some methods (e.g. for tests or private methods)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InvalidCastToInterface.cs
```diff
@@ -33,8 +34,7 @@
 namespace SonarAnalyzer.Rules.CSharp
 {
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
-    [Rule(DiagnosticId)]
```
valhristov: Rule attribute is removed because we have unit tests that ensure each rule is registered in exactly one diagnostic analyzer. InvalidCast is reported from two analyzers.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/AbstractExplodedGraph.cs
```diff
@@ -32,7 +32,7 @@ namespace SonarAnalyzer.SymbolicExecution
 {
     internal abstract class AbstractExplodedGraph
     {
-        internal const int MaxStepCount = 1000;
+        internal const int MaxStepCount = 2000;
```
valhristov: Increased the exploration limit to avoid FNs.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecutionAnalyzer.cs
```diff
@@ -0,0 +1,97 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarAnalyzer.Common;
+using SonarAnalyzer.Helpers;
+using SonarAnalyzer.SymbolicExecution;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(NullPointerDereference.DiagnosticId)]
```
Evangelink: I don't like that the Symbolic Execution Engine has to know about the rules using it. As I told you I prefer to have all analyzer referencing the SEE.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecutionAnalyzer.cs
```diff
@@ -0,0 +1,97 @@
...
+    [Rule(ConditionEvaluatesToConstant.S2589DiagnosticId)]
+    [Rule(InvalidCastToInterface.DiagnosticId)]
+    public class SymbolicExecutionAnalyzer : SonarDiagnosticAnalyzer
+    {
+        private readonly List<ISymbolicExecutionAnalyzerFactory> analyzerFactories;
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
+            analyzerFactories.SelectMany(factory => factory.SupportedDiagnostics).ToImmutableArray();
+
+        public SymbolicExecutionAnalyzer() : this(
```
Evangelink: Same thing here. IMO SEE should not know about other classes.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecutionAnalyzer.cs
```diff
@@ -0,0 +1,97 @@
...
+                explodedGraph.Walk();
+            }
+            finally
+            {
+                explodedGraph.ExplorationEnded -= OnExplorationEnded;
+
+                analyzerInstances.ForEach(analyzer => analyzer.Dispose());
+            }
+
+            void OnExplorationEnded(object sender, EventArgs e) =>
```
Evangelink: IMO this shouldn't be this class reporting issues but the analyzers themselves. This class should only be responsible of doing the symbolic execution (the walk).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecutionAnalyzer.cs
valhristov: +1
It is not difficult to revert the dependency, but it will require significant testing, as we do not have dependent rules now and there could be potentially many cases where it could break. The current solution is cheap and still provides significant performance improvement.

valhristov:
I would consider this PR as POC, we could look for better implementation in the future.

PR #1286, opened 2018-03-28
 Rules should use analysis scope from RSPEC and not be hardcoded

### LINKED ISSUE #1087: Rules should use analysis scope from RSPEC and not be hardcoded
## Description

RSPEC database has a "Analysis Scope" field which specifies whether the rule applies to main source, test sources or both. Currently the C# & VB.NET analysers do not enforce that field systematically.
However since this could create a compatibility issue (the analyzers would suddenly raise a lot of issues on test code where there were none before), this will be evaluated on internally before release and we could have provide a switch to disable that new behaviour.

Note: If the "Analysis Scope" is not properly set (historical reasons), then the default should be "Main Sources" only.

## Tasks
The following tasks should be implemented:

- Update all rules to have the scope metadata
- Remove the specific (hardcoded) checks
-  In the base class implement a standard way of deciding whether the rule should be called or not. Note that for the first implementation the value is immutable (will only change by updating RSPEC and releasing a new version of the analyzer). _I suggest to use the resx as the first implementation as it already contains the rules info._

See [MMF-1160](https://jira.sonarsource.com/browse/MMF-1160)

Depends upon [RULEAPI-482](https://jira.sonarsource.com/browse/RULEAPI-482) (implementation ticket for rule-api)

Evangelink:
Fix #1087


Evangelink:
@valhristov here you go

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/DiagnosticDescriptorBuilder.cs
```diff
@@ -62,6 +65,25 @@ private static IEnumerable<string> BuildCustomTags(string diagnosticId, Resource
                 yield return SonarWayTag;
             }
 
+            var scope = resourceManager.GetString($"{diagnosticId}_Scope");
+            if (scope == "Main")
+            {
+                yield return MainSourceScopeTag;
```
duncanp-sonar: Using _yield_ doesn't really add anything here; it doesn't make the code more readable, but it does add to the execution time.

duncanp-sonar:
This change looks like it's had a significant effect on the expected results in the ITs.
How big is the likely on customers? Is it likely to break their QGs? (i.e. do we need to talk to Support, or make this change in a major version release)

PR #1298, opened 2018-03-29
Invert #if condition to enable faster its

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/DiagnosticDescriptorBuilder.cs
```diff
@@ -91,7 +91,7 @@ private static string[] BuildUtilityCustomTags()
         {
             return new[]
             {
-#if DEBUG
+#if !DEBUG
```
Evangelink: I inverted the condition @michalb-sonar set on my last commit by mistake (while refactoring). This put it back to the right way.

Issue #1295, opened 2018-03-29
Fix S2589: "Condition evaluates to constant" false positive is generated in methods with yield


valhristov:
### Description

The following code generates a FP:

```
using System.Collections.Generic;

public class ZeroGenerator
{
    private bool generate;

    public IEnumerable<int> GetZeroes()
    {
        generate = true;
        while (generate) // Noncompliant, FP
        {
            yield return 0;
        }
    }

    public void Stop()
    {
        generate = false;
    }
}
```

### Known workarounds

Use property instead of a field.

### Related information

* SonarC# any version
* Reported in [Google groups](https://groups.google.com/d/msgid/sonarqube/910df9ed-26bc-4f64-8f5a-4cafba87e240%40googlegroups.com?utm_medium=email&utm_source=footer)

christophe-zurn-sonarsource:
This ticket actually involves cross-procedural symbolic execution: in order to know that the field has changed value in the generator method, we need to know where the generator method is used and whether the field is updated inside the loop using it.
Ex:
```
 foreach (var value in GetZeroes())
{
  Console.WriteLine(value);
}
```
In the above code, the field is never updated inside the loop. So it will result in an infinite loop and the `generate` is actually always true.

christophe-zurn-sonarsource:
As discussed internally, in order to reduce noise, we will currently ignore generator functions for rule S2589. This will unfortunately causes some FN, but it is preferred than having FP

PR #1299, opened 2018-03-29
Adding support for 'yield return' in CFG

valhristov:
Related to #1295 but does not fully implement it.

The `yield return` statements used to generate s `SimpleBlock` in CFG. Now a `JumpBlock` is generated, which "jumps" to the next statement. The `JumpBlock` allows us to do additional processing for `yield return` statements, such as to delete the constraints from class fields.


valhristov:
Yes, it does not implement the ticket (completely). This just changes the representation of `yield return` statements in CFG from a `simple` block to `jump block`. Better loop support is required to resolve the ticket (e.g. the loops should be executed more than once).

PR #1307, opened 2018-04-04
Update CI build script to fix error on versioning with Next

Evangelink:


scripts/build/ci-build.ps1
```diff
@@ -258,7 +258,9 @@ function Get-MavenExpression([string]$exp) {
     $out = Exec { mvn help:evaluate -B -Dexpression="${exp}" }
     Test-ExitCode "ERROR: Evaluation of expression ${exp} FAILED."
 
-    return ($out | Select-String -NotMatch -Pattern '^\[|Download\w\+\:').Line
+    $firstEntry = $out | Select-String -NotMatch -Pattern '^\[|Download\w\+\:' | Select-Object -First 1
```
michalb-sonar: This should also work (so the 2nd line is not needed): `| Select-Object -First 1 -ExpandProperty Line`

PR #1318, opened 2018-04-11
Rule S4432: AES encryption algorithm should be used with secured mode

### LINKED ISSUE #1303: Rule S4432: AES encryption algorithm should be used with secured mode
Implements [RSPEC-4432](https://jira.sonarsource.com/browse/RSPEC-4432)

Evangelink:
Fix #1303


sonaranalyzer-dotnet/rspec/cs/S4432_c#.html
```diff
@@ -0,0 +1,36 @@
...
+{
+  KeySize = 128,
+  BlockSize = 128,
+  Mode = CipherMode.OFB, // Noncompliant
+  Padding = PaddingMode.PKCS7
+};
+</pre>
+<h2>Compliant Solution</h2>
+<pre>
+Cipher c = Cipher.getInstance("AES/GCM/NoPadding");
```
duncanp-sonar: Java-style naming of _getInstance_. Does this matter?

Evangelink: Yes... We don't have any compliant solution for C# xD

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AesEncryptionShouldBeUsedWithSecuredMode.cs
```diff
@@ -0,0 +1,78 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+        private static readonly ISet<string> invalidEnumValues =
+            new HashSet<string>
+            {
+                "CipherMode.CBC",
+                "CipherMode.CFB",
+                "CipherMode.CTS",
+                "CipherMode.ECB",
+                "CipherMode.OFB"
```
duncanp-sonar: It looks like these are all of the cipher modes supported by _AesManaged_ - are there any possible valid uses of the class?

Evangelink: There are none and there is no compliant encryption class built-in the framework, you'll have to use third parties.

duncanp-sonar: Also, according to [MSDN](https://msdn.microsoft.com/en-us/library/system.security.cryptography.aesmanaged.mode(v=vs.110).aspx) the default value for _Mode_ is _CBC_. However, we're only checking for assignments to _Mode_ so we won't catch default usage.

Is that ok i.e. should we just treat it as a false negative? If so, it might be worth documenting somewhere, either as a comment in the code or in the test cases.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/AesEncryptionShouldBeUsedWithSecuredMode.cs
```diff
@@ -0,0 +1,46 @@
...
+            {
+                Mode = CipherMode.CFB // Noncompliant
+            };
+
+            new AesManaged
+            {
+                Mode = CipherMode.CTS // Noncompliant
+            };
+
+            aes.Mode = CipherMode.ECB; // Noncompliant
```
duncanp-sonar: Maybe add a test case to show we are only raising issues for modes we know about e.g.
 _aes.Mode = CipherMode.NewUnknownSecureMode;_

(it's possible that a future version of the framework might include a secure cipher mode)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AesEncryptionShouldBeUsedWithSecuredMode.cs
Evangelink: I forgot to look for the default value. Maybe the best thing is to report on `AesManaged` creation as there is no secure cipher for the time being. WDYT?

duncanp-sonar: It's probably good enough - it will always be correct unless and until a new secure version is included in the framework. We could write a smarter version at that point.

sonaranalyzer-dotnet/rspec/cs/S4432_c#.html
Evangelink: I asked Alex to move the compliant solution to a sub-task so we don't end up with the inherited solution.

valhristov:
This could be easily rewritten using the base class from #1337 and it will support checks for default constructor, object initializers, property settings.

Evangelink:
@valhristov Do you need the code or is it ok to close and delete this PR?

valhristov:
I will try to rewrite the rule, then I will tell you if this should be closed

valhristov:
@Evangelink since there is no way to satisfy this rule with a built-in framework class, why don't we raise every time we see creation of `AesManaged`?

Evangelink:
@valhristov look at the RSPEC (and the email exchange I have initiated). Rule was changed to report primary on `AesManaged` and secondary on the `CipherMode`.

FYI we decided to keep the secondary because `PKCS7` can be valid under certain circumstances (explained in the RSPEC).

Evangelink:
Handled by another PR #1351

PR #1319, opened 2018-04-11
Rule S4564: ASP.NET HTTP request validation feature should not be dis…

### LINKED ISSUE #1313: Rule S4564: ASP.NET HTTP request validation feature should not be disabled
Implements [RSPEC-4564](https://jira.sonarsource.com/browse/RSPEC-4564)

Evangelink:
…abled

Fix #1313

### DIFF
--- a/sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
+++ b/sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
@@ -1,13 +1,20 @@
-<p>ASP.NET is coming with a feature looking at HTTP requests to prevent potentially dangerous content to perform a cross-site scripting (XSS) attack.
-There is no reason to disable this mechanism even if it should not be the only one in place to prevent XSS attacks.</p>
-<p>This rule raises an issue if <code>System.Web.Mvc.ValidateInputAttribute(false)</code> is set on a method.</p>
+<p>ASP.Net has a feature to validate HTTP requests to prevent potentially dangerous content to perform a cross-site scripting (XSS) attack. There is
+no reason to disable this mechanism even if other checks to prevent XXS attacks are in place.</p>
+<p>This rule raises an issue if a method with parameters is marked with <code>System.Web.Mvc.HttpPostAttribute</code> and not
+<code>System.Web.Mvc.ValidateInputAttribute(true)</code>.</p>
 <h2>Noncompliant Code Example</h2>
 <pre>
 public class FooBarController : Controller
 {
-    [HttpPost]
-    [ValidateInput(false)] // Noncompliant
+    [HttpPost] // Noncompliant
+    [ValidateInput(false)]
     public ActionResult Purchase(string input)
+    {
+        return Foo(input);
+    }
+
+    [HttpPost] // Noncompliant
+    public ActionResult PurchaseSomethingElse(string input)
     {
         return Foo(input);
     }
@@ -25,6 +32,8 @@ public class FooBarController : Controller
     }
 }
 </pre>
+<h2>Exceptions</h2>
+<p>Parameterless methods marked with <code>System.Web.Mvc.HttpPostAttribute</code> will not trigger this issue.</p>
 <h2>See</h2>
 <ul>
   <li> <a href="http://cwe.mitre.org/data/definitions/79">MITRE, CWE-79</a> - Improper Neutralization of Input During Web Page Generation ('Cross-site
--- a/sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
+++ b/sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
@@ -1,5 +1,6 @@
 <p>ASP.NET is coming with a feature looking at HTTP requests to prevent potentially dangerous content to perform a cross-site scripting (XSS) attack.
 There is no reason to disable this mechanism even if it should not be the only one in place to prevent XSS attacks.</p>
+<p>This rule raises an issue if <code>System.Web.Mvc.ValidateInputAttribute(false)</code> is set on a method.</p>
 <h2>Noncompliant Code Example</h2>
 <pre>
 public class FooBarController : Controller
--- a/sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
+++ b/sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
@@ -1,0 +1,35 @@
+<p>ASP.NET is coming with a feature looking at HTTP requests to prevent potentially dangerous content to perform a cross-site scripting (XSS) attack.
+There is no reason to disable this mechanism even if it should not be the only one in place to prevent XSS attacks.</p>
+<h2>Noncompliant Code Example</h2>
+<pre>
+public class FooBarController : Controller
+{
+    [HttpPost]
+    [ValidateInput(false)] // Noncompliant
+    public ActionResult Purchase(string input)
+    {
+        return Foo(input);
+    }
+}
+</pre>
+<h2>Compliant Solution</h2>
+<pre>
+public class FooBarController : Controller
+{
+    [HttpPost]
+    [ValidateInput(true)] // Compliant
+    public ActionResult Purchase(string input)
+    {
+        return Foo(input);
+    }
+}
+</pre>
+<h2>See</h2>
+<ul>
+  <li> <a href="http://cwe.mitre.org/data/definitions/79">MITRE, CWE-79</a> - Improper Neutralization of Input During Web Page Generation ('Cross-site
+  Scripting') </li>
+  <li> OWASP Top 10 2017 Category A7 - Cross-Site Scripting (XSS) </li>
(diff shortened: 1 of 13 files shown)


sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
```diff
@@ -0,0 +1,36 @@
...
+        return Foo(input);
+    }
+}
+</pre>
+<h2>Compliant Solution</h2>
+<pre>
+public class FooBarController : Controller
+{
+    [HttpPost]
+    [ValidateInput(true)] // Compliant
```
duncanp-sonar: What if the user hasn't specified the attribute at all? Is it implied, or should we have a separate rule to check that the attribute is applied if there is any user input?

sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
```diff
@@ -0,0 +1,36 @@
+<p>ASP.NET is coming with a feature looking at HTTP requests to prevent potentially dangerous content to perform a cross-site scripting (XSS) attack.
```
duncanp-sonar: Suggested wording change: _ASP.Net has a feature to validate HTTP requests..._

sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
```diff
@@ -0,0 +1,36 @@
...
+There is no reason to disable this mechanism even if it should not be the only one in place to prevent XSS attacks.</p>
```
duncanp-sonar: Suggested wording change: _There is no reason to disable this mechanism even if other checks to prevent XXS attacks are in place_.

sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
Evangelink: Good question. This need to be covered, @agigleux shall we update the rule or add a new one?

sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
Evangelink: @agigleux Could you update the RSPEC?

sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
Evangelink: @agigleux Could you update the RSPEC?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/HttpPostControllerActionShouldValidateInput.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using System.Web.Mvc;
+
+namespace Tests.Diagnostics
+{
+    public class FooBarController1 : Controller
+    {
+        [HttpPost]
+        [ValidateInput(false)] // Noncompliant
```
duncanp-sonar: Why not check that the squigglie is reported on the expected characters?

Evangelink: Cause I went too fast and forgot... Thx

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/HttpPostControllerActionShouldValidateInput.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        [HttpPost]
+        [ValidateInput()] // Compliant - not false
+        public ActionResult Purchase3(string input)
+        {
+            return Foo(input);
+        }
+
+        [ValidateInput(false)] // Noncompliant - no HttpPost
```
duncanp-sonar: I found the comment _Noncompliant - no HttpPost_ confusing initially. It suggests that that the code is invalid because it HttpPost is missing, whereas what I think it means is that the usage of the attribute is invalid, but we'll still report.
A comment like _Noncompliant - reported even though no HttpPost attribute_

Same for the comment in line 39

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/HttpPostControllerActionShouldValidateInput.cs
```diff
@@ -0,0 +1,45 @@
...
+    public class FooBarController1
+    {
+        [HttpPost]
+        [ValidateInput(false)] // Noncompliant - class doesn't inherit from Controller
+        public ActionResult Purchase1(string input)
+        {
+            return Foo(input);
+        }
+    }
+}
```
duncanp-sonar: A couple of other tests cases (all of which are invalid code):
* [ValidateInput("foo")]  //  invalid code - not a boolean
* [ValidateInput(true, false]  //  invalid code - wrong number of attributes
* [Unrecognised]  //  invalid code - unrecognised attribute

sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
Evangelink: Done.

sonaranalyzer-dotnet/rspec/cs/S4564_c#.html
Evangelink: Done.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/HttpPostControllerActionShouldValidateInput.cs
Evangelink: Done.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/HttpPostControllerActionShouldValidateInput.cs
Evangelink: Done.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/HttpPostControllerActionShouldValidateInput.cs
Evangelink: Done.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/HttpPostControllerActionShouldValidateInput.cs
duncanp-sonar: Other test cases:
* method with _[ValidateInput(false)]_ but no _[HttpPost]_ attribute -> shouldn't raise
* fully qualified attribute names: just to check they are handled correctly, since the analyzer is doing _StartsWith_ check on _Name_

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/HttpPostControllerActionShouldValidateInput.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
```
duncanp-sonar: A general question about known/expected false positives: if we can think of cases that would cause false positives, is it worth calling them out explicitly by writing a test case and commenting that's it's a "known" false positive?

An example here would be code with a custom attribute with a similar name that isn't from the core ASP.Net framework attribute e.g. _HttpPostHandler_.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/HttpPostControllerActionShouldValidateInput.cs
```diff
@@ -0,0 +1,92 @@
...
+                        methodDeclaration.ParameterList.Parameters.Count == 0)
+                    {
+                        return;
+                    }
+
+                    var attributes = methodDeclaration.AttributeLists
+                        .SelectMany(list => list.Attributes)
+                        .ToList();
+
+                    var httpPostAttribute = attributes.FirstOrDefault(a => a.Name.ToString().StartsWith("HttpPost"));
```
duncanp-sonar: Is _StartsWith_ case-sensitive?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/HttpPostControllerActionShouldValidateInput.cs
Evangelink: I think that's better to put as many cases as possible. I'll update my implementation to work with symbols for now on and we will see later (we will have to update a lot of rules anyway).

PR #1322, opened 2018-04-12
Use more new C# syntax in the code

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
```diff
@@ -147,10 +147,9 @@ private static bool CheckNullAndValueType(ITypeSymbol typeNull, ITypeSymbol type
             ExpressionSyntax comparedToNull, SemanticModel semanticModel, bool comparedIsNullInTrue, out bool isNullCoalescing)
         {
             isNullCoalescing = false;
-            var return1 = statement1 as ReturnStatementSyntax;
             var return2 = statement2 as ReturnStatementSyntax;
 
-            if (return1 != null && return2 != null)
+            if (statement1 is ReturnStatementSyntax return1 && return2 != null)
```
duncanp-sonar: Why not do the same for _return2_?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassNotInstantiatable.cs
```diff
@@ -100,7 +100,7 @@ private static bool IsNonStaticClassWithNoAttributes(INamedTypeSymbol namedType)
             return typeDeclarations
                 .Select(classDeclaration => new
                 {
-                    SemanticModel = classDeclaration.SemanticModel,
+                    classDeclaration.SemanticModel,
```
duncanp-sonar: Personally I find the previous version of the code easier to read.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplificationCodeFixProvider.cs
```diff
@@ -117,10 +117,9 @@ public class ConditionalSimplificationCodeFixProvider : SonarCodeFixProvider
             ExpressionSyntax condition, ExpressionSyntax compared, SemanticModel semanticModel, SyntaxAnnotation annotation,
             bool isNullCoalescing)
         {
-            var return1 = statement1 as ReturnStatementSyntax;
             var return2 = statement2 as ReturnStatementSyntax;
 
-            if (return1 != null && return2 != null)
+            if (statement1 is ReturnStatementSyntax return1 && return2 != null)
```
duncanp-sonar: Same comment as above - looks odd changing _return1_ but not _return2_.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantJumpStatement.cs
```diff
@@ -129,14 +129,12 @@ private static bool IsOnlyYieldBreak(JumpBlock jumpBlock, int yieldStatementCoun
 
         private static bool IsThrow(JumpBlock jumpBlock)
         {
-            var throwStatement = jumpBlock.JumpNode as ThrowStatementSyntax;
-            return throwStatement != null;
+            return jumpBlock.JumpNode is ThrowStatementSyntax throwStatement;
```
duncanp-sonar: _throwStatement_ isn't used.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
```diff
@@ -421,11 +420,11 @@ private static AccessorDeclarationSyntax GetAccessorSyntax(IMethodSymbol methodS
             }
 
             var candidateSymbols = allNodes
-                .Select(n => new { Syntax = n.SyntaxNode, SemanticModel = n.SemanticModel, Symbol = n.SemanticModel.GetSymbolInfo(n.SyntaxNode) })
+                .Select(n => new { Syntax = n.SyntaxNode, n.SemanticModel, Symbol = n.SemanticModel.GetSymbolInfo(n.SyntaxNode) })
```
duncanp-sonar: _Syntax = n.SyntaxNode_ could be shortened too, couldn't it?

I think it would be better to be consistent and either omit all unnecessary explicit names, or explicitly name everything.

Evangelink: You can't, it would rename `Syntax` into `SyntaxNode`.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantJumpStatement.cs
Evangelink: Good point.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
Evangelink: Automatic code fix on all instances.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplificationCodeFixProvider.cs
Evangelink: Automatic code fix on all instances.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
duncanp-sonar: Why is that a problem? Because it's also a class name?

Evangelink: I'll need to rename usages but except that there is no problem.

PR #1327, opened 2018-04-13
 Changes scope for rule S1192 to look for duplicates at class level

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
```diff
@@ -51,55 +51,45 @@ public sealed class StringLiteralShouldNotBeDuplicated : ParameterLoadingDiagnos
 
         protected override void Initialize(ParameterLoadingAnalysisContext context)
         {
-            context.RegisterCompilationStartAction(
-                csac =>
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
                 {
-                    if (csac.Compilation.IsTest())
```
duncanp-sonar: The RSPEC says this is a "main sources" rule. Is this check for test code no longer required because it is being done in the base class?

duncanp-sonar:
What's the reason reducing the scope of the check from compilation to class?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
Evangelink: Yes this is done on base class.

Evangelink:
@duncanp-sonar Not supported by SonarQube! I need to create a thread to suggest they support this scenario.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
```diff
@@ -51,55 +51,45 @@ public sealed class StringLiteralShouldNotBeDuplicated : ParameterLoadingDiagnos
...
-                    {
-                        return;
-                    }
+                    var stringLiterals = c.Node.DescendantNodes()
```
duncanp-sonar: Just a comment on readability: knowing the kind of the nodes the action is being registered against is important when reading and understanding the code. However, inlining the callback makes this hard to see; the reader has to just forward thirty or so lines to find the kind of nodes, then jump back again to continue reading.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StringLiteralShouldNotBeDuplicated.cs
```diff
@@ -41,4 +41,14 @@ public void Validate(object foobar)
...
+
+    public struct Foo
+    {
+        public const string NameConst = "foobar"; // Noncompliant {{Define a constant instead of using this literal 'foobar' 4 times.}}
+//                                      ^^^^^^^^
+
+        private string name1 = "foobar"; // Secondary
+        private string name2 = "foobar"; // Secondary
+        private string name3 = "foobar"; // Secondary
+    }
```
duncanp-sonar: What about nested classes?
* If a nested class has too many duplicates, will we report two issues, once against the nested class and once against the containing class?
* should we report if e.g. the containing class has three duplicates and the nested class has two?

Evangelink: Again, good catch! I'll report/analyze only from the top-level class, WDYT?

duncanp-sonar: It's not perfect, but it's probably simplest. We can always make it smarter later if necessary.

duncanp-sonar:
Scope: can't we report issues to SonarQube at project/module/compilation level? (?could that be why there wasn't a primary location in the previous code?)

Evangelink:
You can report a primary level issue at module (although we are the only one doing so). The problem is about secondary locations (i.e. all instances of this string) which make it really hard to use. Hence this change of scope.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StringLiteralShouldNotBeDuplicated.cs
duncanp-sonar: ...and I hesitate to say it, but... partial classes? (maybe just document that as a false negative)

duncanp-sonar:
Scope: ok. Might be worth capturing this in a comment so we don't have to remember in the future.

duncanp-sonar:
@Evangelink does this rule correctly handle partial classes?
Either way, it would be good to have it as a test case, even if it currently a false negative.

PR #1328, opened 2018-04-13
Rule S881: Increment (++) and decrement (--) operators should not be …

### LINKED ISSUE #1151: Rule S881: Increment (++) and decrement (--) operators should not be used in a method call or mixed with other operators in an expression
Implements [RSPEC-881](https://jira.sonarsource.com/browse/RSPEC-881)

Evangelink:
…used in a method call or mixed with other operators in an expression

Fix #1151

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DontMixIncrementOrDecrementWithOtherOperators.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DontMixIncrementOrDecrementWithOtherOperators.cs
@@ -1,0 +1,29 @@
+﻿using System;
+
+namespace Tests.Diagnostics
+{
+    public class Program
+    {
+        public void Foo()
+        {
+            int val1 = 0;
+            int val2 = 0;
+
+            var result = ++val1 + val2; // Noncompliant {{Extract this [increment|decrement] operation into a dedicated statement.}}
+//                              ^
+            result = val1++ - 1; // Noncompliant
+            result = val2++ * 2; // Noncompliant
+            result = val2++ / 4; // Noncompliant
+            result = --val1 % 2; // Noncompliant
+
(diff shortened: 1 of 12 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DontMixIncrementOrDecrementWithOtherOperators.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    public class Program
+    {
+        public void Foo()
+        {
+            int val1 = 0;
+            int val2 = 0;
+
+            var result = ++val1 + val2; // Noncompliant {{Extract this [increment|decrement] operation into a dedicated statement.}}
+//                              ^
```
duncanp-sonar: Might be better to put the 	squigglie under the ++val1 since that is what the rule is complaining about (also, the squigglie won't be easy to see under a single character).

duncanp-sonar: Also, it would be clear if for expressions like ++val1 * ++val2

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DontMixIncrementOrDecrementWithOtherOperators.cs
```diff
@@ -0,0 +1,29 @@
...
+            result = val1++ - 1; // Noncompliant
+            result = val2++ * 2; // Noncompliant
+            result = val2++ / 4; // Noncompliant
+            result = --val1 % 2; // Noncompliant
```
duncanp-sonar: How about switching the order of the arguments in a couple of the test cases so the literals come first?

PR #1331, opened 2018-04-16
Rule S4581: 'new Guid()' should not be used

### LINKED ISSUE #1329: Rule S4581: "new Guid()" should not be used
### Description

When the syntax `new Guid()` (i.e. parameterless instantiation) is used, it must be that one of three things is wanted:

- An empty GUID, in which case `Guid.Empty` is clearer.
- A randomly-generated GUID, in which case `Guid.NewGuid()` should be used.
- A new GUID with a specific initialization, in which case the initialization parameter is missing.

This rule raises an issue when a parameterless instantiation of the Guid struct is found.

Implements [RSPEC-4581](https://jira.sonarsource.com/browse/RSPEC-4581)

Evangelink:
Fix #1329


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NewGuidShouldNotBeUsed.cs
```diff
@@ -0,0 +1,17 @@
...
+    public class Program
+    {
+        public void Foo()
+        {
+            var result = new Guid(); // Noncompliant {{Please review this use of 'new Guid()'.}}
+//                       ^^^^^^^^^^
+
+            result = Guid.Empty;
+            result = Guid.NewGuid();
+            result = new Guid(new byte[0]);
```
Evangelink: This raises a runtime argument exception saying byte array should have a length of exactly 16 bytes. Looks like a good candidate for another rule.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NewGuidShouldNotBeUsed.cs
```diff
@@ -0,0 +1,59 @@
...
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public sealed class NewGuidShouldNotBeUsed : SonarDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S4581";
+        private const string MessageFormat = "Please review this use of 'new Guid()'.";
```
valhristov: I would rather suggest what we think is wrong `Use "Guid.NewGuid() or "Guid.Empty" instead of "new Guid()"`

Evangelink: @duncanp-sonar @valhristov Would it be ok: `Use 'Guid.NewGuid()', 'Guid.Empty' or add arguments to this Guid instantiation.`?

PR #1333, opened 2018-04-17
[DO NOT MERGE] Remove Sonar Analysis from Jenkins build

Evangelink:


Evangelink:
We will keep the double stream of analysis for now until we manage to move the full process to VSTS.

PR #1335, opened 2018-04-18
Rule S4586: Non-async 'Task/Task<T>' method should not return null

### LINKED ISSUE #1334: Rule S4586: Non-async "Task/Task<T>" method should not return null
### Description

Returning `null` from a non-`async` `Task`/`Task<T>` method will cause a `NullReferenceException` at runtime. This problem can be avoided while keeping the right intent from the developer by returning `Task.FromResult<T>(null)`.

Evangelink:
Fix #1334


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NonAsyncTaskShouldNotReturnNull.cs
```diff
@@ -0,0 +1,93 @@
...
+                    var methodDeclaration = returnStatement.FirstAncestorOrSelf<MethodDeclarationSyntax>();
+
+                    if (methodDeclaration != null &&
+                        !methodDeclaration.Modifiers.Any(m => m.IsKind(SyntaxKind.AsyncKeyword)) &&
+                        IsTaskReturnType(methodDeclaration, c.SemanticModel))
+                    {
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, returnStatement.Expression.GetLocation()));
+                    }
+                },
+                SyntaxKind.ReturnStatement);
```
Evangelink: I was tempted to look for method declaration to avoid reporting multiple times for the same method but as we don't handle secondary locations correctly in SonarLint/NuGet I decided to go this way.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NonAsyncTaskShouldNotReturnNull.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace Tests.Diagnostics
+{
+    public class CompliantUseCases
+    {
+        public void Foo()
+        {
+            return null; // not Task
```
valhristov: Add test for arrow methods: `public Task Foo() => null;` it is less likely, but possible. Also local properties, functions and lambdas could also return null without having a `return` statement.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NonAsyncTaskShouldNotReturnNull.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        private static ITypeSymbol GetReturnType(ISymbol symbol) =>
+            symbol is IMethodSymbol methodSymbol
+            ? methodSymbol.ReturnType
+            : symbol.GetSymbolType();
+
+        private bool IsSafeTaskReturnType(ISymbol enclosingMemberSymbol) =>
+            // IMethodSymbol also handles lambdas
+            (enclosingMemberSymbol as IMethodSymbol)?.IsAsync
+            ?? false;
```
valhristov: (cosmetic) I would replace with `enclosingMemberSymbol is IMethodSymbol methodSymbol && methodSymbol.IsAsync`

This comment also applies for the `IsTaskReturnType` method. In general I don't like `?? false` constructs when they contain `as` operator and conditional member access. To me they seem less readable than a straight `null` check.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NonAsyncTaskShouldNotReturnNull.cs
```diff
@@ -0,0 +1,114 @@
...
+                    if (!nullLiteral.Parent.IsAnyKind(TrackedNullLiteralLocations))
+                    {
+                        return;
+                    }
+
+                    var enclosingMember = nullLiteral.Ancestors().FirstOrDefault(n => n.IsAnyKind(PossibleEnclosingSyntax));
+                    if (enclosingMember == null ||
+                        enclosingMember.IsKind(SyntaxKind.VariableDeclaration))
+                    {
+                        return;
```
valhristov: I am sorry for not noticing this before, but I suspect we will not catch `return (null);`.

I not sure if it is worth adding support for:
```
return condition ? Task.FromResult(5) : null;
```

PR #1338, opened 2018-04-19
Rule S4433: LDAP connections should be authenticated

### LINKED ISSUE #1310: Rule S4433: LDAP connections should be authenticated
Implements [RSPEC-4433](https://jira.sonarsource.com/browse/RSPEC-4433)

valhristov:
Fix #1310


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotUseUnauthenticatedLdapConnections.cs
```diff
@@ -0,0 +1,67 @@
...
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public sealed class DoNotUseUnauthenticatedLdapConnections : InitializeObjectsWithValueBase
+    {
+        internal const string DiagnosticId = "S4433";
+        private const string MessageFormat = "Set the 'AuthenticationType' property of this DirectoryEntry to 'AuthenticationTypes.Secure'.";
+        private const int AuthenticationTypes_Secure = 1;
```
valhristov: Avoiding reference here, this is the value of `AuthenticationTypes.Secure`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InitializeObjectsWithValueBase.cs
```diff
@@ -119,19 +122,21 @@ private bool IsLaterAssignedWithExpectedValue(ObjectCreationExpressionSyntax obj
...
+                && IsExpectedValue(assignment.Right, semanticModel);
         }
 
         private bool IsTrackedType(ExpressionSyntax expression, SemanticModel semanticModel) =>
             semanticModel.GetSymbolInfo(expression).Symbol
                 .GetSymbolType()
                 .Is(TrackedType);
 
-        private bool IsInitializedAsExpected(ObjectCreationExpressionSyntax objectCreation, SemanticModel semanticModel) =>
+        protected virtual bool IsInitializedAsExpected(ObjectCreationExpressionSyntax objectCreation, SemanticModel semanticModel) =>
```
valhristov: The ctor+initializer check is different, hence the virtual - previously we assumed that the default ctor has the wrong value, but in this rule it is with the correct value

valhristov:
Build is failing because it cannot find the base branch on the server... Go figure...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotUseUnauthenticatedLdapConnections.cs
duncanp-sonar: Might be worth putting the reason in a comment.

PR #1342, opened 2018-04-19
Refactor to simplify testing for attribute(s).

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SymbolHelper.cs
```diff
@@ -45,11 +45,16 @@ public static IEnumerable<INamedTypeSymbol> GetAllNamedTypes(this INamespaceSymb
             }
         }
 
-        public static bool IsObsolete(this ISymbol symbol)
-        {
-            return symbol != null &&
-                symbol.GetAttributes().Any(a => a.AttributeClass.Is(KnownType.System_ObsoleteAttribute));
-        }
+        internal static bool HasAttribute(this ISymbol symbol, KnownType attributeType) =>
+            symbol?.GetAttributes().Any(a => a.AttributeClass.Is(attributeType))
```
Evangelink: I decided to copy the implementation rather than wrapper the item in a set (which can be expensive).

PR #1349, opened 2018-04-23
Refactor code to use ToUpperInvariant instead of ToLowerInvariant

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotHardcodeCredentials.cs
```diff
@@ -139,14 +139,16 @@ private string FindBannedWords(string variableName, string variableValue)
                 .Intersect(splitCredentialWords)
                 .ToHashSet();
 
-            var matches = passwordValuePattern.Matches(variableValue.ToLowerInvariant());
+            var matches = passwordValuePattern.Matches(variableValue);
```
duncanp-sonar: What's the case-sensitivity of this match? You could specify it in the _ToHashSet_ Iine 140 to make it explicit.

PR #1369, opened 2018-04-30
Remove deprecated TEST_SUCCESS_DENSITY

Evangelink:
This is a "finding" we made with Julien when trying to understand what was wrong with reporting metrics at module level when no files are analyzed (triggered more often when analyzing PRs).


sonar-dotnet-tests-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensor.java
```diff
@@ -98,14 +98,6 @@ void analyze(SensorContext context, UnitTestResults unitTestResults) {
         .withValue(executionTime)
         .save();
     }
-
-    if (aggregatedResults.tests() > 0) {
```
Evangelink: There is a warning in logs for quite a long time saying it is deprecated + this value is ignored by the server.

Issue #1365, opened 2018-04-26
MethodId should contain fully qualified method name + parameter types


valhristov:
# C# Method Id

The method Id consists of:
```
TypeId := type.Assembly.Name;[type.Namespace.Name]type.Name[<type.GenericArg1,...>]
ParamId := TypeId(param.Type)
MethodId := TypeId(method.ContainingType).MethodName[<method.GenericArg1,...>](ParamId(method.Parameters[0]),...)
```

# Examples

The following method Ids:

```
"project1.dll;Namespace.Class1.PrimitiveTypes1(mscorlib;string)"
"project1.dll;Namespace.Class1.PrimitiveTypes2(mscorlib;string)"
"project1.dll;Namespace.Class1.PrimitiveTypes3(mscorlib;string)"
"project1.dll;Namespace.Class1.UserType(project1.dll;Namespace.Class1)"
"project1.dll;Namespace.Class1.SystemType(System;System.Uri)"
"project1.dll;Namespace.Class1.GenericMethod1<T1,T2>(T1)"
"project1.dll;Namespace.GenericClass<T1>.GenericMethod2(T1)"
"project1.dll;Namespace.GenericClass<T1>.GenericMethod3<T2>(T1,T2)"
"project1.dll;Namespace.Class1.Class2.InnerClass(project1.dll;Namespace.Class1.Class2)"
"project1.dll;Namespace.BaseClass<T1>.Method1(T1)"
"project1.dll;Namespace.Descendant.Method1(mscorlib;string)"
"project1.dll;Class3.NoNamespace(project1.dll;Class3)"
```

are generated when the following code is part of a project compiled in assembly named "project1.dll"

```
using System;
namespace Namespace
{
    public class Class1
    {
        public void PrimitiveTypes1(string s) { }
        public void PrimitiveTypes2(String s) { }
        public void PrimitiveTypes3(System.String s) { }
        public void UserType(Class1 s) { }
        public void SystemType(Uri uri) { }
        public void GenericMethod1<T1,T2>(T1 x) { }

        public class Class2
        {
            public void InnerClass(Class2 s) { }
        }
    }
    public class GenericClass<T1>
    {
        public void GenericMethod2(T1 x) { }
        public void GenericMethod3<T2>(T1 x, T2 y) { }
    }
    public class BaseClass<T1>
    {
        public virtual void Method1(T1 x) { }
    }
    public class Descendant : BaseClass<string>
    {
        public override void Method1(string x) { }
[...]

PR #1371, opened 2018-05-01
 Setting up the analyzer, creating pb files for each analyzed method

valhristov:
Ignore InstructionBuilder, it is work in progress.

The important changes in this PR:
- adding submodule for sonar-ucfg (I updated the jenkins builds, but I didn't update VSTS)
- new way to read the additional files configuration - could be mocked in tests; I am going to extract this in separate file and update UtilityAnalyzer in another commit
- the protobuf files are generated in `.sonarqube/out/ucfg_cs` the filenames are `ucfg_{config_index}_{counter}.pb`
- MethodID format is according to #1365


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/MethodId.cs
```diff
@@ -0,0 +1,63 @@
...
+            switch (symbol.Kind)
+            {
+                case SymbolKind.Method:
+                    var method = ((IMethodSymbol)symbol).OriginalDefinition;
+                    return $"{Get(method.ContainingType)}.{method.Name}{TypeArguments(method)}({Join(method.Parameters.Select(Get))})";
+                case SymbolKind.Parameter:
+                    var p = (IParameterSymbol)symbol;
+                    return $"{Get(p.Type)}";
+                case SymbolKind.NamedType:
+                    var namedType = (INamedTypeSymbol)symbol;
```
duncanp-sonar: What about nested types?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReviewSqlQueriesForSecurityVulnerabilities.cs
```diff
@@ -37,59 +42,174 @@ public sealed class ReviewSqlQueriesForSecurityVulnerabilities : SonarDiagnostic
...
 
+        private static readonly object fileLock = new object();
+
         private static readonly DiagnosticDescriptor rule =
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
 
-        private static readonly ISet<KnownType> checkedTypes = new HashSet<KnownType>
+        public string ConfigurationIndex { get; private set; }
```
duncanp-sonar: It's not obvious what this property is for, so a doc comment would be useful.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReviewSqlQueriesForSecurityVulnerabilities.cs
```diff
@@ -37,59 +42,174 @@ public sealed class ReviewSqlQueriesForSecurityVulnerabilities : SonarDiagnostic
...
+        private void WriteUCFG(Protobuf.Ucfg.UCFG ucfg)
+        {
+            if (ucfg == null)
+            {
+                return;
+            }
+
+            var pathToWrite = Path.Combine(WorkDirectory, GetNextFileName());
+
+            lock (fileLock)
```
duncanp-sonar: Is this lock necessary? _GetNextFileName_ should return a unique name.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReviewSqlQueriesForSecurityVulnerabilities.cs
```diff
@@ -37,59 +42,174 @@ public sealed class ReviewSqlQueriesForSecurityVulnerabilities : SonarDiagnostic
...
-        private static readonly ISet<KnownType> checkedTypes = new HashSet<KnownType>
+        public string ConfigurationIndex { get; private set; }
+
+        public string WorkDirectory { get; private set; }
+
+        private bool IsEnabled => WorkDirectory != null;
+
+        public IAnalyzerConfiguration Configuration { get; }
+
+        private int index = 0;
```
duncanp-sonar: A doc comment would be useful.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReviewSqlQueriesForSecurityVulnerabilities.cs
```diff
@@ -37,59 +42,174 @@ public sealed class ReviewSqlQueriesForSecurityVulnerabilities : SonarDiagnostic
...
-                    if (c.SemanticModel.GetSymbolInfo(objectCreation).Symbol is IMethodSymbol methodSymbol &&
-                        methodSymbol.IsConstructor() &&
-                        methodSymbol.ContainingType.IsAny(checkedTypes) &&
-                        methodSymbol.Parameters.FirstOrDefault()?.Type.Is(KnownType.System_String) == true &&
-                        !IsSanitizedQuery(objectCreation.ArgumentList.Arguments[0].Expression, c.SemanticModel))
-                    {
-                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, objectCreation.Type.GetLocation()));
-                    }
-                }, SyntaxKind.ObjectCreationExpression);
+                    WriteUCFG(CreateFromMethod(c));
```
duncanp-sonar: Is there any point in registering these actions if _IsEnabled_ is false?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Security/Ucfg/MethodId_Get.cs
```diff
@@ -0,0 +1,95 @@
...
+    }
+    public class Descendant : BaseClass<string>
+    {
+        public override void Method1(string x) { }
+    }
+}
+public class Class3
+{
+    public void NoNamespace(Class3 s) { }
+}";
```
duncanp-sonar: Other test cases:
* nested namespaces
* structs
* array types
* bound generic types e.g. _IList<Class3>_
* extension methods
* nested methods - even if we can't handle these, we want to know they won't fail

I don't think it's necessary to add all of the test cases now, but may add some TODOs to list the cases that aren't covered.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/BlockId.cs
```diff
@@ -0,0 +1,38 @@
...
+using SonarAnalyzer.SymbolicExecution.ControlFlowGraph;
+
+namespace SonarAnalyzer.Security.Ucfg
+{
+    public class BlockId
+    {
+        private readonly Dictionary<Block, string> ids = new Dictionary<Block, string>();
+        private int counter;
+
+        public string Get(Block block) =>
```
duncanp-sonar: Why is the block id a string rather than an int?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/BlockId.cs
```diff
@@ -0,0 +1,38 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+using SonarAnalyzer.Helpers;
+using SonarAnalyzer.SymbolicExecution.ControlFlowGraph;
+
+namespace SonarAnalyzer.Security.Ucfg
+{
+    public class BlockId
```
duncanp-sonar: This name of this class is confusing - it sounds like it's an id for a single block, whereas it's a map of ids->blocks. _BlockIdMapper_, _BlockIdProvider_?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/MethodId.cs
```diff
@@ -0,0 +1,63 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+
+namespace SonarAnalyzer.Security.Ucfg
+{
+    public static class MethodId
+    {
+        public static string Get(ISymbol symbol)
```
duncanp-sonar: Is the size of the method id likely to be an issue?
One of the features we wrote for VS had a similar id creation scheme for methods. When we ran it against the assemblies in the .Net Framework, the longest identifier we found was over 3000 characters (that was a problem for us, because we had to store it in a database. Might not matter here).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,269 @@
...
+
+            private Expression CreateConstant() =>
+                new Expression { Const = new Constant { Value = Const } };
+
+            private ISymbol GetSymbol(SyntaxNode syntaxNode) =>
+                semanticModel.GetSymbolInfo(syntaxNode).Symbol;
+        }
+    }
+
+    public static class StackExtensions
```
Evangelink: [minor] move this class to another file.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,269 @@
...
+    {
+        public static IEnumerable<T> Pop<T>(this Stack<T> stack, int count) =>
+            PopImpl(stack, count).ToList();
+
+        public static IEnumerable<T> PopImpl<T>(this Stack<T> stack, int count)
```
Evangelink: Why is it `public`?

Evangelink: Why do we even need this `PopImpl`?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/DictionaryExtensions.cs
```diff
@@ -39,5 +40,16 @@ public static class DictionaryExtensions
 
             return defaultValue;
         }
+
+        public static TValue GetOrAdd<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key,
+            Func<TKey, TValue> factory)
```
Evangelink: [ignore me] Microsoft names it `valueFactory` :)

duncanp-sonar:
@valhristov it's getting hard to review these changes - the PR seems to cover multiple different tickets, some of which are marked as complete (e.g. https://github.com/SonarSource/sonar-csharp/issues/1365) although the code hasn't been committed. I'm finding it difficult to see what has been completely finished and what is still in progress.
You've created tickets to break the work down into smaller, more isolated pieces. Going forwards, is it possible to send a PR per ticket to make the reviews more manageable and self-contained?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/MethodId.cs
valhristov: There is a test with nested type, it works.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/BlockId.cs
valhristov: The BlockId in the UCFG definition is string.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Common/IAnalyzerConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace SonarAnalyzer.Common
+{
+    public interface IAnalyzerConfiguration
+    {
+        string GetProjectOutput(AnalyzerOptions options);
```
valhristov: Replaceable way to override the project output directory; this could replace parts of the logic in the UtilityAnalyzer base clases.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/ReviewSqlQueriesForSecurityVulnerabilitiesTest.cs
```diff
@@ -44,5 +53,90 @@ public void ReviewSqlQueriesForSecurityVulnerabilities()
...
+        {
+            var workDir = Path.Combine(TestContext.TestRunResultsDirectory, "0");
+            Directory.CreateDirectory(workDir);
+
+            Verifier.VerifyCSharpAnalyzer(@"
+class Program
+{
+    string Property2
+    {
+        set { var x = value; }
```
valhristov: It is beyond me why the setter is not walked when it is defined after the getter.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/SyntaxLocation.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace SonarAnalyzer.Security.Ucfg
+{
+    public static class SyntaxLocation
+    {
+        public static ULocation Get(Location location)
+        {
+            var lineSpan = location.GetLineSpan();
+            return new ULocation
+            {
+                FileId = location.SourceTree.FilePath,
```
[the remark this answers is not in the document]
valhristov: It would be weird if the location points to something that has no source tree. This does not run in SLVS so partially broken code should not be possible.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReviewSqlQueriesForSecurityVulnerabilities.cs
```diff
@@ -55,6 +62,28 @@ public sealed class ReviewSqlQueriesForSecurityVulnerabilities : SonarDiagnostic
...
 
+        private string protobufDirectory;
+        private int protobufFileIndex = 0;
+
+        /// <summary>
+        /// Contains the build configuration name as set by Scanner for MSBuild. Usually it
+        /// is a number. We include this in the protobuf file names because the Sonar Security
+        /// plugin is unable to read files from subfolders.
+        /// </summary>
+        private string configurationName;
```
duncanp-sonar: I found this name really confusing, even with the comment above. I'm trying to work out why that is :-) I think it's because "build configuration" generally means _debug_ or _release_ in MSBuild, and also because it's more like an id than a name.
How about _projectBuildId_ - _a unique identifier generated by the Scanner for MSBuild for this invocation of the compiler for the current project_?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReviewSqlQueriesForSecurityVulnerabilities.cs
```diff
@@ -85,11 +114,90 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                cc =>
+                {
+                    if (!TryReadConfiguration(cc.Options))
+                    {
+                        return;
+                    }
+
+                    protobufFileIndex = 0;
+
+                    cc.RegisterSyntaxNodeActionInNonGenerated(
```
duncanp-sonar: Out of interest, why did you choose to register for callbacks rather than just walking the tree?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReviewSqlQueriesForSecurityVulnerabilities.cs
```diff
@@ -85,11 +114,90 @@ protected override void Initialize(SonarAnalysisContext context)
...
+
+            if (symbol == null ||
+                symbol.IsAbstract ||
+                symbol.IsExtern ||
+                !CSharpControlFlowGraph.TryGet(getBody(declaration), context.SemanticModel, out var cfg))
+            {
+                return;
+            }
+
+            var ucfg = new UCFG
```
duncanp-sonar: It would be neater if the _UCFGBuilder_ did all of the building; it looks a bit odd have this method create the _UCFG_ object then call the builder to do three different jobs.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReviewSqlQueriesForSecurityVulnerabilities.cs
```diff
@@ -85,11 +114,90 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            // the current compilation output dir - "<root>/.sonarqube/out/<index>" where index is 0, 1, 2, etc.
+            if (basePath != null)
+            {
+                // "<root>/.sonarqube/out/0" -> "0" etc.
+                configurationName = Path.GetFileName(basePath);
+
+                // "<root>/.sonarqube/out/0" -> "<root>/.sonarqube/out/ucfg_cs"
+                protobufDirectory = Path.Combine(Path.GetDirectoryName(basePath), $"ucfg_{AnalyzerLanguage.CSharp}");
+
+                Directory.CreateDirectory(protobufDirectory);
```
duncanp-sonar: Might be worth asserting to check the assumption that the directory doesn't already exist.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/KnownMethodId.cs
```diff
@@ -0,0 +1,29 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarAnalyzer.Security.Ucfg
+{
+    public static class KnownMethodId
```
duncanp-sonar: What are these - names as specified in the UCFG? If so, _UCFGMethodId_ / _UCFGMethodKind_ would be clearer.
_Id_ is potentially confusing since there is also a _MethodIdProvider_ that does something quite different. This enum seems more like the Roslyn _SyntaxKind_.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,331 @@
...
+using SonarAnalyzer.Helpers;
+using SonarAnalyzer.Protobuf.Ucfg;
+using SonarAnalyzer.SymbolicExecution.ControlFlowGraph;
+using UcfgLocation = SonarAnalyzer.Protobuf.Ucfg.Location;
+
+namespace SonarAnalyzer.Security.Ucfg
+{
+    public class UniversalControlFlowGraphBuilder
+    {
+        public static readonly string Const = "\"\""; // Same as Java implementation
```
duncanp-sonar: Not at all clear what this is for.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,331 @@
...
+        {
+            this.semanticModel = semanticModel;
+            instructionBuilder = new InstructionBuilder(semanticModel);
+            this.cfg = cfg;
+        }
+
+        public IEnumerable<BasicBlock> Build() =>
+            cfg.Blocks.Where(IsSupportedBlock).Select(CreateBasicBlock);
+
+        public BasicBlock CreateBasicBlock(Block block)
```
duncanp-sonar: Does this need to be public? Looks like it's only called from line 49 above.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,331 @@
...
+    {
+        public static readonly string Const = "\"\""; // Same as Java implementation
+        private readonly BlockIdMap blockId = new BlockIdMap();
+        private readonly SemanticModel semanticModel;
+        private readonly InstructionBuilder instructionBuilder;
+        private readonly IControlFlowGraph cfg;
+
+        public UniversalControlFlowGraphBuilder(SemanticModel semanticModel, IControlFlowGraph cfg)
+        {
+            this.semanticModel = semanticModel;
```
duncanp-sonar: Argument checks/assertions that not null.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,331 @@
...
+        }
+
+        public IEnumerable<BasicBlock> Build() =>
+            cfg.Blocks.Where(IsSupportedBlock).Select(CreateBasicBlock);
+
+        public BasicBlock CreateBasicBlock(Block block)
+        {
+            var basicBlock = new BasicBlock
+            {
+                Id = blockId.Get(block),
```
duncanp-sonar: How about prefixing the CFG variables with _cfg_ to make it easier to tell which call graph it comes from?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,331 @@
...
+
+            if (block is JumpBlock jump)
+            {
+                instructionBuilder.BuildInstruction(jump.JumpNode, basicBlock);
+            }
+            else
+            {
+                if (block.SuccessorBlocks.All(cfg.ExitBlock.Equals))
+                {
+                    basicBlock.Ret = new Return { ReturnedExpression = new Expression { Const = new Constant { Value = Const } } };
```
duncanp-sonar: Could this _Return_ be a singleton? (partly to avoid creating unnecessary objects, but mainly to make it easier to read).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,331 @@
...
+                }
+                else
+                {
+                    basicBlock.Jump = new Jump();
```
duncanp-sonar: Is the _Jump_ class generated by the protobuf?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,331 @@
...
+            !(block is ExitBlock);
+
+        public static UcfgLocation GetLocation(SyntaxNode syntaxNode)
+        {
+            var location = syntaxNode.GetLocation();
+            var lineSpan = location.GetLineSpan();
+            return new UcfgLocation
+            {
+                FileId = location.SourceTree.FilePath,
+                StartLine = lineSpan.StartLinePosition.Line,
```
duncanp-sonar: Is the UCFG positioning zero-based?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,331 @@
...
+                }
+                else
+                {
+                    return CreateExpression(identifier);
+                }
+            }
+
+            private void BuildVariableDeclarator(VariableDeclaratorSyntax variableDeclarator, BasicBlock block)
+            {
+                if (variableDeclarator.Initializer == null)
```
duncanp-sonar: This should never be false, should it? An assertion would do.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,331 @@
...
+                {
+                    Location = GetLocation(returnStatement),
+                    ReturnedExpression = returnStatement.Expression != null ? BuildInstruction(returnStatement.Expression, block) : CreateConstant(),
+                };
+            }
+
+            private Expression BuildInvocation(InvocationExpressionSyntax invocation, BasicBlock block)
+            {
+                var methodSymbol = GetSymbol(invocation) as IMethodSymbol;
+                if (methodSymbol == null)
```
duncanp-sonar: What input would produce null here?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Common/DefaultAnalyzerConfiguration.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            var projectOutputAdditionalFile = options.AdditionalFiles
+                .FirstOrDefault(UtilityAnalyzerBase.IsProjectOutput);
+
+            if (projectOutputAdditionalFile == null)
+            {
+                return null;
+            }
+
+            return File.ReadAllLines(projectOutputAdditionalFile.Path)
```
duncanp-sonar: I don't know if it's better to read the file directly ourselves, or use the _SourceText_ that Roslyn produces (?_SourceText.ToString()_?)

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
```diff
@@ -19,6 +19,10 @@
     <Compile Include="Protobuf\AnalyzerReport.cs">
       <SonarQubeExclude>true</SonarQubeExclude>
     </Compile>
+    <Compile Remove="Protobuf\Ucfg.cs" />
+    <Compile Include="Protobuf\Ucfg.cs">
```
duncanp-sonar: Minor: it should be possible to use Item [Update](https://docs.microsoft.com/en-us/visualstudio/msbuild/item-element-msbuild) here.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
valhristov: Jump is a protobuf generated class.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
valhristov: The initializer is null when `int x;`. When `int x = 1` the initializer is not null.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
valhristov: When the `methodSymbol` is null we will not generate instruction on invocation statement. However, when the method is invoked as an argument of another expression:
`Foo(Bar(x));` we need to return "something".

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
valhristov: No idea, this is worth checking though.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Common/DefaultAnalyzerConfiguration.cs
valhristov: Perhaps. I will create a ticket about that.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReviewSqlQueriesForSecurityVulnerabilities.cs
valhristov: Usually we prefer callbacks because you don't need to create a new walker class. I personally choose walker when some state has to be kept between walking the different syntax nodes.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
valhristov: I don't know if this is going to be correctly serialized to protobuf.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Common/DefaultAnalyzerConfiguration.cs
benzonico: @valhristov  can you please add a link to this ticket ?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Common/IAnalyzerConfiguration.cs
benzonico: @valhristov can you please rename this ? @duncanp-sonar when making such statement it is always nice to come up with a suggestion for the name  : 
What about : `GetProjectOutputPath` ?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/ReviewSqlQueriesForSecurityVulnerabilitiesTest.cs
benzonico: @valhristov what does this comment mean ? is there an issue in Roslyn visit ? or in our analyzer code ? if the latter : do we have a ticket for it ?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
valhristov: Adding null check + a "pointless" test to keep up with the coverage does not seem very reasonable imho (and also there are just a few argument null checks in this project)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
benzonico: @valhristov  we probably can have a test to verify this (correctness of serialization or not) and prevent refactoring that will lead to a wrong serialization (if that is the case).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
benzonico: UCFG relies on lines starting at 1 and column starting at 1 as those are locations in the file and are as such starting with 1. (line 0 makes little sense). 
This question could (should) have been asked to me or @marcelo-sousa-sonarsource so we can improve our definition and documentation of the UCFG to help future developers to map a language to UCFG.

@valhristov can you please create a ticket to test and fix this accordingly and link it into this conversation ? 
So we can first merge the generation and then fix this ticket.
Thanks.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Security/Ucfg/MethodIdProvider_Create.cs
```diff
@@ -0,0 +1,192 @@
...
+using System;
+namespace Namespace
+{
+    public class Class1
+    {
+        public void PrimitiveTypes1(string s) { }
+        public void PrimitiveTypes2(String s) { }
+        public void PrimitiveTypes3(System.String s) { }
+
+        public void ArrayTypes(string[] s) { } // TODO: improve array type signature
```
benzonico: What does improve in that TODO mean ? shouldn't  we have a ticket for this improvement ? (and so it could be put in the TODO comment).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
valhristov: `SemanticModel` cannot be null when it is taken from the analysis context (or at least we never did null checks and we never fail for such reason). The `semanticModel` argument will be `null` if you explicitly pass `null` as parameter... Possible, but unlikely in my opinion.

benzonico:
@valhristov AFAI can tell the testcases on the ucfg looks good.

However this PR and the conversation on it is actually really tough to review and follow. Let's try to not overgrow PRs like that and to have review/conversation that leads to decision and actions.
cc @duncanp-sonar 

@valhristov I think we should address the missing test about the return "singleton" suggested by duncan before merging this pr. (and have the tickets I mention regarding line/colum created to at least add the testcases).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
valhristov: https://jira.sonarsource.com/browse/SONARSEC-48

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
valhristov: It seems the protobuf serializer able to serialize/deserialize multiple references to the same object (why did I had doubts?!). If the constant expression is a singleton though, the deserialized UCFG object will also be a singleton.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Security/Ucfg/MethodIdProvider_Create.cs
valhristov: Currently we generate `__unknown` as TypeId (see the assert below, you could look for "ArrayTypes") but while I am thinking about it, I am more convinced that we should resolve all kinds of types from the beginning: https://jira.sonarsource.com/browse/SONARSEC-49

PR #1372, opened 2018-05-01
Fix the category name

Evangelink:


sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractPropertyDefinitions.java
```diff
@@ -27,14 +27,15 @@
 
 public abstract class AbstractPropertyDefinitions {
 
-  private static final String CATEGORY = "C#";
```
Evangelink: I hardcoded C# on a previous commit when trying to fix how it was displayed for C# without thinking of VB.Net.

PR #1385, opened 2018-05-08
SONARSEC-51 Remove the assembly name from the generated method IDs

valhristov:
We are using the built-in functionality in Roslyn to generate method signatures. This also fixes the other ticket SONARSEC-49 that handles array arguments.


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReviewSqlQueriesForSecurityVulnerabilities.cs
```diff
@@ -151,16 +151,20 @@ private void WriteUCFG<TDeclarationSyntax>(SyntaxNodeAnalysisContext context, Fu
 
             var symbol = context.SemanticModel.GetDeclaredSymbol(declaration);
 
-            if (symbol == null ||
-                symbol.IsAbstract ||
-                symbol.IsExtern ||
+            var methodSymbol = (symbol is IPropertySymbol propertySymbol)
+                ? propertySymbol.GetMethod // We are in PropertyDeclarationSyntax
+                : symbol as IMethodSymbol; // all other are methods
```
valhristov: The `GetMethod` of the `propertySymbol` was retrieved in the `MethodIdProvider` class.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/ReviewSqlQueriesForSecurityVulnerabilitiesTest.cs
```diff
@@ -103,13 +103,13 @@ string Property2
...
-                    "foo.dll;Program.get_Property1()",
-                    "foo.dll;Program.get_Property2()",
-                    "foo.dll;Program.set_Property2(mscorlib;string)",
-                    "foo.dll;Program..ctor(mscorlib;string)",
-                    "foo.dll;Program.op_Addition(foo.dll;Program,mscorlib;string)",
-                    "foo.dll;Program.Method1(mscorlib;string)",
-                    "foo.dll;Program.Method2()",
+                    "Program.Property1.get",
+                    "Program.Property2.get",
+                    "Program.Property2.set",
```
valhristov: The `value` argument is not in the method signature, but that's fine because in C# you cannot declare two properties with the same name (the return type is not included in the method resolution rules)

PR #1395, opened 2018-05-10
Improve S4456 and S4457 detection

### LINKED ISSUE #1380: Fix S4457 - false positive with local functions
### Description

The S4457 rule gives a false positive on an iterator method where the iterator itself is implemented in a local function.

### Repro steps

```
private static IEnumerable<SyntaxTree> CompileSources(IEnumerable<string> sources)
{
    if (sources == null) throw new ArgumentNullException(nameof(sources));
    IEnumerable<SyntaxTree> Iterator()
    {
        foreach (string source in sources)
        {
            yield return CSharpSyntaxTree.ParseText(source);
        }
    }
    return Iterator();
}
```

### Expected behavior

No error in this code

### Actual behavior

The rule s4457 saying that I should split this method

### Known workarounds

Move the local function to a private method

### Related information

* SonarC# Version: ?
* Visual Studio Version: VS15.6.7
* SonarLint Version: 4.0.0.3479 (
* If running through the Scanner for MSBuild, its version: using the Visual Studio plugin

### LINKED ISSUE #1387: Update S4457: make message clearer
Currently the rule reports with the following message: `Split this method into two, one handling parameters check and the other handling the iterator.`. The word `iterator` should be replaced so that it is not confusing what the rule is about.

### LINKED ISSUE #1404: Update S4457: FP when func/action is async/await
### Description

Rule raises FP when using `Func` with the `async/await` pattern.

### Repro steps

```
public Task WithAsyncFunc(string s)
{
    if (s == null)
    {
        throw new ArgumentNullException(nameof(s));
    }

    Func<Task> func = async () => await Task.Delay(0);

    return func();
}
```

Evangelink:
Fix #1380, fix #1387 and fix #1404


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterValidationInYieldShouldBeWrapped.cs
```diff
@@ -81,5 +64,25 @@ public static IEnumerable<string> Get(int val)
 
             yield return "test";
         }
+
+        public static IEnumerable<TSource> TakeWhile2<TSource>(this IEnumerable<TSource> source, Func<TSource, bool> predicate)
+        {
+            if (source == null)
+            { throw new ArgumentNullException(nameof(source)); }
+            if (predicate == null)
+            { throw new ArgumentNullException(nameof(predicate)); }
```
duncanp-sonar: Minor: the indentation looks a bit weird.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterValidationInYieldShouldBeWrapped.cs
```diff
@@ -81,5 +64,25 @@ public static IEnumerable<string> Get(int val)
 
             yield return "test";
         }
+
+        public static IEnumerable<TSource> TakeWhile2<TSource>(this IEnumerable<TSource> source, Func<TSource, bool> predicate)
```
duncanp-sonar: Maybe change the method name to _TakeWhileWithLocalFunction_ to make it immediately obvious what case this method is testing.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParameterValidationInMethodShouldBeWrapped.cs
```diff
@@ -74,5 +74,18 @@ protected override sealed void Initialize(SonarAnalysisContext context)
                 },
                 RegisterSyntaxKinds);
         }
+
+        private static MethodDeclarationSyntax GetFirstEnclosingMethod(TRegisterExpressionSyntax registerExpression)
```
duncanp-sonar: From the name, I wouldn't expect this function to return null for a local function.
Not sure what a good name would be though - _GetMethodIfNotLocalFunction_ is clumsy, but more accurate.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParameterValidationInMethodShouldBeWrapped.cs
```diff
@@ -74,5 +74,18 @@ protected override sealed void Initialize(SonarAnalysisContext context)
...
+        {
+            const int localFunctionStatementRawKind = 8830;
```
duncanp-sonar: Worth calling out that this is a workaround for syntax that Roslyn v1.3.2 can't handle.
Ideally, we want to be able to find all of this type of code easily so we can remove it when we move to a newer version of Roslyn - maybe we should have a well-known comment tag (e.g. _// UNSUPPORTED SYNTAX_), or perhaps we could put all of this type of code in a separate _UnsupportedSyntaxHelper_ class.

Also, it would be useful to comment in the method what version of Roslyn is required to support the syntax.

valhristov: +1 to extract the constant to a separate class and add comments.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterValidationInAsyncShouldBeWrapped.cs
```diff
@@ -90,5 +73,26 @@ public static Task SkipLinesAsync(this TextReader reader, int linesToSkip)
                 throw new Exception("Wrong age");
             }
         }
+
+
+        public static Task SkipLinesAsync(this TextReader reader, int linesToSkip)
+        {
+            if (reader == null)
```
valhristov: If you are going to change this test, I would recommend simplifying it as well - for example, only one argument on the outer method is enough and the loop inside the local function is not necessary.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterValidationInYieldShouldBeWrapped.cs
```diff
@@ -81,5 +64,25 @@ public static IEnumerable<string> Get(int val)
...
+        {
+            if (source == null)
+            { throw new ArgumentNullException(nameof(source)); }
+            if (predicate == null)
+            { throw new ArgumentNullException(nameof(predicate)); }
+
+            return TakeWhileIterator<TSource>(source, predicate);
+
+            IEnumerable<TSource> TakeWhileIterator<TSource>(IEnumerable<TSource> s, Func<TSource, bool> p)
```
valhristov: Still, no need for such complex method imo.
```
IEnumerable<int> Iterator(int x)
{
```

PR #1405, opened 2018-05-16
SONARSEC-53 Add validation to prevent invalid UCFGs from being serialized

valhristov:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReviewSqlQueriesForSecurityVulnerabilities.cs
```diff
@@ -166,6 +182,11 @@ private void WriteUCFG<TDeclarationSyntax>(SyntaxNodeAnalysisContext context, Fu
             var ucfg = new UniversalControlFlowGraphBuilder()
                 .Build(context.SemanticModel, declaration, methodSymbol, cfg);
 
+            if (!IsValid(ucfg))
+            {
+                return;
```
Evangelink: Might be worth tracing somehow. At least adding a TODO.

PR #1407, opened 2018-05-16
Adding test for method ID of extension methods

valhristov:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Security/Ucfg/UcfgBuilder_GetMethodId.cs
```diff
@@ -114,6 +118,7 @@ public class Class3
             MethodId("Method1", skip: 1).Should().Be("Namespace.Descendant.Method1(string)");
             MethodId("InnerNamespace").Should().Be("Namespace.Inner.Class2.InnerNamespace()");
             MethodId("NoNamespace").Should().Be("Class3.NoNamespace(Class3)");
+            MethodId("Extension").Should().Be("Extensions.Extension(string, int)");
```
Evangelink: Is it on purpose that we don't have the namespace?

Evangelink: Ok sorry! Maybe you could add one test with the namespace too?

PR #1409, opened 2018-05-17
SONARSEC-63 Generate UCFG instructions for constructors

valhristov:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
```diff
@@ -194,12 +194,50 @@ private Expression BuildInstructionImpl(SyntaxNode syntaxNode)
...
+                    case SyntaxKind.ObjectCreationExpression:
+                        return BuildObjectCreation((ObjectCreationExpressionSyntax)syntaxNode);
+
                     default:
                         // do nothing
                         return ConstantExpression;
                 }
             }
 
+            private Expression BuildObjectCreation(ObjectCreationExpressionSyntax objectCreation)
```
valhristov: That's almost the same as `BuildInvocation`, will refactor in another PR

PR #1413, opened 2018-05-18
Adding test for method ID generation of nullable arguments

valhristov:


valhristov:
You cannot write `Nullable<Nullable<int>> x = null;` the compiler does not allow it

PR #1414, opened 2018-05-22
SONARSEC-66 Method IDs for methods that explicitly implement an inter…

valhristov:
…face are incorrect


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Security/Ucfg/UniversalControlFlowGraphBuilder.cs
```diff
@@ -149,7 +149,9 @@ private static UcfgLocation GetLocation(SyntaxNode syntaxNode)
         public static string GetMethodId(IMethodSymbol methodSymbol)
         {
             // Generic methods and methods declared in a base class
-            var originalSymbol = methodSymbol?.ReducedFrom ?? methodSymbol?.OriginalDefinition;
+            var originalSymbol = methodSymbol?.ExplicitInterfaceImplementations.FirstOrDefault()
```
Evangelink: Don't bother extracting this into a separate method but please add a comment on top explaining why this order and why this is split into 3 checks. For example which kind of cases do you want to catch when checking `methodSymbol?.ReducedFrom`?

PR #1418, opened 2018-05-23
SONARSEC-67 C# drop the rule S3649

### LINKED ISSUE #1386: Drop the rule S3649
The following rule as currently implemented in SonarCSharp is generating too many false-positives:
* [RSPEC-3649](https://jira.sonarsource.com/browse/RSPEC-3649) SQL queries should not be vulnerable to injection attacks

It is about to be re-implemented using the taint analysis technique in SonarSecurity, for which we expect better results. However, because the same rule key will be re-used, it is a pre-requisite to first remove it from SonarCSharp prior to be able to ship SonarSecurity in SonarQube 7.2.

valhristov:
Fix #1386


sonar-csharp-plugin/src/test/java/org/sonar/plugins/csharp/CSharpSonarWayProfileTest.java
```diff
@@ -31,17 +31,24 @@
...
-  public void test() {
+  public void sonar_security_missing() {
     CSharpSonarWayProfile profileDef = new CSharpSonarWayProfile();
     BuiltInQualityProfilesDefinition.Context context = new BuiltInQualityProfilesDefinition.Context();
     CsRules.ruleKeys = new HashSet<>();
     profileDef.define(context);
     BuiltInQualityProfilesDefinition.BuiltInQualityProfile profile = context.profile("cs", "Sonar way");
     assertThat(profile.language()).isEqualTo(CSharpPlugin.LANGUAGE_KEY);
-    // FIXME : should be null when https://github.com/SonarSource/sonar-csharp/issues/1386 is fixed
-    assertThat(profile.rule(RuleKey.of(CSharpPlugin.REPOSITORY_KEY, "S3649"))).isNotNull();
```
valhristov: Just changing the `isNotNull` to `isNull` does not work because after we set `CsRules.ruleKeys` we need to call `profileDef.define` again to repopulate the quality profile rule list. It fails if you call it twice, hence I had to either recreate the context again, or split the test into two.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReviewSqlQueriesForSecurityVulnerabilities.cs
```diff
@@ -38,29 +38,16 @@
...
-        internal const string DiagnosticId = "S3649";
-        private const string MessageFormat = "Make sure to sanitize the parameters of this SQL command.";
+        internal const string DiagnosticId = "S9999-ucfg-generator";
+        private const string Title = "UCFG generator.";
 
         private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+            DiagnosticDescriptorBuilder.GetUtilityDescriptor(DiagnosticId, Title,
+                new[] { DiagnosticDescriptorBuilder.MainSourceScopeTag });
```
valhristov: - The rule is now "utility", which means that it cannot be configured by the user and always runs.
- The rule should run only on "main" sources.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/DiagnosticDescriptorBuilder.cs
```diff
@@ -30,15 +30,16 @@ public static class DiagnosticDescriptorBuilder
...
+            string[] customTags = null) =>
             new DiagnosticDescriptor(
                 diagnosticId,
                 title,
                 string.Empty,
                 string.Empty,
                 DiagnosticSeverity.Warning,
                 isEnabledByDefault: true,
-                customTags: BuildUtilityCustomTags());
+                customTags: customTags ?? BuildUtilityCustomTags());
```
valhristov: Adding transparent ability to configure which source types should the analyzer run on.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ReviewSqlQueriesForSecurityVulnerabilities.cs
Evangelink: You are now only providing the `MainSourceScopeTag` and not the not configurable one.

PR #1441, opened 2018-06-04
Fix S4586: False positive when returning null from inside Task.Run

### LINKED ISSUE #1436: Rule S4586: False positive when returning null from inside Task.Run ?
### Description

A `null` returned from a non-async lambda expression, executed by `Task.Run<T>()`, is regarded as non-compliant by the current implementation of rule S4586.

### Actual and expected behaviour

    public Task<object> GetFooAsync()
    {
        return Task.Run(() => null) // Should be compliant, but is regarded as non-compliant
    }

### Known workarounds
One could actually adhere to the rule, which would result in using a different overload of `Task.Run`, which is able to unwrap the returned `Task`

Instead of:

    Task<TResult> Run<TResult> (Func<TResult> function)

one would be using

    Task<TResult> Run<TResult> (Func<Task<TResult>> function)

Evangelink:
Fix #1436 

I decided to ignore all return null from within a lambda but we could have looked for some specific invocations instead. This choice is motivated by having a fast fix which would only result in some FN but should cover more FPs.

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NonAsyncTaskShouldNotReturnNull.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NonAsyncTaskShouldNotReturnNull.cs
@@ -28,6 +28,21 @@ namespace Tests.Diagnostics
         }
 
         public Task Foo { get; set; }
+
+        public Task<object> GetFooAsync()
+        {
+            return Task.Run(() =>
+            {
+                if (false)
+                {
+                    return new object();
+                }
+                else
+                {
+                    return null;
+                }
+            });
+        }
     }
 
     public class NonCompliantUseCases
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NonAsyncTaskShouldNotReturnNull.cs
```diff
@@ -28,6 +28,21 @@ public Task GetTask2()
...
+                if (false)
+                {
+                    return new object();
+                }
+                else
+                {
+                    return null;
+                }
+            });
+        }
```
duncanp-sonar: There are two types of ignored enclosing syntax: please add a test for each

Also, please mark those test cases as _// False negative_

Evangelink: This test case is not an FP.

duncanp-sonar: @Evangelink your comment on the original pull request says that this is only a partial fix which will still produce some FNs (and I assumed that this lambda case was an example of an FN).
Either way, if there are known FNs that will still occur, could you call them out in test cases please?

PR #1442, opened 2018-06-05
Adding test for QP with disabled security rules

valhristov:


its/src/test/java/com/sonar/it/csharp/UCFGDeserializationTest.java
```diff
@@ -63,39 +63,41 @@ public static void initializeOrchestrator() throws Exception {
...
     } else {
       // QA environment downloads the plugin built by the CI job
-      csharpLocation = MavenLocation.of("org.sonarsource.dotnet", "sonar-csharp-plugin", securityVersion);
+      csharpLocation = MavenLocation.of("org.sonarsource.dotnet", "sonar-csharp-plugin", csharpVersion);
     }
-    builder.addPlugin(csharpLocation);
+    builder
+      .addPlugin(csharpLocation)
+      .addPlugin(MavenLocation.of("com.sonarsource.security", "sonar-security-plugin", "1.0.0.847"))
+      .addPlugin(MavenLocation.of("com.sonarsource.license", "sonar-dev-license-plugin", "3.3.0.1341"));
```
duncanp-sonar: These two plugin versions are hard-coded but the csharp one is parameterised. Can these be parameterised too?

its/src/test/java/com/sonar/it/csharp/UCFGDeserializationTest.java
```diff
@@ -131,6 +135,28 @@ public void read_ucfgs() {
...
+    orchestrator.executeBuild(getScannerForMSBuild(projectLocation)
+      .addArgument("begin")
+      .setDebugLogs(true)
+      .setProjectKey(PROJECT_KEY)
+      .setProjectName(PROJECT_KEY)
+      .setProjectVersion("1.0"));
+
+    executeDotNetCore(projectLocation, "build");
+
+    orchestrator.executeBuild(getScannerForMSBuild(projectLocation).addArgument("end"));
```
duncanp-sonar: The begin-build-analyse steps are almost identical to the ones in the _rules_enabled_ test. It's probably worth moving them into a help function.

its/src/test/java/com/sonar/it/csharp/UCFGDeserializationTest.java
valhristov: I have no idea. I doubt we need to parametrize the license plugin version - it is for private use and there does not seem to be a need to test against multiple versions.

For the security plugin, perhaps, I don't know. I suggest updating that if there is a specific need.

It seems the parameter for the sonar csharp plugin was added to enable the QA job to run against the version that was built by the CI job... We don't run ITs against multiple versions of the C# plugin.

PR #1443, opened 2018-06-05
C# Property setters of type object are not included in UCFG

valhristov:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Security/Ucfg/UcfgBuilder_Instructions.cs
```diff
@@ -183,6 +184,8 @@ public string Foo(string s)
                                         // %7 = Foo(%6)
                                         // %8 = Class1.Property.set(%7)
 
+            ObjectProperty = s;         // $9 = Class1.ObjectProperty.set(s)
```
duncanp-sonar: What's expected if _ObjectProperty_ is set to other values e.g. null or 123?

valhristov: No instruction should be added to the UCFG, I'll add a test

valhristov:
@duncanp-sonar the other PR has to be merged for the ITs to pass.

PR #1445, opened 2018-06-05
Fix S1125: FP with nullable and VS2015/MSBuild 14

valhristov:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/BooleanLiteralUnnecessary.cs
```diff
@@ -104,4 +107,16 @@ private void M()
             }
         }
     }
+
+    public class SocketContainer
+    {
+        private IEnumerable<Socket> sockets;
+        public bool IsValid =>
+            sockets.All(x => x.IsStateValid == true); // Compliant, this failed when we compile with Roslyn 1.x
```
valhristov: The problem is not reproducible in the unit tests because we compile with Roslyn 2+ and I suspect this was caused by a bug in Roslyn itself. The only way I found to test is to reference the SonarAnalyzer.CSharp nuget in a VS2015 project and paste the SocketContainer/Socket classes.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/BooleanLiteralUnnecessary.cs
```diff
@@ -264,6 +241,14 @@ private static bool IsNullableBoolean(ITypeSymbol type)
             return false;
         }
 
+        private static bool ShouldNotReport(ITypeSymbol typeLeft, ITypeSymbol typeRight)
+        {
+            return typeLeft == null
+                || typeRight == null
+                || IsNullableBoolean(typeLeft)
+                || IsNullableBoolean(typeRight);
+        }
```
valhristov: We should not report if one of the type symbols is unknown (e.g. ErrorType). The actual problem is that the type symbol of the argument in `All(x => x.IsStateValid == true)` was unknown and we raised a FP.

PR #1447, opened 2018-06-06
Fixing coverage and duplication

valhristov:
The catch block was duplicated with SonarJava and in the same time not covered by tests. I replaced the various caught exceptions with Exception and I am planning to mark the SQ issue that will appear as Won't Fix.

### DIFF
--- a/sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/CSharpSonarWayProfile.java
+++ b/sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/CSharpSonarWayProfile.java
@@ -19,6 +19,7 @@
  */
 package org.sonar.plugins.csharp;
 
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/CSharpSonarWayProfile.java
```diff
@@ -44,14 +43,8 @@ public void define(Context context) {
...
-    } catch (ClassNotFoundException e) {
-      LOG.debug("com.sonar.plugins.security.api.CsRules is not found, no security rules added to Sonar way cs profile: " + e.getMessage());
-    } catch (NoSuchMethodException e) {
-      LOG.debug("com.sonar.plugins.security.api.CsRules#getRuleKeys is not found, no security rules added to Sonar way cs profile: " + e.getMessage());
-    } catch (IllegalAccessException e) {
-      LOG.debug("[IllegalAccessException] no security rules added to Sonar way cs profile: " + e.getMessage());
-    } catch (InvocationTargetException e) {
-      LOG.debug("[InvocationTargetException] no security rules added to Sonar way cs profile: " + e.getMessage());
+    } catch (Exception e) {
+      LOG.debug("[Exception] no security rules added to Sonar way cs profile: " + e.getMessage());
```
duncanp-sonar: Could you log the class of the exception, just to preserve a similar level of detail as the previous messages? e.g. _e.getClass().getCanonicalName()_

duncanp-sonar: FYI looks like you can catch multiple exceptions in a single _catch_, as a alternative to catching _Exception_
i.e. catch (ClassNotFoundException|NoSuchMethodException| etc... )

PR #1450, opened 2018-06-06
Fix S2234: Check for parameter types before reporting it as bug

### LINKED ISSUE #496: Fix S2234: Check for parameter types before reporting it as bug
### Description

Currently rule [S2234](https://jira.sonarsource.com/browse/RSPEC-2234) - Parameters should be passed in the correct order - checks the name of the parameters but not their type.  Consider the following method signature

        void Do(DateTime date, CustomTimeType time)

Calling it with inverted parameters name is currently treated as Bug, but caller code wouldn't compile if it was. At most it would be a minor CodeSmell.

### Expected behavior

Calling it with inverted parameters name should not report an error due to type unassignability.

### Actual behavior

Detected as a Bug with default severity Major.

valhristov:
Fix #496

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParametersCorrectOrder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParametersCorrectOrder.cs
@@ -117,15 +117,14 @@ namespace SonarAnalyzer.Rules.CSharp
                 var parameterName = parameter.Name;
 
                 if (string.IsNullOrEmpty(identifierName) ||
-                    !parameterNames.Contains(identifierName))
+                    !parameterNames.Contains(identifierName) ||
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParametersCorrectOrder.cs
```diff
@@ -124,7 +124,7 @@ protected sealed override void Initialize(SonarAnalysisContext context)
 
                 if (identifierArgument is PositionalIdentifierArgument positional &&
                     (parameter.IsParams ||
-                     !identifierNames.Contains(parameterName) ||
+                     !IdentifierWithSameNameAndTypeExists(parameter) ||
```
duncanp-sonar: This is probably the most expensive check of the four being made - might be better to put it last so it only runs when necessary.

valhristov:
Named parameters are also checked now

PR #1458, opened 2018-06-07
Fix S2234: Check for parameter types before reporting it as bug

### LINKED ISSUE #496: Fix S2234: Check for parameter types before reporting it as bug

valhristov:
Fix #496

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParametersCorrectOrder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParametersCorrectOrder.cs
@@ -117,15 +117,14 @@ namespace SonarAnalyzer.Rules.CSharp
                 var parameterName = parameter.Name;
 
                 if (string.IsNullOrEmpty(identifierName) ||
-                    !parameterNames.Contains(identifierName))
+                    !parameterNames.Contains(identifierName) ||
+                    !IdentifierWithSameNameAndTypeExists(parameter))
                 {
                     continue;
                 }
 
                 if (identifierArgument is PositionalIdentifierArgument positional &&
-                    (parameter.IsParams ||
-                     !identifierNames.Contains(parameterName) ||
-                     identifierName == parameterName))
+                    (parameter.IsParams || identifierName == parameterName))
                 {
                     continue;
                 }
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParametersCorrectOrder.cs
```diff
@@ -117,15 +117,14 @@ protected sealed override void Initialize(SonarAnalysisContext context)
                 var parameterName = parameter.Name;
 
                 if (string.IsNullOrEmpty(identifierName) ||
-                    !parameterNames.Contains(identifierName))
+                    !parameterNames.Contains(identifierName) ||
+                    !IdentifierWithSameNameAndTypeExists(parameter))
```
valhristov: We are excluding the parameters that do not have a corresponding variable with the same name and type (including parameters declared in the calling method, fields, etc.).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParametersCorrectOrder.cs
```diff
@@ -149,6 +148,14 @@ protected sealed override void Initialize(SonarAnalysisContext context)
                     additionalLocations: secondaryLocations,
                     messageArgs: methodSymbol.Name));
             }
+
+            bool IdentifierWithSameNameAndTypeExists(IParameterSymbol parameter) =>
+                identifierArguments.Any(ia =>
+                    ia.IdentifierName == parameter.Name &&
+                    GetTypeSymbol(ia.ArgumentSyntax.Expression).DerivesOrImplements(parameter.Type));
```
duncanp-sonar: Could _GetTypeSymbol_ ever return null?

valhristov: It could, but `DerivesOrImplements` returns false when its argument is null (it is an extension method)

PR #1461, opened 2018-06-08
Update rule api jar version

duncanp-sonar:


duncanp-sonar:
Unfortunately, updating the rule API jar to 1.19 causes the script to fail with _Exception in thread "main" java.lang.IllegalStateException: can't find sonarpedia.json file_.
I'll discard this change for now.

PR #1465, opened 2018-06-11
Rule 4143: Dictionary and Collection values should not be replaced un…

### LINKED ISSUE #634: Rule S4143: Dictionary values should not be replaced unconditionally
Implements [RSPEC-4143](https://jira.sonarsource.com/browse/RSPEC-4143)
common rule

valhristov:
…conditionally

Fix #634


sonaranalyzer-dotnet/rspec/cs/S4143_c#.json
```diff
@@ -0,0 +1,16 @@
+{
+  "title": "Dictionary values should not be replaced unconditionally",
```
valhristov: This works for both dictionaries and collections/arrays, we should probably rename in RSPEC... It is weird that the original title says `Map` which imo translates to `Dictionary`, but all other languages renamed to `Collection`.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotOverwriteDictionaryValues.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(c =>
+            {
+                var expression = (ExpressionSyntax)c.Node;
+
+                var result = GetCollectionAndIndex(expression, c.SemanticModel);
```
valhristov: I am open for suggestions about the name of this variable

sonaranalyzer-dotnet/rspec/cs/S4143_c#.json
duncanp-sonar: Agreed - having "dictionary" in the title is definitely confusing.

"Values in indexed/keyed collections should not be..."?
"Values in dictionaries, arrays and collections should not be..."?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteDictionaryValues.cs
```diff
@@ -0,0 +1,103 @@
...
+
+namespace Tests.Diagnostics
+{
+    class TestCases
+    {
+        void Indexers(Dictionary<int, int> items)
+        {
+            int index;
+
+            items[index] = 1; // Secondary
```
duncanp-sonar: Is there a way to check that the secondary location is correct?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteDictionaryValues.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+
+            items[3] = 1; // Secondary
+            items[3] = items[5]; // Noncompliant, not the same item is read
+        }
+
+        void Methods(Dictionary<int, int> items)
+        {
+            items.Add(0, 1); // Secondary
+            items.Add(0, 2); // Noncompliant
```
duncanp-sonar: Out of interest, the second call to _Add_ will throw at runtime, won't it?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteDictionaryValues.cs
```diff
@@ -0,0 +1,103 @@
...
+            index++; // this does not break the search for elements because currently looked up key is const
+            items[1] = 2; // Noncompliant
+
+            items[2] = 1;
+            // existing item is used
+            items.Add(2, items[2]);
+
+            items[3] = 1;
+            // existing item is read
+            items[3] = items[3];
```
duncanp-sonar: Why is this line compliant? Is it a false-negative?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteDictionaryValues.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            int index;
+
+            items[index] = 1; // Secondary
+            items[index] = 2; // Noncompliant {{Verify this is the key that was intended; a value has already been set for it.}}
+//          ^^^^^^^^^^^^^^^^
+
+            items[0] = 1; // Secondary
+            items[1] = 2;
+            items[0] = 3; // Noncompliant
```
duncanp-sonar: How many errors would we expect to see raised if there is more than one duplicate for an index? 
```
items[0] = 1;
items[0] = 2;  // error here?
items[0] = 3; //... and error here?
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteDictionaryValues.cs
valhristov: Yes

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteDictionaryValues.cs
valhristov: It does not change the collection, perhaps we should raise (unless there is another rule that would raise in such cases) but we don't because we detect that the `items[3]` is read on the right hand side of the assignment.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteDictionaryValues.cs
valhristov: We will raise for each except the first. I suspect they did it because of the kill the nouse initiative, but in java they raise secondary locations for each of the subsequent assignments. Not sure if this is a good idea, because the secondary locations would show not only the first initialization, but subsequent overwrites too. I think this would be a rare case.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteDictionaryValues.cs
valhristov: If you are asking about the underline syntax, yes, we can.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotOverwriteCollectionElements.cs
```diff
@@ -0,0 +1,191 @@
...
+        {
+            var constant = semanticModel.GetConstantValue(indexSyntax);
+            if (constant.HasValue)
+            {
+                return constant;
+            }
+
+            var symbol = semanticModel.GetSymbolInfo(indexSyntax).Symbol;
+
+            // Prevent 'dict[i++] = 5' from raising issues
```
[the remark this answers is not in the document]
valhristov: FP

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotOverwriteCollectionElements.cs
```diff
@@ -0,0 +1,191 @@
...
+                    elementAccess.Expression != null &&
+                    elementAccess.ArgumentList?.Arguments.Count == 1)
+                {
+                    collection = semanticModel.GetSymbolInfo(elementAccess.Expression).Symbol;
+                    key = GetConstantOrSymbol(semanticModel, elementAccess.ArgumentList.Arguments[0].Expression);
+                }
+            }
+            else if (assignmentOrInvocation is InvocationExpressionSyntax invocation)
+            {
+                if (invocation.Expression is MemberAccessExpressionSyntax memberAccess &&
```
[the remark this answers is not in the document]
valhristov: if the statements are not merged the code looks better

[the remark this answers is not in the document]
duncanp-sonar: Please fix!

duncanp-sonar:
@valhristov did the ITs pass?

PR #1468, opened 2018-06-12
Rule S2327: "try" statements with identical "catch" and/or "finally" …

### LINKED ISSUE #257: Rule S2327: "try" statements with identical "catch" and/or "finally" blocks should be merged
Implements [RSPEC-2327](https://jira.sonarsource.com/browse/RSPEC-2327)

And that's a good example of a rule for which a quick-fix is useful

valhristov:
…blocks should be merged

Fix #257


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TryStatementsWithIdenticalCatchShouldBeMerged.cs
```diff
@@ -0,0 +1,141 @@
...
+            catch (Exception)
+            {
+            }
+            finally
+            {
+                Console.WriteLine();
+            }
+        }
+
+        void DifferentCatches()
```
duncanp-sonar: Method name is the same as the one above - should be different

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TryStatementsWithIdenticalCatchShouldBeMerged.cs
```diff
@@ -0,0 +1,141 @@
...
+            try { } // Noncompliant {{Combine this 'try' with the one starting on line 9.}}
+            catch (Exception)
+            {
+            }
+            finally { }
+
+            try { }
+            finally { }
+
+            try { } // Noncompliant {{Combine this 'try' with the one starting on line 15.}}
```
duncanp-sonar: The rule says it applies "even if there is intervening code outside the _try_ block". In this case (and in the case in line 30) the intervening code include a _try-catch_ block.
The rule isn't explicit about whether the intervening code must be "normal" code or whether it can include _try-catch_ block. What was the intention of the RSPEC?

Currently, this method is recommending merging blocks (2 and 4) and (3 and 5), which probably isn't possible. Do you think that matters? I think it probably doesn't - once the user has merged one block, the other blocks probably won't be mergeable.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TryStatementsWithIdenticalCatchShouldBeMerged.cs
```diff
@@ -0,0 +1,141 @@
+﻿using System;
+
+namespace Tests.Diagnostics
+{
+    class Program
```
duncanp-sonar: I think it's worth starting the test cases with the compliant and non-compliant examples from the RSPEC, partly so it's clear whether the rule meets the minimal spec-ed behaviour, and also as a sanity check that the RSPEC examples are well-formatted and make sense. WDYT?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TryStatementsWithIdenticalCatchShouldBeMerged.cs
```diff
@@ -0,0 +1,141 @@
...
+            try { } // Noncompliant {{Combine this 'try' with the one starting on line 102.}}
+            catch (Exception)
+            {
+            }
+            catch (ApplicationException)
+            {
+            }
+        }
+    }
+}
```
duncanp-sonar: Some other test cases:
* should this method work for properties too?
* non-compliant examples where the identical catch blocks contain multiple lines
* non-compliant examples where the identical try blocks contain multiple lines
* lambdas containing try-catch blocks that could be merged
* lambdas containing try-catch blocks in between outer blocks that could be merged
* ?nested functions - can we test for those?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TryStatementsWithIdenticalCatchShouldBeMerged.cs
valhristov: - Regarding the first question - I think it doesn't matter what's between the two try blocks.
```
try {} // #1

// whatever is here, even other try blocks

try {} // #2 should be mergeable with #1 even if there are other try blocks in between
```

- It shouldn't matter imo. If you merge two of the blocks the third one will not be mergeable unless it still has the same catch clauses.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TryStatementsWithIdenticalCatchShouldBeMerged.cs
valhristov: - done
- not really relevant imo, changed the non-compliant examples from 1 to 2 lines
- we don't really care about the try block, why write tests?
- done
- [ ] what does this mean?
- this should work
- Unfortunately the `SynaxNode.IsEquivalentTo` considers catch clauses with different variable names of the same exception type as different. E.g. `catch(Exception e)` and `catch(Exception ex)` is a FN.
- done

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TryStatementsWithIdenticalCatchShouldBeMerged.cs
valhristov: I would like this check to be automated :) The unit tests are not much different than the RSPEC examples, or am I missing a test case?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TryStatementsWithIdenticalCatchShouldBeMerged.cs
```diff
@@ -0,0 +1,200 @@
...
+            // FN - the catch clause has a name for the exception, while the try on #36 does not have a name
+            try { }
+            catch (Exception e)
+            {
+            }
+            finally { }
+
+            // exception filter
+            try { }
+            catch (Exception e) when (string.IsNullOrEmpty(e.Message))
```
duncanp-sonar: Test case: non-compliant version where the exception filters are the same

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TryStatementsWithIdenticalCatchShouldBeMerged.cs
duncanp-sonar: 2: my mistake, I think I missed the test case that with a non-empty catch block.
3: try blocks - good point, I was forgetting we didn't care about the content.
4: _lambdas containing..._ I meant a try block containing a lamba with an equivalent try block. But I guess the existing nesting tests are enough.
6: _nested functions ... this should work_ :  I guess this is just another type of nesting.

duncanp-sonar:
One additional test case for matching exception blocks

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TryStatementsWithIdenticalCatchShouldBeMerged.cs
duncanp-sonar: Yes, the RSPEC can change separately, but at least it's a point in time check that it was valid. And the code samples don't change very often.
The unit tests should definitely be far more extensive, but if the RSPEC examples are in the tests then it's a quick sanity that the rule is at least covering the obvious cases. Otherwise, you have to read the code samples in the RSPEC to check the expected behaviour, and then try to find the corresponding unit tests that cover the same logical case.

PR #1469, opened 2018-06-12
Rule S4275: Getters and setters should access the expected fields

### LINKED ISSUE #823: Rule S4275: Getters and setters should access the right field
Implement [RSPEC-4275](https://jira.sonarsource.com/browse/RSPEC-4275)

duncanp-sonar:
Fix #823


sonaranalyzer-dotnet/rspec/cs/S4275_c#.html
```diff
@@ -0,0 +1,52 @@
+<p>Properties provide a way to enforce encapsulation by providing <code>public</code>, <code>protected</code> or <code>internal</code> methods that
+give controlled access to <code>private</code> fields. However in classes with multiple fields it is not unusual that cut and paste is used to quickly
+create the needed properties, which can result in the wrong field being accessed by a getter or setter.</p>
+<p>This rule raises an issue in any of these cases:</p>
+<ul>
+  <li> A setter does not update the field with the corresponding name. </li>
+  <li> A getter does not access the field with the corresponding name. </li>
```
duncanp-sonar: I took out the two more advanced cases so the rule has exactly the same scope as the Java implementation (i.e. doesn't check for getter/setter writing to different fields, and doesn't check if multiple properties set/return the same field)

duncanp-sonar:
The ITs passed i.e. the rule didn't find any issues :-(

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,228 @@
...
+            if (!properties.Any())
+            {
+                return;
+            }
+
+            var fieldToStandardNameMap = new Dictionary<IFieldSymbol, string>();
+            foreach (var field in fields)
+            {
+                fieldToStandardNameMap[field] = GetCanonicalFieldName(field.Name);
+            }
```
duncanp-sonar: Minor optimisation - calculate the names to use for comparisons once rather than each time a comparison is done.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,228 @@
...
+
+            var allPropertyData = CollectPropertyData(properties, context.Compilation);
+
+            // Check that if there is a single matching field name it is used by the property
+            foreach (var data in allPropertyData)
+            {
+                var matchingFields = fieldToStandardNameMap.Keys.Where(k => AreCanonicalNamesEqual(fieldToStandardNameMap[k], data.CanonicalName));
+                if (matchingFields.Count() == 1)
```
duncanp-sonar: If we have multiple candidate field names we give up e.g. 
Fields = a_b and ab,
Property = AB.
We could check that the property at least uses one of the expected fields, but I don't think it's worth it (and it would make the issue message more complicated).

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,256 @@
...
+
+    partial class NonCompliant_PartialClass
+    {
+        private object myProperty;
+    }
+    partial class NonCompliant_PartialClass
+    {
+        private object anotherObject;
+    }
+    partial class NonCompliant_PartialClass
```
duncanp-sonar: I originally wrote the rule to register a syntax node action for class and struct declarations. I thought I was done, then I added a test case for _partial_ classes which failed because the issues were being reported for each partial declaration. 
So I changed the rule to use a symbol action it only fires once per definition.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,228 @@
...
+                            "setter",
+                            field.Name
+                            ));
+                    }
+                }
+            }
+        }
+
+        private static string GetCanonicalFieldName(string name)
+            => name.Replace("_", string.Empty);
```
valhristov: [cosmetic] we usually prefer the `=>` to be on the declaration line

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -116,7 +116,7 @@ public static bool IsNotIssueLocationLine(string line)
             var patternWithoutLineStart = PRECISE_ISSUE_LOCATION_PATTERN.Substring(1);
             if (Regex.IsMatch(lineText, patternWithoutLineStart))
             {
-                Execute.Assertion.FailWith("Line matches expected location pattern, but doesn't start at the beginning of the line.");
+                Execute.Assertion.FailWith($"Line matches expected location pattern, but doesn't start at the beginning of the line. Line: {line.LineNumber}");
```
duncanp-sonar: Miscellaneous: having the line number speeds up fixing formatting issues in the tests.

duncanp-sonar:
The ITs passed i.e. the rule didn't find any issues :-(

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,228 @@
...
+            var properties = symbol.GetMembers()
+                .Where(m => m.Kind == SymbolKind.Property)
+                .OfType<IPropertySymbol>()
+                .Where(p => !p.IsImplicitlyDeclared);
+            if (!properties.Any())
+            {
+                return;
+            }
+
+            var fieldToStandardNameMap = new Dictionary<IFieldSymbol, string>();
```
Evangelink: [minor] Var creation + initialization could be merged into a Linq expression ending with `ToDictionary`.

sonaranalyzer-dotnet/rspec/cs/Sonar_way_recommended_profile.json
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
duncanp-sonar: No idea. I'll remove it.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,228 @@
...
+        private static readonly DiagnosticDescriptor rule =
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            // We want to check the fields read and assigned in all properties in this class
+            // so this is a symbol-level rule (also means the callback is called only once
+            // for partial classes)
+            context.RegisterSymbolAction(c => CheckType(c),
```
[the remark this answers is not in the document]
duncanp-sonar: I've refactored slightly to move the fuzzy matching logic and caching into a helper class.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,228 @@
...
+                SymbolKind.NamedType);
+        }
+
+        private void CheckType(SymbolAnalysisContext context)
```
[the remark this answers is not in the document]
duncanp-sonar: Refactored slightly.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
duncanp-sonar: I've move this cache and the fuzzy matching logic into a separate class.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,258 @@
...
+                semanticModel.GetSymbolInfo(strippedExpression).Symbol is IFieldSymbol field)
+            {
+                return new FieldData(accessorKind, field, strippedExpression.GetLocation());
+            }
+            else
+            {
+                // Check for "this.foo"
+                if (strippedExpression is MemberAccessExpressionSyntax member &&
+                    member.Expression is ThisExpressionSyntax thisExpression &&
+                    semanticModel.GetSymbolInfo(expression).Symbol is IFieldSymbol field2)
```
valhristov: do we care about `base.foo`?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,258 @@
...
+                return null;
+            }
+
+            var strippedExpression = expression.RemoveParentheses();
+
+            // Check for direct field access: "foo"
+            if (strippedExpression is IdentifierNameSyntax &&
+                semanticModel.GetSymbolInfo(strippedExpression).Symbol is IFieldSymbol field)
+            {
+                return new FieldData(accessorKind, field, strippedExpression.GetLocation());
```
valhristov: `GetLocation()` is slow as far as I could tell and it would be better imo to call it only when you are about to report diagnostic.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,258 @@
...
+
+            public IFieldSymbol GetSingleMatchingFieldOrNull(IPropertySymbol propertySymbol)
+            {
+                // We're not caching the property name as only expect to be called once per property
+                var standardisedPropertyName = GetCanonicalFieldName(propertySymbol.Name);
+
+                var matchingFields = fieldToStandardNameMap.Keys
+                    .Where(k => AreCanonicalNamesEqual(fieldToStandardNameMap[k], standardisedPropertyName));
+
+                if (matchingFields.Count() != 1)
```
valhristov: [cosmetic] you could do `ToList()` on line 242 and then check `Count`. This way executing `Single()` at the end will not check the first item one more time.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// </summary>
+        private class PropertyToFieldMatcher
+        {
+            private readonly IDictionary<IFieldSymbol, string> fieldToStandardNameMap;
+
+            public PropertyToFieldMatcher(IEnumerable<IFieldSymbol> fields)
+            {
+                // Calcuate and cache the standardised versions of the field names to avoid
+                // calculating them every time
+                this.fieldToStandardNameMap = fields.ToDictionary(f => f, f => GetCanonicalFieldName(f.Name));
```
valhristov: Is it possible a field to be duplicated in the `fields` collection?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,258 @@
...
+        private void CheckType(SymbolAnalysisContext context)
+        {
+            var symbol = (INamedTypeSymbol)context.Symbol;
+            if (symbol.TypeKind != TypeKind.Class &&
+                symbol.TypeKind != TypeKind.Structure)
+            {
+                return;
+            }
+
+            var fields = symbol.GetMembers().Where(m => m.Kind == SymbolKind.Field).OfType<IFieldSymbol>();
```
valhristov: Is it faster to test for `Kind` first and then cast? Perhaps you should do the same on line 141.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,258 @@
...
+            symbol.GetMembers()
+                .Where(m => m.Kind == SymbolKind.Property)
+                .OfType<IPropertySymbol>()
+                .Where(p => !p.IsImplicitlyDeclared);
+
+        private static void CheckExpectedFieldIsUsed(IFieldSymbol expectedField, FieldData? actualField, SymbolAnalysisContext context)
+        {
+            if (actualField.HasValue && actualField.Value.Field != expectedField)
+            {
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(
```
valhristov: this could be reduced on less lines, even probably only one, if you extract the ternary operator to a variable.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,256 @@
...
+        public int Y
+        {
+            get { return x; }  // Noncompliant {{Refactor this getter so that it actually refers to the field 'y'}}
+//                       ^
+            set { x = value; } // Noncompliant {{Refactor this setter so that it actually refers to the field 'y'}}
+//                ^
+        }
+    }
+
+    class NonCompliantStruct_FromRspec
```
valhristov: perhaps this should be `struct`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PropertiesAccessCorrectField.cs
```diff
@@ -0,0 +1,256 @@
...
+            set { yyy = value; }  // Noncompliant {{Refactor this setter so that it actually refers to the field '__x__X'}}
+//                ^^^
+        }
+
+        public string _Y___Y_Y_
+        {
+            get { return __x__X; } // Noncompliant
+//                       ^^^^^^
+        }
+    }
```
valhristov: I am curious if we could detect the following:
```
private int x;
private int _x;
public int X { get { return _x; } }
public int _X { get { return x; } }
```

duncanp-sonar: We perhaps could if we broke the comparison into two and first looked for a case-insensitive match, and then again after we stripped the underscores. Have you seen many properties with that differ only by underscores in the name? (I hope not!)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
duncanp-sonar: It will be faster since it's just comparing an int.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
duncanp-sonar: I would find that less readable. Does reducing the number of lines matter?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
duncanp-sonar: No, I don't think so. We're currently only looking at fields on the current symbol, not inherited ones.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
duncanp-sonar: I don't believe so: that would mean Rolsyn _GetMembers_ would have to return two different field symbols with the same name.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PropertiesAccessCorrectField.cs
valhristov: I rarely see bad code :)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
valhristov: It is for style consistency, we don't put each argument on a separate line.

PR #1472, opened 2018-06-13
Update S1200: Fixed generic type counting, excluded more types from counting

### LINKED ISSUE #1180: Fix S1200: Should not count generic type parameters of extension methods
### Description

Violation of rule S1200 is raised excessively for some static classes containing extension methods that extend generic types.

### Repro steps

1. Define a class containing a few extension methods for a generic type, such as:
```csharp
public static class DictionaryExtensions
{
    public static TValue GetOrAdd<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key, Func<TValue> createValue)
    {
        if (!dictionary.TryGetValue(key, out var value))
        {
            dictionary[key] = value = createValue();
        }

        return value;
    }

    public static TValue GetValueOrDefault<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key)
    {
        dictionary.TryGetValue(key, out var obj);
        return obj;
    }

    public static TValue GetValueOrDefault<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key, TValue defaultValue)
    {
        return dictionary.TryGetValue(key, out var value) ? value : defaultValue;
    }

    public static TValue GetValueOrDefault<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key, Func<TValue> getDefaultValue)
    {
        return dictionary.TryGetValue(key, out var value) ? value : getDefaultValue();
    }

    public static IEnumerable<TValue> SelectValueByKey<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key)
    {
        if (dictionary.TryGetValue(key, out var value))
        {
            yield return value;
        }
    }

    public static ReadOnlyDictionary<TKey, TValue> AsReadOnly<TKey, TValue>(this IDictionary<TKey, TValue> dictionary)
    {
        return new ReadOnlyDictionary<TKey, TValue>(dictionary);
    }
}
```
2. Run analysis with rule S1200 activated with its default tolerance of 20

### Expected behavior

The analyzer should not request that this class be split, i.e. rule S1200 should not have been violated by that code.

### Actual behavior

Each generic type parameter of a method is considered a different dependency for the class. In turn, other generic types using the method's generic types as type parameters are considered as dependencies.

In the above example, `TKey`, `TValue`, `IDictionary<TKey, TValue>` and `Func<TValue>` are counted as separate dependencies every time they are redefined in a method.

This adds up very quickly to 21 in the above example, above the tolerance of 20, and therefore raises a rule violation.

### Related information

* SonarC# Version 6.8.1.4648
* Visual Studio Version 15.5.4

valhristov:
Fix #1180


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveClassCoupling.cs
```diff
@@ -65,150 +66,176 @@ public sealed class AvoidExcessiveClassCoupling : ParameterLoadingDiagnosticAnal
...
                     {
-                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, classDeclaration.Identifier.GetLocation(),
-                            classCouplingCount, Threshold));
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, typeDeclaration.Identifier.GetLocation(),
+                            dependentTypesCount, Threshold));
                     }
                 },
-                SyntaxKind.ClassDeclaration);
+                SyntaxKind.ClassDeclaration,
+                SyntaxKind.StructDeclaration);
```
valhristov: We are also analyzing structs now.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveClassCoupling.cs
```diff
@@ -65,150 +67,186 @@ public sealed class AvoidExcessiveClassCoupling : ParameterLoadingDiagnosticAnal
...
-                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, classDeclaration.Identifier.GetLocation(),
-                            classCouplingCount, Threshold));
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, typeDeclaration.Identifier.GetLocation(),
+                            typeDeclaration.Keyword.ValueText, dependentTypes.Count, Threshold));
                     }
                 },
-                SyntaxKind.ClassDeclaration);
+                SyntaxKind.ClassDeclaration,
+                SyntaxKind.StructDeclaration,
+                SyntaxKind.InterfaceDeclaration);
```
valhristov: Now handling struct and interface too.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/KnownType.cs
```diff
@@ -170,6 +179,7 @@ internal sealed class KnownType
         internal static readonly KnownType System_IO_StreamReader = new KnownType("System.IO.StreamReader");
         internal static readonly KnownType System_IO_StreamWriter = new KnownType("System.IO.StreamWriter");
         internal static readonly KnownType System_IO_TextWriter = new KnownType("System.IO.TextWriter");
+        internal static readonly KnownType System_Lazy = new KnownType("System.Lazy<T>");
```
valhristov: Decided that Lazy<T> is also worth excluding

sonaranalyzer-dotnet/its/expected/akka.net/Akka.Cluster-{6AB00F61-269A-4501-B06A-026707F000A7}-S1200.json
```diff
@@ -28,33 +28,7 @@
...
-"startColumn":  27,
-"endLine":  39,
-"endColumn":  49
-}
-}
-},
-{
-"id":  "S1200",
-"message":  "Split this class into smaller and more specialized ones to reduce its dependencies on other classes from 30 to the maximum authorized 20 or less.",
+"message":  "Split this class into smaller and more specialized ones to reduce its dependencies on other classes from 31 to the maximum authorized 30 or less.",
```
duncanp-sonar: The number of detected dependencies for this class went up (by one). Is this expected?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveClassCoupling.cs
```diff
@@ -65,150 +67,186 @@ public sealed class AvoidExcessiveClassCoupling : ParameterLoadingDiagnosticAnal
...
+
+        private class TypeDependencyCollector : CSharpSyntaxWalker
         {
-            var propertyReturnType = new[] { model.GetSymbolInfo(property.Type).Symbol as ITypeSymbol };
+            private readonly SemanticModel model;
+            private readonly TypeDeclarationSyntax originalTypeDeclaration;
 
-            var indexerProperty = property as IndexerDeclarationSyntax;
-            if (indexerProperty?.ExpressionBody != null)
+            public IList<ITypeSymbol> DependentTypes { get; } = new List<ITypeSymbol>();
```
duncanp-sonar: Could this be a set rather than a list?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveClassCoupling.cs
```diff
@@ -65,150 +67,186 @@ public sealed class AvoidExcessiveClassCoupling : ParameterLoadingDiagnosticAnal
...
+                SyntaxKind.InterfaceDeclaration);
         }
 
-        private static IEnumerable<ITypeSymbol> CollectCoupledClasses(FieldDeclarationSyntax field,
-            SemanticModel model)
+        private static bool IsCountableType(INamedTypeSymbol namedType) =>
+            namedType.TypeKind != TypeKind.Enum &&
+            !namedType.IsAny(ignoredTypes);
+
+        private static IEnumerable<INamedTypeSymbol> ExpandGenericTypes(INamedTypeSymbol namedType)
```
duncanp-sonar: Some documentation of the intention would be useful e.g. what is expected for _Dictionary<string, MyClass>_?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveClassCoupling.cs
```diff
@@ -65,150 +67,186 @@ public sealed class AvoidExcessiveClassCoupling : ParameterLoadingDiagnosticAnal
...
+                base.VisitPropertyDeclaration(node);
+            }
 
-        private static bool IsCountedType(ITypeSymbol type)
-        {
-            return type != null &&
-                type.TypeKind != TypeKind.Enum &&
-                type.TypeKind != TypeKind.Pointer &&
-                !type.IsAny(typesExcludedFromCoupling);
+            public override void VisitObjectCreationExpression(ObjectCreationExpressionSyntax node)
```
duncanp-sonar: Are calls to static classes captured?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveClassCoupling.cs
```diff
@@ -65,150 +67,186 @@ public sealed class AvoidExcessiveClassCoupling : ParameterLoadingDiagnosticAnal
...
 
-        private static IEnumerable<ITypeSymbol> CollectCoupledClasses(BaseMethodDeclarationSyntax method,
-            SemanticModel model)
-        {
-            return method.DescendantNodes()
-                .SelectMany(node => ExtractTypeSymbolsFromSyntaxNode(node, model));
-        }
+            public override void VisitClassDeclaration(ClassDeclarationSyntax node)
+            {
+                // don't drill down in child classes, but walk the original
```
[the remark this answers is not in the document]
valhristov: if you call the base method the walker will continue through all descendant nodes, including inner classes, thus collecting dependent types from inner classes, which I think is not expected.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveClassCoupling.cs
valhristov: it could probably save some iterations, I guess

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveClassCoupling.cs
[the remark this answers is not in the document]
duncanp-sonar: As discussed this morning, this might be worth adding as a comment in the C# RSPEC sub task

sonaranalyzer-dotnet/its/expected/akka.net/Akka.Cluster-{6AB00F61-269A-4501-B06A-026707F000A7}-S1200.json
valhristov: It is normal, we started to count some things (method return types for example) and stopped counting other - bounded generic types (e.g. IDictionary<int> and IDictionary<string> are counted as +1, not +2 as before).

valhristov:
Added the requested tests. The current class is not counted as dependency.

Attributes are metadata, imo they should not be counted (and are not)

antoinebj:
Hi,

I was the author of #1180 for which this PR represents the fix. I must have missed some updates.

Anyway, if it's not too late, I noticed the following comment in the code:
`void Foo<T>(Dictionary<string,List<T>>) where T : IEnumerable<int> should return Dictionary, string, List, IEnumerable and int`

Have you considered not counting IEnumerable in such cases?
That interface is implemented by Dictionary (and also List), meaning that if the dictionary instance is put in a variable of type IEnumerable, or passed to a private method that only needs to process it as an enumerable, it will count as an additional dependency, which hardly seems fair.
More generically, if a coupling to type A is identified, and A implements or derives from B, should B really also be counted when it is used elsewhere in the class?

valhristov:
Hi @antoinebj it is never late to create a new ticket :) 

I played a bit with Visual Studio's Code Metrics (not sure if it is available in all editions, though) and for the following code it seems to count all classes and interfaces:
```
public class Class1 // Class coupling is 3
{
    public void Foo(Dictionary<string, int> dd, IDictionary<string, int> d, IEnumerable<int> e) { }
}
```

I decided to try to be relatively close to it's way of counting. VS however, counts Task, Action, etc. types, but we are a bit more permissive and do not count some "core" classes.

antoinebj:
@valhristov 
Whether Microsoft's method is the most correct or not, I understand the need for some consistency at the expense of whatever might be perceived as more accurate.
On the other hand, it could be argued that the counting "fairness" matters more in Sonar because it's used for quality gates (in a broad sense) that may frustrate developers, whereas VS Code Metrics seems to serve more as a reporting tool that does not have much impact.
Anyway, I'll try the updated rule when it is made available and open a new issue if I have any further questions or suggestions.
Thanks :+1:

PR #1473, opened 2018-06-14
S3973: Conditionally executed single line should be denoted by indentation

### LINKED ISSUE #221: Rule S3973: Conditionally executed single line should be denoted by indentation
Implements [RSPEC-3973](https://jira.sonarsource.com/browse/RSPEC-3973)

duncanp-sonar:
Fix #221


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/IndentSingleLineFollowingConditional.cs
```diff
@@ -0,0 +1,185 @@
...
+                CheckFor, SyntaxKind.ForStatement);
+
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                CheckForEach, SyntaxKind.ForEachStatement);
+
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                CheckIf, SyntaxKind.IfStatement);
+
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                CheckElse, SyntaxKind.ElseClause);
```
duncanp-sonar: I looked at creating a single _CheckStatement_ method that was parameterised with all of the info required for each type of conditionally, but it was ugly and harder to read.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/IndentSingleLineFollowingConditional.cs
```diff
@@ -0,0 +1,185 @@
...
+                ReportIssue(context, primaryLocation, forEachStatement.Statement, "foreach");
+            }
+        }
+
+        private static void CheckIf(SyntaxNodeAnalysisContext context)
+        {
+            IfStatementSyntax ifStatement = (IfStatementSyntax)context.Node;
+
+            // Special case for "else if" on the same line.
+            // In that case, we'll check that the statement is more indented then the "else", not the "if".
```
duncanp-sonar: This special case was highlighted by ITs - it's used frequently in Nancy and Ember.
Also, it's the default indentation used by VS if you use _else if_ then hit _Enter_.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/IndentSingleLineFollowingConditional.cs
```diff
@@ -0,0 +1,185 @@
...
+            if (!IsStatementIndentationOk(elseClause, elseClause.Statement))
+            {
+                // Just highlight the "else" keyword
+                ReportIssue(context, elseClause.ElseKeyword.GetLocation(), elseClause.Statement, "else");
+            }
+        }
+
+        private static bool IsStatementIndentationOk(SyntaxNode controlNode, SyntaxNode conditionallyExecutedNode) =>
+            conditionallyExecutedNode is BlockSyntax ||
+            GetFirstCharPosition(conditionallyExecutedNode) > GetFirstCharPosition(controlNode);
```
duncanp-sonar: Just checking the character positions is not accurate enough if the indent characters contain a mix of tabs and non-tabs.
Currently, this is causing 6FPs in the ITs: some because Nancy uses Java-style formatting and so has lines like "[tab[tab]} else if", and some because the indent characters are not used consistently i.e. tabs on some lines and spaces on others.

I can fix all the FPs by doing a smarter comparison i.e. fetching the text strings for indents, comparing the number of tabs etc. Do you think it's worth it?

(FYI it looks like the _MultilineBlocksWithoutBraces_ rule might have the same indent-comparison issue, but I guess it's not a problem for the source code used in the ITs).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/IndentSingleLineFollowingConditional.cs
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        private static void CheckWhile(SyntaxNodeAnalysisContext context)
+        {
+            var whileStatement = (WhileStatementSyntax)context.Node;
+            if (!IsStatementIndentationOk(whileStatement, whileStatement.Statement))
+            {
+                // Squiggle - "while (condition1 && condition2)"
+                var primaryLocation = GetLocationBetweenTokens(whileStatement.WhileKeyword, whileStatement.CloseParenToken);
+
```
valhristov: whitespace

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/IndentSingleLineFollowingConditional.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+            else
+            {
+                controlNode = ifStatement;
+                startToken = ifStatement.IfKeyword;
+                conditionLabelText = "if";
+            }
+
+            if (!IsStatementIndentationOk(controlNode, ifStatement.Statement))
+            {
```
valhristov: whitespace

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/IndentSingleLineFollowingConditional.cs
```diff
@@ -0,0 +1,310 @@
...
+        public int Non_compliant_NestedMixedConditionals()
+        {
+            int total = 0;
+            string data = "abc";
+
+            do                
+//          ^^
+            while (total < 10)                  // comments are included in highlighting for compound statements
+//          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Secondary
+//          ^^^^^^^^^^^^^^^^^^ @-1
```
valhristov: It is not obvious if this is a primary location, e.g. usually marked with a `// Noncompliant` comment. As far as I can remember you could do `// ^^^^^^^^ Noncompliant @-1`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/IndentSingleLineFollowingConditional.cs
valhristov: Knowing that the IDE will usually reformat your code, this rule is not the most useful and I would go for the simplest solution and not spend too much time.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/IndentSingleLineFollowingConditional.cs
valhristov: Hopefully SonarQube will not complain about duplications.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/IndentSingleLineFollowingConditional.cs
duncanp-sonar: Thanks, I'll try it. Originally I had _Noncompliant_ on the line containing the error, but the test framework complained that a precise location was being specified more than once and the _Noncompliant_ should be removed. I didn't think about trying it on the ^^^ line.

duncanp-sonar: Update: `// ^^^^^ Noncompliant @-1`  didn't work. The error framework didn't complain, but the tests failed because it "the error wasn't expected" at the marked line.
It works correctly with ` // ^^^^ @-1`.  The issues is definitely being checked for correctly with that syntax (changing the number of ^ is enough to cause the test to fail because the location length isn't correct).

duncanp-sonar:
The RSPEC is about indenting _the line immediately after a conditional_, so I don't think it covers conditionals and statements on the same line.

PR #1481, opened 2018-06-18
Change ucfg test to use a separate test project

### LINKED ISSUE #1478: [Technical] Re-enabling failing UCFG integration tests
The QA build started failing recently due a file locking issue that occurred when running the integration tests.

Disabling the two UCFG fixed the QA build (#1476, 854d18ca3d36b8100b1bda814599c41b68de9173), although those tests had been working correctly before, and there did not appear to be any related code changes.

duncanp-sonar:
Attempting to fix file locking issue in UCFG tests (#1478)

Fix #1478

### DIFF
--- a/its/projects/UcfgNoProtobuf/Class1.cs
+++ b/its/projects/UcfgNoProtobuf/Class1.cs
@@ -1,0 +1,11 @@
+﻿using System;
+
+// Test project used by the test "ucfgs_not_created_when_rules_not_enabled"
+
+namespace UcfgNoProtobuf
+{
+    public class Class1
+    {
+        // TODO
+    }
+}
--- a/its/projects/UcfgNoProtobuf/Class1.cs
+++ b/its/projects/UcfgNoProtobuf/Class1.cs
@@ -6,6 +6,14 @@ namespace UcfgNoProtobuf
 {
     public class Class1
     {
-        // TODO
+        public void DoStuff()
+        {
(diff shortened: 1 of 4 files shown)


its/src/test/java/com/sonar/it/csharp/UCFGDeserializationTest.java
```diff
@@ -127,12 +127,12 @@ public void ucfgs_created_when_rules_enabled() throws IOException {
...
   }
 
-    //@Test
+  @Test
   public void ucfgs_not_created_when_rules_not_enabled() throws IOException {
     // No security rules in QP
     createQP("S100");
 
-    File projectLocation = new File("projects/SimplCommerce");
+    File projectLocation = new File("projects/UcfgNoProtobuf");
```
duncanp-sonar: Just re-enabled the test and got it to build a separate project -> QA passes.

its/projects/UcfgNoProtobuf/Class1.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+// Test project used by the test "ucfgs_not_created_when_rules_not_enabled"
+
+namespace UcfgNoProtobuf
+{
+    public class Class1
```
valhristov: I think there should be at least one method here to be sure that nothing is generated

PR #1482, opened 2018-06-18
Rule S2259: False positive on 'try/catch' with exception filter

### LINKED ISSUE #1324: Rule S2259: False positive on 'try/catch' with exception filter
### Description

Sonarqube raises a false positive when using try/catch-when.

### Repro steps
The following code raises an issue in the catch-body, because `webProxy` can be `null`. This can't happen, because the `when` ensures that the catch body is only executed when `webProxy` is not `null':

    var uri = new Uri("https://www.example.com");
    var webProxy = WebRequest.GetSystemWebProxy();
    var proxyUri = webProxy.GetProxy(uri);
    webProxy = proxyUri != uri && proxyUri != null ? new WebProxy(proxyUri) : null;
    try
    {
        Console.WriteLine($"{webProxy?.Credentials?.ToString() ?? "<no credentials>"}");
    }
    catch (WebException) when (webProxy != null)
    {
        webProxy.Credentials = new NetworkCredential("john.doe@example.com", new SecureString());
    }

### Expected behavior

SonarQube should also use the `when` clause expression.

### Known workarounds

The `when` keyword can be skipped and the code can be rewritten:

    try
    {
       ...
    }
    catch (Exception) when (obj != null)
    {
        obj.test = 0;
    }

as:

    try
    {
       ...
    }
    catch (Exception)
    {
        if (!(obj != null)) throw;
        obj.test = 0;
    }

This is functionally identical, but it introduces a rethrow.

### Related information

* SonarC# Version 6.7.1 (build 35068)
* Visual Studio Version: Visual Studio 2017 (15.6.6)

valhristov:
Fix #1324 

We were not adding the exception filters to the CFG, hence the code was evaluated as unconditional and some false positives were generated. I added `BinaryBranchBlock` to the CFG for each exception filter, connecting the catch block and the finally/after-try block. I had to add support for `CatchFilterSyntax` in the `CSharpExplodedGraph` too.

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -475,7 +475,16 @@ namespace SonarAnalyzer.SymbolicExecution.ControlFlowGraph
 
             var catchBlocks = tryStatement.Catches
                 .Reverse()
-                .Select(catchClause => BuildBlock(catchClause.Block, CreateBlock(catchSuccessor)))
+                .Select(catchClause =>
+                {
+                    Block catchBlock = BuildBlock(catchClause.Block, CreateBlock(catchSuccessor));
+                    if (catchClause.Filter?.FilterExpression != null)
+                    {
+                        catchBlock = BuildExpression(catchClause.Filter.FilterExpression,
+                            CreateBinaryBranchBlock(catchClause.Filter, catchBlock, catchSuccessor));
+                    }
+                    return catchBlock;
+                })
                 .ToList();
 
             // If there is a catch with no Exception filter or equivalent we don't want to
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereference.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereference.cs
@@ -625,19 +625,4 @@ namespace Tests.Diagnostics
         [AttributeUsage(AttributeTargets.Parameter)]
         public sealed class ValidatedNotNullAttribute : Attribute { }
     }
-
-    public class ExceptionFilters // https://github.com/SonarSource/sonar-csharp/issues/1324
-    {
-        public void Run(object o)
-        {
-            try
-            {
-                Console.WriteLine(o?.ToString());
-            }
-            catch (Exception) when (o != null)
-            {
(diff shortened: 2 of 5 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -475,7 +475,16 @@ private Block BuildTryStatement(TryStatementSyntax tryStatement, Block currentBl
 
             var catchBlocks = tryStatement.Catches
                 .Reverse()
-                .Select(catchClause => BuildBlock(catchClause.Block, CreateBlock(catchSuccessor)))
+                .Select(catchClause =>
+                {
+                    Block catchBlock = BuildBlock(catchClause.Block, CreateBlock(catchSuccessor));
+                    if (catchClause.Filter?.FilterExpression != null)
+                    {
+                        catchBlock = BuildExpression(catchClause.Filter.FilterExpression,
```
duncanp-sonar: Does the expression that is built take into account the type of the exception that is being caught when evaluating the expression?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereference.cs
```diff
@@ -625,4 +625,19 @@ public void Guarded(string s1)
...
+    {
+        public void Run(object o)
+        {
+            try
+            {
+                Console.WriteLine(o?.ToString());
+            }
+            catch (Exception) when (o != null)
+            {
+                Console.WriteLine(o.ToString());
```
duncanp-sonar: Unit tests for more complicated cases:
* catch block with condition that doesn't guard against null -> non-compliant
* multiple catch blocks, some with conditions that guard against nulls and some that don't

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
valhristov: This is syntax-based, not types are taken into account here.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereference.cs
duncanp-sonar: To check that the rule is looking at every catch block. None of the tests have multiple catch blocks, so they would all pass even if the rule was only looking at the first/last catch or stopping on the first error.

PR #1483, opened 2018-06-19
Fix FP in S1192

duncanp-sonar:
Fix 1419


sonaranalyzer-dotnet/its/expected/Nancy/Nancy.Testing-{D79203C0-B672-4751-9C95-C3AB7D3FEFBE}-S1192.json
```diff
@@ -50,165 +50,6 @@
 }
 }
 ]
-},
```
duncanp-sonar: All of the removed messages are instances of the [Obsolete("...")] attribute

sonaranalyzer-dotnet/rspec/cs/S1192_c#.html
```diff
@@ -35,5 +35,6 @@ <h2>Exceptions</h2>
 <ul>
   <li> literals with fewer than 5 characters </li>
   <li> literals matching one of the parameter names </li>
+  <li> literals used in attributes </li>
```
duncanp-sonar: I updated the RSPEC sub task and regenerated.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
```diff
@@ -64,7 +64,7 @@ protected override void Initialize(ParameterLoadingAnalysisContext context)
                     }
 
                     var stringLiterals = c.Node
-                        .DescendantNodes()
+                        .DescendantNodes(n => !n.IsKind(SyntaxKind.AttributeList))
```
duncanp-sonar: Ignoring all attributes not just _SuppressMessage_ to kill the noise.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StringLiteralShouldNotBeDuplicated.cs
```diff
@@ -6,34 +6,42 @@ public class Program
     {
         public const string NameConst = "foobar"; // Noncompliant {{Define a constant instead of using this literal 'foobar' 8 times.}}
 //                                      ^^^^^^^^
-        public static readonly string NameReadonly = "foobar"; // Secondary
+        public static readonly string NameReadonly = "foobar";
+//                                                   ^^^^^^^^ Secondary
```
duncanp-sonar: Minor: I updated the existing test to explicitly check the secondary locations.

sonaranalyzer-dotnet/rspec/cs/S1192_c#.html
valhristov: Should we restrict only to certain attributes? Like Obsolete, SuppressMessage, etc.? Attribute strings could be constants declared in the same class and sometimes it makes sense to avoid having literals...

duncanp-sonar: I thought about restricting it to known attributes but the problem was deciding which ones. There are a lot of attributes in the _System.Diagnostics_ namespace that take strings, and strings are commonly used with MEF attributes too. It seemed that it would be difficult to come up with a comprehensive list, so we'd have to revisit this rule repeatedly to add more cases -> reasonable likelihood of other false positives.

Looking at it the other way, there aren't many cases in which you would validly want to define a constant and use it in an attribute. Ids for Visual Studio packages are one, but the id is only used in two places so it wouldn't be caught by the default rule in any case. I couldn't think of any compelling cases in which you'd define a constant and use it in enough places in an attribute and the code to trigger the rule -> low likelihood of false negatives.

PR #1484, opened 2018-06-19
Adding conditional UCFG to DOT serialization

valhristov:
To enable set the following env var `SONARANALYZER_GENERATE_DOT` to `TRUE` (should be case insensitive).

Both `Visitor` classes are not exactly following the visitor pattern; I did not wanted to modify many existing classes and did simpler implementation. If we do a proper `Visitor` implementation it would be of help for the UCFG generation too.

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/BlockIdMap.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/BlockIdMap.cs
@@ -1,0 +1,34 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2018 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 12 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/BlockIdMap.cs
```diff
@@ -0,0 +1,34 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+using SonarAnalyzer.SymbolicExecution.ControlFlowGraph;
+
+namespace SonarAnalyzer.Helpers
+{
+    public  class BlockIdMap
```
valhristov: This could go away if we add an ID property to the CFG blocks

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CfgSerializer.cs
```diff
@@ -0,0 +1,147 @@
...
+                }
+
+                writer.WriteGraphEnd();
+            }
+
+            private void Visit(Block block)
+            {
+                Func<Block, string> getLabel = b => string.Empty;
+
+                if (block is BinaryBranchBlock binaryBranchBlock)
```
duncanp-sonar: Minor: could pattern matching in C#7 be used here instead i.e. a _switch_ instead of multiple _if_s?

PR #1488, opened 2018-06-20
Fix false positive in S2228

### LINKED ISSUE #1279: Fix S2228: Rule should ignore calls within DEBUG preprocessor
### Description

Rule should be tweaked to not report when a call to `Console.Write` or `Console.WriteLine` is done in a method decorated with `[Conditional ("DEBUG")]` or within a `#if DEBUG` preprocessor directive.

duncanp-sonar:
Added helper class to handle pre-processor directives

Fix #1279


duncanp-sonar:
This turned out to be quite complicated - [Conditional(DEBUG)] can be applied at class and method level, and processing hash-if-else-elif-end even approximately is not straightforward.
I should probably have left handling hash-if-else... for another time and looked for some easier tickets.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/IfDirectiveHelper.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// <summary>
+        /// Returns a list of the names of #if [NAME] sections that the specified
+        /// node is contained in.
+        /// </summary>
+        /// <remarks>
+        /// Note: currently we only handle directives with simple identifiers e.g. #if FOO, #elif FOO
+        /// We don't handle logical operators e.g. #if !DEBUG, and we don't handle cases like
+        /// #if !DEBUG ... #else... :DEBUG must be true in the else case.
+        /// </remarks>
+        public static IEnumerable<string> GetActiveConditionalCompilationSections(SyntaxNode node)
```
duncanp-sonar: Helper class to find all of the active directive processor sections before the specified node.

Directive processor syntax is awkward to handle: the keywords are embedded in leading or trailing trivia, which could be attached to a sibling node, not a parent node. Also, _DirectiveSyntaxNodes_ types do exist, but they aren't part of the syntax tree in the normal way (e.g. their _Parent_ property always _null_). It looks like the only way to hold of them is to find the trivia they are embedded in and call _trivia.GetStructure()_.

I ended up using a syntax walker to walk down the tree until reaching the target node. The problem then is that the code in the class code be like this:
```
#if DEBUG
   Console.WriteLine(); // (1) compliant
#endif

Console.WriteLine(); // (2) non-compliant
```

We can't simply stop at the first _#if DEBUG_ and say that we're in a DEBUG block, as then we'd not report on the _Console.WriteLine_ at (2) i.e. we'd have swapped a few false positives for probably a larger number of false negatives.

To be reasonably accurate, we have to keep track of the directive blocks that are currently "open".
The approach I used is:
(1) find all if/else/elif/end blocks above the current node,
(2) walk through the list, pushing onto a stack when a new block opens and popping when a block closes.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/IfDirectiveHelper.cs
```diff
@@ -0,0 +1,193 @@
...
+
+            public override void VisitEndIfDirectiveTrivia(EndIfDirectiveTriviaSyntax node)
+            {
+                AddDirective(node);
+                base.VisitEndIfDirectiveTrivia(node);
+            }
+
+            private void AddDirective(DirectiveTriviaSyntax node)
+            {
+                if (node.IsActive)
```
duncanp-sonar: A syntax node is _active_ if it is in not in an section that is excluded by a conditional compilation block.

```
#if BLOCK1     // BLOCK1 is not defined, but this #if is active
#If BLOCK2     // this #if node is not active since it's in exclude
#endif             // inactive
#endif             // active
```

sonaranalyzer-dotnet/its/expected/Nancy/Nancy-{34576216-0DCA-4B0F-A0DC-9075E75A676F}-S2228.json
```diff
@@ -4,97 +4,6 @@
 "id":  "S2228",
 "message":  "Remove this logging statement.",
 "location":  {
-"uri":  "sources\Nancy\src\Nancy\Json\JsonDeserializer.cs",
```
duncanp-sonar: Yay! we removed seven FPs, all in the same file.

PR #1489, opened 2018-06-21
Adding separate CFG block for finally blocks when there is a return s…

### LINKED ISSUE #1265: Rule S3626: "Jump statements should not be redundant" issues false positive with try/catch/finally clause
### Description

Rule S3626 produces a false positive issue with try/catch/finally clauses, see below example.

### Repro steps

```csharp
try
{
    Console.WriteLine("Try do something");
}
catch (Exception)
{
    Console.WriteLine("do something on fail before cleanup");
    return; // produces warning for S3626 despite code line at very end of snippet
}
finally
{
    Console.WriteLine("Clean up");
}

Console.WriteLine("do something on success after cleanup");
```

The jump statement is not redundant as there is a statement after the try/catch/finally that would only be executed if no exception is thrown.

### Expected behavior

No issues raised

### Actual behavior

Issue S3626 is raised.

### Known workarounds

None.

### Related information

SonarC# Version = 6.7.1
Visual Studio Version = VS 2017

valhristov:
…tatement inside the try/catch statement

Fix #1265

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
@@ -62,10 +62,15 @@ namespace SonarAnalyzer.Rules.CSharp
             var objectDisposedCheck = new ObjectDisposedPointerCheck(explodedGraph);
             explodedGraph.AddExplodedGraphCheck(objectDisposedCheck);
 
+            // Store the nodes that should be reported and ignore duplicate reports for the same node.
+            // This is needed because we generate two CFG blocks for the finally statements and even
+            // though the syntax nodes are the same, when there is a return inside a try/catch block
+            // the walked CFG paths could be different and FPs will appear.
+            var nodesToReport = new Dictionary<SyntaxNode, string>();
+
             void memberAccessedHandler(object sender, ObjectDisposedEventArgs args)
             {
-                context.ReportDiagnosticWhenActive(
-                    Diagnostic.Create(rule, args.Location, args.Name));
+                nodesToReport[args.SyntaxNode] = args.SymbolName;
             }
 
             objectDisposedCheck.ObjectDisposed += memberAccessedHandler;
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
```diff
@@ -62,10 +62,15 @@ private static void CheckForMultipleDispose(CSharpExplodedGraph explodedGraph, S
             var objectDisposedCheck = new ObjectDisposedPointerCheck(explodedGraph);
             explodedGraph.AddExplodedGraphCheck(objectDisposedCheck);
 
+            // Store the nodes that should be reported and ignore duplicate reports for the same node.
+            // This is needed because we generate two CFG blocks for the finally statements and even
+            // though the syntax nodes are the same, when there is a return inside a try/catch block
+            // the walked CFG paths could be different and FPs will appear.
+            var nodesToReport = new Dictionary<SyntaxNode, string>();
```
valhristov: False positive appeared here. We started generating two issues for the same syntax node because it is now present in two CFG blocks (each finally block generates two CFG blocks). We needed a way to deduplicate issues.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -466,11 +466,11 @@ private Block BuildTryStatement(TryStatementSyntax tryStatement, Block currentBl
...
             {
-                // Wire exit in case we have a return inside the try/catch block
-                catchSuccessor = BuildBlock(tryStatement.Finally.Block,
-                    CreateBranchBlock(tryStatement.Finally, new[] { catchSuccessor, ExitTarget.Peek() }));
+                // Create a finally block for the happy path where no exceptions are thrown
+                catchSuccessor = BuildBlock(tryStatement.Finally.Block, CreateBlock(catchSuccessor));
 
-                ExitTarget.Push(catchSuccessor);
+                // Wire another finally block to the exit target stack in case we have a return inside the try/catch block
+                ExitTarget.Push(BuildBlock(tryStatement.Finally.Block, CreateBlock(ExitTarget.Peek())));
```
valhristov: Create a separate CFG blocks for finally blocks that are called after catch and after return. This way we are correctly constructing the code paths.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RedundantJumpStatement.cs
```diff
@@ -115,5 +115,25 @@ void LongChain()
             else
             { }
         }
+
+        // https://github.com/SonarSource/sonar-csharp/issues/1265
+        void FalsePositive()
```
duncanp-sonar: _FalsePositive_ is a bit misleading - _RegressionTestForIssue1265_ might be better, and it means this class won't be found if we are doing a text search for false positives in the tests.

Issue #1459, opened 2018-06-07
S107 should not raise for PInvoke methods


lukas-ais:
### Description

S107 should not raise for methods defined according external signature, like PInvoke methods.
Because the signature is defined externally the USE of the method should not get blamed.

### Repro steps

[DllImport("netapi32.dll", EntryPoint = "NetServerEnum")]
internal static extern int NetServerEnum([MarshalAs(UnmanagedType.LPWStr)] string servername, int level, out IntPtr bufptr, int prefmaxlen, ref int entriesread, ref int totalentries, SV_101_TYPES servertype, [MarshalAs(UnmanagedType.LPWStr)] string domain, int resume_handle);

### Expected behavior

Such methods (e.g. static extern or with DllImportAttribute) should be ignored while checking S107

### Actual behavior

The method gets S107 warning.

### Known workarounds

Using "pragma warning disable" the warning can be turned off locally.

### Related information

* SonarC# Version: 7.1.0.5212

Evangelink:
Hi @lukas-ais,

Thanks for the feedback! You are absolutely right!

lukas-ais:
Okay, I'll try to solve the problem.
Let's see if I can do this...

lukas-ais:
With PR #1493 I fixed the issue, I think.
Should I close the ticket? Or are you doing? (Maybe it is required as marker for release notes or similar.)

Evangelink:
That was a missed action on our side, thanks for the reminder!

PR #1493, opened 2018-06-21
Fix S107: Do not raise for P/Invoke methods (#1459)

lukas-ais:


duncanp-sonar:
Thanks @lukas-ais!

FYI this fix will ship in the next release of the analyzer, which we're aiming to do next week or the first week in July.

lukas-ais:
Hi @duncanp-sonar 

I was wondering why the CI build was not started/completed for about 24 hours. Is this normal? Do I have to do anything else to get the build running?

Do we need any changes on the any other part of SonarQube? According to the contributors document there is a difference between "plugin" and "analyzer".

valhristov:
The CI works only for PRs initiated from the sonar-csharp repository. Otherwise someone could commit script that does bad things to the build machines (formats hard drive, downloads viruses, etc.) that will be automatically executed when a PR is opened.

duncanp-sonar:
@lukas-ais The "plugin" is the Java part of the handles the interaction with SonarQube.
The rules are completely in the C# analyzer part, so since you just changed a rule then there is nothing else to do.

PR #1494, opened 2018-06-21
Fix S2583: Rule should not report false positives when combined with …

### LINKED ISSUE #947: S2583: False-positive when Monitor.Wait is used
### Description

SonarLint detects following code as a violation of [S2583](https://rules.sonarsource.com/csharp/RSPEC-2583) due to a duplicate check with the same expression, but since `Monitor.Wait` is used in the middle the `value` may indeed change (by other thread).

Marking the `value` as `volatile` doesn't help either (but it should have no effect on the result since the lock makes sure all variables are read correctly).

### Repro steps

```c#
private bool value = false;
private readonly object mutex = new object();

private void Foo()
{
	lock (mutex)
	{
		if (value)
			return;
		Monitor.Wait(mutex);
		if (value)
			return;
	}
}
```

### Expected behavior

No issue detected.

### Actual behavior

The last check for `value` is detected as an issue.

### Known workarounds

Suppress the warning manually.

### Related information

* SonarC# 3.7.0.2645
* Visual Studio 15.4.4

### LINKED ISSUE #1400: Fix S2583: Rule should not report false positives when combined with async methods
### Description

Using `await` should stop detection of the same condition checks if a variable (field) from outer scope is used in the condition.

### Repro steps

```csharp
class FooBar
{
  int foo = 0;

  public async Task Bar()
  {
    var i = ++foo;
    await Task.Delay(100);
    if (i != foo)
      return;
    await Task.Delay(100);
    if (i != foo) // <-- S2583 reported here
      return;
    Baz();
  }

  void Baz() {}
}
```

Even though method from another class is called in the example (and without async operation it cannot change the `foo` field) it might get changed before the method resumes asynchronously (`foo` might get changed possibly by calling `Bar` again before the delay has finished).

### Expected behavior

No warnings should be generated.

### Actual behavior

Following warning is generated:

> Warning S2583 Change this condition so that it does not always evaluate to 'false'; some subsequent code is never executed.

### Known workarounds

Add a suppression.

### Related information

* SonarC# 4.0.0.3479
* Visual Studio 15.7.1

valhristov:
…async methods

Fix #1400 

Fix #947

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
@@ -277,8 +277,12 @@ namespace SonarAnalyzer.SymbolicExecution
                 case SyntaxKind.RefValueExpression:
 
                 case SyntaxKind.MemberBindingExpression:
+                    newProgramState = newProgramState.PopValue();
+                    newProgramState = newProgramState.PushValue(new SymbolicValue());
+                    break;
 
                 case SyntaxKind.AwaitExpression:
+                    newProgramState = newProgramState.RemoveSymbols(IsFieldSymbol);
                     newProgramState = newProgramState.PopValue();
                     newProgramState = newProgramState.PushValue(new SymbolicValue());
                     break;
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
@@ -441,7 +441,7 @@ namespace SonarAnalyzer.SymbolicExecution
                         var invocationVisitor = new InvocationVisitor(invocation, SemanticModel, newProgramState);
                         newProgramState = invocationVisitor.ProcessInvocation();
 
-                        if (invocation.Expression.IsOnThis() && !invocation.IsNameof(SemanticModel))
+                        if (!invocation.IsNameof(SemanticModel))
                         {
                             newProgramState = newProgramState.RemoveSymbols(IsFieldSymbol);
                         }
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
```diff
@@ -277,8 +277,12 @@ protected override void VisitInstruction(ExplodedGraphNode node)
                 case SyntaxKind.RefValueExpression:
 
                 case SyntaxKind.MemberBindingExpression:
+                    newProgramState = newProgramState.PopValue();
+                    newProgramState = newProgramState.PushValue(new SymbolicValue());
+                    break;
 
                 case SyntaxKind.AwaitExpression:
+                    newProgramState = newProgramState.RemoveSymbols(IsFieldSymbol);
```
valhristov: Awaiting a task potentially could change the fields, so to avoid FP we clear the field symbols from the current program state.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.cs
```diff
@@ -1471,4 +1471,15 @@ public void FalseNegatives()
...
         }
     }
+
+    class AsyncAwait
+    {
+        string x;
+        async Task Foo(Task t)
+        {
+            x = null;
+            await t; // awaiting clears the constraints
```
duncanp-sonar: Unit test to check awaiting doesn't clear local variables?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
```diff
@@ -437,7 +441,7 @@ protected override void VisitInstruction(ExplodedGraphNode node)
                         var invocationVisitor = new InvocationVisitor(invocation, SemanticModel, newProgramState);
                         newProgramState = invocationVisitor.ProcessInvocation();
 
-                        if (invocation.Expression.IsOnThis() && !invocation.IsNameof(SemanticModel))
+                        if (!invocation.IsNameof(SemanticModel))
```
valhristov: We never know if some static method does not change us...

PR #1495, opened 2018-06-22
S4055 should not raise for string literals in debug code

### LINKED ISSUE #1464: S4055 should not raise issues for string literal used in the 'message' of Debug.Assert
S4055 requests all string literals to be localized. While it's fine to request end-user messages to be localized, It's cumbersome to request all debug logs to be also localized.

S4055 should not raise an issue for such code:

```
void DoStuff()
{
  Debug.Assert(o != null, "Type parameter is null");
  Debug.WriteLine(...)
  Debug.Write(...)
}

[Conditional("DEBUG")]
void WriteDebugLog(string text)
{
  ...
  Console.WriteLine(text);
  ...
}

void Caller()
{
  WriteDebugLog("a message") // ok - calling debug-only method
}
```

duncanp-sonar:
Fix #1464 

As discussed offline:
 - the rule won't raise for any method on _Debug_ (not just _Debug.Assert_)
 - the rule won't raise for any call from or to a method marked with _Conditional("DEBUG")_

We don't check whether code is in  #if DEBUG blocks at the moment. We could add it later if we are still getting FPs.


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/DebugOnlyCodeHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Helpers
+{
+    internal static class DebugOnlyCodeHelper
+    {
+        // Looking for an exact case-sensitive match
+        public static bool IsDebugString(string text) =>
```
duncanp-sonar: Moved from the _ConsoleLogging_ rule into a helper class.

I've deleted some of the code as I realised I'd made a mistake in the original version. The MSDN documentation for _ConditionalAttribute_ says it can be applied to methods or classes. However, the small-print in the documentation says it does nothing at class level unless it's on an attribute class, so I removed the walk up the class hierarchy to check for _[Conditional(DEBUG)]_ at class level.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConsoleLogging.cs
```diff
@@ -76,58 +76,4 @@ public static void LogDebugC(string message)
             Console.WriteLine(); // Noncompliant - wrong case
         }
     }
-
-    [Conditional("DEBUG")]
```
duncanp-sonar: Deleted the now-redundant tests for the attribute being applied at class level.

PR #1497, opened 2018-06-25
Update S4261: should not raise on async unit test methods; rule now a…

### LINKED ISSUE #1366: Update S4261: should not raise on async unit test methods
### Description

Rule S4261 (Methods should be named according to their synchronicities) should not suggest putting Async on the end of method names that are async unit tests. The names of unit tests should be selected based on a strategy that describes what is being tested to produce sensible results in test reports / test explorer.

### Repro steps

    [TestFixture]
    public class NameOfTestClass
    {              
        [Test]
        public async Task NameOfTestMethod()
        {
        }
    }

### Expected behavior

NameOfTestMethod() doesn't cause warning S4261 because it is a unit test method

### Actual behavior

Warning S4261 is applied

### Known workarounds

Suppress the warning in the rules for unit tests projects, but this also suppresses it for other async methods in the project that aren't tests. 

Suppress the warning individually on each test

### Related information

* SonarC# Version 7.1
* Visual Studio Version 15.6

A complication to implementing this would be recognizing it as a unit test. Ignoring methods with MSTest,NUnit and XUnit attributes on the method would cover most cases.

valhristov:
…pplies to test methods as well

Fixed #1366


duncanp-sonar:
Looks good, but the expected results for the ITs need updating

duncanp-sonar:
Since the rule excludes test methods is only raising on test helper methods, I think it's ok if it raises more issues.

valhristov:
@duncanp-sonar it's done.

PR #1499, opened 2018-06-25
Fix S4023: False positive on interfaces that derive from other non-em…

### LINKED ISSUE #1378: Fix S4023: False positive on interfaces that derive from other non-empty interfaces
### Description

Rule S4023 is violated if an interface is defined only as the derivative of other interfaces, without declaring members of its own.

AFAIK, such interfaces are useful for aggregating contracts and do not seem to be part of the code smell of "marker interfaces". For example, I might want to aggregate the contracts of `INotifyCollectionChanged` and `IReadOnlyList<T>` into `INotifyingReadOnlyList<T>`.

If doing that were a code smell, and I am not aware that it is, it would at least be a different code smell from the one mentioned in [RSPEC-4023](https://jira.sonarsource.com/browse/RSPEC-4023). The description of the rule would have to be updated with a workaround that is applicable, because custom attributes cannot fulfill that requirement.

In any case, the interface cannot be considered as empty, it inherits from all the members of the interfaces from which is derives. The algorithm for finding violations of this rule should therefore also count all inherited members.

### Repro steps

1. Add this interface definition to any solution:

```csharp
public interface INotifyingReadOnlyList<out T> : IReadOnlyList<T>, INotifyCollectionChanged
{
}
```

2. Run Sonar analysis

### Expected behavior

Sonar does not detect a violation of rule S4023 by the above code.

### Actual behavior

Sonar detects a violation of rule S4023 by the above code.

### Related information

* SonarC# 6.8.1.4648
* Visual Studio 15.6.7

valhristov:
…pty interfaces

Fix #1378

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InterfacesShouldNotBeEmpty.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InterfacesShouldNotBeEmpty.cs
@@ -45,17 +45,24 @@ namespace SonarAnalyzer.Rules.CSharp
                 c =>
                 {
                     var interfaceDeclaration = (InterfaceDeclarationSyntax)c.Node;
-                    var interfaceSymbol = c.SemanticModel.GetDeclaredSymbol(interfaceDeclaration);
+                    if (interfaceDeclaration.Identifier.IsMissing ||
+                        interfaceDeclaration.Members.Count > 0)
+                    {
+                        return;
+                    }
 
+                    var interfaceSymbol = c.SemanticModel.GetDeclaredSymbol(interfaceDeclaration);
                     if (interfaceSymbol != null &&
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InterfacesShouldNotBeEmpty.cs
```diff
@@ -45,17 +45,24 @@ protected override void Initialize(SonarAnalysisContext context)
                 c =>
                 {
                     var interfaceDeclaration = (InterfaceDeclarationSyntax)c.Node;
-                    var interfaceSymbol = c.SemanticModel.GetDeclaredSymbol(interfaceDeclaration);
+                    if (interfaceDeclaration.Identifier.IsMissing ||
+                        interfaceDeclaration.Members.Count > 0)
```
valhristov: The analyser should be a little faster if we don't immediately go for the symbol

PR #1515, opened 2018-07-02
Add target object as first argument to method call in uCFG

Evangelink:


Evangelink:
I have extracted the few classes that were bothering me. I also undo the changes that were forcing to return a SyntaxNode as it wasn't allowing to pass the first argument (for this + static).

If you could focus your review (first) on test cases that would be nice!

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionBuilder.cs
```diff
@@ -0,0 +1,315 @@
...
+            // A call to a constructor should look like:
+            // %X := new Ctor()
+            // %X+1 := Ctor_MethodId [ params ]
+            // variable := __id [ %X ]
+            // Because all UCFG instructions associate the processed SyntaxNode to the variable created, we need to:
+            // 1. create the variable linked to the new
+            // 2. create the method call
+            // 3. create the new call (this will override to which variable the node is associated)
+            // 4. return the instructions in the right order
+            var initializationVariable = CreateTempVariable();
```
Evangelink: If you think of a better way, please suggest!

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Security/Ucfg/UcfgBuilder_Instructions.cs
```diff
@@ -291,16 +291,10 @@ public class Class1
...
         {
             Class1 c;
-            c = new Class1(s);              // %0 := Namespace.Class1.Class1(string) [ s ]
+            c = new Class1(s);              // %0 := __id [ const ]
+                                            // %1 := Namespace.Class1.Class1(string) [ s ]
+                                            // c := __id [ %0 ]
 
-            c = new Class1();               // ignored, does not accept or return string
-
-            c = new Class1(new Class1(s));  // %1 := Namespace.Class1.Class1(string) [ s ]
```
valhristov: These tests should be present imo

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionBuilder.cs
```diff
@@ -0,0 +1,315 @@
...
+
+        private string CreateTempVariable() =>
+            $"%{tempVariablesCounter++}";
+
+        private IEnumerable<Instruction> CreateFromObjectCreationExpression(ObjectCreationExpressionSyntax objectCreationExpression)
+        {
+            var ctorSymbol = GetSymbol(objectCreationExpression) as IMethodSymbol;
+            if (ctorSymbol == null)
+            {
+                return new[] { CreateConstant(objectCreationExpression) };
```
valhristov: CreateConstant could return IEnumerable

PR #1524, opened 2018-07-03
SONARSEC-105: Translate constructor calls to uCFG

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionBuilder.cs
```diff
@@ -345,15 +346,30 @@ public IEnumerable<Instruction> CreateConstant(SyntaxNode syntaxNode)
         private Instruction CreateInstruction(SyntaxNode syntaxNode, UcfgIdentifier method, string returnVariable,
             params Expression[] arguments)
         {
+            // Associate node to the UCFG variable
+            CreateVariable(syntaxNode, returnVariable);
+
+            // Create the instruction
+            if (syntaxNode is ObjectCreationExpressionSyntax)
```
valhristov: Why not two methods? The only code that's shared is the `CreateVariable`... I would rather prefer `CreateNewObjectInstruction` and `CreateMethodCallInstruction`

PR #1531, opened 2018-07-04
SONARSEC-100 CFG does not correctly represent || and && in if, do, while and ternary operator conditions

valhristov:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -1071,6 +1068,53 @@ private Block BuildVariableDeclarator(VariableDeclaratorSyntax variableDeclarato
...
         #endregion Create*
 
+        #region Condition
+
+        private Block BuildCondition(ExpressionSyntax expression, Block trueSuccessor, Block falseSuccessor)
+        {
+            expression = expression.RemoveParentheses();
+
+            var binaryExpression = expression as BinaryExpressionSyntax;
+            if (binaryExpression != null)
```
Evangelink: You could use the new syntax instead of `as` + null check.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -1071,6 +1068,53 @@ private Block BuildVariableDeclarator(VariableDeclaratorSyntax variableDeclarato
...
+                        return BuildCoalesceCondition(binaryExpression, trueSuccessor, falseSuccessor);
+                }
+            }
+
+            // Fallback to generating an additional branch block for the if statement itself.
+            return BuildExpression(expression,
+                    AddBlock(new BinaryBranchBlock(expression, trueSuccessor, falseSuccessor)));
+        }
+
+        private Block BuildCoalesceCondition(BinaryExpressionSyntax coalesce, Block trueSuccessor, Block falseSuccessor)
```
Evangelink: I don't think these methods brings a lot. As the content is quite small I'd rather have them directly in the cases. IF you prefer to keep the methods maybe they should be inner methods.

PR #1535, opened 2018-07-05
SONARSEC-116: Translate constructor chaining to uCFG

valhristov:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -77,16 +77,39 @@ protected override void PostProcessGraph()
...
+                block = BuildExpression(expression, currentBlock);
             }
 
-            throw new ArgumentException("Neither a statement, nor an expression", nameof(node));
+            if (block == null)
+            {
+                throw new ArgumentException("Neither a statement, nor an expression", nameof(node));
+            }
+
+            if (node.Parent is ConstructorDeclarationSyntax constructorDeclaration &&
```
valhristov: If everything is alright, add the ctor chain to the CFG

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
```diff
@@ -440,6 +440,16 @@ protected override void VisitInstruction(ExplodedGraphNode node)
                     }
                     break;
 
+                case SyntaxKind.BaseConstructorInitializer:
+                case SyntaxKind.ThisConstructorInitializer:
+                    var constructorInitializer = (ConstructorInitializerSyntax)instruction;
+                    if (constructorInitializer.ArgumentList != null)
+                    {
+                        var ctorInifializerArgumentsCount = constructorInitializer.ArgumentList.Arguments.Count;
+                        newProgramState = newProgramState.PopValues(ctorInifializerArgumentsCount);
```
valhristov: Since the initializer arguments were added to the CFG, we generated symbolic values, which are on the stack. To ensure integrity of the stack we need to pop the values when we are walking the initializer instruction, as if we did a method invocation.

PR #1539, opened 2018-07-09
SONARSEC-119: Translate field access to uCFG

Evangelink:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Security/Ucfg/UcfgBuilder_Instructions.cs
```diff
@@ -44,12 +44,15 @@ public void Foo(string s)
...
+            //var x = new string[5];
             int i;
             i = 5;                  // i := __id [ const ]
 
-            field = s;              // ignored
-            Class1.staticField = s; // ignored
+            field = s;              // this.field := __id [ s ]
+            this.field = s;         // this.field := __id [ s ]
+
+            staticField = s;        // Namespace.Class1.staticField := __id [ s ]
```
duncanp-sonar: Edge case - static fields of static classes referenced using the _using static MyNamespace.MyStaticClass;_ syntax

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgExpressionService.cs
```diff
@@ -19,31 +19,66 @@
...
 using SonarAnalyzer.Helpers;
-using SonarAnalyzer.Protobuf.Ucfg;
 
 namespace SonarAnalyzer.ControlFlowGraph.CSharp
 {
     internal class UcfgExpressionService
     {
-        private readonly Dictionary<SyntaxNode, Expression> syntaxNodeToUcfgExpressionCache
-            = new Dictionary<SyntaxNode, Expression>();
+        private static readonly ISet<KnownType> PrimitiveTypes = new[] { KnownType.System_Boolean }
```
duncanp-sonar: Minor question about naming: is the important thing that these are primitive types, or is that these are the variable types we handle in the CFG that just happen to all be primitive types at the moment? If is the latter then something like _SupportedVariableTypes_ might be a better name.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgExpressionService.cs
```diff
@@ -19,31 +19,66 @@
...
+            = new Dictionary<SyntaxNode, UcfgExpression>();
 
-        public void RegisterAsVariable(SyntaxNode syntaxNode, string variableName) =>
-            syntaxNodeToUcfgExpressionCache[syntaxNode] = UcfgExpression.FromVariable(variableName);
+        private int numberedVariableCounter;
 
-        public void RegisterAsConstant(SyntaxNode syntaxNode) =>
-            syntaxNodeToUcfgExpressionCache[syntaxNode] = UcfgExpression.Constant;
+        public void Associate(SyntaxNode syntaxNode, UcfgExpression expression) =>
+            cache[syntaxNode.RemoveParentheses()] = expression;
```
duncanp-sonar: Are there any test cases that cover expressions that differ only by parentheses?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/ControlFlowGraph/UcfgBusinessException.cs
```diff
@@ -0,0 +1,48 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Runtime.Serialization;
+
+namespace SonarAnalyzer.ControlFlowGraph
+{
+    [Serializable]
+    public sealed class UcfgBusinessException : Exception
```
duncanp-sonar: Minor: having "Business" in the name feels a bit odd. Why not just _UcfgException_?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgExpression.cs
```diff
@@ -18,23 +18,191 @@
...
-            new Expression { Var = new Variable { Name = variableName } };
+        public abstract Expression Expression { get; }
+        public ITypeSymbol TypeSymbol { get; }
 
-        public static Expression FromNamedType(INamedTypeSymbol namedType) =>
-            new Expression { Classname = new ClassName { Classname = UcfgIdentifier.CreateTypeId(namedType).ToString() } };
+        public virtual void ApplyAsTarget(Instruction instruction) =>
+            throw new UcfgBusinessException($"Type {GetType().Name} could not be applied as target of the instruction.");
+
+        internal class UnknownExpression : ConstantExpression
```
duncanp-sonar: ?Could be private? The parent class will still be able to construct it.
Not sure if the other sub-classes could be _private_ too or just _sealed_.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgIdentifier.cs
```diff
@@ -95,6 +95,13 @@ public static UcfgIdentifier CreateMethodId(IMethodSymbol methodSymbol)
             }
         }
 
+        public static bool IsMethodId(UcfgIdentifier identifier) =>
```
valhristov: That's not used.

PR #1547, opened 2018-07-10
SONARSEC-144: Build uCFG for operators declaration

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UcfgGenerator.cs
```diff
@@ -95,13 +95,16 @@ protected override void Initialize(SonarAnalysisContext context)
...
-                        SyntaxKind.ConstructorDeclaration,
-                        SyntaxKind.OperatorDeclaration);
+                        SyntaxKind.ConstructorDeclaration);
 
                     cc.RegisterSyntaxNodeActionInNonGenerated(
                         c => WriteUCFG<MethodDeclarationSyntax>(c, x => (CSharpSyntaxNode)x.Body ?? x.ExpressionBody?.Expression),
                         SyntaxKind.MethodDeclaration);
 
+                    cc.RegisterSyntaxNodeActionInNonGenerated(
+                        c => WriteUCFG<OperatorDeclarationSyntax>(c, x => (CSharpSyntaxNode)x.Body ?? x.ExpressionBody?.Expression),
```
Evangelink: You were already handling operators but not when using the expression body so I have updated the generation.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UcfgGenerator.cs
```diff
@@ -95,13 +95,16 @@ protected override void Initialize(SonarAnalysisContext context)
 
                     cc.RegisterSyntaxNodeActionInNonGenerated(
                         c => WriteUCFG<BaseMethodDeclarationSyntax>(c, x => x.Body),
-                        SyntaxKind.ConstructorDeclaration,
-                        SyntaxKind.OperatorDeclaration);
+                        SyntaxKind.ConstructorDeclaration);
```
valhristov: `WriteUCFG<BaseMethodDeclarationSyntax>` should be better with `WriteUCFG<ConstructorDeclarationSyntax>`

PR #1551, opened 2018-07-10
SONARSEC-155: Translate new type array to classic object creation

duncanp-sonar:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Security/Ucfg/UcfgBuilder_Instructions.cs
```diff
@@ -109,7 +109,7 @@ public string Bar(string s, Func<string, string> a)
         }
 
         [TestMethod]
-        public void Assignments_Static_Poperty_On_Generic_Class()
+        public void Assignments_Static_Property_On_Generic_Class()
```
duncanp-sonar: Unrelated: fixing typo in method name

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Security/Ucfg/UcfgBuilder_Instructions.cs
```diff
@@ -787,6 +787,58 @@ public void Foo(Class1 other)
...
+
+            // Array of objects
+            var x1 = new Class1[0];     // %1 := new Namespace.Class1[]
+                                        // x1 := __id [ %1 ]
+
+            // Multi-rank arrays
+            var x2 = new int[10,2];     // %2 := new int[*,*]
+                                        // x2 := __id [ %2 ]
+
+            // Arrau with initializer (initializer is ignored)
```
duncanp-sonar: Oops, typo in "arrau".
The initializer will be handled in a separate ticket (161).

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Security/Ucfg/UcfgBuilder_Instructions.cs
```diff
@@ -787,6 +787,58 @@ public void Foo(Class1 other)
...
+        public void ArrayCreation_WithCreateInstance()
+        {
+            const string code = @"
+namespace Namespace
+{
+    public class Class1
+    {
+        public void Foo()
+        {
+            var a = System.Array.CreateInstance(typeof(string), 10);    // %0 = System.Array.CreateInstance(System.Type, int) [ System.Array.CreateInstance ]
```
duncanp-sonar: I thought the problem was the _typeof_ but it isn't getting that far.
Currently the code is failing when it tries to treat the namespace syntax node for _System_ as a named-type node.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgMethodId.cs
```diff
@@ -117,6 +117,9 @@ public static UcfgMethodId CreateMethodId(IMethodSymbol methodSymbol)
         public static UcfgMethodId CreateTypeId(INamedTypeSymbol typeSymbol) =>
             Create(typeSymbol.ConstructedFrom.ToDisplayString());
 
+        public static UcfgMethodId CreateArrayTypeId(IArrayTypeSymbol arrayTypeSymbol) =>
```
duncanp-sonar: I find it weird that _CreateTypeId_ and _CreateArrayTypeId_ return a type called "method id".

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionFactory.cs
```diff
@@ -180,6 +182,23 @@ private IEnumerable<Instruction> ProcessObjectCreationExpression(ObjectCreationE
                     new[] { expressionService.GetExpression(objectCreationExpression) }.Concat(arguments).ToArray()));
         }
 
+        private IEnumerable<Instruction> ProcessArrayCreationExpression(ArrayCreationExpressionSyntax arrayCreationExpression)
+        {
+            var arrayTypeSymbol = semanticModel.GetTypeInfo(arrayCreationExpression).Type as IArrayTypeSymbol;
+            if (arrayTypeSymbol == null)
+            {
+                return NoInstructions;
```
valhristov: You should associate the `arrayCreationExpression` with `const` here. Otherwise the following could throw:
```
Foo(new string[]);
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Security/Ucfg/UcfgBuilder_Instructions.cs
```diff
@@ -787,6 +787,58 @@ public void Foo(Class1 other)
...
+                                        // x2 := __id [ %2 ]
+
+            // Array with initializer (initializer is ignored currently - issue #161)
+            var x3 = new string[] { ""aaa"", ""bbb"", ""ccc"" };    // %3 := new string[]
+                                                                    // x3 := __id [ %3 ]
+        }
+    }
+}";
+
+            var x3 = new string[] { "aaa", "bbb", "ccc" };
```
duncanp-sonar: Redundant line - I'll remove it.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionFactory.cs
duncanp-sonar: I've added a test that creates an array of a type that doesn't exist, thinking it might hit this code path. It doesn't - we can still get the _ArrayTypeSymbol_, we just don't know what the _ElementType_ is.

PR #1559, opened 2018-07-13
Improve UCFG error reporting

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgFactory.cs
```diff
@@ -66,7 +66,14 @@ public UCFG Create(SyntaxNode syntaxNode, IMethodSymbol methodSymbol, IControlFl
             }
             catch (System.Exception e)
             {
-                throw new UcfgException(e.ToString().Replace("\r\n", " \\ "));
+                var message = $"Type {GetType().Name} could not be applied as target of the instruction. " +
```
Evangelink: This is quite weird beginning of error message. We can have any kind of error so stating this is an error about the target of instruction looks weird to me.

PR #1565, opened 2018-07-16
Fixed SONARSEC-171: unexpected MergedNamespaceSymbol when creating UCFG

duncanp-sonar:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionFactory.cs
```diff
@@ -127,7 +127,7 @@ private IEnumerable<Instruction> ProcessConstructorInitializer(ConstructorInitia
             var chainedCtor = GetSymbol(constructorInitializer) as IMethodSymbol;
             if (chainedCtor == null)
             {
-                return Enumerable.Empty<Instruction>();
+                return NoInstructions;
```
duncanp-sonar: Unrelated - changing existing code to use _NoInstructions_.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Security/Ucfg/UcfgBuilder_Instructions.cs
```diff
@@ -598,6 +598,168 @@ public void Foo()
             UcfgVerifier.VerifyInstructions(code, "Foo");
         }
 
+        [TestMethod]
+        public void Invocations_QualifiedNames_Methods()
```
duncanp-sonar: I've added comprehensive unit tests for methods, property gets and sets 
i.e.
* global methods
* static global methods
* single and multiple namespaces
* nested classes

I haven't added a test for inner functions as they are currently unsupported by the C# CFG.
I'll create a separate SONARSEC ticket for this - the UCFG analyzer throws if it encounters an inner method.

duncanp-sonar:
FYI the QA failure is because one extra protobuf file is now being generated:

```
[ERROR]   UCFGDeserializationTest.ucfgs_created_when_rules_enabled:123 
18:13:33 Expected size:<919> but was:<920> 
```

I'll update the test.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionFactory.cs
```diff
@@ -450,6 +450,14 @@ private IEnumerable<Instruction> ProcessBaseMethodDeclaration(BaseMethodDeclarat
         private IEnumerable<Instruction> ProcessMemberAccessExpression(MemberAccessExpressionSyntax memberAccessExpression)
         {
             var memberAccessSymbol = GetSymbol(memberAccessExpression);
```
Evangelink: Maybe that would be a wrong suggestion but maybe we could update the `GetSymbol` method to include the test of the symbol kind and return null if that's not a kind of symbol we are interested into.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UcfgGeneratorTest.cs
```diff
@@ -169,6 +170,38 @@ public void UcfgGenerator_Disabled_NoOutputFolder()
...
+
+        [TestMethod]
+        public void UcfgGenerator_Enabled_Generate_NoErrorForUnsupportedLanguageConstructs()
+        {
+            var workDir = Path.Combine(TestContext.TestRunResultsDirectory, TestContext.TestName, "0");
+            Directory.CreateDirectory(workDir);
+
+            using (new AssertIgnoreScope())
+            {
+
```
Evangelink: [cosmetic] un-needed empty branch.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UcfgGeneratorTest.cs
```diff
@@ -169,6 +170,38 @@ public void UcfgGenerator_Disabled_NoOutputFolder()
...
         }
 
+
+        [TestMethod]
+        public void UcfgGenerator_Enabled_Generate_NoErrorForUnsupportedLanguageConstructs()
+        {
+            var workDir = Path.Combine(TestContext.TestRunResultsDirectory, TestContext.TestName, "0");
+            Directory.CreateDirectory(workDir);
+
+            using (new AssertIgnoreScope())
```
Evangelink: Might be worth to document which kind of assertion is broken/triggered here (i.e. why do we have this `AssertIgnoreScope`?)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionFactory.cs
```diff
@@ -450,6 +450,14 @@ private IEnumerable<Instruction> ProcessBaseMethodDeclaration(BaseMethodDeclarat
...
+            if (memberAccessSymbol is INamespaceSymbol)
+            {
+                // Every access of a namespace, type or method inside a namespace is represented
+                // by a MemberAccessExpressionSyntax
+                // e.g. Ns1.Ns2.Class1.Class2.Method1(); // has four MemberAccessExpressionSyntax nodes
```
duncanp-sonar: The code previously assumed that every _MemberAccessExpressionSyntax_ node related to an invocation, which isn't the case.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionFactory.cs
duncanp-sonar: Something like _GetSymbol<T>()_? It seems like this is a fairly common pattern, so that would be neater.

Evangelink: Yes. The body of `GetSymbol` would actually retrieve the symbol and then test at the kind of the symbol. If symbol kind is contained in the set of unsupported kinds we return null, otherwise we return the symbol.

PR #1576, opened 2018-07-18
SONARSEC-157: Translate += to concat for strings

Evangelink:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/UcfgBuilder_Instructions.cs
```diff
@@ -1533,6 +1533,201 @@ public string Foo(Class1 c)
...
+                // %6 := Namespace.Class1.PropertyString.get [ this ]
+                // %5 := __concat [ localString %6 ]
+                // localString := __id [ %5 ]
+
+            Foo(localString += ""abc"");
+                // %7 := __concat [ localString const ]
+                // localString := __id [ %7 ]
+                // %8 := Namespace.Class1.Foo(string) [ this localString ]
+
+            localString += localString += localString += ""123"";
```
duncanp-sonar: This test case also applies to fields and properties, doesn't it?

duncanp-sonar: Same for the test above - _Foo(localString += ...)_

Evangelink: Do you think it's worth adding it for all?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/UcfgBuilder_Instructions.cs
```diff
@@ -1533,6 +1533,201 @@ public string Foo(Class1 c)
                 // %0 := object.ToString() [ c ]
         }
     }
+}";
+            UcfgVerifier.VerifyInstructions(code, "Foo");
+        }
+
+        [TestMethod]
+        public void AddAssignment_LocalVariable()
```
duncanp-sonar: Nice. Yet more test cases:
* expressions containing brackets
* local/parameter/field/property += MethodThatReturnsAString()
* += on array elements

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/UcfgBuilder_Instructions.cs
duncanp-sonar: We expect the instructions generated for fields and properties to be different, so yes (although probably two += chained together would be enough).

Also, our translation code is context-sensitive so it matters what the syntax nodes before and after are. Most of the bugs we've had so far are because we didn't expect a node of type _X_ at this point.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionFactory.cs
```diff
@@ -478,6 +473,30 @@ private IEnumerable<Instruction> ProcessMemberAccessExpression(MemberAccessExpre
...
+            if (operatorMethodSymbol == null)
+            {
+                expressionService.Associate(syntaxNode, expressionService.CreateConstant());
+                return NoInstructions;
+            }
+
+            var isStringConcat = operatorMethodSymbol.MethodKind == MethodKind.BuiltinOperator &&
+                operatorMethodSymbol.Parameters.Length == 2 &&
+                operatorMethodSymbol.Name == "op_Addition" &&
+                operatorMethodSymbol.ContainingType.Is(KnownType.System_String);
```
duncanp-sonar: ?Test to check we don't fail for other binary and unary operators: ==, !=, -=, !

PR #1587, opened 2018-07-26
SONARSEC-186: Highlight only method identifier for entry points

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionFactory.cs
```diff
@@ -174,43 +174,68 @@ private void ApplyAsTarget(Expression expression, Instruction instruction)
...
-            BuildFunctionCall(syntaxNode, UcfgBuiltInMethodId.Concatenation, expressionService.CreateVariable(),
-                left, right);
+            BuildFunctionCall(new SyntaxNodeOrToken(syntaxNode), UcfgBuiltInMethodId.Concatenation,
+                expressionService.CreateVariable(), left, right);
 
-        private IEnumerable<Instruction> BuildEntryPointCall(SyntaxNode syntaxNode, ParameterListSyntax parameterList) =>
-            BuildFunctionCall(syntaxNode, UcfgBuiltInMethodId.EntryPoint, expressionService.CreateVariable(),
-                parameterList.Parameters.Select(expressionService.GetOrDefault).ToArray());
+        private IEnumerable<Instruction> BuildEntryPointCall(BaseMethodDeclarationSyntax baseMethodDeclarationSyntax)
+        {
```
duncanp-sonar: This is setting the location on the _AssignCall_ instruction that describes the entry point.
_BasicBlock_ also has a _Location_ property, but this doesn't appear to be set.
Do we need to set _BasicBlock.Location_ too?

PR #1591, opened 2018-07-26
Fix failure when parsing xUnit test result with no total

### LINKED ISSUE #1590: SonarC# fails when parsing xUnit test results with no tests for a test assembly
### Description

When analyzing a project for which one of the test assembly doesn't contain the `total` attribute, the analysis is failing but such xml tag is valid.

```
org.sonar.plugins.dotnet.tests.ParseErrorException: Missing attribute "total" in element <assembly> in PATH_TO_XML_TEST_RESULT at line XXX
```

### Repro steps

```
<assembly name="PATH_TO_DLL" environment="64-bit .NET 4.0.30319.42000 [collection-per-class, parallel (1 threads)]" test-framework="xUnit.net 2.3.1.3858" run-date="2018-07-20" run-time="14:50:31" config-file="PATH_TO_DLL_CONFIG">
```

### Related information

* SonarC# Version 7.2
* SonarQube 6.7.2
* Relates to #406

Evangelink:
Fix #1590


Evangelink:
@duncanp-sonar 7.3 is already release so we can't but as we are about to release 7.4 (which might be what you meant) I think it should be in (no reason not to :)).

PR #1600, opened 2018-07-30
Fix S3400: should not raise issue for virtual methods

andrei-epure-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodShouldNotOnlyReturnConstant.cs
```diff
@@ -69,6 +70,11 @@ protected override void Initialize(SonarAnalysisContext context)
                 SyntaxKind.MethodDeclaration);
         }
 
+        private bool IsVirtual(MethodDeclarationSyntax methodDeclaration)
+        {
+            return methodDeclaration.Modifiers.Any(m => m.IsAnyKind(new SyntaxKind[] { SyntaxKind.VirtualKeyword }));
```
valhristov: `new SyntaxKind[]...` is better to be a `private readonly ISet<SyntaxKind>` or just use `m.IsKind(SyntaxKind.VirtualKeyword)`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodShouldNotOnlyReturnConstant.cs
```diff
@@ -46,7 +47,7 @@ protected override void Initialize(SonarAnalysisContext context)
                 {
                     var methodDeclaration = (MethodDeclarationSyntax)c.Node;
 
-                    if (methodDeclaration.ParameterList?.Parameters.Count > 0)
+                    if (methodDeclaration.ParameterList?.Parameters.Count > 0 || IsVirtual(methodDeclaration))
```
valhristov: Usually we format all additional conditions on new lines:
```
    if (methodDeclaration.ParameterList?.Parameters.Count > 0 ||
        IsVirtual(methodDeclaration))
```

valhristov: On the other hand, for partial methods you could be raising on the parts that do not have `virtual` keyword (I did not check but have the feeling it is possible). We already have the symbol of the method, so it could be safer to use it instead of the syntax.

andrei-epure-sonarsource: As discussed offline, [partial methods](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/partial-method) must be void, so the rule does not apply

Issue #1529, opened 2018-07-04
"Fields should not have public accessibility" should not run against structs


GeertvanHorrik:
### Description

Getting false positives on structs.

### Repro steps

Please provide the steps required to reproduce the problem

```
[StructLayout(LayoutKind.Sequential)]
public struct RECT
{
    public int Left;
    public int Top;
    public int Right;
    public int Bottom;
}
```

### Expected behavior

For structs marked with the _StructLayout_, the usage of public fields should be allowed.

### Actual behavior

Getting vulnerability that fields should not have public accessibility.

### Related information

* SonarC# Version 7.2.1
* Visual Studio Version 2017 15.8 (preview 3)

RSPEC: https://jira.sonarsource.com/browse/RSPEC-1104

duncanp-sonar:
@GeertvanHorrik thanks for the feedback. Excluding all structs from the rule seems too broad and will miss a large number of incorrect uses of public fields. I'm guessing from the repro sample the main issue is with structs used for interop with unmanaged code, which makes perfect sense.
Do you have any other specific cases in which public fields should be allowed on structs?

duncanp-sonar:
Sounds good to me. I'll amend the issue accordingly.

andrei-epure-sonarsource:
Fixed with PR  #1604

PR #1604, opened 2018-07-31
#1529 Fix S1104 - should not report StructLayout classes/structs

andrei-epure-sonarsource:


andrei-epure-sonarsource:
@Evangelink - I added a test with  `struct` (sorry for not having a separate commit with it)

PR #1612, opened 2018-08-03
 Improve the verifier to allow to depends upon multiple version of the… 

### LINKED ISSUE #1611: Improve the verifier to allow to depends upon multiple version of the same NuGet
When writing a rule that depends upon some third-party, we need to support multiple versions (changes in API) of those third-parties. The current solution (i.e. NuGet packages are added to the test project), prevents such feature.

In addition, it would be nice to be able to have some automatic test with the latest version of those dependencies so that we can be notified of any breaking change on some API.

Evangelink:
… same NuGet

Fix #1611


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/KnownType.cs
```diff
@@ -48,8 +48,6 @@ internal sealed class KnownType
         internal static readonly KnownType NUnit_Framework_TestCaseSourceAttribute = new KnownType("NUnit.Framework.TestCaseSourceAttribute");
         internal static readonly KnownType NUnit_Framework_TestFixtureAttribute = new KnownType("NUnit.Framework.TestFixtureAttribute");
         internal static readonly KnownType NUnit_Framework_TheoryAttribute = new KnownType("NUnit.Framework.TheoryAttribute");
-        internal static readonly KnownType Oracle_ManagedDataAccess_Client_OracleCommand = new KnownType("Oracle.ManagedDataAccess.Client.OracleCommand");
```
Evangelink: Dependencies were not removed when we removed the old implementation of the rule.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/AssemblyReferenceService.cs
```diff
@@ -0,0 +1,102 @@
...
+        private static readonly string systemAssembliesFolder =
+            new FileInfo(typeof(object).Assembly.Location).Directory.FullName;
+
+        private static readonly PackageManager packageManager =
+            new PackageManager(CreatePackageRepository(), PackagesFolderRelativePath);
+
+        private static readonly Dictionary<AssemblyReference, MetadataReference> cache =
+            new Dictionary<AssemblyReference, MetadataReference>();
+
+        private static readonly List<string> allowedNugetLibDirectories =
```
Evangelink: Dlls in folder `netstandard1.3`+ results in failures (most likely because of the current fwk version).

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/AssertionArgsShouldBePassedInCorrectOrderTest.cs
```diff
@@ -19,24 +19,49 @@
...
+        public void AssertionArgsShouldBePassedInCorrectOrder_XUnit(string testFwkVersion)
         {
-            Verifier.VerifyAnalyzer(@"TestCases\AssertionArgsShouldBePassedInCorrectOrder.cs",
+            Verifier.VerifyAnalyzer(@"TestCases\AssertionArgsShouldBePassedInCorrectOrder.Xunit.cs",
                 new AssertionArgsShouldBePassedInCorrectOrder(),
                 null,
-                Verifier.MicrosoftVisualStudioTestToolsUnitTestingAssembly,
-                Verifier.NUnitFrameworkAssembly,
-                Verifier.XunitAssertAssembly);
+                AssemblyReference.FromNuGet("xunit.assert.dll", "xunit.assert", testFwkVersion),
```
valhristov: I guess we use relatively few nugets, so we could have some convenience methods:
```
NugetReference.XUnit(string version = NugetInfo.LatestVersion) { ... }
NugetReference.MsTest(string version = NugetInfo.LatestVersion) { ... }
...
FrameworkReference.Mscorlib() => AssemblyReference.FromFramework(...);
FrameworkReference.SystemComponentModelComposition() { ... }
// or properties
FrameworkReference.Mscorlib { get; } = AssemblyReference.FromFramework(...);
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/MethodShouldBeNameAccordingToSynchronicityTest.cs
```diff
@@ -28,16 +28,55 @@ namespace SonarAnalyzer.UnitTest.Rules
     public class MethodShouldBeNameAccordingToSynchronicityTest
     {
         [TestMethod]
+        [DataRow("4.0.0")]
```
valhristov: Did you try if we could use shorter version numbers? This is minor, but I guess it could be convenient to specify just `4.0` or `4.0.*` and nuget to automatically download the latest patch, e.g. `4.0.15`.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
```diff
@@ -238,7 +199,1702 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Compile>
     <Compile Include="**\*.cs" Exclude="**\obj\**\*;**\bin\**\*;**\*.Designer.cs;TestCases\**\*.cs;TestCasesForRuleFailure\**\*.*" />
-    <None Include="TestCases\**\*.*">
+    <None Include="TestCases\AbstractTypesShouldNotHaveConstructors.cs">
```
valhristov: I think the following many lines should be removed.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/AssertionArgsShouldBePassedInCorrectOrderTest.cs
Evangelink: I was thinking about having some `enum` for the framework assemblies as the list is finite. Regarding NuGet we could probably improve the usage even more, I agree.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/MethodShouldBeNameAccordingToSynchronicityTest.cs
Evangelink: I only took the older version available on NuGet for each package.

Evangelink: A bug fix shouldn't change anything on our side. What we care is to say this works on old and works on latest.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
Evangelink: Crap... indeed

PR #1614, opened 2018-08-03
Fix S3242: Rule should not suggest base type for virtual methods

### LINKED ISSUE #1537: Fix S3242: Rule should not suggest base type for virtual methods
### Description

The signature of virtual methods can be considered like a contract between the base class and its subclasses.
If the implementation of the base class does not use the most concrete type of a parameter, a subclass might.

I'm facing this issue with a specialized collection type.
I have type B inheriting from type A.
My collection type inherits from `Collection<B>`, and defines a virtual method that is meant to apply to an item of the collection, so of type B. However the base implementation only accesses members from type A, so rule S3242 kicks in and suggests to take A as parameter. But if I do that, I can imagine that a developer implementing a derived class will go through the first 3 or 4 stages of grief.

### Repro steps

1. Have type B inherit from type A

2. Create a virtual method with a parameter of type B, but whose implementation only uses members of type A

```
using System;
using System.Collections.ObjectModel;

namespace NS1
{
    public class A
    {
        public string ParentProperty { get; set; }
    }

    public class B : A
    {
        public string ChildProperty { get; set; }
    }

    public class MyCollection : Collection<B>
    {
        protected virtual string Manipulate(B item) // <- S3242 raised here
        {
            return item.ParentProperty;
        }
    }
}
```

### Expected behavior

Rule should not raise a violation.

### Actual behavior

Rule raises a violation.

### Related information

* SonarC# Version 7.2
* Visual Studio Version 15.7

Evangelink:
Fix #1537

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldUseBaseTypes.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldUseBaseTypes.cs
@@ -44,21 +44,22 @@ namespace SonarAnalyzer.Rules.CSharp
 
         protected override void Initialize(SonarAnalysisContext context)
         {
-            context.RegisterSyntaxNodeActionInNonGenerated(c =>
-            {
-                FindViolations(c)
-                    .ToList()
-                    .ForEach(d => c.ReportDiagnosticWhenActive(d));
-            },
-            SyntaxKind.MethodDeclaration);
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
+                {
+                    FindViolations(c)
+                        .ToList()
+                        .ForEach(d => c.ReportDiagnosticWhenActive(d));
+                },
+                SyntaxKind.MethodDeclaration);
         }
 
         public IEnumerable<Diagnostic> FindViolations(SyntaxNodeAnalysisContext context)
         {
-
             if (!(context.SemanticModel.GetDeclaredSymbol(context.Node) is IMethodSymbol methodSymbol) ||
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldUseBaseTypes.cs
```diff
@@ -1,299 +1,300 @@
...
+                },
+                SyntaxKind.MethodDeclaration);
+        }
+
+        public IEnumerable<Diagnostic> FindViolations(SyntaxNodeAnalysisContext context)
+        {
+            if (!(context.SemanticModel.GetDeclaredSymbol(context.Node) is IMethodSymbol methodSymbol) ||
+                methodSymbol.Parameters.Length == 0 ||
+                methodSymbol.IsOverride ||
+                methodSymbol.IsVirtual ||
```
Evangelink: this line is the only real change

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodsShouldUseBaseTypes.cs
```diff
@@ -438,13 +438,21 @@ public void Test_16(B foo1, B foo2, B foo3)
...
         int Property { get; }
+    }
+
+    public class A : IA
+    {
+        public int Property { get { return 0; } }
+
+        public void Method() { }
 
-        public virtual void Test_17(B foo) // Noncompliant
```
Evangelink: Code was not compiling so I fixed it at the same time.

PR #1615, opened 2018-08-06
Fix S1450: Implement robust detection whether a local field could be …

### LINKED ISSUE #1607: Fix S1450: Implement robust detection whether a local field could be converted to a local variable
The current implementation of the rule relies on changing and compiling the analyzed class to see if a field could be converted to local variable. This mechanism is not precise and generates a lot of false positives and negatives, and in general the rule does not work well.

The new proposal is to detect if all reads of a field are preceded with unconditional writes. If that's the case, we report, otherwise we don't.

To do that we could collect all read statements and all write statements for particular method or property into separate hash sets, then starting from each read statement we go up the tree and see if there is a preceding statement that's in the write set.

valhristov:
…converted to a local variable

Fix #1607


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -36,340 +36,216 @@ namespace SonarAnalyzer.Rules.CSharp
...
-                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, privateField.Syntax.GetLocation(),
-                            privateField.Symbol.Name));
-                    }
+                    privateFields.Keys
+                        .Where(collector.IsRemovableField)
+                        .ToList()
+                        .ForEach(ReportField);
+
+                    void ReportField(ISymbol fieldSymbol) =>
+                        c.ReportDiagnosticWhenActive(
```
Evangelink: [cosmetic][minor] we usually split on length and not like this.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -36,340 +36,216 @@ namespace SonarAnalyzer.Rules.CSharp
...
-                        .Cast<FieldDeclarationSyntax>()
-                        .Where(f => !f.AttributeLists.Any())
-                        .SelectMany(f => f.Declaration.Variables.Select(
-                            v => new PrivateField(v, semanticModel.GetDeclaredSymbol(v))))
-                        .Where(f => f.Symbol != null && f.Symbol.DeclaredAccessibility == Accessibility.Private)
-                        .ToImmutableDictionary(
-                            f => f.Symbol,
-                            f => f);
+            return typeDeclaration.Members
+                .Where(m => m.IsKind(SyntaxKind.FieldDeclaration))
```
Evangelink: Not sure if that's a lot slower but what about `OfType<FieldDeclarationSyntax>` instead of `Where` + `Cast`?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -36,340 +36,216 @@ namespace SonarAnalyzer.Rules.CSharp
...
-                if (!builder.ContainsKey(enclosingSymbol))
-                {
-                    builder.Add(enclosingSymbol, new Dictionary<SyntaxNode, ISymbol>());
-                }
+                this.semanticModel = semanticModel;
+                this.privateFields = privateFields;
 
-                builder[enclosingSymbol].Add(referenceSyntax, referencedSymbol);
+                privateFieldNames = privateFields.Keys.Select(s => s.Name).ToHashSet();
+                reads = new Dictionary<ISymbol, Dictionary<ISymbol, HashSet<StatementSyntax>>>();
```
Evangelink: Let's have those 2 initialisations inline.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -36,340 +36,216 @@ namespace SonarAnalyzer.Rules.CSharp
...
-                        .Where(m => m.IsKind(SyntaxKind.FieldDeclaration))
-                        .Cast<FieldDeclarationSyntax>()
-                        .Where(f => !f.AttributeLists.Any())
-                        .SelectMany(f => f.Declaration.Variables.Select(
-                            v => new PrivateField(v, semanticModel.GetDeclaredSymbol(v))))
-                        .Where(f => f.Symbol != null && f.Symbol.DeclaredAccessibility == Accessibility.Private)
-                        .ToImmutableDictionary(
-                            f => f.Symbol,
-                            f => f);
+            return typeDeclaration.Members
```
Evangelink: I might be missing something but I think you won't handle properly partial classes/strucs (e.g. where a field is declared on some part but used on others).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -36,340 +36,216 @@ namespace SonarAnalyzer.Rules.CSharp
...
-            foreach (var references in referencesByEnclosingSymbol)
+            public bool IsRemovableField(ISymbol fieldSymbol)
             {
-                if (!classMethods.Contains(references.Key))
-                {
-                    foreach (var reference in references.Value)
-                    {
-                        privateFields[reference.Value].Excluded = true;
-                    }
+                var writesByMethod = writes.GetValueOrDefault(fieldSymbol);
```
Evangelink: I don't understand how you can tell `ByMethod` here!?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -36,340 +36,216 @@ namespace SonarAnalyzer.Rules.CSharp
...
+                var readsByMethod = reads.GetValueOrDefault(fieldSymbol);
```
Evangelink: It's fast access but it's un-needed when `writesByMethod` is `null` so let's move it after the check.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -36,340 +36,216 @@ namespace SonarAnalyzer.Rules.CSharp
...
-                    referencedAtLeastOnceFromClassMethod.Add(reference.Value);
+                // A field is removable when no method read it, or all methods that read it, first overwrite it
+                return readsByMethod == null
+                    || readsByMethod.Keys.All(MethodOverwritesBeforeReading);
 
-                    if (!IsReferenceToSingleFieldValue(reference))
-                    {
-                        privateFields[reference.Value].Excluded = true;
-                    }
+                bool MethodOverwritesBeforeReading(ISymbol methodSymbol)
```
Evangelink: If you are sure this is a method symbol why don't you use `IMethodSymbol`?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -36,340 +36,216 @@ namespace SonarAnalyzer.Rules.CSharp
...
 
-            foreach (var privateField in privateFields.Values)
-            {
-                if (!referencedAtLeastOnceFromClassMethod.Contains(privateField.Symbol))
+                // Returns all statements before the specified statement within the containing method
+                IEnumerable<StatementSyntax> GetPreviousStatements(StatementSyntax statement)
                 {
-                    privateField.Excluded = true;
+                    var previousStatements = statement.Parent.ChildNodes()
+                        .OfType<StatementSyntax>()
```
Evangelink: I might say something a bit stupid but I would put the `TakeWhile` before the `OfType` because we will do less useless casts.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -36,340 +36,216 @@ namespace SonarAnalyzer.Rules.CSharp
...
 
-            foreach (var partiallyRewrittenSymbol in partiallyRewrittenSymbols)
+            private SyntaxNodeWithSymbol<SyntaxNode, ISymbol> GetTopmostSyntaxWithTheSameSymbol(SyntaxNode potentialReference)
             {
-                privateFields[partiallyRewrittenSymbol].Excluded = true;
-            }
+                var referencedSymbol = semanticModel.GetSymbolInfo(potentialReference).Symbol;
 
-            var allSymbolsToRewrite = references.Values.ToHashSet();
+                if (referencedSymbol != null)
```
Evangelink: Let's reduce nesting.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -36,340 +36,216 @@ namespace SonarAnalyzer.Rules.CSharp
...
+                {
+                    while (potentialReference.Parent != null &&
```
Evangelink: [minor][cosmetic] Weird alignment of the second condition of this `while`.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -36,340 +36,216 @@ namespace SonarAnalyzer.Rules.CSharp
...
+                           referencedSymbol.Equals(semanticModel.GetSymbolInfo(potentialReference.Parent).Symbol))
```
Evangelink: Don't we have a better way than to loop through parents + parent symbol? This seems overly expansive.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
valhristov: It could be slower, but is definitely easier to read.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
valhristov: It could be property accessor too... Any suggestion for a common name?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
Evangelink: We bail out earlier when it's partial... ignore this

PR #1617, opened 2018-08-06
 Fix executable lines of code metrics to handle properly ExcludeFromCo…

### LINKED ISSUE #1446: Exclusions from executable lines of code are not handled correctly
Code marked with _[ExcludeFromCodeCoverage](https://msdn.microsoft.com/en-us/library/system.diagnostics.codeanalysis.excludefromcodecoverageattribute(v=vs.110).aspx)_ should not be included in the exectuable_line metric.
There are a few cases that are not handled correctly:
* constructors
* events
* partial classes

Test cases:
```
        [TestMethod]
        public void Test_xx_Constructor()
        {
            AssertLinesOfCode(
              @"using System.Diagnostics.CodeAnalysis;
                class Program
                {
                    int count;

                    [ExcludeFromCodeCoverage]
                    public Program() : this(1)
                    {
                        count = 123;            // excluded
                    }

                    public Program(int initialCount)
                    {
                        count = initialCount;   // +1
                    }
                }",
            14);
        }

        [TestMethod]
        public void Test_xx_Event()
        {
            AssertLinesOfCode(
              @"using System.Diagnostics.CodeAnalysis;
                class EventClass
                {
                    private System.EventHandler _explicitEvent;

                    [System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
                    public event System.EventHandler ExplicitEvent1
                    {
                        add { _explicitEvent += value; }        // Excluded
                        remove { _explicitEvent -= value; }     // Excluded
                    }

                    public event System.EventHandler ExplicitEvent2
                    {
                        add { _explicitEvent += value; }        // +1 ?? TODO check
                        [System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
                        remove { _explicitEvent -= value; }     // excluded
                    }

                    public event System.EventHandler ExplicitEvent3
                    {
                        add { _explicitEvent += value; }        // +1
                        remove { _explicitEvent -= value; }     // +1
                    }
                }",
            15, 22, 23);
        }

        [TestMethod]
        public void Test_xx_PartialClasses()
        {
            // 
            AssertLinesOfCode(
              @"[ExcludeFromCodeCoverage]
                partial class Program
                {
                    int FooProperty { get { return 1; } }   // excluded
                }

                partial class Program
                {
                    void Method1()
                    {
                        System.Console.WriteLine();         // other class partial is excluded -> excluded
                    }
                }

                partial class AnotherClass
                {
                    void Method2()
                    {
                        System.Console.WriteLine();         // +1
                    }
                } ",
              19);
        }
```

Evangelink:
…deCoverageAttribute

Fix #1446

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Metrics/ExecutableLinesWalker.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Metrics/ExecutableLinesWalker.cs
@@ -31,12 +31,12 @@ namespace SonarAnalyzer.Metrics.CSharp
     public class ExecutableLinesWalker : CSharpSyntaxWalker
     {
         private readonly HashSet<int> executableLineNumbers = new HashSet<int>();
+        private readonly SemanticModel semanticModel;
 
-        private static readonly string[] ExcludeFromCoverageAttributes =
-            {
-                "ExcludeFromCodeCoverage",
-                "ExcludeFromCodeCoverageAttribute"
-            };
+        public ExecutableLinesWalker(SemanticModel semanticModel)
+        {
+            this.semanticModel = semanticModel;
+        }
 
         public ICollection<int> ExecutableLines => executableLineNumbers;
 
@@ -50,8 +50,7 @@ namespace SonarAnalyzer.Metrics.CSharp
 
         private bool FindExecutableLines(SyntaxNode node)
         {
-            var kind = node.Kind();
-            switch (kind)
+            switch (node.Kind())
             {
                 case SyntaxKind.CheckedStatement:
                 case SyntaxKind.UncheckedStatement:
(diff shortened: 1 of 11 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Metrics/ExecutableLinesWalkerTest.cs
```diff
@@ -438,5 +434,156 @@ int FooProperty
...
+                    [System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
+                    public event System.EventHandler ExplicitEvent1
+                    {
+                        add { _explicitEvent += value; }        // Excluded
+                        remove { _explicitEvent -= value; }     // Excluded
+                    }
+
+                    public event System.EventHandler ExplicitEvent2
+                    {
+                        add { _explicitEvent += value; }        // +1 ?? TODO check
```
duncanp-sonar: Presumably we know what the expected behaviour is now so this TODO can be removed?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Metrics/ExecutableLinesWalker.cs
```diff
@@ -96,43 +95,71 @@ private bool FindExecutableLines(SyntaxNode node)
...
+                case SyntaxKind.GetAccessorDeclaration:
+                    return !HasExcludedCodeAttribute((AccessorDeclarationSyntax)node, ads => ads.AttributeLists);
 
                 default:
                     return true;
             }
         }
 
-        private bool HasExcludedCodeAttribute(SyntaxList<AttributeListSyntax> attributeSyntaxLists)
+        private bool HasExcludedCodeAttribute<T>(T node, Func<T, SyntaxList<AttributeListSyntax>> getAttributeLists,
```
duncanp-sonar: Why not type the second parameter as _SyntaxList{AttributeListSyntax}_ directly, rather than as function?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Metrics/Metrics.cs
```diff
@@ -32,20 +32,51 @@ namespace SonarAnalyzer.Metrics.CSharp
...
+                SyntaxKind.StructDeclaration,
+                SyntaxKind.InterfaceDeclaration
+            };
+
+        private static readonly ISet<SyntaxKind> FunctionKinds = new
+            HashSet<SyntaxKind>
+            {
+                SyntaxKind.ConstructorDeclaration,
+                SyntaxKind.DestructorDeclaration,
+                SyntaxKind.MethodDeclaration,
```
duncanp-sonar: How are property and explicit event declarations handled?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/MetricsTest.cs
```diff
@@ -581,8 +582,10 @@ private static MetricsBase MetricsFor(AnalyzerLanguage language, string text)
                 throw new ArgumentException("Supplied language is not C# neither VB.Net", nameof(language));
             }
 
+            (var syntaxTree, var semanticModel) = TestHelper.Compile(text);
+
             return language == AnalyzerLanguage.CSharp
-                ? (MetricsBase)new SonarAnalyzer.Metrics.CSharp.Metrics(CSharpSyntaxTree.ParseText(text))
+                ? (MetricsBase)new SonarAnalyzer.Metrics.CSharp.Metrics(syntaxTree, semanticModel)
                 : new SonarAnalyzer.Common.VisualBasic.Metrics(VisualBasicSyntaxTree.ParseText(text));
```
duncanp-sonar: Do we need a separate issue to fix the metrics for VB too?

Evangelink: It's not implemented for VB.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Metrics/ExecutableLinesWalker.cs
duncanp-sonar: That's debatable :)  It does adds unnecessary complexity to this method declaration, and it'll be slower at runtime, although only by a small amount.

PR #1619, opened 2018-08-06
 SONARSEC-193: Handle null-conditional and null-coalescing operators

Evangelink:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/UcfgInstructionTest.Assignment.cs
```diff
@@ -459,16 +459,29 @@ public void TestSimpleFieldAccessAndAssignment()
...
+                // %0 := __id [ this.field ]
                 // variable := __id [ %0 ]
-
+             variable = this.field;
+                // %1 := __id [ this.field ]
+                // variable := __id [ %1 ]
+             variable = baseField;
+                // %2 := __id [ this.baseField ]
+                // variable := __id [ %2 ]
+             variable = base.baseField;
```
Evangelink: They are not directly fixed by this PR but I noticed we were not testing calls to `base`.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/UcfgInstructionTest.Suspect.cs
```diff
@@ -68,6 +68,58 @@ public string Foo(Class1 c)
...
+    {
+        public BaseClass baseField;
+    }
+     public class Class1 : BaseClass
+    {
+        private Class1 field;
+         public void Foo()
+        {
+            var result = field?.field;
+                // %0 := __id [ this.field ]
```
Evangelink: This first access (read) to field comes from the previous block (linked to the null check).

PR #1624, opened 2018-08-07
Fix S1075: Rule should not report on virtual path for asp.net

### LINKED ISSUE #1586: Fix S1075: Rule should not report on virtual path for asp.net
From @kwilliams1987 in https://community.sonarsource.com/t/false-positive-on-s1075-in-asp-net-mvc-with-paths/843
> When using Server.MapPath("~/ ...") to map a relative path to the server root directory, SonarLint raises an “S1075: Refactor your code not to use hardcoded absolute paths or URIs”.
~/ paths are not absolute by their very nature.
Running SonarLint on Visual Studio 2015 Professional, ASP.NET MVC 5

andrei-epure-sonarsource:
Fix #1586

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/UriShouldNotBeHardcodedBase.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/UriShouldNotBeHardcodedBase.cs
@@ -47,6 +47,8 @@ namespace SonarAnalyzer.Rules
             new Regex($"{UriScheme}|{AbsoluteDiskUri}|{AbsoluteMappedDiskUri}|{AbsoluteUnixUri}",
                 RegexOptions.Compiled);
 
+        protected static readonly Regex VirtualPathRegex = new Regex(AbsoluteUnixUri, RegexOptions.Compiled);
+
         protected static readonly Regex PathDelimiterRegex = new Regex(@"^(\\|/)$", RegexOptions.Compiled);
 
         protected static readonly ISet<string> checkedVariableNames =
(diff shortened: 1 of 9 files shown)


andrei-epure-sonarsource:
I must update the RSPEC text

andrei-epure-sonarsource:
I opened [SONARVB-316](https://jira.sonarsource.com/browse/SONARVB-316) to update the vb.net test cases

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/UriShouldNotBeHardcodedBase.cs
```diff
@@ -59,6 +61,21 @@ public abstract class UriShouldNotBeHardcodedBase : SonarDiagnosticAnalyzer
                 "URN",
                 "STREAM"
             };
+
+        internal static readonly ISet<KnownType> AspNetTypes = new HashSet<KnownType>
```
andrei-epure-sonarsource: would deserve a better name and extract from ISet in constant

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UriShouldNotBeHardcodedTest.cs
```diff
@@ -27,12 +27,24 @@ namespace SonarAnalyzer.UnitTest.Rules
...
-        public void UriShouldNotBeHardcoded()
+        [DataTestMethod]
+        [DataRow("3.0.20105.1", "2.0.4", "2.0.3")]
+        [DataRow(AssemblyReference.NuGetInfo.LatestVersion, AssemblyReference.NuGetInfo.LatestVersion, AssemblyReference.NuGetInfo.LatestVersion)]
+        public void UriShouldNotBeHardcoded(string aspNetMvcVersion, string aspNetCoreMvcVersion, string aspNetCoreRoutingVersion)
         {
             Verifier.VerifyAnalyzer(@"TestCases\UriShouldNotBeHardcoded.cs",
-                new UriShouldNotBeHardcoded());
+                new UriShouldNotBeHardcoded(),
+                additionalReferences: new AssemblyReference[] {
```
Evangelink: [minor] You don't need the explicit type of the array.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UriShouldNotBeHardcodedTest.cs
```diff
@@ -27,12 +27,24 @@ namespace SonarAnalyzer.UnitTest.Rules
...
+                    AssemblyReference.FromFramework("System.Web.dll"),
+                    AssemblyReference.FromNuGet("System.Web.Mvc.dll", "Microsoft.AspNet.Mvc", aspNetMvcVersion),
+                    // for VirtualFileResult
+                    AssemblyReference.FromNuGet("Microsoft.AspNetCore.Mvc.Core.dll", "Microsoft.AspNetCore.Mvc.Core", aspNetCoreMvcVersion),
```
Evangelink: [minor][optional] I would have split the test cases into 3 classes, one for each:

- Classic paths
- AspNet Mvc
- AspNet Core Mvc

andrei-epure-sonarsource: Fair point, in case it fails because of the dependency, it's easier to spot

PR #1626, opened 2018-08-07
 SONARSEC-198: Handle cast + safe cast

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionFactory.cs
```diff
@@ -608,6 +612,11 @@ private IEnumerable<Instruction> ProcessBinaryExpression(BinaryExpressionSyntax
             var leftExpression = expressionService.GetOrDefault(binaryExpression.Left);
             var rightExpression = expressionService.GetOrDefault(binaryExpression.Right);
 
+            if (binaryExpression.OperatorToken.IsKind(SyntaxKind.AsKeyword))
```
Evangelink: This is for the as cast.

PR #1627, opened 2018-08-07
SONARSEC-199: Handle event registration

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionFactory.cs
```diff
@@ -577,6 +577,11 @@ private IEnumerable<Instruction> ProcessAssignmentExpression(AssignmentExpressio
                 return instructions;
             }
 
+            if (leftPartSymbol is IEventSymbol)
+            {
+                return instructions;
```
Evangelink: When the assignment is on an event we don't need to do anything.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/UcfgInstructionTest.cs
```diff
@@ -862,5 +862,37 @@ public override bool Equals(object obj)
...
+        {
+            var e = new SocketAsyncEventArgs { UserToken = null };
+                // %0 := new System.Net.Sockets.SocketAsyncEventArgs
+                // %1 := System.Net.Sockets.SocketAsyncEventArgs.SocketAsyncEventArgs() [ %0 ]
+                // %2 := System.Net.Sockets.SocketAsyncEventArgs.UserToken.set [ %0 const ]
+                // e := __id [ %0 ]
+
+            e.Completed += _onComplete;
+                // %3 := __id [ this._onComplete ]
+                // %4 := System.Net.Sockets.SocketAsyncEventArgs.Completed.add [ System.Net.Sockets.SocketAsyncEventArgs const %3 ]
```
duncanp-sonar: It seems a bit odd that there isn't a reference to _e_ in %4. Still, I guess there's no way adding the event handler could taint _e_ (i.e. if _this_ was tainted).

Evangelink: I actually didn't noticed that `e` was missing. It's more corner case so I will ignore it for now but create a ticket about it.

PR #1631, opened 2018-08-08
Improve error reporting for UCFG exceptions

duncanp-sonar:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionFactory.cs
```diff
@@ -53,6 +53,28 @@ public UcfgInstructionFactory(SemanticModel semanticModel, UcfgExpressionService
         }
 
         public IEnumerable<Instruction> CreateFrom(SyntaxNode syntaxNode)
+        {
+            try
+            {
+                return UnsafeCreateFrom(syntaxNode);
```
duncanp-sonar: FYI we can have nested calls to _CreateFrom_, so when looking at the log we need to look for the last _Inner exception_ to find the lowest-level node that caused the failure.

PR #1632, opened 2018-08-08
SONARSEC-205: Ignore instructions for which the target is invalid

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionFactory.cs
```diff
@@ -149,44 +149,44 @@ public IEnumerable<Instruction> CreateFrom(SyntaxNode syntaxNode)
                 .Concat(BuildAnnotationCall(attributeSyntax, targetOfAttribute, expressionService.GetOrDefault(attributeSyntax)));
         }
 
-        private void ApplyAsTarget(Expression expression, Instruction instruction)
+        private IEnumerable<Instruction> ApplyAsTarget(Expression expression, Instruction instruction)
         {
+            var instructionsToReturn = new[] { instruction };
```
duncanp-sonar: Am I misreading, or is _instructionsToReturn_ always empty?
If so, what's the difference between returning _instructionsToReturn_ and _NoInstructions_?

PR #1637, opened 2018-08-08
Added UCFG unit tests

duncanp-sonar:


duncanp-sonar:
No. The coverage on the _InstructionFactory_ is still too low.
Most of the new lines are in the try-catch-decorate from inside _CreateFrom_ (i.e. that reports exactly which node fails), and I'm not sure I can easily test that.

There are a few other new lines scattered throughout the class that aren't executed, which is a bit of a surprise given that we had regressions tests to start with.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/UcfgInstructionFactoryTests.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace SonarAnalyzer.UnitTest.ControlFlowGraph
+{
+    [TestClass]
+    public class UcfgInstructionFactoryTests
+    {
+        [TestMethod]
+        public void CreateFrom_DecoratesAndReThrowsExceptions()
+        {
+            // Arrange
+            var mockSemanticModel = new Mock<SemanticModel>(MockBehavior.Strict); // strict -> throw on any usage
```
duncanp-sonar: This should push the code coverage high enough (although there are still uncovered lines in some fairly complex bits of the file that may or may not be required or working as expected).

The instruction factory and expression service aren't easy to test; they're internal so they can't be mocked and they don't have virtual methods, so it's difficult to inject behaviours (e.g. throwing an exception when testing).

_SyntaxNode_ can't be mocked either since it only has internal constructors.
Fortunately, _SemanticModel_ is public and abstract so it can be mocked.

I looked at creating an _IUcfgExpressionService_ interface, but that would have involved a lot more churn for very little value (also, the _UcfgExpressionService_ uses default arguments for a couple of methods, so I'd had to have added extra methods or changed all the callers).

PR #1639, opened 2018-08-08
Add a warning to notify user that no coverage report file was found f…

### LINKED ISSUE #1638: Add a warning to notify user that no coverage report file was found for the given pattern
Currently when the user pass a coverage report file or pattern to the scanner for msbuild and we cannot find any matching file we don't have any warning showing up.

Let's improve this and add a warning message.

Evangelink:
…or the given pattern

Fix #1638

### DIFF
--- a/sonar-dotnet-tests-library/src/main/java/org/sonar/plugins/dotnet/tests/CoverageAggregator.java
+++ b/sonar-dotnet-tests-library/src/main/java/org/sonar/plugins/dotnet/tests/CoverageAggregator.java
@@ -20,12 +20,17 @@
 package org.sonar.plugins.dotnet.tests;
 
 import java.io.File;
+import java.util.Set;
 import java.util.function.Predicate;
 import org.sonar.api.batch.ScannerSide;
 import org.sonar.api.config.Configuration;
+import org.sonar.api.utils.log.Logger;
+import org.sonar.api.utils.log.Loggers;
 
 @ScannerSide
 public class CoverageAggregator {
+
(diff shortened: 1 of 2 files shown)


sonar-dotnet-tests-library/src/main/java/org/sonar/plugins/dotnet/tests/CoverageAggregator.java
```diff
@@ -110,8 +115,13 @@ Coverage aggregate(WildcardPatternFileProvider wildcardPatternFileProvider, Cove
...
     for (String reportPathPattern : reportPaths) {
       if (!reportPathPattern.isEmpty()) {
-        for (File reportFile : wildcardPatternFileProvider.listFiles(reportPathPattern)) {
-          aggregatedCoverage.mergeWith(coverageCache.readCoverageFromCacheOrParse(parser, reportFile));
+        Set<File> filesMatchingPattern = wildcardPatternFileProvider.listFiles(reportPathPattern);
+        if (filesMatchingPattern.isEmpty()) {
+          LOG.warn("Could not find any coverage report file matching the pattern '{}'.", reportPathPattern);
+        } else {
+          for (File reportFile : filesMatchingPattern) {
+            aggregatedCoverage.mergeWith(coverageCache.readCoverageFromCacheOrParse(parser, reportFile));
```
Evangelink: Testing here if the file exists is pointless because the `wildcardPatternFileProvider` already has filtered out files.

PR #1640, opened 2018-08-08
Update S3881: Rule should allow abstract IDisposable implementations

### LINKED ISSUE #1460: Update S3881: Rule should allow abstract IDisposable implementations
### Description

The following code should not raise issues:

```
public abstract class DisposableBase : IDisposable
{
    public void Dispose()
    {
        Dispose(true);
        GC.SuppressFinalize(this);
    }

    protected abstract void Dispose(bool disposing);
}
```

because it is equivalent to:
```
public abstract class DisposableBase : IDisposable
{
    public void Dispose()
    {
        Dispose(true);
        GC.SuppressFinalize(this);
    }

    protected virtual void Dispose(bool disposing)
	{
	}
}
```

andrei-epure-sonarsource:
Fix #1460

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
@@ -41,7 +41,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private static readonly ISet<SyntaxKind> notAllowedDisposeModifiers = new HashSet<SyntaxKind>
         {
             SyntaxKind.VirtualKeyword,
-            SyntaxKind.ProtectedKeyword
+            SyntaxKind.AbstractKeyword
         };
 
         private static readonly DiagnosticDescriptor rule =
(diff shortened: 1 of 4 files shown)


andrei-epure-sonarsource:
ITs are failing locally

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
```diff
@@ -174,12 +174,13 @@ private void VerifyDisposeOverride(MethodDeclarationSyntax disposeMethod)
 
             private void VerifyDispose(MethodDeclarationSyntax disposeMethod, bool isSealedClass)
             {
-                if (disposeMethod?.Body == null)
+                if (disposeMethod == null)
```
andrei-epure-sonarsource: because of this change, ITs are failing with

> CSC : error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.ImplementIDisposableCorrectly' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'. [C:\Workspace\sonar-csharp\sonaranalyzer-dotnet\its\sources\akka.net\src\core\Akka\Akka.csproj]

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
```diff
@@ -174,12 +174,13 @@ private void VerifyDisposeOverride(MethodDeclarationSyntax disposeMethod)
 
             private void VerifyDispose(MethodDeclarationSyntax disposeMethod, bool isSealedClass)
             {
-                if (disposeMethod?.Body == null)
```
andrei-epure-sonarsource: because of this change, ITs are failing with

> CSC : error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.ImplementIDisposableCorrectly' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'. [C:\Workspace\sonar-csharp\sonaranalyzer-dotnet\its\sources\akka.net\src\core\Akka\Akka.csproj]

andrei-epure-sonarsource:
I created branch andrei/fix-S3881  to avoid breaking the CI loop

PR #1646, opened 2018-08-09
Fix S4261: Do not report on async Main

### LINKED ISSUE #1644: Fix S4261: False positive on async Main
### Description

A false positive S4261 is thrown on the line:

### Repro steps

```csharp
static async Task Main(string[] args)
```

### Expected behavior

Don't suggest to rename method Main

### Actual behavior

image

### Related information

* SonarC# Version 7.3 (build 5690)
* SonarScanner for MSBuild 4.1.1.1164 - .NET Fwk 4.6 (Jenkins plugin)

Evangelink:
Fix #1644

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNameAccordingToSynchronicity.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNameAccordingToSynchronicity.cs
@@ -69,14 +69,14 @@ namespace SonarAnalyzer.Rules.CSharp
                 c =>
                 {
                     var methodDeclaration = (MethodDeclarationSyntax)c.Node;
-                    if (methodDeclaration.Identifier.IsMissing ||
-                        methodDeclaration.Identifier.ValueText == "Main")
+                    if (methodDeclaration.Identifier.IsMissing)
                     {
                         return;
                     }
 
                     var methodSymbol = c.SemanticModel.GetDeclaredSymbol(methodDeclaration);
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNameAccordingToSynchronicity.cs
```diff
@@ -69,7 +69,8 @@ protected override void Initialize(SonarAnalysisContext context)
                 c =>
                 {
                     var methodDeclaration = (MethodDeclarationSyntax)c.Node;
-                    if (methodDeclaration.Identifier.IsMissing)
+                    if (methodDeclaration.Identifier.IsMissing ||
+                        methodDeclaration.Identifier.ValueText == "Main")
```
Evangelink: I could have done a more strict check to ensure this is a real main method but I think that's good enough like this.

duncanp-sonar: Why not call _SonarAnalyzer.Helpers.KnownMethods.IsMainMethod_?
It uses the symbol, but the code fetches that in line 78 anyway so it isn't going to affect perf much.

Evangelink: Well except that for all methods named `Main` you will actually not call the semantic model so you are faster.

PR #1652, opened 2018-08-09
Merging commits from branch-7.3.2

duncanp-sonar:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/UcfgInstructionFactory.cs
```diff
@@ -53,6 +53,28 @@ public UcfgInstructionFactory(SemanticModel semanticModel, UcfgExpressionService
         }
 
         public IEnumerable<Instruction> CreateFrom(SyntaxNode syntaxNode)
+        {
+            try
+            {
+                return UnsafeCreateFrom(syntaxNode);
+            }
+            catch (System.Exception e)
```
Evangelink: I didn't paid attention earlier but we are catching everything here. So if we have a NPE we will transform it into a `UcfgException` that will be swallowed.

duncanp-sonar: Before this change, every exception would have been caught by the higher-level try-catch in _UcfgFactoryCreate_ which converts everything into _UcfgException_ anyway. So this shouldn't change the overall behaviour.

PR #1664, opened 2018-08-10
Update RSPEC metadata + rules in SonarWay

### LINKED ISSUE #1658: Update S4261: Default severity should be Code Smell
Update the default severity for the rule.

### LINKED ISSUE #1660: Update S4524: Rule should be in the default quality profile (SonarWay)

### LINKED ISSUE #1661: Update S2255: Rule should be in the default quality profile (SonarWay)

### LINKED ISSUE #1662: Update S2245: Rule should be in the default quality profile (SonarWay)

### LINKED ISSUE #1663: Fix S4433: Remediation cost should be constant

Evangelink:
Fix #1663 
Fix #1662 
Fix #1661 
Fix #1660 
Fix #1658


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/RspecStrings.resx
```diff
@@ -10240,7 +10240,7 @@
     <value>switch can contain a default clause for various reasons: to handle unexpected values, to show that all the cases were properly considered.</value>
```
andrei-epure-sonarsource: I think this should be updated as well (locally, for me, it's updated)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/RspecStrings.resx
```diff
@@ -3262,7 +3262,7 @@
     <value>Attackers can use widely-available tools to view cookies and read any sensitive information they may contain. Even if the information is encoded in a way that is not human-readable, certain techniques could determine which encoding is being used, then decode the information.</value>
   </data>
   <data name="S2255_IsActivatedByDefault" xml:space="preserve">
-    <value>False</value>
+    <value>True</value>
   </data>
   <data name="S2255_Remediation" xml:space="preserve">
     <value>Constant/Issue</value>
```
andrei-epure-sonarsource: For \<data name="S4144_Description" xml:space="preserve"\>, locally I have    

> <value>When two methods have the same implementation, either it was a mistake - something else was intended - or the duplication was intentional, but may be confusing to maintainers. In the latter case, one implementation should invoke the other.</value>

andrei-epure-sonarsource:
Update: the changes which I reported in my previous comments must be reviewed, because they may imply a change in behavior which may have to be implemented before updating the RSPEC.

PR #1665, opened 2018-08-13
Fix S1144: Develop robust mechanism to detect if a private or internal member could be removed

### LINKED ISSUE #904: Fix S1144: rule should not report false positives with constants
In the below example, the `Sample.Constants` class is faded out by S1144.

```C#
using System;

class Program
{
    static void Main(string[] args)
    {
        Console.WriteLine(Sample.Constants.X);
    }
}

public static class Sample
{
    internal static class Constants // Class reported as unused
    {
        public const string X = nameof(X);
    }
}
```

### LINKED ISSUE #1195: Fix S1144: Issues is raised while method is being used (DebuggerDisplayAttribute)
Copy-paste from https://github.com/SonarSource/sonarlint-visualstudio/issues/571

### Description

There are moments when .Net uses reflection as the primary access point for interaction.  Sometimes it is beneficial to take advantage of this pattern, but that is neither here nor there.  I am not sure if it is even possible to track this, but might as well mention it.
When using reflection for class member access it can sometimes appear that a method/properties is never referenced.   (RSPEC-1144)

### Repro steps

1. DebuggerDisplayAttribute
```
    [DebuggerDisplay("{DebugDisplay()}")]
    public class Foo
    {
        public string Bar { get; set; }

        string DebugDisplay()
        {
            return $"Bar: {this.Bar}";
        }
    }

```
2. Build with SonarAnalyzer - Appears in Messages (not warnings)

### Expected behavior

Should not be flagged as "unreferenced"

### Actual behavior

warning S1144: Remove the unused private method 'DebugDisplay'.

### Known workarounds

Ignore it.

### Related information

* SonarLint for VS Version: 6.8.2.4717
* Visual Studio Version: 14.0.25424.00 UPdate 3
* SonarQube server version: 5.6.7 
* Languages in your solution: C#

### LINKED ISSUE #1225: Fix S1144: False Positive on Inner Classes
Please see closed ticket #1026.  Please re-open this ticket or create another.  Attached is a better example of this false positive.  Notice in the attached pic and sample code that there's 1 reference to OtherInnerClass, its grayed out and it's throwing up the S1144 message.

Code:
[TestClass.zip](https://github.com/SonarSource/sonar-csharp/files/1798756/TestClass.zip)

Pic:
s1144falsepositive

* SonarC# Version - 3.10.0.3095
* Visual Studio Version - 2017, 15.6.1

### LINKED ISSUE #1398: S1144 False positive for protected ctor
```
public class Foo {

	private abstract class Bar
	{
		private readonly bool _stopOnFirstViolation;

// this cannot be removed!
		protected Bar(
			bool stopOnFirstViolation
		)
		{
			_stopOnFirstViolation = stopOnFirstViolation;
		}

		public abstract string ItemsName { get; }
	}

	private class Baz : Bar
	{
		public Bar(bool stopOnFirstViolation)
			: base(stopOnFirstViolation)
		{
		}

		public override string ItemsName => "Users";
	}
}
```

### LINKED ISSUE #1623: Update S1144: Develop robust mechanism to detect when a class member is unused
The current implementation of the rule generates many false positives and is very difficult to change.

valhristov:
Fix #1623 

As well as the following tickets:
Fix #1398
Fix #904 
Fix #1195 
Fix #1225


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UnusedPrivateMemberTest.cs
```diff
@@ -29,9 +29,637 @@ public class UnusedPrivateMemberTest
...
+    {
+        new[] { 1, 2, 3 }.Select(Method12);
+    }
+}
+", new UnusedPrivateMember());
+        }
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void DebuggerDisplay_Attribute()
```
duncanp-sonar: You've labelled one of the regression tests with the related bug number, which is really useful when reviewing and for traceability later. Could you tag the tests for the other tickets with the related bug number too please?

This one is https://github.com/SonarSource/sonar-csharp/issues/1195

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/SymbolUsageCollector.cs
```diff
@@ -0,0 +1,284 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace SonarAnalyzer.Helpers
+{
+    public class SymbolUsageCollector : CSharpSyntaxWalker
```
duncanp-sonar: Does this class need to be public? The other collector classes are internal.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/RemovableSymbolCollector.cs
```diff
@@ -0,0 +1,189 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.Common;
+
+namespace SonarAnalyzer.Helpers
+{
+    internal class RemovableSymbolCollector : CSharpSyntaxWalker
```
duncanp-sonar: It would be neater if the "walker" part was implemented in an inner class that was only visible to the outer collector. The public API of the collector would then only be the public methods that intended to be called directly, and wouldn't include all public methods of CSharpSyntaxWalker.
It would also mean the logic in UnusedPrivateMembers.VisitDeclaringReferences method could be moved to the outer collector.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/RemovableSymbolCollector.cs
```diff
@@ -0,0 +1,189 @@
...
+        private bool IsRemovableType(ISymbol typeSymbol)
+        {
+            var accessibility = typeSymbol.GetEffectiveAccessibility();
+            return typeSymbol.ContainingType != null
+                && (accessibility == Accessibility.Private || accessibility == Accessibility.Internal)
+                && IsRemovable(typeSymbol);
+        }
+
+        private bool IsRemovableMember(ISymbol symbol) =>
+            symbol.GetEffectiveAccessibility() == Accessibility.Private &&
```
duncanp-sonar: Does this mean that internal properties, fields and events will only be considered removable? If so, why are they treated differently from types and methods?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UnusedPrivateMemberTest.cs
```diff
@@ -29,9 +29,637 @@ public class UnusedPrivateMemberTest
...
+    protected int protectedField;
+    protected internal int protectedInternalField;
+    public int publicField;
+    internal static int internalStaticField;
+    protected static int protectedStaticField;
+    protected internal static int protectedInternalStaticField;
+    public static int publicStaticField;
+
+    public class InnerPublicClass
+    {
```
duncanp-sonar: What about private and private static fields on inner classes - are they reported on? Same for the corresponding properties test.
Either way, we should have a test case that covers those cases to documented what the expected behaviour is.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UnusedPrivateMemberTest.cs
```diff
@@ -29,9 +29,640 @@ public class UnusedPrivateMemberTest
...
+    private class InnerPrivateClass // Noncompliant
+    {
+        internal int internalField; // Noncompliant
+        protected int protectedField; // Noncompliant
+        protected internal int protectedInternalField; // Noncompliant
+        public int publicField; // Noncompliant
+        internal static int internalStaticField; // Noncompliant
+        protected static int protectedStaticField; // Noncompliant
+        protected internal static int protectedInternalStaticField; // Noncompliant
+        public static int publicStaticField; // Noncompliant
```
duncanp-sonar: If an inner class is unused, could we avoid reporting on any unused member in that type too reduce the noise?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/SymbolUsageCollector.cs
```diff
@@ -0,0 +1,284 @@
...
+                {
+                    // Prop += value --> get/set
+                    return AccessorAccess.Both;
+                }
+            }
+
+            // nameof(Prop) --> get/set
+            if (node.IsInNameofCall(getSemanticModel(node)))
+            {
+                return AccessorAccess.Both;
```
duncanp-sonar: Why is nameof(X) treated as accessing both get and set?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
```diff
@@ -80,462 +65,182 @@ protected sealed override void Initialize(SonarAnalysisContext context)
...
-        private static bool IsInternalVisibleToAttribute(AttributeSyntax attribute, SemanticModel semanticModel)
+        private static IEnumerable<Diagnostic> GetDiagnostics(SymbolUsageCollector usageCollector, ISet<ISymbol> removableSymbols,
+            string accessibility, BidirectionalDictionary<ISymbol, SyntaxNode> fieldLikeSymbols)
         {
+            var unusedSymbols = removableSymbols
+                .Except(usageCollector.UsedSymbols)
+                .Where(symbol => !MentionedInDebuggerDisplay(symbol))
+                .ToHashSet();
+
+            var propertiesWithUnusedAccessor = removableSymbols
```
duncanp-sonar: Why is this value calculated here and passed in to _GetDiagnosticsForPropertyAccessors_, rather than just being calculated in _GetDiagnosticsForPropertyAccessors_?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
```diff
@@ -80,462 +65,182 @@ protected sealed override void Initialize(SonarAnalysisContext context)
...
                 });
         }
 
-        private static bool IsInternalVisibleToAttribute(AttributeSyntax attribute, SemanticModel semanticModel)
+        private static IEnumerable<Diagnostic> GetDiagnostics(SymbolUsageCollector usageCollector, ISet<ISymbol> removableSymbols,
+            string accessibility, BidirectionalDictionary<ISymbol, SyntaxNode> fieldLikeSymbols)
         {
+            var unusedSymbols = removableSymbols
+                .Except(usageCollector.UsedSymbols)
+                .Where(symbol => !MentionedInDebuggerDisplay(symbol))
```
duncanp-sonar: Can properties be mentioned in DebuggerDisplay? Looks like this check is not be applied when reporting diagnostics on properties.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/SymbolUsageCollector.cs
valhristov: Which one should it be? :)

If we assume that when a property should not be removed when it is used in a `nameof` expression, I think we should mark both accessors as used. On the other hand, perhaps we should suggest members for removal if they are only used in `nameof`... I didn't check if the ITs will raise many issues when the `nameof` exclusion is removed.

This functionality was added two years ago when the property support was implemented and I decided to keep it like this.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
```diff
@@ -80,462 +65,182 @@ protected sealed override void Initialize(SonarAnalysisContext context)
...
         }
 
-        private static void ReportUnusedPropertyAccessors(SymbolAnalysisContext context, HashSet<ISymbol> usedSymbols,
-            HashSet<ISymbol> declaredPrivateSymbols,
+        private static IEnumerable<Diagnostic> GetDiagnosticsForPropertyAccessors(IEnumerable<IPropertySymbol> onlyOneAccessorAccessed,
             Dictionary<IPropertySymbol, AccessorAccess> propertyAccessorAccess)
         {
-            var usedPrivatePropertis = declaredPrivateSymbols.Intersect(usedSymbols).OfType<IPropertySymbol>();
-            var onlyOneAccessorAccessed = usedPrivatePropertis.Where(p => propertyAccessorAccess.ContainsKey(p));
+            return onlyOneAccessorAccessed.SelectMany(GetDiagnosticsForProperty).ToList();
```
duncanp-sonar: I find this style of using inner functions much harder to read and understand (particularly here, where the method is one line, but contains a twenty line inner function and another inner function after that. I have a similar problem with the  _GetDiagnosticsForMembers_ method; there's a _return_ statement half-way down the method, and everything after that is inner functions.

I'm not sure what a good style guideline would be. Perhaps limiting local functions to things that can be expressed in one or two lines using expression syntax?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
valhristov: The code is not much better when the collection is moved inside the other method

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/RemovableSymbolCollector.cs
duncanp-sonar: Naming: a doc comment on the class explaining that this class collects _potentially_ removable symbols (i.e. symbols that could be removed if they are not used) would help.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/SymbolUsageCollector.cs
```diff
@@ -0,0 +1,284 @@
...
+        public HashSet<ISymbol> UsedSymbols { get; } =
+            new HashSet<ISymbol>();
+
+        public HashSet<string> DebuggerDisplayValues { get; } =
+            new HashSet<string>();
+
+        public Dictionary<IPropertySymbol, AccessorAccess> PropertyAccess { get; } =
+            new Dictionary<IPropertySymbol, AccessorAccess>();
+
+        public SymbolUsageCollector(Func<SyntaxTree, bool, SemanticModel> getSemanticModel, HashSet<string> removableSymbolNames)
```
duncanp-sonar: Naming: _knownIdentifierNames_ would be better than _removableSymbolNames_ - from the point of view of this class, it doesn't know or care about removability.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
valhristov: +1, it looks better with less nested methods.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
valhristov: Also, when you mentioned the `MentionedInDebuggerDisplay`, I remember it is actually the reason for calculating the collections outside...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/RemovableDeclarationCollector.cs
```diff
@@ -147,13 +143,5 @@ private static SyntaxNodeSymbolSemanticModelTuple SelectNodeTuple(SyntaxNode nod
             return container.SyntaxNode.DescendantNodes(IsNodeContainerTypeDeclaration)
                 .Where(node => kinds.Contains(node.Kind()));
         }
-
-        private static bool IsSymbolNotInInterfaceAndRemovable(ISymbol symbol, Accessibility maxAccessibility)
```
duncanp-sonar: It would help the review process if you could add context to some of the changes e.g. _Not related to the current bug fix. IsRemovable already checks IsInterface etc so IsSymbolNotInInterfaceAndRemovable is redundant._

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UnusedPrivateMemberTest.cs
valhristov: We don't report internal members other than types. There is no ticket about that and the functionality was not present before and I decided to not increase the scope of the current work. There are test cases for internal fields, properties and methods, even though there are no comments saying why they are not reported.

valhristov: I generally think that reporting internal members should not be part of this rule (because the rule is about private members and internal is not private). I did not check if the other languages report on members with accessibility that corresponds to internal, though.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UnusedPrivateMemberTest.cs
valhristov: Sounds like a good candidate for a new ticket.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/RemovableDeclarationCollector.cs
valhristov: point taken

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UnusedPrivateMemberTest.cs
```diff
@@ -29,9 +29,640 @@ public class UnusedPrivateMemberTest
...
+public class PrivateMembers
+{
+    private int privateField; // Noncompliant {{Remove the unused private field 'privateField'.}}
+//  ^^^^^^^^^^^^^^^^^^^^^^^^^
+    private static int privateStaticField; // Noncompliant
+
+    private class InnerPrivateClass // Noncompliant
+    {
+        internal int internalField; // Noncompliant
+        protected int protectedField; // Noncompliant
```
duncanp-sonar: Do we report on unused _protected_ methods in internal classes?

valhristov: No, we report on members with "effective" visibility "private" or "internal" when they are classes.

valhristov: "Effective" is computed visibility

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
```diff
@@ -80,462 +65,182 @@ protected sealed override void Initialize(SonarAnalysisContext context)
...
-                           }
-
-                           foreach (var diagnostic in possibleUnusedInternalMembers)
-                           {
-                               cc.ReportDiagnosticIfNonGenerated(diagnostic, cc.Compilation);
-                           }
-                       });
+                        cc =>
+                        {
+                            var foundInternalsVisibleTo = cc.Compilation.Assembly
```
duncanp-sonar: This would probably be a useful common helper function.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/SymbolUsageCollector.cs
```diff
@@ -27,6 +27,10 @@
 
 namespace SonarAnalyzer.Helpers
 {
+    /// <summary>
+    /// Collects all symbol usages from a class declaration. Ignores symbols whose names are not present
+    /// in the removableSymbolNames collection for performance reasons.
```
duncanp-sonar: _knownSymbolNames_ now :)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
```diff
@@ -44,31 +43,17 @@ public class UnusedPrivateMember : SonarDiagnosticAnalyzer
...
-                                .Any(a => IsInternalVisibleToAttribute(a, cc.SemanticModel));
-                            if (isInternalsVisibleToAttributeFound)
-                            {
-                                shouldRaise = false;
-                            }
-                        });
+                    // Collect potentially removable internal members from the project to evaluate when
+                    // the compilation is over, depending on whether InternalsVisibleTo attribute is present
+                    // or not.
+                    var removableInternalMembers = new ConcurrentBag<ISymbol>();
```
duncanp-sonar: If where only interested in removable internal _types_ then this variable should be renamed.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
```diff
@@ -80,486 +65,209 @@ protected sealed override void Initialize(SonarAnalysisContext context)
...
+                                .Any();
 
-        private static bool IsInternalVisibleToAttribute(AttributeSyntax attribute, SemanticModel semanticModel)
-        {
+                            if (foundInternalsVisibleTo)
+                            {
+                                return;
+                            }
+
+                            var usageCollector = new SymbolUsageCollector(
```
duncanp-sonar: Optimisation - no point in re-processing all named types if there are no removable internal types.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
```diff
@@ -80,486 +65,209 @@ protected sealed override void Initialize(SonarAnalysisContext context)
...
+                                c.Compilation.GetSemanticModel,
+                                removableInternalMembers.Select(s => s.Name).ToHashSet());
 
-            return semanticModel.GetSymbolInfo(attribute).Symbol is IMethodSymbol attributeConstructor &&
-                attributeConstructor.ContainingType.Is(KnownType.System_Runtime_CompilerServices_InternalsVisibleToAttribute);
+                            foreach (var symbol in allNamedTypes)
```
duncanp-sonar: Corner case: walking named types won't catch references to fields/properties/methods that are in assembly-level attributes e.g. _AssemblyVersion_. The only way to catch those would be to walk all of the syntax trees in the assembly. I can't think of any other ways of refering to a field/property etc outside of type, but there might be some.

I suspect we will get some false-positives from this, but not many so there's probably no need to handle this case now. However, it would be adding a test case and marking it as a known false positive.

'''
[assembly: AssemblyCompany(Ns1.MyAppDetails.AppName)]
namespace Ns1
{
    internal static class MyAppDetails
    {
        public const string AppName = "myApp";
    }
}
'''

PR #1666, opened 2018-08-13
Update test helpers to simplify working with NuGet on test cases

Evangelink:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/AssemblyReference.cs
```diff
@@ -18,62 +18,93 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System;
+using System.Collections.Generic;
```
Evangelink: This is the only class who really changed. The rest is just related to using this new class.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/AssemblyReference.cs
```diff
@@ -18,62 +18,93 @@
...
-        public NuGetInfo NuGet { get; }
+            return aggregateRepository;
+        }
 
-        private AssemblyReference(string assemblyName, NuGetInfo nuGetInfo = null)
+        public static MetadataReference FromFrameworkAssembly(string assemblyName)
         {
-            Name = assemblyName;
-            NuGet = nuGetInfo;
+            return metadataReferenceCache.GetOrAdd(assemblyName, ProcessAssembly).Single();
```
valhristov: Could the cache entry contain zero or more than one items? Why not convert to non-array entries?

Evangelink: I am not expecting something else than one + if that's the case I want to know about it hence this failure.

valhristov: Please, add a proper check with a descriptive error message then. It would be much easier for the developer who will encounter the Linq exception after several months.

PR #1671, opened 2018-08-13
Deprecate S2758 in favor of S3923

### LINKED ISSUE #1670: Deprecate S2758 in favor of S3923
### Description

This rule is now deprecated and is replaced by S3923.

Evangelink:
Fix #1670

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AllBranchesShouldNotHaveSameImplementation.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AllBranchesShouldNotHaveSameImplementation.cs
@@ -100,7 +100,8 @@ namespace SonarAnalyzer.Rules.CSharp
                 {
                     var ternaryStatement = (ConditionalExpressionSyntax)c.Node;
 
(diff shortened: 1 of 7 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AllBranchesShouldNotHaveSameImplementation.cs
```diff
@@ -100,7 +100,8 @@ protected override void Initialize(SonarAnalysisContext context)
                 {
                     var ternaryStatement = (ConditionalExpressionSyntax)c.Node;
 
-                    if (ternaryStatement.WhenTrue.IsEquivalentTo(ternaryStatement.WhenFalse, false))
+                    if (ternaryStatement.WhenTrue.RemoveParentheses().IsEquivalentTo(
+                            ternaryStatement.WhenFalse.RemoveParentheses(), false))
```
valhristov: Is this related to the current change set?

Evangelink: Yes when copying the test cases from the deprecated rules I have noticed we were not handling parenthesis (on the new rule) while the deprecated one was.

PR #1678, opened 2018-08-13
Fix S1144: Unused private members should not report false positives w…

### LINKED ISSUE #159: Fix S1144: Unused private members should not report false positives with Unity classes
We should investigate the options:
* whitelist affected classes, such as https://docs.unity3d.com/ScriptReference/MonoBehaviour.html, docs.unity3d.com/ScriptReference/AssetPostprocessor.html
* introduce a whitelist parameter which can be set by the user in SQ. Note that the list might be very long, (I'm not sure if there's a limit in SQ for parameter length), and that this rule should work out of the box with SonarLint with no parameters (at least for non Unity projects). 

See [SONARCS-677](https://jira.sonarsource.com/browse/SONARCS-677) for the full thread of discussion

valhristov:
…ith Unity classes

Fix #159 

The ticket mentions more types than the added to the IgnoredTypes collection. They however, are all subclasses of `MonoBehaviour` and `ScriptableObject`, hence they will be automatically ignored.

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UnusedPrivateMemberTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UnusedPrivateMemberTest.cs
@@ -716,6 +716,46 @@ public class NewClass
 
         [TestMethod]
         [TestCategory("Rule")]
+        public void Unity3D_Ignored()
+        {
+            Verifier.VerifyCSharpAnalyzer(@"
+// https://github.com/SonarSource/sonar-csharp/issues/159
+public class UnityMessages1 : UnityEngine.MonoBehaviour
+{
+    private void SomeMethod(bool hasFocus) { } // Compliant
+}
+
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UnusedPrivateMemberTest.cs
```diff
@@ -714,6 +714,46 @@ public class NewClass
...
+{
+    private void SomeMethod(bool hasFocus) { } // Compliant
+}
+
+public class UnityMessages4 : UnityEditor.AssetModificationProcessor
+{
+    private void SomeMethod(bool hasFocus) { } // Compliant
+}
+
+// Unity3D does not seem to to available as a nuget package and we cannot use the original classes
```
duncanp-sonar: FYI I don't think it not being available as a NuGet package is a problem. Since we don't check the assembly the class is defined in, I don't think there's much value in referencing the real package over defining our own classes with the expected names anyway - it just adds bloat to the tests, and doesn't scale well.

PR #1680, opened 2018-08-14
Fix S4143: False Positive when variable is reassigned

### LINKED ISSUE #1643: Fix S4143: False Positive when variable is reassigned
### Description

S4143 Rule causes a lot of issues throughout our application analyses, however, most of them are false positive. This rule is unable to detect object c

### Repro steps

```csharp
dr["CurrencyCode"] = General.GeneralParametersFinder.LocalCurrencyCode();
dt.Rows.Add(dr);
dr = dt.NewRow();
dr["CurrencyCode"] = Framework.GeneralParameters.E_U_R;  // False issue raised here. dr refers to a new DataRow because of the previous line
```

### Related information

From https://community.sonarsource.com/t/sonarc-7-3-collection-elements-should-not-be-replaced-unconditionally-false-positive-problem/1622

Evangelink:
Fix #1643

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotOverwriteCollectionElements.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotOverwriteCollectionElements.cs
@@ -98,7 +98,8 @@ namespace SonarAnalyzer.Rules.CSharp
 
                    if (methodName == null ||
                        methodName != "Add" ||
-                       invokedOn == null)
+                       invokedOn == null ||
+                       !IsDictionaryAdd(c.SemanticModel.GetSymbolInfo(invocation).Symbol as IMethodSymbol))
                    {
                        return;
                    }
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotOverwriteCollectionElements.cs
```diff
@@ -44,110 +44,131 @@ public sealed class DoNotOverwriteCollectionElements : SonarDiagnosticAnalyzer
...
+                   {
+                       return;
+                   }
+
+                   var keyValue = invocation.ArgumentList.Arguments[0].ToString();
+
+                   var previousAddInvocationOnVariable = GetPreviousStatements(invocation)
+                        .OfType<ExpressionStatementSyntax>()
+                        .Select(ess => ess.Expression)
+                        .TakeWhile(e => IsInvocationOnSameItem(e, invokedOn))
```
valhristov: We don't detect changes like this anymore, do we?
```
dictionary.Add("key", "something");
dictionary["key"] = "other";
```

PR #1696, opened 2018-08-15
Fix False Positives related to mutable fields

### LINKED ISSUE #505: Fix S2386: Rule should not report when field is readonly and ...
... initialized with immutable type in ctor

### Description

When a `public static readonly` mutable field is initialized through the constructor with an immutable type the rule should not raise an issue.

### Expected behavior

Issue should not be reported.

### Actual behavior

Issue is reported.

### Related information
[RSPEC-2386](https://jira.sonarsource.com/browse/RSPEC-2386)

Related bug: #1491 (same issue, the rules share a base class)

### LINKED ISSUE #1491: Fix S3887: Rule should not report when field is readonly and initialized with immutable type in ctor
### Description

When a `public readonly` mutable field is initialized through the constructor with an immutable type the rule should not raise an issue.

### Expected behavior

Issue should not be reported.

### Actual behavior

Issue is reported.

### Related information
https://jira.sonarsource.com/browse/RSPEC-3887

Related bug: #505 (same issue, the rules share a base class)

### LINKED ISSUE #1693: Fix S2386: Rule should handle effective accessiblity
### Description

The rule currently look only at the field modifier without considering the class accessibility which can lead to some FP.

### LINKED ISSUE #1694: Fix S3887: Rule should handle effective accessiblity
### Description

The rule currently look only at the field modifier without considering the class accessibility which can lead to some FP.

### LINKED ISSUE #1695: Fix S3887: Rule should not raise for uninitialized readonly fields
### Description

When a mutable `readonly` field is not initialized, this is equivalent to being initialized with `null`. As the rule doesn't report on the latter it should not report on uninitialized fields.

Evangelink:
Fix #505 
Fix #1693
Fix #1491
Fix #1694
Fix #1695


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MutableFieldsShouldNotBe.cs
```diff
@@ -58,76 +60,146 @@ public abstract class MutableFieldsShouldNotBe : SonarDiagnosticAnalyzer
 
         protected sealed override void Initialize(SonarAnalysisContext context)
         {
-            context.RegisterSyntaxNodeActionInNonGenerated(CheckForIssue, SyntaxKind.FieldDeclaration);
+            context.RegisterSyntaxNodeActionInNonGenerated(CheckForIssue,
+                SyntaxKind.ClassDeclaration,
+                SyntaxKind.StructDeclaration);
         }
 
         private void CheckForIssue(SyntaxNodeAnalysisContext analysisContext)
```
valhristov: We should start standardizing the names of these methods...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MutableFieldsShouldNotBe.cs
```diff
@@ -58,76 +60,146 @@ public abstract class MutableFieldsShouldNotBe : SonarDiagnosticAnalyzer
...
-            var symbolInfo = model.GetSymbolInfo(initializer);
+            if (typeSymbol is INamedTypeSymbol namedTypeSymbol)
+            {
+                return !namedTypeSymbol.ConstructedFrom.DerivesOrImplementsAny(MutableBaseTypes) ||
+                    namedTypeSymbol.ConstructedFrom.DerivesOrImplementsAny(ImmutableBaseTypes);
+            }
 
-            return symbolInfo.Symbol ?? symbolInfo.CandidateSymbols.FirstOrDefault();
+            return !typeSymbol.DerivesOrImplementsAny(MutableBaseTypes) ||
+                typeSymbol.DerivesOrImplementsAny(ImmutableBaseTypes);
```
valhristov: The two returns could be reduced to one:
```
if (typeSymbol is INamedTypeSymbol namedTypeSymbol)
{
    typeSymbol = namedTypeSymbol.ConstructedFrom;
}
return !typeSymbol.Derives... || ...;
```

PR #1703, opened 2018-08-15
[DO NOT SQUASH] Fix S3242: Rule should not suggest base type resulting in inconsistent accessibility

### LINKED ISSUE #1610: Fix S3242: Rule should not suggest base type resulting in inconsistent accessibility (bis)
### Description

This is similar to #863, for which the fix apparently did not cover all situations.

### Repro steps

1. Create a solution with 2 projects

2. In one project, add the following code:
```csharp
internal interface IFoo
{
    bool IsFoo { get; }
}

public class Foo : IFoo
{
    public bool IsFoo { get; set; }
}
```

3. In the other project, reference the project that defines `Foo`, and add the following code:
```csharp
internal class Bar
{
    public void MethodOne(Foo foo)
    {
        var x = foo.IsFoo;
    }
}
```

### Expected behavior

A violation of S3242 should not be raised, because `IFoo` is not accessible from the project defining Bar.

### Actual behavior

A violation of S3242 is raised.
Note: If the `Bar` class is made public, that does not happen.

### Related information

* SonarC# Version 7.3
* Visual Studio Version 15.7

valhristov:
+ Creating NewVerifier in an attempt to replace the existing Verifier with more modular framework. NOTE: this is work in progress and some of the code in Verifier will be replaced with calls to the new classes.

Fix #1610

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldUseBaseTypes.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldUseBaseTypes.cs
@@ -45,13 +45,14 @@ namespace SonarAnalyzer.Rules.CSharp
         protected override void Initialize(SonarAnalysisContext context)
         {
             context.RegisterSyntaxNodeActionInNonGenerated(
-                c => FindViolations(c).ForEach(d => c.ReportDiagnosticWhenActive(d)),
+                c => FindViolations((MethodDeclarationSyntax)c.Node, c.SemanticModel)
+                        .ForEach(d => c.ReportDiagnosticWhenActive(d)),
                 SyntaxKind.MethodDeclaration);
         }
 
-        private List<Diagnostic> FindViolations(SyntaxNodeAnalysisContext context)
+        private List<Diagnostic> FindViolations(MethodDeclarationSyntax methodDeclaration, SemanticModel semanticModel)
         {
-            if (!(context.SemanticModel.GetDeclaredSymbol(context.Node) is IMethodSymbol methodSymbol) ||
+            if (!(semanticModel.GetDeclaredSymbol(methodDeclaration) is IMethodSymbol methodSymbol) ||
                 methodSymbol.Parameters.Length == 0 ||
                 methodSymbol.IsOverride ||
                 methodSymbol.IsVirtual ||
@@ -69,11 +70,10 @@ namespace SonarAnalyzer.Rules.CSharp
                 .GroupBy(p => p.Name)
                 .ToDictionary(p => p.Key, p => new ParameterData(p.First(), methodAccessibility));
 
-            var parameterUsesInMethod = context
-                .Node
+            var parameterUsesInMethod = methodDeclaration
                 .DescendantNodes()
                 .OfType<IdentifierNameSyntax>()
-                .Where(id => parametersToCheck.Values.Any(p => p.MatchesIdentifier(id, context.SemanticModel)));
+                .Where(id => parametersToCheck.Values.Any(p => p.MatchesIdentifier(id, semanticModel)));
 
             foreach (var identifierReference in parameterUsesInMethod)
             {
@@ -92,7 +92,7 @@ namespace SonarAnalyzer.Rules.CSharp
                     continue;
                 }
 
-                var symbolUsedAs = FindParameterUseAsType(identifierReference, context.SemanticModel);
+                var symbolUsedAs = FindParameterUseAsType(identifierReference, semanticModel);
                 if (symbolUsedAs != null)
                 {
                     paramData.AddUsage(symbolUsedAs);
@@ -114,7 +114,7 @@ namespace SonarAnalyzer.Rules.CSharp
                    !type.Is(KnownType.System_String);
         }
 
-        private SyntaxNode GetNextUnparenthesizedParent(SyntaxNode node)
+        private static SyntaxNode GetFirstNonParenthesizedParent(SyntaxNode node)
         {
             if (node is ExpressionSyntax expression)
             {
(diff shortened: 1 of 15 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldUseBaseTypes.cs
```diff
@@ -197,24 +197,41 @@ private ITypeSymbol HandlePropertyOrField(IdentifierNameSyntax identifier, ISymb
...
+
+            // Do not suggest internal types that are declared in an assembly different than
+            // the one that's declaring the parameter. Such types should not be suggested at
+            // all if there is no InternalsVisibleTo attribute present in the compilation.
+            // Since the check for the attribute must be done in CompilationEnd thus making
+            // the rule unusable in Visual Studio, we will not suggest such classes and will
+            // generate some False Negatives.
+            bool IsNotInternalOrSameAssembly(INamedTypeSymbol namedTypeSymbol) =>
+                namedTypeSymbol.ContainingAssembly == usageSite ||
+                namedTypeSymbol.GetEffectiveAccessibility() != Accessibility.Internal;
```
valhristov: That's the important change - we ignore internal types that originate from another assembly than the one that is declaring the analyzed parameter

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldUseBaseTypes.cs
```diff
@@ -114,7 +114,7 @@ private static bool IsTrackedParameter(IParameterSymbol parameterSymbol)
                    !type.Is(KnownType.System_String);
         }
 
-        private SyntaxNode GetNextUnparenthesizedParent(SyntaxNode node)
+        private static SyntaxNode GetNextUnparenthesizedParent(SyntaxNode node)
```
Evangelink: I think we have an helper method doing the exact same thing.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldUseBaseTypes.cs
```diff
@@ -124,45 +124,43 @@ private SyntaxNode GetNextUnparenthesizedParent(SyntaxNode node)
...
                 if (conditionalAccess.WhenNotNull is MemberBindingExpressionSyntax binding)
                 {
-                    var name = binding.Name;
-                    if (name == null)
-                    {
-                        return null;
-                    }
-
-                    return HandlePropertyOrField(identifier, semanticModel.GetSymbolInfo(name).Symbol);
+                    return binding.Name == null
```
Evangelink: [minor] I find it more readable with the opposite condition (i.e. when `null` is the else part).

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        public IReadOnlyList<ProjectId> ProjectIds => solution.ProjectIds;
+
+        public ImmutableArray<ParseOptions> DefaultParseOptions { get; }
+
+        private SolutionBuilder(Solution solution)
+        {
+            this.solution = solution;
+
+            DefaultParseOptions = ImmutableArray.Create<ParseOptions>(
```
Evangelink: [cosmetic] could be initialized inline.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+
+namespace SonarAnalyzer.UnitTest.TestFramework
+{
+    public struct SolutionBuilder
+    {
+        private readonly Solution solution;
+
+        public IReadOnlyList<ProjectId> ProjectIds => solution.ProjectIds;
+
+        public ImmutableArray<ParseOptions> DefaultParseOptions { get; }
```
Evangelink: It could be `static`.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            return solution
+                .Projects
+                .SelectMany(project => parseOptions
+                    .Where(GetParseOptionsFilter(project.Language))
+                    .Select(options => GetCompilation(project, options)))
+                .ToList()
+                .AsReadOnly();
+        }
+        private static Compilation GetCompilation(Project project, ParseOptions options) =>
```
Evangelink: [cosmetic] missing blank line before.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+            project
+                .WithParseOptions(options)
+                .GetCompilationAsync()
+                .Result;
```
Evangelink: Why are you not using `await`?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using SonarAnalyzer.Common;
+using CS = Microsoft.CodeAnalysis.CSharp;
+using VB = Microsoft.CodeAnalysis.VisualBasic;
+
+namespace SonarAnalyzer.UnitTest.TestFramework
+{
+    public struct SolutionBuilder
```
Evangelink: Could be `internal`.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/ProjectBuilder.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        public Project Project { get; }
+
+        public SolutionBuilder GetSolution() =>
+            solutionWrapper.Value;
+
+        private ProjectBuilder(Project project)
+        {
+            Project = project;
+            solutionWrapper = new Lazy<SolutionBuilder>(() => SolutionBuilder.FromSolution(project.Solution));
```
Evangelink: Could be initialized inline.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/ProjectBuilder.cs
```diff
@@ -0,0 +1,147 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using Microsoft.CodeAnalysis;
+
+namespace SonarAnalyzer.UnitTest.TestFramework
+{
+    public struct ProjectBuilder
```
Evangelink: Could be `internal`.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
valhristov: I don't want to change all tests to asynchronous, we could do it in the future, even though it is not required.

Evangelink: We cannot update the version of Roslyn we are using for unit test because of some await (and probably something else).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldUseBaseTypes.cs
valhristov: Not exactly ;) This one is returning the parent of the top parenthesized expression...

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
valhristov: Created test for that #1730

PR #1704, opened 2018-08-15
Fixed #3433 - added support for all types of test methods

### LINKED ISSUE #3433: S4261: FP on methods which return IAsyncEnumerable<T>
### Description

Issues are raised for methods which end with "Async" and return `IAsyncEnumerable<T>`.

### Repro steps

```
using System;

namespace Tests.Diagnostics
{
    public abstract class IAsyncEnumerableReproducer
    {
        protected abstract IAsyncEnumerable<int> GetFilesAsync(string folderName); // Noncompliant - FP, async method
    }
}
```

### Expected behavior

Issue is not raised for methods which end with "Async" and return `IAsyncEnumerable<T>`.

### Actual behavior

Since the rule does not consider `IAsyncEnumerable<T>` return type, it will raise a FP in this case.

### Related information

* C#/VB.NET Plugins Version: https://github.com/SonarSource/sonar-dotnet/releases/tag/8.9.0.19135
* Source: https://community.sonarsource.com/t/rule-s4261-net-core-3-1-false-positive-on-iasyncenumerable/27326
* Rule: https://rules.sonarsource.com/csharp/RSPEC-4261

duncanp-sonar:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldHaveCorrectSignature.cs
```diff
@@ -44,83 +43,93 @@ public sealed class TestMethodShouldHaveCorrectSignature : SonarDiagnosticAnalyz
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
 
-        protected override void Initialize(SonarAnalysisContext context)
+        protected override void Initialize(SonarAnalysisContext context) =>
```
duncanp-sonar: I almost completely re-wrote the rule, so it's probably best not to try to compare it to the previous version.
There were two reasons for the re-write:
1) the previous version was looking at class declarations, then using _GetMethods_ on the class symbol to the methods.
2) I had to add support for another four or five types of test method attributes, some of which have different constraints on the signature e.g. Theories can be generic but Facts can't, so the validation checking got more complicated.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
```diff
@@ -95,5 +95,57 @@ public static bool IsMsTestOrNUnitTestIgnored(this IMethodSymbol method)
                     a.AttributeClass.Is(KnownType.Xunit_FactAttribute) ||
                     a.AttributeClass.Is(KnownType.Xunit_TheoryAttribute) ||
                     a.AttributeClass.Is(KnownType.LegacyXunit_TheoryAttribute));
+
+        public static KnownType FindTestAttribute(this IMethodSymbol method)
```
duncanp-sonar: This is effective the old _ToKnownTestType_ method from the test, but expanded to cover all of the known test method types.

I don't like having a big lookup method like this one, but couldn't think of a simpler way of translating from attribute symbols to _KnownTypes_.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldHaveCorrectSignature.Misc.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    using System.Threading.Tasks;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    public partial class TestThatIssuesAreOnlyReportedOnceForPartialClasses
+    {
+    }
+
+    [TestClass]
+    public partial class TestThatIssuesAreOnlyReportedOnceForPartialClasses
```
duncanp-sonar: The previous version would report each issue multiple times for partial classes since it was starting at the _ClassDeclarationSyntax_ node, then calling _GetMethods_ on the _ClassSymbol_.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldHaveCorrectSignature.MsTest.cs
```diff
@@ -44,6 +45,59 @@ class MsTestTest
...
+    }
+}
+
+namespace Tests.Diagnostics
+{
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using System.Threading.Tasks;
+
+    [TestClass]
+    public class MsTestTest_DataTestMethods
```
duncanp-sonar: MSTest - same test cases, but repeated for the _DataTestMethod_ attribute.

The tests for the other frameworks follow the same pattern - duplicating the existing test cases for the other attributes.

duncanp-sonar:
I've updated the rspec: https://jira.sonarsource.com/browse/RSPEC-3433

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
Evangelink: I suggest to rename to `FindFirstTestAttribute` or maybe even to return the collection of test attributes and to filter at the caller place.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
```diff
@@ -95,5 +95,57 @@ public static bool IsMsTestOrNUnitTestIgnored(this IMethodSymbol method)
...
+                        {
+                            return KnownType.Xunit_TheoryAttribute;
+                        }
+                        if (attribute.AttributeClass.Is(KnownType.LegacyXunit_TheoryAttribute))
+                        {
+                            return KnownType.LegacyXunit_TheoryAttribute;
+                        }
+
+                        return (KnownType)null;
+                    }).FirstOrDefault();
```
Evangelink: It's easier to read when dotted method calls are aligned. Please align this `FirstOrDefault` with the `Select`.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
```diff
@@ -95,5 +95,57 @@ public static bool IsMsTestOrNUnitTestIgnored(this IMethodSymbol method)
                     a.AttributeClass.Is(KnownType.Xunit_FactAttribute) ||
                     a.AttributeClass.Is(KnownType.Xunit_TheoryAttribute) ||
                     a.AttributeClass.Is(KnownType.LegacyXunit_TheoryAttribute));
+
+        public static KnownType FindTestAttribute(this IMethodSymbol method)
+        {
+            return method.GetAttributes()
+                .Select(
```
Evangelink: Please create a inner method for the content of this `Select`. It helps reading the chain of linq operations.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
```diff
@@ -95,5 +95,57 @@ public static bool IsMsTestOrNUnitTestIgnored(this IMethodSymbol method)
                     a.AttributeClass.Is(KnownType.Xunit_FactAttribute) ||
                     a.AttributeClass.Is(KnownType.Xunit_TheoryAttribute) ||
                     a.AttributeClass.Is(KnownType.LegacyXunit_TheoryAttribute));
+
+        public static KnownType FindTestAttribute(this IMethodSymbol method)
+        {
+            return method.GetAttributes()
```
Evangelink: I have the feeling it would be clearer the other way around. Something like (I am not writing from the IDE so code might not compile):

```csharp
supportedTestAttributes.FirstOrDefault(testAttribute => 
	method.GetAttributes().Any(attribute => attribute.Is(testAttribute)));
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TestMethodShouldHaveCorrectSignature.cs
```diff
@@ -63,5 +63,28 @@ public void TestMethodShouldHaveCorrectSignature_Xunit(string testFwkVersion)
                     .Concat(MetadataReferenceHelper.FromNuGet("xunit.extensibility.core", testFwkVersion))
                     .ToArray());
         }
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void TestMethodShouldHaveCorrectSignature_Xunit_Legacy()
```
Evangelink: I haven't tested but maybe we could still have a `DataTestMethod` testing with version `1.7.0.1540` (older) and `1.9.1`.

Also that's surprising the version `1.9.2` doesn't also have this behavior.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldHaveCorrectSignature.cs
```diff
@@ -44,83 +43,116 @@ public sealed class TestMethodShouldHaveCorrectSignature : SonarDiagnosticAnalyz
...
-                            return KnownType.Xunit_FactAttribute;
-                        }
-                        else
-                        {
-                            return null;
-                        }
-                    })
-                .FirstOrDefault();
+            // Find the first matching attribute type in the table
+            var type = method.FindTestAttribute();
```
Evangelink: The variable name is not really explicit, I had to look at the return type of the `FindTestAttribute` method to understand. Maybe something like `attributeKnownType`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldHaveCorrectSignature.cs
```diff
@@ -44,83 +43,116 @@ public sealed class TestMethodShouldHaveCorrectSignature : SonarDiagnosticAnalyz
...
+            if (type == null)
+            {
+                return NullValidator;
+            }
+
+            attributeToConstraintsMap.TryGetValue(type, out var validator);
```
Evangelink: Please use the extension method `GetValueOrDefault` instead.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldHaveCorrectSignature.cs
```diff
@@ -44,83 +43,116 @@ public sealed class TestMethodShouldHaveCorrectSignature : SonarDiagnosticAnalyz
...
             {
                 yield return MakeNonAsyncOrTaskMessage;
             }
         }
+
+        /// <summary>
+        /// Validation method. Checks the supplied method and returns the error message,
+        /// or null if there is no issue.
+        /// </summary>
+        private delegate string SignatureValidator(IMethodSymbol method);
```
Evangelink: Why using the old `delegate` instead of a `Func`?

duncanp-sonar: Using the delegate provides more information about what the function is for: seeing _Func<IMethodSymbol, string>_ in the code doesn't tell me anything about what the function is supposed to do. _SignatureValidator_ gives more of an idea.

Also, the code at the point of use is easier to read with the delegate
e.g.
```
var x = new Dictionary<KnownType, Func<IMethodSymbol, string>>();
// or
var x = new Dictionary<KnownType, SignatureValidator>();
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TestMethodShouldHaveCorrectSignature.cs
duncanp-sonar: I don't understand - what behaviour? Xunit doesn't have _DataTestMethod_.
The main purpose of this test is to check that we recognize the legacy XUnit attributes. I don't think there's any value in checking more than one old version, unless we know that there are multiple old versions of the API that we need to support.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldHaveCorrectSignature.cs
```diff
@@ -44,79 +43,111 @@ public sealed class TestMethodShouldHaveCorrectSignature : SonarDiagnosticAnalyz
...
-                },
-                SyntaxKind.ClassDeclaration);
+            var methodSymbol = c.SemanticModel.GetDeclaredSymbol(c.Node) as IMethodSymbol;
+            if (methodSymbol == null)
+            {
+                return;
+            }
+
+            var validator = GetValidator(methodSymbol);
+            var message = validator(methodSymbol);
```
Evangelink: This seems like the 2 previous statements could be merged. Ofc you will need to update the signature of `GetValidator`.

PR #1709, opened 2018-08-16
Improve time to run unit tests for rules

Evangelink:


duncanp-sonar:
What has changed to improve performance?

PR #1712, opened 2018-08-17
Fix S1226: should detect if param was read before being assigned

### LINKED ISSUE #1669: Fix S1226: rule doesn't detect correctly that param was read before being assigned
### Description

I have seen a couple of cases where the rule reports but shouldn't because the value was correctly read before being assigned a new value.

I suspect that the current implementation doesn't respect the definition and reports everytime a parameter (not `out` nor `ref`) is assigned.

[Example on Next](https://next.sonarqube.com/sonarqube/project/issues?id=sonaranalyzer-csharp-vbnet&issues=AWUhBkuhm8XJvoLlaEPX&open=AWUhBkuhm8XJvoLlaEPX)

andrei-epure-sonarsource:
Fix #1669


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParameterAssignedTo.cs
```diff
@@ -72,13 +65,68 @@ protected override bool IsAssignmentToParameter(ISymbol symbol)
         {
             var parameterSymbol = symbol as IParameterSymbol;
             var result = parameterSymbol?.RefKind == RefKind.None;
-
             return result;
         }
 
+        protected override bool IsReadBefore(SemanticModel semanticModel, ISymbol symbol, AssignmentExpressionSyntax assignment)
+        {
+            return GetPreviousStatements(assignment).Union(new[] { assignment.Right })
```
Evangelink: [cosmetic] Please move the `Union` statement to the next line. It's easier to read chained elements.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParameterAssignedTo.cs
```diff
@@ -72,13 +65,68 @@ protected override bool IsAssignmentToParameter(ISymbol symbol)
...
+            public override void VisitIdentifierName(IdentifierNameSyntax node)
+            {
+                var nodeSymbol = semanticModel.GetSymbolInfo(node).Symbol;
+                IsRead |= (parameterSymbol == nodeSymbol && IsReadAccess(node));
+            }
+
+            private bool IsReadAccess(IdentifierNameSyntax node)
+            {
+                return !IsLeftSideOfAssignment(node);
+                bool IsLeftSideOfAssignment(SyntaxNode syntaxNode) =>
```
Evangelink: [cosmetic] Please add a blank line before this inner method declaration.

Evangelink: Actually you don't even need this sub-method.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/ParameterAssignedTo.cs
```diff
@@ -85,6 +85,12 @@ protected override bool IsAssignmentToParameter(ISymbol symbol)
 
         protected override SyntaxNode GetAssignedNode(AssignmentStatementSyntax assignment) => assignment.Left;
 
+        protected override bool IsReadBefore(SemanticModel semanticModel, ISymbol symbol, AssignmentStatementSyntax assignment)
+        {
+            // FIXME implement logic
```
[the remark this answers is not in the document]
Evangelink: Let's add a link to the ticket to fix this.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParameterAssignedTo.cs
```diff
@@ -72,13 +65,68 @@ protected override bool IsAssignmentToParameter(ISymbol symbol)
...
+                .OfType<StatementSyntax>()
+                .TakeWhile(x => x != statement)
+                .Reverse();
+
+            return statement.Parent is StatementSyntax parentStatement
+                ? previousStatements.Union(GetPreviousStatements(parentStatement))
+                : previousStatements;
+        }
+
+        private class ParameterAccessCollector : CSharpSyntaxWalker
```
valhristov: Wouldn't this be simpler without a CSharpSyntaxWalker? Just something like `node.Descendants<Identifier>.Any(IsRead)`?

valhristov: Actually this would have better to be on line 80...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParameterAssignedTo.cs
andrei-epure-sonarsource: Because the function is affirmative, but inside it has a negation, I should either add a small comment or use a named method. I prefer the latter.

Evangelink: @andrei-epure-sonarsource you can update the condition not to require the not.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParameterAssignedTo.cs
andrei-epure-sonarsource: That's ok, I've changed it to use just LINQ.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ParameterAssignedTo.cs
```diff
@@ -72,13 +64,51 @@ protected override bool IsAssignmentToParameter(ISymbol symbol)
...
+            return GetPreviousStatements(assignment)
+                .Union(new[] { assignment.Right })
+                .SelectMany(s => s.DescendantNodes(n => true))
+                .OfType<IdentifierNameSyntax>()
+                .Select(node =>
+                {
+                    var nodeSymbol = semanticModel.GetSymbolInfo(node).Symbol;
+                    return parameterSymbol.Equals(nodeSymbol) && IsReadAccess(node);
+                })
+                .Aggregate(false, (isRead, accumulator) => accumulator || isRead);
```
valhristov: Could this become:
```
.Any(node =>
{
    var symbol = semanticModelGetSymbolInfo(node).Symbol;
	return parameterSymbol.Equals(symbol) && IsReadAccess(node);
});
```

PR #1713, opened 2018-08-17
Fixed #1486 - exception for classes that contain assembly-level setup/cleanup methods

### LINKED ISSUE #1486: S2187 Does not allow a class with only assembly related attributes
### Description

When a class only contains test assembly initialization / cleanup code by using methods that are annotated with the [AssemblyInitialize](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.assemblyinitializeattribute?view=mstest-net-1.2.0) / [AssemblyCleanup ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.assemblycleanupattribute?view=mstest-net-1.2.0) attributes it is also required for that class to be annotated with the [TestClass](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.testclassattribute?view=mstest-net-1.2.0) attribute (otherwise the assembly initialize / cleanup methods are not executed by MSTest). However Sonar will report S2187 on such a class because it expects test methods to be present whenever the [TestClass](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.testclassattribute?view=mstest-net-1.2.0) attribute is used.

### Repro steps

Create a new C# file with the following code and run analyzes on it:

```

namespace Foo.Bar
{
    [TestClass]
    public static class AssemblyHooks
    {
        [AssemblyInitialize]
        public static void BeforeTestRun(TestContext context)
        {
            // setup resources before testing
        }

        [AssemblyCleanup]
        public static void AfterTestRun()
        {
            // dispose resources after testing
        }
    }
}
```

### Expected behavior

This scenario should not result in warning S2187.

### Actual behavior

The class gets warning S2187.

### Known workarounds

Move the assembly initialization / cleanup code to a class which contains test methods.

### Related information

* SonarC# version 6.8.2 (build 4717)
* Scanner for MSBuild version 4.0.2.892

duncanp-sonar:
Fixed #1486

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestClassShouldHaveTestMethod.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestClassShouldHaveTestMethod.cs
@@ -43,6 +43,7 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static readonly ISet<KnownType> HandledTestClassAttributes = new HashSet<KnownType>
         {
+            // xUnit does not have have attributes to identity test classes
             KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_TestClassAttribute,
             KnownType.NUnit_Framework_TestFixtureAttribute
         };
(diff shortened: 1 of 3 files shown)


duncanp-sonar:
I updated the RSPEC: https://jira.sonarsource.com/browse/RSPEC-3412

duncanp-sonar:
I think we might still have false positives in other cases for MSTest and NUnit, namely where a test class only contains test instance-level setup and clean-up methods. However, that hasn't been reported as an issue. I'll do some more investigation and create a separate ticket if necessary.

PR #1719, opened 2018-08-20
Update S2971: Rule should not only suggest to remove call to ToList o…

### LINKED ISSUE #1711: Update S2971: Rule should not only suggest to remove call to ToList or ToArray
### Description

Using ToList() in the middle of  a LINQ to Entities call chain query should be replaced with AsEnumerable() 
instead of just being dropped.

### Repro steps
1. Step A
Run analysis on a LINQ to Entities query

```
context.MyEntities.Where(x => x.Foo == 'bar').ToList().Select(e => new Dto 
{ 
   MyProperty = GetValue(e) 
})} 

public int GetValue(Foo bar){ ... }
```

2. Step B

### Expected behavior

Message from S2971 should be something similiar to "Replace ToList() with AsEnumerable()"

Since materializing the query to memory is necessary to do before running GetValue

### Actual behavior

Sonarqube reports 
"Drop 'ToList' from the middle of the call chain. (S2971)"

Dropping ToList from the call chain will cause a NotSupportedException (LINQ to Entities)

### Known workarounds

Please provide a description of any known workarounds.

### Related information

* SonarC# Version
* Visual Studio Version

Evangelink:
…r ToArray

Fix #1711

### DIFF
--- a/sonaranalyzer-dotnet/its/expected/akka.net/Akka-{5DEDDF90-37F0-48D3-A0B0-A5CBD8A7E377}-S2971.json
+++ b/sonaranalyzer-dotnet/its/expected/akka.net/Akka-{5DEDDF90-37F0-48D3-A0B0-A5CBD8A7E377}-S2971.json
@@ -2,7 +2,7 @@
 "issues":  [
 {
 "id":  "S2971",
-"message":  "Drop 'ToArray' from the middle of the call chain.",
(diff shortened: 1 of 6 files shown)


duncanp-sonar:
The change looks ok, but why is the PR is targetting _master_ instead of a feature branch?

PR #1720, opened 2018-08-20
Fix #1705: support recognized test attributes

### LINKED ISSUE #1705: Fix S1607: : handle all test method types for supported test frameworks
S1607 does not handle the following test method types:
* _DataTestMethod_ (MSTest)
* Nunit _Theory_

Also, it only handles the MSTest _Ignore_ attribute. It does not handle the "ignore" mechanism in NUnit (_NUnit.Framework.Ignore_).

Finally, the rule runs against xUnit v2 _Fact_ and _Theory_ test methods, even though the MSTest _Ignore_ attribute has no effect on xUnit tests.

Currently the rule has both false positives and false negatives e.g.
* FN - not recognizing _DataTestMethod_ as indicating a test method
* FN - not recognizing the NUnit _Ignore_ attribute
* FP - raising against MSTest _Ignore_ attributes used on _xUnit_ tests

duncanp-sonar:
Fixed #1705 

* add support for NUnit _Theory_ and MSTest _DataTestMethod_ tests
* exclude xUnit from rule to remove FPs

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
@@ -44,13 +44,19 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static readonly ISet<KnownType> TrackedTestMethodAttributes = new HashSet<KnownType>
         {
+            // xUnit has it's own "ignore" mechanism (by providing a (Skip = "reason") string in
+            // the attribute, so there is always an explanation for the test being skipped).
+            // However, it is possible to mix test frameworks i.e. have an xUnit [Fact] that has an 
+            // MSTest [Ignore]. This seems like a weird corner case, but the rule handles it.
             KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_TestMethodAttribute,
+            KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_DataTestMethodAttribute,
             KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_TestClassAttribute,
             KnownType.NUnit_Framework_TestAttribute,
             KnownType.NUnit_Framework_TestCaseAttribute,
             KnownType.NUnit_Framework_TestCaseSourceAttribute,
             KnownType.Xunit_FactAttribute,
-            KnownType.Xunit_TheoryAttribute
+            KnownType.Xunit_TheoryAttribute,
+            KnownType.LegacyXunit_TheoryAttribute
         };
 
         protected override void Initialize(SonarAnalysisContext context)
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
@@ -46,17 +46,13 @@ namespace SonarAnalyzer.Rules.CSharp
         {
             // xUnit has it's own "ignore" mechanism (by providing a (Skip = "reason") string in
             // the attribute, so there is always an explanation for the test being skipped).
-            // However, it is possible to mix test frameworks i.e. have an xUnit [Fact] that has an 
-            // MSTest [Ignore]. This seems like a weird corner case, but the rule handles it.
             KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_TestMethodAttribute,
             KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_DataTestMethodAttribute,
             KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_TestClassAttribute,
             KnownType.NUnit_Framework_TestAttribute,
             KnownType.NUnit_Framework_TestCaseAttribute,
             KnownType.NUnit_Framework_TestCaseSourceAttribute,
-            KnownType.Xunit_FactAttribute,
-            KnownType.Xunit_TheoryAttribute,
-            KnownType.LegacyXunit_TheoryAttribute
+            KnownType.NUnit_Framework_TheoryAttribute
         };
 
         protected override void Initialize(SonarAnalysisContext context)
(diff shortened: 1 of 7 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldNotBeIgnored.cs
```diff
@@ -44,13 +44,15 @@ public sealed class TestMethodShouldNotBeIgnored : SonarDiagnosticAnalyzer
...
+            // the attribute, so there is always an explanation for the test being skipped).
             KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_TestMethodAttribute,
+            KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_DataTestMethodAttribute,
             KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_TestClassAttribute,
             KnownType.NUnit_Framework_TestAttribute,
             KnownType.NUnit_Framework_TestCaseAttribute,
             KnownType.NUnit_Framework_TestCaseSourceAttribute,
-            KnownType.Xunit_FactAttribute,
-            KnownType.Xunit_TheoryAttribute
+            KnownType.NUnit_Framework_TheoryAttribute
```
duncanp-sonar: Added support for two missing attributes types.
Removed xUnit attributes from the scope of the rule; it isn't possible to ignore an xUnit test without providing a reason, and applying MSTest/NUnit _Ignore_ attributes against an xUnit test does nothing.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldNotBeIgnored.MsTest.cs
```diff
@@ -37,24 +37,63 @@ public void Foo5()
         }
     }
 
+    [TestClass]
+    public class MsTestClass_DataTestMethods
```
duncanp-sonar: Duplicated the existing test cases for the _DataTestMethod_ attribute

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldNotBeIgnored.MsTest.cs
```diff
@@ -1,9 +1,9 @@
-﻿using System;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
 
-namespace Tests.Diagnostics
+namespace Tests.Diagnostics.TestMethods
 {
-    class MsTestClass
+    [TestClass]
```
duncanp-sonar: I modified the existing test code so that the test code is valid, so it can be cut-and-paste into a test project and executed to check that it is working as expected.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldNotBeIgnored.NUnit.cs
```diff
@@ -1,11 +1,14 @@
...
+using MSTest = Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Tests.Diagnostics
 {
-    class NUnitClass
+    [TestFixture]
+    public class NUnitClass
     {
+        // False positive: using an MSTest Ignore against an NUnit test does nothing,
+        // but we still raise an issue.
```
duncanp-sonar: It's unlikely anyone would do this in practice so I don't think it's worth worrying about.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldNotBeIgnored.NUnit.cs
```diff
@@ -14,51 +17,95 @@ public void Foo1()
         }
 
         [Test]
-        [Microsoft.VisualStudio.TestTools.UnitTesting.Ignore] // This test is ignored because 'blah blah'
+        [NUnit.Framework.Ignore] // This test is ignored because 'blah blah'
```
duncanp-sonar: Previously all of these tests were using the MSTest _Ignore_ attribute, even though it does not work with NUnit tests, and NUnit has its own _Ignore_ attribute with the same semantics.
I've changed the remaining tests to use the NUnit attribute.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldNotBeIgnored.NUnit.cs
```diff
@@ -14,51 +17,95 @@ public void Foo1()
...
-            new object[] { 12, 3, 4 },
-            new object[] { 12, 2, 6 },
-            new object[] { 12, 4, 3 }
+            new object[] { 12, 3, 4 }
         };
+
+        [Datapoint]
+        public int Data = 42;
+
+        [Theory]
```
duncanp-sonar: Added tests for the _Theory_ attribute since that is now handled.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldNotBeIgnored.NUnit.cs
```diff
@@ -14,51 +17,95 @@ public void Foo1()
...
+        [Ignore("a reason")]
+        public void Theory3(int arg)
+        {
+        }
+    }
+
+    [TestFixture, Ignore]
+    public class IgnoredClass1
```
duncanp-sonar: Added tests for class-level NUnit _Ignore_ (same as corresponding MSTest class-level tests).

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TestMethodShouldNotBeIgnoredTest.cs
```diff
@@ -66,5 +66,17 @@ public void TestMethodShouldNotBeIgnored_Xunit(string testFwkVersion)
...
                     .ToArray());
         }
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void TestMethodShouldNotBeIgnored_Xunit_v1()
+        {
+            Verifier.VerifyAnalyzer(@"TestCases\TestMethodShouldNotBeIgnored.Xunit.v1.cs",
+                new TestMethodShouldNotBeIgnored(),
+                additionalReferences: MetadataReferenceHelper.FromNuGet("MSTest.TestFramework", "1.1.11")
```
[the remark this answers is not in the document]
duncanp-sonar: I'll rebase _branch-testDetection_ when I've finished all of the outstanding tickets.

PR #1721, opened 2018-08-20
Update S3237: Rule should handle expression body accessors

### LINKED ISSUE #1021: Update S3237: Rule should handle expression body accessors
### Description

The rule currently won't report on properties using the new syntax (expression body accessor).

### Repro steps

```
namespace ConsoleApp8
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello, world!");
        }

        public Program(string id)
        {
            if (string.IsNullOrEmpty(id)) throw new ArgumentNullException(nameof(id));
        }

        private string id;

        public string Id
        {
            get => null;
            set => id = 10; // Rule won't raise if the value is not used here
        }
    }
}
```

### Expected behavior

The issue should be raised.

### Actual behavior

The issue is not raised.

Evangelink:
Fix #1021

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UseValueParameter.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UseValueParameter.cs
@@ -19,12 +19,14 @@
  */
 
 using System.Collections.Immutable;
+using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.ShimLayer.CSharp;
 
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UseValueParameter.cs
```diff
@@ -37,61 +39,51 @@ public sealed class UseValueParameter : SonarDiagnosticAnalyzer
...
                     {
                         return;
                     }
 
-                    var interfaceMember = cbc.SemanticModel.GetDeclaredSymbol(accessorDeclaration)
-                        .GetInterfaceMember();
+                    var interfaceMember = c.SemanticModel.GetDeclaredSymbol(accessor).GetInterfaceMember();
                     if (interfaceMember != null &&
-                        accessorDeclaration.Body.Statements.Count == 0)
+                        accessor?.Body.Statements.Count == 0) // No need to check ExpressionBody, it can't be empty
```
valhristov: Are we going to raise on incomplete syntax? E.g. in VS, while typing the code:
```
public int Foo
{
    get => _foo;
    set => 
}
// or
public int Bar
{
	get => _bar;
	set
}
```

PR #1731, opened 2018-08-20
Add test to show that S3251 handles ExpressionBody

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PartialMethodNoImplementation.cs
```diff
@@ -57,8 +57,10 @@ private static void CheckForCandidatePartialDeclaration(SyntaxNodeAnalysisContex
         {
             var declaration = (MethodDeclarationSyntax)context.Node;
             var partialKeyword = declaration.Modifiers.FirstOrDefault(m => m.IsKind(SyntaxKind.PartialKeyword));
-            if (declaration.Body != null ||
-                partialKeyword == default(SyntaxToken))
+
+            if (partialKeyword == default(SyntaxToken)||
+                declaration.Body != null ||
+                declaration.ExpressionBody != null)
```
Evangelink: It was working without the new check but I think that's better to do early return whenever possible.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PartialMethodNoImplementation.cs
```diff
@@ -74,29 +76,32 @@ private static void CheckForCandidatePartialDeclaration(SyntaxNodeAnalysisContex
         private static void CheckForCandidatePartialInvocation(SyntaxNodeAnalysisContext context)
         {
             var invocation = (InvocationExpressionSyntax)context.Node;
+
+            if (!(invocation.Parent is StatementSyntax statement))
```
Evangelink: Move the syntax check before checking for symbol so that we do as much as possible on the non-costy side.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PartialMethodNoImplementation.cs
```diff
@@ -74,29 +76,32 @@ private static void CheckForCandidatePartialDeclaration(SyntaxNodeAnalysisContex
...
             // from the method symbol it's not possible to tell if it's a partial method or not.
             // https://github.com/dotnet/roslyn/issues/48
-
             var partialDeclarations = methodSymbol.DeclaringSyntaxReferences
                 .Select(r => r.GetSyntax())
                 .OfType<MethodDeclarationSyntax>()
-                .Where(method => method.Body == null && method.Modifiers.Any(SyntaxKind.PartialKeyword));
+                .Where(method => method.Modifiers.Any(SyntaxKind.PartialKeyword) &&
+                    method.Body == null &&
+                    method.ExpressionBody == null);
```
Evangelink: It was working without the new check but I think that's better to do early return whenever possible.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PartialMethodNoImplementation.cs
```diff
@@ -74,29 +76,32 @@ private static void CheckForCandidatePartialDeclaration(SyntaxNodeAnalysisContex
...
                 !partialDeclarations.Any())
             {
                 return;
             }
 
-            if (!(invocation.Parent is StatementSyntax statement))
-            {
-                return;
-            }
             context.ReportDiagnosticWhenActive(Diagnostic.Create(rule, statement.GetLocation(), "the", MessageAdditional));
```
valhristov: That's weird use of string format (see the definition of MessageAdditional)

valhristov: Yeah, I know, I'm just mentioning it.

PR #1732, opened 2018-08-20
Small refactoring for S4200

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NativeMethodsShouldBeWrapped.cs
```diff
@@ -45,33 +45,31 @@ public sealed class NativeMethodsShouldBeWrapped : SonarDiagnosticAnalyzer
 
         protected override void Initialize(SonarAnalysisContext context)
         {
-            context.RegisterSyntaxNodeAction(ReportPublicExternalMethods, SyntaxKind.MethodDeclaration);
+            context.RegisterSymbolAction(ReportPublicExternalMethods, SymbolKind.Method);
```
Evangelink: We were only doing checks on symbol so I changed the type of the registration.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NativeMethodsShouldBeWrapped.cs
```diff
@@ -45,33 +45,31 @@ public sealed class NativeMethodsShouldBeWrapped : SonarDiagnosticAnalyzer
...
+                    c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, methodDeclaration.Identifier.GetLocation(),
+                        MakeThisMethodPrivateMessage));
+                }
             }
         }
 
         private static void ReportTrivialWrappers(SyntaxNodeAnalysisContext c)
         {
             var methodDeclaration = (MethodDeclarationSyntax)c.Node;
-            var methodSymbol = c.SemanticModel.GetDeclaredSymbol(methodDeclaration);
```
Evangelink: Move symbol checks later in the code to ensure we do as many tests as possible on the syntax before getting the symbol.

PR #1736, opened 2018-08-20
Fixed #1498: centralise test detection code

### LINKED ISSUE #1498: Test method detection code is not consistent across rules
A number of rules contain checks to decide whether a method is a test method or not (either to exclude or include it).

However, the rules aren't consistent in how they decide e.g. _TestClassShouldHaveTestMethod_ doesn't handle _XUnit_, but _TestMethodShouldContainAssertion_ does.

Centralising the detection code will also make it easier to update when e.g. the namespace of the _XUnit_ attributes changes.

The following rules are test-only rules:
* [S1607](https://jira.sonarsource.com/browse/RSPEC-1607) Tests should not be ignored
* [S2187](https://jira.sonarsource.com/browse/RSPEC-2187) Test cases should contain tests
* [S2699](https://jira.sonarsource.com/browse/RSPEC-2699) Tests should include assertions
* [S2701](https://jira.sonarsource.com/browse/RSPEC-2701) Literal boolean values should not be used in assertions
* [S3415](https://jira.sonarsource.com/browse/RSPEC-3415) Assertion arguments should be passed in the correct order
* [S3431](https://jira.sonarsource.com/browse/RSPEC-3431) [ExpectedException] should not be used
* [S3433](https://jira.sonarsource.com/browse/RSPEC-3433) Test method signatures should be correct

duncanp-sonar:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ExpectedExceptionAttributeShouldNotBeUsed.cs
```diff
@@ -41,12 +40,6 @@ public sealed class ExpectedExceptionAttributeShouldNotBeUsed : SonarDiagnosticA
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
 
-        private static readonly ISet<KnownType> HandledExpectedExceptionAttributes = new HashSet<KnownType>
```
duncanp-sonar: Now using the definition in _UnitTestHelper_.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestClassShouldHaveTestMethod.cs
```diff
@@ -49,14 +49,10 @@ public sealed class TestClassShouldHaveTestMethod : SonarDiagnosticAnalyzer
...
-            KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_DataTestMethodAttribute,
-            KnownType.NUnit_Framework_TestAttribute,
-            KnownType.NUnit_Framework_TestCaseAttribute,
-            KnownType.NUnit_Framework_TestCaseSourceAttribute,
-            KnownType.NUnit_Framework_TheoryAttribute
-        };
+        (
+            UnitTestHelper.KnownTestMethodAttributes_MSTest
+            .Concat(UnitTestHelper.KnownTestMethodAttributes_NUnit)
+        );
```
duncanp-sonar: Can't use _UnitTestHelper.IsTestMethod_ here since rule doesn't apply to _xUnit_.
I considered adding separate helper methods for the various test frameworks, but it didn't seem worth it. Instead, I've exposed the commonly-used groups of attributes in the _UnitTestHelper_ so they only need to be updated in one place.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SymbolHelper.cs
```diff
@@ -49,7 +49,7 @@ public static IEnumerable<INamedTypeSymbol> GetAllNamedTypes(this INamespaceSymb
             symbol?.GetAttributes().Where(a => a.AttributeClass.Is(attributeType))
                 ?? Enumerable.Empty<AttributeData>();
 
-        internal static IEnumerable<AttributeData> GetAttributes(this ISymbol symbol, ISet<KnownType> attributeTypes) =>
+        internal static IEnumerable<AttributeData> GetAttributes(this ISymbol symbol, IEnumerable<KnownType> attributeTypes) =>
```
duncanp-sonar: Changing the signature to use _IEnumerable_ so the method can be called from some of the methods in UnitTestHelper_ that don't use _ISet_.
Same for the method in _TypeHelper_.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
```diff
@@ -29,34 +29,49 @@ namespace SonarAnalyzer.Helpers
 {
     internal static class UnitTestHelper
     {
-        private static readonly ISet<KnownType> KnownTestMethodAttributes = new HashSet<KnownType>
+        public static readonly IEnumerable<KnownType> KnownTestMethodAttributes_MSTest = new HashSet<KnownType>
```
duncanp-sonar: Changed to _IEnumerable_ rather than _ISet_ since the field is now publicly visible (otherwise we get "use an immutable type" issues).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestClassShouldHaveTestMethod.cs
valhristov: The rule does not apply to XUnit because it does not have equivalent to the [TestClass] attribute and all XUnit test classes already have tests by definition. Imo it does not hurt if we use the general IsTestMethod even though a few of the checked attributes will be redundant...

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
```diff
@@ -29,34 +29,49 @@ namespace SonarAnalyzer.Helpers
...
-        private static readonly ISet<KnownType> KnownIgnoreAttributes = new HashSet<KnownType>
+        public static readonly IEnumerable<KnownType> KnownIgnoreAttributes = new HashSet<KnownType>
         {
             // Note: XUnit doesn't have a separate "Ignore" attribute. It has a "Skip" parameter
             // on the test attribute
             KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_IgnoreAttribute,
             KnownType.NUnit_Framework_IgnoreAttribute
         };
 
+        private static readonly ISet<KnownType> KnownTestMethodAttributes = new HashSet<KnownType>
```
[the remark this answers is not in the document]
duncanp-sonar: The field is private so the "immutable field" rule didn't raise an issue so I didn't change it. However, I'll make them all the same.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestClassShouldHaveTestMethod.cs
duncanp-sonar: Ok. I'll move the "IsTestClass" code into _UnitTestHelper_ too.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
```diff
@@ -85,7 +85,21 @@ internal static class UnitTestHelper
                 "VERIFY",
                 "VALIDATE"
             };
-        
+
+        private static readonly IEnumerable<KnownType> KnownTestClassAttributes = new HashSet<KnownType>
+        {
+            // xUnit does not have have attributes to identity test classes
```
valhristov: typo on `identity` should be `identify`, but don't bother changing :)

PR #1744, opened 2018-08-20
Update S3881: Rule should handle ExpressionBody

### LINKED ISSUE #1728: Update S3881: Rule should handle ExpressionBody
Relates to #1714

valhristov:
Fix #1728


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
```diff
@@ -286,6 +287,10 @@ private bool HasVirtualDisposeBool(INamedTypeSymbol typeSymbol)
                     .WhereNotNull()
                     .Any(methodSym => !methodSym.IsAbstract);
             }
+
+            private static bool HasBodyOrExpressionBody(BaseMethodDeclarationSyntax methodDeclaration) =>
+                methodDeclaration != null &&
```
Evangelink: It's ok like this but you could have use null-conditional on both path and forget about the first AND

PR #1745, opened 2018-08-20
Merge test fixes to master

duncanp-sonar:
All of the fixes have been reviewed individually.

The merge fix-ups are in the tests around the provisioning of NuGet packages. No further product changes.


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/NuGetMetadataReference.cs
```diff
@@ -45,6 +45,7 @@ internal static class NuGetMetadataReference
             {
                 "lib",
                 "portable-net45",
+                "net20",
```
duncanp-sonar: xUnit v1 targets net20.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/NuGetMetadataReference.cs
```diff
@@ -174,6 +181,11 @@ private static MetadataReference[] CreateReferences(string packageId, string pac
         internal static Dictionary<string, MetadataReference[]> XunitExtensibilityCore =>
             lazyXunitExtensibilityCore.Value;
 
+        internal static MetadataReference[] XunitV1Packages =>
+            lazyXunitV1.Value.Values.First()
+            .Concat(lazyXunitExtensionsV1.Value.Values.First())
+            .ToArray();
+
```
duncanp-sonar: There are now a few tests that specifically check the legacy (v1) _xUnit_ framework. There's no particular reason to test multiple versions of the legacy framework, and there's no need reason for the individual tests to have to know which version of _xUnit_ is installed, so the property exposes an array of references rather than a dictionary.

PR #1757, opened 2018-08-21
Update S2190: Rule should handle ExpressionBody

### LINKED ISSUE #1755: Update S2190: Rule should handle ExpressionBody
Relates to #1714

Evangelink:
Fix #1755


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.cs
```diff
@@ -105,6 +109,20 @@ int Prop2
...
+        int Prop5 => Prop5; // Noncompliant
+
+        int Prop6
+        {
+            get => 42;
+        }
+
+        int Prop6
+        {
+            get => Prop6; // Compliant - FP
```
valhristov: please use `False Positive` instead of `FP`

PR #1779, opened 2018-08-21
Rule S3937: Number patterns should be regular

### LINKED ISSUE #249: Rule S3937: Number patterns should be regular
Implements [RSPEC-3937](https://jira.sonarsource.com/browse/RSPEC-3937)

Evangelink:
Fix #249

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
@@ -1,0 +1,92 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2018 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -0,0 +1,92 @@
...
+        private static readonly DiagnosticDescriptor rule =
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
+                {
+                    if ((c.Compilation as CSharpCompilation)?.LanguageVersion.CompareTo(LanguageVersionEx.CSharp7) < 0)
```
valhristov: This looks like a good candidate for an extension method(s)
```
if (c.LanguageIs(LanguageVersionEx.CSharp7)) // or something better
```
Perhaps in another PR, though.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NumberPatternShouldBeRegular.cs
```diff
@@ -0,0 +1,30 @@
...
+    class Program
+    {
+        private int field = 1_000;
+
+        public void CompliantCases()
+        {
+            var thousand = 1000;
+
+            // decimal notation
+            var balance = 2_435_951.68;
```
valhristov: You could have `_` after the decimal point, like `111_2.22_333` - what should be the correct behavior in such cases? I feel that we should threat the two sides around the decimal point as different numbers, but still have the same group length... Please, add some tests.

PR #1783, opened 2018-08-23
Follow IDE styling recommendations

Evangelink:


.editorconfig
```diff
@@ -64,7 +64,7 @@ indent_size = 2
 dotnet_sort_system_directives_first = true
 
 # Avoid "this." and "Me." if not necessary
-dotnet_style_qualification_for_field = false:suggestion
```
valhristov: I don't like this setting, everything else is good

PR #1788, opened 2018-08-23
Pattern matching support for conditions and switch statements [NO SQUASH]

### LINKED ISSUE #157: Add support for pattern matching in conditions in CFG
Code like this should be correctly analyzed

```
public void Patterns(object o)
{
    if (o is null)
    {
        return;
    }

    if (o is int i)
    {
        Console.WriteLine(i);
    }
}
```

Apart from if statements, the other language elements that should support pattern matching are:
ternary operators, for loops, while loops, do loops

Switch statements are handled in the referenced ticket #1775

### LINKED ISSUE #1775: Add support for switch statements pattern matching in CFG
### Description

This is split from #157 and is related only to pattern matching in `switch` statements:

```
switch (o)
{
    case int n1:
        Console.Write(n1);
        break;
    case string s when s.Length > 0:
        Console.WriteLine(s);
        break;
    case null:
        Console.WriteLine("null");
        break;
}
```

valhristov:
Fix #157 
Fix #1775

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -25,6 +25,7 @@ using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.ShimLayer.CSharp;
 
 namespace SonarAnalyzer.ControlFlowGraph.CSharp
 {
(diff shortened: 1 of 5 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -612,7 +642,12 @@ private Block BuildGotoStatement(GotoStatementSyntax statement, Block currentBlo
 
         #region Build switch
 
-        private Block BuildSwitchStatement(SwitchStatementSyntax switchStatement, Block currentBlock)
+        private Block BuildSwitchStatement(SwitchStatementSyntax switchStatement, Block currentBlock) =>
+            switchStatement.Sections.All(s => s.Labels.All(l => l.IsAnyKind(csharp6SwitchLabelKinds)))
+                ? BuildSwitchStatement_Legacy(switchStatement, currentBlock)
+                : BuildSwitchStatement_CSharp7(switchStatement, currentBlock);
```
valhristov: The pattern-enabled switch statements now takes into account the order of the case section declarations (the non-pattern switch doesn't), hence two completely different ways to handle.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -442,6 +448,30 @@ private Block BuildExpression(ExpressionSyntax expression, Block currentBlock)
                             parent.ArgumentList?.Arguments.Select(a => a.Expression));
                     }
 
+                case SyntaxKindEx.IsPatternExpression:
+                    var isPatternExpression = (IsPatternExpressionSyntaxWrapper)expression;
```
Evangelink: Would be nice to have some examples of the different cases.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp7.cs
```diff
@@ -2,19 +2,103 @@
...
+                    s.ToString();
+                }
+            }
+        }
 
-        public bool Prop => TRUE || myBool; // Noncompliant
-                                            // Secondary@-1
+        void VariableDesignationPattern_Source(object o)
+        {
+            // We can set NotNull constraint only for one of the variables in the if condition
```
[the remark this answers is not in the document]
valhristov: To specify a relationship between two variables (e.g. `a > b` or `a != null`) we have a special SymbolicValue. To specify relationship between three variables (e.g. `a != null and b != null) we would need a new SV. Creating a new SV is difficult and not absolutely required to enable the new syntax. Perhaps there is another way, but I could not find it so far and I am open for ideas.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
```diff
@@ -450,6 +451,33 @@ protected override void VisitInstruction(ExplodedGraphNode node)
...
+                    {
+                        // "x is string s"
+                        // VisitDeclarationPattern() expects SV_s on top of the stack, hence we pop SV_x
+                        newProgramState = newProgramState.PopValue();
+
+                        newProgramState = VisitDeclarationPattern((DeclarationPatternSyntaxWrapper)isPatternExpression.Pattern, newProgramState);
+                    }
+                    else
+                    {
+                        goto default;
```
Evangelink: For a one line throw you can duplicate the code.

PR #1790, opened 2018-08-23
Fix how module levels issues are reported

### LINKED ISSUE #1789: Module levels issues are not reported correctly
### Description

It seems that when we updated the sensors to be global we broke the way the module level issues are reported.

Evangelink:
Fix #1789

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractPropertiesSensor.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractPropertiesSensor.java
@@ -21,6 +21,7 @@ package org.sonarsource.dotnet.shared.plugins;
 
 import java.nio.file.Path;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.sonar.api.batch.ScannerSide;
 import org.sonar.api.batch.sensor.Sensor;
 import org.sonar.api.batch.sensor.SensorContext;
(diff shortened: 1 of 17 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/SarifParserCallbackImpl.java
```diff
@@ -44,16 +48,17 @@ public SarifParserCallbackImpl(SensorContext context, Map<String, String> reposi
...
     if (repositoryKey == null) {
       return;
     }
+
     NewIssue newIssue = context.newIssue();
     newIssue
       .forRule(RuleKey.of(repositoryKey, ruleId))
       .at(newIssue.newLocation()
-        .on(context.module())
+        .on(inputModule)
```
Evangelink: Report on the right module.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractPropertiesSensor.java
```diff
@@ -55,7 +56,7 @@ public void execute(SensorContext context) {
 
     List<Path> roslynReportPaths = configuration.roslynReportPaths();
     if (!roslynReportPaths.isEmpty()) {
-      reportPathCollector.addRoslynDirs(roslynReportPaths);
+      reportPathCollector.addRoslynDirs(roslynReportPaths.stream().map(path -> new RoslynReport(context.module(), path)).collect(Collectors.toList()));
```
Evangelink: Collect the reports and their module.

PR #1795, opened 2018-08-27
Fix S3246: False positive on tuples

### LINKED ISSUE #1513: Fix S3246: False positive on tuples
### Description

S3246 suggests making a generic type parameter T contravariant (`<in T>`) when T is only used as a parameter type in the interface.

It fails to recognize when T is also returned as part of a tuple, so it generates a suggestion to modify the code to a state where it won't compile.

### Repro steps
Create an interface like below, using T as a parameter type and having the method return a tuple of T and anything else.

    public interface IFoo<T>
    {
        (T, object) Process(T item);
    }

### Expected behavior

S3246 should **not** trigger a suggestion to modify `IFoo<T>` to `IFoo<in T>`.

### Actual behavior

It does actually suggest to add `in` to make T contravariant
After a quick look in the repo, it seems return types are only considered as a whole, which would explain why using T in a tuple is not recognized correctly.

### Related information

SonarScanner for MSBuild 4.3.1, which runs as a build step in VSTS

valhristov:
Fix #1513

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterInOut.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterInOut.cs
@@ -103,8 +103,8 @@ namespace SonarAnalyzer.Rules.CSharp
             foreach (var typeParameter in declaredSymbol.TypeParameters
                 .Where(typeParameter => typeParameter.Variance == VarianceKind.None))
             {
-                var canBeIn = CheckTypeParameter(typeParameter, VarianceKind.In, declaredSymbol, returnType, parameterSymbols);
-                var canBeOut = CheckTypeParameter(typeParameter, VarianceKind.Out, declaredSymbol, returnType, parameterSymbols);
+                var canBeIn = CheckTypeParameter(typeParameter, VarianceKind.In, returnType, parameterSymbols);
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterInOut.cs
```diff
@@ -126,14 +127,14 @@ private static void CheckDelegateVariance(DelegateDeclarationSyntax declaration,
             }
 
             canBe = CanTypeParameterBeVariant(typeParameter, variance, returnType,
-                true, false, delegateType);
```
valhristov: This type parameter was only passed, but never used

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterInOut.cs
```diff
@@ -341,6 +336,11 @@ private static void ReportIssue(ITypeParameterSymbol typeParameter, VarianceKind
                     return true;
             }
 
+            if (namedType.IsTupleType())
```
valhristov: This is the actual fix - when a generic argument is used in a tuple, that argument cannot be neither `in` or `out`

PR #1802, opened 2018-08-28
The case null section handling in switch with patterns is wrong; fixe…

### LINKED ISSUE #1801: Create BrancBlock for "case null" sections to avoid exceptions in the exploded graph
`case null` should generate BranchBlock instead of BinaryBranchBlock when inside a `switch` block with patterns.

The currently created BinaryBranchBlock causes an exception to be thrown when it is visited by the ExplodedGraph. The exception is thrown because the BinaryBranchBlock does not contain an instruction for the branch condition. Such branch condition would be difficult to be created because we cannot easily put more complex expressions in the CFG:
```
switch (a ?? b)
{
	case null:
		break;
}
```

Using a simple `BranchBlock` will generate some false negatives, but will work without exceptions and will be generally consistent with the other pattern sections (they don't apply constraints on the switch expression, but only on the declared variable).

Example False Negative for the condition evaluates to constant rule:

```
switch (o)
{
	case null:
		if (o == null) // False Negative, we don't know anything for "o" at this point
		{
		}
		break;
}
```
```
switch (o)
{
	case null:
		return o.ToString(); // False Negative, we don't know anything for "o" at this point
}
```

valhristov:
…d default section detection;

Fix #1801

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -749,8 +749,14 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
                         caseSwitchLabel.Value != null &&
                         caseSwitchLabel.Value.IsKind(SyntaxKind.NullLiteralExpression))
                     {
-                        currentSectionBlock = CreateBinaryBranchBlock(caseSwitchLabel,
-                            trueSuccessor: sectionBlock, falseSuccessor: currentSectionBlock);
+                        // Creating BranchBlock instead of BinaryBranchBlock will generate some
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -749,8 +749,8 @@ private Block BuildSwitchStatement_CSharp7(SwitchStatementSyntax switchStatement
                         caseSwitchLabel.Value != null &&
                         caseSwitchLabel.Value.IsKind(SyntaxKind.NullLiteralExpression))
                     {
-                        currentSectionBlock = CreateBinaryBranchBlock(caseSwitchLabel,
-                            trueSuccessor: sectionBlock, falseSuccessor: currentSectionBlock);
+                        currentSectionBlock = CreateBranchBlock(caseSwitchLabel,
```
Evangelink: Please add a comment with what you explained to me.

PR #1806, opened 2018-08-29
Improve debug logging when importing code coverage and test coverage

### LINKED ISSUE #1780: Improve debug logging when importing code coverage and test coverage
### Description

The C# plugin does not provide detailed logs when importing code or test coverage, even when the logging level is set to _debug_. This make is difficult to diagnose why results are not being uploaded.

### Repro steps

Example of current output where results were not uploaded with _debug_ logging turned on:

```
…
22:05:01.93 22:05:01.93 sonar.verbose=true was specified - setting the log verbosity to ‘Debug’
…
22:24:28.615 INFO: Sensor C# Tests Coverage Report Import [csharp]
22:24:28.619 INFO: Sensor C# Tests Coverage Report Import [csharp] (done) | time=4ms
```

### Expected behavior

Debug level logging should provide more information e.g.
* if coverage data correspond to the sources that have been analyzed
* if there is skipped data and why
* if the report is empty

### Actual behavior

No additional diagnostic information is logged.

### Known workarounds

None

### Related information

* SonarC# Version: all

Evangelink:
Fix #1780

### DIFF
--- a/sonar-dotnet-tests-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResults.java
+++ b/sonar-dotnet-tests-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResults.java
@@ -25,15 +25,13 @@ import javax.annotation.Nullable;
 public class UnitTestResults {
 
   private int tests;
-  private int passed;
   private int skipped;
   private int failures;
   private int errors;
   private Long executionTime;
 
-  public void add(int tests, int passed, int skipped, int failures, int errors, @Nullable Long executionTime) {
+  public void add(int tests, int skipped, int failures, int errors, @Nullable Long executionTime) {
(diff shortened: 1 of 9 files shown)


sonar-dotnet-tests-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResults.java
```diff
@@ -50,10 +48,6 @@ public int tests() {
     return tests;
   }
 
-  double passedPercentage() {
```
valhristov: Is this related to the current ticket?

Evangelink: It wasn't used so I removed it at the same time.

PR #1813, opened 2018-08-31
[WIP] Converting ProgramState to struct

valhristov:


valhristov:
Out of memory still there...

PR #1816, opened 2018-08-31
SONARVB-319: Fix the VBC error happening when the line on which an is…

Evangelink:
…sue is raised contains the error pattern.


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/VbcHelper.cs
```diff
@@ -43,16 +43,16 @@ public static class VbcHelper
         /// </returns>
         public static bool IsTriggeringVbcError(Diagnostic diagnostic)
         {
-            var rootNode = diagnostic.Location?.SourceTree?.GetRoot();
-
-            if (rootNode != null &&
-                rootNode.Language == LanguageNames.VisualBasic &&
-                IsTextMatchingVbcErrorPattern(rootNode.FindNode(diagnostic.Location.SourceSpan)?.ToString()))
```
Evangelink: This was just finding the node on which we were reporting the issue but that doesn't mean this is the node containing the error pattern... We need to look at the full line.

PR #1817, opened 2018-08-31
Update the plugin to not fail for SonarLint on SonarCloud

### LINKED ISSUE #1824: SonarC# NPE with SonarLint for VS connected mode
### Description

The way we are currently building the ruleset for the connected mode of SonarLint for Visual Studio doesn't always work for SonarCloud, resulting in a NullPointerException being raised.

Example of URI call:
`http://sonarcloud.io/api/qualityprofiles/export?exporterKey=sonarlint-vs-cs&language=cs&qualityProfile=Sonar%20way&organization=my-org`

Evangelink:
Fix #1824 

We need to make test to ensure I didn't break anything for SonarLint + let's do one quick test to check Scanner for MSBuild (it seems to be using a different class but let's be sure).

### DIFF
--- a/sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/SonarLintProfileExporter.java
+++ b/sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/SonarLintProfileExporter.java
@@ -19,13 +19,14 @@
  */
 package org.sonar.plugins.csharp;
 
+import org.sonar.api.rules.RuleFinder;
 import org.sonarsource.dotnet.shared.plugins.AbstractSonarLintProfileExporter;
 
 public class SonarLintProfileExporter extends AbstractSonarLintProfileExporter {
   private static final String PROFILE_KEY = "sonarlint-vs-cs";
   private static final String PROFILE_NAME = "SonarLint for Visual Studio Rule Set";
 
-  public SonarLintProfileExporter(CSharpSonarRulesDefinition csharpRulesDefinition) {
-    super(csharpRulesDefinition, PROFILE_KEY, PROFILE_NAME, CSharpPlugin.LANGUAGE_KEY, CSharpPlugin.SONARANALYZER_NAME, CSharpPlugin.REPOSITORY_KEY);
+  public SonarLintProfileExporter(RuleFinder ruleFinder) {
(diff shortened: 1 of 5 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractRulesDefinition.java
```diff
@@ -92,21 +91,13 @@ public void define(Context context) {
...
     for (NewRule rule : repository.rules()) {
       updateMetadata(rule);
-      allRuleKeys.add(rule.key());
     }
 
-    allRuleKeys = Collections.unmodifiableSet(allRuleKeys);
     repository.done();
   }
 
-  public Set<String> allRuleKeys() {
```
Evangelink: `define` wasn't always called before. Based on the recommendations from JulienH I have changed the code to user the `RuleFinder`.

sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/SonarLintProfileExporter.java
```diff
@@ -19,13 +19,14 @@
  */
 package org.sonar.plugins.csharp;
 
+import org.sonar.api.rules.RuleFinder;
```
duncanp-sonar: It looks like this API has been deprecated since 5.1 : http://javadocs.sonarsource.org/5.3/apidocs/org/sonar/api/rules/RuleFinder.html

Do we have a long-term plan to replace this with a supported API, or a ticket to track that further work is required?

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractSonarLintProfileExporter.java
```diff
@@ -23,29 +23,31 @@
...
-    this.rulesDefinition = rulesDefinition;
     this.analyzerName = analyzerName;
     this.repositoryKey = repositoryKey;
   }
 
   @Override
   public void exportProfile(RulesProfile ruleProfile, Writer writer) {
     Set<String> disabledRuleKeys = new LinkedHashSet<>();
-    disabledRuleKeys.addAll(rulesDefinition.allRuleKeys());
+    disabledRuleKeys.addAll(ruleFinder.findAll(RuleQuery.create().withRepositoryKey(repositoryKey)).stream().map(Rule::getKey).collect(Collectors.toList()));
```
duncanp-sonar: As discussed, this would be easier to understand if it was in a separate method e.g. _GetAllRuleKeys_

PR #1823, opened 2018-09-03
Make SonarVB open source

valhristov:


sonar-vbnet-plugin/src/test/resources/Program.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
```
Evangelink: This should be a VB.Net class.

sonar-vbnet-plugin/src/main/java/org/sonar/plugins/vbnet/VbNetSonarWayProfile.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @Override
+  public void define(Context context) {
+    NewBuiltInQualityProfile sonarWay = context.createBuiltInQualityProfile("Sonar way", VbNetPlugin.LANGUAGE_KEY);
+    BuiltInQualityProfileJsonLoader.load(sonarWay, VbNetPlugin.REPOSITORY_KEY, "org/sonar/plugins/vbnet/Sonar_way_profile.json");
+    getSecurityRuleKeys().forEach(key -> sonarWay.activateRule(VbNetPlugin.REPOSITORY_KEY, key));
+    sonarWay.done();
+  }
+
+  private static Set<String> getSecurityRuleKeys() {
```
Evangelink: We don't have SonarSecurity for VB.Net so please remove this method.

duncanp-sonar:
FYI this PR has been superceded by #1864 - same set of commits, rebased on master, with additional commits to fix the ITs. There are no additional product changes.

PR #1850, opened 2018-09-12
Import third party Roslyn issues (fix #1825)

### LINKED ISSUE #1825: Update SonarC# to allow import of other roslyn issues
### Description

As valuable as they can be, SonarC# and SonarVBNet roslyn analyzers are "only" some generic rule engines able to find some common code quality issues in C# and VB.Net source files. In other words, SonarC# and SonarVB.Net are part of a highly complementary ecosystem of roslyn analyzers. They might be some functional overlaps between those analyzers but usually they are more complementary than in competition.

### Tasks to do

- Support a new parameter `sonar.cs.roslyn.ignoreIssues` which should act as a switch to allow importing external issues into SonarQube/SonarCloud

- Support the following parameters: `sonar.cs.roslyn.bugCategories`, `sonar.cs.roslyn.codeSmellCategories` and `sonar.cs.roslyn.vulnerabilityCategories` which will allow users to specify which category this issue should belong to.

- Detect source of the issue to feed the correct API:
    - Sonar
    - Roslyn SDK
    - External (adhoc)

- With the new update of the scanner for msbuild we won't prevent analyzers added to the projects to run during the build (whether or not the `ignoreIssues` flag is set) so we will need to filter out results in the roslyn json.

Relates to [MMF-1332](https://jira.sonarsource.com/browse/MMF-1332)

henryju:
This is still work in progress. I still have to add more UT and do a manual test. But still you can maybe start having a look.

### DIFF
--- a/sonar-dotnet-shared-library/pom.xml
+++ b/sonar-dotnet-shared-library/pom.xml
@@ -14,7 +14,7 @@
   <description>Share code between C# and VB.NET plugins</description>
 
   <properties>
-    <sonar.version>7.4.0.16968</sonar.version>
+    <sonar.version>7.4-alpha1</sonar.version>
     <protobuf.version>3.1.0</protobuf.version>
     <protobuf.compiler>${settings.localRepository}/com/google/protobuf/protoc/${protobuf.version}/protoc-${protobuf.version}-${os.detected.classifier}.exe</protobuf.compiler>
     <!-- used for deployment to SonarSource Artifactory -->
--- a/sonar-dotnet-shared-library/pom.xml
+++ b/sonar-dotnet-shared-library/pom.xml
@@ -14,7 +14,7 @@
   <description>Share code between C# and VB.NET plugins</description>
 
   <properties>
-    <sonar.version>7.4.0.16659</sonar.version>
+    <sonar.version>7.4.0.16968</sonar.version>
     <protobuf.version>3.1.0</protobuf.version>
     <protobuf.compiler>${settings.localRepository}/com/google/protobuf/protoc/${protobuf.version}/protoc-${protobuf.version}-${os.detected.classifier}.exe</protobuf.compiler>
     <!-- used for deployment to SonarSource Artifactory -->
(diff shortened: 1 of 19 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractConfiguration.java
```diff
@@ -173,4 +167,20 @@ private static void ifNotSilent(boolean silent, Runnable r) {
       }
     }
   }
+
+  public boolean importAllIssues() {
+    return configuration.getBoolean(AbstractPropertyDefinitions.getImportAllIssuesProperty(languageKey)).orElse(false);
+  }
+
+  public Set<String> bugCategories() {
+    return new HashSet<>(Arrays.asList(configuration.getStringArray(AbstractPropertyDefinitions.getBugCategoriesProperty(languageKey))));
```
julienlancelot: Minor -> you can statically import Arrays

sonar-dotnet-shared-library/pom.xml
```diff
@@ -14,7 +14,7 @@
   <description>Share code between C# and VB.NET plugins</description>
 
   <properties>
-    <sonar.version>7.3</sonar.version>
+    <sonar.version>7.4.0.16659</sonar.version>
```
duncanp-sonar: Previously it looks like this version was just _major.minor_. Out of interest, is there a reason for moving to a four-part version now?

henryju: This is a QA build, we'll move to an official release when available.

henryju:
@duncanp-sonar I'm a bit uncertain about the log. There are so many requirements (SQ 7.4+ or SonarCloud, latest version of the scanner for msbuild, ...) that I'm a bit afraid of doing more harm than good. We can solve that by documenting the feature (it is easier to fix than writing in logs).

duncanp-sonar:
@henryju re logging: ok.
FYI we raised and fixed #1780 recently to meet a direct request from the support team to provide more detailed logs to help them diagnose customer issues.
I've added more logging to the Scanner for MSBuild to make it easier to see how the analysis settings are calculated (i.e. the set of analyzers used during the build step and the ruleset configuration).

PR #1853, opened 2018-09-13
SONARSEC-223 Add security rules to SonarWay

andrei-epure-sonarsource:


sonar-csharp-plugin/src/test/java/org/sonar/plugins/csharp/CSharpSonarWayProfileTest.java
```diff
@@ -21,78 +21,131 @@
...
+    sonarWay.done();
+    CsRules.ruleKeys = Sets.newHashSet("S2");
+
+    CSharpSonarWayProfile profileDef = new CSharpSonarWayProfile(SonarVersion.SQ_73_RUNTIME);
+    profileDef.define(context);
+  }
+
   @Test
   public void sonar_security_missing() {
-    CSharpSonarWayProfile profileDef = new CSharpSonarWayProfile();
```
andrei-epure-sonarsource: the below changes are only styling changes + passing the `SonarVersion` to `CSharpSonarWayProfile`

sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/CSharpSonarWayProfile.java
```diff
@@ -23,19 +23,37 @@
...
-    getSecurityRuleKeys().forEach(key -> sonarWay.activateRule(CSharpPlugin.REPOSITORY_KEY, key));
+    final String repositoryKey;
+    if (supportsSecurityFrontend) {
+      repositoryKey = "roslyn.sonaranalyzer.security.cs";
+    } else {
+      repositoryKey = CSharpPlugin.REPOSITORY_KEY;
+    }
+    try {
+      getSecurityRuleKeys().forEach(key -> sonarWay.activateRule(repositoryKey, key));
+    } catch (IllegalArgumentException e) {
```
m-g-sonar: We should also catch `IllegalStateException` here just to be sure that we are not making startup of SQ crash.

I know the conditions to raise such exception are going to be hard to combine, but we could still fail if we are using a `SQ >= 7.4` and old version of SonarSecurity (returning some keys). If it's the case, we are then failing with the following:
```
java.lang.IllegalStateException: Rule with key 'roslyn.sonaranalyzer.security.cs:S3649' not found
```

sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/CSharpSonarWayProfile.java
```diff
@@ -23,19 +23,37 @@
...
+  }
 
   @Override
   public void define(Context context) {
     NewBuiltInQualityProfile sonarWay = context.createBuiltInQualityProfile("Sonar way", CSharpPlugin.LANGUAGE_KEY);
     BuiltInQualityProfileJsonLoader.load(sonarWay, CSharpPlugin.REPOSITORY_KEY, "org/sonar/plugins/csharp/Sonar_way_profile.json");
-    getSecurityRuleKeys().forEach(key -> sonarWay.activateRule(CSharpPlugin.REPOSITORY_KEY, key));
+    final String repositoryKey;
+    if (supportsSecurityFrontend) {
+      repositoryKey = "roslyn.sonaranalyzer.security.cs";
```
valhristov: It would be better if we retrieve the repository key from the SonarSecurity's class through reflection. In case when SonarSecurity is old version and the method returning the repository key is missing we should fallback to SonarC#'s repository key (csharpsquid)

duncanp-sonar:
@andrei-epure-sonarsource Please commit this to a feature branch rather than to _master_. We're planning on doing a separate release of SonarC# next week containing the external issue changes.

duncanp-sonar:
@andrei-epure-sonarsource the failing check for _SonarAnalyzer for .NET + C# plugin_ is unrelated to this change so you can ignore it.
I've disabled PR integration for that check for the time being and created #1855 to track that issue.

PR #1858, opened 2018-09-17
Update rspec.ps1 to avoid manual change of sonarpedia.json file

andrei-epure-sonarsource:


scripts/rspec/rspec.ps1
```diff
@@ -322,13 +327,20 @@ function ReplaceTokens($text) {
 
 ### SCRIPT START ###
 
+$sonarpediaFolder = $sonarpediaMap.Get_Item($language)
+Write-Host "Will change directory to $sonarpediaFolder to run rule-api"
+cd $sonarpediaFolder
```
valhristov: you could use `pushd $sonarpediafolder` and then `popd` instead of `cd ../../..`

valhristov: also, would it be possible to do the same without changing folders?

PR #1859, opened 2018-09-18
Switched to msbuild-only protobuf generation (removed build.cmd, move…

valhristov:
…d the Google.Protobuf.Tools dependency to Common)

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
@@ -43,14 +43,23 @@
   </ItemGroup>
 
   <!-- Generate protobuf classes, supports windows and linux execution -->
-  <Target Name="GenerateProtobufClasses" BeforeTargets="CoreCompile" Outputs="./Protobuf/Ucfg.cs;.\Protobuf\AnalyzerReport.cs" Inputs="$(UcfgProtoPath)ucfg.proto;.\Protobuf\AnalyzerReport.proto">
+  <Target Name="GenerateProtobufClasses" BeforeTargets="CoreCompile"
+          Outputs="./Protobuf/Ucfg.cs;./Protobuf/AnalyzerReport.cs"
+          Inputs="$(UcfgProtoPath)ucfg.proto;./Protobuf/AnalyzerReport.proto">
     <Message Importance="high" Text="Generating protobuf classes..." />
+    <!-- There is no built-in way to distinguish between Linux and OSX -->
     <PropertyGroup>
-      <!-- This path is platform dependent, it should become conditional when we decide we want to build on Linux or OSX -->
-      <ProtocExe>"$(UserProfile)/.nuget/packages/Google.Protobuf.Tools/3.4.0/tools/windows_x64/protoc.exe"</ProtocExe>
+      <IsWindows Condition="'$(OS)'=='Windows_NT'">true</IsWindows>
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
```diff
@@ -31,13 +33,25 @@
...
   <ItemGroup>
-    <None Include="..\..\..\sonar-ucfg\src\main\protobuf\ucfg.proto" Link="Protobuf\ucfg.proto" />
+    <None Include="$(UcfgProtoPath)ucfg.proto" Link="Protobuf\ucfg.proto" />
   </ItemGroup>
 
-  <Target Name="GenerateProtobufClasses" BeforeTargets="PrepareForBuild" Outputs="Protobuf\AnalyzerReport.cs,Protobuf\Ucfg.cs" Inputs="Protobuf\AnalyzerReport.proto,..\..\..\sonar-ucfg\src\main\protobuf\ucfg.proto">
-    <Message Importance="high" Text="Generating protobuf class..." />
-    <Exec WorkingDirectory="Protobuf" Command="build.bat" />
+  <!-- Generate protobuf classes, supports windows and linux execution -->
+  <Target Name="GenerateProtobufClasses" BeforeTargets="CoreCompile" Outputs="./Protobuf/Ucfg.cs;.\Protobuf\AnalyzerReport.cs" Inputs="$(UcfgProtoPath)ucfg.proto;.\Protobuf\AnalyzerReport.proto">
```
valhristov: This should also fix the unconditional generation, the Inputs/Outputs separator was `,`, but it should have been `;`

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
```diff
@@ -31,13 +33,25 @@
...
+    <Message Importance="high" Text="Generating protobuf classes..." />
+    <PropertyGroup>
+      <!-- This path is platform dependent, it should become conditional when we decide we want to build on Linux or OSX -->
+      <ProtocExe>"$(UserProfile)/.nuget/packages/Google.Protobuf.Tools/3.4.0/tools/windows_x64/protoc.exe"</ProtocExe>
```
valhristov: In SonarSecurity this is multi-platform, I removed the rest of the code because we don't plan to build on Linux so far...

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
```diff
@@ -10,6 +10,8 @@
 
   <ItemGroup>
     <PackageReference Include="Google.Protobuf" Version="3.1.0" />
+    <!-- When changing this reference update ProtocExe property too! -->
+    <PackageReference Include="Google.Protobuf.Tools" Version="3.4.0" />
```
valhristov: Moved the `Tools` dependency here as it is the more appropriate place

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
duncanp-sonar: JulienH was trying to build on Linux last week and this was the main thing that stopped him. If we already have a cross-plat version that works we should use it here too.

valhristov: it won't work for the VSIX project as far as I could imagine... I will copy the cross-plat code here, though.

PR #1880, opened 2018-09-24
Fix S3168: Rule should ignore MSTest V1 cleanup and initialize attrib…

### LINKED ISSUE #1847: Fix S3168: Rule should ignore MSTest V1 cleanup and initialize attributes
### Description

The MSTest `ClassCleanup` method must be _"static, public, does not return a value and should not take any parameter"_ (an exception is thrown otherwise). The `void` return constraint is still in effect when the method is `async`, which triggers `S3168: "async" methods should not return "void"`.

### Repro steps

```cs
[ClassCleanup]
public static async void Cleanup() { ... }
```

### Expected behavior

Sonar should detect the `ClassCleanup` attribute and realize a `Task` could not be returned. Ideally, it would actually enforce the signature that is required by MSTest.

### Actual behavior

An error is triggered by `S3168: "async" methods should not return "void"`.

### Known workarounds

```cs
[SuppressMessage("Sonar.Bug", "S3168:AsyncMethodsShouldNotReturnVoid", Justification = "MSTest")]
```

### Related information

* SonarC# Version: 7.5.0.6605
* Visual Studio Version: 2017 (15.8.3)

Evangelink:
…utes

Fix #1847


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/AsyncVoidMethod_MsTestV2.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Tests.Diagnostics
+{
+    [TestClass]
+    public class MyUnitTests // MSTest V2 has proper support for async so people should avoid async void
+    {
+        [AssemblyCleanup]
+        public static async void AssemblyCleanup() { } // Noncompliant
```
valhristov: Please, add several compliant tests for V2

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/AsyncVoidMethod.cs
```diff
@@ -66,7 +77,8 @@ protected override void Initialize(SonarAnalysisContext context)
 
         private static bool IsExceptionToTheRule(MethodDeclarationSyntax methodDeclaration, IMethodSymbol methodSymbol) =>
             methodSymbol.IsEventHandler() ||
-            IsUsedAsEventHandler(methodDeclaration);
+            IsUsedAsEventHandler(methodDeclaration) ||
+            HasAnyMsTestV1AllowedAttribute(methodSymbol);
```
valhristov: This could probably would have been simpler if it was the other way around - report only on methods with attributes coming from the assembly of MSTest V2. Yes, it will not be future proof, because some day MS could release V3, but I think it would be far in the future... Anyway, this is just a note and I don't think the implementation should be changed if you have doubts.

Evangelink:
I asked builders to update the version of `nuget.exe` installed on CI machines because the latest version might fix this long path issue.

PR #1882, opened 2018-09-24
Rule 1301 for VB:  "switch" statements should have at least 3 "case" clauses

loic-joly-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/SwitchCasesMinimumThreeBase.cs
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules
+{
+
```
Evangelink: [cosmetic] un-needed blank line.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/SwitchCasesMinimumThreeTest.cs
```diff
@@ -33,5 +33,14 @@ public void SwitchCasesMinimumThree()
         {
```
Evangelink: Please add `_CS` to the test name.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/SwitchCasesMinimumThree.cs
```diff
@@ -0,0 +1,62 @@
...
+                    var selectNode = (SelectBlockSyntax)c.Node;
+                    if (!HasAtLeastThreeLabels(selectNode))
+                    {
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, selectNode.SelectStatement.SelectKeyword.GetLocation()));
+                    }
+                },
+                SyntaxKind.SelectBlock);
+        }
+
+        private static bool HasAtLeastThreeLabels(SelectBlockSyntax node)
```
Evangelink: [minor] We try to use arrowed body for short methods like this one.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/SwitchCasesMinimumThree.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            return node.CaseBlocks.Sum(caseBlock => caseBlock.CaseStatement.Cases.Count) >= 3;
```
Evangelink: Could you check the C# implementation to see if we do count multiple values on the same case? If not, could you update (and create a ticket) so that we have the same behavior for C# and VB.Net

loic-joly-sonarsource: `Case 1, 2` is the equivalent of `case 1: case 2:` in C#, which is already counted as 2 cases. Was that what you has in mind when saying *"if we do count multiple values on the same case"*?

PR #1898, opened 2018-09-26
Rule S1481: Unused local variables should be removed

valhristov:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/VariableUnusedBase.cs
```diff
@@ -0,0 +1,76 @@
...
+                yield return symbolInfo.Symbol;
+            }
+
+            foreach (var candidate in symbolInfo.CandidateSymbols.WhereNotNull())
+            {
+                yield return candidate;
+            }
+        }
+    }
+}
```
Evangelink: Missing empty line at the end of the file.

PR #1902, opened 2018-09-26
 Rule S1313: IP addresses should not be hardcoded

valhristov:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/HardcodedIpAddressTest.cs
```diff
@@ -19,8 +19,10 @@
  */
 
 extern alias csharp;
+extern alias vbnet;
```
andrei-epure-sonarsource: why do we use this?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/HardcodedIpAddress.vb
```diff
@@ -0,0 +1,32 @@
...
+        End Sub
+    End Class
+
+    Public Class HardcodedIpAddress
+        Private Shared Sub WriteAssemblyInfo(ByVal assemblyName As String, ByVal version As String, ByVal author As String, ByVal description As String, ByVal title As String)
+        End Sub
+
+        <SomeAttribute("127.0.0.1")>
+        Public Sub Foo()
+            Dim ip As String = "192.168.0.1" 'Noncompliant {{Make sure using this hardcoded IP address '192.168.0.1' is safe here.}}
```
andrei-epure-sonarsource: could you add a test with 
`Dim urlWithPort As String = "http://192.168.0.1:8000/admin.html"; // Noncompliant `

Also, from the RSPEC

> Strings of the form 2.5.<number>.<number> as they often match Object Identifiers (OID).

sonaranalyzer-dotnet/rspec/vbnet/S1313_vb.net.html
```diff
@@ -0,0 +1,39 @@
...
+<p>Last but not least it has an effect on application security. Attackers might be able to decompile the code and thereby discover a potentially
+sensitive address. They can perform a Denial of Service attack on the service at this address or spoof the IP address. Such an attack is always
+possible, but in the case of a hardcoded IP address the fix will be much slower, which will increase an attack's impact.</p>
+<h2>Recommended Secure Coding Practices</h2>
+<ul>
+  <li> make the IP address configurable. </li>
+</ul>
+<h2>Noncompliant Code Example</h2>
+<pre>
+String ip = "192.168.12.42"; // Noncompliant
```
andrei-epure-sonarsource: this should be VB.NET code

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/HardcodedIpAddressTest.cs
valhristov: We introduced this because we needed to test classes that have the same name and namespace in the VB and C# projects... If I remember correctly we decided to use the aliases for all `using`s to avoid confusion. I admit it is not always needed, for example in this particular test class we can very well work without the alias.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/HardcodedIpAddress.vb
valhristov: The URI will be caught by [S1075](https://jira.sonarsource.com/browse/RSPEC-1075)

valhristov: I updated the compliant/noncompliant examples and the exceptions to be the same as C#

PR #1904, opened 2018-09-26
[DO NOT MERGE] Remove everything that relates to uCFG

Evangelink:


sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/CSharpSonarWayProfile.java
```diff
@@ -19,53 +19,15 @@
...
 import org.sonarsource.analyzer.commons.BuiltInQualityProfileJsonLoader;
 
 public class CSharpSonarWayProfile implements BuiltInQualityProfilesDefinition {
-  private static final Logger LOG = Loggers.get(CSharpSonarWayProfile.class);
 
   @Override
   public void define(Context context) {
     NewBuiltInQualityProfile sonarWay = context.createBuiltInQualityProfile("Sonar way", CSharpPlugin.LANGUAGE_KEY);
     BuiltInQualityProfileJsonLoader.load(sonarWay, CSharpPlugin.REPOSITORY_KEY, "org/sonar/plugins/csharp/Sonar_way_profile.json");
-    final String repositoryKey = getRepositoryKey();
```
valhristov: The changes in this file are still required. Since we SQ does not provide API to add rules to the default quality profile from more than one plugin, we need to keep this hack here... At some point in the future SonarSecurity will become "engine extension" which operates on much lower level than the normal plugins and will be able to register rules in the SonarC#'s Sonar Way profile. The corresponding tests should be kept as well.

PR #1910, opened 2018-09-27
Rule S2068: Credentials should not be hard-coded

valhristov:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/DoNotHardcodeCredentialsBase.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        private string credentialWords;
+        private IEnumerable<string> splitCredentialWords;
+        private Regex passwordValuePattern;
+        
+        [RuleParameter("credentialWords", PropertyType.String,
+            "Comma separated list of words identifying potential credentials", DefaultCredentialWords)]
+        public string CredentialWords
+        {
+            get => this.credentialWords;
```
andrei-epure-sonarsource: this isn't covered by tests - is it used?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/DoNotHardcodeCredentialsBase.cs
```diff
@@ -0,0 +1,127 @@
...
+                            Diagnostic.Create(rule, declarator.GetLocation(), string.Join(", ", bannedWords)));
+                    }
+                };
+
+            protected IEnumerable<string> FindBannedWords(string variableName, string variableValue)
+            {
+                if (string.IsNullOrWhiteSpace(variableName) ||
+                    string.IsNullOrWhiteSpace(variableValue))
+                {
+                    return Enumerable.Empty<string>();
```
andrei-epure-sonarsource: could you add a test to cover this line? probably an empty string value?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/DoNotHardcodeCredentialsBase.cs
```diff
@@ -0,0 +1,127 @@
...
+                    string.Join("|", this.splitCredentialWords)), RegexOptions.Compiled | RegexOptions.IgnoreCase);
+            }
+        }
+
+        public DoNotHardcodeCredentialsBase()
+        {
+            CredentialWords = DefaultCredentialWords;
+        }
+
+        protected abstract class BannedWordsFinderBase<TSyntaxNode>
```
andrei-epure-sonarsource: given that this looks for `CredentialWords`, could you use the same terminology for both ?

Or maybe instead of `{Credential|Banned}Word`, use `CredentialIdentifier`

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/DoNotHardcodeCredentialsBase.cs
valhristov: It is not used indeed, but it is silly to create a property with just a setter (SQ will also complain)... Also CredentialWords is required to be a property, so we don't have much choice.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/DoNotHardcodeCredentialsBase.cs
valhristov: Apparently the check for `variableName` is not needed... I put it because we don't support assignments like `this.passwords = "foo"` and I was getting NRE in a previous step in the execution, but apparently `variableName` is ok to be null, because `SplitCamelCaseToWords` will return empty enumerable...

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/DoNotHardcodeCredentialsBase.cs
andrei-epure-sonarsource: I'd rather have SQ complaining (and mark as Won't fix) than having useless code

PR #1913, opened 2018-09-27
Fix unit test related to opt out of external issues

Evangelink:


sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/AbstractConfigurationTest.java
```diff
@@ -177,17 +177,17 @@ public void onlyOldProtobufReportsPresent() throws IOException {
   }
 
   @Test
-  public void externalIssueIsFalseByDefault() throws IOException {
+  public void externalIssueIsFalseByDefault() {
```
henryju: s/externalIssueIsFalseByDefault/ignoreExternalIssuesIsFalseByDefault

Evangelink: done

PR #1915, opened 2018-09-27
SONARVB-297 Rule S2757: "=+" should not be used instead of "+="

andrei-epure-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ReversedOperatorsBase.cs
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+using System;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules
+{
+    public abstract class ReversedOperatorsBase : SonarDiagnosticAnalyzer
```
valhristov: Why do we need two base classes?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ReversedOperators.vb
```diff
@@ -0,0 +1,32 @@
...
+      target=- num ' Noncompliant
+
+      target = - num
+      target = -num ' Compliant intent To assign inverse value Of num Is clear
+      target =-num ' Compliant most probably intent To assign inverse value?
+      target=-num ' Compliant most probably intent To assign inverse value?
+
+      target += num
+
+      target += -num
```
loic-joly-sonarsource: What about `target +=- num`?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ReversedOperatorsBase.cs
loic-joly-sonarsource: I'm not sure about the value of this class. Instead of:
```
SonarDiagnosticAnalyzer<-ReversedOperatorsBase <- ReversedOperatorsBase<-TUnaryExpressionSyntax><-vb/c# version
// Why not
SonarDiagnosticAnalyzer<-ReversedOperatorsBase<-TUnaryExpressionSyntax><-vb/c# version
```
(I've seen this pattern used for other rules, it's just that I don't see why there too)

sonaranalyzer-dotnet/rspec/vbnet/S2757_vb.net.html
```diff
@@ -0,0 +1,21 @@
+<p>The use of operators pairs ( <code>=+</code>, <code>=-</code> or <code>=!</code> ) where the reversed, single operator was meant (<code>+=</code>,
+<code>-=</code> or <code>!=</code>) will compile and run, but not produce the expected results.</p>
+<p>This rule raises an issue when <code>=+</code>, <code>=-</code>, or <code>=!</code> is used without any spacing between the two operators and when
```
loic-joly-sonarsource: There should not be a reference to `=!`

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/ReversedOperators.cs
```diff
@@ -0,0 +1,54 @@
...
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                GetAnalysisAction(rule),
+                SyntaxKind.UnaryMinusExpression,
+                SyntaxKind.UnaryPlusExpression);
```
loic-joly-sonarsource: Maybe a comment saying that, in contrast with C#, the `Not` operator is not concerned? (or a comment in the C# version?)
This looks like an important difference between both implementation, and the difference is IMO too subtle to see in the source code.

sonaranalyzer-dotnet/rspec/vbnet/S2757_vb.net.html
andrei-epure-sonarsource: good catch!

PR #1917, opened 2018-09-28
Rule S3464: Type inheritance should not be recursive

loic-joly-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/GenericInheritanceShouldNotBeRecursive.cs
```diff
@@ -0,0 +1,59 @@
...
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
+                {
+                    var typeDeclaration = (TypeBlockSyntax)c.Node;
+                    var typeSymbol = c.SemanticModel.GetDeclaredSymbol(typeDeclaration);
+                    if (IsRecursiveInheritance(typeSymbol))
+                    {
+                        c.ReportDiagnosticWhenActive(
+                            Diagnostic.Create(rule, typeDeclaration.BlockStatement.Identifier.GetLocation(),
+                            typeDeclaration.BlockStatement.DeclarationKeyword));
```
andrei-epure-sonarsource: nitpick: this should be indented

PR #1923, opened 2018-09-28
Fix S4457: False positive when ArgumentException thrown after awaited…

### LINKED ISSUE #1819: Fix S4457: False positive when ArgumentException thrown after awaited call
```cs
        public async Task Handle()
        {
            var customers = await _loader.GetAllCustomers();

            if (customers.Count == 3)
                throw new ArgumentException("oops");
        }
```

7.4.0.6452
VS 15.8.1

Evangelink:
… call

Fix #1819


Evangelink:
@duncanp-sonar the OOM are a bit random and flickering but I have been able to link it to the update of Roslyn on the Unit Tests. I wanted to create an issue on Roslyn side but I haven't had time yet.

PR #1929, opened 2018-10-01
 SONARVB-333 Rule S3869: "SafeHandle.DangerousGetHandle" should not be called

andrei-epure-sonarsource:
Please **squash** commits when merging.


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -204,6 +204,23 @@ public static bool IsStringEmpty(this ExpressionSyntax expression, SemanticModel
...
+        {
+            if (invocation == null)
+            {
+                return null;
+            }
+            if (invocation.Expression is IdentifierNameSyntax directMethodCall)
+            {
+                return directMethodCall.Identifier;
+            }
+            if (invocation.Expression is MemberAccessExpressionSyntax memberAccessCall)
```
Evangelink: You probably want one more case with `MemberBindingExpressionSyntax`.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Helpers/VisualBasicSyntaxHelper.cs
```diff
@@ -81,5 +81,22 @@ public static StatementSyntax GetSucceedingStatement(this StatementSyntax curren
...
+        {
+            if (invocation == null)
+            {
+                return null;
+            }
+            if (invocation.Expression is IdentifierNameSyntax directMethodCall)
+            {
+                return directMethodCall.Identifier;
+            }
+            if (invocation.Expression is MemberAccessExpressionSyntax memberAccessCall)
```
Evangelink: Same as for C# here.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/RspecStrings.resx
```diff
@@ -1174,7 +1174,7 @@
     <value>Because it is easy to extract strings from a compiled application, credentials should never be hard-coded. Do so, and they're almost guaranteed to end up in the hands of an attacker. This is particularly true for applications that are distributed.</value>
   </data>
   <data name="S2068_IsActivatedByDefault" xml:space="preserve">
-    <value>True</value>
+    <value>False</value>
```
Evangelink: I don't think this change should be done in this PR.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DangerousGetHandleShouldNotBeCalledTest.cs
```diff
@@ -29,9 +29,17 @@ public class DangerousGetHandleShouldNotBeCalledTest
     {
         [TestMethod]
         [TestCategory("Rule")]
-        public void DangerousGetHandleShouldNotBeCalled()
+        public void DangerousGetHandleShouldNotBeCalled_CS()
         {
-            Verifier.VerifyAnalyzer(@"TestCases\DangerousGetHandleShouldNotBeCalled.cs", new DoNotCallDangerousGetHandleMethod());
+            Verifier.VerifyAnalyzer(@"TestCases\DangerousGetHandleShouldNotBeCalled.cs", new csharp.DangerousGetHandleShouldNotBeCalled());
```
Evangelink: [cosmetic][minor] Please move the new analyzer part on the next line.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DangerousGetHandleShouldNotBeCalledTest.cs
```diff
@@ -29,9 +29,17 @@ public class DangerousGetHandleShouldNotBeCalledTest
...
+        }
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void DangerousGetHandleShouldNotBeCalled_VB()
+        {
+            Verifier.VerifyAnalyzer(@"TestCases\DangerousGetHandleShouldNotBeCalled.vb", new vbnet.DangerousGetHandleShouldNotBeCalled());
```
Evangelink: [cosmetic][minor] Please move the new analyzer part on the next line.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Helpers/VisualBasicSyntaxHelper.cs
andrei-epure-sonarsource: in vb.net when using the null-conditional operator (e.g. `handle?.IsClosed`), the parser will generate a `SimpleMemberAccessExpression` as well (as opposed to C#)

Evangelink: Please add this comment in the code directly. It will avoid wondering why we didn't handled this case.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotCallMethodsCsharpBase.cs
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    public abstract class DoNotCallMethodsCsharpBase : DoNotCallMethodsBase<InvocationExpressionSyntax>
```
Evangelink: [minor] Please use `CSharp` instead of `Csharp`.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DangerousGetHandleShouldNotBeCalled.cs
```diff
@@ -29,19 +30,17 @@ namespace SonarAnalyzer.Rules.CSharp
 {
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     [Rule(DiagnosticId)]
-    public sealed class DoNotCallDangerousGetHandleMethod : DoNotCallMethodsBase
+    public sealed class DangerousGetHandleShouldNotBeCalled : DangerousGetHandleShouldNotBeCalledBase<InvocationExpressionSyntax>
```
Evangelink: Is it normal that this class doesn't inherit the CSharp base class?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -204,6 +204,27 @@ public static bool IsStringEmpty(this ExpressionSyntax expression, SemanticModel
                    (methodDeclaration as DestructorDeclarationSyntax)?.Identifier;
         }
 
+        public static SyntaxToken? GetMethodCallIdentifier(this InvocationExpressionSyntax invocation)
+        {
+            if (invocation == null)
+            {
+                return null;
+            }
+            if (invocation.Expression is IdentifierNameSyntax directMethodCall)
```
Evangelink: I think it would be easier to read with a switch instead of all those `if`s.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DangerousGetHandleShouldNotBeCalled.cs
andrei-epure-sonarsource: `DangerousGetHandleShouldNotBeCalledBase` (which inherits `DoNotCallMethodsBase`)  is common for the CSharp and VB.net implementations.

In order to use `DoNotCallMethodsCSharpBase` and also use the common code from `DangerousGetHandleShouldNotBeCalledBase`, I would have to use an inner class inheriting `DoNotCallMethodsBase`, and making that refactoring for all the rules (that are currently implemented only in C#) does not make sense at the moment.

PR #1932, opened 2018-10-01
RSPEC-1048 Destructors should not throw exceptions (For VB)

loic-joly-sonarsource:
This is not a direct translation, since VB does not have the notion of destructor, so review should be more careful.


sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DoNotThrowFromDestructors.cs
```diff
@@ -0,0 +1,73 @@
...
+using SonarAnalyzer.Rules.Common;
+
+namespace SonarAnalyzer.Rules.VisualBasic
+{
+    [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
+    [Rule(DiagnosticId)]
+    public sealed class DoNotThrowFromDestructors : DoNotThrowFromDestructorsBase
+    {
+        private const string MessageFormat = "Remove this 'Throw' statement.";
+        private static readonly DiagnosticDescriptor rule =
```
Evangelink: [minor][cosmetic] Please add a blank line between the `const` and `static readonly`.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DoNotThrowFromDestructors.cs
```diff
@@ -0,0 +1,73 @@
...
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+
```
Evangelink: [minor][cosmetic] un-needed blank line.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DoNotThrowFromDestructors.cs
```diff
@@ -0,0 +1,73 @@
...
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(c =>
```
Evangelink: [minor][cosmetic] we try (at least for new rules) to move the `c =>` to the next line.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DoNotThrowFromDestructors.cs
```diff
@@ -0,0 +1,73 @@
...
+                SyntaxKind.ThrowStatement);
+        }
+
+        private bool IsFinalizer(MethodBlockSyntax methodBlockSyntax)
+        {
+            if (methodBlockSyntax == null)
+            {
+                return false;
+            }
+            var decl = methodBlockSyntax.SubOrFunctionStatement;
```
Evangelink: [minor][cosmetic] please add a blank line after the `if` block.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DoNotThrowFromDestructors.cs
Evangelink: Also please fix the indent of the lambda block.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DoNotThrowFromDestructors.cs
```diff
@@ -0,0 +1,73 @@
...
+            var noParam = decl.ParameterList == null || decl.ParameterList.Parameters.Count == 0;
+            var noTypeParam = decl.TypeParameterList == null || decl.TypeParameterList.Parameters.Count == 0;
+            var isSub = decl.SubOrFunctionKeyword.Kind() == SyntaxKind.SubKeyword;
```
Evangelink: Please use `IsKind()` instead of `Kind() ==`.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DoNotThrowFromDestructors.cs
Evangelink: Also please use a full name and not `decl`.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DoNotThrowFromDestructors.cs
```diff
@@ -0,0 +1,73 @@
...
+            var isProtected = decl.Modifiers.Any(testc => testc.Kind() == SyntaxKind.ProtectedKeyword);
```
Evangelink: Same here, `IsKind()` instead of `Kind() ==`. Also I think there is an overload to `Any` or something like `AnyKind` which does the same.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/RspecStrings.resx
```diff
@@ -1174,7 +1204,7 @@
     <value>Because it is easy to extract strings from a compiled application, credentials should never be hard-coded. Do so, and they're almost guaranteed to end up in the hands of an attacker. This is particularly true for applications that are distributed.</value>
   </data>
   <data name="S2068_IsActivatedByDefault" xml:space="preserve">
-    <value>True</value>
+    <value>False</value>
```
Evangelink: This change should not be part of this PR.

loic-joly-sonarsource:
@Evangelink (not sure why, I cannot directly answer your comment...)
> Are we handling the fact that an exception thrown is caught in the destructor?
No, we are not. This is consistent with the C# behaviour, with a specific unit test about that, that claims: 
```c#
 // Noncompliant, generally a bad idea to throw and catch in the same method
```

I did not feel a strong need to worry about that because finalizers should not be common, finalizers that throw even less (for instance, the rule is never triggered in integration tests), and finalizers that throw but then do it in a try/catch are a corner or a corner case...

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DoNotThrowFromDestructors.cs
loic-joly-sonarsource: I could not find `AnyKind` or anything similar...

Evangelink: If you look at `Metrics.cs` line `129` you can see that there is a built-in extension method (for C#) called `Any` that takes directly a `SyntaxKind`. I don't know if such extension methods exists for VB.NET. If it doesn't then please ignore this comment.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DoNotThrowFromDestructors.cs
loic-joly-sonarsource: Is there a reference explaining the team logic about blank lines? Because you often ask to remove "unneeded" blank lines or to add blanks lines (that I would consider unneeded), and I fail to see where.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/RspecStrings.resx
loic-joly-sonarsource: Do you know how this can happen?

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DoNotThrowFromDestructors.cs
Evangelink: No we don't have one because we naturally had the same logic within the team (for spacing) so we didn't do anything.

Evangelink:
@loic-joly-sonarsource regarding the resx change, there is a bug in the `rspec.ps1` script.

And regarding the throw/catch, I wasn't asking for a change of behavior. I just wanted to confirm that the unit test you wrote was what we were expecting :)

PR #1933, opened 2018-10-02
RSPEC-2551 Types and "Me" should not be used for locking (for VB)

loic-joly-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/RspecStrings.resx
```diff
@@ -1174,7 +1174,7 @@
     <value>Because it is easy to extract strings from a compiled application, credentials should never be hard-coded. Do so, and they're almost guaranteed to end up in the hands of an attacker. This is particularly true for applications that are distributed.</value>
   </data>
   <data name="S2068_IsActivatedByDefault" xml:space="preserve">
-    <value>True</value>
+    <value>False</value>
```
Evangelink: Undo this change.

PR #1935, opened 2018-10-02
 SONARVB-332 Rule S3889: Do not use Thread.{Resume|Suspend}

andrei-epure-sonarsource:
please **squash** when committing


sonaranalyzer-dotnet/src/SonarAnalyzer.Utilities/Rules.Description/S3889_vb.html
```diff
@@ -0,0 +1,18 @@
+<p><code>Thread.Suspend</code> and <code>Thread.Resume</code> can give unpredictable results, and both methods have been deprecated. Indeed, if
+<code>Thread.Suspend</code> is not used very carefully, a thread can be suspended while holding a lock, thus leading to a deadlock. Other safer
+synchronization mechanisms should be used, such as <code>Monitor</code>, <code>Mutex</code>, and <code>Semaphore</code>.</p>
+<h2>Noncompliant Code Example</h2>
+<pre>
+static void Main(string[] args)
```
andrei-epure-sonarsource: must update with for vb.net code

Evangelink: Do you take care of it?

PR #1936, opened 2018-10-02
SONARVB-331: Rule S4159: Classes should implement their 'ExportAttrib…

Evangelink:
…ute' interfaces


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ShouldImplementExportedInterfaces.cs
```diff
@@ -81,78 +69,11 @@ protected override void Initialize(SonarAnalysisContext context)
...
-
-        private static bool IsContractTypeNamedArgument(AttributeArgumentSyntax argument) =>
-            argument.NameColon?.Name.Identifier.ValueText == "contractType";
-
-        private static bool IsOfExportType(ITypeSymbol type, INamedTypeSymbol exportedType) =>
-            type.GetSelfAndBaseTypes()
-                .Union(type.AllInterfaces)
-                .Any(currentType => currentType.Equals(exportedType));
+        protected override string GetIdentifier(AttributeArgumentSyntax argumentSyntax) =>
+            argumentSyntax.NameColon?.Name.Identifier.ValueText;
```
loic-joly-sonarsource: Two lines lower, you use `argumentSyntax?`, which implies that this argument might be null, but not here.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShouldImplementExportedInterfacesBase.cs
```diff
@@ -0,0 +1,126 @@
...
+ */
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.Common
+{
+    public abstract class ShouldImplementExportedInterfacesBase : SonarDiagnosticAnalyzer
```
loic-joly-sonarsource: What is the use of this class? Why not implementing those into the generic class?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShouldImplementExportedInterfacesBase.cs
```diff
@@ -0,0 +1,126 @@
...
+    {
+        protected static bool IsOfExportType(ITypeSymbol type, INamedTypeSymbol exportedType) =>
+            type.GetSelfAndBaseTypes()
+                .Union(type.AllInterfaces)
+                .Any(currentType => currentType.Equals(exportedType));
+
+        protected INamedTypeSymbol GetExportedTypeSymbol(SeparatedSyntaxList<TArgumentSyntax>? attributeArguments,
+            SemanticModel semanticModel)
+        {
+            if (attributeArguments == null)
```
loic-joly-sonarsource: I believe using `!attributeArguments.HasValue` would be clearer.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShouldImplementExportedInterfacesBase.cs
```diff
@@ -0,0 +1,126 @@
...
+            if (exportedTypeSyntax == null)
+            {
+                return null;
+            }
+
+            return semanticModel.GetSymbolInfo(exportedTypeSyntax).Symbol as INamedTypeSymbol;
+        }
+
+        private TArgumentSyntax GetArgumentFromNamedArgument(IEnumerable<TArgumentSyntax> arguments) =>
+            arguments.FirstOrDefault(x => GetIdentifier(x) == "contractType");
```
loic-joly-sonarsource: Don't we have issues with VB.NET case-insensitivity here?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShouldImplementExportedInterfacesBase.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        private TArgumentSyntax GetArgumentFromSingleArgumentAttribute(IEnumerable<TArgumentSyntax> arguments)
+        {
+            if (arguments.ElementAtOrDefault(1) != null)
```
loic-joly-sonarsource: Wouldn't it be clearer to use `arguments.Count()`?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShouldImplementExportedInterfacesBase.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        private TArgumentSyntax GetArgumentFromDoubleArgumentAttribute(IEnumerable<TArgumentSyntax> arguments,
+            SemanticModel semanticModel)
+        {
+            var firstArgument = GetExpression(arguments.ElementAtOrDefault(0));
+            if (firstArgument != null &&
+                semanticModel.GetConstantValue(firstArgument).Value is string)
+            {
+                // Two arguments, second should be typeof expression
+                return arguments.ElementAtOrDefault(1);
```
loic-joly-sonarsource: We are enumerating several times this IEnumerable (agreed, it's not that large). wouldn't it be simple to pss the argument as `SeparatedSyntaxList<TArgumentSyntax>` and use the indexer directly?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShouldImplementExportedInterfacesBase.cs
```diff
@@ -0,0 +1,126 @@
...
+            {
+                return null;
+            }
+
+            return semanticModel.GetDeclaredSymbol(attributeTarget) as ITypeSymbol;
+        }
+
+        protected abstract string GetIdentifier(TArgumentSyntax argumentSyntax);
+        protected abstract TExpressionSyntax GetExpression(TArgumentSyntax argumentSyntax);
+        protected abstract SyntaxNode GetExportedTypeSyntax(TExpressionSyntax expressionSyntax);
```
loic-joly-sonarsource: The purpose of this function is not really obvious... Maybe a small example would help?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ShouldImplementExportedInterfaces.vb
```diff
@@ -0,0 +1,115 @@
...
+    End Class
+    Public class BaseThing2
+    End Class
+
+    <Export(GetType(ISomething(Of BaseThing)))>
+    Public Class BaseSomethingImplementation
+        Implements ISomething(Of BaseThing)
+    End Class
+
+    <Export(GetType(ISomething(Of BaseThing)))> ' Noncompliant
```
loic-joly-sonarsource: I think this case deserves to specify the expected message. Will it contain `ISomething<BaseThing>` or `ISomething(Of BaseThing)`?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShouldImplementExportedInterfacesBase.cs
Evangelink: We have a rule that reports when using static on generic class because you end up with a value that is not actually shared across all instances.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShouldImplementExportedInterfacesBase.cs
Evangelink: Good point, I will make it case insensitive.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShouldImplementExportedInterfacesBase.cs
Evangelink: I assume this also checks that the item in position 1 is not null itself.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShouldImplementExportedInterfacesBase.cs
Evangelink: It would be more complex because we would have to test the size before doing the access.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShouldImplementExportedInterfacesBase.cs
Evangelink: Adding a comment.

PR #1937, opened 2018-10-02
SONARVB-339: Rule S4277: 'Shared' parts should not be created with 'new'

Evangelink:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DoNotInstantiateSharedClassesTest.cs
```diff
@@ -38,11 +38,29 @@ public void DoNotInstantiateSharedClasses()
...
+                additionalReferences: FrameworkMetadataReference.SystemComponentModelComposition);
+        }
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void DoNotInstantiateSharedClasses_VB_InTest()
+        {
+            Verifier.VerifyNoIssueReportedInTest(@"TestCases\DoNotInstantiateSharedClasses.vb",
+                new SonarAnalyzer.Rules.VisualBasic.DoNotInstantiateSharedClasses(),
+                additionalReferences: FrameworkMetadataReference.SystemComponentModelComposition);
```
loic-joly-sonarsource: Why do we need two test cases? What difference does it make?

loic-joly-sonarsource: Sorry if I've been unclear... I was just wondering what `VerifyNoIssueReportedInTest` means...

Evangelink: Oh sorry! Even if SonarQube/SonarCloud doesn't yet support the difference of Main/Test sources, in our analyzer we do. As this rule is only reporting on Main sources we have a test (this one) that ensures that no issue is being reported on a test project for this rule.

PR #1944, opened 2018-10-03
SONARVB-340 Rule S4210: Windows Forms entry points should be marked with STAThread

andrei-epure-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MarkWindowsFormsMainWithStaThreadBase.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    public abstract class MarkWindowsFormsMainWithStaThreadBase<TMethodSyntax> : SonarDiagnosticAnalyzer
+        where TMethodSyntax : SyntaxNode
+    {
+        internal const string DiagnosticId = "S4210";
+        protected const string MessageFormat = "{0}";
+        protected const string AddStaThreadMessage = "Add the 'STAThread' attribute to this entry point.";
+        protected const string ChangeMtaThreadToStaThreadMessage = "Change the 'MTAThread' attribute of this entry point to 'STAThread'.";
+
+        protected abstract IMethodSymbol GetSymbol(SemanticModel model, TMethodSyntax method);
```
Evangelink: I don't understand why you are doing this abstract method? You could call `c.SemanticModel.GetDeclaredSymbol(methodDeclaration)` directly on line `43`.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MarkWindowsFormsMainWithStaThreadBase.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                string message = methodSymbol.GetAttributes(KnownType.System_MTAThreadAttribute).Any()
+                    ? ChangeMtaThreadToStaThreadMessage
+                    : AddStaThreadMessage;
+
+                Report(c, methodDeclaration, message);
+            }
+        }
+
+        protected static bool IsAssemblyReferencingWindowsForms(Compilation compilation)
```
Evangelink: [minor][cosmetic] We usually go for arrowed body when we have a one liner method.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/MarkWindowsFormsMainWithStaThread.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+
+        protected override IMethodSymbol GetSymbol(SemanticModel model, MethodBlockSyntax method) =>
+            model.GetDeclaredSymbol(method);
+
+        protected override void Report(SyntaxNodeAnalysisContext c, MethodBlockSyntax method, string message) =>
+            c.ReportDiagnosticWhenActive(
+                Diagnostic.Create(rule,
+                    method.SubOrFunctionStatement.Identifier.GetLocation(),
```
Evangelink: You could have an abstract method in the base class to get the identifier location. 

FYI instead of using `rule` when you will move this code to the base class you can use `SupportedDiagnostics[0]`.

PR #1947, opened 2018-10-03
SONARVB-343: Rule S3598: One-way 'OperationContract' methods should h…

Evangelink:
…ave 'void' return type


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/WcfNonVoidOneWay.vb
```diff
@@ -0,0 +1,34 @@
...
+        Public Property isOneWay As Boolean
+    End Class
+
+    <ServiceContract>
+    Interface IMyService1
+        <OperationContract>
+        Function MyServiceMethod() As Integer
+
+        <OperationContract(IsOneWay:=True)>
+        Function MyServiceMethod2() As Integer ' Noncompliant {{This method can't return any values because it is marked as one-way operation.}}
```
Evangelink: I wasn't sure if it was better to report on `As Integer` or just `Integer`.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/WcfNonVoidOneWay.vb
```diff
@@ -0,0 +1,34 @@
...
+'                                      ^^^^^^^
+
+        <OperationContract(IsOneWay:=False)>
+        Function MyServiceMethod3() As Integer
+
+        <OperationContract(IsTerminating:=True)>
+        <My(IsOneWay:=True)>
+        Function MyServiceMethod4() As Integer
+
+        <OperationContract(IsOneWay:="mistake")>
```
andrei-epure-sonarsource: could also add a test with `<OperationContract(Action:="Action", IsOneWay:="mistake")>` (where IsOneWay is not the first property in the list)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/WcfNonVoidOneWayTest.cs
```diff
@@ -29,12 +29,20 @@ public class WcfNonVoidOneWayTest
...
                 new WcfNonVoidOneWay(),
                 additionalReferences: FrameworkMetadataReference.SystemServiceModel);
         }
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void WcfNonVoidOneWay_VB()
+        {
+            Verifier.VerifyAnalyzer(@"TestCases\WcfNonVoidOneWay.vb",
+                new SonarAnalyzer.Rules.VisualBasic.WcfNonVoidOneWay(),
```
andrei-epure-sonarsource: nitpick: use alias for `SonarAnalyzer.Rules.VisualBasic`

PR #1948, opened 2018-10-03
SONARVB-342 Rule S4428: "PartCreationPolicyAttribute" should be used with "ExportAttribute"

andrei-epure-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/PartCreationPolicyShouldBeUsedWithExportAttributeBase.cs
```diff
@@ -0,0 +1,69 @@
...
+        protected void Action(SyntaxNodeAnalysisContext c)
+        {
+            var attribute = (TAttributeSyntax)c.Node;
+            if (!IsPartCreationPolicyAttribute(attribute))
+            {
+                return;
+            }
+
+            TClassSyntax classDeclaration = GetClassDeclaration(attribute);
+            if (classDeclaration != null)
```
Evangelink: [minor] I usually prefer early exit compared to nesting.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PartCreationPolicyShouldBeUsedWithExportAttributeTest.cs
```diff
@@ -35,5 +35,15 @@ public void PartCreationPolicyShouldBeUsedWithExportAttribute()
                 new PartCreationPolicyShouldBeUsedWithExportAttribute(),
```
Evangelink: [minor] Please add the `_CS` suffix to the other method name.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/PartCreationPolicyShouldBeUsedWithExportAttributeBase.cs
```diff
@@ -0,0 +1,69 @@
...
+        where TClassSyntax : SyntaxNode
+    {
+        internal const string DiagnosticId = "S4428";
+
+        protected const string MessageFormat = "Add the 'ExportAttribute' or remove 'PartCreationPolicyAttribute'" +
+            " to/from this class definition.";
+
+        protected abstract TClassSyntax GetClassDeclaration(TAttributeSyntax attribute);
+
+        protected void Action(SyntaxNodeAnalysisContext c)
```
Evangelink: [minor] What about `AnalyzeNode` instead? It's still quite generic but it's easier to understand this is going to be the roslyn callback method.

PR #1950, opened 2018-10-04
Rule S3603: Methods with "Pure" attribute should return a value

valhristov:


sonaranalyzer-dotnet/rspec/vbnet/Sonar_way_profile_no_hotspot.json
```diff
@@ -5,8 +5,12 @@
                      "S112",
                      "S114",
                      "S117",
+                     "S1048",
```
valhristov: This is generated by rspec.ps1, I guess this PR was the first to run it with the new changes we made yesterday

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/RspecStrings.resx
```diff
@@ -3319,7 +3319,7 @@
     <value>Minor Security Hotspot</value>
   </data>
   <data name="S2255_Description" xml:space="preserve">
-    <value>Using cookies is security-sensitive. It has led in the past to the following vulnerabilities:</value>
```
Evangelink: Please undo these changes (or move them to a separate PR).

sonaranalyzer-dotnet/rspec/vbnet/Sonar_way_profile_no_hotspot.json
Evangelink: As we have other changes which don't belong to this PR, I think it's better to have only the rule S3603 added here.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/PureAttributeOnVoidMethod.cs
```diff
@@ -0,0 +1,35 @@
...
+using SonarAnalyzer.Common;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.VisualBasic
+{
+    [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
+    [Rule(DiagnosticId)]
+    public sealed class PureAttributeOnVoidMethod : PureAttributeOnVoidMethodBase
+    {
+        protected override DiagnosticDescriptor Rule { get; } =
```
Evangelink: Not sure of the real impact but for all rules we have a static field so that this creation is not done every time. If we think this is equivalent then we should do a PR to update all classes.

Bottom line, please use the static field here.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/PureAttributeOnVoidMethodBase.cs
```diff
@@ -0,0 +1,60 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules
```
Evangelink: AFAIK we always have put the rules in `SonarAnalyzer.Rules.Common`.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PureAttributeOnVoidMethod.cs
```diff
@@ -31,39 +27,9 @@ namespace SonarAnalyzer.Rules.CSharp
...
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     [Rule(DiagnosticId)]
-    public sealed class PureAttributeOnVoidMethod : SonarDiagnosticAnalyzer
+    public sealed class PureAttributeOnVoidMethod : PureAttributeOnVoidMethodBase
     {
-        internal const string DiagnosticId = "S3603";
-        private const string MessageFormat = "Remove the 'Pure' attribute or change the method to return a value.";
-
-        private static readonly DiagnosticDescriptor rule =
+        protected override DiagnosticDescriptor Rule { get; } =
```
Evangelink: See comment on the VB.Net class.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/PureAttributeOnVoidMethodBase.cs
Evangelink: Ok all namespaces in the common don't make sense... Let's fix that later.

PR #1951, opened 2018-10-04
SONARVB-346: Rule S3926: Deserialization methods should be provided f…

Evangelink:
…or 'OptionalField' members

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ProvideDeserializationMethodsForOptionalFieldsBase.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ProvideDeserializationMethodsForOptionalFieldsBase.cs
@@ -41,44 +41,39 @@ namespace SonarAnalyzer.Rules.Common
                 KnownType.System_Runtime_Serialization_OnDeserializedAttribute
             };
 
-    }
-
-    public abstract class ProvideDeserializationMethodsForOptionalFieldsBase<TClassDeclarationSyntax, TClassDeclarationSyntaxKind> :
-        ProvideDeserializationMethodsForOptionalFieldsBase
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/rspec/vbnet/Sonar_way_profile_no_hotspot.json
```diff
@@ -5,8 +5,12 @@
                      "S112",
                      "S114",
                      "S117",
+                     "S1048",
```
valhristov: These should be reverted...

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ProvideDeserializationMethodsForOptionalFieldsBase.cs
```diff
@@ -0,0 +1,111 @@
...
+        where TClassDeclarationSyntaxKind : struct
+    {
+        protected sealed override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                GeneratedCodeRecognizer,
+                c =>
+                {
+                    var classDeclaration = (TClassDeclarationSyntax)c.Node;
+                    if (!(c.SemanticModel.GetDeclaredSymbol(classDeclaration) is INamedTypeSymbol classSymbol))
```
valhristov: This could be improved by either:
- use `(INamedTypeSymbol)c.SemanticModel.GetDeclaredSymbol(c.Node)` and have only one abstract method `GetClassDeclarationIdentifierLocation(SyntaxNode)`, thus removing all generic arguments except `TSyntaxKind`
- try to convert to RegisterSymbolAction and use the symbol's declaring syntax reference to obtain the identifier (still you'll need the virtual method to get the location), thus removing all generic arguments

PR #1955, opened 2018-10-04
Rule S3903: Types should be defined in named namespaces

valhristov:


sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DeclareTypesInNamespaces.cs
```diff
@@ -0,0 +1,64 @@
...
+            if (declaration.Parent.Parent is TypeBlockSyntax || // Inner type
+                declaration.Parent.Parent is NamespaceBlockSyntax) // Outer type within a namespace
+            {
+                return false;
+            }
+
+            // If we are an outer type without a declared namespace,
+            // make sure there is no Root Namespace set in the project
+            var typeSymbol = (INamedTypeSymbol)semanticModel.GetDeclaredSymbol(declaration);
+            return typeSymbol != null && typeSymbol.ContainingNamespace.IsGlobalNamespace;
```
valhristov: VB.NET has this "Root Namespace" setting that allows you not to declare types in namespaces and they will automatically go into that namespace. It is optional, though, so we need to check the symbol.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DeclareTypesInNamespaces.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        protected override void Initialize(SonarAnalysisContext context) =>
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                GetAnalysisAction(rule),
+                SyntaxKind.ClassStatement,
+                SyntaxKind.StructureStatement);
```
valhristov: The `*Statement` is the first line of a `Class/StructBlock`. Registering for the inner node spares us one null check...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DeclareTypesInNamespaces.cs
```diff
@@ -30,32 +30,25 @@ namespace SonarAnalyzer.Rules.CSharp
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } =
+            ImmutableArray.Create(rule);
+
+        protected override SyntaxToken GetTypeIdentifier(SyntaxNode declaration) =>
+            ((TypeDeclarationSyntax)declaration).Identifier;
 
-                    c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, declaration.Identifier.GetLocation(),
-                        declaration.Identifier.ValueText));
-                },
+        protected override bool IsOuterTypeWithoutNamespace(SyntaxNode declaration, SemanticModel semanticModel) =>
```
Evangelink: There is something I don't understand, methods indicate that type is NOT in a namespace and yet we return `true` is node parent is `NamespaceDeclarationSyntax`. What am I missing?

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DeclareTypesInNamespaces.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        protected override bool IsOuterTypeWithoutNamespace(SyntaxNode declaration, SemanticModel semanticModel)
+        {
+            if (declaration.Parent.Parent is TypeBlockSyntax || // Inner type
+                declaration.Parent.Parent is NamespaceBlockSyntax) // Outer type within a namespace
+            {
+                return false;
+            }
+
+            // If we are an outer type without a declared namespace,
```
Evangelink: Missing words on your sentence.

Evangelink: [cosmetic][minor] I don't understand why you line break so soon compared to your method signature length.

valhristov: To have similar length of the two lines. There are no missing words :)

PR #1956, opened 2018-10-04
SONARVB-348: Rule S3927: Serialization event handlers should be imple…

Evangelink:
…mented correctly


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ImplementSerializationMethodsCorrectlyBase.cs
```diff
@@ -0,0 +1,98 @@
...
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules
+{
+    public abstract class ImplementSerializationMethodsCorrectlyBase : SonarDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S3927";
+        protected const string MessageFormat = "Make this method {0}.";
+        private const string problemMakePrivateText = "'private'";
+        private const string problemReturnVoidText = "return 'void'";
```
valhristov: A more precise message would be "Make this function a sub" or "This method should not return a value" or something like this...

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ImplementSerializationMethodsCorrectlyBase.cs
```diff
@@ -0,0 +1,98 @@
...
+                    var issues = FindIssues(methodSymbol).ToList();
+                    if (issues.Count == 0)
+                    {
+                        return;
+                    }
+
+                    var location = GetIdentifierLocation(methodSymbol);
+                    if (location != null)
+                    {
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(SupportedDiagnostics[0], location, issues.ToSentence()));
```
valhristov: I probably made the wrong recommendation yesterday, we used to not report on generated code, now we do. I think you could use the `ReportDiagnosticIfNonGenerated` extension method and provide C# or VB.NET `GeneratedCodeRecognizer`...

PR #1957, opened 2018-10-04
SONARVB-350: Rule S4586: Non-async 'Task/Task<T>' methods should not …

Evangelink:
…return null


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/NonAsyncTaskShouldNotReturnNullBase.cs
```diff
@@ -0,0 +1,63 @@
...
+        private static readonly ISet<KnownType> TaskTypes =
+            new HashSet<KnownType>
+            {
+                KnownType.System_Threading_Tasks_Task,
+                KnownType.System_Threading_Tasks_Task_T
+            };
+
+        protected static bool IsInvalidEnclosingSymbolContext(SyntaxNode enclosingMember, SemanticModel model)
+        {
+            var enclosingMemberSymbol = model.GetDeclaredSymbol(enclosingMember);
```
loic-joly-sonarsource: Wouldn't it be better to cast to `IMethodSymbol` only one, here, instead of doing it in the next two following functions?

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/RspecStrings.resx
```diff
@@ -2547,36 +2547,6 @@
   <data name="S3598_Type" xml:space="preserve">
     <value>BUG</value>
   </data>
-  <data name="S3464_Category" xml:space="preserve">
```
loic-joly-sonarsource: Is that some unrelated change?

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/NonAsyncTaskShouldNotReturnNull.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                return false;
+            }
+        }
+
+        private static SyntaxNode GetEnclosingMember(LiteralExpressionSyntax literal)
+        {
+            foreach (var ancestor in literal.Ancestors())
+            {
+                if (ancestor.IsKind(SyntaxKind.MultiLineFunctionLambdaExpression) ||
```
loic-joly-sonarsource: Would it be clearer to have a  `switch (ancestor.Kind())` (and remove the hashset :) ) ?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/NonAsyncTaskShouldNotReturnNullBase.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace SonarAnalyzer.Rules
+{
+    public abstract class NonAsyncTaskShouldNotReturnNullBase : SonarDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S4586";
+        protected const string MessageFormat = "Do not return null from this method, instead return 'Task.FromResult<T>(null)', " +
+            "'Task.CompletedTask' or 'Task.Delay(0)'.";
+
+        private static readonly ISet<KnownType> TaskTypes =
```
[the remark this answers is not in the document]
Evangelink: I'll keep it like this for now (for consistency) but you are right that we could improve/optimize some parts.

PR #1961, opened 2018-10-08
Rule S3449: Right operands of shift operators should be integers

valhristov:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShiftDynamicNotIntegerBase.cs
```diff
@@ -0,0 +1,83 @@
...
+            var constValue = semanticModel.GetConstantValue(expression);
+            return constValue.HasValue && constValue.Value == null
+                ? "null"
+                : typeOfRight.ToMinimalDisplayString(semanticModel, expression.SpanStart);
+        }
+
+        private bool IsErrorType(TExpressionSyntax expression, SemanticModel semanticModel, out ITypeSymbol type)
+        {
+            type = semanticModel.GetTypeInfo(expression).Type;
+            return type is IErrorTypeSymbol;
```
Evangelink: You can use a one liner and also please use `.Kind == SymbolKind.Error` instead of `is ErrorTypeSymbol`.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/ShiftDynamicNotInteger.cs
```diff
@@ -0,0 +1,71 @@
...
+            !IsConvertibleToInt(right, semanticModel);
+
+        protected override bool CanBeConvertedTo(ExpressionSyntax expression, ITypeSymbol type, SemanticModel semanticModel) =>
+            expression.IsKind(SyntaxKind.NothingLiteralExpression) || // x >> Nothing will not throw, so ignore
+            semanticModel.GetTypeInfo(expression).Type.IsAny(KnownType.IntegralNumbers);
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                Helpers.VisualBasic.GeneratedCodeRecognizer.Instance,
```
Evangelink: I don't understand why we have the `GeneratedCodeRecognizer` here.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShiftDynamicNotIntegerBase.cs
valhristov: `type` is an `out` parameter, I don't think you could assign it in an expression bodied method... Will replace the cast.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShiftDynamicNotIntegerBase.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace SonarAnalyzer.Rules
+{
+    public abstract class ShiftDynamicNotIntegerBase<TExpressionSyntax> : SonarDiagnosticAnalyzer
+        where TExpressionSyntax : SyntaxNode
+    {
+        internal const string DiagnosticId = "S3449";
+        protected const string MessageFormat = "Remove this erroneous shift, it will fail because '{0}' can't be implicitly converted to 'int'.";
+
+        protected abstract DiagnosticDescriptor Rule { get; }
```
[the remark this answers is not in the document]
valhristov: I don't like using `SupportedDiagnostics[0]` because it implies that there could be more than one supported diagnostics and is weird why we chose the first to report and not the second... I know it is used in some rules, but in general I prefer the more explicit single `Rule` override, which is also used in many rules.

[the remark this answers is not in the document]
Evangelink: But reporting with `Rule` doesn't prevent to register more than ` rule in the base class collection.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ShiftDynamicNotIntegerBase.cs
Evangelink: I missed the `out`.

PR #1973, opened 2018-10-09
Fix S1449: Do not raise issues inside of an Expression

### LINKED ISSUE #1841: Fix S1449: Rule should not report for objects when ToUpper is inside expression
### Description

S1449 incorrectly reports for the `ToUpper` invocation below. If the code is changed as per the recommendation, the query will cease working.

Note: the issue was originally reported for VB.NET, but it should also apply for C#

```
Dim londonCustomers = From customer in Customers _
	Where customer.City = “London” && _
	customer.code.ToUpper().Trim() == “Somthing”
Select customer
```

Original community thread:
https://community.sonarsource.com/t/c-rule-culture-should-be-specified-for-string-operations/2364

### LINKED ISSUE #1994: Fix S3254: Correctly detect Expression in function calls to silence the rule
Issues for [S3254](https://jira.sonarsource.com/browse/RSPEC-3254) are not raised when the call is inside of an Expression. However, currently, Expressions are only detected if they are assigned to a variable, not if they are directly used in a function call.

loic-joly-sonarsource:
~~No change in behaviour, the unit test was already working :)~~
Sorry, I was wrong about that... Here is a proper correction.

Related to community thread:
https://community.sonarsource.com/t/c-rule-culture-should-be-specified-for-string-operations/2364

closes #1841 
closes #1994

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
@@ -262,39 +262,29 @@ namespace SonarAnalyzer.Helpers
             // * VariableDeclarationSyntax (for local variable or field),
             // * PropertyDeclarationSyntax,
             // * SimpleAssigmentExpression
-            return node.Ancestors()
-                .Select(GetExpressionTree)
-                .FirstOrDefault(tuple => tuple.Item1)
-                ?.Item2?.IsKnownType(KnownType.System_Linq_Expressions_Expression_T, semanticModel)
-                ?? false;
-
-            Tuple<bool, SyntaxNode> GetExpressionTree(SyntaxNode syntaxNode)
+            foreach (var n in node.Ancestors())
             {
-                switch (syntaxNode.Kind())
+                switch (n.Kind())
                 {
-                    case SyntaxKind.VariableDeclaration:
-                        return new Tuple<bool, SyntaxNode>(true, ((VariableDeclarationSyntax)syntaxNode).Type);
-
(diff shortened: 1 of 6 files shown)


Evangelink:
I am a bit bothered by the fact we are fully excluding the check when it's inside a lambda because I think we are cutting too many cases while still not handling everything.

For example, if you have a method that is handling the lambda of the where part then we will still report.

```csharp
var londonCustomers = allCustomers.Where(LiveInLondon);

bool LiveInLondon(Customer c)
{
    return c.City.ToUpper() == "LONDON"; // Issue is raised here but we cannot change the implementation
}
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -262,39 +262,30 @@ public static bool IsInExpressionTree(this SyntaxNode node, SemanticModel semant
...
+                    case SyntaxKind.OrderByClause:
+                    case SyntaxKind.SelectClause:
+                    case SyntaxKind.GroupClause:
+                        // For those clauses, we don't know how to differentiate an expression tree from a delegate,
+                        // so we assume we are in the (more restricted) expression tree
+                        return true;
+                    case SyntaxKind.SimpleLambdaExpression:
+                    case SyntaxKind.ParenthesizedLambdaExpression:
+                        return semanticModel.GetTypeInfo(n).ConvertedType?.OriginalDefinition.Is(KnownType.System_Linq_Expressions_Expression_T)
+                            ??false;
```
Evangelink: [cosmetic] You are going to say I am boring but please add a whitespace between `??` and `false`.

PR #1974, opened 2018-10-10
Upgrading Microsoft.VSSDK.BuildTools to be compatible with latest VS updates

valhristov:
Updated single nuget


sonaranalyzer-dotnet/src/SonarAnalyzer.Vsix/packages.SonarAnalyzer.Vsix.config
```diff
@@ -1,5 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Google.Protobuf" version="3.6.1" targetFramework="net461" />
-  <package id="Microsoft.VSSDK.BuildTools" version="15.5.72" targetFramework="net46" developmentDependency="true" />
+  <package id="Microsoft.VisualStudio.SDK.Analyzers" version="15.8.33" targetFramework="net461" developmentDependency="true" />
```
valhristov: it doesn't really matter, but I made the new analyzer packages that are referenced `developmentDependency=true`

PR #1975, opened 2018-10-10
Fix NullReferenceException when invoked function is not a method (e.g. when it is a lambda)

valhristov:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/DoNotOverwriteCollectionElementsBase.cs
```diff
@@ -173,8 +173,8 @@ protected abstract class InvocationVerifierBase
             private bool IsInvocationOnSameItem(SyntaxNode expression, string invokedOn) =>
                 expression is TInvocationSyntax ies &&
                 HasNumberOfArguments(ies, 2) &&
-                GetMethodName(ies).Equals("Add", StringComparison.OrdinalIgnoreCase) &&
-                getInvokedOnName(ies).Equals(invokedOn, StringComparison.OrdinalIgnoreCase);
+                "Add".Equals(GetMethodName(ies), StringComparison.OrdinalIgnoreCase) &&
```
valhristov: GetMethodName returns `null` when `ies` is not a method

PR #1998, opened 2018-10-12
Loic/#1964 compare generic types

### LINKED ISSUE #1964: Fix S3427: Rule should not generated FPs for generic parameters
### Description

When running against a class with methods that differ by multiple generic params, a false positive is raised.

### Repro steps

```csharp
public class MyClass {
    public static Property FromDto(this KeyValuePair<PropertyKey, CreatePropertyRequest> kvp, DateTimeOffset? effectiveDate = null){ ...}
    public static PerpetualProperty FromDto(this KeyValuePair<PropertyKey, CreatePerpetualPropertyRequest> kvp) {...}
}
```

In this case, the KVP's differ by the second generic type.
repro zip attached.
[repro.zip](https://github.com/SonarSource/sonar-csharp/files/2455184/repro.zip)

### Expected behavior

No issue raised.

### Actual behavior

Issue identified.

### Known workarounds

Other than silencing the issue, none.

### Related information

* SonarC# Version : 6.8.1
* Visual Studio Version N/A
* If running through the Scanner for MSBuild, its version 15.x

loic-joly-sonarsource:
Fixes #1964

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadOptionalParameter.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadOptionalParameter.cs
@@ -148,15 +148,31 @@ namespace SonarAnalyzer.Rules.CSharp
         {
             return candidateHidingMethod.Parameters
                     .Zip(method.Parameters, (param1, param2) => new { param1, param2 })
-                    .All(p => AreParameterTypesEqual(p.param1, p.param2) &&
+                    .All(p => AreTypesEqual(p.param1.Type, p.param2.Type) &&
                               p.param1.IsOptional == p.param2.IsOptional);
         }
 
-        private static bool AreParameterTypesEqual(IParameterSymbol p1, IParameterSymbol p2)
+        private static bool AreTypesEqual(ITypeSymbol t1, ITypeSymbol t2)
         {
-            return Equals(p1.Type, p2.Type) ||
-                   p1.Type.Is(TypeKind.TypeParameter) && p2.Type.Is(TypeKind.TypeParameter) ||
-                   Equals(p1.Type.OriginalDefinition, p2.Type.OriginalDefinition);
+            return Equals(t1, t2) ||
+                   t1.Is(TypeKind.TypeParameter) && t2.Is(TypeKind.TypeParameter) ||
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadOptionalParameter.cs
```diff
@@ -148,15 +148,31 @@ private static bool IsMethodHidingOriginal(IMethodSymbol candidateHidingMethod,
...
+                   AreGenericInstancesTypesEqual(t1, t2);
+        }
+
+        private static bool AreGenericInstancesTypesEqual(ITypeSymbol t1, ITypeSymbol t2)
+        {
+            if (t1.OriginalDefinition != t2.OriginalDefinition)
+            {
+                return false;
+            }
+            var named1 = t1 as INamedTypeSymbol;
```
Evangelink: I would write the following as
```csharp
if (t1 is INamedTypeSymbol named1 &&
    t2 is INamedTypeSymbol named2)
{
    return named1.TypeArguments.SequenceEqual(named2.TypeArguments,
       (arg1, arg2) => AreTypesEqual(arg1, arg2));
}

return false;
```

pms1969: I know this is merged now, but what about types with more than 2 generic parameters?

Evangelink: Hi @pms1969,

I think you might have misread the code. Indeed we are actually covering n parameters. If you look the code we actually group (zip) parameters from the method with the potential matching method and then for each parameter we check whether the `TypeArguments` are all equals (`named1.TypeArguments.SequenceEqual(named2.TypeArguments`).

Assuming we have the following 2 methods:
```csharp
void M1(Tuple<int, string, double> t, string msg = null) {}
void M1(Tuple<int, string, double> t) {}
```
We will do a grouping like this:
```csharp
(Tuple<int, string, double>, Tuple<int, string, double>), (string, nothing)
```
As the first group has generic parameters, we compare them all using `SequenceEquals`.

Hope this makes sense now?

PR #2005, opened 2018-10-16
Fix S4261: rule should not report for MVC controller methods

### LINKED ISSUE #1875: Fix S4261: rule should not report for MVC controller methods
### Description

Rule S4261 suggests method Xyz should renamed to XyzAsync if it is an async method. This rule should be suppressed for any public method on a class that derives from the MVC Controller base class, as this may interfere with routing.

Original request: https://community.sonarsource.com/t/s4261-methods-should-be-named-according-to-their-synchronicities-should-not-apply-to-controllers/2436

valhristov:
Fix #1875


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/MethodShouldBeNameAccordingToSynchronicityTest.cs
```diff
@@ -40,6 +40,34 @@ public void MethodShouldBeNameAccordingToSynchronicity(string tasksVersion)
                 additionalReferences: NuGetMetadataReference.SystemThreadingTasksExtensions(tasksVersion));
         }
 
+        [TestMethod]
+        [DataRow("3.0.20105.1")]
+        [DataRow(Constants.NuGetLatestVersion)]
+        [TestCategory("Rule")]
+        public void MethodShouldBeNameAccordingToSynchronicity_MVC(string mvcVersion)
```
loic-joly-sonarsource:
```suggestion
        public void MethodShouldBeNamedAccordingToSynchronicity_MVC(string mvcVersion)
```
(and apply the name change everywhere...)

PR #2017, opened 2018-10-23
Fix S2701: rule should ignore bool? assertions

### LINKED ISSUE #1874: Fix S2701: rule should ignore bool? assertions
# Description

The following causes rule S2701 to fire:

            bool? x = false;
            Assert.AreEqual(false, x);

However, the suggested change fails to compile:

            Assert.IsFalse(x);

Original request: https://community.sonarsource.com/t/should-s2701-apply-to-bool-datatypes/2435

valhristov:
Fix #1874


valhristov:
I already reviewed...

PR #2022, opened 2018-10-26
Detect compilation errors in unit test code

duncanp-sonar:
Same code as https://github.com/SonarSource/sonar-csharp/pull/2014

Updated the test framework to fail if there are compilation errors in the code being tested.
Added new markup to indicate expected errors.

Current status: some but not all not all of the sample test code has been fixed up so that it compiles correctly. The remaining tests may still contain uncompilable code.


duncanp-sonar:
Closing this PR - backend QA build is not being triggered correctly. Replaced by https://github.com/SonarSource/sonar-csharp/pull/2023

PR #2031, opened 2018-10-31
[DO NOT MERGE] Rule S4829: Reading the Standard Input is security-sensitive

valhristov:
This is one of the implementations I came with for factorizing the common code for the Hotspots. I would like to spend some time to create a good framework for creating Hotspot rules in the future.


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/HotspotHelper.cs
```diff
@@ -0,0 +1,59 @@
...
+
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarAnalyzer.Common;
+
+namespace SonarAnalyzer.Helpers
+{
+    public abstract class HotspotHelper
```
duncanp-sonar: This class doesn't seem to have any hotspot-specific functionality; it could just as easily be used for non-hotspot rules, couldn't it?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpHotspotHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            bool IsTrackedMethod(InvocationExpressionSyntax invocation, SemanticModel semanticModel)
+            {
+                var identifier = invocation.Expression.GetIdentifier();
+                return identifier != null &&
+                    trackedMethods.Any(MethodSignatureHelper.IsSameMethod(identifier, semanticModel));
+            }
+        }
+
+        public override void TrackPropertyAccess(SonarAnalysisContext context, DiagnosticDescriptor rule, params MethodSignature[] trackedProperties)
```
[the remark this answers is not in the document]
duncanp-sonar: Might be better to rename the param in the base class.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/MethodSignatureHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+        public static Func<MethodSignature, bool> IsSameProperty(SimpleNameSyntax identifierName, SemanticModel semanticModel)
+        {
+            var identifierText = identifierName.Identifier.ValueText;
+            var identifierSymbol = new Lazy<IPropertySymbol>(() => semanticModel.GetSymbolInfo(identifierName).Symbol as IPropertySymbol);
+            // This function will be called multiple times for each tracked MethodSignature,
+            // hence we cache as much as possible before.
+            return (methodSignature) =>
+                identifierText == methodSignature.Name &&
+                identifierSymbol.Value != null &&
+                identifierSymbol.Value.ContainingType.Is(methodSignature.ContainingType);
```
duncanp-sonar: _IsSameProperty_ and _IsSameMethod_ are identical apart from the symbol type, which isn't visible outside the methods, and presumably there are equivalent methods for VB. Can we avoid the repetition?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ReadingStandardInput.vb
```diff
@@ -0,0 +1,42 @@
...
+            ' Other console methods
+            Console.Write(1)
+            Console.WriteLine(1)
+            ' Other classes
+            MyConsole.Read()
+            MyConsole.[In].Read()
+        End Sub
+    End Class
+
+    Module MyConsole
```
duncanp-sonar: Are we matching the whole namespace, or just the containing type name i.e. would _MyNamespace.Console.Read_ give an FP?

valhristov: We are matching the fully qualified type name (but not the assembly name), so FPs are possible but that would be rare

PR #2035, opened 2018-11-01
Refactor hotspot helpers

valhristov:
Basically split the implementation into several classes:
- to have smaller, specialized classes
- to be more modular (e.g. we could add a flag to the `InvocationTracker` to ignore invocations if their result is ignored and we could have two instances with different settings)
- to easily add new "trackers" without changing the same file
- perhaps more...

I think this would be better if amended with the previous commit


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SyntaxTrackerBase.cs
```diff
@@ -50,13 +54,7 @@ protected bool IsEnabled(AnalyzerOptions options)
                 return false;
             }
 
-            return supportedDiagnostics.Any(d => analysisConfiguration.IsEnabled(d.Id));
+            return analysisConfiguration.IsEnabled(Rule.Id);
```
duncanp-sonar: Not changed in this PR, but there are two checks for _analysisConfiguration.EnabledRules == null_ earlier in this method (lines 47 and 52).

PR #2063, opened 2018-11-02
Modify invocation tracker classes

duncanp-sonar:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/InvocationTracker.cs
```diff
@@ -18,14 +18,15 @@
...
 
-using System;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 
 namespace SonarAnalyzer.Helpers
 {
+    public delegate bool InvocationCondition(InvocationContext invocationContext);
```
valhristov: I know someone who will like this a lot ;)

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Helpers/VisualBasicInvocationTracker.cs
```diff
@@ -43,7 +43,32 @@ public VisualBasicInvocationTracker(IAnalyzerConfiguration analysisConfiguration
...
-        protected override Func<MethodSignature, bool> IsSame<TSymbolType>(SyntaxNode identifier, SemanticModel semanticModel) =>
-            MethodSignatureHelper.IsSame<TSymbolType>((SimpleNameSyntax)identifier, semanticModel);
+
+        #region Syntax-level checking methods
+
+        public override InvocationCondition MatchSimpleNames(params MethodSignature[] methods)
+        {
+            return Check;
+
+            bool Check(InvocationContext invocationContext)
```
valhristov: you could directly `return (invocationContext) => { ... }`, I don't think this will reduce the readability a lot...

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/InvocationTracker.cs
duncanp-sonar: ;) Delegates are a better choice here - _InvocationCondition_ more descriptive and shorter to type than _Func<InvocationContext, bool>_

PR #2064, opened 2018-11-02
Refactor property tracker to match invocation tracker

duncanp-sonar:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/PropertyAccessContext.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System;
+using Microsoft.CodeAnalysis;
+
+namespace SonarAnalyzer.Helpers
+{
+    /// <summary>
+    /// Syntax and semantic information about a single property access
+    /// </summary>
+    public class PropertyAccessContext
```
duncanp-sonar: I've created a separate context class and delegate for the property tracker for the time being. I considered creating a share _SyntaxTrackerContext_ class, but decided not to to start with, but we might want to change it after we've written a few rules.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Helpers/MethodSignatureHelper.cs
```diff
@@ -19,24 +19,28 @@
...
-            // This function will be called multiple times for each tracked MethodSignature,
-            // hence we cache as much as possible before.
-            return (methodSignature) =>
-                identifierText == methodSignature.Name &&
-                identifierSymbol.Value != null &&
-                identifierSymbol.Value.ContainingType.Is(methodSignature.ContainingType);
+            return methods.Any(m =>
+                identifierText == m.Name &&
+                symbolFetcher.Value != null &&
+                symbolFetcher.Value.ContainingType.Is(m.ContainingType));
```
valhristov: I cannot believe I am saying that :), but I feel this LINQ would be probably better if written with a `foreach` (or even `for`) loop:
```
foreach (var m in methods)
{
  if (identifierText != m.Name)
    continue;
  if (symbolFetcher.Value == null)
    break; // No need to continue looping if the symbol is null
  if (symbolFetcher.Value.ContainingType.Is(m.ContainingType))
    return true;
}
return false;
```
Again, this looks like a micro-optimization, but could be something that we could change after we have more rules implemented.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/MethodSignatureHelper.cs
```diff
@@ -19,25 +19,28 @@
...
-            // This function will be called multiple times for each tracked MethodSignature,
-            // hence we cache as much as possible before.
-            return (methodSignature) =>
-                identifierText == methodSignature.Name &&
-                identifierSymbol.Value != null &&
-                identifierSymbol.Value.ContainingType.Is(methodSignature.ContainingType);
+            return methods.Any(m =>
+                identifierText == m.Name &&
+                symbolFetcher.Value != null &&
+                symbolFetcher.Value.ContainingType.Is(m.ContainingType));
```
valhristov: Other classes that match MethodSignature usually use `symbol.ContainingType.ConstructedFrom.Is(knownType)`... This could be fixed later when we find it, but it is definitely something we should have in mind with generic types.

duncanp-sonar: I'll change it now so we don't forget.

duncanp-sonar:
Closing this PR - changes merged in a separate branch with cleaned-up history in PR #2067

PR #2069, opened 2018-11-05
Update test project to the new csproj format

Evangelink:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ResourceTests/GeneratedResourcesTest.cs
```diff
@@ -37,7 +37,7 @@ namespace SonarAnalyzer.UnitTest.ResourceTests
     [TestClass]
     public class GeneratedResourcesTest
     {
-        private const string RspecRelativeFolderPath = @"..\..\..\..\rspec\";
+        private const string RspecRelativeFolderPath = @"..\..\..\..\..\rspec\";
```
Evangelink: the framework is now present in the bin path so we need to get up one more folder.

PR #2072, opened 2018-11-05
Add FluentAssertions analyzer and fix raised issues

Evangelink:


duncanp-sonar:
Looks ok, but it's going to cause merge conflicts with the work in the hotspots branch.
Does this neeed to be done now?

PR #2075, opened 2018-11-05
Stop relying on submodule for ruling projects

Evangelink:
We used to have the projects used as part of the ruling in a separate GitHub repository and used here as a submodule. This way of working wasn't providing much benefits and was actually more of a pain.

This PR copies the different projects directly within this project so that changes can be made more easily.


Evangelink:
@valhristov @duncanp-sonar PR is now ready. I'll let you decide what's the best time to merge to master so you don't have conflict with your work.

FYI I don't expect any conflict as this is just moving the submodule (readonly) to this repo.

PR #2076, opened 2018-11-05
Update project key on SonarCloud

Evangelink:


scripts/build/ci-build.ps1
```diff
@@ -131,7 +131,7 @@ function Invoke-SonarBeginAnalysis([array][parameter(ValueFromRemainingArguments
     }
 
     Exec { & (Get-ScannerMsBuildPath) begin `
-        /k:sonaranalyzer-csharp-vbnet `
+        /k:sonaranalyzer-dotnet `
```
Evangelink: Project key on SonarCloud was different from Next. Use the key of SonarCloud so that we don't lose history.

PR #2077, opened 2018-11-05
Added param-not-constant check to invocation tracker

duncanp-sonar:
Modified _MethodSignatureHelper_ to handle method overrides
Added unit tests for _MethodSignatureHelper_

Test infrastructure:
* added _SnippetCompiler_ class
* refactored _SymbolHelperTests_ to use the new class

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpInvocationTracker.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpInvocationTracker.cs
@@ -45,8 +45,8 @@ namespace SonarAnalyzer.Helpers
 
         public override InvocationCondition MatchSimpleNames(params MethodSignature[] methods)
         {
-            return (context) => MethodSignatureHelper.IsExactMatch(context.Identifier as SimpleNameSyntax,
-                    context.Model, context.InvokedMethodSymbol, methods);
+            return (context) => MethodSignatureHelper.IsMatch(context.Identifier as SimpleNameSyntax,
+                    context.Model, context.InvokedMethodSymbol, true, methods);
         }
 
         public override bool FirstParameterIsStringAndIsNotConstant(InvocationContext context)
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpInvocationTracker.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpInvocationTracker.cs
@@ -70,6 +70,11 @@ namespace SonarAnalyzer.Helpers
             }
 
             var strippedExpression = expression.RemoveParentheses();
+            if (strippedExpression is CastExpressionSyntax castExpression)
+            {
+                strippedExpression = castExpression.Expression.RemoveParentheses();
+            }
+
             if (strippedExpression is LiteralExpressionSyntax)
             {
                 return true;
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpInvocationTracker.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpInvocationTracker.cs
@@ -45,7 +45,7 @@ namespace SonarAnalyzer.Helpers
 
         public override InvocationCondition MatchSimpleNames(params MethodSignature[] methods)
         {
-            return (context) => MethodSignatureHelper.IsMatch(context.Identifier as SimpleNameSyntax,
+            return (context) => MethodSignatureHelper.IsExactMatch(context.Identifier as SimpleNameSyntax,
                     context.Model, context.InvokedMethodSymbol, methods);
         }
 
@@ -55,8 +55,6 @@ namespace SonarAnalyzer.Helpers
             {
                 return false;
             }
-            IMethodSymbol m;
-            
             var argumentsSyntax = (context.Invocation as InvocationExpressionSyntax)?
                 .ArgumentList?.Arguments.FirstOrDefault();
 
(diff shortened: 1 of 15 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpInvocationTracker.cs
```diff
@@ -47,7 +46,19 @@ public CSharpInvocationTracker(IAnalyzerConfiguration analysisConfiguration, Dia
         public override InvocationCondition MatchSimpleNames(params MethodSignature[] methods)
         {
             return (context) => MethodSignatureHelper.IsMatch(context.Identifier as SimpleNameSyntax,
-                    context.Model, context.InvokedMethodSymbol, methods);
+                    context.Model, context.InvokedMethodSymbol, true, methods);
+        }
+
+        public override bool FirstParameterIsStringAndIsNotConstant(InvocationContext context)
```
duncanp-sonar: New check required by the "executing XPath" rule.
I've added an _IsConstant_ extension method to the _xxxSyntaxHelper_ classes.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/KnownType.cs
```diff
@@ -392,6 +392,12 @@ internal sealed class KnownType
 
         private readonly bool isSpecialType;
         private readonly SpecialType specialType;
+
+        static internal KnownType CreateKnownTypeForTesting(string typeName)
```
duncanp-sonar: I was going to delete this because in the end I didn't need it for the tests I've created.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/SymbolHelperTest.cs
```diff
@@ -73,193 +73,120 @@ public interface IInterface
...
-
-            this.baseClassDeclaration = this.tree.GetRoot().DescendantNodes().OfType<ClassDeclarationSyntax>()
-                .First(m => m.Identifier.ValueText == "Base");
-            this.derivedClassDeclaration1 = this.tree.GetRoot().DescendantNodes().OfType<ClassDeclarationSyntax>()
-                .First(m => m.Identifier.ValueText == "Derived1");
-            this.derivedClassDeclaration2 = this.tree.GetRoot().DescendantNodes().OfType<ClassDeclarationSyntax>()
-                .First(m => m.Identifier.ValueText == "Derived2");
-            this.interfaceDeclaration = this.tree.GetRoot().DescendantNodes().OfType<InterfaceDeclarationSyntax>()
-                .First(m => m.Identifier.ValueText == "IInterface");
+            this.testCode = new SnippetCompiler(TestInput, ignoreErrors: true, language: AnalyzerLanguage.CSharp);
```
duncanp-sonar: Refactored an existing test class to use _SnippetCompiler_. No other changes to the tests, or the product code they test.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/ProjectBuilder.cs
```diff
@@ -49,10 +49,18 @@ private ProjectBuilder(Project project)
...
+            var project = parseOptions != null
+                ? Project.WithParseOptions(parseOptions)
+                : Project;
+
+            var compilation = project.GetCompilationAsync().Result;
+
+            return compilationOptions == null
+                ? compilation
+                : compilation.WithOptions(compilationOptions);
+        }
```
duncanp-sonar: Added the ability to pass in _CompilationOptions_ so the caller can specify e.g. that the assembly is a DLL (to prevent the compile warning about _main method not found_).

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SnippetCompiler.cs
```diff
@@ -0,0 +1,200 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using CSharpSyntax = Microsoft.CodeAnalysis.CSharp.Syntax;
+using VBSyntax = Microsoft.CodeAnalysis.VisualBasic.Syntax;
+using SonarAnalyzer.Common;
+using Microsoft.CodeAnalysis.VisualBasic;
+
+namespace SonarAnalyzer.UnitTest.TestFramework
+{
+    /// <summary>
+    /// Helper class compiles snippets of code on the fly
```
duncanp-sonar: Works for C# and VB.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Helpers/VisualBasicSyntaxHelper.cs
```diff
@@ -151,6 +151,17 @@ public static SimpleNameSyntax GetIdentifier(this ExpressionSyntax expression)
...
 
+        public static bool IsConstant(this ExpressionSyntax expression, SemanticModel semanticModel)
+        {
+            if (expression == null)
+            {
+                return false;
+            }
+
+            var strippedExpression = expression.RemoveParentheses();
+            return semanticModel.GetConstantValue(strippedExpression).HasValue;
```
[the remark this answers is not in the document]
duncanp-sonar: Probably. I'll try it.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/MethodSignatureHelper.cs
```diff
@@ -56,7 +56,27 @@ internal static class MethodSignatureHelper
                 {
                     return true;
                 }
+                if (checkOverriddenMethods && IsOverride(m))
+                {
```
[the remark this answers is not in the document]
duncanp-sonar: In most cases but not all. It'll behave differently if the derived class has added a method with the same name as a method in the base class:

```

  MyBase:Foo(string)

  MyClass:Foo(string)        ' IsOverride will detect this as an override...
  MyClass:Foo(string, int)  ' ... but not this.
```

I'll add another test for this case.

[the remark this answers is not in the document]
duncanp-sonar: I switched to using _DerivesOrImplements_ and added a test case to document the false-positive corner case.

PR #2083, opened 2018-11-07
[NO SQUASH] Rule S4823: Using command line arguments is security-sensitive

### LINKED ISSUE #1986: Rule S4823: Using command line arguments is security-sensitive
[RSPEC-4823](https://jira.sonarsource.com/browse/RSPEC-4823)

Using command line arguments is security-sensitive. It has led in the past to the following vulnerabilities:

valhristov:
Fix #1986


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/MethodDeclarationContext.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// Syntax and semantic information about a single method declaration
+    /// </summary>
+    public class MethodDeclarationContext
+    {
+        private readonly Compilation compilation;
+
+        public IMethodSymbol MethodSymbol { get; }
+
+        public SemanticModel GetSemanticModel(SyntaxNode syntaxNode) =>
+            compilation.GetSemanticModel(syntaxNode.SyntaxTree);
```
valhristov: Partial class support

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/MethodDeclarationTracker.cs
```diff
@@ -0,0 +1,80 @@
...
+                });
+
+            void TrackMethodDeclaration(SymbolAnalysisContext c)
+            {
+                if (IsTrackedMethod((IMethodSymbol)c.Symbol, c.Compilation))
+                {
+                    foreach (var declaration in c.Symbol.DeclaringSyntaxReferences)
+                    {
+                        var methodIdentifier = GetMethodIdentifier(declaration.GetSyntax());
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, methodIdentifier.GetLocation()));
```
duncanp-sonar: Seems odd raising the same issue in multiple locations, but since it will only happen for _partial_ methods it probably doesn't matter.

sonaranalyzer-dotnet/rspec/vbnet/S4823_vb.net.html
```diff
@@ -0,0 +1,26 @@
+<pre>
+Module Program
+    Sub Main(args As String()) ' Questionable as there is a reference to "args" in the procedure.
+        Console.WriteLine(args(0))
```
duncanp-sonar: The example might be better if the args were being used in a more obviously unsafe way e.g. passing them to another method in class. Passing them to _Console.WriteLine_ doesn't seem to be unsafe so it's easy to overlook the value of the rule i.e. the examples should maximize the risk to highlight the value of the rules. @nicolas-harraudeau-sonarsource, what do you think?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/UsingCommandLineArgumentsBase.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        protected MethodDeclarationTracker<TSyntaxKind> MethodDeclarationTracker { get; set; }
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            MethodDeclarationTracker.Track(context,
+                MethodDeclarationTracker.MatchSimpleNames("Main"),
+                MethodDeclarationTracker.IsMainMethod(),
+                MethodDeclarationTracker.ParameterAtIndexIsUsed(0));
+        }
```
duncanp-sonar: I like this way of defining rules: short, declarative, easy to understand. No doubt we could still make improvements, but I thinks it is better than the approach of having each rule directly registering callbacks against the supplied analyzer context.
Let's hope it doesn't have a negative impact on performance...

sonaranalyzer-dotnet/rspec/vbnet/S4823_vb.net.html
nicolas-harraudeau-sonarsource: @duncanp-sonar This is of course open to discussion but I would keep this topic for later. In my experience every person has a different point of view on what examples should represent. Some see it as specification of what the rule highlight, others see it as a minimalistic example showing realistic code.
For these rules I chose to show what Hotspot highlight. The point of a Security Hotspot is to ask for a review of sensitive code. It does not pinpoint vulnerabilities. Showing only vulnerabilities could make the person think that the rule is smarter than it actually is.
It is the role of the textual description to highlight the risks and explain how to avoid them. There are also links to CERT, OWASP and SANS-TOP-25 showing vulnerability examples.

valhristov: This is from the unit tests, the idea was to check if we catch more than simple identifier syntax (this is ElementAccessExpressionSyntax)

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/UsingCommandLineArgumentsBase.cs
valhristov: Perhaps these are the slowest rules we have ;)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UsingCommandLineArgumentsTest.cs
```diff
@@ -0,0 +1,119 @@
...
+{
+    static partial void Main(params string[] args) // Noncompliant
+    {
+        System.Console.WriteLine(args);
+    }
+}")
+                .AddSnippet(@"
+partial class Program1
+{
+    static partial void Main(params string[] args); // Compliant, we raise only on methods with implementation
```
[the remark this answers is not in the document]
valhristov: The partial class is declared in different files, hence there are different syntax trees and different semantic models, hence the IMethodSymbol instances are different for both parts. We don't report for the IMethodSymbol that is for the "empty" part. If the parts were in the same file I think we should have reported twice.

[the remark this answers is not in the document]
valhristov: So far I suggest leaving it like this, e.g. not hardcoding to report only for non-empty methods. We can refactor later when we decide what to do with the reporting locations...

PR #2089, opened 2018-11-08
Rule S4787: encrypting data is security-sensitive

duncanp-sonar:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/EncryptingDataBase.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace SonarAnalyzer.Rules
+{
+    public abstract class EncryptingDataBase<TSyntaxKind> : SonarDiagnosticAnalyzer
+        where TSyntaxKind : struct
+    {
+        protected const string DiagnosticId = "S4787";
+        protected const string MessageFormat = "Make sure that encrypting data is safe here.";
+
+        private static readonly ImmutableArray<KnownType> forbiddenBaseTypes =
```
duncanp-sonar: I used _ImmutableArray_ here since that is what is expected by the existing helper method _DerivesFromAny_.
I considered changing the helper method signature to use _IEnumerable_, but that would have been inconsistent with the other helper methods in the class and I didn't want to change them all.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/EncryptingData.cs
```diff
@@ -0,0 +1,107 @@
...
+            myRSA.Decrypt(data, padding);   // Noncompliant
+            myRSA.DecryptValue(data);       // Noncompliant
+
+            RSACryptoServiceProvider myRSAC = new RSACryptoServiceProvider();
+            // Review the use of any TryEncrypt/TryDecrypt and specific Encrypt/Decrypt of RSA subclasses.
+            myRSAC.Encrypt(data, false);    // Noncompliant
+            myRSAC.Decrypt(data, false);    // Noncompliant
+            int written;
+
+            // Note: TryEncrypt/TryDecrypt are only in .NET Core 2.1+
```
duncanp-sonar: See below for a hacky test for these methods.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/EncryptingData.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            // Should raise on the Try* methods added in NET Core 2.1
+            // Note: this test is cheating - we can't currently referencing the
+            // real 2.1 assemblies since the test project is targetting an older
+            // NET Framework, so we're testing against a custom subclass
+            // to which we've added the new method names.
+            customAsymProvider.TryEncrypt();        // Noncompliant
+            customAsymProvider.TryEncrypt(null);    // Noncompliant
+            customAsymProvider.TryDecrypt();        // Noncompliant
+            customAsymProvider.TryDecrypt(null);    // Noncompliant
```
duncanp-sonar: This is the hacky test workaround. I spent a bit of time trying to work out if we could easily change the framework version used in each test, but could find a simple way to do it.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/EncryptingDataBase.cs
valhristov: Also ImmutableArray<T> is a struct and changing the parameter to IEnumerable<T> would cause boxing...

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/BaseTypeTracker.cs
```diff
@@ -112,8 +113,24 @@ bool IsTrackedRelationship(SyntaxNode contextNode, SemanticModel semanticModel,
...
             };
-        
-        
+
+        internal BaseClassCondition WhenDerivesFrom(ImmutableArray<KnownType> types) =>
+            (BaseTypeContext context, out Location issueLocation) =>
+            {
+                foreach (var baseTypeNode in context.AllBaseTypeNodes)
+                {
+                    if (context.Model.GetTypeInfo(baseTypeNode).Type?.DerivesFromAny(types) ?? false)
```
valhristov: Perhaps this does not matter much with the current rule, but I think it makes sense to cache the results of GetTypeInfo in Lazy or something else, so that we don't obtain the symbol for each of the checked types

PR #2094, opened 2018-11-10
Rule S1523: Dynamic execution is security-sensitive

### LINKED ISSUE #1987: Rule S1523: Dynamically executing code is security-sensitive
[RSPEC-1523](https://jira.sonarsource.com/browse/RSPEC-1523)

Executing code dynamically is security-sensitive. It has led in the past to the following vulnerabilities:

duncanp-sonar:
Fix #1987


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DynamicExecutionBase.cs
```diff
@@ -0,0 +1,104 @@
...
+    {
+        protected const string DiagnosticId = "S1523";
+        protected const string MessageFormat = "Make sure that this dynamic injection or execution of code is safe.";
+
+        protected abstract InvocationTracker<TSyntaxKind> CreateInvocationTracker();
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            // Special case - Assembly.Load
+            var tracker1 = CreateInvocationTracker();
```
valhristov: Why do we need different tracker instances? It should work with only one I think...

PR #2096, opened 2018-11-12
Rule S4834: Controlling permissions is security-sensitive

### LINKED ISSUE #1992: Rule S4834: Controlling permissions is security-sensitive
[RSPEC-4834](https://jira.sonarsource.com/browse/RSPEC-4834)

Controlling permissions is security-sensitive. It has led in the past to the following vulnerabilities:

valhristov:
Fix #1992


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpMethodDeclarationTracker.cs
```diff
@@ -51,7 +51,8 @@ public CSharpMethodDeclarationTracker(IAnalyzerConfiguration analysisConfigurati
                 }
 
                 var methodDeclaration = context.MethodSymbol.DeclaringSyntaxReferences
-                    .Select(r => (MethodDeclarationSyntax)r.GetSyntax())
+                    .Select(r => r.GetSyntax())
+                    .OfType<BaseMethodDeclarationSyntax>()
```
valhristov: The property getters and setters are also methods, hence we filter only "normal" methods, constructors and destructors here.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/BaseTypeTracker.cs
```diff
@@ -103,7 +103,7 @@ bool IsTrackedRelationship(SyntaxNode contextNode, SemanticModel semanticModel,
             {
                 foreach(var baseTypeNode in context.AllBaseTypeNodes)
                 {
-                    if (context.Model.GetTypeInfo(baseTypeNode).Type?.DerivesFrom(type) ?? false)
+                    if (context.Model.GetTypeInfo(baseTypeNode).Type?.DerivesOrImplements(type) ?? false)
```
valhristov: I need to support interfaces here too.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/MethodDeclarationTracker.cs
```diff
@@ -35,7 +36,7 @@ protected MethodDeclarationTracker(IAnalyzerConfiguration analysisConfiguration,
         {
         }
 
-        protected abstract SyntaxToken GetMethodIdentifier(SyntaxNode methodDeclaration);
+        protected abstract SyntaxToken? GetMethodIdentifier(SyntaxNode methodDeclaration);
```
valhristov: MethodIdentifier is null when the method is not a "normal" method, e.g. a property for example

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Helpers/VisualBasicSyntaxHelper.cs
```diff
@@ -165,6 +165,22 @@ public static SimpleNameSyntax GetIdentifier(this ExpressionSyntax expression)
             }
         }
 
+        public static SyntaxToken? GetIdentifierOrDefault(this MethodBaseSyntax methodDeclaration)
```
valhristov: From the C# syntax helper

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/MethodDeclarationTracker.cs
```diff
@@ -55,20 +56,40 @@ void TrackMethodDeclaration(SymbolAnalysisContext c)
...
+                            c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, methodIdentifier.Value.GetLocation()));
+                        }
                     }
                 }
             }
 
             bool IsTrackedMethod(IMethodSymbol methodSymbol, Compilation compilation)
             {
+                if (methodSymbol.MethodKind != MethodKind.Ordinary &&
+                    methodSymbol.MethodKind != MethodKind.ReducedExtension) // Just methods
```
duncanp-sonar: Why not make these conditions? This seems like quite a big constraint to bake into the base tracker class.

Also, the _MethodKind_ enum has values for _Constructor_, _Desctructor_, _PropertyGet_, _PropertySet_ etc.  Presumably this check is filtering out all of those types so the changes above to handle those kinds of method symbols aren't required if this check is done?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/ControllingPermissionsBase.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            MethodDeclarationTracker.Track(context,
+                MethodDeclarationTracker.DecoratedWithAnyAttribute(
+                    KnownType.System_Security_Permissions_PrincipalPermissionAttribute));
+
+            BaseTypeTracker.Track(context,
+                BaseTypeTracker.WhenDerivesFrom(
+                    KnownType.System_Security_Principal_IIdentity));
+
+            BaseTypeTracker.Track(context,
```
duncanp-sonar: There is now an overload of _BaseTypeTracker.WhenDerivesFrom_ that accepts a list of types (although it might be better if it was renamed to _WhenDerivesFromAny_).

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/MethodDeclarationTracker.cs
valhristov: I added logic to obtain identifier from all method symbol kinds and extracted the method kind check as a condition.

PR #2104, opened 2018-11-14
Fix test framework bug (partial matching of NuGet package ids)

duncanp-sonar:
Partially fixed by a4c64d09. There were two locations in the code that were calling Directory.GetDirectories to matching package folders. a4c64d09 fixed one of them to correctly match package ids. Now the matching logic only exists in one place.

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/NuGetMetadataReference.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/NuGetMetadataReference.cs
@@ -24,6 +24,7 @@ using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Text.RegularExpressions;
 using Microsoft.CodeAnalysis;
 using NuGet;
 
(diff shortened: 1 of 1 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/NuGetMetadataReference.cs
```diff
@@ -163,19 +166,41 @@ private static void EnsurePackageIsInstalled(string packageId, string packageVer
...
-            // Package directory names are in the form "{package id}.{package version}".
-            // Sorting the names orders them by version.
-
             const string LastUpdateFileName = "LastCheckedForUpdate.txt";
 
-            var directory = Directory.GetDirectories(PackagesFolderRelativePath, $"{packageId}.*")
-                .OrderByDescending(name => name)
-                .FirstOrDefault();
+            var directory = GetSortedPackageFolders(packageId)
+                .LastOrDefault();
```
duncanp-sonar: FYI I was getting a test failure because I didn't have the _Microsoft.AspNetCore.2.1.6_ package installed. The test framework was doing a check to see if it needed to check for the latest version, and this method finding the _LastCheckedForUpdate.txt_ for one of the other AspNetCore packages e.g. (_Microsoft.AspNetCore.Diagnostics.2.1.1_) and deciding that it didn't need to check for the latest package -> exception because the package was not installed.

PR #2110, opened 2018-11-16
Refactoring *Tracker classes, improving quality of code

valhristov:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
```diff
@@ -0,0 +1,96 @@
...
+
+                    enabledRules = new HashSet<string>();
+
+                    var sonarLintAdditionalFile = options.AdditionalFiles
+                        .FirstOrDefault(f => ParameterLoader.ConfigurationFilePathMatchesExpected(f.Path));
+                    if (sonarLintAdditionalFile != null)
+                    {
+                        var sonarLintXml = XDocument.Load(sonarLintAdditionalFile.Path);
+
+                        var rulesInXml = sonarLintXml.Descendants("Rule")
```
Evangelink: It's probably minor but I would add a comment here to explain that we expect the `SonarLint.xml` file to contain only rules that are active. At least it there is a change later on, when reading this it will be obvious.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
```diff
@@ -0,0 +1,96 @@
...
+                            .Select(r => r.Element("Key")?.Value)
+                            .WhereNotNull();
+
+                        enabledRules.UnionWith(rulesInXml);
```
Evangelink: I don't understand why you create the hashset on line 78 and use `UnionWith` here. You could directly create the hashset here: `enabledRules = <rulesInXml_query>.ToHashSet();`

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    public class AnalyzerConfiguration
+    {
+        private static readonly Lazy<HotspotConfiguration> hotspotConfiguration =
+            new Lazy<HotspotConfiguration>(() => new HotspotConfiguration());
+
+        private static readonly Lazy<IAnalyzerConfiguration> alwaysEnabledConfiguration =
+            new Lazy<IAnalyzerConfiguration>(() => new AlwaysEnabledConfiguration());
+
+        public static IAnalyzerConfiguration Hotspot =>
```
Evangelink: Based on the usage of this property I don't really see what the lazy field brings.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Xml.Linq;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Common
+{
+    public class AnalyzerConfiguration
+    {
+        private static readonly Lazy<HotspotConfiguration> hotspotConfiguration =
+            new Lazy<HotspotConfiguration>(() => new HotspotConfiguration());
```
Evangelink: It's probably minor but as the code of the `HotspotConfiguration` relies on the class being a singleton I would hardcode the singleton in the class (even if you have the property on this outer class). It would make it obvious and avoid (potential later) wrong change.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpFieldAccessTracker.cs
```diff
@@ -44,8 +44,8 @@ public CSharpFieldAccessTracker(IAnalyzerConfiguration analyzerConfiguration, Di
         protected override GeneratedCodeRecognizer GeneratedCodeRecognizer { get; } =
             CSharp.GeneratedCodeRecognizer.Instance;
 
-        protected override SyntaxNode GetIdentifier(SyntaxNode expression) =>
-            ((ExpressionSyntax)expression).GetIdentifier();
+        protected override string GetFieldName(SyntaxNode expression) =>
+            ((ExpressionSyntax)expression).GetIdentifier()?.Identifier.ValueText;
```
Evangelink: [minor] I am playing belt and braces here but I think it would be better to do a `as` cast.

Evangelink: Or can we force the type to be `ExpressionSyntax`?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/MethodSignatureHelper.cs
```diff
@@ -20,43 +20,33 @@
...
                 if (symbolFetcher.Value == null)
                 {
                     return false; // No need to continue looping if the symbol is null
                 }
-                if (!checkOverriddenMethods && symbolFetcher.Value.ContainingType.ConstructedFrom.Is(m.ContainingType))
-                {
-                    return true;
-                }
-                if (checkOverriddenMethods && symbolFetcher.Value.ContainingType.ConstructedFrom.DerivesOrImplements(m.ContainingType))
+                var containingType = symbolFetcher.Value.ContainingType.ConstructedFrom;
```
Evangelink: `ContainingType` could be null which would throw a NRE.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SyntaxTrackers/InvocationTracker.cs
```diff
@@ -62,59 +62,63 @@ void TrackInvocationExpression(SyntaxNodeAnalysisContext c)
...
-            ParameterAtIndexIsConstant(0);
-
-        public abstract InvocationCondition MethodNameIs(string methodName);
+            ArgumentAtIndexIsConstant(0);
 
-        public abstract InvocationCondition ParameterAtIndexIsConstant(int index);
+        public abstract InvocationCondition ArgumentAtIndexIsConstant(int index);
 
-        public abstract InvocationCondition ParameterAtIndexIsString(int index, string value);
+        public abstract InvocationCondition ArgumentAtIndexIsString(int index, string value);
```
Evangelink: Maybe I am going to far but wouldn't it be better to have the method being something like `ArgumentAtIndexIsOfType<T>`?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/MemberDescriptor.cs
```diff
@@ -0,0 +1,35 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarAnalyzer.Helpers
+{
+    public class MemberDescriptor
```
Evangelink: We are not providing something meant to be reused by someone else so it's probably easier to have the type `internal` and keep members `public`.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/MemberDescriptorHelper.cs
```diff
@@ -0,0 +1,86 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+
+namespace SonarAnalyzer.Helpers
+{
+    internal static class MemberDescriptorHelper
+    {
+        public static bool IsMatch<TSymbolType>(this MemberDescriptor member, string identifierName,
```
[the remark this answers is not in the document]
Evangelink: Does it need to be an extension method?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/MemberDescriptorHelper.cs
```diff
@@ -19,22 +19,50 @@
...
+            if (identifierName == null)
+            {
+                return false;
+            }
+
+            if (identifierName != member.Name)
+            {
+                return false;
+            }
+            if (symbol.Value == null)
```
Evangelink: [cosmetic] Missing blank line before this `if`.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/ControllingPermissionsBase.cs
```diff
@@ -68,7 +72,7 @@ protected override void Initialize(SonarAnalysisContext context)
                     KnownType.System_Security_Permissions_PrincipalPermissionAttribute));
 
             BaseTypeTracker.Track(context,
-                BaseTypeTracker.WhenDerivesFrom(
+                BaseTypeTracker.MatchSubclassesOf(
```
Evangelink: I am a bit confused by the naming. The previous method has been reworded to use `WhenDerived...` and this is to `MatchSubclasses`. I'd prefer having the same kind of name in all contexts.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DeliveringDebugFeaturesInProductionBase.cs
```diff
@@ -28,14 +28,17 @@ public abstract class DeliveringDebugFeaturesInProductionBase<TSyntaxKind> : Son
         protected const string DiagnosticId = "S4507";
         protected const string MessageFormat = "Make sure this debug feature is deactivated before delivering the code in production.";
 
+        protected static readonly MemberDescriptor isDevelopmentMethod =
```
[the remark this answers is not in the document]
valhristov: Imo the idea of the rule is to protect the developers from modifying mutable static value in a generic class and being surprised that they have different instance per constructed type (e.g. with generic parameters applied). When the value is immutable, having two instances that occupy a few bytes should be minor.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
valhristov: refactored the whole file

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpFieldAccessTracker.cs
valhristov: The `expression` is the value of `context.Node`, which is the syntax representing the SyntaxKind enumerations above. Generally it should not throw "randomly", because if you add more SyntaxKind items, you will "usually" write a test...

PR #2116, opened 2018-11-20
Rule S5042: Expanding archive files is security-sensitive

valhristov:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ExpandingArchives.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            stream = entry.Open(); // Noncompliant
+
+            entry.Delete(); // Compliant, method is not tracked
+
+            fullName = entry.FullName; // Compliant, properties are not tracked
+
+            ExtractToFile(entry); // Compliant, method is not tracked
+
+            Invoke(ZipFileExtensions.ExtractToFile); // Compliant, not an invocation, but could be considered as FN
```
valhristov: Note: I could not convert that to VB

PR #2129, opened 2018-11-27
Prepare support for SonarQube 7.5

henryju:
[SonarQube 7.5 will change InputFileFilter](https://jira.sonarsource.com/browse/SONAR-11484) extension point to evaluate it at project level. As a consequence it is no more possible to access module level components. Getting Configuration will allow only to get root level properties, that won't contain protobuf report paths that are currently provided by the Scanner for MSBuild as module properties.

Waiting for a cleaner fix, here is a workaround, using deprecated ProjectBuilder to read module properties.

Note: I have not yet done a real life test.


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Protobuf/AnalyzerReport.proto
```diff
@@ -25,7 +25,7 @@ option csharp_namespace = "SonarAnalyzer.Protobuf";
 option java_package = "org.sonarsource.dotnet.protobuf";
 option java_outer_classname = "SonarAnalyzer";
 
-option optimize_for = LITE_RUNTIME;
+option optimize_for = SPEED;
```
Evangelink: Is it related to this change/PR?

henryju: Nope, except that it was preventing to load the project in Eclipse. that's why I kept it as a separate commit.

Evangelink: Do you know if it has any impact or if we can merge without fear?

henryju: The size of the generated code is a bit bigger. To me it is even safer than the previous flag, since it is the default. LITE_RUNTIME was introduce to target Android, and will be moved to a protoc plugin.

PR #2135, opened 2018-12-04
Do not report Security Hotspots as external rules when analyzing PRs and short-living branches

### LINKED ISSUE #2131: Do not report Security Hotspots as external rules when analyzing PRs and short-living branches
### Description

Issues raised by SonarC# rules are reported as "external" in PRs.

### [Example PR](https://github.com/SonarSource/sonar-scanner-msbuild/pull/625):

image

### Related information

* SonarC# Version 7.10 (build 7601)
* SonarQube DCE Version 7.5 (build 19742)

* this seems to only affect hotspot rules.
* it looks like issues are reported correctly when the branch is analysed (i.e. only a problem on PRs)

valhristov:
Fix #2131

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/SarifParserCallbackImpl.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/SarifParserCallbackImpl.java
@@ -254,7 +254,7 @@ public class SarifParserCallbackImpl implements SarifParserCallback {
   }
 
   private boolean shouldCreateExternalIssue(String ruleId) {
-    return !ignoreThirdPartyIssues && !ruleId.matches("^S\\d{3,5}$");
+    return !ignoreThirdPartyIssues && !ruleId.matches("^S\\d{3,4}$");
   }
 
   private RuleType mapRuleType(@Nullable String category, String defaultLevel) {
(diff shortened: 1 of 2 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/SarifParserCallbackImpl.java
```diff
@@ -248,11 +248,15 @@ public void onRule(String ruleId, @Nullable String shortDescription, @Nullable S
...
       .severity(mapSeverity(defaultLevel))
       .name(shortDescription != null ? shortDescription : ruleId)
-      .description(fullDescription != null ? fullDescription : null)
+      .description(fullDescription)
       .type(ruleType)
       .save();
   }
 
+  private boolean shouldCreateExternalIssue(String ruleId) {
+    return !ignoreThirdPartyIssues && !ruleId.matches("^S\\d{3,5}$");
```
valhristov: We decided to go with the simplest check - if the RuleId matches our own format we do not report it as external.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/SarifParserCallbackImpl.java
```diff
@@ -248,11 +248,15 @@ public void onRule(String ruleId, @Nullable String shortDescription, @Nullable S
       .ruleId(ruleId)
       .severity(mapSeverity(defaultLevel))
       .name(shortDescription != null ? shortDescription : ruleId)
-      .description(fullDescription != null ? fullDescription : null)
+      .description(fullDescription)
```
valhristov: Not related

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/SarifParserCallbackImpl.java
henryju: Are we confident enough that other analyzers don't use this pattern? If yes, then LGTM ;)

PR #2145, opened 2018-12-11
SonarVB: Feed metric 'executable_lines_data' when SQ >= 6.2

### LINKED ISSUE #2044: SonarVB: Feed metric 'executable_lines_data' when SQ >= 6.2
### Description

Mimic C# behavior

Evangelink:
Fix #2044


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Metrics/ExecutableLinesMetric.cs
```diff
@@ -0,0 +1,173 @@
...
+                    case SyntaxKind.SetAccessorDeclaration:
+                    case SyntaxKind.GetAccessorDeclaration:
+                        return !HasExcludedCodeAttribute((AccessorDeclarationSyntax)node, ads => ads.AttributeLists);
+
+                    default:
+                        return true;
+                }
+            }
+
+            private bool HasExcludedCodeAttribute<T>(T node, Func<T, SyntaxList<AttributeListSyntax>> getAttributeLists,
```
Evangelink: This is duplicated between C# and VB.NET but trying to extract a common method was actually more painful.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Metrics/ExecutableLinesMetric.cs
```diff
@@ -0,0 +1,173 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Metrics.CSharp
+{
+    public static class ExecutableLinesMetric
+    {
+        public static ImmutableArray<int> GetLineNumbers(SyntaxTree syntaxTree, SemanticModel semanticModel)
```
Evangelink: This method and the outer class are new, the rest is the content of the old class.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Metrics/ExecutableLinesMetric.cs
```diff
@@ -0,0 +1,182 @@
...
+            {
+                switch (node.Kind())
+                {
+                    // The following C# constructs have no equivalent in VB.NET:
+                    // - checked
+                    // - unchecked
+                    // - fixed
+                    // - unsafe
+                    // - yield break
+                    // - yield return
```
valhristov: `yield return something` should be just `Yield something` https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/statements/yield-statement

`yield break` is `Exit Function` or `Return` (same article)... I am not sure if using `Exit Function` is a good idea in general, but I guess we should handle it as executable.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Metrics/ExecutableLinesMetric.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            private bool HasExcludedCodeAttribute<T>(T node, Func<T, SyntaxList<AttributeListSyntax>> getAttributeLists,
+                bool canBePartial = false)
+                where T : SyntaxNode
+            {
+                var hasExcludeFromCodeCoverageAttribute = getAttributeLists(node)
+                    .SelectMany(attributeList => attributeList.Attributes)
+                    .Any(HasExcludedAttribute);
+
+                if (!canBePartial)
```
valhristov: I think you could add `&& !node.Modifiers.HasAny(SyntaxKind.PartialKeyword)` (the code is from the top of my head, it should be changed) here to avoid getting the symbols for each method

its/src/test/java/com/sonar/it/csharp/Tests.java
```diff
@@ -91,8 +91,7 @@ public static void runNuGet(Orchestrator orch, Path projectDir, String... argume
   }
 
   private static Path getNuGetPath(Orchestrator orch) {
-    String toolsFolder = Paths.get("tools").resolve("nuget.exe").toAbsolutePath().toString();
```
Evangelink: Location no longer exists (not related to this change) so this line is pointless.

PR #2153, opened 2018-12-13
Update S1200: Rule should ignore 'nameof()' references

### LINKED ISSUE #2123: Update S1200: Rule should ignore 'nameof()' references
### Description

Using the `nameof()` keyword operator counts against the threshold of Rule S1200. 

While it is technically true that a type used as a `nameof()` parameter will be coupled to the class you are writing, I think it is not a problematic dependency because it is only about the name of the type and not about behavior or responsibility. 

### Repro steps

Create a class and write some code containing several `nameof()`'s that point to several other classes and namespaces than the one you are writing.

### Expected behavior

The analyzer should not demand to split the class into smaller and more specialized ones.

### Actual behavior

The analyzer demands to split the class into smaller and more specialized ones.

### Known workarounds

None

### Related information

* SonarC# Version 7.8.0.7320
* Visual Studio Version 15.9.2

Evangelink:
Fix #2123

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/AvoidExcessiveClassCouplingTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/AvoidExcessiveClassCouplingTest.cs
@@ -468,7 +468,7 @@ public class Self // Compliant, attributes are not counted
         public void AvoidExcessiveClassCoupling_Nameof()
         {
             Verifier.VerifyCSharpAnalyzer(@"
-public class A // Compliant, attributes are not counted
+public class A // Compliant, types referenced by the nameof expression are not counted
 {
     public A()
     {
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/AvoidExcessiveClassCouplingTest.cs
```diff
@@ -458,6 +459,23 @@ void M1()
...
+",
+                new AvoidExcessiveClassCoupling { Threshold = 0 });
+        }
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void AvoidExcessiveClassCoupling_Nameof()
+        {
+            Verifier.VerifyCSharpAnalyzer(@"
+public class A // Compliant, attributes are not counted
```
valhristov: [minor] the comment is wrong :)

PR #2157, opened 2018-12-14
Do not warn for missing "projectOutPaths" in test projects

### LINKED ISSUE #2021: Do not warn for missing "projectOutPaths" when the property is not generated on purpose
### Description

For test projects we don't generate `projectOutPaths` on purpose and thus we should not write warnings to the output when the property is missing.

### Repro steps

Scan a solution containing a unit test project.
A warning about missing `projectOutPaths` is written to the output.

valhristov:
Fix #2021


valhristov:
We should have had a test for that. This ticket is about preventing a warning from being displayed for projects that don't generate a property.

PR #2165, opened 2018-12-18
Adding TestLocationAnalyzer to build a map between fully qualified test method names and the source files

valhristov:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Utilities/TestLocationAnalyzer.cs
```diff
@@ -0,0 +1,64 @@
...
+            collector.Visit(syntaxTree.GetRoot());
+
+            return collector.ClassDeclarations.Select(classDeclaration => semanticModel.GetDeclaredSymbol(classDeclaration));
+        }
+
+        private class ClassDeclarationCollector : CSharpSyntaxWalker
+        {
+            public List<ClassDeclarationSyntax> ClassDeclarations { get; } = new List<ClassDeclarationSyntax>();
+
+            public override void VisitClassDeclaration(ClassDeclarationSyntax node)
```
valhristov: Little optimization to not drill down into the whole tree.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TestLocationAnalyzerTest.cs
```diff
@@ -0,0 +1,407 @@
...
+                        "Tests.Diagnostics.Foo_WhenFoo_ExpectsFoo.Foo_WhenFoo_ExpectsFoo"
+                        );
+
+                    messages[2].FilePath.Should().EndWith("TestMethodShouldHaveCorrectSignature.MsTest.cs");
+                    messages[2].TestNames.Should().OnlyContain(
+                        "Tests.Diagnostics.MsTestTest.PrivateTestMethod",
+                        "Tests.Diagnostics.MsTestTest.ProtectedTestMethod",
+                        "Tests.Diagnostics.MsTestTest.InternalTestMethod",
+                        "Tests.Diagnostics.MsTestTest.AsyncTestMethod",
+                        "Tests.Diagnostics.MsTestTest.GenericTestMethod",
```
valhristov: We report all kinds of methods because XUnit can execute private, generic, etc. methods as tests... We don't distinguish between test frameworks and report all attributed methods.

PR #2166, opened 2018-12-18
[WIP] Display list of unit tests

### LINKED ISSUE #886: Update plugin: display list of unit tests
Currently once I pass path to unit test report to MSBuils sonar scanner (I'm using latest c# plugin) it gets imported in SonarQube but only as static number. If I click on it does not list any file.
unit_tests
 Is there a way to display all affected files? Especially important for failed tests.

Evangelink:
Fix #886


pavel-mikula-sonarsource:
Closing PR after closing original issue as Won't fix.
cc @andrei-epure-sonarsource

PR #2177, opened 2018-12-19
Update executable lines metric

### LINKED ISSUE #2176: Variable assignment are not counted as executable lines of code in VB
### Description

The executable lines of code metric is not giving the same result for C# and VB. This is because variable assignment are not counted.

### Repro steps

Analyze the following C# code:
```
public class Class2
{
    public void Foo(bool flag)
    {
        if (flag)
        {
            flag = true; flag = false; flag = true;
        }
    }
    private void Bar(string[] Array)
    {
        foreach (string a in Array)
        {
            foreach (string b in Array)
            {
            }
        }
    }
}
```
And this VB  code
```
Public Class Class2
    Public Sub Foo(ByVal flag _
                   As Boolean)
        If flag _
            Then
            flag = True : flag = False : flag = True
        End If
    End Sub
    Private Sub Bar(ByVal Array)
        For Each a As String In Array
            For Each b As String In Array
            Next
        Next
    End Sub
End Class
```

The VB code has 3 uncovered lines whereas the C# one has 4 uncovered lines, yet the two examples are equivalent.

### Expected behavior

Both C# and VB should have the same number of uncovered lines of code.

Evangelink:
Fix #2176

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Metrics/VisualBasicExecutableLinesMetric.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Metrics/VisualBasicExecutableLinesMetric.cs
@@ -88,6 +88,7 @@ namespace SonarAnalyzer.Metrics.VisualBasic
 
                     case SyntaxKind.EmptyStatement:
                     case SyntaxKind.ExpressionStatement:
+                    case SyntaxKind.SimpleAssignmentStatement:
 
                     case SyntaxKind.DoUntilStatement:
                     case SyntaxKind.DoWhileStatement:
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Metrics/VisualBasicExecutableLinesMetric.cs
```diff
@@ -88,6 +88,7 @@ private bool FindExecutableLines(SyntaxNode node)
 
                     case SyntaxKind.EmptyStatement:
                     case SyntaxKind.ExpressionStatement:
+                    case SyntaxKind.SimpleAssignmentStatement:
```
Evangelink: In VB.NET, `SimpleAssignmentStatement` is not wrapped within an `ExpressionStatement` like it is for C#.

PR #2178, opened 2018-12-19
Fix S122: FPs with Select case statement

### LINKED ISSUE #1475: S122 in VB.NET (Select case statement)
Issue S122 keeps appearing in "Select Case" statements in VB.NET warning to reformat the code to have only one statement per line. This issue appears for all types of variables, booleans, strings, integers, everything.
I am attaching the enum case

enum

SonarQube: 5.6.1
Visual Studio: 15.7.2

Evangelink:
Fix #1475


sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/SingleStatementPerLine.cs
```diff
@@ -79,7 +78,8 @@ private static bool StatementIsSingleInLambda(StatementSyntax st)
             st is UsingBlockSyntax ||
             st is WithBlockSyntax ||
             st is MethodBaseSyntax ||
-            st is InheritsOrImplementsStatementSyntax;
+            st is InheritsOrImplementsStatementSyntax ||
+            st is SelectBlockSyntax;
```
Evangelink: This is the actual change. The rest is only about readability.

PR #2180, opened 2018-12-20
Fix S4144: FPs for semantically different methods

### LINKED ISSUE #2015: Fix S4144: FPs for semantically different methods
### Description

`S4144` is reported when implementations are identical textually but not semantically equivalent (in other words, no code sharing between them is possible). 

### Repro steps

```cs
public class Foo 
{
    public void Test(string str)
    {
        Console.WriteLine(str);
    }
}

public class Bar
{
    public void Test(string str)
    {
        throw new Exception(str);
    }
}

public static void TestFoo(Foo x)
{
    x.Test("hello");
    x.Test("world");
}

public static void TestBar(Bar x) // <= S4144 
{
    x.Test("hello");
    x.Test("world");
}
```

### Expected behavior

`S4144` should not be triggered because `Foo` and `Bar` are unrelated classes and calling `Test` on each has completely different meaning.

### Actual behavior

`S4144` is reported.

### Known workarounds

Changing the name of the variable so that the method is not textually identical e.g. `public static void TestBar(Bar bar)`.

### Related information

* 7.7.0.7192 (latest at the time of writing)
* VS 2017 (15.8.7)

Evangelink:
Fix #2015


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldNotHaveIdenticalImplementations.cs
```diff
@@ -53,7 +53,20 @@ protected override bool AreDuplicates(MethodDeclarationSyntax firstMethod, Metho
                 secondMethod.Body != null &&
                 firstMethod.Body.Statements.Count >= 2 &&
                 firstMethod.Identifier.ValueText != secondMethod.Identifier.ValueText &&
+                HaveSameParameters(firstMethod.ParameterList?.Parameters, secondMethod.ParameterList?.Parameters) &&
                 firstMethod.Body.IsEquivalentTo(secondMethod.Body, false);
+
+            bool HaveSameParameters(SeparatedSyntaxList<ParameterSyntax>? leftParameters, SeparatedSyntaxList<ParameterSyntax>? rightParameters)
```
Evangelink: C# and VB.NET methods are exactly the same BUT the `ParameterSyntax` object is not coming from the same dll so we cannot share the code.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodsShouldNotHaveIdenticalImplementations.cs
```diff
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
```
Evangelink: I don't know why there is this change.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodsShouldNotHaveIdenticalImplementations.cs
```diff
@@ -96,5 +94,33 @@ string Qux2(int val)
...
+        {
+            x.Test("hello");
+            x.Test("world");
+        }
+
+        public static void TestBar(Bar x)
+        {
+            x.Test("hello");
+            x.Test("world");
+        }
```
valhristov: Please, write tests with more than one argument and tests with different number of arguments.

PR #2183, opened 2018-12-20
TypeHelper.IsMatch should check OriginalDefinition as well to work wi…

### LINKED ISSUE #2132: TypeHelper.IsMatch should check OriginalDefinition as well to work with generics
[TypeHelper.IsMatch](https://github.com/SonarSource/sonar-dotnet/blob/5e4c6988273ab51d1974dce014b2feaeafeebb1d/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/TypeHelper.cs#L60) checks only the typeSymbol.

However, this won't work for generic collections for which the `typeSymbol.SpecialType` contains the concrete type - e.g. `List<string>`, whereas the generic type definition is `List<T>`.

To fix this, the `OriginalDefinition.SpecialType` should be compared as well, like below.

        return type.Matches(typeSymbol.SpecialType) ||
                type.Matches(typeSymbol.OriginalDefinition.SpecialType) ||
                type.Matches(typeSymbol.ToDisplayString()) ||
                type.Matches(typeSymbol.OriginalDefinition.ToDisplayString());
        }

Evangelink:
…th generics

Fix #2132

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/TypeHelper.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/TypeHelper.cs
@@ -59,7 +59,10 @@ namespace SonarAnalyzer.Helpers
 
         private static bool IsMatch(ITypeSymbol typeSymbol, KnownType type)
         {
-            return type.Matches(typeSymbol.SpecialType) || type.Matches(typeSymbol.ToDisplayString());
+            return type.Matches(typeSymbol.SpecialType) ||
+                type.Matches(typeSymbol.OriginalDefinition.SpecialType) ||
+                type.Matches(typeSymbol.ToDisplayString()) ||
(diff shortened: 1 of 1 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/TypeHelper.cs
```diff
@@ -59,7 +59,10 @@ public static bool Is(this ITypeSymbol self, TypeKind typeKind)
 
         private static bool IsMatch(ITypeSymbol typeSymbol, KnownType type)
         {
-            return type.Matches(typeSymbol.SpecialType) || type.Matches(typeSymbol.ToDisplayString());
+            return type.Matches(typeSymbol.SpecialType) ||
+                type.Matches(typeSymbol.OriginalDefinition.SpecialType) ||
```
valhristov: Are any of the "special" types generic?

PR #2197, opened 2019-01-02
Fix S1192: Rule should display string as defined in the source code

### LINKED ISSUE #2191: Fix S1192: Rule should display string as defined in the source code
### Description

Analyzing Lucene.Net fails when reporting the issue about duplicated string for the char \u0000 on file [HTMLStripCharFilter.cs](https://raw.githubusercontent.com/apache/lucenenet/a3a12967b250e8e7e5f623f0ba7572ec64f479ac/src/Lucene.Net.Analysis.Common/Analysis/CharFilter/HTMLStripCharFilter.cs)

See scanner end step logs https://cirrus-ci.com/task/5111004193095680

Evangelink:
Fix #2191

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
@@ -77,7 +77,8 @@ namespace SonarAnalyzer.Rules.CSharp
             var stringWithLiterals = new Dictionary<string, List<LiteralExpressionSyntax>>();
             foreach (var literal in stringLiterals)
             {
-                var stringValue = literal.Token.ValueText;
+                // Remove leading and trailing double quotes
+                var stringValue = ExtractStringContent(literal.Token.Text);
 
                 if (stringValue != null &&
                     stringValue.Length >= MinimumStringLength &&
@@ -104,11 +105,23 @@ namespace SonarAnalyzer.Rules.CSharp
             }
         }
 
-        private bool IsMatchingMethodParameterName(LiteralExpressionSyntax literalExpression) =>
+        private static bool IsMatchingMethodParameterName(LiteralExpressionSyntax literalExpression) =>
             literalExpression.FirstAncestorOrSelf<BaseMethodDeclarationSyntax>()
                 ?.ParameterList
                 ?.Parameters
                 .Any(p => p.Identifier.ValueText == literalExpression.Token.ValueText)
                 ?? false;
+
+        private static string ExtractStringContent(string literal)
+        {
+            if (literal.StartsWith("@\""))
+            {
+                return literal.Substring(2, literal.Length - 3);
+            }
+            else
+            {
+                return literal.Substring(1, literal.Length - 2);
+            }
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
```diff
@@ -74,21 +74,22 @@ private void ReportOnViolation(SyntaxNodeAnalysisContext context)
                 .Cast<LiteralExpressionSyntax>();
 
             // Collect duplications
-            var stringWithLiterals = new Dictionary<string, List<LiteralExpressionSyntax>>();
+            var stringWithLiterals = new Dictionary<string, List<Location>>();
```
Evangelink: We only use the `Location` so there is no need to store the full object.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
```diff
@@ -74,21 +74,22 @@ private void ReportOnViolation(SyntaxNodeAnalysisContext context)
...
             foreach (var literal in stringLiterals)
             {
-                var stringValue = literal.Token.ValueText;
+                // Remove leading and trailing double quotes
+                var stringValue = literal.Token.Text.Substring(1, literal.Token.Text.Length - 2);
```
Evangelink: Actual change.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StringLiteralShouldNotBeDuplicated.cs
```diff
@@ -4,6 +4,7 @@ namespace Tests.Diagnostics
 {
     public class Program
     {
+        private string empty = "";
```
Evangelink: This is to ensure that the substring doesn't fail for an empty string.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
valhristov: Does it work with strings like `@"asdasd"` or `$@"adasdasd"`? (not sure if those are different syntax kinds)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
```diff
@@ -74,21 +74,22 @@ private void ReportOnViolation(SyntaxNodeAnalysisContext context)
...
                     !IsMatchingMethodParameterName(literal))
                 {
                     if (!stringWithLiterals.ContainsKey(stringValue))
                     {
-                        stringWithLiterals[stringValue] = new List<LiteralExpressionSyntax>();
+                        stringWithLiterals[stringValue] = new List<Location>();
                     }
 
-                    stringWithLiterals[stringValue].Add(literal);
+                    stringWithLiterals[stringValue].Add(literal.GetLocation());
```
valhristov: I think `GetLocation` is generally slow

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
Evangelink: `$@""` and `$""` are not `StringLiteralExpression` and so not handled by this rule.

PR #2198, opened 2019-01-02
Fix S3937: FP for numbers with type suffixes

### LINKED ISSUE #2196: Fix S3937: FP for numbers with type suffixes (i.e. 1_000_000UL)
### Description

All numbers with literal suffix and thousands separator cause false positive S3937 error.

### Repro steps

Compile the code:
ulong a = 1_000_000UL;

### Expected behavior

There should be no errors.

### Actual behavior

I'm getting the following error message:
S3937 Review this number; its irregular pattern indicates an error.
The use of punctuation characters to separate subgroups in a number can make the number more readable. For instance consider 1,000,000,000 versus 1000000000. But when the grouping is irregular, such as 1,000,00,000; it indicates an error. 

### Related information

* SonarC# Version 7.9.1.7622
* Visual Studio Version 2017 15.9.4

Evangelink:
Fix #2196

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -73,5 +75,21 @@ protected override void Initialize(SonarAnalysisContext context)
                 },
                 SyntaxKind.NumericLiteralExpression);
         }
+
+        private static string ClearNumberTypeSuffix(string numberLiteral)
+        {
+            if (numberLiteral.EndsWith("UL"))
```
valhristov: the string checks should be case insensitive (in the if below too)
- integers - https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#integer-literals
- real numbers - https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#real-literals

PR #2199, opened 2019-01-03
Fix S2699: Rule doesn't raise when the code contains any invocation

### LINKED ISSUE #2130: Fix S2699: Rule doesn't raise when the code contains any invocation
### Description

Since the tests refactoring the rule is broken and no longer raises any issue on methods containing invocations.

### Repro steps

```csharp
namespace UnitTestProject1
{
    public class Class1
    {
        public int Add(int a, int b) => a + b;
    }

    [TestClass]
    public class UnitTest1
    {
        [TestMethod]
        public void TestMethod1() // No issue raised
        {
            var result = new Class1().Add(1, 2);
        }
    }
}
```

Evangelink:
Fix #2130

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
@@ -63,7 +63,7 @@ namespace SonarAnalyzer.Rules.CSharp
 
                     var hasAnyAssertion = methodDeclaration.DescendantNodes()
                         .OfType<InvocationExpressionSyntax>()
-                        .Any(invocation => IsAssertion(invocation, c.SemanticModel));
+                        .Any(invocation => IsAssertion(invocation));
                     if (!hasAnyAssertion)
                     {
                         c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, methodDeclaration.Identifier.GetLocation()));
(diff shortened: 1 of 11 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
```diff
@@ -87,19 +87,11 @@ public static bool IsTestIgnored(IMethodSymbol method)
                 factAttributeSyntax.ArgumentList.Arguments.Any(x => x.NameEquals.Name.Identifier.ValueText == "Skip");
         }
 
-        private static bool IsAssertion(InvocationExpressionSyntax invocation, SemanticModel model)
```
Evangelink: The behavior of this method was incorrect so I rewrote it.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
```diff
@@ -77,16 +77,16 @@ internal static class UnitTestHelper
...
         /// List of partial names that are assumed to indicate an assertion method
         /// </summary>
-        private static readonly IEnumerable<string> AssertionPartialMethodName =
-            new List<string>
-            {
+        public static readonly ImmutableArray<string> KnownAssertionMethodParts =
+            ImmutableArray.Create(
                 "ASSERT",
-                "SHOULD",
+                "CHECK",
```
Evangelink: Add `CHECK` to reduce number of FPs.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
```diff
@@ -108,9 +108,6 @@ internal static class UnitTestHelper
         public static bool HasExpectedExceptionAttribute(this IMethodSymbol method) =>
             method.GetAttributes().Any(a => a.AttributeClass.IsAny(KnownExpectedExceptionAttributes));
 
-        public static bool IsAssertionMethodName(string name) =>
-            name.SplitCamelCaseToWords().Union(AssertionPartialMethodName).Any();
```
Evangelink: This method was doing an `Union` not an `Intersect` and so ALL methods were considered correct.

PR #2201, opened 2019-01-03
Update S2436: Rule should be syntax based and not symbol based

### LINKED ISSUE #2202: Update S2436: Rule should handle struct and interface types
### Description

Currently, the rule detects classes and methods with too many generic args. We could extend the rule to cover `struct` and `interface`.

Evangelink:
Fix #2202

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TooManyGenericParameters_DefaultValues.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TooManyGenericParameters_DefaultValues.cs
@@ -55,3 +55,21 @@ namespace Tests.Diagnostics
     {
     }
 }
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TooManyGenericParameters_DefaultValues.cs
```diff
@@ -55,3 +55,21 @@ class Bar<T1, T2, T3, T4, T5> // Noncompliant
...
+    {
+
+    }
+}
+
+namespace TheProject
+{
+    using MyLib;
+
+    public class Impl : FrameworkBaseClass<int, double, bool>
```
Evangelink: Adding a new test case to ensure we report only on the type definition.

PR #2251, opened 2019-01-29
Rule S107: Functions should not have too many parameters

### LINKED ISSUE #2058: Rule S107: Functions should not have too many parameters
### Description

Implements https://jira.sonarsource.com/browse/RSPEC-107

andrei-epure-sonarsource:
Fix #2058


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/TooManyParametersBase.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                GeneratedCodeRecognizer,
+                c =>
+                {
+                    var parameterListNode = (TParameterListSyntax)c.Node;
+                    var parametersCount = CountParameters(parameterListNode);
+
+                    if (parametersCount > Maximum &&
+                        parameterListNode.Parent != null &&
```
Evangelink: What's the goal of this check? Is it purely defensive or are you trying to handle some known use-case?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/TooManyParametersBase.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace SonarAnalyzer.Rules
+{
+    public abstract class TooManyParametersBase<TSyntaxKind, TParameterListSyntax> : ParameterLoadingDiagnosticAnalyzer
+        where TSyntaxKind : struct
+        where TParameterListSyntax : SyntaxNode
+    {
+        protected const string DiagnosticId = "S107";
+        protected const string MessageFormat = "{2} has {1} parameters, which is greater than the {0} authorized.";
+
+        protected abstract DiagnosticDescriptor Rule { get; }
```
Evangelink: Un-needed.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/TooManyParametersBase.cs
```diff
@@ -0,0 +1,69 @@
...
+                {
+                    var parameterListNode = (TParameterListSyntax)c.Node;
+                    var parametersCount = CountParameters(parameterListNode);
+
+                    if (parametersCount > Maximum &&
+                        parameterListNode.Parent != null &&
+                        CanBeChanged(parameterListNode.Parent, c.SemanticModel) &&
+                        Mapping.TryGetValue(ParentType(parameterListNode), out var declarationName))
+                    {
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, parameterListNode.GetLocation(),
```
Evangelink: Use `SupportedDiagnostics[0]` here instead of `Rule`.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/TooManyParametersBase.cs
```diff
@@ -0,0 +1,69 @@
...
+        where TSyntaxKind : struct
+        where TParameterListSyntax : SyntaxNode
+    {
+        protected const string DiagnosticId = "S107";
+        protected const string MessageFormat = "{2} has {1} parameters, which is greater than the {0} authorized.";
+
+        protected abstract DiagnosticDescriptor Rule { get; }
+        protected abstract TSyntaxKind[] SyntaxKinds { get; }
+        protected abstract GeneratedCodeRecognizer GeneratedCodeRecognizer { get; }
+        protected abstract Dictionary<TSyntaxKind, string> Mapping { get; }
```
Evangelink: Mapping is a really vague name. I had to look at the code to understand what it is used for.

I'd suggest `NodeToDeclarationName` instead.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
```diff
@@ -0,0 +1,118 @@
...
+                return false;
+            }
+
+            if (symbol != null)
+            {
+                // Not a declaration, such as Action
+                return true;
+            }
+
+            if ((node as SubNewStatementSyntax)?.ParameterList?.Parameters.Count is int parameterCount &&
```
Evangelink: Except for this part, which uses the syntax, all the rest is relying on symbol and is the exact same code for C# and VB.NET. I think it should be extracted into the base class.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/TooManyParametersBase.cs
Evangelink: Also, I think that it should be of type `ImmutableDictionary`.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/TooManyParametersBase.cs
```diff
@@ -0,0 +1,69 @@
...
+        protected abstract TSyntaxKind ParentType(TParameterListSyntax parameterList);
```
Evangelink: Please add an empty line between the `abstract` properties and methods.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
```diff
@@ -0,0 +1,118 @@
...
+using SonarAnalyzer.Helpers.VisualBasic;
+
+namespace SonarAnalyzer.Rules.VisualBasic
+{
+    [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
+    [Rule(DiagnosticId)]
+    public sealed class TooManyParameters : TooManyParametersBase<SyntaxKind, ParameterListSyntax>
+    {
+        protected override GeneratedCodeRecognizer GeneratedCodeRecognizer => VisualBasicGeneratedCodeRecognizer.Instance;
+        protected override SyntaxKind[] SyntaxKinds => new SyntaxKind[] { SyntaxKind.ParameterList };
```
Evangelink: Please either:

- create a `static readonly` field containing this array and return it with the arrow

- change the `=>` to a `{ get; } =`

The problem with the arrowed body is that every time this is called a new array is created.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/TooManyParametersBase.cs
```diff
@@ -0,0 +1,69 @@
...
+                GeneratedCodeRecognizer,
+                c =>
+                {
+                    var parameterListNode = (TParameterListSyntax)c.Node;
+                    var parametersCount = CountParameters(parameterListNode);
+
+                    if (parametersCount > Maximum &&
+                        parameterListNode.Parent != null &&
+                        CanBeChanged(parameterListNode.Parent, c.SemanticModel) &&
+                        Mapping.TryGetValue(ParentType(parameterListNode), out var declarationName))
```
Evangelink: The `ParentType` method is used only here and the `Mapping` property too. I would suggest to get rid of the 2 and to declare a `GetParentDeclarationName(TParameterListSyntax parameterList)` method.

Evangelink: Also, I think that it is a bit confusing to have this last call part of the `if`. Indeed, all the rest is to check that there is a violation and this one is there to say there is a violation but I code defensive in case I forgot to handle one case.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
```diff
@@ -0,0 +1,118 @@
...
+                // P/Invoke method is defined externally.
+                // Do not raise
+                return false;
+            }
+
+            return declaredSymbol.GetOverriddenMember() == null &&
+                   declaredSymbol.GetInterfaceMember() == null;
+        }
+
+        private static readonly DiagnosticDescriptor rule =
```
Evangelink: Move this one upper in the class so that C# and VB.NET classes are alike in term of ordering.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
```diff
@@ -0,0 +1,118 @@
...
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager,
+                isEnabledByDefault: false);
+
+        private static bool ContainsMyBaseNewInvocation(ConstructorBlockSyntax constructorBlock, int maximum) =>
+                constructorBlock.Statements.Any(s => s is ExpressionStatementSyntax expression &&
+                    expression.Expression is InvocationExpressionSyntax invocation &&
```
Evangelink: No need for this indentation for this and the other statements.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TooManyParameters.vb
```diff
@@ -0,0 +1,110 @@
...
+'               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  End Sub
+
+  Public Sub F1(ByVal p1 As Integer, ByVal p2 As Integer, ByVal p3 As Integer) Implements MyInterface.F1
+  End Sub
+
+  Public Sub F2(ByVal p1 As Integer, ByVal p2 As Integer, ByVal p3 As Integer, ByVal p4 As Integer)  Implements MyInterface.F2 ' Compliant, interface implementation
+  End Sub
+
+  Public Sub F1(ByVal p1 As Integer, ByVal p2 As Integer, ByVal p3 As Integer, ByVal p4 As Integer, ByVal p5 As String) ' Noncompliant {{Sub has 5 parameters, which is greater than the 3 authorized.}}
```
Evangelink: Please add some tests with `ByRef`

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/TooManyParametersBase.cs
andrei-epure-sonarsource: ok... I remember it's a pattern we use but it's quite cryptic

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
andrei-epure-sonarsource: all expressions are an argument for the `Any` call, I prefer to indent to make it clear

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/TooManyParametersBase.cs
andrei-epure-sonarsource: Ok, so I should do a check if the syntax kind is in the dictionary inside `CanBeChanged`

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
andrei-epure-sonarsource: good catch!

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/TooManyParametersBase.cs
andrei-epure-sonarsource: it's defensive

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
```diff
@@ -35,70 +35,39 @@ namespace SonarAnalyzer.Rules.VisualBasic
...
                 node.Parent is ConstructorBlockSyntax constructorBlock &&
                 ContainsMyBaseNewInvocation(constructorBlock, Maximum))
             {
                 return false;
             }
-
-            if (declaredSymbol.IsExtern &&
-                declaredSymbol.IsStatic &&
-                declaredSymbol.GetAttributes(KnownType.System_Runtime_InteropServices_DllImportAttribute).Any())
+            if (node.IsAnyKind(LambdaHeaders))
```
andrei-epure-sonarsource: defensive programming (not covered by tests)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TooManyParameters.cs
```diff
@@ -20,69 +20,34 @@
...
     {
-        internal const string DiagnosticId = "S107";
-        private const string MessageFormat = "{2} has {1} parameters, which is greater than the {0} authorized.";
-
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager,
-                isEnabledByDefault: false);
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        protected override GeneratedCodeRecognizer GeneratedCodeRecognizer { get; } = CSharpGeneratedCodeRecognizer.Instance;
+        protected override SyntaxKind[] SyntaxKinds { get; } = new SyntaxKind[] { SyntaxKind.ParameterList };
```
Evangelink: [cosmetic] the type before the array creation is not needed.
```suggestion
        protected override SyntaxKind[] SyntaxKinds { get; } = new[] { SyntaxKind.ParameterList };
```

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        protected override int CountParameters(ParameterListSyntax parameterList) => parameterList.Parameters.Count;
+
+        protected override bool CanBeChanged(SyntaxNode node, SemanticModel semanticModel)
+        {
+            if (!nodeToDeclarationName.ContainsKey(node.Kind()))
+            {
+                return false;
+            }
+            if ((node as SubNewStatementSyntax)?.ParameterList?.Parameters.Count is int parameterCount &&
```
Evangelink: [cosmetic] Add an empty line between the ifs.
```suggestion

            if ((node as SubNewStatementSyntax)?.ParameterList?.Parameters.Count is int parameterCount &&
```

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
```diff
@@ -0,0 +1,95 @@
...
+                parameterCount > Maximum &&
+                node.Parent is ConstructorBlockSyntax constructorBlock &&
+                ContainsMyBaseNewInvocation(constructorBlock, Maximum))
+            {
+                return false;
+            }
+            if (node.IsAnyKind(LambdaHeaders))
```
Evangelink: [cosmetic] Add an empty line between the ifs
```suggestion

            if (node.IsAnyKind(LambdaHeaders))
```

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                return true;
+            }
+            return VerifyCanBeChangedBySymbol(node, semanticModel);
```
Evangelink: [cosmetic] Add an empty line before this return
```suggestion

            return VerifyCanBeChangedBySymbol(node, semanticModel);
```

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
Evangelink: This comment wasn't fixed. We usually put the fields on top of the class (at least before method definitions).

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        private static readonly DiagnosticDescriptor rule =
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager,
+                isEnabledByDefault: false);
+
+        private static readonly ImmutableDictionary<SyntaxKind, string> nodeToDeclarationName = new Dictionary<SyntaxKind, string>
```
Evangelink: Same here.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
```diff
@@ -0,0 +1,95 @@
...
+            { SyntaxKind.SubNewStatement, "Constructor" },
+            { SyntaxKind.FunctionStatement, "Function" },
+            { SyntaxKind.SubStatement, "Sub" },
+            { SyntaxKind.DelegateFunctionStatement, "Delegate" },
+            { SyntaxKind.DelegateSubStatement, "Delegate" },
+            { SyntaxKind.SubLambdaHeader, "Lambda" },
+            { SyntaxKind.FunctionLambdaHeader, "Lambda" },
+        }.ToImmutableDictionary();
+
+        private static readonly SyntaxKind[] LambdaHeaders = new SyntaxKind[]
```
Evangelink: Same here.

Evangelink: You could simplify the array creation.
```suggestion
        private static readonly SyntaxKind[] LambdaHeaders = new[]
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TooManyParametersTest.cs
```diff
@@ -29,10 +31,19 @@ public class TooManyParametersTest
     {
         [TestMethod]
         [TestCategory("Rule")]
-        public void TooManyParameters()
+        public void TooManyParameters_CS()
         {
-            var diagnostic = new TooManyParameters { Maximum = 3 };
-            Verifier.VerifyAnalyzer(@"TestCases\TooManyParameters.cs", diagnostic);
+            Verifier.VerifyAnalyzer(@"TestCases\TooManyParameters.cs",
+                new CSharp.TooManyParameters { Maximum = 3 });
```
Evangelink: It's not always applied but I like to have one `_DefaultValue` test and one `_CustomValue` test. This allows avoiding un-seen regression if the default value changes.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TooManyParameters.vb
Evangelink: I was actually meaning that it would be nice to have a failing test with `ByRef`. See suggestion.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TooManyParameters.vb
```diff
@@ -0,0 +1,123 @@
...
```
Evangelink:
```suggestion
  Public Sub F1(ByVal p1 As Integer, ByVal p2 As Integer, ByRef p3 As Integer, ByVal p4 As Integer, ByVal p5 As String) ' Noncompliant {{Sub has 5 parameters, which is greater than the 3 authorized.}}
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TooManyParameters.vb
andrei-epure-sonarsource: I added tests with `ByRef` at lines 87 and 88 (the issues are raised on the interface definitions)

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/TooManyParameters.cs
andrei-epure-sonarsource: Ok. I think all these conventions have to be formalized. Smth like:
- Properties sorted by visibility
- Fields sorted by visibility
- Methods sorted by visibility

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TooManyParameters.vb
Evangelink: I missed them! I cannot find any mixing though.

PR #2256, opened 2019-01-30
Rule S1151: 'Select...Case' clauses should not have too many lines of…

### LINKED ISSUE #2060: Rule S1151: 'Select...Case' clauses should not have too many lines of code
### Description

Implements https://jira.sonarsource.com/browse/RSPEC-1151

Evangelink:
… code

Fix #2060

### DIFF
--- a/sonaranalyzer-dotnet/rspec/vbnet/S1151_vb.net.html
+++ b/sonaranalyzer-dotnet/rspec/vbnet/S1151_vb.net.html
@@ -1,6 +1,6 @@
-<p>The <code>switch</code> statement should be used only to clearly define some new branches in the control flow. As soon as a <code>case</code>
-clause contains too many statements this highly decreases the readability of the overall control flow statement. In such case, the content of the
-<code>case</code> clause should be extracted into a dedicated method.</p>
+<p>The <code>Select...Case</code> statement should be used only to clearly define some new branches in the control flow. As soon as a case clause
+contains too many statements this highly decreases the readability of the overall control flow statement. In such case, the content of the case clause
+should be extracted into a dedicated procedure.</p>
 <h2>Noncompliant Code Example</h2>
 <p>With the default threshold of 3:</p>
 <pre>
--- a/sonaranalyzer-dotnet/rspec/vbnet/S1151_vb.net.html
+++ b/sonaranalyzer-dotnet/rspec/vbnet/S1151_vb.net.html
@@ -4,35 +4,30 @@ clause contains too many statements this highly decreases the readability of the
 <h2>Noncompliant Code Example</h2>
 <p>With the default threshold of 3:</p>
 <pre>
-switch (myVariable)
-{
-    case 0: // Noncompliant: 5 statements in the case
-        methodCall1("");
-        methodCall2("");
-        methodCall3("");
-        methodCall4("");
-        break;
-    case 1:
-        ...
-}
+Select Case number
+    Case 1 To 5 // Noncompliant: 4 statements in the case
+        MethodCall1("")
+        MethodCall2("")
+        MethodCall3("")
+        MethodCall4("")
+    Case Else
+        ' ...
+End Select
 </pre>
 <h2>Compliant Solution</h2>
 <pre>
-switch (myVariable)
-{
-    case 0:
+Select Case number
+    Case 1 To 5
         DoSomething()
-        break;
-    case 1:
-        ...
-}
+    Case Else
+        ' ...
+End Select
 ...
-private void DoSomething()
-{
-    methodCall1("");
-    methodCall2("");
-    methodCall3("");
-    methodCall4("");
-}
+Sub DoSomething()
+    MethodCall1("")
+    MethodCall2("")
+    MethodCall3("")
+    MethodCall4("")
+End Sub
 </pre>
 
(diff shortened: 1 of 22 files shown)


sonaranalyzer-dotnet/rspec/vbnet/S1151_vb.net.html
```diff
@@ -0,0 +1,33 @@
+<p>The <code>switch</code> statement should be used only to clearly define some new branches in the control flow. As soon as a <code>case</code>
```
andrei-epure-sonarsource: s/switch/Select

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/SwitchSectionShouldNotHaveTooManyStatementsBase.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using SonarAnalyzer.Common;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules
+{
+    public abstract class SwitchSectionShouldNotHaveTooManyStatementsBase : ParameterLoadingDiagnosticAnalyzer
+    {
+        protected const string DiagnosticId = "S1151";
+        protected const string MessageFormat = "Reduce this {2} number of statements from {0} to at most {1}, for example by extracting code into a {3}.";
```
andrei-epure-sonarsource: nitpick: would be better to have {0} {1} {2} {3} inside the message (lower brain overload)

sonaranalyzer-dotnet/src/SonarAnalyzer.Utilities/Rules.Description/S1151_vb.html
```diff
@@ -0,0 +1,33 @@
+<p>The <code>switch</code> statement should be used only to clearly define some new branches in the control flow. As soon as a <code>case</code>
```
andrei-epure-sonarsource: Select instead of switch

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SwitchSectionShouldNotHaveTooManyStatements_DefaultValue.vb
```diff
@@ -0,0 +1,48 @@
...
+                    Test()
+                    Test()
+                    Test()
+                    Test()
+                    Test()
+                    Test()
+                    Test()
+                    Test()
+                    Test()
+                Case 6 ' Noncompliant {{Reduce this 'Case' block number of statements from 12 to at most 8, for example by extracting code into a procedure.}}
```
andrei-epure-sonarsource: I would add comments below to specify what is ignored and what is not (to be easier to verify the number)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SwitchSectionShouldNotHaveTooManyStatements_DefaultValue.vb
```diff
@@ -0,0 +1,48 @@
...
+                    If False
```
andrei-epure-sonarsource: You should add coverage for the following as well
                    case SyntaxKind.CatchBlock:
                    case SyntaxKind.FinallyBlock:
                    case SyntaxKind.ForEachBlock:
                    case SyntaxKind.SyncLockBlock:
                    case SyntaxKind.TryBlock:
                    case SyntaxKind.UsingBlock:
                    case SyntaxKind.WithBlock:
                    case SyntaxKind.EndSyncLockStatement:
                    case SyntaxKind.EndTryStatement:
                    case SyntaxKind.EndUsingStatement:
                    case SyntaxKind.EndWithStatement:

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/SwitchSectionShouldNotHaveTooManyStatements.cs
```diff
@@ -0,0 +1,123 @@
...
+                    {
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, caseBlock.CaseStatement.GetLocation(),
+                            statementsCount, Threshold, "'Case' block", "procedure"));
+                    }
+                },
+                SyntaxKind.CaseBlock,
+                SyntaxKind.CaseElseBlock);
+        }
+
+        private IEnumerable<StatementSyntax> GetSubStatements(StatementSyntax statement)
```
andrei-epure-sonarsource: Given that `Sub` is the name of the VB 'void' procedure, I'd rename this to `GetInnerStatements`

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/SwitchSectionShouldNotHaveTooManyStatementsBase.cs
Evangelink: That's what I did first but it was looking less nice at the place where we pass the parameters: string, int, int, string.

andrei-epure-sonarsource: ok, feel free to leave it as is

PR #2260, opened 2019-02-04
Rule S138: Procedures should not have too many lines of code

### LINKED ISSUE #2051: Rule S138: Rule: Functions should not have too many lines of code
### Description

Implements https://jira.sonarsource.com/browse/RSPEC-138

Evangelink:
Fix #2051


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodsShouldNotHaveTooManyLines_CustomValues.vb
```diff
@@ -0,0 +1,54 @@
+﻿Imports System
+Imports System.Collections.Generic
+Imports System.Linq
+Imports System.Text
+
+Namespace Tests.TestCases
+    Class Foo
+        Public Sub Test() ' Noncompliant {{This method 'Test' has 6 lines, which is greater than the 2 lines authorized. Split it into smaller procedures.}}
```
andrei-epure-sonarsource: should add test with 2 lines as well

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/MethodsShouldNotHaveTooManyLinesTest.cs
```diff
@@ -29,26 +29,51 @@ public class MethodsShouldNotHaveTooManyLinesTest
...
+            Verifier.VerifyNoIssueReportedInTest(@"TestCases\MethodsShouldNotHaveTooManyLines_DefaultValues.cs",
                 new MethodsShouldNotHaveTooManyLines());
         }
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void MethodsShouldNotHaveTooManyLines_DefaultValues_VB()
+        {
+            Verifier.VerifyAnalyzer(@"TestCases\MethodsShouldNotHaveTooManyLines_DefaultValues.vb",
+                new SonarAnalyzer.Rules.VisualBasic.MethodsShouldNotHaveTooManyLines());
```
andrei-epure-sonarsource: extract namespace in nice and pretty

```
using CS = SonarAnalyzer.Rules.CSharp;
using VB = SonarAnalyzer.Rules.VisualBasic;
```

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/MethodsShouldNotHaveTooManyLines.cs
```diff
@@ -0,0 +1,93 @@
...
+namespace SonarAnalyzer.Rules.VisualBasic
+{
+    [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
+    [Rule(DiagnosticId)]
+    public sealed class MethodsShouldNotHaveTooManyLines
+        : MethodsShouldNotHaveTooManyLinesBase<SyntaxKind, MethodBlockBaseSyntax>
+    {
+        private static readonly DiagnosticDescriptor rule =
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
```
andrei-epure-sonarsource: above is static field, this is property, ideally should have a line between

we should change the generator to generate code with `{ get; } = value;` syntax to initialize it

andrei-epure-sonarsource:
```suggestion
        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
```

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/MethodsShouldNotHaveTooManyLines.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        protected override GeneratedCodeRecognizer GeneratedCodeRecognizer =>
```
andrei-epure-sonarsource: `         protected override GeneratedCodeRecognizer GeneratedCodeRecognizer { get; } = VisualBasicGeneratedCodeRecognizer.Instance;`

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/MethodsShouldNotHaveTooManyLines.cs
```diff
@@ -0,0 +1,93 @@
...
+                SyntaxKind.FunctionBlock
+            };
+
+        protected override string MethodKeyword { get; } = "procedures";
+
+        protected override IEnumerable<SyntaxToken> GetMethodTokens(MethodBlockBaseSyntax baseMethodDeclaration) =>
+            baseMethodDeclaration.Statements.SelectMany(s => s.DescendantTokens())
+                ?? Enumerable.Empty<SyntaxToken>();
+
+        protected override SyntaxToken? GetMethodIdentifierToken(MethodBlockBaseSyntax baseMethodDeclaration) =>
```
andrei-epure-sonarsource: can the method not have an identifier? did you add test coverage for this case?

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/MethodsShouldNotHaveTooManyLines.cs
```diff
@@ -0,0 +1,93 @@
...
+            var declaration = identifierToken.Parent;
+            if (declaration.IsKind(SyntaxKind.SubNewStatement))
+            {
+                return $"constructor";
+            }
+
+            var identifierName = identifierToken.ValueText;
+            if (identifierName == null)
+            {
+                return "procedure";
```
andrei-epure-sonarsource: add test for this

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodsShouldNotHaveTooManyLinesBase.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        protected sealed override void Initialize(ParameterLoadingAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                GeneratedCodeRecognizer,
+                c =>
+                {
+                    if (Max < 2)
+                    {
+                        throw new ArgumentOutOfRangeException(
```
andrei-epure-sonarsource: add test

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldNotHaveTooManyLines.cs
```diff
@@ -34,86 +33,60 @@ namespace SonarAnalyzer.Rules.CSharp
...
+        protected override SyntaxToken? GetMethodIdentifierToken(BaseMethodDeclarationSyntax baseMethodDeclaration) =>
+            baseMethodDeclaration?.GetIdentifierOrDefault();
+
+        protected override string GetMethodKindAndName(SyntaxToken identifierToken)
         {
-            var identifierName = baseMethodDeclaration.GetIdentifierOrDefault()?.ValueText;
+            var identifierName = identifierToken.ValueText;
             if (identifierName == null)
             {
                 return "method";
```
andrei-epure-sonarsource: is this covered with a test?

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/MethodsShouldNotHaveTooManyLines.cs
Evangelink: There is no point in changing this `=>` to a `{ get; } =` as this is just returning the value of another property.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/MethodsShouldNotHaveTooManyLines.cs
Evangelink: I always put them without any blank line because they are really tight. 

Regarding the change from `=>` to `{ get; }` this is a good suggestion but I'd rather do that in a separate commit changing all instances instead of having different styling.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldNotHaveTooManyLines.cs
andrei-epure-sonarsource: you should also add a test for C# with NullOrEmpty identifier name (fix the leak!)

Evangelink: @andrei-epure-sonarsource I don't see why this is showing `== null` because the code is `NullOrEmpty` check.

PR #2272, opened 2019-02-08
Add code coverage for S2234 and S1151

andrei-epure-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ParametersCorrectOrderBase.cs
```diff
@@ -67,6 +67,7 @@ public abstract class ParametersCorrectOrderBase<TArgumentSyntax> : SonarDiagnos
                 HasIncorrectlyOrderedParameters(argumentIdentifiers, argumentParameterMappings, parameterNames, identifierNames,
                     analysisContext.SemanticModel))
             {
+                // for VB the symbol does not contain the method syntax reference
```
Evangelink: I don't understand the comment.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParametersCorrectOrder.vb
```diff
@@ -53,7 +53,7 @@
         Sub FooBar(ByVal value As Integer, ByVal ParamArray args() As String)
         End Sub
 
-        Public Function Divide(ByVal divisor As Integer, ByVal dividend As Integer) As Double
+        Public Function Divide(ByVal divisor As Integer, ByVal dividend As Integer) As Double ' Should be secondary-location for [1], but does not work
```
Evangelink: Good catch!

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ParametersCorrectOrderBase.cs
andrei-epure-sonarsource: I could leave just the comment in `SonarAnalyzer.UnitTest/TestCases/ParametersCorrectOrder.vb` ?

The idea is that one would expect `secondaryLocations` to be populated for both CS and VB, but in fact it is not. And the VB implementation of `GetMethodDeclarationIdentifierLocation` is called only with null values, so [VisualBasicSyntaxHelper.FindIdentifierLocation](https://sonarcloud.io/code?id=sonaranalyzer-dotnet&line=201&selected=sonaranalyzer-dotnet%3Asrc%2FSonarAnalyzer.VisualBasic%2FHelpers%2FVisualBasicSyntaxHelper.cs) is not called at all

Evangelink: Ah! There is probably something off with the structure of my code then...

andrei-epure-sonarsource: I'd say we can merge this and you can open a ticket if you want to fix the secondary location issue.

PR #2273, opened 2019-02-08
Update RSPEC for VB

andrei-epure-sonarsource:


sonaranalyzer-dotnet/rspec/vbnet/Sonar_way_recommended_profile.json
```diff
@@ -0,0 +1,6 @@
+{
```
Evangelink: What's that?

andrei-epure-sonarsource: I have no idea. I just ran `.\scripts\rspec\rspec.ps1 vbnet`. But I removed it as we don't know exactly what it is

PR #2294, opened 2019-02-19
Update S2933 to not suggest readonly for structs

### LINKED ISSUE #2291: Update S2933: Rule should not suggest adding readonly modifier for fields of type struct
### Description

Sonar suggest adding `readonly` modifier to field of type `SpinLock`, but this should not be done and is mentioned in .NET documentation https://docs.microsoft.com/en-us/dotnet/api/system.threading.spinlock?view=netframework-4.7.2 in remarks section.

### Repro steps

Have class with `SpinLock` field that is assigned only in constructor or initialized directly

`private System.Threading.SpinLock _lock = new System.Threading.SpinLock();`

### Expected behavior

do not raise S2933 in this situation, perhaps this should be done for all structs?

### Actual behavior

S2933 is raised

### Known workarounds

no workaround

### Related information

* SonarC# Version 4.8.0.4040
* Visual Studio Version VS2017

freza-tm:
Fixes  #2291


Evangelink:
Hi @freza-tm,

Thanks for this PR. I took inspiration from it and created another PR (linked).

PR #2300, opened 2019-02-22
 Update S100: Allow two-letter acronyms

### LINKED ISSUE #2290: Update S100: Allow two-letter acronyms
### Description

Rule S101 was updated to allow two-letter acronyms (#174). I think the same should be done with rule S100, with the same justification:

From [Microsofts Capitalization Conventions](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions)

> A special case is made for two-letter acronyms in which both letters are capitalized, as shown in the following identifier: `IOStream`

Also the message of the analyzer should be corrected (see screenshot). The method name should be in pascal case (instead of camel case):

image

Evangelink:
Fix #2290 

Note that the 2nd commit is a refactoring only commit.


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs
```diff
@@ -33,25 +33,25 @@
...
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
-    [Rule(DiagnosticId_TypeName)]
-    [Rule(DiagnosticId_MethodName)]
+    [Rule(TypeNameDiagnosticId)]
+    [Rule(MethodNameDiagnosticId)]
     public sealed class ClassAndMethodName : SonarDiagnosticAnalyzer
     {
-        internal const string DiagnosticId_MethodName = "S101";
-        internal const string DiagnosticId_TypeName = "S100";
+        internal const string MethodNameDiagnosticId = "S100";
```
Evangelink: Values were inconsistent with RSPEC but...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs
```diff
@@ -33,25 +33,25 @@
...
+        internal const string TypeNameDiagnosticId = "S101";
 
         private const string MessageFormat = "Rename {0} '{1}' to match camel case naming rules, {2}.";
         internal const string MessageFormatNonUnderscore = "consider using '{0}'";
         internal const string MessageFormatUnderscore = "trim underscores from the name";
 
-        private static readonly DiagnosticDescriptor rule_MethodName =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId_TypeName, MessageFormat, RspecStrings.ResourceManager);
+        private static readonly DiagnosticDescriptor methodNameRule =
+            DiagnosticDescriptorBuilder.GetDescriptor(MethodNameDiagnosticId, MessageFormat, RspecStrings.ResourceManager);
```
Evangelink: ...There was a second mismatch here that was making things work properly.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs
```diff
@@ -33,25 +33,25 @@
...
 
-        private static readonly DiagnosticDescriptor rule_TypeName =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId_MethodName, MessageFormat, RspecStrings.ResourceManager);
+        private static readonly DiagnosticDescriptor typeNameRule =
+            DiagnosticDescriptorBuilder.GetDescriptor(TypeNameDiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
-            => ImmutableArray.Create(rule_MethodName, rule_TypeName);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; }
```
Evangelink: Use `{ get; }` to avoid recreating the array every time.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs
```diff
@@ -254,21 +254,18 @@ private static bool IsMemberNameValid(string identifierName, out string suggesti
...
-            {
-                suggestionToProcess = FirstCharToUpper(input.Substring(1));
-            }
-            else
-            {
-                suggestionToProcess = FirstCharToUpper(input);
-            }
-
-            return SuggestFixedCaseName(suggestionToProcess, maxUppercase);
+                !IsCharUpper(input, 1);
```
duncanp-sonar: Not changed as a result of this refactoring, but checking `!IsCharUpper(input, 1)` seems to be redundant since `input.Length == 1` has alreayd been checked.

duncanp-sonar:
Should the description in the issue refer to "Pascal" case rather than "camel case"?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs
Evangelink: This class seems really complex for what it is actually doing but I don't think it's a high priority to rewrite it now.

PR #2304, opened 2019-02-26
Update scripts and update rule API version

duncanp-sonar:


scripts/rspec/create-github-issue.ps1
```diff
@@ -33,9 +33,9 @@ $languageMap = @{
 }
 
 # Update the following variable when a new version of rule-api has to be used.
-$rule_api_version = "1.17.0.1017"
+$rule_api_version = "1.22.2.1215"
 $rule_api_error = "Download Rule-api from " + `
-    "'https://repox.sonarsource.com/sonarsource-private-releases/com/sonarsource/rule-api/rule-api/${rule_api_version}' " +`
+    "'https://repox.jfrog.io/repox/sonarsource-private-releases/com/sonarsource/rule-api/rule-api/${rule_api_version}' " +`
```
Evangelink: Why are you not using the environment variable for the first part of the URL?

scripts/rspec/rspec.ps1
```diff
@@ -36,9 +36,9 @@ $ErrorActionPreference = "Stop"
 $RuleTemplateFolder = "${PSScriptRoot}\\rspec-templates"
 
 # Update the following variable when a new version of rule-api has to be used.
-$rule_api_version = "1.22.0.1199"
+$rule_api_version = "1.22.2.1215"
 $rule_api_error = "Download Rule-api from " + `
-    "'https://repox.sonarsource.com/sonarsource-private-releases/com/sonarsource/rule-api/rule-api/${rule_api_version}' " +`
+    "'https://repox.jfrog.io/repox/sonarsource-private-releases/com/sonarsource/rule-api/rule-api/${rule_api_version}' " +`
```
Evangelink: Same here, why are you not using the environment variable for the first part of the URL?

duncanp-sonar: These scripts are run on dev machines so I wouldn't expect the variable to be set.
And the messages are for info only, telling the dev how to manually set up their machine, so I didn't think it was worth parameterising and having to handle the case of the environment variable not being set.

PR #2309, opened 2019-02-27
Fix S3440: FP on properties affectation.

### LINKED ISSUE #2313: Update S3440: Allow property check before assignment
### Description
We get a false positive for Remove this useless conditional.

### Repro steps

In the following example we guard against assigning the same  value to a property with side affects.

According to the documentation:
Properties are excluded from this rule because they could have side effects and removing the check could lead to undesired side effects.

```csharp
namespace WindowsFormsApp3
{
  public class ClassWithProperty
  {
    private string myProperty;

    private void DoIncridibleExpensiveOperationHere(string argument)
    {
      myProperty = argument;
    }

    public string MyProperty
    {
      get
      {
        return myProperty;
      }

      set
      {
        DoIncridibleExpensiveOperationHere(value);
      }
    }
  }

  public class S3440Test
  {
    private readonly ClassWithProperty test = new ClassWithProperty();

    public void ReproduceWarning(string argument)
    {
      if (test.MyProperty != argument) // Remove this useless conditional.
      {
        test.MyProperty = argument;
      }
    }
  }
}
```

### Expected behavior
No warning

### Actual behavior
Warning: Remove this useless conditional.

### Known workarounds
Mark item as false positive

### Related information
SonarC# Version 7.11.0.8083
Visual Studio Version 2017
If running through the Scanner for MSBuild, its version 4.5.0.1761

Reproduced in VS2019 with SonarAnalyzer.CSharp 7.11.0.8083

kvpt:
The rule should ignore property checking before assignment because of possible side effect, but  it's not the case.
The test is wrong, the verification is made for a check and assignment inside a property and not for a check and assignment of a property.

Fix #2313

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantConditionalAroundAssignment.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantConditionalAroundAssignment.cs
@@ -47,14 +47,22 @@ namespace SonarAnalyzer.Rules.CSharp
                 c =>
                 {
                     var ifStatement = (IfStatementSyntax)c.Node;
+
                     if (ifStatement.Else != null ||
-                        ifStatement.Parent is ElseClauseSyntax ||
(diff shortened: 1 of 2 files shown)


kvpt:
Ok, no problem.

PR #2319, opened 2019-03-06
1h with dev change 

andrei-epure-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MemberShouldBeStatic.cs
```diff
@@ -51,6 +51,15 @@ public sealed class MemberShouldBeStatic : SonarDiagnosticAnalyzer
             SymbolKind.Method
         };
 
+        public static int Method()
```
andrei-epure-sonarsource: this is wrong!

andrei-epure-sonarsource:
demo ended :)

PR #2320, opened 2019-03-06
Fix S1144: False Positive when private member is only used in a gener…

### LINKED ISSUE #2312: Fix S1144: False Positive when private member is only used in a generated file
### Description

We get a false positive for Unused private types or members should be removed

### Repro steps

We have a private function that is called from the Dispose function that is in the designer file

```csharp
namespace WindowsFormsApp3
{
  public partial class Form1 : Form
  {
    public Form1()
    {
      InitializeComponent();
    }

    private void DoMoreDisposing()
    {
    }
  }
}

namespace WindowsFormsApp3
{
  partial class Form1
  {
    /// <summary>
    /// Required designer variable.
    /// </summary>
    private System.ComponentModel.IContainer components = null;

    /// <summary>
    /// Clean up any resources being used.
    /// </summary>
    /// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
    protected override void Dispose(bool disposing)
    {
      if (disposing && (components != null))
      {
        components.Dispose();
      }

      DoMoreDisposing();
      base.Dispose(disposing);
    }

    #region Windows Form Designer generated code

    /// <summary>
    /// Required method for Designer support - do not modify
    /// the contents of this method with the code editor.
    /// </summary>
    private void InitializeComponent()
    {
      this.SuspendLayout();
      // 
      // Form1
      // 
      this.AutoScaleDimensions = new System.Drawing.SizeF(6F, 13F);
      this.AutoScaleMode = System.Windows.Forms.AutoScaleMode.Font;
      this.ClientSize = new System.Drawing.Size(800, 450);
      this.Name = "Form1";
      this.Text = "Form1";
      this.ResumeLayout(false);

    }

    #endregion
  }
}
```

### Expected behavior
No warning

### Actual behavior
Message	S1144	Remove the unused private method 'DoMoreDisposing'.

### Known workarounds
Move the dispose method or mark item as false positive

### Related information
* SonarC# Version 7.11.0.8083
* Visual Studio Version 2017
* If running through the Scanner for MSBuild, its version 4.5.0.1761

Reproduced in VS2019 with SonarAnalyzer.CSharp 7.11.0.8083

Evangelink:
…ated file

Fix #2312


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
```diff
@@ -85,7 +85,7 @@ protected override void Initialize(SonarAnalysisContext context)
                             // Collect symbols of private members that could potentially be removed
                             var removableSymbolsCollector = new CSharpRemovableSymbolWalker(c.Compilation.GetSemanticModel);
 
-                            if (!VisitDeclaringReferences(namedType, removableSymbolsCollector, c.Compilation))
+                            if (!VisitDeclaringReferences(namedType, removableSymbolsCollector, c.Compilation, includeGeneratedFile: false))
```
Evangelink: We don't need to collect the members declared in generated files as we don't want to report about them...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
```diff
@@ -100,7 +100,7 @@ protected override void Initialize(SonarAnalysisContext context)
                                 c.Compilation.GetSemanticModel,
                                 removableSymbolsCollector.PrivateSymbols.Select(s => s.Name).ToHashSet());
 
-                            if (!VisitDeclaringReferences(namedType, usageCollector, c.Compilation))
+                            if (!VisitDeclaringReferences(namedType, usageCollector, c.Compilation, includeGeneratedFile: true))
```
Evangelink: ... But we do want to collect the usages of other members even in generated files.

PR #2321, opened 2019-03-06
Update S1905: False Positive with 'default'

### LINKED ISSUE #2295: Update S1905: False Positive with 'default'
### Description

This code throws S1905 but when I make the recommended change the code fails to compile with CS0411. Is this a false positive?

```csharp
using System;

namespace S1905_False_Positive
{
    internal static class Class
    {
        public static void RunAction(Action action)
        {
            RunFunc(() => { action(); return default; }, (bool)default);
        }

        public static T RunFunc<T>(Func<T> func, T returnvalue = default)
        {
            try
            {
                return func();
            }
            catch (Exception)
            {
                return returnvalue;
            }
        }
    }
}
```

### Related information

* SonarC# Version: Latest
* Visual Studio Version : 15.9.7

Evangelink:
Fix #2295

### DIFF
--- a/sonaranalyzer-dotnet/rspec/cs/S1905_c#.html
+++ b/sonaranalyzer-dotnet/rspec/cs/S1905_c#.html
@@ -3,22 +3,29 @@
 <pre>
 public int Example(int i)
 {
-  return (int) (i + 42); // Noncompliant
+    return (int) (i + 42); // Noncompliant
 }
+
 public IEnumerable&lt;int&gt; ExampleCollection(IEnumerable&lt;int&gt; coll)
 {
-  return coll.Reverse().OfType&lt;int&gt;(); // Noncompliant
+    return coll.Reverse().OfType&lt;int&gt;(); // Noncompliant
 }
 </pre>
 <h2>Compliant Solution</h2>
 <pre>
 public int Example(int i)
 {
-  return i + 42;
+    return i + 42;
 }
+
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantCast.cs
```diff
@@ -83,7 +84,8 @@ protected override void Initialize(SonarAnalysisContext context)
                 return;
             }
 
-            if (expressionType.Equals(castType))
+            if (expressionType.Equals(castType)
+                && !expression.IsKind(SyntaxKindEx.DefaultLiteralExpression))
```
Evangelink: As discussed, we exclude all `default` expressions.

sonaranalyzer-dotnet/rspec/cs/S1905_c#.html
```diff
@@ -3,22 +3,29 @@ <h2>Noncompliant Code Example</h2>
...
 }
+
 public IEnumerable&lt;int&gt; ExampleCollection(IEnumerable&lt;int&gt; coll)
 {
-  return coll.Reverse();
+    return coll.Reverse();
 }
 </pre>
+<h2>Exceptions</h2>
+<p>Issues are not raised against C# 7.1 `default` literal.</p>
```
Evangelink: Updates the RSPEC to mention the exclusion.

PR #2332, opened 2019-03-11
Remove executable_lines_data IT

andrei-epure-sonarsource:
It makes master QA fail.
It is not clear if it is an intended drop of API or a regression on the SQ side.


andrei-epure-sonarsource:
the breaking change has been reverted SonarSource/sonar-enterprise#1344 , so will close this PR until further notice

PR #2345, opened 2019-03-15
Update S3874: FP with interface implementations

### LINKED ISSUE #2344: Update S3874: FP with interface implementations
### Description
When implementing an interface or abstract method (or overriding a method), you should not be faced with an error on the defined contract by that interface or method that is overwritten. It is not the place where it can be fixed, and in a lot of cases it even impossible because that contract is not in your control.

### Repro steps
As an example I took rule S3874, which triggered me:

``` CSharp
{
    public class S3874 : I3874
    {
        public void Set(ref S3874 obj) // compliant
        {
            obj = new S3874();
        }
    }

    public interface I3874
    {
        void Set(ref S3874 obj); // non-compliant
    }
}
```
### Related information
* SonarC# Version 7.11.0.8083
* https://github.com/Corniel/SonarAnalyzer.Experiments/blob/master/src/SonarAnalyzer.Experiments.CSharp/S3874.cs

Evangelink:
Fix #2344

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotUseOutRefParameters.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotUseOutRefParameters.cs
@@ -61,7 +61,8 @@ namespace SonarAnalyzer.Rules.CSharp
                     if (!(parameterSymbol?.ContainingSymbol is IMethodSymbol containingMethod) ||
                         containingMethod.IsOverride ||
                         !containingMethod.IsPubliclyAccessible() ||
-                        IsTryPattern(containingMethod, modifier))
+                        IsTryPattern(containingMethod, modifier) ||
+                        containingMethod.GetInterfaceMember() != null)
                     {
                         return;
                     }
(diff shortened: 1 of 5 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotUseOutRefParameters.cs
```diff
@@ -71,16 +72,13 @@ protected override void Initialize(SonarAnalysisContext context)
                 SyntaxKind.Parameter);
         }
 
-        private bool IsTryPattern(IMethodSymbol method, SyntaxToken modifier)
-        {
-            return method.Name.StartsWith("Try", StringComparison.Ordinal) &&
-                method.ReturnType.Is(KnownType.System_Boolean) &&
-                modifier.IsKind(SyntaxKind.OutKeyword);
-        }
+        private bool IsTryPattern(IMethodSymbol method, SyntaxToken modifier) =>
```
Evangelink: [cosmetic] Use expression body

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotUseOutRefParameters.cs
```diff
@@ -71,16 +72,13 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            method.Name.StartsWith("Try", StringComparison.Ordinal)
+            && method.ReturnType.Is(KnownType.System_Boolean)
+            && modifier.IsKind(SyntaxKind.OutKeyword);
 
-        private static bool IsRefOrOut(SyntaxToken token)
-        {
-            return token.IsKind(SyntaxKind.RefKeyword) || token.IsKind(SyntaxKind.OutKeyword);
-        }
+        private static bool IsRefOrOut(SyntaxToken token) =>
```
Evangelink: [cosmetic] Use expression body

PR #2360, opened 2019-03-28
Generate required xml for the plugin as post-build event

Evangelink:


sonaranalyzer-dotnet/src/SonarAnalyzer.RuleDescriptorGenerator/SonarAnalyzer.RuleDescriptorGenerator.csproj
```diff
@@ -24,6 +24,10 @@
   <PropertyGroup>
     <WarningsAsErrors />
     <NoWarn>NU1605</NoWarn>
+    <RunPostBuildEvent>OnBuildSuccess</RunPostBuildEvent>
```
Evangelink: Run on all successful build as this is an inexpensive call.

PR #2367, opened 2019-04-03
Fix S4275: FP with multiple assigments and ref parameters

### LINKED ISSUE #2140: Fix S4275: FP when using MVVM frameworks
### Description

Rule S4275 gives a false positive when the setter is implemented by a member method taking in a reference such as the one implemented for the MVVM Light framework.

### Repro steps

```csharp
public double MyProp
{
     get => this._myProp;
     set
     {
          if (this.Set(ref this._myProp, value))
          {
                this._isDirty = true;
          }
     }
}
```

### Related information

SonarLint version: 4.7.0.3891
SonarQube c# plugin: 7.9.1 (build 7622) 
Visual Studio 15.8.2
SonarQube scanner 4.3.2

### LINKED ISSUE #2195: Fix S4275: FP when multiple assignments inside the setter
It also fails for something like this.

```csharp
public int Field
{
  get
  {
    return _field;
  }
  set
  {
    _suppressEvents = true;
    _field = value;
    [...]
    _suppressEvents = false;
  }
}
```

_Originally posted by @marcells in https://github.com/SonarSource/sonar-dotnet/issues/2140#issuecomment-447310336_

andrei-epure-sonarsource:
Fix #2140
Fix #2195

I've put the alphabetical order refactor in a separate commit for an easy review of the actual changes. **Please squash commits** when merging.

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
@@ -53,7 +53,7 @@ namespace SonarAnalyzer.Rules.CSharp
                 {
                     assignments.Add(ExtractFieldFromExpression(AccessorKind.Setter, assignment.Left, compilation));
                 }
-                else if (node is ArgumentSyntax argument && argument.RefOrOutKeyword.IsKind(SyntaxKind.RefKeyword))
+                else if (node is ArgumentSyntax argument && IsRefOrOut(argument.RefOrOutKeyword))
                 {
                     assignments.Add(ExtractFieldFromExpression(AccessorKind.Setter, argument.Expression, compilation));
                 }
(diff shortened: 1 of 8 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/PropertiesAccessCorrectFieldBase.cs
```diff
@@ -83,16 +84,21 @@ protected void CheckType(SymbolAnalysisContext context)
                 .OfType<IPropertySymbol>()
                 .Where(p => !p.IsImplicitlyDeclared);
 
-        protected void CheckExpectedFieldIsUsed(IFieldSymbol expectedField, FieldData? actualField, SymbolAnalysisContext context)
+        protected void CheckExpectedFieldIsUsed(IFieldSymbol expectedField, IEnumerable<FieldData?> actualFields, SymbolAnalysisContext context)
         {
-            if (actualField.HasValue && actualField.Value.Field != expectedField)
+            var expectedFieldIsUsed = actualFields.Any(a => a.HasValue && a.Value.Field == expectedField);
```
duncanp-sonar: What will the rule do if the code being tested doesn't refer to any fields? (e.g. just has a method call)

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/PropertiesAccessCorrectFieldBase.cs
```diff
@@ -83,16 +84,21 @@ protected void CheckType(SymbolAnalysisContext context)
...
+            if (!expectedFieldIsUsed)
             {
-                context.ReportDiagnosticWhenActive(Diagnostic.Create(
-                    Rule,
-                    actualField.Value.LocationNode.GetLocation(),
-                    actualField.Value.AccessorKind == AccessorKind.Getter ? "getter" : "setter",
-                    expectedField.Name
-                    ));
+                var fieldWithValue = actualFields.FirstOrDefault(a => a.HasValue);
```
duncanp-sonar: I'm guessing it's a deliberate choice to report on the first field usage in the event of the expected field not being used, but it's not clear. It's worth calling out in a comment if this is the case.

duncanp-sonar:
This also fixes #2195, doesn't it?

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -68,6 +76,22 @@ public sealed class PropertiesAccessCorrectField : PropertiesAccessCorrectFieldB
...
 
+        private static FieldData? ExtractFieldFromRefArgument(ArgumentSyntax argument, Compilation compilation)
+        {
+            var semanticModel = compilation.GetSemanticModel(argument.SyntaxTree);
+            if (semanticModel != null && argument.Parent is ArgumentListSyntax argList)
+            {
+                var argumentIndex = argList.Arguments.IndexOf(argument);
+                if (semanticModel.GetSymbolInfo(argList.Parent).Symbol is IMethodSymbol methodSymbol &&
+                    argumentIndex < methodSymbol?.Parameters.Length &&
+                    methodSymbol?.Parameters[argumentIndex]?.RefKind == RefKind.Ref)
```
duncanp-sonar: What about _out_ parameters? Looks like an _out_ parameter would also update the field.
Either way, there should be a test case for it.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -39,33 +40,80 @@ public sealed class PropertiesAccessCorrectField : PropertiesAccessCorrectFieldB
...
+        }
+
+        protected override bool ShouldIgnoreAccessor(IMethodSymbol accessorMethod)
+        {
+            if (!(accessorMethod?.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax() is AccessorDeclarationSyntax accessor))
+            {
+                // no accessor
+                return true;
+            }
+            if (accessor.DescendantNodes().Count(n => n is StatementSyntax) > 2)
```
duncanp-sonar: The code in lines 103-108 made my head hurt; I had to debug through the test to work out that the special case is "_ignore the accessor if it only contains throw_".

As it is, the code still raises an FP if the accessor is expression-bodied e.g.

```
private object baz2;
public object Baz2
{
    get /* 123 */ => /* 456 */ throw new System.InvalidOperationException(""); // Compliant
    set => throw new System.InvalidOperationException(""); // Compliant
}
```

The following code seems to work:
```
            // Special case: ignore the accessor if the only statement/expression is a throw.
            if (accessor.Body == null)
            {
                // Expression-bodied syntax
                return accessor.DescendantNodes().FirstOrDefault() is ArrowExpressionClauseSyntax arrowClause
                    && ShimLayer.CSharp.ThrowExpressionSyntaxWrapper.IsInstance(arrowClause.Expression);
            }

            // Statement-bodied syntax
            return (accessor.Body.DescendantNodes().Count(n => n is StatementSyntax) == 1 &&
                accessor.Body.DescendantNodes().Count(n => n is ThrowStatementSyntax) == 1);
```

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/PropertiesAccessCorrectFieldBase.cs
```diff
@@ -71,28 +73,57 @@ protected void CheckType(SymbolAnalysisContext context)
...
+            {
+                Location location = null;
+                string accessorType = null;
+                if (fields.Count() == 1)
+                {
+                    var fieldWithValue = fields.First();
+                    location = fieldWithValue.LocationNode.GetLocation();
+                    accessorType = fieldWithValue.AccessorKind == AccessorKind.Getter ? "getter" : "setter";
+                }
+                else if (method != null)
```
duncanp-sonar: Is this code just being defensive, or are there valid cases where there are multiple fields but `method` is null?

If it's just defensive then I'd add an assertion to highlight that we've ended up in an unexpected situation.

PR #2374, opened 2019-04-10
Fix S3240: FP with lambdas due to type conversion

### LINKED ISSUE #2244: Fix S3240: FP when using having lambdas within the IF-ELSE
### Description

False positive in [S3240](https://jira.sonarsource.com/browse/RSPEC-3240) when using `Task.Run(Action)`

### Repro steps

```C#
if(condition())
{
    Task.Run(() => A();
}
else
{
    Task.Run(() => B();
}
```

### Expected behavior

Nothing.

### Actual behavior

Suggested refactoring:
```C#
Task.Run(
    condition()
    ? () => A()
    : () => B());
```

Which leads to:
`CS0173	Type of conditional expression cannot be determined because there is no implicit conversion between 'lambda expression' and 'lambda expression`

### Related information

* **SonarC# Version** 7.8 (build 7320)
* **Visual Studio Version** VS 2017 15.9.2

andrei-epure-sonarsource:
Fixes #2244


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalSimplification.cs
```diff
@@ -158,17 +158,74 @@ int Test(object a, object b, object y, bool condition)
...
+            {
+                Foo(1, "2", () => X());
+            }
+            else
+            {
+                Foo(1, "2", () => Y());
+            }
+
+            Action x;
+            if (action != null)
```
andrei-epure-sonarsource: this is for the check that suggests to use `??` , the above are for the ternary operator `?:` suggestions

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
```diff
@@ -119,6 +119,11 @@ private static bool AreTypesCompatible(ExpressionSyntax expression1, ExpressionS
             var type1 = semanticModel.GetTypeInfo(expression1).Type;
             var type2 = semanticModel.GetTypeInfo(expression2).Type;
 
+            if (expression1 is AnonymousFunctionExpressionSyntax || expression2 is AnonymousFunctionExpressionSyntax)
```
duncanp-sonar: Minor optimisation: this check can be done right at the start of the method, before the calls to get `type1` and `type2` (rule of thumb is to do syntax checks first before touching the semantic model since fetching from the semantic model is always slower)

PR #2376, opened 2019-04-11
Fix S4210: not suggest STA for async main

### LINKED ISSUE #2352: Update S4210: Rule should not suggest STAThreadAttribute for async main
S4210 is flagged on ```Main``` on an async WinForms program but even if I add the ```STAThread``` attribute the ApartmentState is actually MTA. Is S4210 a false positive for async mains?

```csharp
using System.Threading;
using System.Threading.Tasks;

namespace S4210_False_Positive
{
    static class Program
    {
        static async Task Main()
        {
            System.Windows.Forms.MessageBox.Show(Thread.CurrentThread.GetApartmentState().ToString());
            await Task.CompletedTask;
        }
    }
}
```

### Related information

* SonarC# Version: Latest
* Visual Studio Version: 15.9.9

andrei-epure-sonarsource:
because async main is always MTA
Fixes #2352

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MarkWindowsFormsMainWithStaThread.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MarkWindowsFormsMainWithStaThread.cs
@@ -97,4 +97,23 @@ namespace Tests.Diagnostics
             await Task.CompletedTask;
         }
     }
+
+    class Program_11
+    {
+        [STAThread]
+        public static async Task Main()
+        {
+            await Task.CompletedTask;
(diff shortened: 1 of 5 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MarkWindowsFormsMainWithStaThread.cs
```diff
@@ -87,4 +89,12 @@ public static int Main(string[] args)
             return 1;
         }
     }
+
+    class Program_10
```
duncanp-sonar: It would be good to have a test with both `async` and `[STAThread(...)]`.

PR #2377, opened 2019-04-12
Fix S3459: FP with interop on private class members

### LINKED ISSUE #2236: Fix S3459: False Positive with interop methods
### Description

See #242 

The current algorithm will still report on fields that are actually used but not assigned.

andrei-epure-sonarsource:
Fixes #2236


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -112,6 +114,9 @@ protected override void Initialize(SonarAnalysisContext context)
                 SymbolKind.NamedType);
         }
 
+        private static bool HasStructLayoutAttribute(INamedTypeSymbol namedTypeSymbol) =>
```
andrei-epure-sonarsource: I could move that in the SyntaxHelper tbh

PR #2379, opened 2019-04-15
Fix S4056: FP for overloads with different signature

### LINKED ISSUE #2139: Fix S4056: Rule recommends overload with completely different signature
### Description

The issue is raised on call of methods with several overloads when one of them has CultureInfo parameter, while others, with different signatures, don't. 

Is it expected behavior?
If yes, I would like to ask, if it can be configured somehow? And what use cases for that behavior are?
Now I have 3-party lib that has a class with such method and the method is used very often in my projects. So, every time I get false positive issue on it.

### Repro steps

```
string Get(int)
string Get(Enum, CultureInfo)
```

### Expected behavior

No error. 

### Actual behavior
False detection of Error S4056: Use the overload that takes a 'CultureInfo' or 'IFormatProvider' parameter.

### Related information

* SonarC# 7.9.1
* Scanner for MSBuild 4.0.2.892

andrei-epure-sonarsource:
Fixes #2139


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SpecifyIFormatProviderOrCultureInfo.cs
```diff
@@ -41,10 +42,11 @@ public sealed class SpecifyIFormatProviderOrCultureInfo : SonarDiagnosticAnalyze
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
 
-        private static readonly ImmutableArray<KnownType> formatAndClutureTypes =
+        private static readonly ImmutableArray<KnownType> formatAndCultureType =
             ImmutableArray.Create(
                 KnownType.System_IFormatProvider,
-                KnownType.System_Globalization_CultureInfo
+                KnownType.System_Globalization_CultureInfo,
+                KnownType.System_StringComparison
```
andrei-epure-sonarsource: needed to pass existing test `"".StartsWith("")`

andrei-epure-sonarsource: removed it after all, it should be a separate rule (I started a discussion with NicoH)

sonaranalyzer-dotnet/its/expected/akka.net/Akka.MultiNodeTestRunner.Shared-{964F0EC5-FBE6-47C5-8AE6-145114D5DB8C}-S4056.json
```diff
@@ -134,175 +56,6 @@
...
-"endLine":  26,
-"endColumn":  84
-}
-}
-},
-{
-"id":  "S4056",
-"message":  "Use the overload that takes a 'CultureInfo' or 'IFormatProvider' parameter.",
-"location":  {
-"uri":  "sources\akka.net\src\core\Akka.MultiNodeTestRunner.Shared\Sinks\ConsoleMessageSinkActor.cs",
```
andrei-epure-sonarsource: many of the "lost" issues are `string.StartsWith` which is ok, but many are not raising on `string.Format("Results for {0}", data.FactName))` (like the current line on `ConsoleMessageSinkActor.cs`) - I added tests for `string.Format` and I don't understand the behavior

By looking inside the `mscorlib.dll` contents for `System.string` I see only

```

        public static String Format(IFormatProvider provider, String format, params object[] args);

        public static String Format(String format, params object[] args);

        public static String Format(String format, object arg0, object arg1, object arg2);

        public static String Format(String format, object arg0);

        public static String Format(String format, object arg0, object arg1);
```

Case in which it depends how Roslyn identifies the `invocation` - it may overlap with the `params` overload or not.

But the official docs from https://docs.microsoft.com/en-us/dotnet/api/system.string.format?view=netframework-4.7.2 show overloads for each of the above signatures, like below. If the `mscorlib.dll` would respect the official docs, we wouldn't have FN. And I don't want to add complexity to my implementation because of this.

```
Format(String, Object) 
Format(String, Object[])
Format(IFormatProvider, String, Object) 
Format(IFormatProvider, String, Object[])
Format(String, Object, Object)
Format(IFormatProvider, String, Object, Object)
Format(String, Object, Object, Object)
Format(IFormatProvider, String, Object, Object, Object)
```

andrei-epure-sonarsource: although `string.Format` is quite used and is quite the "predilect target" of this rule... so I'm not completely sure.
I would rather open a separate ticket for the FN rather than fixing it in this ticket...

andrei-epure-sonarsource: after discussing with @duncanp-sonar, I should fix the FNs before merging this ticket to avoid a bad experience for users

sonaranalyzer-dotnet/its/expected/Automapper/AutoMapper-025B9F93-7953-454B-AF29-E6A55B2AA58F-S4056.json
```diff
@@ -17,97 +17,6 @@
 "id":  "S4056",
 "message":  "Use the overload that takes a 'CultureInfo' or 'IFormatProvider' parameter.",
 "location":  {
-"uri":  "sources\Automapper\src\AutoMapper\AutoMapperConfigurationException.cs",
```
andrei-epure-sonarsource: `StartsWith` - TP

sonaranalyzer-dotnet/its/expected/Automapper/AutoMapper-025B9F93-7953-454B-AF29-E6A55B2AA58F-S4056.json
```diff
@@ -17,97 +17,6 @@
...
-"endLine":  121,
-"endColumn":  88
-}
-}
-},
-{
-"id":  "S4056",
-"message":  "Use the overload that takes a 'CultureInfo' or 'IFormatProvider' parameter.",
-"location":  {
-"uri":  "sources\Automapper\src\AutoMapper\Configuration\Conventions\ReplaceName.cs",
```
andrei-epure-sonarsource: `Replace` - no overload

sonaranalyzer-dotnet/its/expected/Automapper/AutoMapper-025B9F93-7953-454B-AF29-E6A55B2AA58F-S4056.json
```diff
@@ -147,32 +56,6 @@
 "id":  "S4056",
 "message":  "Use the overload that takes a 'CultureInfo' or 'IFormatProvider' parameter.",
 "location":  {
-"uri":  "sources\Automapper\src\AutoMapper\Mappers\TypeConverterMapper.cs",
```
andrei-epure-sonarsource: `ConvertTo(object value, Type destinationType);` vs `ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType);` - not compatible

sonaranalyzer-dotnet/its/expected/Ember-MM/Ember.Plugins-{9496C697-5AFD-4813-AEDC-AF33FACEADF0}-S4056.json
```diff
@@ -19,19 +19,6 @@
 "location":  {
 "uri":  "sources\Ember-MM\Ember.Plugins\Ember\EmberWorkAroundPlugin.cs",
 "region":  {
-"startLine":  63,
```
andrei-epure-sonarsource: `EndsWith(String value);` vs `EndsWith(String value, bool ignoreCase, CultureInfo culture);` - not direct overload

andrei-epure-sonarsource: answer: we have a rule for StringComparison https://rules.sonarsource.com/csharp/RSPEC-4058

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SpecifyIFormatProviderOrCultureInfo.cs
```diff
@@ -92,18 +92,99 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                }
+            }
+            return true;
+
+            ITypeSymbol GetParamsType(ITypeSymbol typeSymbol) => (typeSymbol as IArrayTypeSymbol)?.ElementType;
+        }
+
+        /**
+         * Checks that the invocation parameters are of the same type (or a subtype) with the overload parameters
+         * Assumption: the two lists have the same number of elements.
```
duncanp-sonar: A `Debug.Assert` to check the assumption would be better than a comment.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SpecifyIFormatProviderOrCultureInfo.cs
```diff
@@ -92,18 +92,99 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            for (; i < invocationParameters.Count; i++)
+            {
+                if (!invocationParameters[i].Type.DerivesOrImplements(paramsType))
+                {
+                    return false;
+                }
+            }
+            return true;
+
+            ITypeSymbol GetParamsType(ITypeSymbol typeSymbol) => (typeSymbol as IArrayTypeSymbol)?.ElementType;
```
duncanp-sonar: This line is coded defensively to handle `typeSymbol` not being an array type. If we think this could happen in practice (e.g. we're passed an invalid syntax tree that somehow gives a valid symbol that is a `params` symbol but not an array type) then it would be better to do the check right at the start of the method and exit early if we can't work out the type.
If we don't think it could happen in practice then the defensive code isn't needed.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SpecifyIFormatProviderOrCultureInfo.cs
```diff
@@ -92,18 +92,99 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            {
+                return false;
+            }
+
+            // once we filter out the FormatOrCulture argument, we have two possibilities:
+            // - possibleOverload has a 'params' argument which matches the invocationParameters
+            // - the number of parameters is the same and of same type
+
+            var invocationParametersNumber = invocationParameters.Count();
+            var parametersWithoutFormatCultureNumber = parametersWithoutFormatCulture.Count();
```
duncanp-sonar: `parametersWithoutFormatCulture.Count()` has already been used in line 116 above. It would be better to move this assignment before line 116 to avoid lazily evaluating the LINQ expression again.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SpecifyIFormatProviderOrCultureInfo.cs
```diff
@@ -92,18 +92,99 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            if (parametersWithoutFormatCulture.Count() == possibleOverload.GetParameters().Count())
+            {
+                return false;
+            }
+
+            // once we filter out the FormatOrCulture argument, we have two possibilities:
+            // - possibleOverload has a 'params' argument which matches the invocationParameters
+            // - the number of parameters is the same and of same type
+
+            var invocationParametersNumber = invocationParameters.Count();
```
duncanp-sonar: `Count` is more descriptive of what the variable represents than `Number`.
i.e.  `invocationParametersCount` rather than `invocationParametersNumber`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SpecifyIFormatProviderOrCultureInfo.cs
```diff
@@ -92,18 +92,99 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            var parametersWithoutFormatCultureNumber = parametersWithoutFormatCulture.Count();
+
+            if (parametersWithoutFormatCultureNumber <= invocationParametersNumber &&
+                parametersWithoutFormatCultureNumber > 0 &&
+                parametersWithoutFormatCulture.Last() is IParameterSymbol lastParameter &&
```
duncanp-sonar: `parametersWithoutFormatCulture.Last()` will lazily-evaluate the expression again.
Strictly speaking we can determine if the last parameter of the possible overload is `params` by checking `possibleOverload.GetParameters().Last().IsParams` which would avoid re-evaluating the expression.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SpecifyIFormatProviderOrCultureInfo.cs
```diff
@@ -92,18 +92,99 @@ protected override void Initialize(SonarAnalysisContext context)
...
+
+            // must have same number of arguments + 1 (the format or culture argument) OR is params argument
+            bool IsCompatibleOverload(IMethodSymbol m) =>
+                (m.GetParameters().Count() - invocation.ArgumentList.Arguments.Count == 1) ||
+                (m.GetParameters().Any() && m.GetParameters().Last().IsParams);
+        }
+
+        private static bool SameParametersExceptFormatOrCulture(IMethodSymbol possibleOverload, IEnumerable<IParameterSymbol> invocationParameters)
+        {
+            var parametersWithoutFormatCulture = possibleOverload.GetParameters().Where(p => !p.Type.IsAny(formatAndCultureType));
```
duncanp-sonar: This expression is evaluated multiple times in the rest of the method. Most of the following comments are about avoiding this.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SpecifyIFormatProviderOrCultureInfo.cs
andrei-epure-sonarsource: this sounds like a good rule to RSPEC

PR #2380, opened 2019-04-15
Added xunit.runner.visualstudio v2.4.1

Corniel:
Currently at my Visual Studio 2017, the unit tests did not longer run. After adding xunit.runner.visualstudio that works again.


Corniel:
Hmm, strangly this works, but as XUnit is not used, it should use another test runner.

Corniel:
Kill. I'm not sure what went wrong locally, but without, now I can run my tests too. Sorry for the inconvenience.

Issue #1548, opened 2018-07-10
New analyzer idea: replace empty array instantiation by Array.Empty<>()


GeertvanHorrik:
### Description

To prevent unneeded allocations, empty array allocations should be replaced by `Array.Empty<T>()` so it will re-use the same instance over and over again. This will especially benefit high-performance code.

```
myMethod.Invoke(someObject, new Type[] { });
```

Should be written as:

```
myMethod.Invoke(someObject, Array.Empty<Type>());
```

pavel-mikula-sonarsource:
There's a PR from @Corniel to implement this. (This is a useless message that I need to get Corniel on Kanban :) )

pavel-mikula-sonarsource:
Kanban cheat didn't work :(

Corniel:
@pavel-mikula-sonarsource Too bad. :(

pavel-mikula-sonarsource:
@Corniel Now it works, it needed your post in the thread ✔️

Corniel:
This one can be closed. This rule exists already.

PR #2383, opened 2019-04-16
Implementation of new rule Use Array.Empty<T>() 

Corniel:
As proposed via #1548.


Corniel:
@GeertvanHorrik I gave your suggestion a try.

Corniel:
Note that `Array.Empty<T>()` is available since .NET 4.6 See: https://docs.microsoft.com/en-us/dotnet/api/system.array.empty?view=netframework-4.6.

andrei-epure-sonarsource:
thanks, @Corniel . Before we could merge this, we'd need to create a rule specification (RSPEC) for #1548. The `RspecStrings.resx` files are machine-generated from our [rule repository](https://jira.sonarsource.com/browse/RSPEC-5324?jql=project%20%3D%20RSPEC%20AND%20resolution%20%3D%20Unresolved%20ORDER%20BY%20updated%20DESC).

No ETA for that, it will have to be put on the roadmap at some point.
Thank you for the patience.

(cc @nicolas-harraudeau-sonarsource )

Corniel:
@andrei-epure-sonarsource : I know there is a process for that, and that it will take time. BTW, I could not find how to make a rule that is only applicable based on a version of .NET. Can you enlighten me on that?

andrei-epure-sonarsource:
hi @Corniel , thanks for the follow-up. The short answer is that we currently don't have rules that run only in a certain .NET version.

However, we do have tests that parse based on specific language versions - you can check how `ParseOptionsHelper` is used in various unit tests.

Also, if you're willing to explore, `Microsoft.CodeAnalysis.SemanticModel` has a reference to `Microsoft.CodeAnalysis.Compilation` which has a reference to `Microsoft.CodeAnalysis.CompilationOptions` which *may* contain the language version, but I haven't personally used that (yet).

Corniel:
@andrei-epure-sonarsource The `Microsoft.CodeAnalysis.CompilationOptions` do not provide the .NET framework version, I found out.

What might be worth a try is the following:

``` C#
var version = c.Compilation.Assembly.GetAttributes().GetCustomAttribute<TargetFrameworkAttribute>()?.FrameworkName;
```
But this requires potentially some extra thoughts. You could think about a custom version type that is IComparable, that allows you to define per rule, which target framework(s) apply.

Might be better to put that in a different pull request.

Corniel:
@andrei-epure-sonarsource What is the status of this?

andrei-epure-sonarsource:
@Corniel - thanks for following up. In short, we've been busy with making sure that SonarQube 7.9 LTS (release should be next week) will be the best LTS ever.

Sometime this summer I will create rule specifications (RSPEC) in JIRA for this and other rule suggestions we got over the time. This rule has priority as it already has a PR out. I'll then code review it. Then, I will need to run an internal tool to generate the meta-files for the rule and update the PR with that.

thanks for the patience 😃

Corniel:
@andrei-epure-sonarsource For now, no framework detection?

andrei-epure-sonarsource:
@Corniel  regarding framework detection, we're going to need it for #3096

Corniel:
@andrei-epure-sonarsource this one is blocked till #3096 is finished?

andrei-epure-sonarsource:
> @andrei-epure-sonarsource this one is blocked till #3096 is finished?

- it's blocked until we make the rule specification in our rule repository for #1548
- then that rule will be included in an MMF ("Minimal Marketable Feature") and shipped together with other rules
- then we need to cherry-pick these commits on a PR started by one of us (the full CI works only on PRs started by sonarsources), add the rule specs (which are using an internal API to query the rule repository to create the metadata files) and do the code review

To sum up, there's more than just code review involved.

andrei-epure-sonarsource:
@Corniel , this FI kinda messed up the history of this PR.
we do rebase on master to keep a clean history

PR #2400, opened 2019-04-26
Fix #2384: Update Rule S2255 (remove cookie reading)

### LINKED ISSUE #2384: Update S2255:  Remove Cookie reading from Rule S2255
In order to make security hotspot rules more useful to developers, we are refocusing them on issues which are not already covered by our taint analysis engine. Thus we now remove cookie reading from Rule S2255.

For C#, this was added recently by #1842. Note that the PR cannot be fully reverted as the rule was new for SonarVB.net.

christophe-zurn-sonarsource:
Fixes #2384

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
@@ -55,5 +55,28 @@ namespace SonarAnalyzer.Helpers
                     constantValue.Value is string constant &&
                     constant == value;
             };
+
+        #region Syntax-level checking methods
+
+        public override ElementAccessCondition MatchSetter() =>
+            (context) => ((ExpressionSyntax)context.Expression).IsLeftSideOfAssignment();
+
+        public override ElementAccessCondition MatchProperty(MemberDescriptor member)
+        {
+            return (context) => {
+                var accessSyntax = context.Expression as ElementAccessExpressionSyntax;
+                if(accessSyntax.Expression.IsKind(SyntaxKind.SimpleMemberAccessExpression))
+                {
+                    var memberAccessSyntax = accessSyntax.Expression as MemberAccessExpressionSyntax;
+                    var propertyName = memberAccessSyntax.Name.Identifier.ValueText;
+                    var enclosingClassType = context.SemanticModel.GetTypeInfo(memberAccessSyntax.Expression);
+                    return enclosingClassType.Type != null && member.IsMatch(propertyName, enclosingClassType.Type, false);
+                }
+                return false;
+            };
+
+        }
+
+        #endregion
     }
 }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
@@ -63,18 +63,15 @@ namespace SonarAnalyzer.Helpers
 
         public override ElementAccessCondition MatchProperty(MemberDescriptor member)
         {
-            return (context) => {
-                var accessSyntax = context.Expression as ElementAccessExpressionSyntax;
-                if(accessSyntax.Expression.IsKind(SyntaxKind.SimpleMemberAccessExpression))
-                {
-                    var memberAccessSyntax = accessSyntax.Expression as MemberAccessExpressionSyntax;
-                    var propertyName = memberAccessSyntax.Name.Identifier.ValueText;
-                    var enclosingClassType = context.SemanticModel.GetTypeInfo(memberAccessSyntax.Expression);
-                    return enclosingClassType.Type != null && member.IsMatch(propertyName, enclosingClassType.Type, false);
-                }
-                return false;
-            };
-
+            return (context) =>
+                ((ElementAccessExpressionSyntax)context.Expression).Expression is MemberAccessExpressionSyntax memberAccess &&
+                 memberAccess.IsKind(SyntaxKind.SimpleMemberAccessExpression) &&
+                 memberAccess.Name is SimpleNameSyntax memberName &&
+                 memberName.Identifier is SyntaxToken memberIdentifier &&
+                 context.SemanticModel.GetTypeInfo(memberAccess.Expression) is TypeInfo enclosingClassType &&
+                 memberIdentifier.ValueText != null &&
+                 enclosingClassType.Type != null &&
+                 member.IsMatch(memberIdentifier.ValueText, enclosingClassType.Type, false);
         }
 
         #endregion
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
@@ -70,7 +70,7 @@ namespace SonarAnalyzer.Helpers
                  context.SemanticModel.GetTypeInfo(memberAccess.Expression) is TypeInfo enclosingClassType &&
                  memberIdentifier.ValueText != null &&
                  enclosingClassType.Type != null &&
-                 member.IsMatch(memberIdentifier.ValueText, enclosingClassType.Type, false);
+                 member.IsMatch(memberIdentifier.ValueText, enclosingClassType.Type);
 
         #endregion
     }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
@@ -61,9 +61,8 @@ namespace SonarAnalyzer.Helpers
         public override ElementAccessCondition MatchSetter() =>
             (context) => ((ExpressionSyntax)context.Expression).IsLeftSideOfAssignment();
 
-        public override ElementAccessCondition MatchProperty(MemberDescriptor member)
-        {
-            return (context) =>
+        public override ElementAccessCondition MatchProperty(MemberDescriptor member) =>
+            (context) =>
                 ((ElementAccessExpressionSyntax)context.Expression).Expression is MemberAccessExpressionSyntax memberAccess &&
                  memberAccess.IsKind(SyntaxKind.SimpleMemberAccessExpression) &&
                  memberAccess.Name is SimpleNameSyntax memberName &&
@@ -72,7 +71,6 @@ namespace SonarAnalyzer.Helpers
                  memberIdentifier.ValueText != null &&
                  enclosingClassType.Type != null &&
                  member.IsMatch(memberIdentifier.ValueText, enclosingClassType.Type, false);
-        }
 
         #endregion
     }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/MemberDescriptor.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/MemberDescriptor.cs
@@ -35,20 +35,19 @@ namespace SonarAnalyzer.Helpers
 
         internal string Name { get; }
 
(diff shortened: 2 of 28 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
```diff
@@ -55,5 +55,28 @@ public CSharpElementAccessTracker(IAnalyzerConfiguration analyzerConfiguration,
                     constantValue.Value is string constant &&
                     constant == value;
             };
+
+        #region Syntax-level checking methods
+
+        public override ElementAccessCondition MatchSetter() =>
+            (context) => ((ExpressionSyntax)context.Expression).IsLeftSideOfAssignment();
```
andrei-epure-sonarsource: is this safe? is there a check before to ensure it won't be a cast error?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
```diff
@@ -55,5 +55,28 @@ public CSharpElementAccessTracker(IAnalyzerConfiguration analyzerConfiguration,
...
+
+        public override ElementAccessCondition MatchProperty(MemberDescriptor member)
+        {
+            return (context) => {
+                var accessSyntax = context.Expression as ElementAccessExpressionSyntax;
+                if(accessSyntax.Expression.IsKind(SyntaxKind.SimpleMemberAccessExpression))
```
andrei-epure-sonarsource: it may be null after cast, you can replace the if with

```
if (context.Expression is ElementAccessExpressionSyntax accessSyntax &&
    accessSyntax.Expression.IsKind(SyntaxKind.SimpleMemberAccessExpression))
{}
```
this ensures the null-check

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
```diff
@@ -55,5 +55,28 @@ public CSharpElementAccessTracker(IAnalyzerConfiguration analyzerConfiguration,
...
+                {
+                    var memberAccessSyntax = accessSyntax.Expression as MemberAccessExpressionSyntax;
```
andrei-epure-sonarsource: and you can chain this with `is` inside the above if statement

andrei-epure-sonarsource: actually I tend to think all of this can be shortened into a single arrow-body function like
```
public override ElementAccessCondition MatchProperty(MemberDescriptor member) =>
  (context) =>
    context.Expression is ElementAccessExpressionSyntax accessSyntax &&
    accessSyntax.Expression.IsKind(SyntaxKind.SimpleMemberAccessExpression) &&
    accessSyntax.Expression is MemberAccessExpressionSyntax memberAccessSyntax &&
    memberAccessSyntax.Name.Identifier.ValueText is SyntaxNode propertyName &&
    context.SemanticModel.GetTypeInfo(memberAccessSyntax.Expression) is ITypeInfo enclosingClassType &&
    enclosingClassType.Type != null && 
    member.IsMatch(propertyName, enclosingClassType.Type, false);
```

andrei-epure-sonarsource: applies both to C# and VB impl

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
christophe-zurn-sonarsource: I inspired this from https://github.com/SonarSource/sonar-dotnet/blob/22c93d85f30e0952788e878b8ab002faa0d958f5/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpPropertyAccessTracker.cs#L60
I assumed that trackers were already filtered on correct element type, but I actually cannot find anywhere in the code where this is done. I ll add a check beforehand unless I can find why/where there is already a check on the type

christophe-zurn-sonarsource: I checked. This is safe, as tracker is registered on `ElementAccessExpression` nodes (https://github.com/SonarSource/sonar-dotnet/blob/27fdfc70a0d643ab01cd3f04288bb24a2aa11007/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs#L36).
So all `context.Expression` elements will be a subclass of `ExpressionSyntax` in this tracker

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
christophe-zurn-sonarsource: I added many checks and switched to single-arrow function, so as to be on the safer side. Only the 1st one is not necessary, as the first cast is based on the fact that we know the type of node the checker is registered on

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
```diff
@@ -63,18 +63,15 @@ public CSharpElementAccessTracker(IAnalyzerConfiguration analyzerConfiguration,
...
-                {
-                    var memberAccessSyntax = accessSyntax.Expression as MemberAccessExpressionSyntax;
-                    var propertyName = memberAccessSyntax.Name.Identifier.ValueText;
-                    var enclosingClassType = context.SemanticModel.GetTypeInfo(memberAccessSyntax.Expression);
-                    return enclosingClassType.Type != null && member.IsMatch(propertyName, enclosingClassType.Type, false);
-                }
-                return false;
-            };
-
+            return (context) =>
```
andrei-epure-sonarsource: nitpick: you can replace the body of the method, as it has only 1 return
```
public override ElementAccessCondition MatchProperty(MemberDescriptor member) =>
	(context) =>
		// ... code
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UsingCookies_NetCore.vb
```diff
@@ -8,7 +8,7 @@ Namespace Tests.Diagnostics
             ' Response headers
             response.Headers.Add("Set-Cookie", "") ' Noncompliant
             response.Headers("Set-Cookie") = "" ' Noncompliant
-            value = response.Headers("Set-Cookie") ' Noncompliant
+            value = response.Headers("Set-Cookie") ' Compliant
 
             ' Not the Set-Cookie header
             response.Headers.Add("something", "")
```
andrei-epure-sonarsource: as Vb.net is case insensitive, it would be nice to have tests for lower case method calls as well like

```
Dim responseHeaders2 = response.HEADERS
responseHeaders2.ADD("Set-Cookie", "") ' Noncompliant

Dim responseCookies2 = response.cookies
responseCookies2.append("", "") ' Noncompliant
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpElementAccessTracker.cs
christophe-zurn-sonarsource: Good to know :)

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/MemberDescriptor.cs
```diff
@@ -35,20 +35,19 @@ internal MemberDescriptor(KnownType containingType, string name)
...
         internal string Name { get; }
 
-        public bool IsMatch(string memberName, ITypeSymbol containingType, bool checkOverriddenMethods) =>
-            memberName == Name &&
-            containingType.Is(ContainingType);
+        public bool IsMatch(string memberName, ITypeSymbol containingType, bool caseInsensitiveComparison) =>
+            containingType.Is(ContainingType) && HasSameName(memberName, Name, caseInsensitiveComparison);
 
-        public bool IsMatch<TSymbolType>(string memberName, Lazy<TSymbolType> memberSymbol, bool checkOverriddenMethods)
+        public bool IsMatch<TSymbolType>(string memberName, Lazy<TSymbolType> memberSymbol, bool caseInsensitiveComparison)
```
andrei-epure-sonarsource: as a nitpick, you could choose a default value (for example, `bool caseInsensitiveComparison = false`) and then in the CSharp `DeliveringDebugFeaturesInProduction.cs` you could just call `isDevelopmentMethod.IsMatch(methodName, methodSymbol);` and pass the 3rd parameter only for the VBnet implementation

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/MemberDescriptor.cs
```diff
@@ -35,20 +35,19 @@ internal MemberDescriptor(KnownType containingType, string name)
...
         {
-            if (memberName != Name ||
+            if (!HasSameName(memberName, Name, caseInsensitiveComparison) ||
                 memberSymbol.Value == null)
             {
                 return false;
             }
 
-            if (HasSameContainingType(memberSymbol.Value, checkOverriddenMethods))
+            if (HasSameContainingType(memberSymbol.Value, true))
```
andrei-epure-sonarsource: nitpick: could use named arguments `HasSameContainingType(memberSymbol.Value, checkOverriddenMethods: true)`

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/SyntaxTrackers/VisualBasicFieldAccessTracker.cs
```diff
@@ -30,7 +30,7 @@ namespace SonarAnalyzer.Helpers
     public class VisualBasicFieldAccessTracker : FieldAccessTracker<SyntaxKind>
     {
         public VisualBasicFieldAccessTracker(IAnalyzerConfiguration analyzerConfiguration, DiagnosticDescriptor rule)
-            : base(analyzerConfiguration, rule)
+            : base(analyzerConfiguration, rule, true)
```
andrei-epure-sonarsource: nitpick: could use named arguments for `true`, would make code more readable

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/SyntaxTrackers/VisualBasicInvocationTracker.cs
```diff
@@ -86,8 +85,7 @@ public override InvocationCondition MatchProperty(MemberDescriptor member)
                 context.SemanticModel.GetTypeInfo(propertyMemberAccess.Expression) is TypeInfo enclosingClassType &&
                 propertyMemberIdentifier.ValueText != null &&
                 enclosingClassType.Type != null &&
-                member.IsMatch(propertyMemberIdentifier.ValueText, enclosingClassType.Type, false);
-        }
+                member.IsMatch(propertyMemberIdentifier.ValueText, enclosingClassType.Type, true);
```
andrei-epure-sonarsource: nitpick would me more readable with named arg `caseInsensitiveComparison: true`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/MethodSignatureHelperTest.cs
```diff
@@ -322,6 +322,43 @@ End Namespace
             DoCheckMatch_InterfaceMethods(snippet);
         }
 
+        [TestMethod]
+        public void CheckMatch_CaseInsensitivity()
```
andrei-epure-sonarsource: nice!

PR #2406, opened 2019-04-30
Convert Unit test project to Visual Studio 2017 structure

Corniel:
I had some issues (again) with running/debugging tests. So I decided to put some effort in updating the tests. I also had to update `System.Collections.Immutable` to 1.2.0 as VS was complaining (error) that it had to downgrade to 1.1.37.

I removed signing of the unit test project (for now). Some tests fail now as it (apparently) now throws an `XUnit.Exception` (I don't understand why).

I don't have any `AssemblyInfo.cs` in this project anymore. Therefor some other tests seems to fail. It is not up to me to judge if this is an issue or not.

I hope this change it helps the team.


andrei-epure-sonarsource:
Thanks, @Corniel , sorry for the late reply on this.
I think this will be a big help for us when we're going to migrate to the new csproj structure. However, we didn't make progress on this as there is a testing overhead (like you've mentioned yourself) and hence it's still on the todo list.

Corniel:
I'll close this one. Given all the FI you have to do, it is better to just try it again from the latest master.

andrei-epure-sonarsource:
Sorry for that, @Corniel 

To give some (late) feedback on this, regarding the upgrade for `<PackageReference Include="System.Collections.Immutable" Version="1.2.0" />` in `SonarAnalyzer.CSharp.csproj` - we kept the old one because of a problem in Visual Studio 2015 SP3 regression regarding this specific package (without doing this we couldn't run w/ MSBuild 14). I know it gives a build warning but it's on purpose.

We'll look into improving how we deal with external PRs.

PR #2409, opened 2019-04-30
Update S4784: target only hard-coded regex

### LINKED ISSUE #2386: Rule S4784: Update implementation to target only hard-coded regex
In order to make security hotspot rules more useful to developers, we are refocusing them on issues which are not already covered by our taint analysis engine.

S4784's description has been updated to match only hardcoded regular expressions which have specific characteristics. The goal is to detect potential ReDos vulnerabilities which are not due to regular expression injection.

andrei-epure-sonarsource:
Fixes #2386


andrei-epure-sonarsource:
Quality Gate is failing on PR due to code duplication between C# and VB.NET impl, but that's expected

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UsingRegularExpressions.cs
```diff
@@ -7,65 +7,81 @@ namespace Tests.Diagnostics
...
-            r = new Regex("", RegexOptions.Compiled, TimeSpan.Zero); // Noncompliant
+            r = new Regex(""); // Compliant, less than 3 characters
+            r = new Regex("**"); // Compliant, less than 3 characters
+            r = new Regex("+*"); // Compliant, less than 3 characters
+            r = new Regex("abcdefghijklmnopqrst"); // Compliant, does not have the special characters
+            r = new Regex("abcdefghijklmnopqrst+"); // Compliant, has only 1 special character
+            r = new Regex("{abc}+defghijklmnopqrst"); // Noncompliant
+            r = new Regex("{abc}+{a}"); // Noncompliant {{Make sure that using a regular expression is safe here.}}
+//              ^^^^^^^^^^^^^^^^^^^^^^
+            r = new Regex("+++"); // Noncompliant
```
duncanp-sonar: I guess we'll raise false positives in the cases where special characters exist in the string but those characters have been escaped. 
That's probably ok for now (unless it's easy to fix), but it would be worth adding a a couple of test cases to document the FNs.

PR #2412, opened 2019-05-02
S2068: Update message and fix FP for empty strings

### LINKED ISSUE #2387: S2068 FP: should not raise for empty string
### Description

Rule https://rules.sonarsource.com/csharp/RSPEC-2068

If a `password` field is initialized with `null` or an empty string, it should not be considered as a hardcoded password.

### Repro steps

Add the below line in sonar-dotnet\sonaranalyzer-dotnet\tests\SonarAnalyzer.UnitTest\TestCases\DoNotHardcodeCredentials_DefaultValues.cs

```
            string password2 = "";
```

### Expected behavior

Should not raise.

### Actual behavior

Raises `'Remove hard-coded password(s): 'password'.'`
Also, it should raise `password2` rather than `password` (the variable name, not the substring name we search with).

### Known workarounds

None

### Related information

* SonarC# Version: 7.13

christophe-zurn-sonarsource:
Fix #2387

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotHardcodeCredentials.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotHardcodeCredentials.cs
@@ -55,7 +55,7 @@ namespace SonarAnalyzer.Rules.CSharp
             public VariableDeclarationBannedWordsFinder(DoNotHardcodeCredentialsBase analyzer) : base(analyzer) { }
 
             protected override string GetAssignedValue(VariableDeclaratorSyntax syntaxNode) =>
-                (syntaxNode.Initializer?.Value as LiteralExpressionSyntax)?.Token.ValueText;
+                syntaxNode.Initializer?.Value.GetStringValue();
 
             protected override string GetVariableName(VariableDeclaratorSyntax syntaxNode) =>
                 syntaxNode.Identifier.ValueText;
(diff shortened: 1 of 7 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotHardcodeCredentials.cs
```diff
@@ -55,7 +55,7 @@ private class VariableDeclarationBannedWordsFinder : CredentialWordsFinderBase<V
             public VariableDeclarationBannedWordsFinder(DoNotHardcodeCredentialsBase analyzer) : base(analyzer) { }
 
             protected override string GetAssignedValue(VariableDeclaratorSyntax syntaxNode) =>
-                syntaxNode.Initializer?.Value.ToString();
+                (syntaxNode.Initializer?.Value as LiteralExpressionSyntax)?.Token.ValueText;
```
andrei-epure-sonarsource: you could use the freshly added `GetStringValue` extension method from `CSharpSyntaxHelper` (has null check inside)

```suggestion
                syntaxNode.Initializer?.Value.GetStringValue();
```

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/DoNotHardcodeCredentials.cs
```diff
@@ -55,7 +55,8 @@ private class VariableDeclarationBannedWordsFinder : CredentialWordsFinderBase<V
             public VariableDeclarationBannedWordsFinder(DoNotHardcodeCredentialsBase analyzer) : base(analyzer) { }
 
             protected override string GetAssignedValue(VariableDeclaratorSyntax syntaxNode) =>
-                syntaxNode.Initializer?.Value.ToString();
+                (syntaxNode.Initializer?.Value as LiteralExpressionSyntax)?.Token.ValueText;
```
andrei-epure-sonarsource: could use `GetStringValue()` extension method from `VisualBasicSyntaxHelper`
I think it should work like
```suggestion
                syntaxNode.Initializer?.Value.GetStringValue();
```

PR #2415, opened 2019-05-02
S2201: Update message and description for string.Intern

### LINKED ISSUE #2382: Fix S2201: FP for string.Intern which has side effect
### Description

`string.Intern` has a side-effect: it actually interns the string if it’s not interned already. Hence, this should not be marked as a bug by [S2201](https://rules.sonarsource.com/csharp/RSPEC-2201).

### Repro steps

Add `string.Intern("foo");` inside [ReturnValueIgnored.cs](https://github.com/SonarSource/sonar-dotnet/blob/7.12.0.8217/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ReturnValueIgnored.cs) and run the `ReturnValueIgnored` UT.

### Expected behavior

Should not raise an issue.

### Actual behavior

Raises an issue `'Use the return value of method 'Intern', which has no side effect.'`

### Related information

* SonarC# Version: 7.12

Originally reported on [community.sonarsource.com](https://community.sonarsource.com/t/string-intern-does-have-a-side-effect-but-sonarqube-says-it-doesnt/7217)

christophe-zurn-sonarsource:
Fix #2382

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Utilities/Rules.Description/S2201.html
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Utilities/Rules.Description/S2201.html
@@ -7,8 +7,8 @@ the function call is useless and should be dropped or the source code doesn't be
   <li> any method on <code>string</code>, <code>int</code>, ..., <code>System.Collections.Immutable.ImmutableArray&lt;T&gt;</code>,
   <code>ImmutableHashSet&lt;T&gt;</code>, ... </li>
 </ul>
-<p>Note: although <code>string.Intern</code> method has a side effect, ignoring its return value is still suspicious as it is the only reference
-ensured to point to the intern pool.</p>
+<p>Note: although <code>string.Intern</code> has a side effect, ignoring its return value is still suspicious as it is the only reference ensured to
+point to the intern pool.</p>
 <h2>Noncompliant Code Example</h2>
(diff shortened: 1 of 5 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Utilities/Rules.Description/S2201.html
```diff
@@ -7,6 +7,8 @@
   <li> any method on <code>string</code>, <code>int</code>, ..., <code>System.Collections.Immutable.ImmutableArray&lt;T&gt;</code>,
   <code>ImmutableHashSet&lt;T&gt;</code>, ... </li>
 </ul>
+<p>Note: although <code>string.Intern</code> method has a side effect, ignoring its return value is still suspicious as it is the only reference
```
andrei-epure-sonarsource: I believe you should either use "although **the** `string.Intern` method has a side effect" OR "although `string.Intern` has a side effect" (preferably the 2nd as it has less words)

christophe-zurn-sonarsource: I actually had the same reflection when modifying RSPEC, and ended up confusing myself 😅.
Updated RSPEC to 2nd solution

Issue #2324, opened 2019-03-07
Add an option to enable the analysis of generated code


nicolas-harraudeau-sonarsource:
### Problem

We currently ignore some generated files during analysis (see #905). This is a hardcoded behavior.

In the following cases this behavior can be problematic:
* a file contains both generated and non-generated code.
* developers want to know if the generated code contains issues.

### Solution

In order to provide an alternative for these users we will add an option to the dotnet analyzers. This option will enable the analysis of generated files. By default the behavior will remain the same, i.e. generated files won't be analyzed.

If the option is enabled, it is up to the developers to add the exclusions manually.

### Priority

This is a low priority issue. We do not intend to have it for Sonarqube LTS.

andrei-epure-sonarsource:
also related to #2228

PR #2436, opened 2019-05-28
Analyze generated code in dotnet plugins

andrei-epure-sonarsource:
second bullet point in #2324

I recommend reviewing commit by commit


andrei-epure-sonarsource:
The quality gate is failing due to pre-existing lack of coverage. That specific code is actually covered by analyzer (C#) and plugin (Java) integration tests.

sonar-dotnet-shared-library/pom.xml
```diff
@@ -14,7 +14,7 @@
   <description>Share code between C# and VB.NET plugins</description>
 
   <properties>
-    <sonar.version>7.4-alpha1</sonar.version>
+    <sonar.version>7.6</sonar.version>
```
andrei-epure-sonarsource: I have updated the API just to have the latest Dependency Injection related code (see the commit "Read the analyzeGenerateCode setting in the plugin").

One could argue I could drop this change from this PR and make it separately.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/RoslynProfileExporter.java
```diff
@@ -217,6 +229,15 @@ private String analysisSettings(boolean includeSettings, boolean ignoreHeaderCom
       appendLine(sb, "  </Settings>");
     }
 
+    if (includeSettings) {
```
andrei-epure-sonarsource: I could have avoided this change, but until we remove this file, it's best to be as "up to date" as possible, to avoid confusion

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -44,6 +44,8 @@ public abstract class UtilityAnalyzerBase : SonarDiagnosticAnalyzer
 
         protected string WorkDirectoryBasePath { get; set; }
 
+        protected virtual bool AnalyzeAutogeneratedCode { get; set; }
```
andrei-epure-sonarsource: all utility analyzers except the `FileMetadataAnalyzer` were ignoring generated code. This change makes them read the setting value and generate the metrics/cpd/symbol/token metadata for generated files as well

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractProjectConfiguration.java
```diff
@@ -46,15 +46,15 @@
 @ScannerSide
 @InstantiationStrategy(InstantiationStrategy.PER_PROJECT)
 @Phase(name = Phase.Name.POST)
-public abstract class AbstractConfiguration {
-  private static final Logger LOG = Loggers.get(AbstractConfiguration.class);
+public abstract class AbstractProjectConfiguration {
```
andrei-epure-sonarsource: I can split this commit in a separate PR if you want...

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/AbstractSolutionConfigurationTest.java
```diff
@@ -0,0 +1,63 @@
...
+  public void analyzeGeneratedCodeIsTrue() {
+    AbstractSolutionConfiguration config = new AbstractSolutionConfiguration(
+      createSettings("true").asConfig(),
+      "cs") {
+    };
+    assertThat(config.analyzeGeneratedCode()).isTrue();
+  }
+
+  @Test
+  public void analyzeGeneratedCodeIsFalse() {
```
andrei-epure-sonarsource: this has been moved here from `AbstractProjectConfigurationTest.java`

its/src/test/java/com/sonar/it/csharp/AutoGeneratedTest.java
```diff
@@ -47,35 +53,96 @@ public void init() {
...
+      }
 
       TestUtils.runMSBuild(orchestrator, projectDir, "/t:Rebuild");
 
       orchestrator.executeBuild(TestUtils.newScanner(projectDir)
         .addArgument("end"));
     }
+
+  private static void setAnalyzeGeneratedCodeToTrue(Path projectDir) throws IOException {
+    // FIXME unfortunately, setting the property inside SQ does not propagate in due time to these files
```
andrei-epure-sonarsource: S4MSB queries the SQ API to get the settings and creates the 2 XML files. Unfortunately, during the ITs, these files do not contain the up-to-date `sonar.cs.analyzeGeneratedCode` setting, and this is why I am manually changing the file.

When testing manually, this problem does not occur.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/CoverageReportImportSensorTest.java
```diff
@@ -124,7 +124,7 @@ public void execute_warn_about_deprecated_integration_tests() throws IOException
     new CoverageReportImportSensor(coverageConf, coverageAggregator, "cs", "C#", true)
       .execute(context);
 
-    assertThat(logTester.logs(LoggerLevel.WARN)).containsOnly("Starting with SonarQube 6.2 separation between Unit Tests and Integration Tests "
+    assertThat(logTester.logs(LoggerLevel.WARN)).contains("Starting with SonarQube 6.2 separation between Unit Tests and Integration Tests "
```
duncanp-sonar: What other messages are now being produced as a result of this commit?

sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/CSharpPlugin.java
```diff
@@ -60,6 +60,7 @@ public void define(Context context) {
       DotNetSensor.class,
       CSharpConfiguration.class,
       CSharpGlobalProtobufFileProcessor.class,
+      CSharpSolutionConfiguration.class,
```
duncanp-sonar: What's the difference between this class and the existing `CSharpConfiguration.class`?
Are the settings for the new class per-SQ-project, whereas the settings for the `CSharpConfiguration.class` are per-SQ-module?

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/GeneratedFileFilter.java
```diff
@@ -32,13 +32,23 @@
   private static final Logger LOG = Loggers.get(GeneratedFileFilter.class);
 
   private final AbstractGlobalProtobufFileProcessor globalReportProcessor;
+  private final boolean analyzeGeneratedCode;
 
-  public GeneratedFileFilter(AbstractGlobalProtobufFileProcessor globalReportProcessor) {
+  public GeneratedFileFilter(AbstractGlobalProtobufFileProcessor globalReportProcessor, AbstractSolutionConfiguration configuration) {
```
duncanp-sonar: I'm confused. The new `analyzeGeneratedCode` settings has been added to both the existing `AbstractConfiguration` and the new `AbstractSolutionConfiguration`, but this filter only applies to the new class. Why is that?

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/RoslynProfileExporter.java
```diff
@@ -286,8 +307,12 @@ private static void appendParameters(StringBuilder sb, Map<String, String> param
     return result;
   }
 
-  private List<String> allRuleKeysByRepositoryKey(String repositoryKey) {
+  private List<String> allRuleKeysByRepositoryKey(@Nullable String repositoryKey) {
```
duncanp-sonar: Why is this nullable now but not before?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
duncanp-sonar: Naming consistency: why is this property called `Autogenerated` code rather than just `Generated`?

its/src/test/java/com/sonar/it/csharp/AutoGeneratedTest.java
duncanp-sonar: Let's have a chat about this to see if we can work out what is going on (having a test that we don't understand is a red flag).

From the look of it, I would expect `orchestrator.executeBuild(...).setProperty` to set a command line argument that is passed to the scanner. I'd expect the value for that property to be used by the scanner-cli, as any command line property will should override the server-side property value. I can't see any code seems to be setting the property value on the server.

However, I think there might be a bug in the way the Scanner for MSBuild generates the `SonarLint.xml` file and analysis config file - it may only be using the values from the server, and not overriding them with the value from the command line.

So, my guess (I haven't debugged through) is that what is happening is:
* this test is setting the value on the command line, not on the server.
* this Scanner for MSBuild isn't writing the correct value to either the `SonarLint.xml` file or the analysis config file.
* the code in this method is fixing up both files with the correct value.
* the `sonar-project.properties` file is generated from the analysis config file, so it contains the correct value.
* the scanner-cli API is querying for the setting value, and is correctly overriding the default from the server with the value from the properties file.

sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/CSharpPlugin.java
henryju: @duncanp-sonar yes, that's the idea.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractConfiguration.java
```diff
@@ -42,6 +44,8 @@
 import static org.sonarsource.dotnet.shared.plugins.AbstractPropertyDefinitions.getRoslynJsonReportPathProperty;
 
 @ScannerSide
+@InstantiationStrategy(InstantiationStrategy.PER_PROJECT)
+@Phase(name = Phase.Name.POST)
```
henryju: Remove this Phase annotation, it is useless.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/GeneratedFileFilter.java
henryju: I agree, to make things clearer I would keep only the getter in `AbstractSolutionConfiguration`

andrei-epure-sonarsource: sorry, I already did that in commit _Add AbstractSolutionConfiguration UT (e81d89f507116f75a41abf58bfbba665582271ac)_

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/RoslynProfileExporter.java
andrei-epure-sonarsource: it was before as well, triggering `squid:S4449`. I just cleaned it as I was modifying the file...

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/CoverageReportImportSensorTest.java
andrei-epure-sonarsource: >"Failed to set working directory hidden: C:\Users\Andrei\AppData\Local\Temp\junit4479429460697924433\junit8811176124595669393\.sonar"

This is happening inside [AbstractProjectOrModule](https://github.com/SonarSource/sonarqube/blob/master/sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/AbstractProjectOrModule.java#L101) when it tries to
```
        Files.createDirectories(workingDir);
        Files.setAttribute(workingDir, "dos:hidden", true, LinkOption.NOFOLLOW_LINKS);
```

its/src/test/java/com/sonar/it/csharp/AutoGeneratedTest.java
andrei-epure-sonarsource: as discussed with @duncanp-sonar , this problem may be related to SonarSource/sonar-scanner-msbuild#699

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/CoverageReportImportSensorTest.java
henryju: @dbmeneses could you please help to investigate this `Failed to set working directory hidden`. Is it fixed in more recent versions of the sonar-plugin-api?

its/src/test/java/com/sonar/it/csharp/AutoGeneratedTest.java
andrei-epure-sonarsource: In the end I decided to use the SQ WS API

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/CoverageReportImportSensorTest.java
dbmeneses: It's done in the scanner-engine. At one point it was failing sometimes when the directory didn't exist and it was fixed. @andrei-epure-sonarsource if possible, could you run it with verbose to get a stack trace?

andrei-epure-sonarsource: @dbmeneses it's because I was using SQ API 7.6 instead of 7.7. Updating to 7.7 removed the extra warnings

PR #2438, opened 2019-05-28
Add integration test for counting header comments

andrei-epure-sonarsource:
This is quite hard to test inside the C# analyzer, as the `SonarLint.xml` file is passed as _Additional file_ to roslyn by S4MSB, which is quite hard to mock.

### DIFF
--- a/its/src/test/java/com/sonar/it/csharp/MetricsIncludeHeaderCommentTest.java
+++ b/its/src/test/java/com/sonar/it/csharp/MetricsIncludeHeaderCommentTest.java
@@ -1,0 +1,201 @@
+/*
+ * SonarSource :: C# :: ITs :: Plugin
+ * Copyright (C) 2011-2019 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


its/src/test/java/com/sonar/it/csharp/MetricsTest.java
```diff
@@ -305,38 +304,38 @@ public void complexityDistributionAtFileLevel() {
...
     assertThat(value).contains("32=1");
     assertThat(value).contains("33=1");
     assertThat(value).contains("34=1");
+    assertThat(value).contains("35=1");
+    assertThat(value).contains("36=1");
+    assertThat(value).contains("37=1");
+    assertThat(value).contains("38=1");
 
-    assertThat(value.length()).isEqualTo(143); // No other line
+    assertThat(value.length()).isEqualTo(144); // No other line
```
andrei-epure-sonarsource: this is because there's one value line with two-digit line number (as the header shifted the positions)

its/src/test/java/com/sonar/it/csharp/MetricsIncludeHeaderCommentTest.java
```diff
@@ -0,0 +1,201 @@
...
+
+  @ClassRule
+  public static RuleChain chain = getRuleChain();
+
+  private static RuleChain getRuleChain() {
+    assertThat(SystemUtils.IS_OS_WINDOWS).withFailMessage("OS should be Windows.").isTrue();
+
+    // Scanner for MSBuild caches the analyzer, so running the test twice in a row means the old binary is used.
+    // This code deletes the cache, but there should be a way to run without cache.
+    // Ticket: https://jira.sonarsource.com/browse/SONARMSBRU-346
```
andrei-epure-sonarsource: this has actually been closed as "Won't fix" https://jira.sonarsource.com/browse/SONARMSBRU-346

PR #2439, opened 2019-06-04
[moved] Read analyzeGeneratedCode setting in analyzers

andrei-epure-sonarsource:
first bullet point in #2324


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpDiagnosticAnalyzerContextHelper.cs
```diff
@@ -115,10 +115,11 @@ internal static class CSharpDiagnosticAnalyzerContextHelper
 
         #region SyntaxTree.IsGenerated
 
-        internal static bool IsGenerated(this SyntaxTree tree,
+        internal static bool ShouldAnalyze(this SyntaxTree tree, AnalyzerOptions options,
```
duncanp-sonar: The #region description in lines 116 and 125 is not longer accurate.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/PropertiesHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+        internal const string AnalyzeGeneratedCodeCSharp = "sonar.cs.analyzeGeneratedCode";
+        internal const string AnalyzeGeneratedCodeVisualBasic = "sonar.vbnet.analyzeGeneratedCode";
+
+        // TODO should make this a singleton class or make sure somehow to not read every time the setting
+
+        public static bool ShouldAnalyzeGeneratedCode(this AnalyzerOptions options, string language)
+        {
+            return ReadProperty(GetSettings(options), GetProperty());
+
+            string GetProperty() =>
```
duncanp-sonar: I don't think an nested function adds much here; it's more convoluted and less readable. Also, `GetPropertyName` would  be a more accurate name.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/PropertiesHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace SonarAnalyzer.Helpers
+{
+    internal static class PropertiesHelper
+    {
+        internal const string AnalyzeGeneratedCodeCSharp = "sonar.cs.analyzeGeneratedCode";
+        internal const string AnalyzeGeneratedCodeVisualBasic = "sonar.vbnet.analyzeGeneratedCode";
+
+        // TODO should make this a singleton class or make sure somehow to not read every time the setting
```
duncanp-sonar: Perhaps this could be part of the `IAnalyzerConfiguration`

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/PropertiesHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        private static IEnumerable<XElement> GetSettings(AnalyzerOptions options)
+        {
+            var sonarLintAdditionalFile = options.AdditionalFiles
+                .FirstOrDefault(f => ParameterLoader.IsSonarLintXml(f.Path));
+
+            if (sonarLintAdditionalFile == null)
+            {
+                return new List<XElement>();
```
duncanp-sonar: `Enumerable.Empty<XElement>` would be better.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/PropertiesHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+            }
+
+            var xml = XDocument.Load(sonarLintAdditionalFile.Path);
```
duncanp-sonar: The code that reads from the XML file is inconsistently defensive; it guards against some conditions, but not other e.g. it won't fail if a property value isn't a boolean, but it will fail if the file isn't XML.
It would be safer to be defensive on this line too.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpDiagnosticAnalyzerContextHelper.cs
```diff
@@ -115,10 +115,11 @@ internal static class CSharpDiagnosticAnalyzerContextHelper
...
             Compilation compilation)
         {
-            return tree.IsGenerated(CSharp.CSharpGeneratedCodeRecognizer.Instance, compilation);
+            return !tree.IsGenerated(CSharp.CSharpGeneratedCodeRecognizer.Instance, compilation) ||
+                options.ShouldAnalyzeGeneratedCode(LanguageNames.CSharp);
```
duncanp-sonar: TODO: once the options check has been optimized, it would be better to do the options check first as it is a global check rather than a per-file check. It would also be worth recording this in a comment e.g.

` // PERF: do the global setting check before the per-file check. `

duncanp-sonar: Also, couldn't this method call `DiagnosticAnalyzerContextHelper.ShouldAnalyze` instead of repeating the same checks?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/PropertiesHelperTest.cs
```diff
@@ -0,0 +1,76 @@
...
+        public void ShouldAnalyzeGeneratedCode_WithTrueSetting_ReturnsTrue()
+        {
+            var result = GetSetting("ResourceTests\\AnalyzeGenerated\\SonarLint.xml");
+            result.Should().BeTrue();
+        }
+
+        [TestMethod]
+        public void ShouldAnalyzeGeneratedCode_WithFalseSetting_ReturnsFalse()
+        {
+            var result = GetSetting("ResourceTests\\SonarLint.xml");
```
duncanp-sonar: It's usually better to avoid sharing static resource files between different tests to improve the test isolation.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -40,7 +40,7 @@ internal static class DiagnosticAnalyzerContextHelper
             context.RegisterSyntaxNodeAction(
                 c =>
                 {
-                    if (!c.GetSyntaxTree().IsGenerated(generatedCodeRecognizer, c.SemanticModel.Compilation))
+                    if (ShouldAnalyze(generatedCodeRecognizer, c.GetSyntaxTree(), c.Compilation, c.Options))
```
duncanp-sonar: Need at least one integration test (in C#) that check the end-to-end analysis i.e. that the analyzers actually analyze/skip based on the setting in SonarLint.xml.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/PropertiesHelper.cs
duncanp-sonar: Update: I see there's a malformed XML test. I don't understand why `XDocument.Load` isn't throwing with that input. I've just tried calling `XDocument.Load` with a missing file, an empty file, and the content of the malformed XML file, and it threw in every case.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/DiagnosticAnalyzerContextHelper.cs
andrei-epure-sonarsource: fair point, I only added one for the plugin

andrei-epure-sonarsource:
moved to  #2443

PR #2445, opened 2019-06-11
S2077: Fix False Negative for + operator

### LINKED ISSUE #2417: False negative on Rule S2077: the use of operator + should raise an issue in VB.Net
Rule S2077 should raise when the SQL query is concatenated using the `+` operator in VB.Net.

christophe-zurn-sonarsource:
Fix #2417


sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/ExecutingSqlQueries.cs
```diff
@@ -95,25 +95,29 @@ protected override bool IsStringMethodInvocation(string methodName, ExpressionSy
         private static bool AllConstants(List<ArgumentSyntax> arguments, SemanticModel semanticModel) =>
             arguments.All(a => a.GetExpression().IsConstant(semanticModel));
 
+        private static bool IsConcatenationOperator(SyntaxNode node)
```
andrei-epure-sonarsource: this can use `=>` syntax to improve readability

Issue #420, opened 2017-06-12
S3897 is an unsafe suggestion in my opinion


jabbera:
I posted this in the google groups, but no one responded. Through I might get more traction here.

Hi all,

I wanted to discuss: S3897 (https://sonarcloud.io/organizations/default/rules#rule_key=csharpsquid%3AS3897)

I think it is an unsafe suggestion on it's surface if the class is not sealed. Especially as a major. The reason is fairly well described here: http://blog.mischel.com/2013/01/05/inheritance-and-iequatable-do-not-mix/

But the important quote: A class that implements IEquatable<T> is saying, “I know how to compare two instances of type T or any type derived from T for equality.” After all, if type B derives from A, then B is-a A. That’s what inheritance means! The problem with implementing IEquatable<T> on a non-sealed type is that a base class cannot know how to compare instances of derived types. Trying to implement IEquatable<T> on a non-sealed class ends up producing inconsistent results when comparing instances of derived types.

Here is an example.

```
class Base : IEquatable<Base> {
  bool Equals(Base b) => return true;
}

class A : Base, IEquatable<A> {
  bool Equals(A a) => return false;
}

class B : Base, IEquatable<B> {
  bool Equals(B b) => return false;
}

A a = new A();
B b = new B();

Console.WriteLine(a.Equals(b)); // This should obviously return false as they are different classes however because of the existance of Base:Equals(Base b) it will return true.

```

Those classes are obviously not equal but implementing the suggestion would make the runtime behave as such.

Thanks,
Mike

Evangelink:
Hi @jabbera,

Indeed we have seen your thread on Google Group but we haven't had the opportunity to have an internal discussion meeting on this subject. At first glance your suggestion make sense but we could also allow a virtual implementation of the equal. So we have to have a little talk to see whether we can think of a best solution.

Cheers,
Amaury

jabbera:
Thanks for getting back to me! Glad to know it's being looked at. 

Cheers,
Mike

Evangelink:
Hi @jabbera!

We have been discussing this subject with the team and we came up with the following solution:
- Keep this rule as it is
- Add a new rule (description [here](https://jira.sonarsource.com/browse/RSPEC-4035)) to recommend sealing the class.

We have decided to keep the rule as it is because it suggests you to implement `IEquatable<T>` only when you override `Equals(object)` or have created a `Equals(T)` method. On the example you gave, overriding `Equals(object)` will lead to the same problem.

[...]

PR #2449, opened 2019-06-12
S3897: FP due to not checking IEquatable in inheritance chain

### LINKED ISSUE #1960: Fix S3897: FP due to not checking IEquatable in inheritance chain
I need to "override" `Equals(T other)`, i.e., hide base implementation with new:

```
public sealed class PersonNames : ValueObject<PersonNames>
{
	public PersonNames(string firstName, string middleName, string lastName)
	{
		FirstName = firstName;
		MiddleName = middleName;
		LastName = lastName;
	}

	public string FirstName { get; }
	public string LastName { get; }
	public string MiddleName { get; }

	public new bool Equals(PersonNames other)
	{
		return string.Equals(FirstName, other.FirstName, StringComparison.OrdinalIgnoreCase)
			&& string.Equals(MiddleName, other.MiddleName, StringComparison.OrdinalIgnoreCase)
			&& string.Equals(LastName, other.LastName, StringComparison.OrdinalIgnoreCase);
	}

	public override string ToString()
	{
		return $"{{{FirstName},{MiddleName},{LastName}}}";
	}
}

public abstract class ValueObject<T> : IEquatable<T>
        where T : ValueObject<T>
{
...

	public bool Equals(T other)
	{
		...
	}
}
```

SonarC# 7.4.0.6452 says me that I should implement `IEquatable<T>`

christophe-zurn-sonarsource:
Fixes #1960
Duplication is from already existing code


sonaranalyzer-dotnet/rspec/cs/S3897_c#.html
```diff
@@ -5,27 +5,13 @@
 <h2>Noncompliant Code Example</h2>
 <pre>
 class MyClass  // Noncompliant
-{
-  public override bool Equals(object other)
```
andrei-epure-sonarsource: MyClass does not implement IEquatable. Why did you remove this example?

christophe-zurn-sonarsource: It follows the discussion here: https://github.com/SonarSource/sonar-dotnet/issues/420
Implementation was already updated not to raise issue in that case, only the code examples were not up-to-date

PR #2450, opened 2019-06-12
S2219 FP on 'is object' used as null-check

### LINKED ISSUE #2424: S2219 FP: `is object` is not always allowed for null-check
### Description

`is object` is the canonical non-null test in C#, but in SonarQube it throws a clumsy warning.

### Repro steps

```
if (obj is object)
{
....
}
```

Throws a "Use a 'null' check instead"

### Expected behavior

Should not throw a null check exception.

### Actual behavior

Throws a "Use a 'null' check instead"

### Known workarounds

none

### Related information

More information here https://twitter.com/FransBouma/status/1115201371182071808 but essentually... using `is object` isstead of != null gives better IL and has less side effects.

christophe-zurn-sonarsource:
Fixes #2424

### DIFF
--- a/sonaranalyzer-dotnet/rspec/cs/S2219_c#.html
+++ b/sonaranalyzer-dotnet/rspec/cs/S2219_c#.html
@@ -86,6 +86,6 @@ condition = i is int?; // true
 <p>No issue is reported on the following expressions:</p>
 <ul>
   <li> <code>expr is T</code>&nbsp;when either operand of the <code>is</code> operator is a value type. In that case CS0183 or CS0184 reports </li>
-  <li> <code>expr is object</code>, as this is a common and efficient pattern to do non-null checks </li>
+  <li> <code>expr is object</code>, as this is a common and efficient pattern to do null checks </li>
 </ul>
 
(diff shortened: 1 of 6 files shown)


sonaranalyzer-dotnet/rspec/cs/S2219_c#.html
```diff
@@ -83,6 +83,9 @@ <h2>Exceptions</h2>
 bool condition = i.GetType() == typeof(int?); // false;
 condition = i is int?; // true
 </pre>
-<p>No issue is reported on <code>expr is T</code> expressions if either operand of the <code>is</code> operator is a value type. In that case CS0183
-or CS0184 reports.</p>
+<p>No issue is reported on the following expressions:</p>
+<ul>
+  <li> <code>expr is T</code>&nbsp;when either operand of the <code>is</code> operator is a value type. In that case CS0183 or CS0184 reports </li>
+  <li> <code>expr is object</code>, as this is a common and efficient pattern to do non-null checks </li>
```
andrei-epure-sonarsource: s/non-null checks/null checks

PR #2452, opened 2019-06-13
Avoid failing analysis when XML imports fails

### LINKED ISSUE #2259: Trx parsing fails: Illegal character entity: expansion character (code 0xc
### Description

Original post:
https://github.com/SonarSource/sonar-scanner-msbuild/issues/652

I am using the sonarcloud 1.5.1 vats plugin but analysis fails due to a problem reading the trx file.

From log
```
Error while parsing the XML file: D:\a\1\s\TestResults\VssAdministrator_fv-az40 2019-01-09 21_39_02.trx

ERROR: Caused by: Illegal character entity: expansion character (code 0xc

at [row,col {unknown-source}]: [1179,22]
```

For full log and all steps see https://svensson-daniel.visualstudio.com/OpenRiaServices/_build/results?buildId=239

I have searched a bit for the error message and it might be the case that the XML needs to be opened as XML 1.1 and not 1.0

### Repro steps

Try to parse trx file attached:
[VssAdministrator_fv-az40.2019-01-09.21_39_02.zip](https://github.com/SonarSource/sonar-dotnet/files/2821514/VssAdministrator_fv-az40.2019-01-09.21_39_02.zip)

Long steps:
1. Fork openriaserices/openriaservices at github
2. Edit azure pipelines yaml so that it uses vstest v1 instead of v2
3. Build via azure pipelines (need to setup a sonarcloud endpoint first)

### Expected behavior

I expect 
1. analysis to be uploaded to sonarcloud without issues.
2. Trx parsing to succeed

### Actual behavior

Trx reading fails and while upload analysis is aborted.

### Known workarounds

None 

### Related information

* SonarC#
* Usage: vsts sonarcloud 1.5.1

christophe-zurn-sonarsource:
Fixes #2259

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/XmlParserHelper.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/XmlParserHelper.java
@@ -19,19 +19,25 @@
  */
 package org.sonar.plugins.dotnet.tests;
 
-import javax.annotation.Nullable;
-import javax.xml.stream.XMLStreamConstants;
-import javax.xml.stream.XMLStreamException;
-import javax.xml.stream.XMLStreamReader;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
-import java.io.InputStreamReader;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
+import java.util.Objects;
+import javax.annotation.Nullable;
+import javax.xml.stream.Location;
+import javax.xml.stream.XMLStreamConstants;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamReader;
+import org.sonar.api.utils.log.Logger;
+import org.sonar.api.utils.log.Loggers;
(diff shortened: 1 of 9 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/XmlParserHelper.java
```diff
@@ -60,17 +66,43 @@ void checkRootTag(String name) {
...
+        Location currentLocation = stream.getLocation();
+        if (isSameLocation(lastLocation, currentLocation)) {
+          // if the next() method throws exception before moving XML pointer forward, we fail here
+          LOG.warn("Unable to get next XML event while parsing file '{}'", file.toString());
+          throw e;
+        }
+        lastLocation = currentLocation;
+      }
+    }
+    return -1;
```
andrei-epure-sonarsource: this line could be covered by adding a test with an empty file

christophe-zurn-sonarsource: Empty file actually does not work here for testing this part. It actually throws an exception when encountering an unexpected EOF token and does not put the reader pointer forward, so the exception gets rethrown as it falls into the case handled by the catch clause.
I added mock tests instead for coverage

PR #2453, opened 2019-06-14
Use the AnalysisContext cache for the analyzeGeneratedCode setting value

andrei-epure-sonarsource:
4th bullet point in #2324


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -191,9 +176,21 @@ internal static class DiagnosticAnalyzerContextHelper
...
             = new ConditionalWeakTable<Compilation, ConcurrentDictionary<SyntaxTree, bool>>();
 
-        public static bool ShouldAnalyze(GeneratedCodeRecognizer generatedCodeRecognizer, SyntaxTree syntaxTree, Compilation c, AnalyzerOptions options) =>
-            !syntaxTree.IsGenerated(generatedCodeRecognizer, c) ||
-            options.ShouldAnalyzeGeneratedCode(c.Language);
+        private static bool ShouldAnalyze(SonarAnalysisContext context, GeneratedCodeRecognizer generatedCodeRecognizer, SyntaxTree syntaxTree, Compilation c, AnalyzerOptions options) =>
+            context.ShouldAnalyzeGenerated(c, options) ||
+            !syntaxTree.IsGenerated(generatedCodeRecognizer, c);
+
+        private static bool ShouldAnalyze(CompilationStartAnalysisContext context, GeneratedCodeRecognizer generatedCodeRecognizer, SyntaxTree syntaxTree, Compilation c, AnalyzerOptions options) =>
```
andrei-epure-sonarsource: I get a roslyn warning on this one (`'analysisContext' does not register any analyzer actions. Consider moving actions registered in 'Initialize' that depend on this start action to 'analysisContext'.`) but I don't really understand why

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -70,6 +72,21 @@ public class SonarAnalysisContext
         /// </remarks>
         public static Action<IReportingContext> ReportDiagnostic { get; set; }
 
+        public static bool ShouldAnalyzeGenerated(AnalysisContext analysisContext, Compilation c, AnalyzerOptions options) =>
+            TryGetSonarLintXml(options, out var sonarLintXml) &&
+            analysisContext.TryGetValue(sonarLintXml.GetText(), GetProvider(c.Language), out var shouldAnalyzeGeneratedCode) &&
+            shouldAnalyzeGeneratedCode;
+
+        public static bool ShouldAnalyzeGenerated(CompilationStartAnalysisContext analysisContext, Compilation c, AnalyzerOptions options) =>
```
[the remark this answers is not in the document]
duncanp-sonar: Me neither, even after reading the rule [description](https://github.com/mavasani/roslyn/commit/7824bb4c1e226dfcc1f197a243458943ef5a5be8). I think it's an FP.
If I remember correctly, the standard pattern is that XXXStart and XXXEnd callbacks are used in pairs by stateful analyzers, with the idea being that in the XXXStart callback you register the `end` callback. Here, we've got the XXXStart analysis context but we aren't registering another callback -> RS1012 is raised (but with a message that doesn't make much sense).

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -70,6 +72,21 @@ public class SonarAnalysisContext
...
+
+        public static bool ShouldAnalyzeGenerated(CompilationStartAnalysisContext analysisContext, Compilation c, AnalyzerOptions options) =>
+            TryGetSonarLintXml(options, out var sonarLintXml) &&
+            analysisContext.TryGetValue(sonarLintXml.GetText(), GetProvider(c.Language), out var shouldAnalyzeGeneratedCode) &&
+            shouldAnalyzeGeneratedCode;
+
+        public static bool ShouldAnalyzeGenerated(CompilationAnalysisContext analysisContext, Compilation c, AnalyzerOptions options) =>
+            TryGetSonarLintXml(options, out var sonarLintXml) &&
+            analysisContext.TryGetValue(sonarLintXml.GetText(), GetProvider(c.Language), out var shouldAnalyzeGeneratedCode) &&
+            shouldAnalyzeGeneratedCode;
```
duncanp-sonar: We need tests to check that the caching actually works.
I've taken the liberty of adding some in another commit that I've pushed to this branch.

andrei-epure-sonarsource: You, dear Sir, are a true gentleman!

duncanp-sonar: You're welcome :-)

PR #2454, opened 2019-06-14
S4058 FP on overloaded methods without a StringComparison alternative

### LINKED ISSUE #2099: S4058 is raised incorrectly in case of Xunit Assert.Contains
### Description

I got a [S4058](https://rules.sonarsource.com/csharp/RSPEC-4058) warning when using this Xunit method

`Assert.Contains(anyString, arrayOfString, StringComparer.OrdinalIgnoreCase)`

### Repro steps

```csharp
var returnedNames = new[] { "Any name", "Any other name" };
var expectedName = "Any name";

Assert.Contains(expectedName , returnedNames , StringComparer.OrdinalIgnoreCase);
```

### Expected behavior

No warning should be raised

### Actual behavior

An issue is raised as:

Warning	S4058	Change this call to 'Assert.Contains' to an overload that accepts a 'StringComparison' as a parameter.

Actually, XUnit does not have any overload of 'Assert.Contains' that takes 'StringComparision' as parameter.

### Related information

* SonarAnalyzer.CSharp - 7.8.0.7320
* Visual Studio - 15.8.9
* XUnit - 2.4.1

christophe-zurn-sonarsource:
Fixes #2099


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpOverloadHelper.cs
```diff
@@ -0,0 +1,111 @@
...
+            semanticModel.GetMemberGroup(invocation.Expression)
+                .OfType<IMethodSymbol>()
+                .Where(m => !m.GetAttributes(KnownType.System_ObsoleteAttribute).Any())
+                .Where(method => IsCompatibleOverload(invocation, method))
+                .Any(methodSignature => SameParametersExceptWantedType(methodSignature, GetInvocationParameters(invocation, semanticModel), types));
+
+        private static bool SameParametersExceptWantedType(IMethodSymbol possibleOverload, IEnumerable<IParameterSymbol> invocationParameters, ImmutableArray<KnownType> types)
+        {
+            var parametersWithoutWantedType = possibleOverload.GetParameters().Where(p => !p.Type.IsAny(types)).ToList();
+            var parametersWithoutWantedTypeCount = parametersWithoutWantedType.Count();
```
andrei-epure-sonarsource: you can use the IList `Count` property instead of the LINQ Count()

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SpecifyStringComparison.cs
```diff
@@ -35,7 +44,10 @@ void ValidCalls()
             "".StartsWith("", StringComparison.CurrentCulture); // Compliant
             "".StartsWith("", true, CultureInfo.CurrentCulture); // Compliant - CultureInfo implies string formatting
 
+            MyMethod1("", "");
```
andrei-epure-sonarsource: Also add `MyMethod1(""); // Noncompliant`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SpecifyStringComparison.cs
```diff
@@ -35,7 +44,10 @@ void ValidCalls()
...
             MyMethod2("");
+
+            Contains("", new[] { "" }, StringComparer.OrdinalIgnoreCase); // Compliant
```
andrei-epure-sonarsource: also add Noncompliant test

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SpecifyStringComparison.cs
```diff
@@ -8,12 +9,20 @@ class Program
         public void MyMethod(string value) { }
         public void MyMethod(string value, StringComparison comparisonType) { }
 
+        public void MyMethod1(string value1, string value2) { }
```
andrei-epure-sonarsource: should add tests with `params` as well

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SpecifyStringComparison.cs
christophe-zurn-sonarsource: I will let it like this here, as this is the real case that related to the issue (The `Contains` methods here have the exact same signatures as the ones defined by the framework).
I'll add another similar method that does have a `Noncompliant` variant as well

PR #2467, opened 2019-06-25
Add Jenkinsfile using jdk 11 and maven 3.6

tomverin:


Jenkinsfile
```diff
@@ -0,0 +1,83 @@
...
+          agent {
+            label 'vs2017'
+          }
+          steps {
+            runITs("DEV","")
+          }
+        }     
+        stage('LTS') {
+          agent {
+            label 'linux'
```
andrei-epure-sonarsource: I don't understand - why is LTS and SONARANALYZER linux?

andrei-epure-sonarsource:
Well, probably my question above is a silly one, as those axes are not failing. But there are failing ITs.

Issue #2474, opened 2019-07-03
Analysis of Entity Framework migration files never completes


duncanp-sonar:
### Description
Analysis of large files never completes.
See https://github.com/SonarSource/sonar-scanner-msbuild/issues/590

### TODO
Investigate which analysis rules are causing the issue. UPDATE: S1144 and S4487 (same implementation class)
~~Potentially related to https://github.com/SonarSource/sonar-dotnet/issues/2429, which relates to utility rules taking too long. However, the repro here is more extreme - effectively the build never completes and the user has to kill it.~~

### Repro
1. clone https://github.com/ksitarek/sonarscannersample
2. build the project using the sonar-dotnet analyzers

Note: the bug repros with just the file `20181005212624_InitialCreate.cs` in project `vtb.Warehouse.Data`

### Expected
The build completes successfully in a reasonable time.

### Actual
The build does not end.

### Known workarounds
A) Exclude the project from analysis by adding the `SonarQubeExclude` property:

```
  <PropertyGroup>
    <SonarQubeExclude>true</SonarQubeExclude> 
  </PropertyGroup>
```

or 

B) Disable rule S1144 (`Unused private types or members should be removed`) and rule S4487 (`Unread "private" fields should be removed`) (both implemented by `UnusedPrivateMember.cs`.

### Notes
The problematic project in the sample project is `vtb.Warehouse.Data`. This is an Entity Framework project that contains migrations, all of which are several thousand lines long.

duncanp-sonar:
Update: `S1144 Unused "private" methods should be removed` performs particularly badly against some of the files in the repro solution. Disabling this rule will improve analysis performance.

poggdogg:
I had some experience with this.  It appears to be hanging up on specific rules in very specific scenarios.  

By lots of elimination I was able to remove two "Sonar Way" rules and advised developers to inspect manually:
- Unread "private" fields should be removed
- Unused private types or members should be removed

I just hit the same issue again so I suspect there is another malformed rule.

PS - I hate to have to point this out, but...
The listed "known workaround" isn't really a work-around and very unhelpful.  You're asking users to not use your product for this stack.

It's like when someone says, "My car will start but my truck will not.  I need my truck to move a bed." 
The mechanic's work-around is, "Then don't use your truck. We don't think you don't need a bed that bad."

poggdogg:
[...]

PR #2475, opened 2019-07-04
Added a perf baseline unit test

duncanp-sonar:
Related to #2474


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Misc/MiscellaneousTests.cs
```diff
@@ -0,0 +1,245 @@
...
+{
+    [TestClass]
+    public class MiscellaneousTests
+    {
+        public TestContext TestContext { get; set; }
+
+        [TestMethod]
+        // This test takes around 40s-60s to run, which is as long as all of the other tests.
+        // This category allows us to exclude it from the set of tests to run.
+        // In the IDE: set the test filter in the Test Explorer to the following
```
duncanp-sonar: This test is currently enabled so it adds about 45 seconds to the unit test run (not a lot, but all of the other tests combined only take about 60 seconds). I've left it enabled by default for now.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Misc/MiscellaneousTests.cs
```diff
@@ -0,0 +1,245 @@
...
+        {
+            // The test file comes from a repro project for https://github.com/SonarSource/sonar-dotnet/issues/2474
+            // The single file is enough to cause the issue described in the bug, although there
+            // are other large files in the project that might also be problematic.
+
+            // Analyzers known to timeout against this test code
+            // (note: this test doesn't cover the utility analyzers):
+            var knownSlowAnalyzers = new[] { "SonarAnalyzer.Rules.CSharp.UnusedPrivateMember" };
+            
+            int analysisTimeoutInMs = 700;
```
duncanp-sonar: Any analyser that takes longer than 700ms will be aborted.
To stop the test from failing continually, there's a list of analyzers that are known to be slow.

The test will fail if either:
* any of the aborted analyzers are not in the "known slow analyzer" list, or
* a "known slow analyzer" is not in the list (i.e. that analyzer is now faster so the "slow list" needs to be updated)

This test is just a starting point, created to help diagnose #2474. It was enough to identify that only one of the "normal" (i.e. non-utility analysers) has a problem processing this file.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Misc/MiscellaneousTests.cs
```diff
@@ -0,0 +1,245 @@
...
+                var newExecutionResult = new ExecutionResult(analyzerType, timer.ElapsedMilliseconds, !succeeded);
+                executionResults.Add(newExecutionResult);
+
+                if (!succeeded)
+                {
+                    Debug.WriteLine($"   -> TIMED OUT");
+                }
+            }
+
+            DumpExecutionResults(executionResults);
```
duncanp-sonar: Writes the execution results to the test output, sorted by execution time (descending).

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -137,7 +138,8 @@ private static void DumpActualDiagnostics(IEnumerable<Diagnostic> diagnostics)
         }
 
         public static ImmutableArray<Diagnostic> GetAllDiagnostics(Compilation compilation,
-            IEnumerable<DiagnosticAnalyzer> diagnosticAnalyzers, CompilationErrorBehavior checkMode)
+            IEnumerable<DiagnosticAnalyzer> diagnosticAnalyzers, CompilationErrorBehavior checkMode,
+            CancellationToken? cancellationToken = null)
```
duncanp-sonar: Tweaked the method to support cancellation

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Misc/MiscellaneousTests.cs
```diff
@@ -0,0 +1,244 @@
...
+            return compilation;
+        }
+
+        private static IEnumerable<MetadataReference> GetReferencesNetCore(string entityFrameworkVersion) =>
+            Enumerable.Empty<MetadataReference>()
+                .Concat(FrameworkMetadataReference.Netstandard)
+                .Concat(NuGetMetadataReference.MicrosoftEntityFrameworkCoreSqlServer(entityFrameworkVersion))
+                .Concat(NuGetMetadataReference.MicrosoftEntityFrameworkCoreRelational(entityFrameworkVersion));
+
+        private Type[] GetSonarAnalyzerTypes(Assembly assembly)
```
andrei-epure-sonarsource: could be a local function inside `TestForSlowAnalyzers_EFMigrationSource_2474`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Misc/MiscellaneousTests.cs
```diff
@@ -0,0 +1,244 @@
...
+
+            return allAnalyzerTypes;
+        }
+
+        private bool IsSonarAnalyzer(Type type) =>
+            (!type.IsAbstract && IsDerivedFromSonarDiagnosticAnalyzer(type));
+
+        private static readonly Type SonarDiagnosticAnalyzerType = typeof(SonarDiagnosticAnalyzer);
+
+        private bool IsDerivedFromSonarDiagnosticAnalyzer(Type type)
```
andrei-epure-sonarsource: could be local function

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Misc/MiscellaneousTests.cs
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        private void AssertNoUnexpectedSlowAnalyzers(IEnumerable<ExecutionResult> executionResults, IEnumerable<string> expectedSlowAnalyzers)
+        {
+            var actualSlowAnalyzers = executionResults.Where(x => x.TimedOut).Select(x => x.AnalyzerType.FullName);
+            AssertNoUnexpectedSlowAnalyzers(actualSlowAnalyzers.Except(expectedSlowAnalyzers, StringComparer.OrdinalIgnoreCase));
+
+            var unexpectedlyFastAnalyzers = expectedSlowAnalyzers.Except(actualSlowAnalyzers, StringComparer.OrdinalIgnoreCase);
+
+            if (!unexpectedlyFastAnalyzers.Any())
```
andrei-epure-sonarsource: to avoid multiple negation ("unexpected fast" and "! Any()") which are harder to follow, I would swap the logic and use

```
if (unexpectedlyFastAnalyzers.Any()) {
	// ... fail the analysis due to faster than expected
} else {
	// ... Log the slow analyzers
}
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Misc/MiscellaneousTests.cs
```diff
@@ -0,0 +1,245 @@
...
+            {
+                Log(FormatListToMessage(expectedSlowAnalyzers, "The expected analyzers were slow:"));
```
andrei-epure-sonarsource: would it be better "The following analyzers were slow (as expected):"

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Misc/MiscellaneousTests.cs
```diff
@@ -0,0 +1,245 @@
...
+                return;
+            }
+
+            string unexpectedlyFastMessage = FormatListToMessage(unexpectedlyFastAnalyzers,
+                "The following analyzers where unexpectedly fast - the baseline may need to be updated:");
```
andrei-epure-sonarsource: would it be better : `The following analyzers where expected to be slow, but performed better than expected - the baseline may need to be updated:`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Misc/MiscellaneousTests.cs
```diff
@@ -0,0 +1,245 @@
...
+                if (ex.InnerExceptions.Any(x => x is OperationCanceledException))
+                {
+                    return false;
+                }
+                throw ex;
+            }
+            return true;
+        }
+
+        private void AssertNoUnexpectedSlowAnalyzers(IEnumerable<ExecutionResult> executionResults, IEnumerable<string> expectedSlowAnalyzers)
```
andrei-epure-sonarsource: this also asserts no unexpected fast analyzers, maybe it could be `AssertAnalyzerExpectedTime` ?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Misc/MiscellaneousTests.cs
```diff
@@ -0,0 +1,245 @@
...
+                Log($"  {result.AnalyzerType.FullName}, {result.ElapsedTimeInMs}ms  {(result.TimedOut ? "TIMED OUT" : "")}");
+            }
+        }
+
+        private static string FormatListToMessage(IEnumerable<string> list, string messagePrefix = null) =>
+            messagePrefix + Environment.NewLine
+                + string.Join(", " + Environment.NewLine, list);
+
+        private void Log(string message)
+        {
```
andrei-epure-sonarsource: nitpick: arrow notation?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Misc/MiscellaneousTests.cs
```diff
@@ -0,0 +1,244 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarAnalyzer.Helpers;
+using SonarAnalyzer.UnitTest.TestFramework;
+
+namespace SonarAnalyzer.UnitTest.Misc
+{
+    [TestClass]
+    public class MiscellaneousTests
```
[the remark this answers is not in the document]
duncanp-sonar: It evolved into a performance test; it wasn't one to start with. I'll rename the file and folder.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Misc/MiscellaneousTests.cs
duncanp-sonar: My assumption is that we'll add more performance tests against different sources as we find other problem cases so `GetSonarAnalyzerTypes` will be called from multiple tests.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Misc/MiscellaneousTests.cs
```diff
@@ -0,0 +1,244 @@
...
+        {
+            var allAnalyzerTypes = assembly.GetTypes()
+                .Where(IsSonarAnalyzer).ToArray();
+
+            Log($"Analyzer count: {allAnalyzerTypes.Length}");
+
+            return allAnalyzerTypes;
+        }
+
+        private bool IsSonarAnalyzer(Type type) =>
```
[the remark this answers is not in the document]
duncanp-sonar: Yes, it could be. I'm not really a fan of local functions; I don't find that they really help readability or maintainability.

I've made `IsDerivedFromSonarDiagnosticAnalyzer` a local function.

PR #2477, opened 2019-07-08
Integration of CBDE and the rule S3949 using it

loic-joly-sonarsource:
We only generate code for a small subset of C# programs...


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -0,0 +1,328 @@
...
+            Environment.GetEnvironmentVariable("CIRRUS_WORKING_DIR") ?? // For Cirrus
+            Environment.GetEnvironmentVariable("WORKSPACE") ?? // For Jenkins
+            Path.GetTempPath(); // By default
+        private static readonly string mlirProcessSpecificPath = Path.Combine(mlirPath, $"CBDE_{Process.GetCurrentProcess().Id}");
+        private static readonly string mlirLogFile = Path.Combine(mlirProcessSpecificPath, "cbdeHandler.log");
+        private static readonly string mlirMetricsLogFile = Path.Combine(mlirProcessSpecificPath, "metrics.log");
+        private static readonly string mlirPerfLogFile = Path.Combine(mlirProcessSpecificPath, "performances.log");
+        private static void GlobalLog(string s)
+        {
+            lock (logFileLock)
```
andrei-epure-sonarsource: should you do this only when it runs in DEBUG mode?
or when some property is turned on (and it should be turned off by default, in PROD)?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -0,0 +1,328 @@
...
+                        }
+                    }
+                    catch(Exception e)
+                    {
+                        GlobalLog("An exception has occured: " + e.Message + "\n" + e.StackTrace);
+                        throw;
+                    }
+                });
+        }
+        private string ManglePath(string path)
```
andrei-epure-sonarsource: it would be nice to comment why this is necessary

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.ControlFlowGraph;
+using SonarAnalyzer.ControlFlowGraph.CSharp;
+using SonarAnalyzer.Helpers;
+using SonarAnalyzer.ShimLayer.CSharp;
+
+namespace SonarAnalyzer
+{
+    public static class SyntaxNodeExtension
```
andrei-epure-sonarsource: if this is public, it should be in its own file
why not put this together with the other extension methods?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
+﻿using System;
```
andrei-epure-sonarsource: missing file header (not the only one)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+        }
+
+        private bool HasNoReturn(MethodDeclarationSyntax method)
+        {
+            return semanticModel.GetTypeInfo(method.ReturnType).Type.SpecialType == SpecialType.System_Void;
+        }
+
+        private void ExportBlock(Block block, MethodDeclarationSyntax parentMethod, string functionReturnType)
+        {
+            writer.WriteLine($"^{BlockId(block)}: // {block.GetType().Name}"); // TODO: Block arguments...
```
andrei-epure-sonarsource: are you going to address this TODO?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+                    {
+                        Debug.Assert(bbb.BranchingNode.Kind() == SyntaxKind.ForStatement);
+                        writer.WriteLine($"br ^{BlockId(bbb.TrueSuccessorBlock)}");
+                    }
+                    else
+                    {
+                        var id = EnforceBoolOpId(cond as ExpressionSyntax);
+                        writer.WriteLine($"cond_br %{id}, ^{BlockId(bbb.TrueSuccessorBlock)}, ^{BlockId(bbb.FalseSuccessorBlock)} {GetLocation(cond)}");
+                    }
+                    /*
```
andrei-epure-sonarsource: please remove the commented code and eventually add it to a CBDE ticket

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+            var args = method.ParameterList.Parameters.Select(
+                p => {
+                    ++paramCount;
+                    var paramName = string.IsNullOrEmpty(p.Identifier.ValueText) ?
+                        ".param" + paramCount.ToString() :
+                        EncodeName(p.Identifier.ValueText);
+                    return $"%{paramName} : {MLIRType(p)}";
+                }
+                );
+            return '(' + string.Join(", ", args) + ')';
```
andrei-epure-sonarsource: use string interpolation instead? (applies to all document)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CbdeHandlerRule.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Runtime.InteropServices;
+using System.Diagnostics.CodeAnalysis;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(S3949DiagnosticId)]
+    public sealed class CbdeHandlerRule : SonarDiagnosticAnalyzer
+    {
+        // Note: For now, this rule actually runs only under windows and outside of SonarLint
```
andrei-epure-sonarsource: is this comment related to the field `unitTest` ? if not, should it be above the class?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CbdeHandlerRule.cs
```diff
@@ -0,0 +1,106 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(ruleS3949);
+
+        private readonly ImmutableDictionary<string, DiagnosticDescriptor> ruleIdToDiagDescriptor = ImmutableDictionary<string, DiagnosticDescriptor>.Empty
+            .Add("S3949", ruleS3949);
+        public CbdeHandlerRule() : this(false) {}
+        private CbdeHandlerRule(bool unitTest)
+        {
+            this.unitTest = unitTest;
+        }
+        public static CbdeHandlerRule MakeUnitTestInstance()
```
andrei-epure-sonarsource: should this only be exposed to the unit test library, rather than being public?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CbdeHandlerRule.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        internal const string ConfigurationAdditionalFile = "ProjectOutFolderPath.txt";
+
+        [ExcludeFromCodeCoverage]
+        internal static bool IsProjectOutput(AdditionalText file) =>
+           ParameterLoader.ConfigurationFilePathMatchesExpected(file.Path, ConfigurationAdditionalFile);
+
+        // The logic used here is based on detecting a file which is present only when not run from SonarLint
+        // The logic is copied from FirstOrDefault(IsProjectOutput)
```
andrei-epure-sonarsource: could you add a `/// <see cref=/>` comment to point to the exact namespace+class+member you copied this from?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
```diff
@@ -0,0 +1,107 @@
...
+    <CBDEFolder>$(MSBuildProjectDirectory)\$(RelativeCBDEFolder)</CBDEFolder>
+    <SpecificCBDEVersion>1.0.0.13341</SpecificCBDEVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <CBDEUrls Include="http://repox.jfrog.io/repox/sonarsource-public-builds/org/sonarsource/cbde/$(SpecificCBDEVersion)/cbde-windows.$(SpecificCBDEVersion).zip" >
+      <SubFolder>windows\</SubFolder>
+      <MainExe>dotnet-symbolic-execution.exe</MainExe>
+      <ZipFileName>cbde-windows.zip</ZipFileName>
+    </CBDEUrls>
+    <!-- Only Windows is supported for now
```
andrei-epure-sonarsource: I guess the best is to remove this ? (having commented out logic is not great)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
```diff
@@ -0,0 +1,107 @@
...
+
+    <!-- Specifies the CBDE files that should be included in the project and copied to the output directory -->
+    <ItemGroup>
+      <Content Include="$(RelativeCBDEFolder)\windows\**\*.exe" >
+        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      </Content>
+      <Content Include="$(RelativeCBDEFolder)\windows\**\*.txt" >
+        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      </Content>
+      <Content Include="$(RelativeCBDEFolder)\linux\**" Exclude="$(RelativeCBDEFolder)\linux\**\*.zip">
```
andrei-epure-sonarsource: same here, should be removed before merging to master

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
```diff
@@ -0,0 +1,107 @@
...
+       these files are missing. Duplication the item definition here avoid this situation
+       -->
+  <ItemGroup>
+    <Content Include="$(RelativeCBDEFolder)\windows\**\*.exe" >
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="$(RelativeCBDEFolder)\windows\**\*.txt" >
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="$(RelativeCBDEFolder)\linux\**" Exclude="$(RelativeCBDEFolder)\linux\**\*.zip">
```
andrei-epure-sonarsource: remove?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
```diff
@@ -0,0 +1,107 @@
...
+    </Content>
+  </ItemGroup>
+
+  
+
+
+  <Target Name="FetchCBDEBinaries" BeforeTargets="BeforeBuild" Outputs="%(CBDEUrls.SubFolder)/%(CBDEUrls.MainExe)">
+    <PropertyGroup>
+      <PowerShellExe Condition="'$(PowerShellExe)'==''">%WINDIR%\System32\WindowsPowerShell\v1.0\powershell.exe</PowerShellExe>
+      <UnzipScript>$(MSBuildProjectDirectory)\..\..\..\Scripts\unzip.ps1</UnzipScript>
```
andrei-epure-sonarsource: there are built-in MSBuild tasks for 
- Download https://docs.microsoft.com/en-us/visualstudio/msbuild/downloadfile-task?view=vs-2019 
- Unzip https://docs.microsoft.com/en-us/visualstudio/msbuild/unzip-task?view=vs-2019

I think it's better to use that rather then maintaining a set of scripts

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/CBDE/CheckMLIRWithReference.cs
```diff
@@ -0,0 +1,176 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace SonarAnalyzer.UnitTest.CBDE
+{
+    [TestClass]
+    public class CheckMlirWithReference
```
andrei-epure-sonarsource: should you also add tests for
- Bitwise AND, OR, XOR operations?
- multiplication, division , modulo?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/CBDE/MlirExportTest.cs
```diff
@@ -0,0 +1,1343 @@
+﻿extern alias csharp;
+using System.IO;
+using csharp::SonarAnalyzer.ControlFlowGraph.CSharp;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace SonarAnalyzer.UnitTest.CBDE
+{
+    [TestClass]
+    public class MlirExportTest
```
andrei-epure-sonarsource: if this is the test that just verifies the generated code to respect the MLIR syntax, please add a comment above the class to specify that

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/CBDE/MlirTestUtilities.cs
```diff
@@ -0,0 +1,114 @@
...
+using csharp::SonarAnalyzer.ControlFlowGraph.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarAnalyzer.ControlFlowGraph;
+
+namespace SonarAnalyzer.UnitTest.CBDE
+{
+    public static class MlirTestUtilities
+    {
+        private static string cbdeDialectCheckerPath = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), @"CBDE\windows\cbde-dialect-checker.exe");
```
andrei-epure-sonarsource: _nitpick_ long line

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CbdeHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+            i -=100; // Noncompliant {{This operation always overflows}}
+        }
+        public void DoSomething() { throw new NotSupportedException(); }
+        public void DoSomething2() { throw new NotSupportedException(); }
+        public void DoSomething3() {
+            goto Label;
+        Label:
+            return;
+        }
+        public void DoSomethingArgs(bool a, bool b,int i)
```
andrei-epure-sonarsource: I guess you can remove this test?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CbdeHandlerRule.cs
andrei-epure-sonarsource: as discussed, it can be `internal` and it will be visible to the UT project (see https://github.com/SonarSource/sonar-dotnet/blob/master/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Properties/AssemblyInfo.cs#L28 )

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
andrei-epure-sonarsource: see [PropertiesHelper](https://github.com/SonarSource/sonar-dotnet/blob/master/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/PropertiesHelper.cs#L30) and how it's used

.gitignore
```diff
@@ -62,3 +62,4 @@ packages/
 # rule-api temp artifacts
 *.restext
 RspecStrings.resx
+/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE
```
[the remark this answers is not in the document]
loic-joly-sonarsource: It should not, only the windows subfolder should be ignored... I don't know why this reappeared...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
loic-joly-sonarsource: Code referenced in https://github.com/SonarSource/SonarCBDE/issues/67

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
loic-joly-sonarsource: See #2885

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/CBDE/CheckMLIRWithReference.cs
loic-joly-sonarsource: I don't really think so, because those operations are handled in the same way as operators that are tested here, and in the same functions (`ExtractBinaryExpression` & `ExtractBinaryAssignmentExpression`), so adding tests here would not increase coverage.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
andrei-epure-sonarsource: in the end we agreed to use an env var, as the usage is mainly internal. It will also be easy to use in case of support diagnostics

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/CBDE/CheckMLIRWithReference.cs
andrei-epure-sonarsource: I'm not talking about increasing coverage (_pleasing the tool_), I am talking about testing the behavior of our analyzers.

I believe tests should not be made based on implementation, but based on functionality. If the implementation is refactored (e.g. a `switch` will be replaced with `if-else` branches), tests should stay the same.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
loic-joly-sonarsource: Is switching it to `internal` enough to you?
I'd like to avoid making it accessible to anything but `MLIRExporter`, because what it does looks very specific to our use case to me...

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/CBDE/CheckMLIRWithReference.cs
geoffray-adde-sonarsource: tests were added

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ResourceTests/GeneratedResourcesTest.cs
```diff
@@ -42,9 +42,10 @@ public class GeneratedResourcesTest
         [TestMethod]
         public void AnalyzersHaveCorrespondingResource_CSharp()
         {
-            var rulesFromResources = GetRulesFromResources(RspecRelativeFolderPath + "cs");
+            // NOTE: we allow one rule to be implemented by several classes (because of CBDE that can overlap with a C# class)
```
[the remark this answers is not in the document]
geoffray-adde-sonarsource: reverted

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
loic-joly-sonarsource: It will probably naturally come when handling `A ? B : C`, so no need to keep this TODO for now.

andrei-epure-sonarsource:
@geoffray-adde-sonarsource  for the conflicting `packages.lock.json`, you can talk with @costin-zaharia-sonarsource how to update the file

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
loic-joly-sonarsource: Move it to a non-code comment that indicates how the code works.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
loic-joly-sonarsource: I usually try to use it, but in this case, I believe the concatenation is as clear (as well as being slightly faster). Maybe because I don't see this code as "a string with changing parts", but as "a string surrounded by parentheses"

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -0,0 +1,436 @@
...
+        {
+            string jsonFileContent;
+            List<List<JObject>> jsonIssues;
+            LogIfFailure($"- parsing json file {cbdeJsonOutputPath}");
+            try
+            {
+                jsonFileContent = File.ReadAllText(cbdeJsonOutputPath);
+                jsonIssues = JsonConvert.DeserializeObject<List<List<JObject>>>(jsonFileContent);
+            }
+            catch
```
andrei-epure-sonarsource: what do you want to catch? please add a specific exception and details about the exception cause in the error message

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -0,0 +1,436 @@
...
+            {
+                LogIfFailure($"- error parsing json file {cbdeJsonOutputPath}");
+                Log(logStringBuilder.ToString());
+                Console.Error.WriteLine($"Error parsing output from CBDE, more details in {cbdeProcessSpecificPath}");
+                return;
+            }
+
+            foreach (var issue in jsonIssues.First())
```
andrei-epure-sonarsource: what happens with what's after First?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -0,0 +1,436 @@
...
+            {
+                LogIfFailure($"  * processing token {issue.ToString()}");
+                try
+                {
+                    RaiseIssueFromJToken(issue, context);
+                }
+                catch
```
andrei-epure-sonarsource: please catch a specific Exception and add failure details in the error message

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+using System.Text.RegularExpressions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.ControlFlowGraph;
+using SonarAnalyzer.ControlFlowGraph.CSharp;
+using SonarAnalyzer.Helpers;
+using SonarAnalyzer.ShimLayer.CSharp;
+
+namespace SonarAnalyzer
```
andrei-epure-sonarsource: please add CBDE or CommonBugDetectionEngine in the namespace of the files inside the CBDE folder

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+        }
+
+        public void ExportFunction(MethodDeclarationSyntax method)
+        {
+            if (method.Body == null)
+            {
+                return;
+            }
+
+            if (IsTooClomplexForMLIROrTheCFG(method))
```
andrei-epure-sonarsource: typo - `Clomplex` -> `Complex`?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+            {
+                return OpId(getAssignmentValue(op));
+            }
+            var newId = UniqueOpId();
+            writer.WriteLine($"%{newId} = cbde.unknown : {resultType} {GetLocation(op)} // Dummy variable because type of %{OpId(getAssignmentValue(op))} is incompatible");
+            return newId;
+        }
+
+        private readonly TextWriter writer;
+        private readonly SemanticModel semanticModel;
```
andrei-epure-sonarsource: please add fields at the beginnign of the class - see our C# Coding Style https://github.com/SonarSource/sonar-dotnet/blob/master/docs/coding-style.md#class-members

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+{
+    internal static class SyntaxNodeExtension
+    {
+        public static string Dump(this SyntaxNode node)
+        {
+            return Regex.Replace(node.ToString(), @"\t|\n|\r", " ");
+        }
+    }
+
+    public class MLIRExporter
```
andrei-epure-sonarsource: are you checking the SonarCloud issues? https://sonarcloud.io/project/issues?id=sonaranalyzer-dotnet&open=AW8AX2qMPTv0sE1_NOv1&pullRequest=2477&resolved=false

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+                    sb.Append('.');
+                }
+                else
+                {
+                    sb.Append('$');
+                }
+            }
+            return sb.ToString();
+        }
+        private bool IsTooClomplexForMLIROrTheCFG(MethodDeclarationSyntax method)
```
andrei-epure-sonarsource: please add empty line above

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+            }
+            // MLIR encodes blocks relationships in operations, not in blocks themselves
+            // So we need to add the corresponding operations at the end...
+            switch (block)
+            {
+                case JumpBlock jb:
+                    switch (jb.JumpNode)
+                    {
+                        case ReturnStatementSyntax ret:
+                            if (ret.Expression == null)
```
andrei-epure-sonarsource: please move this logic in a method, it's too big for a switch case

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+                            }
+                            break;
+                        case BreakStatementSyntax breakStmt:
+                            writer.WriteLine($"br ^{BlockId(jb.SuccessorBlock)} {GetLocation(breakStmt)} // break");
+                            break;
+                        case ContinueStatementSyntax continueStmt:
+                            writer.WriteLine($"br ^{BlockId(jb.SuccessorBlock)} {GetLocation(continueStmt)} // continue");
+                            break;
+                        case ThrowStatementSyntax throwStmt:
+                            // TODO : Should we transfert to a catch block if we are inside a try/catch?
```
andrei-epure-sonarsource: please open github issue and put link here if there is an action to be done in the future... otherwise what's the purpose of the TODO? https://sonarcloud.io/project/issues?id=sonaranalyzer-dotnet&open=AW8AX2qMPTv0sE1_NOvx&pullRequest=2477&resolved=false

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+        }
+
+        private void ExtractBinaryExpression(ExpressionSyntax expr, ExpressionSyntax lhs, ExpressionSyntax rhs)
+        {
+            if (!AreTypesSupported(lhs, rhs, expr))
+            {
+                writer.WriteLine($"%{OpId(expr)} = cbde.unknown : {MLIRType(expr)} {GetLocation(expr)} // Binary expression on unsupported types {expr.Dump()}");
+                return;
+            }
+            // TODO : C#8 : Use switch expression
```
andrei-epure-sonarsource: it's possible if you rebase on master

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+            writer.WriteLine($"%{OpId(op)} = cmpi \"{compName}\", %{OpId(getAssignmentValue(binExpr.Left))}, %{OpId(getAssignmentValue(binExpr.Right))} : {MLIRType(binExpr.Left)} {GetLocation(binExpr)}");
+        }
+
+        private string GetLocation(SyntaxNode node)
+        {
+            if (!exportsLocations)
+            {
+                return string.Empty;
+            }
+            // TODO: We should decide which of GetLineSpan or GetMappedLineSpan is better to use
```
andrei-epure-sonarsource: please open a github issue for this

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -0,0 +1,853 @@
...
+        private readonly TextWriter writer;
+        private readonly SemanticModel semanticModel;
+        private readonly bool exportsLocations;
+        private readonly Dictionary<Block, int> blockMap = new Dictionary<Block, int>();
+        private int blockCounter = 0;
+        private readonly Dictionary<SyntaxNode, int> opMap = new Dictionary<SyntaxNode, int>();
+        private int opCounter = 0;
+        private readonly Encoding encoder = System.Text.Encoding.GetEncoding("ASCII", new PreservingEncodingFallback(), DecoderFallback.ExceptionFallback);
+        private MlirExporterMetrics exporterMetrics;
+        public static readonly List<SyntaxKind> unsupportedSyntaxes = new List<SyntaxKind>
```
andrei-epure-sonarsource: same here - please order fields based on their accessibility

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/PreservingEncodingFallback.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SonarAnalyzer
+{
+    internal class PreservingEncodingBuffer : EncoderFallbackBuffer
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: you've solved this, but I don't see any comment on why this is necessary

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/PreservingEncodingFallback.cs
```diff
@@ -0,0 +1,71 @@
...
+        private string buffer;
+        private int currentChar;
+    }
+
+    /// <summary>
+    /// C# source code can contain any character, but MLIR only handle 8-bits chars. We must therefore encode C# names
+    /// so that two different strings in C# always result in two different strings in the generated code (by default, all
+    /// unknown characters would be translated to the same one)
+    /// </summary>
+    internal class PreservingEncodingFallback : EncoderFallback
```
andrei-epure-sonarsource: please move it in a separate file, we don't keep different types in a single file in our codebase and this is confusing

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/PreservingEncodingFallback.cs
```diff
@@ -0,0 +1,71 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Text;
+
+namespace SonarAnalyzer
+{
+    internal class PreservingEncodingBuffer : EncoderFallbackBuffer
```
andrei-epure-sonarsource: also, please move it in a separate file, we don't keep different types in a single file in our codebase and this is confusing

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
```diff
@@ -0,0 +1,80 @@
...
+  <ItemGroup>
+    <Content Include="$(RelativeCBDEFolder)\windows\**\*.exe" >
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="$(RelativeCBDEFolder)\windows\**\*.txt" >
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
+
+
```
andrei-epure-sonarsource: nitpick: are all these empty lines necessary here? can we just keep one?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CbdeHandlerRule.cs
```diff
@@ -0,0 +1,110 @@
...
+        private const string S3949MessageFormat = "{0}";
+        private static readonly DiagnosticDescriptor ruleS3949 = DiagnosticDescriptorBuilder.GetDescriptor(S3949DiagnosticId, S3949MessageFormat, RspecStrings.ResourceManager, fadeOutCode: true);
+        private static string WorkDirectoryBasePath;
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(ruleS3949);
+
+        private readonly ImmutableDictionary<string, DiagnosticDescriptor> ruleIdToDiagDescriptor = ImmutableDictionary<string, DiagnosticDescriptor>.Empty
+            .Add("S3949", ruleS3949);
+        public CbdeHandlerRule() : this(false) {}
+        private CbdeHandlerRule(bool unitTest)
```
andrei-epure-sonarsource: please add empty lines between methods

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CbdeHandlerRule.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            this.unitTest = unitTest;
+        }
+        internal static CbdeHandlerRule MakeUnitTestInstance()
```
andrei-epure-sonarsource: please add empty lines between methods

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CbdeHandlerRule.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            return new CbdeHandlerRule(true);
+        }
+        protected sealed override void Initialize(SonarAnalysisContext context)
```
andrei-epure-sonarsource: please add empty lines between methods

sonaranalyzer-dotnet/src/SonarAnalyzer.RuleDescriptorGenerator/Program.cs
```diff
@@ -1,4 +1,4 @@
-/*
+﻿/*
```
andrei-epure-sonarsource: can this be reverted?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/CBDE/MlirTestUtilities.cs
```diff
@@ -0,0 +1,134 @@
...
+            Assert.AreEqual(trimmedExpected.Trim(), trimmedActual.Trim());
+        }
+
+        public static IControlFlowGraph GetCfgForMethod(string code, string methodName)
+        {
+            (var method, var semanticModel) = TestHelper.Compile(code).GetMethod(methodName);
+
+            return CSharpControlFlowGraph.Create(method.Body, semanticModel);
+        }
+        public static string GetCfgGraph(string code, string methodName)
```
andrei-epure-sonarsource: please add empty lines between methods

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/CBDE/MlirTestUtilities.cs
```diff
@@ -0,0 +1,134 @@
...
+            Path.Combine(
+                Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location),
+                @"CBDE\windows\cbde-dialect-checker.exe");
+        public static void checkExecutableExists()
+        {
+            Assert.IsTrue(File.Exists(cbdeDialectCheckerPath),
+                $"We need cbde-dialect-checker.exe to validate the generated IR, searched in path {cbdeDialectCheckerPath}");
+        }
+
+        private static string ValidateIR(string path)
```
andrei-epure-sonarsource: please order methods based on visibility , see our coding style https://github.com/SonarSource/sonar-dotnet/blob/master/docs/coding-style.md#class-members

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -0,0 +1,436 @@
...
+
+        // cbdePath is inside .sonarqube/out/<n>/
+        // cbdeDirectoryRoot contains mlir files and results for each assembly
+        // cbdeProcessSpecificPath is the $"{cbdePath}/CBDE_{pid}/" folder
+        // cbdeLogFile, cbdeMetricsLogFile and cbdePerfLogFile are inside cbdeProcessSpecificPath
+        private string cbdePath;
+        private string cbdeDirectoryRoot;
+        private string cbdeDirectoryAssembly;
+        private string cbdeJsonOutputPath;
+        private static string cbdeProcessSpecificPath;
```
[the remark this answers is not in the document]
geoffray-adde-sonarsource: I am not sure if this class can be instantiated multiple times by the same process.
I am assuming than the static variables are shared across only on process.
So, everything that is linked to one process could be static but.... a few things (like some path) are process-specific but also instance specific (like the some_rpoject/out/<n>/CBDE_<pid>/.
All that indeed creates a sort of strange mix-up that I did not make very clear in the code.
I just changed the one field you were mentioning as non static for the sake of clarity.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -0,0 +1,436 @@
...
+                pProcess.StartInfo.RedirectStandardError = true;
+                pProcess.StartInfo.UseShellExecute = false;
+                pProcess.Start();
+                long totalProcessorTime=0, peakPagedMemory = 0, peakWorkingSet=0;
+                while(!pProcess.WaitForExit(processStatPeriodMs))
+                {
+                    try
+                    {
+                        pProcess.Refresh();
+                        totalProcessorTime = (long)pProcess.TotalProcessorTime.TotalMilliseconds;
```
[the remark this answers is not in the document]
geoffray-adde-sonarsource: Because it comes from a double

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
geoffray-adde-sonarsource: I added a comment to explain this.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
geoffray-adde-sonarsource: Nope. I will start from now on.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/PreservingEncodingFallback.cs
loic-joly-sonarsource: The comment is on line 57, because both class work together for a single purpose...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/PreservingEncodingFallback.cs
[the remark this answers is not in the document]
loic-joly-sonarsource: The comment is on line 57, because both class work together for a single purpose...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
loic-joly-sonarsource: Unfortunately, new cases were added that no longer fall into the switch expression... So removing this outdated TODO

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
andrei-epure-sonarsource: or, as discussed with @loic-joly-sonarsource offline, add the EPIC/MMF jira id in the comment, to be able to see if it's still in progress or already done

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -366,27 +371,37 @@ private void RaiseIssuesFromJSon(CompilationAnalysisContext context)
...
             {
-                LogIfFailure($"- error parsing json file {cbdeJsonOutputPath}");
-                Log(logStringBuilder.ToString());
-                Console.Error.WriteLine($"Error parsing output from CBDE, more details in {cbdeProcessSpecificPath}");
-                return;
+                if (exception is JsonException || exception is IOException)
+                {
+                    LogIfFailure($"- error parsing json file {cbdeJsonOutputPath}: {exception.ToString()}");
+                    Log(logStringBuilder.ToString());
+                    Console.Error.WriteLine($"Error parsing output from CBDE, more details in {cbdeProcessSpecificPath}");
```
andrei-epure-sonarsource: don't forget to replace this with throwing an exception, if you want something to show in the users console (AFAIK, Console output will be eaten by roslyn)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -366,27 +371,37 @@ private void RaiseIssuesFromJSon(CompilationAnalysisContext context)
...
                 {
-                    LogIfFailure($"  * error reporting token {cbdeJsonOutputPath}");
-                    Log(logStringBuilder.ToString());
-                    Console.Error.WriteLine($"Error raising issue from CBDE, more details in {cbdeProcessSpecificPath}");
-                    continue;
+                    if (e is SystemException || e is JsonException)
+                    {
+                        LogIfFailure($"  * error reporting token {cbdeJsonOutputPath}: {e.ToString()}");
+                        Log(logStringBuilder.ToString());
+                        Console.Error.WriteLine($"Error raising issue from CBDE, more details in {cbdeProcessSpecificPath}");
```
andrei-epure-sonarsource: this may not show in the users' console (Roslyn eats standard output)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -159,6 +206,75 @@ private bool HasNoReturn(MethodDeclarationSyntax method)
...
+                    ExportReturnStatement(ret, functionReturnType);
+                    break;
+                case BreakStatementSyntax breakStmt:
+                    writer.WriteLine($"br ^{BlockId(jb.SuccessorBlock)} {GetLocation(breakStmt)} // break");
+                    break;
+                case ContinueStatementSyntax continueStmt:
+                    writer.WriteLine($"br ^{BlockId(jb.SuccessorBlock)} {GetLocation(continueStmt)} // continue");
+                    break;
+                case ThrowStatementSyntax throwStmt:
+                    // TODO : Should we transfert to a catch block if we are inside a try/catch?
```
andrei-epure-sonarsource: typo in transfert
is there an existing ticket for this (looks like a design question?)

sonaranalyzer-dotnet/src/SonarAnalyzer.RuleDescriptorGenerator/Program.cs
```diff
@@ -1,4 +1,4 @@
-﻿/*
+/*
```
andrei-epure-sonarsource: 😄

geoffray-adde-sonarsource: Sorry for that. If we squash and merge. It would not appear anymore. Would it?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
loic-joly-sonarsource: I already added the reference to https://github.com/SonarSource/SonarCBDE/issues/111. Would you like something else?
And transfert is not a typo, it's just a french word we used because... well because... well, ok, it's a typo :)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
loic-joly-sonarsource: Maybe it shows in unit tests, which would be better than nothing?
Otherwise, do you know a way to log info to user?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
geoffray-adde-sonarsource: removed and replaced by throwing an exception

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
geoffray-adde-sonarsource: removed and replaced by throwing an exception

andrei-epure-sonarsource:
Sorry, I've been reviewing individual commits and hence missing it.

No, the link to the issue is perfect.

Indeed, I need to become more French :)

andrei-epure-sonarsource:
The only ways I know are
- raising an issue
- throwing an exception

If it doesn't work for users (and we know it), than you can use
System.Diagnostics.Debug for logging for UTs

Otherwise, people who read this might except to be logged for users

sonaranalyzer-dotnet/src/SonarAnalyzer.RuleDescriptorGenerator/Program.cs
andrei-epure-sonarsource: dunno

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -358,7 +366,7 @@ private void LogFailedCbdeRun(Process pProcess)
             failureString.Append("  content of stderr is:\n" + pProcess.StandardError.ReadToEnd());
             failureString.Append("  content of the CBDE handler log file is :\n" + logStringBuilder.ToString());
             Log(failureString.ToString());
-            Console.Error.WriteLine($"Error when executing CBDE, more details in {cbdeProcessSpecificPath}");
+            throw new Exception($"CBDE external process reported an error.");
```
andrei-epure-sonarsource: why remove the path? it may be useful...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
andrei-epure-sonarsource: it seems github is not smart enough to add the comment here when it's sent via email

the link is good

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
andrei-epure-sonarsource: or when throwing, the logs won't be dumped to the file anymore?

geoffray-adde-sonarsource: because the path will be there only if the logging is enabled by an internal env variable that we would use as a support tool, right?
I can add it only if the env var is defined.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeException.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
```
andrei-epure-sonarsource: header?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
geoffray-adde-sonarsource: done

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MlirExporter.cs
```diff
@@ -0,0 +1,872 @@
...
+using SonarAnalyzer.Helpers;
+using SonarAnalyzer.ShimLayer.CSharp;
+
+namespace SonarAnalyzer.CBDE
+{
+    public class MlirExporter
+    {
+        public static readonly ImmutableList<SyntaxKind> unsupportedSyntaxes = new List<SyntaxKind>
+        {
+            SyntaxKind.ForEachStatement,
```
andrei-epure-sonarsource: _nitpick_ it's better to order alphabetically, when navigating the code

andrei-epure-sonarsource: and please reduce the visibility to private or internal - if it's public, it should be a { get; } property instead

PR #2485, opened 2019-07-15
Adding a note related to encoding

abbas-sabra-sonarsource:


docs/contributing-analyzer.md
```diff
@@ -46,6 +46,9 @@ If the script ends with `SUCCESS: No differences were found!` (or exit code 0),
 
 If the script ends with `ERROR: There are differences between the actual and the expected issues.` (or exit code 1),
 the changes you have made have impacted one or many issues raised by the rules.
+
+Note: if you are facing compilation errors due to unknown characters, Disable `beta use unicode utf-8 for worldwide language support` in your `Region Settings`.
```
andrei-epure-sonarsource: I think you should be more explicit:

- mention that this happens on Windows 10
- mention that `Region Settings` is inside Control Panel (ideally say how to reach that UI)

PR #2487, opened 2019-07-16
S2583: Fix reported FP

### LINKED ISSUE #2369: Fix S2583: FP with null-coalescing operator
Copied from https://github.com/SonarSource/sonarlint-visualstudio/issues/866 created by @peska

### Description

Warning S2583 Change this condition so that it does not always evaluate to 'false'; some subsequent code is never executed.

### Repro steps

Sample class to reproduce. Warning in line 5:

```
public class TestClass
{
	public void TestMethod(bool? input)
	{
		if (input ?? false)
			Console.WriteLine("input is true");
		else
			Console.WriteLine("input is false");
	}
}
```

### Expected behavior

No warning given. This is a false positive.

### Actual behavior

Warning S2583 Change this condition so that it does not always evaluate to 'false'; some subsequent code is never executed.

### Known workarounds

Use:

`if (input.GetValueOrDefault())`

### Related information

* SonarLint for VS Version: 4.9.0.4314 https://github.com/SonarSource/sonarlint-visualstudio/releases/tag/v4.9-rc2
* Visual Studio Version Enterprise 2019 16.0.0
* SonarQube server version: ---
* Languages in your solution:

abbas-sabra-sonarsource:
Fixes #2369

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -1256,11 +1256,17 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
                             BuildCondition(binaryExpression.Right, trueSuccessor, falseSuccessor));
 
                     case SyntaxKind.LogicalAndExpression:
+                        return BuildCondition(
+                        binaryExpression.Left,
+                        BuildCondition(binaryExpression.Right, trueSuccessor, falseSuccessor),
+                        falseSuccessor);
+
                     case SyntaxKind.CoalesceExpression:
                         return BuildCondition(
-                            binaryExpression.Left,
-                            BuildCondition(binaryExpression.Right, trueSuccessor, falseSuccessor),
-                            falseSuccessor);
+                        binaryExpression.Left,
+                        BuildCondition(binaryExpression.Right, trueSuccessor, falseSuccessor),
+                        CreateBranchBlock(binaryExpression.Left,successors: new[] { trueSuccessor, falseSuccessor }));
+
                 }
             }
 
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.cs
```diff
@@ -1556,4 +1556,33 @@ void Second(FooContainer fooContainer)
...
+
+  public class TestNullCoalescing
+  {
+    public void CompliantMethod(bool? input)
+    {
+      if (input ?? false) // Compliant S2583
+      {
+        Console.WriteLine("input is true");
+      }
+
```
andrei-epure-sonarsource: I guess you can remove this empty line

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.cs
```diff
@@ -1556,4 +1556,33 @@ void Second(FooContainer fooContainer)
             }
         }
     }
+
+  public class TestNullCoalescing
```
andrei-epure-sonarsource: could you add some tests with the null coalescing operator
- inside a `while`
- inside a ternary condition (maybe like in line 1540? if it makes sense...)
- inside boolean expression (see examples Foo1-Foo7 lines 85-140)

Also, please add some tests inside `ConditionEvaluatesToConstant.CSharp7.cs`

Also, please have a look at the test at line 1332 to see if it does what it should (it's also using the null coalescing operator `if (b ?? false)`)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -1256,10 +1256,15 @@ private Block BuildCondition(ExpressionSyntax expression, Block trueSuccessor, B
                             BuildCondition(binaryExpression.Right, trueSuccessor, falseSuccessor));
 
                     case SyntaxKind.LogicalAndExpression:
+                        return BuildCondition(
```
andrei-epure-sonarsource: please verify existing tests and most likely add some new tests for both LogicalAnd and NullCoalescing expressions in `ControlFlowGraphTest.cs`

Issue #2311, opened 2019-03-01
Fix S1854 FP: `case when` syntax not supported when pattern is a constant


ValentijnMakkenze:
### Description
We get a false positive for Dead stores should be removed

### Repro steps
```
  private void TextEditorKeydown(object sender, KeyEventArgs e)
  {
    var tab = new string(' ', 2);  // Remove this useless assignment to local variable 'tab'

    switch (e.KeyCode)
    {
    case Keys.Tab:
    {
      e.SuppressKeyPress = true;
      var ultraTextEditor = (UltraTextEditor)sender;
      var caretPos = ultraTextEditor.Editor.SelectionStart;
      ultraTextEditor.Text = ultraTextEditor.Text.Insert(caretPos, tab);  // We use tab here
      ultraTextEditor.Editor.SelectionStart = caretPos + tab.Length;
      ultraTextEditor.ScrollToCaret();
      break;
    }
  …
```

### Expected behavior
No warning

### Actual behavior
Warning: Remove this useless assignment to local variable 'tab'.

### Known workarounds
Mark item as false positive

### Related information
* SonarC# Version 7.11.0.8083
* Visual Studio Version 2017
* If running through the Scanner for MSBuild, its version 4.5.0.1761

andrei-epure-sonarsource:
hi @ValentijnMakkenze . I cannot repro this FP. I don't have the library for `UltraTextEditor`, so I used the following snippet (with SonarC# 7.14). 

```
    public class WindowsFormsExample
    {
        private string TextEditorKeydown(object sender, KeyEventArgs e)
        {
            var tab = new string(' ', 2);  // Remove this useless assignment to local variable 'tab'

            switch (e.KeyCode)
            {
                case Keys.Tab:
                {
                    e.SuppressKeyPress = true;
                    var text = new string(' ', 1);
                    text = text.Insert(1, tab);  // We use tab here
                    return text;
                }
            }
            return "";
        }
    }
```

Could you please provider a smaller reproducer w/o the need of that `UltraTextEditor` library?

ValentijnMakkenze:
Hi Andrei,

My example was reduced too much, it’s the ’case when’ that came after it that triggers the warning without it the warning disappears.

    private void TextEditorKeydown(object sender, KeyEventArgs e)
    {
      var tab = new string(' ', 2);
      switch (e.KeyCode)
      {
        case Keys.Tab:
        {
          Console.WriteLine(tab);
          break;
        }

[...]

PR #2493, opened 2019-07-18
Fix Case-when false positives for S1172 and S1854 

### LINKED ISSUE #2010: S1172 False positive when pattern matching with parameter (C# 7.7.0.7192)
### Description

When a function parameter only is used in het when clause of pattern matching in a switch statement it incorrecty marks that parameter as unused (S1172)

### Repro steps

Example code:
```csharp
static int PatternMatch(StringSplitOptions splitOptions, int i)
{
    switch (splitOptions)
    {
        case StringSplitOptions.None
            when i > 0:
            return 1;
        default:
            return 0;
    }
}
```

Parameter _i_ is marked unused.

### Expected behavior

No warning from sonar analyzers.

### Actual behavior

S1172 warning generated:
`
Warning	S1172	Remove this parameter 'i', whose value is ignored in the method.
`

### Known workarounds

Suppress warning.

### Related information

* SonarC# Version: 7.7.0.7192
* Visual Studio Version: Visual Studio Professional 2017 (15.8.7 and 15.9.0 Preview 3)

andrei-epure-sonarsource:
Fixes #2010 and #2311


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -2807,7 +2807,7 @@ public void Cfg_Switch_Patterns_Default()
 
         [TestMethod]
         [TestCategory("CFG")]
-        public void Cfg_Switch_Patterns_When()
+        public void Cfg_Switch_Patterns_Two_Case_When()
```
andrei-epure-sonarsource: @duncanp-sonar - do you reckon there is some `switch` scenario (pre-C#8) which is not covered by the UTs?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
```diff
@@ -617,6 +617,11 @@ private void VisitBinaryBranch(BinaryBranchBlock binaryBranchBlock, ExplodedGrap
             var ps = node.ProgramState;
             SymbolicValue sv;
 
+            if (!ps.HasValue)
```
abbas-sabra-sonarsource: I think you can move this to be before the declaration of `sv`.

duncanp-sonar:
Looks good. A couple of other test cases to consider:

1) default statement first
```
int index = 0;
Exception ex = null;
switch (index)
{
    default:
        break;
    case 0 when ex is InvalidOperationException:
        ex = null;
        break;
}```

2) multiple cases with the same action

```

int index = 0;
Exception ex = null;
switch (index)
{
    case 0 when ex is ArgumentException:
    case 1:
    case 2 when ex is InvalidOperationException:
        // do stuff
        break;
}
```

PR #2502, opened 2019-07-24
fix-fp-S2583

### LINKED ISSUE #1347: Rule S2583: False positive when using ?? with an empty string
### Description

Sonarqube raises a false positive when using an ?? operation in combination with `string.IsNullOrEmpty`.

### Repro steps

The following code raises a code-smell with the `string.IsNullOrEmpty` check, because it assumes the expression always true. When `parameter` is in fact an empty string (not `null`, but `string.Empty`), then the expression will be false.

    public static string ShowCSharpSquidS2583(string parameter)
    {
        var useParameter = parameter ?? "non-empty";

        // The following line shows an error that the condition is always true,
        // but if 'parameter' has value 'string.Empty' then it will be false.
        if (!string.IsNullOrEmpty(useParameter))
            return "non-empty";

        return "empty";
    }

The same code-smell is given for `string.IsNullOrWhiteSpace` (which is a bit harder to prevent).

### Expected behavior

No code-smell should be given.

### Actual behavior

Code smell csharpsquid:S2583 is raised.

### Known workarounds

In this case we could check if the string is not empty `!useParameter.Equals(string.Empty)`.

### Related information

* SonarC# Version 6.7.1 (build 35068)
* Visual Studio Version: Visual Studio 2017 (15.6.6)

abbas-sabra-sonarsource:
Fixes #1347

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -1056,7 +1056,10 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
         private Block BuildCoalesceExpression(BinaryExpressionSyntax expression, Block currentBlock)
         {
             var rightBlock = BuildExpression(expression.Right, CreateBlock(currentBlock));
-            return BuildExpression(expression.Left, CreateBinaryBranchBlock(expression, rightBlock, currentBlock));
+            var leftCurrentBlock = expression.Parent.Kind() == SyntaxKind.EqualsValueClause ? currentBlock : CreateBlock(currentBlock);
+            var leftBlock = BuildExpression(expression.Left, leftCurrentBlock);
+            return BuildExpression(expression.Left,
+                CreateBinaryBranchBlock(expression, rightBlock, leftBlock));
         }
 
         private Block BuildLogicalAndExpression(BinaryExpressionSyntax expression, Block currentBlock)
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
@@ -549,8 +549,6 @@ namespace SonarAnalyzer.SymbolicExecution
             if (ShouldConsumeValue(parenthesizedExpression))
             {
                 var newProgramState = programState.PopValue();
-                System.Diagnostics.Debug.Assert(!newProgramState.HasValue);
-
                 return newProgramState;
             }
 
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
@@ -1361,7 +1361,7 @@ namespace NS
 
             VerifyAllInstructions(branchBlock, "b");
             VerifyAllInstructions(bNullBlock, "c");
-            VerifyAllInstructions(afterOp, "a = b ?? c");
+            VerifyAllInstructions(afterOp, "b", "a = b ?? c");
         }
 
         [TestMethod]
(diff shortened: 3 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -1056,7 +1056,10 @@ private Block BuildConditionalExpression(ConditionalExpressionSyntax conditional
         private Block BuildCoalesceExpression(BinaryExpressionSyntax expression, Block currentBlock)
         {
             var rightBlock = BuildExpression(expression.Right, CreateBlock(currentBlock));
-            return BuildExpression(expression.Left, CreateBinaryBranchBlock(expression, rightBlock, currentBlock));
+            var leftCurrentBlock = expression.Parent.Kind() == SyntaxKind.EqualsValueClause ? currentBlock : CreateBlock(currentBlock);
```
andrei-epure-sonarsource: could you please add a comment why you do this here and not in the SE engine (what we discussed y-day)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
```diff
@@ -549,8 +549,6 @@ private ProgramState EnsureStackState(ExpressionSyntax parenthesizedExpression,
             if (ShouldConsumeValue(parenthesizedExpression))
             {
                 var newProgramState = programState.PopValue();
-                System.Diagnostics.Debug.Assert(!newProgramState.HasValue);
```
andrei-epure-sonarsource: why did you remove the assert? it doesn't affect prod, it will only run if compiling with `/d:DEBUG`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -1361,30 +1361,35 @@ public void Cfg_Coalesce()
 
             VerifyAllInstructions(branchBlock, "b");
```
andrei-epure-sonarsource: now that you touch this code, could you please replace `branchBlock.SuccessorBlocks.Should().OnlyContainInOrder(bNullBlock, afterOp)` with `TrueSuccessor.Should()...` and `FalseSuccessor.Should()...`)?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -1361,30 +1361,35 @@ public void Cfg_Coalesce()
...
             VerifyAllInstructions(bNullBlock, "c");
-            VerifyAllInstructions(afterOp, "a = b ?? c");
+            VerifyAllInstructions(afterOp, "b", "a = b ?? c");
```
andrei-epure-sonarsource: this actually makes sense, because `b` is the value that gets assigned to `a` in this block 👍

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -1361,30 +1361,35 @@ public void Cfg_Coalesce()
...
-            var dBlock = blocks[2];
-            var bcdBlock = blocks[3];   // b ?? c ?? d
+            var cResultBlock = blocks[2];
+            var dBlock = blocks[3];
+            var bcdBlock = blocks[4];   // b ?? c ?? d
             var exitBlock = cfg.ExitBlock;
 
             bBlock.SuccessorBlocks.Should().OnlyContainInOrder(cBlock, bcdBlock);
-            cBlock.SuccessorBlocks.Should().OnlyContainInOrder(dBlock, bcdBlock);
+            cBlock.SuccessorBlocks.Should().OnlyContainInOrder(dBlock, cResultBlock);
```
andrei-epure-sonarsource: same here, if you could check the `TrueSuccessor` and `FalseSuccessor` instead of `SuccessorBlocks`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -1361,30 +1361,35 @@ public void Cfg_Coalesce()
...
             dBlock.SuccessorBlocks.Should().OnlyContain(bcdBlock);
             bcdBlock.SuccessorBlocks.Should().OnlyContain(exitBlock);
 
-            bcdBlock.Instructions.Should().ContainSingle();
-            bcdBlock.Instructions.Should().Contain(i => i.ToString() == "a = b ?? c ?? d");
+            VerifyAllInstructions(bBlock, "b");
+            VerifyAllInstructions(cBlock, "c");
+            VerifyAllInstructions(cResultBlock, "c");
+            VerifyAllInstructions(dBlock, "d");
+            VerifyAllInstructions(bcdBlock, "b", "a = b ?? c ?? d");
```
andrei-epure-sonarsource: I am not sure I am following here... is it possible to have `"c", "b", "a = b ?? c ?? d"`
OR to have inside `cResultBlock` something like `"c", "a = b ?? c ?? d"` ?

if not, that's fine

PR #2504, opened 2019-07-24
Validate generated IR

loic-joly-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/MlirExportTest.cs
```diff
@@ -34,50 +64,38 @@ int Cond2(int i)
...
-                ExportFunction(code, writer, "Cond");
-                ExportFunction(code, writer, "Cond2");
-            }
-
-            var dot = CfgSerializer.Serialize("Cond2", GetCfgForMethod(code, "Cond2"));
+            ValidateCodeGeneration(code);
         }
 
-        private static void ExportFunction(string code, TextWriter writer, string functionName)
+        private static void ExportMethod(string code, TextWriter writer, string functionName)
```
amelie-renard-sonarsource: Unused method?

loic-joly-sonarsource: Yes, no longer used... But I'd like to keep it a little bit before removing it, if you don't mind... I think it will become useful again if we do more accurate tests...

amelie-renard-sonarsource: I don't mind :)

PR #2505, opened 2019-07-25
Handle exit block

loic-joly-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/MLIRExporter.cs
```diff
@@ -100,10 +106,14 @@ private void ExportBlock(Block block, bool isEntryBlock, MethodDeclarationSyntax
                     }*/
                     break;
                 case SimpleBlock sb:
-                    writer.WriteLine("return");
+                    writer.WriteLine($"br ^{BlockId(sb.SuccessorBlock)}");
                     break;
                 case ExitBlock eb:
-                    writer.WriteLine("return");
+                    if (HasNoReturn(parentMethod))
```
amelie-renard-sonarsource: If my understanding is correct, considering previous line:
`if (block is ExitBlock && !HasNoReturn(parentMethod))` -> return
`HasNoReturn` should always return true here

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/MlirExportTest.cs
```diff
@@ -94,8 +94,39 @@ private void ValidateCodeGeneration(string code)
...
+
+int WithReturn(int i) {
+    if (i == 0) {
+        if (i != 0) {
+            int j = 2 * i;
+        }
+    }
+    return i;
+}";
+            var dot = CfgSerializer.Serialize("WithReturn", GetCfgForMethod(code, "UselessCondition"));
```
amelie-renard-sonarsource: I would need some explanation about this line...

Issue #2393, opened 2019-04-18
S1854 FP: `try-catch` inside a loop


remcoros:
### Description

I'm getting a false positive on S1854

### Repro steps

```csharp
    public class Foo
    {
        public void Bar()
        {
            bool isFirst = true;
            foreach (var i in Enumerable.Range(1, 10))
            {
                try
                {
                    if (isFirst)
                    {
                        Console.WriteLine("First");
                    }
                    else
                    {
                        Console.WriteLine("Not first");
                    }
                }
                finally
                {
                    isFirst = false;
                }
            }
        }
    }
```

### Expected behavior

No warning

### Actual behavior

![image](https://user-images.githubusercontent.com/130373/56343966-8be07600-61bc-11e9-9f92-ebe477a37ad4.png)

### Known workarounds

n/a

### Related information

* SonarLint for VS Version: 4.9.0.4316
* Visual Studio 2019 / C#

Corniel:
To me this is valid warning. The finally will only be reached to clean up some stuff. Afterwards, the Exception is still thrown.

remcoros:
It's a simple example (I didn't add a 'catch' in this), but it makes no difference if I add it.

If I remove this "useless assignment", the loop/isFirst check doesn't work anymore, so it's not a valid suggestion.

Corniel:
Ah, now I see. it is within a `foreach` loop. Me bad. So yes, it is a false positive. However, in real examples, I would argue it would be better to put the complex try-catch code within the loop in an separate function to improve readability.

remcoros:
Yes, you're absolutely right. I just wanted to capture the issue here in a minimal example.

andrei-epure-sonarsource:
Adding an example with try-catch FPs inside a loop from @killergege as mentioned in #697

```
public static void RetryOnException(int retries, TimeSpan delay, Action operation)
{
    var attempts = 0;
    do
    {
        try
        {
            attempts++; //Remove this useless assignment to local variable 'attempts'.
            operation();
            break;
        }
        catch (Exception ex)
        {
            if (attempts > retries)
                throw;
            Logger.LogError($"Exception caught on attempt {attempts} - will retry after delay {delay}", ex);
            Task.Delay(delay).Wait();
        }
    } while (true);
}
```

PR #2506, opened 2019-07-25
Fix S1854 FP due to wrong CFG shape when `throw` is inside `try`

### LINKED ISSUE #697: S1854 FP: `try` block not connected to `catch` when exception is thrown inside
### Description

Variables that are only used in exception filters incorrectly trigger S1854

### Repro steps

```cs
bool shouldCatch = false;
try
{
    shouldCatch = true; //S1854 Remove this useless assignment to local variable 'shouldCatch'
    throw new InvalidOperationException("bar");
}
catch (Exception) when(shouldCatch)
{
    Console.WriteLine("Error");
}
```
### Expected behavior

S1854 should not trigger because the `shouldCatch` assignment is not useless. Indeed, without it the exception would pass through.

### Actual behavior

S1854 is triggered.

### Related information

* SonarC# Version 6.3.0.2862
* Visual Studio Version 2017 (15.2)

andrei-epure-sonarsource:
Fixes #697

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
@@ -267,7 +267,7 @@ namespace SonarAnalyzer.Helpers
                 return true;
             }
 
-            var exceptionTypeName = catchClause.Declaration.Type.GetText().ToString();
+            var exceptionTypeName = catchClause.Declaration.Type.GetText().ToString().Trim();
 
             return catchClause.Filter == null &&
                 (exceptionTypeName == "Exception" || exceptionTypeName == "System.Exception");
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
@@ -1056,6 +1056,30 @@ namespace NS
 
         [TestMethod]
         [TestCategory("CFG")]
+        public void Cfg_Return_JustBeforeExit()
+        {
+            var cfg = Build(@"
+            return;
+            cw0();
+            return;");
+
+            VerifyCfg(cfg, 3);
+
+            var blocks = cfg.Blocks.ToList();
+
+            var block1 = (JumpBlock) blocks[0];
+            var block2 = (SimpleBlock) blocks[1];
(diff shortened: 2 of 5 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
```diff
@@ -467,7 +467,67 @@ private void SwitchCase(object sender, KeyEventArgs e)
...
+                throw new InvalidOperationException("bar");
+            }
+            catch (Exception) when (shouldCatch)
+            {
+                Console.WriteLine("Error");
+            }
+        }
+    }
 
+    public class ReproGithubIssue2393
```
andrei-epure-sonarsource: You can ignore these, as they're for a different ticket

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -2592,6 +2592,132 @@ public void Cfg_TryCatch_Exception_Filter()
...
+            afterTryBlock.Should().BeOfType<SimpleBlock>();
+            VerifyAllInstructions(afterTryBlock, "cw5", "cw5()");
+            afterTryBlock.SuccessorBlocks.Should().OnlyContain(exit);
+
+            exit.Should().BeOfType<ExitBlock>();
+        }
+
+        [TestMethod]
+        [TestCategory("CFG")]
+        public void Cfg_TryCatch_Inside_DoWhile()
```
andrei-epure-sonarsource: this is for #2393 as well

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -267,7 +267,7 @@ public static bool IsCatchingAllExceptions(this CatchClauseSyntax catchClause)
                 return true;
             }
 
-            var exceptionTypeName = catchClause.Declaration.Type.GetText().ToString();
+            var exceptionTypeName = catchClause.Declaration.Type.GetText().ToString().Trim();
```
andrei-epure-sonarsource: because of this, "Exception " (which a space in it) was compared with "Exception" below and the code thought that some `catch` blocks weren't catching everything, so the ExitBlock was connected wrongly. This fixes this problem

andrei-epure-sonarsource: and now that I look at it, type comparison should not be done using `ToString`...

PR #2511, opened 2019-07-29
Fix S1854 FP: `try-catch` inside of loop and with return

### LINKED ISSUE #2393: S1854 FP: `try-catch` inside a loop

andrei-epure-sonarsource:
Fixes #2393


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -558,25 +558,49 @@ private Block BuildTryStatement(TryStatementSyntax tryStatement, Block currentBl
                 tryEndStatementConnections.Add(this.ExitTarget.Peek());
             }
 
-            var tryBody = BuildBlock(tryStatement.Block,
-                CreateBranchBlock(tryStatement, tryEndStatementConnections.Distinct()));
-
-            var tryStartStatementConnections = catchBlocks.ToList();
-            tryStartStatementConnections.Add(tryBody); // try body
-
-            if (!areAllExceptionsCaught) // unexpected exception thrown, go to exit (through finally if present)
```
andrei-epure-sonarsource: I would need to find a test case where this is necessary to bring it back

andrei-epure-sonarsource: the `areAllExceptionsCaught` variable is still used at line 556 where it is used for the `tryEndStatementConnections`.

The main change of this PR is that I removed from the `beforeTry` block (initially called `tryStartBlock`) the edges to the `catch` and `finally` blocks - what happens outside of a `try` block (and `tryStartBlock` is outside the try block) won't be caught by `catch` and won't enter `finally` !

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -558,25 +558,49 @@ private Block BuildTryStatement(TryStatementSyntax tryStatement, Block currentBl
...
 
-            var tryBody = BuildBlock(tryStatement.Block,
-                CreateBranchBlock(tryStatement, tryEndStatementConnections.Distinct()));
-
-            var tryStartStatementConnections = catchBlocks.ToList();
-            tryStartStatementConnections.Add(tryBody); // try body
-
-            if (!areAllExceptionsCaught) // unexpected exception thrown, go to exit (through finally if present)
+            Block tryBody;
+            if (tryStatement.Block.Statements.Any(s=>s.IsKind(SyntaxKind.ReturnStatement)))
```
andrei-epure-sonarsource: before this change, when there was a `return` statement inside the body of the `try`, it would short-circuit the `catch` blocks, having only one possible successor: the `exit` block. This was wrong, because if an exception is thrown before the `return`, it can still reach `catch` blocks

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -558,25 +558,49 @@ private Block BuildTryStatement(TryStatementSyntax tryStatement, Block currentBl
...
+                tryBody = CreateBranchBlock(tryStatement, connections);
+            }
+            else
             {
-                tryStartStatementConnections.Add(this.ExitTarget.Peek());
+                tryBody = BuildBlock(tryStatement.Block, CreateBranchBlock(tryStatement,
+                    tryEndStatementConnections.Distinct()));
             }
 
-            var tryStartBlock = CreateBranchBlock(tryStatement, tryStartStatementConnections.Distinct());
```
andrei-epure-sonarsource: `tryStartBlock` has been renamed to `beforeTryBlock`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -836,6 +860,15 @@ private Block BuildBreakStatement(BreakStatementSyntax breakStatement, Block cur
             }
 
             var target = this.BreakTarget.Peek();
+            if (currentBlock is BranchBlock possibleTryBlock &&
```
andrei-epure-sonarsource: until this, when a `break` happened inside a `try`, the block was short-circuited with an only successor: the **ExitBlock**

this change ensures that, in case of a `break` inside a `try`, the edges to `catch` blocks are kept

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
```diff
@@ -528,27 +528,54 @@ public static void RetryOnException(int retries)
...
+        }
+
         public void Bar()
         {
             bool isFirst = true;
             foreach (var i in System.Linq.Enumerable.Range(1, 10))
             {
                 try
                 {
-                    if (isFirst)
```
andrei-epure-sonarsource: this is just a simplification of the test reported in #2393 (it did not affect the reproducibility of the FP)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
```diff
@@ -416,6 +416,65 @@ public virtual int Create()
...
+    {
+        internal void OnlyFinally(object actor)
+        {
+            try
+            {
+                Foo(actor);
+            }
+            finally
+            {
+                actor = null; // FN
```
andrei-epure-sonarsource: I iterated over some solutions to fix this FN, however I gave up because it was adding needless complexity and it wasn't covering all cases

the root cause of this situation is the fact that:
- for each `finally` clause, to separate blocks are generated (see comments in `DeadStore` class)
- when analyzing a block that follow a `try` statement, we cannot tell whether it's a `catch`, a `finally` or just the code which follows the `try-catch-finally` statement

sonaranalyzer-dotnet/its/expected/Ember-MM/Ember.Plugins-{9496C697-5AFD-4813-AEDC-AF33FACEADF0}-S2259.json
```diff
@@ -1,17 +0,0 @@
-{
-"issues":  [
-{
-"id":  "S2259",
-"message":  "'className' is null on at least one execution path.",
-"location":  {
-"uri":  "sources\Ember-MM\Ember.Plugins\PluginSectionHandler.cs",
```
andrei-epure-sonarsource: [Here](https://github.com/SonarSource/sonar-dotnet/blob/master/sonaranalyzer-dotnet/its/sources/Ember-MM/Ember.Plugins/PluginSectionHandler.cs#L91) we used to raise a null pointer exception. HOWEVER, it was an FP because

- `InvalidConfigException` which can be thrown when `className` is still null is caught by the previous `catch`.
- when `UnknownClassException` is thrown, we are sure that `className` is not null because of line 66

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
duncanp-sonar: Will `tryStatement.Block.Statements.Any(...)` find all `return` statements inside the `try-catch`, or only those that are direct children?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -2780,59 +2954,106 @@ public void Cfg_TryCatchFinally_Return_Nested()
...
+            var cfg = Build(@"
+                var number = 5;
+                try
+                {
+                    bar();
+                    return 0;
+                }
+                catch
+                {
+                    return number;
```
duncanp-sonar: Additional tests:
* non-direct child `return` e.g.
```
catch
{
  if (retries > 5) { return; }
}
```

* `return` when there is a `finally` block

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
duncanp-sonar: It might be worth capturing why this FN occurs in the test.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -2669,7 +2715,68 @@ public void Cfg_TryCatch_ThrowInsideTry()
...
+            var cfg = Build(@"
+            var attempts = 0;
+            do
+            {
+                cw0();
+                try
+                {
+                    attempts++;
+                    cw1();
+                    break;
```
duncanp-sonar: Is there a test where there is a `break` inside a `try-catch-finally` (or `try-finally`, and `try-catch-catch-finally`)?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
andrei-epure-sonarsource: fair point, I'll add these comments in the test

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
andrei-epure-sonarsource: Good point, I'll add a test for this use case

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
abbas-sabra-sonarsource: As discussed this FN can be fixed by adding finally clause in the instructions of the finally block to distinguish catch from finally.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
andrei-epure-sonarsource: in addition, I'll have to add tests with a `continue` inside the `try-finally`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
andrei-epure-sonarsource: `return` cannot be inside `finally` - https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0157

PR #2515, opened 2019-07-29
Fix switch case cfg

### LINKED ISSUE #2240: Fix S2259: False positive with switch and pattern matching case blocks
### Description

The following FP is raised because we do not put `Not Null` constraint on `o` when performing symbolic execution. We put `Not Null` only on the matched variables.

```
switch (o)
{
    case Foo f:
        ...
        break;

    default:
        return;
}

var s = o.ToString(); // 'o' is null in at least one execution path.
```

The same limitation could appear in other rules too. For example we could get false negatives in S2583 or S2589.

### Related information

* SonarC# Version 7.5+ (after pattern matching support was added to CFG and symbolic execution)

### LINKED ISSUE #2338: Rule S2259: False positive on switch statements
### Description
The rule does not recognize which cases of a switch statement are safe (all except the one that contains the `null` case).

### Repro steps
``` Csharp
using System;

{
    public class S2259
    {
        public ConsoleColor Color { get; set; }

        public void Method1(S2259 obj)
        {
            switch (obj?.Color)
            {
                case null:
                    Console.ForegroundColor = obj.Color; //not compliant
                    break;
                case ConsoleColor.Red:
                    Console.ForegroundColor = obj.Color; //compliant
                    break;
                default:
                    Console.WriteLine($"Color {obj.Color} is not supported."); //compliant
                    break;
            }
        }

        public void Method2(S2259 obj)
        {
            switch (obj?.Color)
            {
                case ConsoleColor.Red:
                    Console.ForegroundColor = obj.Color; //compliant
                    break;
                default:
                    Console.WriteLine($"Color {obj.Color} is not supported."); //not compliant
                    break;
            }
        }
    }
}
```

### Expected Result
Only the `null` cases should report S2259.

### Related information
* SonarC# Version 7.11.0.8083

abbas-sabra-sonarsource:
Fixes #2338
Fixes #2240

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -638,10 +638,7 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
 
         #region Build switch
 
-        private Block BuildSwitchStatement(SwitchStatementSyntax switchStatement, Block currentBlock) =>
-            BuildSwitchStatement_CSharp7(switchStatement, currentBlock);
-
-        private Block BuildSwitchStatement_CSharp7(SwitchStatementSyntax switchStatement, Block currentBlock)
+        private Block BuildSwitchStatement(SwitchStatementSyntax switchStatement, Block currentBlock)
         {
             var caseBlocksByValue = new Dictionary<object, Block>();
             BreakTarget.Push(currentBlock);
@@ -677,6 +674,7 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
 
                 foreach (var label in section.Labels.Reverse())
                 {
+                    // Handle C#7 pattern matching case Block 
                     if (CasePatternSwitchLabelSyntaxWrapper.IsInstance(label))
                     {
                         var casePatternSwitchLabel = (CasePatternSwitchLabelSyntaxWrapper)label;
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -41,11 +41,6 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
         private readonly Dictionary<string, JumpBlock> LabeledStatements = new Dictionary<string, JumpBlock>();
         private static readonly object GotoDefaultEntry = new object();
         private static readonly object GotoNullEntry = new object();
-        private static readonly HashSet<SyntaxKind> csharp6SwitchLabelKinds = new HashSet<SyntaxKind>
-        {
-            SyntaxKind.CaseSwitchLabel,
-            SyntaxKind.DefaultSwitchLabel,
-        };
 
 
         internal CSharpControlFlowGraphBuilder(CSharpSyntaxNode node, SemanticModel semanticModel)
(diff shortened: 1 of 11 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -644,69 +639,17 @@ private Block BuildGotoStatement(GotoStatementSyntax statement, Block currentBlo
...
-
-            if (!caseBlocksByValue.ContainsKey(GotoDefaultEntry))
-            {
-                caseBlocks.Add(currentBlock);
-            }
-
-            var switchBlock = CreateBranchBlock(switchStatement, caseBlocks);
-            return BuildExpression(switchStatement.Expression, switchBlock);
-        }
+            BuildSwitchStatement_CSharp7(switchStatement, currentBlock);
```
andrei-epure-sonarsource: I guess you can now rename it to `BuildSwitchStatement`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -737,58 +680,24 @@ private Block BuildSwitchStatement_CSharp7(SwitchStatementSyntax switchStatement
...
+                        currentSectionBlock = BuildExpression(switchStatement.Expression, CreateBinaryBranchBlock(simpleCaseLabel,
+                            sectionBlock, currentSectionBlock));
+                        var key = GetCaseIndexer(simpleCaseLabel.Value);
+                        caseBlocksByValue[key] = sectionBlock;
                     }
                 }
             }
 
             BreakTarget.Pop();
-
```
andrei-epure-sonarsource: [nitpick] please revert removing the empty lines - it adds some noise to the diff

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
```diff
@@ -137,6 +137,10 @@ protected override void VisitBinaryBranch(BinaryBranchBlock binaryBranchBlock, E
                     VisitBinaryBranch(binaryBranchBlock, node, ((CatchFilterClauseSyntax)binaryBranchBlock.BranchingNode).FilterExpression);
                     return;
 
+                case SyntaxKind.CaseSwitchLabel when !CasePatternSwitchLabelSyntaxWrapper.IsInstance(binaryBranchBlock.BranchingNode):
```
andrei-epure-sonarsource: sorry, it's hard to tell what `!CasePatternSwitchLabelSyntaxWrapper.IsInstance(binaryBranchBlock.BranchingNode)` - could you add a brief comment saying why this `when` filter is used?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
```diff
@@ -616,6 +620,32 @@ private void VisitConditionalAccessBinaryBranch(BinaryBranchBlock binaryBranchBl
...
+                    EnqueueNewNode(new ProgramPoint(branchBlock.FalseSuccessorBlock), ps);
+                }
+            }
+            else
+            {
+                foreach (var newProgramState in sv.TrySetOppositeConstraint(ObjectConstraint.Null, ps))
+                {
+                    EnqueueNewNode(new ProgramPoint(branchBlock.TrueSuccessorBlock), newProgramState);
+                }
+                EnqueueNewNode(new ProgramPoint(branchBlock.FalseSuccessorBlock), ps);
```
andrei-epure-sonarsource: why there is no `foreach` needed here, for the `FalseSuccessorBlock`?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/SymbolicValues/SymbolicValue.cs
```diff
@@ -241,7 +241,6 @@ protected IEnumerable<ProgramState> ThrowIfTooMany(IEnumerable<ProgramState> sta
                 {
                     return Enumerable.Empty<ProgramState>();
                 }
-
```
andrei-epure-sonarsource: could you please revert this empty line change?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.cs
```diff
@@ -1236,41 +1236,111 @@ public static void M(MyStructWithOperator a)
             }
         }
 
-        public class NullableCases
+    public class NullableCases
+    {
+      void Case1()
```
andrei-epure-sonarsource: is there any change here in `Case1()`? if there's only whitespace changes, please revert

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
andrei-epure-sonarsource: ping

PR #2520, opened 2019-07-30
Fix the cfg of ref arg

### LINKED ISSUE #2476: False positive S3900
### Description

I observed a false positive for [rule 3900](https://rules.sonarsource.com/csharp/RSPEC-3900). 

### Repro steps
Consider the following method:

```
public void Foo(params string[] infixes)
{
    if(infixes == null  || infixes.Length == 0) {
        infixes = new string[1];
    }
    else
    {
        Array.Resize(ref infixes, infixes.Length + 1); // S3900, Refactor this method to add validation of parameter 'infixes' before using it.
    }
    // more stuff
}

```

Run it through Sonar.

### Expected behavior

No warning, I dont see how infixes could be null in the else branch.

### Actual behavior

S3900 violation is reported by Sonar

### Known workarounds

-

### Related information

* SonarC# Version: 7.15 build 8572
* Visual Studio Version: 16.1.5
* Scanner for MSBuild Version: 4.6.2

abbas-sabra-sonarsource:
fixes #2476

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -1011,11 +1011,29 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
         private Block BuildInvocationLikeExpression(ExpressionSyntax parent, Block currentBlock,
             ExpressionSyntax child, IEnumerable<ArgumentSyntax> arguments)
         {
-            var args = arguments == null
-                ? Enumerable.Empty<ExpressionSyntax>()
-                : arguments.Select(a => a.Expression);
+            currentBlock.ReversedInstructions.Add(parent);
+            var isNameof = parent is InvocationExpressionSyntax invocation
+                && invocation.IsNameof(this.semanticModel);
 
-            return BuildSimpleNestedExpression(parent, currentBlock, new[] { child }.Concat(args));
+            if (isNameof)
+                return currentBlock;
+
+            foreach (var arg in arguments.Reverse())
+            {
+                if (arg.RefOrOutKeyword.IsKind(SyntaxKind.RefKeyword))
+                {
+                    currentBlock = BuildExpression(arg.Expression, currentBlock);
+                }
+            }
+
+            foreach (var arg in arguments.Reverse())
+            {
+                if (!arg.RefOrOutKeyword.IsKind(SyntaxKind.RefKeyword))
+                {
+                    currentBlock = BuildExpression(arg.Expression, currentBlock);
+                }
+            }
+            return BuildExpression(child, currentBlock);
         }
 
         private Block BuildObjectCreationExpression(ObjectCreationExpressionSyntax expression, Block currentBlock)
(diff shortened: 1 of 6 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -1011,11 +1011,29 @@ private Block BuildImplicitElementAccessExpression(ImplicitElementAccessSyntax e
...
-            var args = arguments == null
-                ? Enumerable.Empty<ExpressionSyntax>()
-                : arguments.Select(a => a.Expression);
+            currentBlock.ReversedInstructions.Add(parent);
+            var isNameof = parent is InvocationExpressionSyntax invocation
+                && invocation.IsNameof(this.semanticModel);
 
-            return BuildSimpleNestedExpression(parent, currentBlock, new[] { child }.Concat(args));
+            if (isNameof)
+                return currentBlock;
```
andrei-epure-sonarsource: how is this reflected in the tests? can you add a comment to explain ?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -3561,6 +3561,26 @@ public void Cfg_Default_Statement_First()
...
+        {
+            var cfg = Build(@"Bye(ref x0, x1, x2, x3, ref x4);}");
+            VerifyCfg(cfg, 2);
+            var instructionsBlock = (SimpleBlock)cfg.Blocks.ElementAt(0);
+            VerifyAllInstructions(instructionsBlock, "Bye", "x1", "x2", "x3", "x0", "x4", "Bye(ref x0, x1, x2, x3, ref x4)");
+        }
+
+        [TestMethod]
+        [TestCategory("CFG")]
+        public void Cfg_Ref_Arg_Should_Be_Last_Instruction2()
```
andrei-epure-sonarsource: maybe s/`Cfg_Ref_Arg_Should_Be_Last_Instruction2`/`Cfg_Ref_Arg_Should_Be_Last_Instruction_WithMethodCallOnObject`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -3561,6 +3561,26 @@ public void Cfg_Default_Statement_First()
             defaultBlock.SuccessorBlock.Should().Be(exitBlock);
         }
 
+        [TestMethod]
+        [TestCategory("CFG")]
+        public void Cfg_Ref_Arg_Should_Be_Last_Instruction1()
```
andrei-epure-sonarsource: these 2 tests should not be inside the `switch` region, but rather in a region of their own

TBH we could split this file in multiple files using the `partial class` feature, the same way we do with the SonarSecurity InstructionFactory tests; but that would be in a different PR

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -1011,11 +1011,29 @@ private Block BuildImplicitElementAccessExpression(ImplicitElementAccessSyntax e
...
+
+            foreach (var arg in arguments.Reverse())
+            {
+                if (arg.RefOrOutKeyword.IsKind(SyntaxKind.RefKeyword))
```
andrei-epure-sonarsource: why are you putting `ref` arguments first and non-`ref` after? please add comments

abbas-sabra-sonarsource: ill add a comment. ref argument is put last not first, since it is in reverse ;)

PR #2521, opened 2019-07-30
Fix nullable struct S2589 FP

### LINKED ISSUE #2496: S2583,S2589 FP: with nullable struct - update symbolic execution engine
### Description

FP for rule [Boolean expressions should not be gratuitous](https://rules.sonarsource.com/csharp/RSPEC-2589) when a nullable struct like [Span](https://docs.microsoft.com/en-us/dotnet/api/system.span-1?view=netstandard-2.1) or `ReadOnlySpan` is null-checked inside an `if`.

### Repro steps

```
public void EmptyCollectionCheck(ReadOnlySpan<byte> span)
{
    // Check for empty collection with == override
    if (span == null) // FP S2589 Change this condition so it does not always evaluate to true
    {
        return;
    }
}
```

### Expected behavior

No issue

### Related information

* SonarC# Version 7.15

Initially reported [on community](https://community.sonarsource.com/t/c-ref-struct-span-and-readonlyspan-false-positive-as-sonarlint-understands-it-cannot-be-null/10090)

abbas-sabra-sonarsource:
fixes #2496

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/AbstractExplodedGraph.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/AbstractExplodedGraph.cs
@@ -436,7 +436,11 @@ namespace SonarAnalyzer.SymbolicExecution
         {
             return type != null &&
                 type.IsValueType &&
-                !type.OriginalDefinition.Is(KnownType.System_Nullable_T);
+                !type.OriginalDefinition.Is(KnownType.System_Nullable_T)
+                && !(type.OriginalDefinition.TypeKind == TypeKind.Struct && type.SpecialType == SpecialType.None);
+                // if it is a non nullable struct comparing it to null is a compilation error.
+                // This will eliminate FP and should not generate FN when the Program compiles.
+                // TODO: remove last condition when we can compile ref struct (nullable struct).
         }
     }
 }
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/ConditionEvaluatesToConstantTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/ConditionEvaluatesToConstantTest.cs
@@ -23,6 +23,8 @@ using csharp::SonarAnalyzer.Rules.CSharp;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarAnalyzer.UnitTest.TestFramework;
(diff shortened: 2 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/AbstractExplodedGraph.cs
```diff
@@ -436,7 +436,11 @@ private static bool IsNonNullableValueType(ITypeSymbol type)
         {
             return type != null &&
                 type.IsValueType &&
-                !type.OriginalDefinition.Is(KnownType.System_Nullable_T);
+                !type.OriginalDefinition.Is(KnownType.System_Nullable_T)
+                && !(type.OriginalDefinition.TypeKind == TypeKind.Struct && type.SpecialType == SpecialType.None);
+                // if it is a non nullable struct comparing it to null is a compilation error.
+                // This will eliminate FP and should not generate FN when the Program compiles.
+                // TODO: remove last condition when we can compile ref struct (nullable struct).
```
andrei-epure-sonarsource: should there be a ticket for this?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/ConditionEvaluatesToConstantTest.cs
```diff
@@ -46,5 +48,15 @@ public void ConditionEvaluatesToConstant_FromCSharp7()
...
         }
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void ConditionEvaluatesToConstant_FromLatest()
+        {
+            Verifier.VerifyAnalyzer(@"TestCases\ConditionEvaluatesToConstant.CSharpLatest.cs",
+                new ConditionEvaluatesToConstant(),
+                ImmutableArray.Create(new CS.CSharpParseOptions(CS.LanguageVersion.Latest)),
+                CompilationErrorBehavior.Ignore); // Fix ignoring compilation when nullable struct is supported
```
andrei-epure-sonarsource: is the comment a TODO or it explains what the code is doing?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/AbstractExplodedGraph.cs
abbas-sabra-sonarsource: Yes

andrei-epure-sonarsource:
@abbas-sabra-sonarsource if this PR is not relevant anymore (the ticket is put in the backlog and not worked on), please close it.

PR #2524, opened 2019-07-31
Fix S2583/S2589 FP due to foreach loops executed only once (#1160, #1187, #2348)

### LINKED ISSUE #1160: Loops should be executed more than once
### Description

The Data Flow engine executes loops only once and this sometimes results in false positives. The loops should be executed twice (or more) so that we can catch more states of the Boolean variables.

### Repro steps

The following code generates false positive:

```
public bool AreFullyQualified(string[] files)
{
    bool anyPathRooted = false;
    bool allPathsRooted = true;
    foreach (var file in files)
    {
        if (Path.IsPathRooted(file))
        {
            anyPathRooted = true;
        }
        else
        {
            allPathsRooted = false;
        }
    }
    if (anyPathRooted && !allPathsRooted)
//                        ^^^^^^^^^^^^^^ Change this condition so that it does not always evaluate to 'false'; some subsequent code is never executed
    {
        throw new InvalidOperationException("Paths must be all rooted or all unrooted");
    }
    return allPathsRooted;
}
```

### Expected behavior

No issue is raised.

### LINKED ISSUE #1187: Fix S2583: False positive for variables updated in foreach loop
### Description

There's S2583 generated for method below for line
if (do1 && do2)
while actually the conditon evaluates to true

### Repro steps

```
        public void DoTest()
        {
            bool do1 = false;
            bool do2 = false;
            var items = new int[] {1, 2, 3};
            foreach (var item in items)
            {
                switch (item)
                {
                    case 1:
                        do1 = true;
                        break;
                    case 2:
                        do2 = true;
                        break;
                }
            }

            if (do1 && do2)
            {
                throw new InvalidOperationException();
            }
        }
```
### Expected behavior

No S2583 generated for if (do1 && do2)

### Actual behavior

Warning	S2583	Change this condition so that it does not always evaluate to 'false'; some subsequent code is never executed.	

### Related information

* SonarC# Version 6.8.1.4648
* Visual Studio Version 15.5.7

### LINKED ISSUE #2348: Fix S2589: False Positive with foreach loop
The following code flags S2589 on the return line. Is this a false positive?

```csharp

namespace S2589_False_Positive
{
    static class Class
    {
        private static bool ListContainsAtLeastOnePositiveAndNoNegatives(List<int> list)
        {
            bool containspositive = false;
            bool containsnegative = false;

            foreach (int value in list)
            {
                if (value > 0)
                {
                    containspositive = true;
                }
                else if (value < 0)
                {
                    containsnegative = true;
                }
            }

            return containspositive && !containsnegative;
        }
    }
}
```

### Related information

* SonarC# Version: Latest
* Visual Studio Version: 15.9.9

christophe-zurn-sonarsource:
fixes #1160
fixes #1187 
fixes #2348

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/AbstractExplodedGraph.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/AbstractExplodedGraph.cs
@@ -316,6 +316,8 @@ namespace SonarAnalyzer.SymbolicExecution
                     .Contains(field.ContainingType));
         }
 
+        protected abstract Block GetForEachExitBlock(Block block);
+
         #endregion Visit*
 
         #region Enqueue exploded graph node
@@ -358,14 +360,24 @@ namespace SonarAnalyzer.SymbolicExecution
                 this.programPoints[pos] = pos;
             }
 
+            ExplodedGraphNode newNode = null;
             if (programState.GetVisitedCount(pos) >= MaxProgramPointExecutionCount)
             {
                 OnProgramPointVisitCountExceedLimit(pos, programState);
-                return;
+
+                // reached the max number of visit by program point, so we take the foreach loop false branch with current program state
+                var exitLoopPoint = GetForEachExitBlock(programPoint.Block);
+                if (exitLoopPoint != null)
+                {
+                    newNode = new ExplodedGraphNode(new ProgramPoint(exitLoopPoint), programState);
+                }
+            }
+            else
+            {
+                newNode = new ExplodedGraphNode(pos, programState.AddVisit(pos));
             }
 
-            var newNode = new ExplodedGraphNode(pos, programState.AddVisit(pos));
-            if (this.nodesAlreadyInGraph.Add(newNode))
+            if (newNode != null && this.nodesAlreadyInGraph.Add(newNode))
             {
                 this.workList.Enqueue(newNode);
             }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/AbstractExplodedGraph.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/AbstractExplodedGraph.cs
@@ -365,11 +365,11 @@ namespace SonarAnalyzer.SymbolicExecution
             {
                 OnProgramPointVisitCountExceedLimit(pos, programState);
 
-                // reached the max number of visit by program point, so we take the foreach loop false branch with current program state
-                var exitLoopPoint = GetForEachExitBlock(programPoint.Block);
-                if (exitLoopPoint != null)
+                // reached the max number of visit by program point, in the case of ForEach blocks, we take the foreach loop false branch with current program state,
+                // if it is not a foreach loop, newNode will be null and we will stop exploring the path
+                if (GetForEachExitBlock(programPoint.Block) is Block forEachExitBlock)
                 {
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/AbstractExplodedGraph.cs
```diff
@@ -358,14 +360,24 @@ protected void EnqueueNewNode(ProgramPoint programPoint, ProgramState programSta
                 this.programPoints[pos] = pos;
             }
 
+            ExplodedGraphNode newNode = null;
             if (programState.GetVisitedCount(pos) >= MaxProgramPointExecutionCount)
             {
                 OnProgramPointVisitCountExceedLimit(pos, programState);
-                return;
+
+                // reached the max number of visit by program point, so we take the foreach loop false branch with current program state
```
andrei-epure-sonarsource: given that this gets executed for everything, not only `ForEach` blocks, I would change the comment to emphasize that if *might* be a for each...

and also please add a comment that if it is _not_ a foreach block AND it has reached the MaxNumberOfSteps, the `newNode` will be null, thus it will end the exploration

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/AbstractExplodedGraph.cs
```diff
@@ -358,14 +360,24 @@ protected void EnqueueNewNode(ProgramPoint programPoint, ProgramState programSta
...
 
+            ExplodedGraphNode newNode = null;
             if (programState.GetVisitedCount(pos) >= MaxProgramPointExecutionCount)
             {
                 OnProgramPointVisitCountExceedLimit(pos, programState);
-                return;
+
+                // reached the max number of visit by program point, so we take the foreach loop false branch with current program state
+                var exitLoopPoint = GetForEachExitBlock(programPoint.Block);
+                if (exitLoopPoint != null)
```
andrei-epure-sonarsource: you can improve the readability by using `is`

```
if (GetForEachExitBlock(programPoint.Block) is Block forEachExitBlock)
{
  newNode = new ExplodedGraphNode(new ProgramPoint(forEachExitBlock), programState);
}
```

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SymbolicExecution/AbstractExplodedGraph.cs
```diff
@@ -358,14 +360,24 @@ protected void EnqueueNewNode(ProgramPoint programPoint, ProgramState programSta
                 this.programPoints[pos] = pos;
             }
 
+            ExplodedGraphNode newNode = null;
```
abbas-sabra-sonarsource: MaxProgramPointExecutionCount is currently equal to 2. Did you look at the possibility of increasing it to catch more possible states? other than that the PR looks good to me

christophe-zurn-sonarsource: Increasing it would have an impact on performance, so I believe it's a discussion to have outside this ticket.
I added a test case in the PR to show the limitation (and FP) it can have

abbas-sabra-sonarsource: Great!

Issue #474, opened 2017-06-26
Improve CFG for `try` statement: connect finally statements after break/continue/goto


valhristov:
Consider the following code:
```
while (a)
{
    try
    {
        continue; // for simplicity this is not conditional
    }
    finally
    {
        ...
    }
}
```

## Actual Behavior
The `finally` block in the CFG will not be connected to the `continue` block.

## Expected Behavior
The `finally` block is connected to each path that exits the `try/catch` statement, including `return` (already handled), `continue`, `break` and `throw` for unhandled exceptions. Consider supporting `goto` as well.

## Possible Solution
We could investigate if we duplicate the `finally` block for each exit path will have negative side effects.

andrei-epure-sonarsource:
Part of this (maybe split into sub-tasks) should be:
- remove the 2 finally catch blocks
- fix the weird shape where there is a `break` inside the try inside a loop

andrei-epure-sonarsource:
due to the complexity of the necessary changes, we will probably need to do a sprint only on improving the CFG representation of the try.

See #2550 for explanations for the `try-catch` part. This is related.

gregory-paidis-sonarsource:
Closed as we do not work anymore in the old SE engine.

PR #2526, opened 2019-07-31
Fix FN introduced in #2511 (commit 4bd9516)

andrei-epure-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DeadStores.cs
```diff
@@ -111,24 +111,51 @@ private static void CheckForDeadStores(CSharpSyntaxNode node, ISymbol declaratio
...
                     possibleTry.BranchingNode.IsKind(SyntaxKind.TryStatement))
                 {
-                    // we ignore the "uncaught exception flow" block because it gives FPs when inside a loop
-                    // this leads to FNs because we cannot tell what block is actually a `finally` block
-                    // so any block which follows a Try statement and points to EXIT will be ignored
-                    continue;
+                    var instruction = block.Instructions.First();
+                    while (instruction.Parent != null)
+                    {
+                        if (instruction.IsKind(SyntaxKind.FinallyClause) &&
```
andrei-epure-sonarsource: I guess an optimisation is to stop when the `FinallyClause` is reached, in any case (either return `true`, either stop)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DeadStores.cs
```diff
@@ -111,24 +111,58 @@ private static void CheckForDeadStores(CSharpSyntaxNode node, ISymbol declaratio
...
+             * It will return `false` when it finds a second `finally` block that leads to EXIT
+             * (a "normal flow" can lead to EXIT as well, in case the `finally` is at the end of the method)
+             */
+            bool IsUncaughtExceptionFinallyBlock(Block block)
+            {
                 if (block is SimpleBlock simpleBlock &&
                     !(block is JumpBlock) &&
                     simpleBlock.SuccessorBlock is ExitBlock &&
+                    // we need an instruction to see if it's inside a finally clause
+                    block.Instructions.Count > 0 &&
```
abbas-sabra-sonarsource: we need instruction to know because we haven't added the final clause in the cfg. Did you look at that possibility to avoid this problem( but indeed empty finally block is rare?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DeadStores.cs
```diff
@@ -111,24 +111,58 @@ private static void CheckForDeadStores(CSharpSyntaxNode node, ISymbol declaratio
...
                     possibleTry.BranchingNode.IsKind(SyntaxKind.TryStatement))
                 {
-                    // we ignore the "uncaught exception flow" block because it gives FPs when inside a loop
-                    // this leads to FNs because we cannot tell what block is actually a `finally` block
-                    // so any block which follows a Try statement and points to EXIT will be ignored
-                    continue;
+                    var instruction = block.Instructions.First();
+                    while (instruction.Parent != null)
+                    {
+                        if (instruction.IsKind(SyntaxKind.FinallyClause))
```
abbas-sabra-sonarsource: should we check that the kind is not a try or catch? because it might be a try-catch inside a finally.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DeadStores.cs
andrei-epure-sonarsource: I tried to add the `final` clause, but it seems hacky and quite an expensive hack for what I hope to fix in #474

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DeadStores.cs
andrei-epure-sonarsource: good catch, I added a test for that which indeed is a problem

PR #2533, opened 2019-08-02
fix 2442

### LINKED ISSUE #2442: S2583 FP: certains configuration of conditions lead to wrong detection of dead code
**Repro steps**

```
/// <summary>
/// A certain combination of condition wrongly considers the else code as dead.
/// </summary>
private static void FalsePositive()
{
	FalsePositive2_Sub(true, false, false);
	FalsePositive2_Sub(true, false, true);

	FalsePositive2_Sub(true, true, true);
	FalsePositive2_Sub(true, true, false);

	FalsePositive2_Sub(false, false, false);
	FalsePositive2_Sub(false, false, true);

	FalsePositive2_Sub(false, true, true);
	FalsePositive2_Sub(false, true, false);

	// Outcome.
	Console.WriteLine((_test1 && _test2 && _test3 && _test4) ? "Went through each test condition" : "Missed at least one test condition");        //NOSONAR.
}

private static bool _test1 = false;
private static bool _test2 = false;
private static bool _test3 = false;
private static bool _test4 = false;

private static void FalsePositive_Sub(bool testCondition1, bool testCondition2, bool testCondition3)
{
	bool condition1 = testCondition1;
	bool condition2 = testCondition2;
	bool condition3 = condition2 && testCondition3;

	if (condition2 && condition3 && condition1)
	{
		_test1 = true;
	}

	else if (!condition2 && !condition1)
	{
		_test2 = true;
	}

	else if (condition2 && condition1 && !condition3)
	{
		_test3 = true;
	}

	else
	{
		_test4 = true;
	}
}
```

**Expected behavior**

Rule [S2583](https://rules.sonarsource.com/csharp/RSPEC-2583) should not be triggered

**Actual Behavior**

The else code should not be considered as dead code, as at least one combination of conditions triggers it.

**Related information**

Sonar C# version  7.11

abbas-sabra-sonarsource:
fixes #2442


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
```diff
@@ -195,7 +195,15 @@ private static IEnumerable<Location> GetUnreachableLocations(SyntaxNode constant
                 // parent has an unreachable branch.
                 if (constantExpression is ExpressionSyntax expression)
                 {
-                    unreachableExpressions.Add(expression);
+                    bool unreachableCodeIsDetectable = constantValue ?
+                    !expression.Parent.IsKind(SyntaxKind.LogicalAndExpression)
```
christophe-zurn-sonarsource: When looking at the parent expression, we should skip parenthesis, otherwise we will still have FP for unreachable code in this case:
```
            if (unknown && (trueConstant)) // Noncompliant
            {

            }
            else
            { // Secondary is raised here for dead code, but it may not be depending on the unknown value

            }
```

abbas-sabra-sonarsource: good point

PR #2544, opened 2019-08-05
Fix regression (Symbolic Execution exception which breaks cirrus QA) introduced in #2511

andrei-epure-sonarsource:
Note: this was making many rules fail: NullPointerDereference, ObjectsShouldNotBeDisposedMoreThanOnce, EmptyNullableValueAccess, ConditionEvaluatesToConstant, InvalidCastToInterface, EmptyCollectionsShouldNotBeEnumerated

I have added just some partial UTs that go over the exploded graph


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharpExplodedGraph.cs
```diff
@@ -393,6 +393,8 @@ protected override void VisitInstruction(ExplodedGraphNode node)
 
                 case SyntaxKind.CheckedExpression:
                 case SyntaxKind.UncheckedExpression:
+
+                case SyntaxKind.BreakStatement:
```
andrei-epure-sonarsource: this is probably a wrong fix, the problem is the `break` being part of the instructions of the `JumpBlock`

PR #2545, opened 2019-08-06
Switch statements and lambda expressions

amelie-renard-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/MLIRExporter.cs
```diff
@@ -76,7 +76,10 @@ private bool IsTooClomplexForMLIROrTheCFG(MethodDeclarationSyntax method)
             n.IsKind(SyntaxKind.LogicalOrExpression) ||
             n.IsKind(SyntaxKind.ConditionalExpression) ||
             n.IsKind(SyntaxKind.ConditionalAccessExpression) ||
-            n.IsKind(SyntaxKind.CoalesceExpression)
+            n.IsKind(SyntaxKind.CoalesceExpression) ||
+            n.IsKind(SyntaxKind.SwitchStatement) ||
+            n.IsKind(SyntaxKind.ParenthesizedLambdaExpression) ||
+            n.IsKind(SyntaxKind.SimpleLambdaExpression)
```
loic-joly-sonarsource: I don't see a test case for `SimpleLambdaExpression`

PR #2557, opened 2019-08-06
Fix FP S3966: Rule should consider the `leaveOpen` parameter

### LINKED ISSUE #2491: FP S3966: Rule should consider the `leaveOpen` parameter
### Description

Rule [Objects should not be disposed more than once](https://sonarcloud.io/organizations/budzow-github/rules?open=csharpsquid%3AS3966&q=S3966) raises an issue when `Dispose` would be called twice. However, we do not take into consideration the `leaveOpen` parameter of the `StreamWriter` class, which ensures that disposed will not be called on the `stream` parameter.

From the [StreamWriter documentation](https://docs.microsoft.com/en-us/dotnet/api/system.io.streamwriter.-ctor?view=netframework-4.8)

> leaveOpen : true to leave the stream open after the StreamWriter object is disposed; otherwise, false.

> Unless you set the leaveOpen parameter to true, the StreamWriter object calls Dispose() on the provided Stream object when StreamWriter.Dispose is called. 

### Repro steps

```
    public class ReproFromSonarCloud
    {
        public void S3966FalsePositive()
        {
            // Rule S3966 is a false positive:
            // The StreamWriter constructor used sets 'leaveOpen' to true, which prevents the
            // StreamWriter from disposing the MemoryStream. Dispose is only called due to the
            // 'using'-statement.

            using (MemoryStream memoryStream = new MemoryStream())
            using (StreamWriter writer = new StreamWriter(memoryStream, new System.Text.UTF8Encoding(false), 1024, leaveOpen: true)) // Noncompliant FP
            {
                //...
            }

        }
    }
```

### Expected behavior

The rule should check if the `leaveOpen` parameter is provided and `true` and should not raise this FP.

### Related information

* SonarC# Version: 7.15

This FP was raised by a customer who provided us with a very helpful [reproducer on SonarCloud](https://sonarcloud.io/code?id=ConsoleApp32&selected=ConsoleApp32%3AS3966.cs). Thank you!

christophe-zurn-sonarsource:
Fixes #2491

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
@@ -29,6 +29,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.ControlFlowGraph.CSharp;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.Helpers.CSharp;
 using SonarAnalyzer.SymbolicExecution;
 using SonarAnalyzer.SymbolicExecution.Constraints;
 
@@ -40,6 +41,7 @@ namespace SonarAnalyzer.Rules.CSharp
     {
         internal const string DiagnosticId = "S3966";
         private const string MessageFormat = "Refactor this code to make sure '{0}' is disposed only once.";
+        private const string LeaveOpenParameterName = "leaveOpen";
 
         private static readonly ImmutableArray<KnownType> typesDisposingUnderlyingStream =
             ImmutableArray.Create(
@@ -177,20 +179,43 @@ namespace SonarAnalyzer.Rules.CSharp
             private ProgramState ProcessStreamDisposingTypes(ProgramState programState, SyntaxNode usingExpression)
             {
                 var newProgramState = programState;
+                var objectCreationExpressions = usingExpression.DescendantNodes()
+                    .OfType<ObjectCreationExpressionSyntax>();
 
-                var arguments = usingExpression.DescendantNodes()
-                    .OfType<ObjectCreationExpressionSyntax>()
-                    .Where(IsStreamDisposingType)
-                    .Select(FirstArgumentOrDefault)
-                    .WhereNotNull();
-
-                foreach (var argument in arguments)
+                foreach (var objectCreationExpression in objectCreationExpressions)
                 {
-                    var streamSymbol = this.semanticModel.GetSymbolInfo(argument.Expression).Symbol;
-                    newProgramState = ProcessDisposableSymbol(newProgramState, argument.Expression, streamSymbol);
+                    var objectType = this.semanticModel.GetSymbolInfo(objectCreationExpression.Type)
+                    .Symbol
+                    .GetSymbolType();
+
+                    if (objectType.DerivesOrImplementsAny(typesDisposingUnderlyingStream) &&
+                        objectCreationExpression.ArgumentList?.Arguments is IEnumerable<ArgumentSyntax> argumentList)
+                    {
+                        // Checks for the 'leaveOpen' argument. It is at position 4 if present for StreamReader constructor, while it is at position 3 for StreamWriter.
+                        // See #2491 for more details
+                        var leaveOpenArgumentIndex = objectType.DerivesOrImplements(KnownType.System_IO_StreamReader) ? 4 : 3;
+                        if (!HasLeaveOpenArgument(argumentList, leaveOpenArgumentIndex))
+                        {
+                            var argument = argumentList.First();
+                            var streamSymbol = this.semanticModel.GetSymbolInfo(argument.Expression).Symbol;
+                            newProgramState = ProcessDisposableSymbol(newProgramState, argument.Expression, streamSymbol);
+                        }
+                    }
                 }
 
                 return newProgramState;
+            }
+
+            private static bool HasLeaveOpenArgument(IEnumerable<ArgumentSyntax> argumentList, int leaveOpenArgumentIndex)
+            {
+                var leaveOpenArgument = argumentList.FirstOrDefault(argument => LeaveOpenParameterName.Equals(argument.NameColon?.Name.Identifier.ValueText));
+                if (leaveOpenArgument == null)
+                {
+                    leaveOpenArgument = argumentList.ElementAtOrDefault(leaveOpenArgumentIndex);
+                }
+
+                return leaveOpenArgument != null &&
+                    CSharpEquivalenceChecker.AreEquivalent(leaveOpenArgument.Expression, CSharpSyntaxHelper.TrueLiteralExpression);
             }
 
             private ProgramState ProcessDisposableSymbol(ProgramState programState, SyntaxNode disposeInstruction,
@@ -224,15 +249,6 @@ namespace SonarAnalyzer.Rules.CSharp
                 }
                 return disposableSymbol.SetConstraint(DisposableConstraint.Disposed, newProgramState);
             }
-
-            private static ArgumentSyntax FirstArgumentOrDefault(ObjectCreationExpressionSyntax objectCreation) =>
-                objectCreation.ArgumentList?.Arguments.FirstOrDefault();
-
-            private bool IsStreamDisposingType(ObjectCreationExpressionSyntax objectCreation) =>
-                this.semanticModel.GetSymbolInfo(objectCreation.Type)
-                    .Symbol
-                    .GetSymbolType()
-                    .DerivesOrImplementsAny(typesDisposingUnderlyingStream);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
```diff
@@ -177,22 +179,45 @@ private ProgramState VisitInvocationExpression(InvocationExpressionSyntax instru
...
-                    .Select(FirstArgumentOrDefault)
-                    .WhereNotNull();
-
-                foreach (var argument in arguments)
+                foreach (var objectCreationExpression in objectCreationExpressions)
                 {
-                    var streamSymbol = this.semanticModel.GetSymbolInfo(argument.Expression).Symbol;
-                    newProgramState = ProcessDisposableSymbol(newProgramState, argument.Expression, streamSymbol);
+                    var objectType = this.semanticModel.GetSymbolInfo(objectCreationExpression.Type)
+                    .Symbol
```
andrei-epure-sonarsource: indent please

andrei-epure-sonarsource: also, I think `Symbol` could be null if rolsyn doesn't manage to resolve it (`SymbolInfo` is not null because it's a `struct` but `Symbol` can be null, and we're usually defensive because the analyzer can run on the fly inside the IDE when not everything compiles)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
```diff
@@ -177,22 +179,45 @@ private ProgramState VisitInvocationExpression(InvocationExpressionSyntax instru
...
+                    .GetSymbolType();
+
+                    if (objectType.DerivesOrImplementsAny(typesDisposingUnderlyingStream) &&
+                        objectCreationExpression.ArgumentList?.Arguments is IEnumerable<ArgumentSyntax> argumentList)
+                    {
+                        // Checks for the 'leaveOpen' argument. It is at position 4 if present for StreamReader constructor, while it is at position 3 for StreamWriter.
+                        // See #2491 for more details
+                        var leaveOpenArgumentIndex = objectType.DerivesOrImplements(KnownType.System_IO_StreamReader) ? 4 : 3;
```
andrei-epure-sonarsource: I would keep this info in a map where key is the type and values is the index. Like this, if a new type will be added, everything will be in the same place. And you can still use `Dictionary.Keys` instead of `typesDisposingUnderlyingStream` for the call `DerivesOrImplementsAny`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
```diff
@@ -177,22 +179,45 @@ private ProgramState VisitInvocationExpression(InvocationExpressionSyntax instru
...
+                        if (!HasLeaveOpenArgument(argumentList, leaveOpenArgumentIndex))
+                        {
+                            var argument = argumentList.First();
```
andrei-epure-sonarsource: could `argumentList` be empty in some future API version? it's safer to check the ArgumentList size before accessing the list

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
```diff
@@ -177,22 +179,45 @@ private ProgramState VisitInvocationExpression(InvocationExpressionSyntax instru
...
+                foreach (var objectCreationExpression in objectCreationExpressions)
                 {
-                    var streamSymbol = this.semanticModel.GetSymbolInfo(argument.Expression).Symbol;
-                    newProgramState = ProcessDisposableSymbol(newProgramState, argument.Expression, streamSymbol);
+                    var objectType = this.semanticModel.GetSymbolInfo(objectCreationExpression.Type)
+                    .Symbol
+                    .GetSymbolType();
+
+                    if (objectType.DerivesOrImplementsAny(typesDisposingUnderlyingStream) &&
+                        objectCreationExpression.ArgumentList?.Arguments is IEnumerable<ArgumentSyntax> argumentList)
```
andrei-epure-sonarsource: You could also check the  number of arguments, if it makes sense... Can you call the method with 2 arguments and still pass the required argument? (I'm not sure, so feel free to ignore the comment if not relevant)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ObjectsShouldNotBeDisposedMoreThanOnce.cs
```diff
@@ -165,21 +165,116 @@ public void DoSomething()
...
+            using (StreamWriter writer = new StreamWriter(memoryStream, new System.Text.UTF8Encoding(false), leaveOpen: true, bufferSize: 1024))
+            using (StreamReader reader = new StreamReader(memoryStream, new System.Text.UTF8Encoding(false), bufferSize: 2014, leaveOpen: true, detectEncodingFromByteOrderMarks: false))
+            {
+                string str = null;
+            }
+        }
+
+        public void LeaveOpenTrue7()
+        {
+            using (MemoryStream memoryStream = new MemoryStream()) // Noncompliant
```
andrei-epure-sonarsource: could you add a test where `leaveOpen: true` is passed as the first argument?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
```diff
@@ -184,19 +184,21 @@ private ProgramState ProcessStreamDisposingTypes(ProgramState programState, Synt
...
+                        argumentList.Any())
                     {
                         // Checks for the 'leaveOpen' argument. It is at position 4 if present for StreamReader constructor, while it is at position 3 for StreamWriter.
                         // See #2491 for more details
-                        var leaveOpenArgumentIndex = objectType.DerivesOrImplements(KnownType.System_IO_StreamReader) ? 4 : 3;
-                        if (!HasLeaveOpenArgument(argumentList, leaveOpenArgumentIndex))
+                        var leaveOpenArgumentIndex = typesDisposingUnderlyingStreamWithLeaveOpenArgumentIndex
+                            .Where(entry => objectType.DerivesOrImplements(entry.Key))
+                            .Select(entry => entry.Value)
+                            .First();
```
andrei-epure-sonarsource: this will throw `InvalidOperationException` if the list is empty - should it be `FirstOrDefault`?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
```diff
@@ -184,19 +184,21 @@ private ProgramState ProcessStreamDisposingTypes(ProgramState programState, Synt
...
+                        symbol.GetSymbolType() is ITypeSymbol objectType &&
+                        objectType.DerivesOrImplementsAny(typesDisposingUnderlyingStreamWithLeaveOpenArgumentIndex.Keys.ToImmutableArray()) &&
+                        objectCreationExpression.ArgumentList?.Arguments is IEnumerable<ArgumentSyntax> argumentList &&
+                        argumentList.Any())
                     {
                         // Checks for the 'leaveOpen' argument. It is at position 4 if present for StreamReader constructor, while it is at position 3 for StreamWriter.
                         // See #2491 for more details
-                        var leaveOpenArgumentIndex = objectType.DerivesOrImplements(KnownType.System_IO_StreamReader) ? 4 : 3;
-                        if (!HasLeaveOpenArgument(argumentList, leaveOpenArgumentIndex))
+                        var leaveOpenArgumentIndex = typesDisposingUnderlyingStreamWithLeaveOpenArgumentIndex
```
andrei-epure-sonarsource: isn't his a Dictionary, can't you just get the element by key?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
christophe-zurn-sonarsource: This should always work, because of the outer if-condition `objectType.DerivesOrImplementsAny(typesDisposingUnderlyingStreamWithLeaveOpenArgumentIndex.Keys.ToImmutableArray())`. It implies this `Where` query will always have (at least, although in this case it is exactly) one element

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
christophe-zurn-sonarsource: Retrieving the element by key is not the same as getting it for types that `type.DerivesOrImplements` one of these types. I did not want to change the logic, but we can be more strict and only consider these 2 exact types here

andrei-epure-sonarsource: sorry, I missed that

PR #2564, opened 2019-08-07
Improve CFG for `try` statement: replace the 2 finally blocks with one finally block

### LINKED ISSUE #2549: Improve CFG for `try` statement: replace the 2 finally blocks with one finally block
For the `finally` block, 2 different blocks are created in the CFG:

- one block for the "uncaught exception flow", which will always point to EXIT
- one block for the "normal flow", which points to the block after the `try-finally`

I believe we should have only one finally block that is a `BranchBlock` and leads to what's after the `try` and could also lead to `exit`.

christophe-zurn-sonarsource:
Fixes #2549

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
@@ -2030,26 +2030,24 @@ namespace NS
                 AfterForeach();
             ");
 
-            VerifyCfg(cfg, 9);
+            VerifyCfg(cfg, 8);
             var blocks = cfg.Blocks.ToList();
 
             var beforeForeach = (ForeachCollectionProducerBlock)blocks[0];
             var foreachDecision = (BinaryBranchBlock)blocks[1];
             var beforeTry = (SimpleBlock)blocks[2];
             var insideTry = (BranchBlock)blocks[3];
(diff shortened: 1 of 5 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -2808,38 +2783,36 @@ public void Cfg_TryCatchFinally_InsideLoop_WithBreakInsideTry_AndContinueInsideC
...
 
             VerifyAllInstructions(doBeforeTry, "cw0", "cw0()");
             doBeforeTry.SuccessorBlocks.Should().OnlyContainInOrder(tryStatement);
 
             VerifyAllInstructions(tryStatement, "cw1", "cw1()");
-            tryStatement.SuccessorBlocks.Should().OnlyContainInOrder(catchBody, finallyBody_NormalFlow, finallyBlock_Exit, afterDoWhile);
+            tryStatement.SuccessorBlocks.Should().OnlyContainInOrder(catchBody, finallyBlock, afterDoWhile);
 
-            tryBody.SuccessorBlocks.Should().OnlyContainInOrder(catchBody, finallyBody_NormalFlow, finallyBlock_Exit);
+            tryBody.SuccessorBlocks.Should().OnlyContainInOrder(catchBody, finallyBlock);
```
andrei-epure-sonarsource: [unrelated to your change] this `tryBody` is so weird, I hope we'll fix it

PR #2573, opened 2019-08-09
Generate Mlir from C# unit tests

amelie-renard-sonarsource:


loic-joly-sonarsource:
This experiment was concluded successfully, and is no longer required.

PR #2612, opened 2019-08-29
report cbde build failures

geoffray-adde-sonarsource:
un peu plus que 2 lignes au final :-/

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
@@ -31,6 +31,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using System.Collections.Immutable;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Text;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
@@ -45,6 +46,8 @@ namespace SonarAnalyzer.Rules.CSharp
         private string mlirDirectoryRoot;
         private string mlirDirectoryAssembly;
         private string cbdeJsonOutputPath;
+        private string logFilePath;
+        protected HashSet<string> csSourceFileNames= new HashSet<string>();
         protected Dictionary<string, int> fileNameDuplicateNumbering = new Dictionary<string, int>();
         private StreamWriter logFile;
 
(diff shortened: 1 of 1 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -183,6 +184,32 @@ private static Diagnostic CreateDiagnosticFromJToken(JToken token)
...
+            var begin = new LinePosition(1,1);
+            var end = new LinePosition(1,2);
+            var loc = Location.Create(filename, TextSpan.FromBounds(0, 0), new LinePositionSpan(begin, end));
+
+            return Diagnostic.Create(rule, loc);
+        }
+        private void RaiseIssueFromFailedCbdeRun(CompilationAnalysisContext context)
+        {
+            string failureString = "CBDE Failure Report :\n  C# souces files involved are:\n";
+            foreach (var fileName in csSourceFileNames)
```
loic-joly-sonarsource: Use `string.join`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -183,6 +184,32 @@ private static Diagnostic CreateDiagnosticFromJToken(JToken token)
...
+
+            var begin = new LinePosition(1,1);
+            var end = new LinePosition(1,2);
+            var loc = Location.Create(filename, TextSpan.FromBounds(0, 0), new LinePositionSpan(begin, end));
+
+            return Diagnostic.Create(rule, loc);
+        }
+        private void RaiseIssueFromFailedCbdeRun(CompilationAnalysisContext context)
+        {
+            string failureString = "CBDE Failure Report :\n  C# souces files involved are:\n";
```
loic-joly-sonarsource: Use a `StringBuilder`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
geoffray-adde-sonarsource: I am just using the StringBuilder as suggested below as re-configuring the code to use string.Join would be far less readable for a negligible perf boost.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
geoffray-adde-sonarsource: done

PR #2623, opened 2019-09-06
Fix CFG for conditional access while avoiding regression

christophe-zurn-sonarsource:


sonaranalyzer-dotnet/its/expected/Automapper/AutoMapper-025B9F93-7953-454B-AF29-E6A55B2AA58F-S3900.json
```diff
@@ -327,19 +327,6 @@
 },
 {
 "id":  "S3900",
-"message":  "Refactor this method to add validation of parameter 'typeMap' before using it.",
-"location":  {
-"uri":  "sources\Automapper\src\AutoMapper\ICtorParamConfigurationExpression.cs",
```
andrei-epure-sonarsource: I'm not sure this is right... `typeMap` is not really validated above, is it? It's only dereferenced.

But I think this is still fine because we do raise at line 56, so there is a warning for validating `typeMap`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.cs
```diff
@@ -90,7 +90,7 @@ string Prop0
 
         object Prop1
         {
-             get // Noncompliant {{Add a way to break out of this property accessor's recursion.}}
```
andrei-epure-sonarsource: if this is an accepted regression and FN, you should add a comment specifying the limitation

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereferenceCSharp7.cs
```diff
@@ -166,30 +166,53 @@ void Switch_Pattern(object o)
...
+                {
+
+                }
+            }
         }
-      }
-    }
 
-    public void Method1(object obj)
+        public void Method1(object obj)
```
andrei-epure-sonarsource: extra whitespace?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.cs
```diff
@@ -90,7 +90,7 @@ string Prop0
...
+            get
             {
                 return (new InfiniteRecursion())?.Prop1;
```
andrei-epure-sonarsource: you could replace this instantiation with adding a field `InfiniteRecursion instanceField` and call `return instanceField?.Prop1` ?

sonaranalyzer-dotnet/its/expected/Automapper/AutoMapper-025B9F93-7953-454B-AF29-E6A55B2AA58F-S3900.json
christophe-zurn-sonarsource: Rule S3900 relies on `NullPointerDereference.NullPointerCheck` to check whether a public argument can be null or not. Once a member select on a identifier has been visited, [it will set a non-null constraint](https://github.com/SonarSource/sonar-dotnet/blob/ff7413922f6565eccccda028d0d89f22d2a11c68/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NullPointerDereference.cs#L245) on the underlying identifier, in order to avoid duplicate issue.
So this was indeed a FP, as the constraint was no longer compared correctly due to the wrong CFG shape

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.cs
christophe-zurn-sonarsource: I added a test with a field access, but I kept this instanciation for reference

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.cs
christophe-zurn-sonarsource: Yes this is an accepted regression, as it was unintentional to fix this with previous PR, and the reason for the fix is because of the strange shape of CFG at that point.

PR #2625, opened 2019-09-09
Rule S4201: Null checks should not be redundant

christophe-zurn-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheckCodeFixProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        internal const string Title = "Remove this unnecessary null check";
+        public override ImmutableArray<string> FixableDiagnosticIds
+        {
+            get
+            {
+                return ImmutableArray.Create(RedundantNullCheck.DiagnosticId);
+            }
+        }
+        public override FixAllProvider GetFixAllProvider()
```
andrei-epure-sonarsource: nitpick empty line above

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RedundantNullCheck.Fixed.Batch.cs
```diff
@@ -0,0 +1,121 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Tests.Diagnostics
+{
+    public class RedundantNullCheck
```
andrei-epure-sonarsource: silly question from me: what's the difference between Batch and non-Batch?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheck.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        protected override SyntaxNode GetNonNullCheckVariable(SyntaxNode node) => GetNullCheckVariableForKind(node, SyntaxKind.NotEqualsExpression);
+
+        protected override SyntaxNode GetIsOperatorCheckVariable(SyntaxNode node)
+        {
+            var innerExpression = RemoveParentheses(node);
+            if (innerExpression is BinaryExpressionSyntax binaryExpression && binaryExpression.IsKind(SyntaxKind.IsExpression))
+            {
+                return GetLeftNode(binaryExpression);
+            } else if (innerExpression.IsKind(SyntaxKindEx.IsPatternExpression))
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: Put `else if` on new line 👍 

However, the expression cannot be replaced by `innerExpression is IsPatternExpressionSyntaxWrapper isPatternExpression` because it is part of the ShimLayer, and `IsPatternExpressionSyntaxWrapper` is not a subclass of `SyntaxNode`, but a wrapper on it

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RedundantNullCheck.Fixed.Batch.cs
christophe-zurn-sonarsource: Files are the same. From what I udnerstand, it's the way the code fix provider is executed that is differrent: depending whether the user wants to fix all the issues at the same time or not.
But resulting fixed file here are indeed identical

PR #2630, opened 2019-09-10
Rule S3881: Makes message clearer on what is missing

### LINKED ISSUE #2339: Rule S3881: message needs to be improved
### Description
See the code below. I assume it is not related to issue #1728. As far as I know I followed all recommended guidelines, and just cleaned up some code (like comments), renamed a variable, and made Dispose() using an expression body (but that is not the reason it failed as far as I know).

### Repro steps
``` CSharp
using System;

{
    public class S3881 : IDisposable //compliant
    {
        public S3881(IDisposable prop) => Property = prop;

        public IDisposable Property { get; }

        public void Dispose() => Dispose(true);

        protected virtual void Dispose(bool disposing)
        {
            if (!isDisposed)
            {
                if (disposing)
                {
                    Property.Dispose();
                }
                isDisposed = true;
            }
        }
        private bool isDisposed;
    }
}
```

### Expected Result
I would expect not to see a warning at all, obviously, but because this is a quite complex check, at least a hint on which part of the pattern is missing.

### Related information
* SonarC# Version 7.11.0.8083
* https://github.com/Corniel/SonarAnalyzer.Experiments/blob/master/src/SonarAnalyzer.Experiments.CSharp/S3881.cs

alban-auzeill:
Fixes #2339

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ImplementIDisposableCorrectly.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ImplementIDisposableCorrectly.cs
@@ -9,7 +9,7 @@ namespace Tests.Diagnostics
 
     public class SimpleDisposable : IDisposable // Noncompliant {{Fix this implementation of 'IDisposable' to conform to the dispose pattern.}}
     {
-        public void Dispose() // Secondary {{'SimpleDisposable.Dispose()' should only invoke 'Dispose(true)' and 'GC.SuppressFinalize(this)'.}}
+        public void Dispose() // Secondary {{'SimpleDisposable.Dispose()' should also call 'GC.SuppressFinalize(this)'.}}
         {
             Dispose(true);
         }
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ImplementIDisposableCorrectly.cs
```diff
@@ -103,7 +115,7 @@ public virtual void Dispose()
     public class WithFinalizer : IDisposable // Noncompliant
     {
         public void Dispose()
-//                  ^^^^^^^ Secondary {{'WithFinalizer.Dispose()' should only invoke 'Dispose(true)' and 'GC.SuppressFinalize(this)'.}}
+//                  ^^^^^^^ Secondary {{'WithFinalizer.Dispose()' should also call 'GC.SuppressFinalize(this)'.}}
         {
             Dispose(true);
         }
```
andrei-epure-sonarsource: please also add the snippet from the ticket (if not already present) with the arrow expression syntax
`public void Dispose() => Dispose(true);`

alban-auzeill: You are right, there was no `=> Dispose(true)`.
And because there are several Dispose methods containing only `Dispose(true);`, let's convert one of them with this trendy arrow expression syntax.

PR #2644, opened 2019-09-12
[C#] Rule S2252: Loop conditions should be true at least once

### LINKED ISSUE #616: Rule S2252: Loop conditions should be true at least once
[RSPEC-2252](https://jira.sonarsource.com/browse/RSPEC-2252)

If a for loop's condition is false before the first loop iteration, the loop will never be executed. Such loops are almost always bugs, particularly when the initial value and stop conditions are hard-coded.

christophe-zurn-sonarsource:
Fixes #616


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopConditionAlwaysFalse.cs
```diff
@@ -0,0 +1,187 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarAnalyzer.Common;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: As mentioned during standup, logic is a bit different for VB.NET. So implementing the rule for VB.NET deserves its own separate ticket:  see #2648

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ForLoopConditionAlwaysFalse.cs
```diff
@@ -50,6 +50,10 @@ void LoopTest(int x, int y)
             for (var i = 0; i > 0x10;) { } // Noncompliant
             for (var i = 0; i < 0b10;) { }
             for (var i = 1; i <= 0Xffff; i++) { }
+
+            var z = 0;
+            for (int i = 0; i < z; i++) { } // FN - we only check for literals in the condition
+            for (; z < 0; z++) { } // FN - we only check for initilaizers inside the loop statement
```
andrei-epure-sonarsource: s/initilaizers/initializers

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopConditionAlwaysFalse.cs
```diff
@@ -0,0 +1,218 @@
...
+                 ));
+
+        private bool IsAlwaysTrueCondition(ExpressionSyntax condition) =>
+            condition.IsKind(SyntaxKind.TrueLiteralExpression) ||
+                (
+                    IsLogicalNot(condition, out var logicalNode) &&
+                    IsAlwaysFalseCondition(logicalNode.Operand.RemoveParentheses())
+                );
+
+        private static bool IsLogicalNot(ExpressionSyntax expression, out PrefixUnaryExpressionSyntax logicalNot)
```
andrei-epure-sonarsource: instance `private` methods are mixed with `private static` methods - could you please put the static ones first, and then the non-static ones?

PR #2645, opened 2019-09-12
Rule S1199: Nested code blocks should not be used

### LINKED ISSUE #768: Rule S1199: Nested code blocks should not be used
[RSPEC-1199](https://jira.sonarsource.com/browse/RSPEC-1199)

Nested code blocks can be used to create a new scope and restrict the visibility of the variables defined inside it. Using this feature in a method typically indicates that the method has too many responsibilities, and should be refactored into smaller methods.

alban-auzeill:
Fixes #768


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NestedCodeBlock.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+
+namespace Tests.Diagnostics
+{
+    class Program
+    {
+        public void Method(int value)
+        { // Compliant, parent is MethodDeclaration
```
andrei-epure-sonarsource: I would also add **compliant** tests with nested

- `using`,
- `try-catch-finally`
- `do-while`
- `foreach`
- `unsafe { }`
- `checked {}`
- `lock (this) {}`
- `unchecked { }`
- `event EventHandler<LoadEventArgs> LoadedData = delegate { };` (on this one I'm note sure it 
makes sense, so if it doesn't skip it)
- _enum_
- a local function
```
public void Method1()
{
  var x = LocalFoo();
  bool LocalFoo()
  {
    return true;
  }
}
```

Also, **Noncompliant** tests with nested blocks inside
- _constructor_
- _property (setter and getter)_
- _struct_

PR #2646, opened 2019-09-12
remove F# from ITs

andrei-epure-sonarsource:
we don't analyze the code and it's an unnecessary dependency in our toolset

we will still have the cirrus ci jobs with the akka.net project making sure the our analyzer works with a solution with F# projects in it


sonaranalyzer-dotnet/its/sources/akka.net/src/Akka.sln
```diff
@@ -13,16 +13,6 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ChatClient", "examples\Chat
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ChatServer", "examples\Chat\ChatServer\ChatServer.csproj", "{7FD886CC-216B-4324-9AB2-241494598B06}"
 EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "FSharp", "FSharp", "{E814B432-DBC2-4372-92C6-70B9198613BD}"
-EndProject
-Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "FSharp.Api", "examples\FSharp.Api\FSharp.Api.fsproj", "{19993C0B-8F9D-4FE4-B047-45965EA66BA2}"
-EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "RemoteDeploy", "RemoteDeploy", "{68D89B2A-6770-483E-B7B0-C7EDAE0C2A6D}"
```
alban-auzeill: Do not remove `RemoteDeploy`, it does not contain *.fs files but *.cs

andrei-epure-sonarsource: it's already at line 24 - `Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "RemoteDeploy", "RemoteDeploy", "{B555BA52-1911-426B-8018-D303D7228748}"`

it seems it was a duplicate

andrei-epure-sonarsource:
regarding the `build.fsx` script, it's part of the build chain (see in build.cmd - `src\packages\FAKE\tools\FAKE.exe build.fsx` )

I might as well delete the build scripts because we don't use them in the ITs but that would be a separate PR

PR #2652, opened 2019-09-16
Rule S4136: Method overloads should be grouped together

### LINKED ISSUE #1168: Rule S4136: Method overloads should be grouped together
[RSPEC-4136](https://jira.sonarsource.com/browse/RSPEC-4136)

For clarity, all overloads of the same method should be grouped together. That lets both users and maintainers quickly understand all the current available options.

christophe-zurn-sonarsource:
Fixes #1168


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -0,0 +1,115 @@
...
+            }
+            else
+            {
+                return nameSyntaxNode?.ValueText?.ToLowerInvariant();
+            }
+        }
+
+        protected Location GetLocation(TMemberDeclarationSyntax member) => GetNameSyntaxNode(member)?.GetLocation();
+
+        protected StringComparison CaseSensitivity
```
andrei-epure-sonarsource: nitpick: the `protected` property should be above the group of `protected` methods

sonaranalyzer-dotnet/its/expected/Automapper/AutoMapper-025B9F93-7953-454B-AF29-E6A55B2AA58F-S4136.json
```diff
@@ -0,0 +1,421 @@
+{
+"issues":  [
+{
+"id":  "S4136",
+"message":  "All 'MapFrom' signatures should be adjacent",
+"location":  [
+{
+"uri":  "sources\Automapper\src\AutoMapper\Configuration\MemberConfigurationExpression.cs",
```
andrei-epure-sonarsource: this is actually a different method than the one at line 135, because it's generic. I think we should differentiate them

sonaranalyzer-dotnet/its/expected/akka.net/Akka.TestKit-{0D3CBAD0-BBDB-43E5-AFC4-ED1D3ECDC224}-S4136.json
```diff
@@ -0,0 +1,238 @@
...
+"startLine":  94,
+"startColumn":  21,
+"endLine":  94,
+"endColumn":  25
+}
+},
+{
+"uri":  "sources\akka.net\src\core\Akka.TestKit\TestActorRefBase.cs",
+"region":  {
+"startLine":  195,
```
andrei-epure-sonarsource: `void ICanTell.Tell(object message, IActorRef sender)` - should we treat [explicit interface](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation) implementations the same?

sonaranalyzer-dotnet/its/expected/akka.net/Akka.Cluster-{6AB00F61-269A-4501-B06A-026707F000A7}-S4136.json
```diff
@@ -0,0 +1,100 @@
...
+}
+}
+]
+},
+{
+"id":  "S4136",
+"message":  "All 'Equals' signatures should be adjacent",
+"location":  [
+{
+"uri":  "sources\akka.net\src\core\Akka.Cluster\ClusterMetricsCollector.cs",
```
andrei-epure-sonarsource: this is a `private` method, whereas at line 324, the method is `public` - it is common to group together based on the visibility

christophe-zurn-sonarsource:
> you should add more tests with:
> 
>     * different accessibilities for overloads (public, protected, private) - they should be able to be grouped separately

Ok, it makes sense. I'll update implementation accordingly.

>     * explicit interface implementations - `void ICanTell.Tell(object message, IActorRef sender)` and `void Tell(object message, IActorRef sender)` - I imagine someone can want to group together explicit interface implementations in a class, based on an interface (thus the overloads would be separated)

Mmm. Ok. Makes sense also. I'll add an exception for this, in both the implementation and the RSPEC description.


>     * generic overloads - `Foo<Type>()` and `Foo()` - I guess they should be grouped together, but it's good to document that in the UTs

I'll add UT with this case

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -0,0 +1,90 @@
...
+                return constructorDeclaration.Identifier;
+            }
+            else if (member is MethodDeclarationSyntax methodDeclaration)
+            {
+                return methodDeclaration.Identifier;
+            }
+            return null;
+        }
+
+        protected override void Initialize(SonarAnalysisContext context)
```
alban-auzeill: `ClassDeclarationSyntax`, `InterfaceDeclarationSyntax` and `StructDeclarationSyntax` inherit from `TypeDeclarationSyntax` that has a `Members` method. We could simplify with:
```
        {
            context.RegisterSyntaxNodeActionInNonGenerated(c => CheckMembers(c, ((TypeDeclarationSyntax)c.Node).Members) ,
                SyntaxKind.ClassDeclaration, SyntaxKind.InterfaceDeclaration, SyntaxKind.StructDeclaration);
        }
```

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace SonarAnalyzer.Rules
+{
+    public abstract class MethodOverloadsShouldBeGroupedBase<TMemberDeclarationSyntax> : SonarDiagnosticAnalyzer
+        where TMemberDeclarationSyntax : SyntaxNode
+    {
+        internal const string DiagnosticId = "S4136";
+        protected const string MessageFormat = "All '{0}' signatures should be adjacent";
```
alban-auzeill: I don't understand easily what the word `signatures` means. I would understand faster a message like:
```
All '{0}' method overloads should be adjacent.
```
Bonus: Don't forget the '.' at the end of each message, you win 10 `Ann Happy Points` with a low effort.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodOverloadsShouldBeGrouped.vb
```diff
@@ -0,0 +1,134 @@
...
+
+        Function tEst(ByVal a As Integer) As Integer
+
+        Event MyEvent2(ByVal Success As Boolean)
+    End Interface
+
+    Structure F
+        Public myField1 As String
+
+        Public Sub Test() ' Noncompliant {{All 'test' signatures should be adjacent}}
```
alban-auzeill: I would have tried to not use the lower case representation `test` of the method name in the message, but the first occurrence `Test`. For example, if the method name is `AMethodNameEasyToRead`, it's not nice to display `amethodnameeasytoread`.

PR #2661, opened 2019-09-18
Dump metrics

amelie-renard-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MLIRExporter.cs
```diff
@@ -97,48 +103,16 @@ private bool IsTooClomplexForMLIROrTheCFG(MethodDeclarationSyntax method)
             {
                 return true;
```
loic-joly-sonarsource: Shouldn't you call `exporterMetrics.AddUnsupportedFunction` here too?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MlirExporterMetrics.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        public void AddSupportedFunction()
+        {
+            ++nbSupportedFunctions;
+        }
+        public void AddUnsupportedFunction(SyntaxKind unsupportedSyntax)
+        {
+            ++nbUnsupportedFunctions;
+            if (!unsupportedSyntaxes.ContainsKey(unsupportedSyntax))
```
loic-joly-sonarsource: Why not code like that:

```
unsupportedSyntaxes[unsupportedSyntax] = unsupportedSyntax.tryGetValue(unsupportedSyntax, out int currentVal) ? currentVal + 1 : 1;
```
Or maybe even:
```
unsupportedSyntaxes.TryAdd(unsupportedSyntax, 0); // Nop if value already exists
++unsupportedSyntaxes[unsupportedSyntax]
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/MlirExporterMetrics.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                unsupportedSyntaxes.Add(unsupportedSyntax, 1);
+                return;
+            }
+            ++unsupportedSyntaxes[unsupportedSyntax];
+        }
+
+        public String Dump()
```
loic-joly-sonarsource: Could you add a unit test for that?

PR #2666, opened 2019-09-20
Rule S4583: Calls to delegate's method "BeginInvoke" should be paired with calls to "EndInvoke"

### LINKED ISSUE #1330: Rule S4583: "BeginInvoke" should be paired with "EndInvoke"
### Description

Using `BeginInvoke` without a closing `EndInvoke` can result in thread leaks. That's because when the invoked method throws an exception or returns a value, the thread that is used is never released back to the `ThreadPool`.

This rule raises an issue when a call to `BeginInvoke` is not paired with a call to `EndInvoke`.

Implements [RSPEC-4583](https://jira.sonarsource.com/browse/RSPEC-4583)

alban-auzeill:
Fixes #1330


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/BeginInvokePairedWithEndInvoke.cs
```diff
@@ -0,0 +1,230 @@
...
+            calledMethod = ReplaceNewDelegateByMethodDeclaration(calledMethod, semantic);
+            if (calledMethod != null && ParentDeclarationKinds.Contains(calledMethod.Kind()))
+            {
+                return GetEndInvokeList(calledMethod, semantic).Count > 0;
+            }
+
+            return true;
+        }
+
+        private static SyntaxNode ReplaceNewDelegateByMethodDeclaration(SyntaxNode node, SemanticModel semantic)
```
andrei-epure-sonarsource: The name is confusing - I expect a side effect from the name of the method. Also, please add a comment - which method declaration is this method looking for?

Also , would be clearer if instead of SyntaxNode, it would receive an ObjectCreationExpressionSyntax parameter

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/BeginInvokePairedWithEndInvoke.cs
```diff
@@ -0,0 +1,230 @@
...
+                    && variableDeclarator.Initializer is EqualsValueClauseSyntax equalsValueClause)
+                {
+                    return equalsValueClause.Value.RemoveParentheses();
+                }
+            }
+
+            return node;
+        }
+
+        private static bool IsNullLiteral(ExpressionSyntax callbackArg) => callbackArg.IsKind(SyntaxKind.NullLiteralExpression);
```
andrei-epure-sonarsource: I think this could be called on a `SyntaxNode`, and you could make it an extension method (see the `Helpers` namespace to add it to SyntaxNodeHelpers

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/BeginInvokePairedWithEndInvoke.cs
```diff
@@ -0,0 +1,230 @@
...
+        {
+            var callbackArgPos = invocationExpression.ArgumentList.Arguments.Count - 2;
+            var callbackArg = GetArgumentExpressionByNameOrPosition(invocationExpression, "callback", callbackArgPos)
+                ?.RemoveParentheses();
+            return callbackArg;
+        }
+
+        private static bool IsCallbackMayContainEndInvoke(SyntaxNode callbackArg, SemanticModel semantic)
+        {
+            var calledMethod = callbackArg.RemoveParentheses();
```
andrei-epure-sonarsource: At this point, you don't know it's a called method, `argument` would be more accurate

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/BeginInvokePairedWithEndInvoke.cs
```diff
@@ -0,0 +1,230 @@
...
+                    return symbol?.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax();
+                }
+            }
+
+            return node;
+        }
+
+        private static SyntaxNode ReplaceIdentifierByItsInitializer(SyntaxNode node, SemanticModel semantic)
+        {
+            if (node != null && node.IsKind(SyntaxKind.IdentifierName))
```
andrei-epure-sonarsource: You could move the `IsKind` check in the parent, and make this method just for IdentifierNameSyntax parameter instead of SyntaxNode - like this it would be clear from the signature what it should do

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/BeginInvokePairedWithEndInvoke.cs
```diff
@@ -0,0 +1,230 @@
...
+            SyntaxKind.ClassDeclaration,
+            SyntaxKind.CompilationUnit,
+            SyntaxKind.ConstructorDeclaration,
+            SyntaxKind.DelegateDeclaration,
+            SyntaxKind.DestructorDeclaration,
+            SyntaxKind.InterfaceDeclaration,
+            SyntaxKind.MethodDeclaration,
+            SyntaxKind.ParenthesizedLambdaExpression,
+            SyntaxKind.SimpleLambdaExpression,
+            SyntaxKind.StructDeclaration
```
andrei-epure-sonarsource: Did you add a test for StructDeclaration?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/BeginInvokePairedWithEndInvoke.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        private ExpressionSyntax GetCallbackArg(InvocationExpressionSyntax invocationExpression)
+        {
+            var callbackArgPos = invocationExpression.ArgumentList.Arguments.Count - 2;
+            var callbackArg = GetArgumentExpressionByNameOrPosition(invocationExpression, "callback", callbackArgPos)
+                ?.RemoveParentheses();
+            return callbackArg;
+        }
+
+        private static bool IsCallbackMayContainEndInvoke(SyntaxNode callbackArg, SemanticModel semantic)
```
andrei-epure-sonarsource: regarding the name, I think you can skip the `Is` because it's a bit unnatural - `CallbackContainsEndInvoke`  should suffice

sonaranalyzer-dotnet/src/SonarAnalyzer.Utilities/Rules.Description/S4583.html
```diff
@@ -0,0 +1,79 @@
+<p>Calling the <code>BeginInvoke</code> method of a delegate will allocate some resources that are only freed-up when <code>EndInvoke</code> is
+called. This is why you should always pair <code>BeginInvoke</code> with an <code>EndInvoke</code> to complete your asynchronous call.</p>
+<p>This rule raises an issue when:</p>
+<p> * the <code>BeginInvoke</code> method is called without any callback and it is not paired with a call to <code>EndInvoke</code> in the same
```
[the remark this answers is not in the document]
alban-auzeill: You have good eyes. :confused: it's a bug in rule-api that does not support the space character ' ' before the star '*'. [RULEAPI-520](https://jira.sonarsource.com/browse/RULEAPI-520)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/BeginInvokePairedWithEndInvoke.cs
alban-auzeill: I didn't find `SyntaxNodeHelpers` but `SonarAnalyzer.CSharp\Helpers\CSharpSyntaxHelper`. I hope it's what you wanted to say. I created an extension method.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/BeginInvokePairedWithEndInvoke.cs
alban-auzeill: Ok but I used `CallbackMayContainEndInvoke` and not not `CallbackContainsEndInvoke` because:
* false means we resolved callback code and didn't find `EndInvoke`
* true means we found `EndInvoke` or didn't resolve the callback code.

andrei-epure-sonarsource: ok, then it's worth adding these 2 points in the method comment

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/BeginInvokePairedWithEndInvoke.cs
```diff
@@ -96,48 +96,52 @@ private ExpressionSyntax GetCallbackArg(InvocationExpressionSyntax invocationExp
...
+                callback = LookupIdentifierInitializer((IdentifierNameSyntax)callback, semantic);
             }
 
-            return true;
-        }
-
-        private static SyntaxNode ReplaceNewDelegateByMethodDeclaration(SyntaxNode node, SemanticModel semantic)
-        {
-            if (node != null && node.IsKind(SyntaxKind.ObjectCreationExpression))
+            if (callback.IsKind(SyntaxKind.ObjectCreationExpression))
```
andrei-epure-sonarsource: should this be an `else if`? or are you checking the result of `LookupIdentifierInitializer`?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/BeginInvokePairedWithEndInvoke.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+        private static readonly ISet<SyntaxKind> ParentDeclarationKinds = new HashSet<SyntaxKind>
+        {
+            SyntaxKind.AnonymousMethodExpression,
+            SyntaxKind.ClassDeclaration,
+            SyntaxKind.CompilationUnit,
+            SyntaxKind.ConstructorDeclaration,
+            SyntaxKind.DelegateDeclaration,
```
andrei-epure-sonarsource: is DelegateDeclaration needed here anymore?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/BeginInvokePairedWithEndInvoke.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        private ExpressionSyntax GetCallbackArg(InvocationExpressionSyntax invocationExpression)
+        {
+            var callbackArgPos = invocationExpression.ArgumentList.Arguments.Count - 2;
+            var callbackArg = GetArgumentExpressionByNameOrPosition(invocationExpression, "callback", callbackArgPos)
+                ?.RemoveParentheses();
+            return callbackArg;
+        }
+
+        /// <returns>
```
andrei-epure-sonarsource: I would add a comment like
```
This method is looking for the callback code which can be:
- in a identifier initializer (like a lambda)
- passed directly as a lambda argument
- passed as a new delegate instantiation (and the code can be inside the method declaration)
- a mix of the above
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/BeginInvokePairedWithEndInvoke.cs
alban-auzeill: no

PR #2668, opened 2019-09-23
 Rule S5034: "ValueTask" should be consumed correctly

### LINKED ISSUE #2093: Rule S5034: "ValueTask" should be consumed correctly
### Description

There don't appear to be rules that cover the safe usage of `ValueTask<T>`. These rules have recently been documented on the [.NET Blog](https://blogs.msdn.microsoft.com/dotnet/2018/11/07/understanding-the-whys-whats-and-whens-of-valuetask/) (there may be other locations).

### Repro steps

See the linked blog post for details including code samples, but in summary code containing `ValueTask<T>` where the following operations are performed:

* Awaiting a `ValueTask` / `ValueTask<TResult>` more than once
* Awaiting a `ValueTask` / `ValueTask<TResult>` concurrently
* Using `.GetAwaiter().GetResult()` when the operation may not have completed

### Expected behavior

Warnings at appropriate levels.

### Actual behavior

No warnings.

### Known workarounds

None known.

### Related information

* SonarC# Version 7.8
* Visual Studio Version 2017
* If running through the Scanner for MSBuild, its version 4.4.2.1543

andrei-epure-sonarsource:
Fixes #2093


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConsumeValueTaskCorrectly.cs
```diff
@@ -0,0 +1,277 @@
...
+    class Tests
+    {
+        // see https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/
+        // and https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1?view=netstandard-2.1
+
+        public async void Foo1(ValueTaskProvider stream)
+        {
+            var valueTask = stream.ReadAsync();
+            var once = await valueTask;
+//                           ^^^^^^^^^ Noncompliant
```
andrei-epure-sonarsource: add the expected text

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConsumeValueTaskCorrectly.cs
```diff
@@ -0,0 +1,278 @@
...
+
+    class Tests
+    {
+        // see https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/
+        // and https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1?view=netstandard-2.1
+
+        public async void Foo1(ValueTaskProvider stream)
+        {
+            var valueTask = stream.ReadAsync();
+            var once = await valueTask; // Noncompliant { "Refactor this 'ValueTask' usage to consume it only once." }
```
andrei-epure-sonarsource: the syntax for this should actually be `{{Refactor this 'ValueTask' usage to consume it only once.}}
` , so this doesn't check the message...

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConsumeValueTaskCorrectly.cs
```diff
@@ -0,0 +1,201 @@
...
+            }
+
+            /**
+             * Check if it's the wanted method on a ValueTask
+             * - we treat AsTask() like await - always add it to the list
+             * - for GetAwaiter().GetResult() - ignore the call if it's called inside an 'if (valueTask.IsCompletedSuccessfully)'
+             */
+            public override void VisitInvocationExpression(InvocationExpressionSyntax node)
+            {
+                if (node.Expression is MemberAccessExpressionSyntax memberAccess)
```
alban-auzeill: it will be easier to read if you use helper methods to match method signature `AsTask` and `GetResult` like `SonarAnalyzer.Helpers.CSharpSyntaxHelper.IsMethodInvocation` (and add another helper method for the `Result` property)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConsumeValueTaskCorrectly.cs
```diff
@@ -0,0 +1,225 @@
...
+            }
+
+            /**
+             * Check if ".Result" is accessed on a 'ValueTask'
+             * - ignore the call if it's called inside an 'if (valueTask.IsCompletedSuccessfully)'
+             */
+            public override void VisitMemberAccessExpression(MemberAccessExpressionSyntax node)
+            {
+                if (node.IsPropertyInvocation(ValueTaskTypes, "Result", semanticModel) &&
+                    GetLeftMostIdentifier(node) is IdentifierNameSyntax identifierName &&
```
andrei-epure-sonarsource: I need to get the identifier so that I can get the symbol of the variable, and keep a list of its usages (to see if it's used twice).

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -256,10 +256,23 @@ public static bool IsStringEmpty(this ExpressionSyntax expression, SemanticModel
         }
 
         public static bool IsMethodInvocation(this InvocationExpressionSyntax expression, KnownType type, string methodName, SemanticModel semanticModel) =>
+            expression.NameIs(methodName) &&
             semanticModel.GetSymbolInfo(expression).Symbol is IMethodSymbol methodSymbol &&
             methodSymbol.IsInType(type) &&
             methodName.Equals(methodSymbol.Name, StringComparison.InvariantCulture);
 
+        public static bool IsMethodInvocation(this InvocationExpressionSyntax expression, ImmutableArray<KnownType> types, string methodName, SemanticModel semanticModel) =>
```
alban-auzeill: A tiny maintenance improvement could be to have only one `IsMethodInvocation` implementation with a variable number of `types` (if there no performance drawback)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -256,10 +256,23 @@ public static bool IsStringEmpty(this ExpressionSyntax expression, SemanticModel
...
+            expression.NameIs(methodName) &&
+            semanticModel.GetSymbolInfo(expression).Symbol is IMethodSymbol methodSymbol &&
+            methodSymbol.IsInType(types) &&
+            methodName.Equals(methodSymbol.Name, StringComparison.InvariantCulture);
+
+        public static bool IsPropertyInvocation(this MemberAccessExpressionSyntax expression, ImmutableArray<KnownType> types, string propertyName, SemanticModel semanticModel) =>
+            expression.NameIs(propertyName) &&
+            semanticModel.GetSymbolInfo(expression).Symbol is IPropertySymbol propertySymbol &&
+            propertySymbol.IsInType(types) &&
+            propertyName.Equals(propertySymbol.Name, StringComparison.InvariantCulture);
```
alban-auzeill: like the above comment, the latest `Equals` seems to be useless.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConsumeValueTaskCorrectly.cs
```diff
@@ -0,0 +1,225 @@
...
+    {
+        internal const string DiagnosticId = "S5034";
+
+        // 'await', 'AsTask', 'Result' and '.GetAwaiter().GetResult()' should be called only once on a ValueTask
+        private const string ConsumeOnlyOnceMessage = "Refactor this 'ValueTask' usage to consume it only once.";
+
+        private static readonly DiagnosticDescriptor ConsumeOnlyOnce =
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, ConsumeOnlyOnceMessage, RspecStrings.ResourceManager);
+
+        // This should be called only when 'readTask.IsCompletedSuccessfully' is not called before
```
alban-auzeill: I don't like the statement `is not called before` in the above comment.
It makes me believe that the following case is supported:
```cs
if (!valueTask.IsCompletedSuccessfully)
{
  return weNeedToAwait(valueTask);
}
return valueTask.Result;
```
Is the above case supported or is it a FP?
I think statements used in other comments better represent the logic `called inside an 'if (valueTask.IsCompletedSuccessfully)'`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
andrei-epure-sonarsource: indeed the recommendation (e.g. [article](https://www.dotnetperls.com/params)) is to have overloads for small number of parameters
- this is what [String.Concat](https://referencesource.microsoft.com/#mscorlib/system/string.cs,3045) does

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -367,7 +365,7 @@ public static bool NameIs(this InvocationExpressionSyntax invocation, string nam
             {
                 invocationName = identifierName.Identifier.ValueText;
             }
-            return invocationName.Equals(invocationName, StringComparison.InvariantCulture);
+            return invocationName.Equals(name, StringComparison.InvariantCulture);
```
andrei-epure-sonarsource: yes this was a hidden bug

alban-auzeill: :scream:

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
andrei-epure-sonarsource: damn, it seems I missed this one. I'll clean up soon

PR #2669, opened 2019-09-24
Fix boom due to release of dotnet core 3.0

andrei-epure-sonarsource:
Yesterday dotnet core 3.0 got released.
So some APIs got removed, and tests started failing because we were referencing NuGetLatestVersion.


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/ConfiguringLoggersTest.cs
```diff
@@ -31,6 +31,8 @@ namespace SonarAnalyzer.UnitTest.Rules
     [TestClass]
     public class ConfiguringLoggersTest
     {
+        private const string DOT_NET_CORE_VERSION = "2.2.0";
```
alban-auzeill: I would prefer having the version of .NET Core declared only at one location in `Constants`

Issue #2675, opened 2019-09-27
C# 7 tuples in foreach makes CFG creation fail


alban-auzeill:
### Description

CSharpControlFlowGraphBuilder fails to create the CFG of tuples in `foreach`.

### Repro steps

Create the CFG of the following `Bar` method:
```cs
namespace NS
{
  public class Foo
  {
    public void Bar((string key, string value)[] bars)
    {
      foreach (var (key, value) in bars) { }
    }
  }
}
```

### Expected behavior

Should not throw an exception.

### Actual behavior

```
System.NotSupportedException: ForEachVariableStatement
  at SonarAnalyzer.ControlFlowGraph.CSharp.CSharpControlFlowGraphBuilder.BuildStatement(StatementSyntax statement, Block currentBlock) in sonaranalyzer-dotnet\src\SonarAnalyzer.CSharp\ControlFlowGraph\CSharpControlFlowGraphBuilder.cs:line 209
  at SonarAnalyzer.ControlFlowGraph.CSharp.CSharpControlFlowGraphBuilder.BuildStatements(IEnumerable`1 statements, Block currentBlock) in sonaranalyzer-dotnet\src\SonarAnalyzer.CSharp\ControlFlowGraph\CSharpControlFlowGraphBuilder.cs:line 486
  at SonarAnalyzer.ControlFlowGraph.CSharp.CSharpControlFlowGraphBuilder.BuildBlock(BlockSyntax block, Block currentBlock) in sonaranalyzer-dotnet\src\SonarAnalyzer.CSharp\ControlFlowGraph\CSharpControlFlowGraphBuilder.cs:line 975
  at SonarAnalyzer.ControlFlowGraph.CSharp.CSharpControlFlowGraphBuilder.BuildStatement(StatementSyntax statement, Block currentBlock) in sonaranalyzer-dotnet\src\SonarAnalyzer.CSharp\ControlFlowGraph\CSharpControlFlowGraphBuilder.cs:line 126
  at SonarAnalyzer.ControlFlowGraph.CSharp.CSharpControlFlowGraphBuilder.BuildStatements(IEnumerable`1 statements, Block currentBlock) in sonaranalyzer-dotnet\src\SonarAnalyzer.CSharp\ControlFlowGraph\CSharpControlFlowGraphBuilder.cs:line 486
[...]

PR #2676, opened 2019-09-27
[Do not merge] We should allow CFG build exceptions to find out what syntax we do not support

### LINKED ISSUE #2541: We should allow CFG build exceptions to find out what syntax we do not support
Currently in [CSharpControlFlowGraph.cs ](https://github.com/SonarSource/sonar-dotnet/blob/7.15.0.8572/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraph.cs#L53) we are eating the exceptions we throw during the Control Flow Graph build process.

This means we have hidden errors or syntax we don't support that we may not be aware of.

We should log the stacktrace (on a single line) and enable these failures to catch them during our QA steps.

SonarCsharp version 7.15

alban-auzeill:
Fixes #2541

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraph.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/ControlFlowGraph/CSharpControlFlowGraph.cs
@@ -19,17 +19,42 @@
  */
 
 using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Diagnostics;
+using System.Text;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
+using SonarAnalyzer.Helpers;
 
 namespace SonarAnalyzer.ControlFlowGraph.CSharp
 {
     public static class CSharpControlFlowGraph
     {
+        private static readonly ISet<string> UNSUPPORTED_SYNTAX_NODE_KINDS = new HashSet<string>
+        {
+           "DeclarationExpression",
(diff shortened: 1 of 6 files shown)


alban-auzeill:
This PR helps us highlighting the syntax nodes that are not fully supported by our CFG builder.
The result is:
* DeclarationExpression
* DefaultLiteralExpression
* LocalFunctionStatement
* ForEachVariableStatement
* ThrowExpression
* TupleExpression

But we will not merge it onto master, because:
* The PR is not complete, we need to raise issues only in a `Hard Fail` mode defined by property, like the property `"sonar.security.hardfail"` in sonar-security.
* There's a workaround to avoid throwing exceptions for the above unsupported syntax nodes list. Without this workaround, ITs fails. Removing this workaround means fixing the CFG implementation, and we don't want to fix it in `sonar-dotnet`. Currently, the most up-to-date C# CFG implementation is in `sonar-security`, it's the one that will be extracted as a shared library and later used by `sonar-dotnet`.

alban-auzeill:
ForEachVariableStatement should be fixed by #2675

alban-auzeill:
DefaultLiteralExpression should be fixed by #2678

alban-auzeill:
LocalFunctionStatement should be fixed by #2679

alban-auzeill:
ThrowExpression should be fixed by #2680

alban-auzeill:
TupleExpression should be fixed by #2681

andrei-epure-sonarsource:
currently  #2541 is paused, we can start over from this branch later on

the idea is that we need to factor out the CFG from sonar-security and used it here, so until now it's not worth spending time fixing the issues

PR #2696, opened 2019-10-03
Rule S2479: Whitespace and control characters in string literals should be explicit

### LINKED ISSUE #2685: Rule S2479: Whitespace and control characters in string literals should be explicit
### Description

Rule spec: https://jira.sonarsource.com/browse/RSPEC-2479
C# snippet: https://jira.sonarsource.com/browse/RSPEC-5476

alban-auzeill:
Fixes #2685


andrei-epure-sonarsource:
For the test file
- [line 196](https://github.com/SonarSource/sonar-dotnet/blob/5a5846c29820e8016c1a9262a1691727c47be649/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ControlCharacterInString.cs#L196) - could you also specify in the comment what character is in that string (to make it easy to read that it's not an empty string there)
- could you add a test with a multi-line verbatim string?
- could you add 3 tests with a "hello" string with control characters at the beginning, the middle and the end of the normal string - and what position the issue gets raised?
- could you add a test with string concatenation?
- did you add compliant tests with characters in the range from \u0080 to \u009F , and also write in a comment in the UT like the comment inside the RSPEC to let the dev that will read/break them know easily why we left them out?

PR #2697, opened 2019-10-07
A "for" loop update clause should move the counter in the right…

### LINKED ISSUE #2683: Rule S2251 A "for" loop update clause should move the counter in the right direction
### Description
Rules description: https://jira.sonarsource.com/browse/RSPEC-2251
C# details https://jira.sonarsource.com/browse/RSPEC-5475

costin-zaharia-sonarsource:
Fix #2683


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSignCheck.cs
```diff
@@ -34,25 +34,206 @@ namespace SonarAnalyzer.Rules.CSharp
...
+
+                    if (incrementorData.Operation == ArithmeticOperation.Addition &&
+                        condition == Condition.Greater)
+                    {
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, forNode.Condition.GetLocation(), incrementorData.IdentifierName, "inc"));
+                    }
+                    else if (incrementorData.Operation == ArithmeticOperation.Substraction &&
+                             condition == Condition.Less)
+                    {
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, forNode.Condition.GetLocation(), incrementorData.IdentifierName, "dec"));
```
andrei-epure-sonarsource: in [sonar-php](https://github.com/SonarSource/sonar-php/blob/3875b7574ba3be9f2fb6ce21be815547b6f2642a/php-checks/src/main/java/org/sonar/php/checks/ForLoopIncrementSignCheck.java) there is a secondary location labeled "stop condition" - by adding the primary location as the `indexer` and the secondary location as the `condition`, the message will make more sense

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSignCheck.cs
```diff
@@ -34,25 +34,206 @@ namespace SonarAnalyzer.Rules.CSharp
...
+                    if (GetOperandFromUnaryExpression(incrementor) is IdentifierNameSyntax operand)
+                    {
+                        identifierName = operand.Identifier.ValueText;
+                        opp =  incrementorToOperation.GetValueOrDefault(incrementorKind, ArithmeticOperation.None);
+                    }
+                    break;
+
+                case SyntaxKind.SubtractAssignmentExpression:
+                case SyntaxKind.AddAssignmentExpression:
+                    if (incrementor is AssignmentExpressionSyntax add && add.Left is IdentifierNameSyntax leftId &&
```
andrei-epure-sonarsource: nitpick: put the `add.Left is IdentifierNameSyntax leftId` on a new line (it's easier to follow if one line has one condition)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSignCheck.cs
```diff
@@ -34,25 +34,206 @@ namespace SonarAnalyzer.Rules.CSharp
...
+                    break;
+
+                default:
+                    break;
+            }
+
+            return new IncrementData(identifierName, opp);
+        }
+
+        private Condition GetCondition(BinaryExpressionSyntax conditionSyntax, string identifierName)
```
andrei-epure-sonarsource: it would be useful to add a comment specifying that this method treats conditions on both sides of the binary expression (e.g. `x < 1` and `1 > x` are the same thing, and this considers these cases)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSignCheck.cs
```diff
@@ -34,25 +34,206 @@ namespace SonarAnalyzer.Rules.CSharp
...
+                {
+                    return Condition.Less;
+                }
+            }
+
+            return Condition.None;
+        }
+
+        private bool IsVariableAndLiteralBinaryExpression(BinaryExpressionSyntax binaryExpression, string identifier) =>
+            (binaryExpression.Left is IdentifierNameSyntax leftId && leftId.NameIs(identifier) && binaryExpression.Right.IsKind(SyntaxKind.NumericLiteralExpression)) ||
```
andrei-epure-sonarsource: having a simple method `IsIdentifier(SyntaxNode x, string identifier)` would make this method more readable

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ForLoopIncrementSignCheck.cs
```diff
@@ -6,13 +6,23 @@ public class ForLoopIncrementSignCheck
     {
         public void TestMethod(int x, int y, int z)
```
andrei-epure-sonarsource: I would also add 
- a `for` with 2 iterators
- a `for` with 2 initializers

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -367,6 +367,9 @@ public static bool NameIs(this InvocationExpressionSyntax invocation, string nam
             return invocationName.Equals(name, StringComparison.InvariantCulture);
         }
 
+        public static bool NameIs(this IdentifierNameSyntax identifierName, string name)
+            => identifierName?.Identifier.ValueText == name;
```
alban-auzeill: Could you validate with @andrei-epure-sonarsource that using `==` is ok here. Because I see in the previous method that we use `xxx.Equals(xxx, StringComparison.InvariantCulture)` to compare `IdentifierNameSyntax` name. So I don't understand which is the best.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSignCheck.cs
costin-zaharia-sonarsource: Done

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSignCheck.cs
costin-zaharia-sonarsource: Done

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSignCheck.cs
costin-zaharia-sonarsource: Done

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSignCheck.cs
```diff
@@ -0,0 +1,243 @@
+﻿/*
```
alban-auzeill: In all our analyzer we end the rule implementation name with "Check". But I see that `sonar-dotnet` repository is not using this convention.  There's only 1 other file out of 408 in this directory that has its name ending with "Check.cs". Can you validate with @andrei-epure-sonarsource that we want to use this naming convention here?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSignCheck.cs
costin-zaharia-sonarsource: Done. Simplified a bit also **GetCondition** with this.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSignCheck.cs
```diff
@@ -34,25 +34,206 @@ namespace SonarAnalyzer.Rules.CSharp
...
+            (binaryExpression.Left.IsKind(SyntaxKind.NumericLiteralExpression) && binaryExpression.Right is IdentifierNameSyntax rightId && rightId.NameIs(identifier));
+
+        private IdentifierNameSyntax GetOperandFromUnaryExpression(ExpressionSyntax syntax)
+        {
+            switch (syntax.Kind())
+            {
+                case SyntaxKind.PreIncrementExpression:
+                case SyntaxKind.PreDecrementExpression:
+                    var prefix = (PrefixUnaryExpressionSyntax)syntax;
+                    if (prefix.Operand.IsKind(SyntaxKind.IdentifierName))
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Done.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ForLoopIncrementSignCheck.cs
costin-zaharia-sonarsource: Done.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSignCheck.cs
andrei-epure-sonarsource: good catch! it should be `ForLoopIncrementSignTest`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
andrei-epure-sonarsource: `ValueText` is a `string`, just like `name`, so in this case `==` should be fine because [operator ==](https://referencesource.microsoft.com/#mscorlib/system/string.cs,705) is just a syntactic sugar for `String.Equals`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSignCheck.cs
costin-zaharia-sonarsource: Done.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSignCheck.cs
andrei-epure-sonarsource: nice 😄

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ForLoopIncrementSign.cs
```diff
@@ -0,0 +1,243 @@
...
+            identifier.NameIs(identifierName);
+
+        private IdentifierNameSyntax GetIdentifierFromUnaryExpression(ExpressionSyntax syntax)
+        {
+            switch (syntax.Kind())
+            {
+                case SyntaxKind.PreIncrementExpression:
+                case SyntaxKind.PreDecrementExpression:
+                    var prefix = (PrefixUnaryExpressionSyntax)syntax;
+                    return (IdentifierNameSyntax)prefix.Operand;
```
andrei-epure-sonarsource: you should add a test with `for(;; this.number++)`
I think it will throw because you are forcing a cast here and `MemberAccess` is not an `IdentifierName` (I think)

if you use the [`as` operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast#as-operator) instead of the [Cast operator ()](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast#cast-operator-), `null` will be returned if the cast fails. And as you null-check in the caller, all will be fine

costin-zaharia-sonarsource: Indeed it fails in this case. I've added the test and updated the code.

PR #2699, opened 2019-10-08
Update the expected path for AnalyzeGeneratedVb

andrei-epure-sonarsource:
This VB file was forgotten to be updated - the equivalent CS file has 'replaced' in URI

(sonaranalyzer-dotnet\its\expected\AnalyzeGenerated\AnalyzeGeneratedFiles-6389f4a4-59fd-497b-bb52-4414b1320d29-S1451.json)


costin-zaharia-sonarsource:
The change looks good to me but it seems that we need to update the expected results for VB tests. (ERROR: There are differences between the actual and the expected issues.)

andrei-epure-sonarsource:
done in #2832

PR #2700, opened 2019-10-08
Add a manually non compliant issue for S2251

costin-zaharia-sonarsource:
- add file headers to get rid of the warning
- move assembly attributes in own file


andrei-epure-sonarsource:
This is related to #2683

alban-auzeill:
LGTM, one issue per related file, that's perfect :sparkles:

PR #2702, opened 2019-10-09
Rule S4635: String offset-based methods should be preferred

### LINKED ISSUE #2687: Rule S4635: String offset-based methods should be preferred for finding substrings from offsets
### Description

Rule spec: https://jira.sonarsource.com/browse/RSPEC-4635
C# details: https://jira.sonarsource.com/browse/RSPEC-4638

costin-zaharia-sonarsource:
Fix #2687


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StringOffsetMethods.cs
```diff
@@ -0,0 +1,78 @@
...
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public sealed class StringOffsetMethods : SonarDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S4635";
+        private const string MessageFormat = "A new string is going to be created by 'Substring'.";
```
andrei-epure-sonarsource: the message in the rspec is `Replace "XXX" with the overload that accepts an offset parameter.` , please update it :)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StringOffsetMethods.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+
+namespace Tests.Diagnostics
+{
+    public class StringOffsetMethods
+    {
+        public void StringOffsetMethodsCases()
```
andrei-epure-sonarsource: please also add tests with

- a variable of type string `x.Substring(1).IndexOf('t')`
- using `x.Substring(int startIndex, int length);`
- using other overloads of IndexOf chained to the `Substring` call
- using overloads with `StringComparison` as well

also, I would add some Noncompliant tests where this gets triggered inside a Constructor, a Getter and a lambda

also, I would add a test like
`x.Substring(1).Remove(1).IndexOf('t')`
and
`x.Remove(1).IndexOf('t')`

to document the behavior in these cases

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StringOffsetMethods.cs
```diff
@@ -4,41 +4,76 @@ namespace Tests.Diagnostics
...
     {
-        public void StringOffsetMethodsCases()
+        public StringOffsetMethods()
+        {
+            "Test".Substring(1).IndexOf('t'); // Noncompliant {{Replace 'IndexOf' with the overload that accepts an offset parameter.}}
+//          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+        }
+
+        public int GetIndex =>
+            "Test".Substring(1).IndexOf('t'); // Noncompliant {{Replace 'IndexOf' with the overload that accepts an offset parameter.}}
```
andrei-epure-sonarsource: as discussed, you can have only one or two tests for the message and issue highlighting, and for the rest test simply with `// Noncompliant`

PR #2707, opened 2019-10-09
Fix S2699 FP: NUnit 3.0 TestCase + ExpectedResult

### LINKED ISSUE #2451: S2699 false positive after upgrade
On version 7.9.1.7622 this is ok:

```
using System;
using NUnit.Framework;

namespace ByValue
{
    [TestFixture]
    public class TypeExtensionsTests
    {
        [TestCase(typeof(string), ExpectedResult = "System.String")]
        [TestCase(typeof(string[]), ExpectedResult = "System.String[]")]
        [TestCase(typeof(List<string>), ExpectedResult = "List<System.String>")]
        [TestCase(typeof(List<(string, int)>), ExpectedResult = "List<ValueTuple<System.String,System.Int32>>")]
        public string GetFriendlyFullName(Type type)
        {
            return type.GetFriendlyFullName();
        }
    }
}
```

after upgrade to 7.14.0.8411, I got S2699 

* Visual Studio 2019

andrei-epure-sonarsource:
Also upgrade UT to run with NUnit 3.11 as baseline

Fix  #2451

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
@@ -111,7 +111,12 @@ namespace SonarAnalyzer.Helpers
             method.GetAttributes().Any(a =>
                 a.AttributeClass.Is(KnownType.NUnit_Framework_TestCaseAttribute) &&
                 a.NamedArguments.Any(arg => arg.Key == "ExpectedResult"));
+public static bool HasAssertionInAttribute(this IMethodSymbol method) =>
+    method.GetAttributes().Any(IsTestCaseAttributeWithExpectedResult);
 
+private static bool IsTestCaseAttributeWithExpectedResult(AttributeData a) =>
+    a.AttributeClass.Is(KnownType.NUnit_Framework_TestCaseAttribute) &&
+    a.NamedArguments.Any(arg => arg.Key == "ExpectedResult");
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
```diff
@@ -107,6 +107,11 @@ internal static class UnitTestHelper
         public static bool HasExpectedExceptionAttribute(this IMethodSymbol method) =>
             method.GetAttributes().Any(a => a.AttributeClass.IsAny(KnownExpectedExceptionAttributes));
 
+        public static bool HasAssertionInAttribute(this IMethodSymbol method) =>
+            method.GetAttributes().Any(a =>
+                a.AttributeClass.Is(KnownType.NUnit_Framework_TestCaseAttribute) &&
+                a.NamedArguments.Any(arg => arg.Key == "ExpectedResult"));
+
```
costin-zaharia-sonarsource: We can extract a small method here to improve a bit readability:
```suggestion
public static bool HasAssertionInAttribute(this IMethodSymbol method) =>
    method.GetAttributes().Any(IsTestCaseAttributeWithExpectedResult);

private static bool IsTestCaseAttributeWithExpectedResult(AttributeData a) =>
    a.AttributeClass.Is(KnownType.NUnit_Framework_TestCaseAttribute) &&
    a.NamedArguments.Any(arg => arg.Key == "ExpectedResult");
```

PR #2712, opened 2019-10-10
Update S2699: to support NSubstitute 4.2.1

### LINKED ISSUE #2709: Update S2699: to support NSubstitute 4.2.1
### Description

Extend https://jira.sonarsource.com/browse/RSPEC-3413 by adding support for [NSubstitute](https://nsubstitute.github.io/).

The latest stable version at the moment is **4.2.1**.

This can be done by checking that at least one of the following assert methods is called in a test:
```
NSubstitute.Received.InOrder()
NSubstitute.SubstituteExtensions.DidNotReceive()
NSubstitute.SubstituteExtensions.DidNotReceiveWithAnyArgs()
NSubstitute.SubstituteExtensions.Received()
NSubstitute.SubstituteExtensions.ReceivedWithAnyArgs()
```

costin-zaharia-sonarsource:
The checks for the NSubstitute assertions are done using the actual method symbol instead of searching for keywords like it's done for the other assertions. This allows for more granular and precise matching.

Fixes #2709

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TestMethodShouldContainAssertionTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TestMethodShouldContainAssertionTest.cs
@@ -56,6 +56,7 @@ namespace SonarAnalyzer.UnitTest.Rules
                 new TestMethodShouldContainAssertion(),
                 additionalReferences: NuGetMetadataReference.NUnit(testFwkVersion)
                     .Concat(NuGetMetadataReference.FluentAssertions(fluentVersion))
+                    .Concat(NuGetMetadataReference.NSubstitute("4.2.1"))
                     .Concat(FrameworkMetadataReference.SystemReflection)
                     .Concat(FrameworkMetadataReference.SystemXmlXDocument)
                     .Concat(FrameworkMetadataReference.SystemXmlLinq)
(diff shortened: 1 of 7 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TestMethodShouldContainAssertionTest.cs
```diff
@@ -56,6 +56,7 @@ public void TestMethodShouldContainAssertion_NUnit(string testFwkVersion, string
                 new TestMethodShouldContainAssertion(),
                 additionalReferences: NuGetMetadataReference.NUnit(testFwkVersion)
                     .Concat(NuGetMetadataReference.FluentAssertions(fluentVersion))
+                    .Concat(NuGetMetadataReference.NSubstitute("4.2.1"))
```
andrei-epure-sonarsource: maybe put the version together with `testFwkVersion` and `fluentVersion`?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.NUnit.cs
```diff
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using FluentAssertions;
+    using NSubstitute;
```
andrei-epure-sonarsource: it would make sense to add a simple test for `xunit` or other libraries that can be used together with `nsubstitute`, something like 

```
        [TestCase]
        public void ReceivedNameSpace() => NSubstitute.SubstituteExtensions.Received(calculator).Add(1, 2);

```
in each, just to make sure that we don't raise on other libraries then nunit

and then this file will still have extensive tests for nsubstitute specifically (and you can add a comment in the xunit tests that the extensive tests are in this file - so that people know to modify only these in case of a change of nsubstitute API which will need more extensive tests

PR #2715, opened 2019-10-11
Update S2699: add support for the latest version of FluentAsser…

### LINKED ISSUE #2714: Update S2699: add support for the latest version of FluentAssertions (5.9.0)
### Description

Update TestMethodShouldContainAssertion and tests to support FluentAssertions 5.

- ShouldThrow() and ShouldNotThrow() are replaced by Should().Throw() and Should().NotThrow()

costin-zaharia-sonarsource:
- extract constants
- add test for legacy versions
- remove the tests which are using the old assertions (ShouldThrow, ShouldNotThrow)

- fix #2714


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TestMethodShouldContainAssertionTest.cs
```diff
@@ -58,22 +79,19 @@ public void TestMethodShouldContainAssertion_NUnit(string testFwkVersion, string
...
         }
 
         [DataTestMethod]
-        [DataRow("2.5.7.10213", "4.19.4")]
-        [DataRow("2.7.0", "4.19.4")]
+        [DataRow(NUnitVersions.Ver25)]
+        [DataRow(NUnitVersions.Ver27)]
         [TestCategory("Rule")]
-        public void TestMethodShouldContainAssertion_NUnit_V2Specific(string testFwkVersion, string fluentVersion)
+        public void TestMethodShouldContainAssertion_NUnit_V2Specific(string testFwkVersion)
```
andrei-epure-sonarsource: from talk with Costin - `fluentVersion` wasn't being used in the test

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TestMethodShouldContainAssertionTest.cs
```diff
@@ -142,8 +155,51 @@ public void TestMethodShouldContainAssertion_Xunit_Legacy()
...
                 additionalReferences: NuGetMetadataReference.XunitFrameworkV1
-                    .Concat(NuGetMetadataReference.FluentAssertions("1.6.0"))
+                    .Concat(NuGetMetadataReference.FluentAssertions(Constants.NuGetLatestVersion))
+                    .Concat(FrameworkMetadataReference.SystemXml)
+                    .Concat(FrameworkMetadataReference.SystemXmlLinq)
                     .ToArray());
         }
+
+        [DataTestMethod]
+        [DataRow(NUnitVersions.Ver25, FluentAssertionVersions.Ver1)]
```
andrei-epure-sonarsource: curiosity (no action needed) : was there a reason not to combine other `NUnitVersions` in this test?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.MsTest.cs
```diff
@@ -43,20 +43,6 @@ public void TestMethod5()
             x.Should().Be(42);
         }
 
-        [TestMethod]
```
andrei-epure-sonarsource: this was moved in `TestMethodShouldContainAssertion_NUnit_FluentAssertionsLegacy`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.Xunit.Legacy.cs
```diff
@@ -1,231 +1,201 @@
-﻿namespace TestXunitLegacy
```
andrei-epure-sonarsource: should try to avoid this confusing diff

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TestMethodShouldContainAssertionTest.cs
costin-zaharia-sonarsource: The main reason is that the nunit version should not be relevant for this test because what we want to verify is that we support old versions of FluentAssertions.

A second reason is that by using the latest version of NUnit I get an error message: "Issue with message 'CS0012: The type 'Attribute' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard...."

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.Xunit.Legacy.cs
costin-zaharia-sonarsource: Yes :). Should be ok now

PR #2727, opened 2019-10-15
Fix broken icon URL in nuspec files

### LINKED ISSUE #2370: Replace expired logo link with github static resource
### Description

The `nuspec` files (SonarAnalyzer.CSharp.nuspec, SonarAnalyzer.VisualBasic.nuspec) point to
```
    <iconUrl>http://dist.sonarsource.com/csharp/download/ico-sonarlint-64.png</iconUrl>
```
That's a 404 and because of that the SL icon isn't shown on the [nuget page](https://www.nuget.org/packages/SonarAnalyzer.CSharp/) and inside the Visual Studio Marketplace.

Instead, we get the default 
image

### Expected behavior

Should put the needed resource (sonarlint logo) here https://github.com/SonarSource/analyzers-github-static-resources as a static github pages website, and reference them from there. Github raw functionality will be deprecated, so use github pages.

### Related information

* SonarC# Version  7.12

andrei-epure-sonarsource:
fix  #2370

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Vsix/SonarAnalyzer.CSharp.nuspec
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Vsix/SonarAnalyzer.CSharp.nuspec
@@ -8,7 +8,7 @@
     <owners>SonarSource</owners>
     <license type="file">License.txt</license>
     <projectUrl>http://vs.sonarlint.org</projectUrl>
-    <icon>https://sonarsource.github.io/analyzers-github-static-resources/sonar-dotnet/SonarLint-icon-64px.png</icon>
+    <iconUrl>https://sonarsource.github.io/analyzers-github-static-resources/sonar-dotnet/SonarLint-icon-64px.png</iconUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <summary>Analyzers which spot bugs and code smells in your code. This package is best used together with SonarLint for Visual Studio and/or the SonarQube platform.</summary>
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Vsix/SonarAnalyzer.CSharp.nuspec
```diff
@@ -8,7 +8,7 @@
     <owners>SonarSource</owners>
     <license type="file">License.txt</license>
     <projectUrl>http://vs.sonarlint.org</projectUrl>
-    <iconUrl>http://dist.sonarsource.com/csharp/download/ico-sonarlint-64.png</iconUrl>
+    <icon>https://sonarsource.github.io/analyzers-github-static-resources/sonar-dotnet/SonarLint-icon-64px.png</icon>
```
christophe-zurn-sonarsource: @andrei-epure-sonarsource In order to use the `<icon>` element, we need to have at least version 5.3 of NuGet (See https://github.com/NuGet/Home/issues/8660).
I suggest we revert the update of `<iconUrl>` to `<icon>` and create a ticket for updating NuGet version and `iconUrl` element. WDYT ?

christophe-zurn-sonarsource: @andrei-epure-sonarsource As discussed and agrred on, we will keep `<iconUrl>` element until we update NuGet version on our build pipeline. I will update the PR accordingly

andrei-epure-sonarsource: thanks! 👍

Issue #2426, opened 2019-05-15
`VarPattern` not supported in the analyzer (when building with MSBuild 16)


YairHalberstadt:
### Description

Error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.EmptyNullableValueAccess' threw an exception of type 'System.NotSupportedException' with message 'VarPattern'.

### Repro steps

Compile the following with SonarAnalyzer.CSharp nuget package installed:

```csharp
using System;
public class C {
    public int M<T>() {
		switch (typeof(T))
		{
			case var x when x == typeof(Foo): return 1;
			case var x when x == typeof(Bar): return 2;
			default: throw new InvalidOperationException();
		}
    }
}

public class Foo {}
public class Bar {}
```

* SonarC# Version: 7.14.0.8411
* Visual Studio Version: VS 2019 16.0.3

andrei-epure-sonarsource:
Hi @YairHalberstadt . I'm trying to reproduce the issue but cannot. Could you please tell me what version of the .NET framework you are using?

RamanBut-Husaim:
@andrei-epure-sonarsource Hello.

I've attached the project that fails on my local machine VS 2019 16.0.3 installed.

More precisely, the build fails on MSBuild 16. In my case it is situated in `C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin\MSBuild.exe`.

MSBuild 15+ works with no issues.

The list of errors inside the repro solution on my local machine:
```
  CSC: error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.ConditionEvaluatesToConstant' threw an exception of type 'System.NotSupportedException' with message 'VarPattern'.
  CSC: error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.EmptyCollectionsShouldNotBeEnumerated' threw an exception of type 'System.NotSupportedException' with message 'VarPattern'.
  CSC: error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.EmptyNullableValueAccess' threw an exception of type 'System.NotSupportedException' with message 'VarPattern'.
  CSC: error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.InvalidCastToInterface' threw an exception of type 'System.NotSupportedException' with message 'VarPattern'.
[...]

PR #2729, opened 2019-10-16
Add support for VarPattern

costin-zaharia-sonarsource:
- needed when running the analyzer with MSBuild 16

Fix: #2426


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
```diff
@@ -758,5 +758,20 @@ public int Start()
         }
 
         public void Archive() {}
+
+        public class TestClass
```
duncanp-sonar: This new test class is nested inside a test class for a different bug (`public class FooBarBaz`), which is confusing.
It should be top-level class, and it would be useful to have a link to the bug (same as in line 737)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
```diff
@@ -758,5 +758,20 @@ public int Start()
...
+        {
+            public int VarPatternCheck(string value)
+            {
+                switch (value)
+                {
+                    case var x when x == "one":
```
duncanp-sonar: Is this a regression test for the bug? I tried executing this test with the new `VarPatternSyntaxWrapper` class commented out and the DeadStores test didn't fail.

costin-zaharia-sonarsource: Indeed it only fails when running with MSBuild 16.

The way I tested it was to create a small project with this class and run the analyzer on it:
"%MSBUILD16%" "ConsoleApp.sln" /t:Clean,Build

and then update the .csproj to use the local analyzer and run again
```
 <ItemGroup>
    <Analyzer Include="..\..\..\..\..\..\src\sonar-dotnet\sonaranalyzer-dotnet\src\SonarAnalyzer.CSharp\bin\Debug\net46\SonarAnalyzer.dll" />
    <Analyzer Include="..\..\..\..\..\..\src\sonar-dotnet\sonaranalyzer-dotnet\src\SonarAnalyzer.CSharp\bin\Debug\net46\SonarAnalyzer.CSharp.dll" />
  </ItemGroup>
```

The worst part now is that we get 0 coverage for the new code since it will never run with the current build due to the fact that we use MsBuild 15.

duncanp-sonar: Thanks for the explanation. As discussed offline, please create a ticket about also running tests using MSBuild16.

Issue #2594, opened 2019-08-20
vbnet:S4144 :  throws an exception of type 'System.NullReferenceException'


W3bCorp:
### Description

When i build with TFS2018/VS2017/MsBuild 15 with SonarQube 7.9.1 and Sonar-scanner 3.3.0.1492, i got the error : 

2019-08-20T10:03:09.9054057Z ##[error]vbc(0,0): Error AD0001: Analyzer 'SonarAnalyzer.Rules.VisualBasic.MethodsShouldNotHaveIdenticalImplementations' threw an exception of type 'System.NullReferenceException' with message 'La référence d'objet n'est pas définie à une instance d'un objet.'.
2019-08-20T10:03:09.9054717Z ##[debug]Processed: ##vso[task.logissue type=Error;sourcepath=vbc;linenumber=0;columnnumber=0;code=AD0001;]Analyzer 'SonarAnalyzer.Rules.VisualBasic.MethodsShouldNotHaveIdenticalImplementations' threw an exception of type 'System.NullReferenceException' with message 'La référence d'objet n'est pas définie à une instance d'un objet.'.
2019-08-20T10:03:09.9060360Z vbc : error AD0001: Analyzer 'SonarAnalyzer.Rules.VisualBasic.MethodsShouldNotHaveIdenticalImplementations' threw an exception of type 'System.NullReferenceException' with message 'La référence d'objet n'est pas définie à une instance d'un objet.'. [G:\Builds\agents\day01\_work\1\s\IsiFrameWork\IsiBusinessLogic\IsiBusinessLogic.vbproj]
2019-08-20T10:03:09.9268948Z ##[debug]Processed: ##vso[task.logdetail id=9765be38-5431-4b86-b27f-4e154cb645ee;parentid=a570503f-71e3-4c21-adc8-32d46823bbd8;type=Build;result=Failed;finishtime=2019-08-20T10:03:09.9265529Z;progress=100;state=Completed;parentid=a570503f-71e3-4c21-adc8-32d46823bbd8;name=;]

### Repro steps

Standard build, with standard "Sonar way" quality profile. 
The repro is difficult because +100k of lines of code
When i disabled the rule "vbnet:S4144", all is ok.

I try to diagnostic with "sonar.verbose=true" and msbuild "-v:diag" args but i have no more informations. Visibly, the error occurs randomly and not always in the same file

### Related information

* SonarQube : 7.9.1
* Sonar-scanner : 3.3.0.1492
[...]

PR #2735, opened 2019-10-17
Fix NullReferenceException issues (#2594)

costin-zaharia-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Helpers/VisualBasicSyntaxHelper.cs
```diff
@@ -103,10 +103,12 @@ public static StatementSyntax GetSucceedingStatement(this StatementSyntax curren
 
         public static SyntaxToken? GetMethodCallIdentifier(this InvocationExpressionSyntax invocation)
         {
-           if (invocation == null)
+            if (invocation == null ||
+                invocation.Expression == null)
```
costin-zaharia-sonarsource: This fixes the exceptions from ThreadResumeOrSuspendShouldNotBeCalled and DangerousGetHandleShouldNotBeCalled from our cirrus ci job.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/MethodsShouldNotHaveIdenticalImplementations.cs
```diff
@@ -51,12 +51,17 @@ protected override IEnumerable<MethodBlockSyntax> GetMethodDeclarations(SyntaxNo
...
-                firstMethod.SubOrFunctionStatement.Identifier.ValueText != secondMethod.SubOrFunctionStatement.Identifier.ValueText &&
-                HaveSameParameters(firstMethod?.BlockStatement?.ParameterList.Parameters, secondMethod?.BlockStatement?.ParameterList.Parameters) &&
-                VisualBasicEquivalenceChecker.AreEquivalent(firstMethod.Statements, secondMethod.Statements);
+                   firstMethod.GetIdentifierText() != secondMethod.GetIdentifierText() &&
+                   HaveSameParameters(firstMethod.BlockStatement?.ParameterList.Parameters, secondMethod.BlockStatement?.ParameterList.Parameters) &&
+                   VisualBasicEquivalenceChecker.AreEquivalent(firstMethod.Statements, secondMethod.Statements);
 
             bool HaveSameParameters(SeparatedSyntaxList<ParameterSyntax>? leftParameters, SeparatedSyntaxList<ParameterSyntax>? rightParameters)
             {
+                if (leftParameters == null && rightParameters == null)
```
costin-zaharia-sonarsource: Since we cannot reproduce this issue and I don't have the full stack-trace I had to do a bit of guess work here. We seem to handle the cases when only one of the parameters (left or right) is null but not when both are null. I assume that this might be the reason why we get the exception but I'm not sure yet.

andrei-epure-sonarsource: From [the docs ](https://docs.microsoft.com/en-us/dotnet/visual-basic/programming-guide/language-features/procedures/sub-procedures), the declaration syntax for Sub is `[ modifiers ] Sub subname [( parameterlist )]`

from what I understand, it means the `()` are optional, so you can add tests in [MethodsShouldNotHaveIdenticalImplementations.vb](https://github.com/SonarSource/sonar-dotnet/blob/c6abd066b90df6d5b9f2fe484fb21d61ea404a30/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodsShouldNotHaveIdenticalImplementations.vb) where you have some subs like

```
        Sub FooNoParens1
            Dim s As String = "test"
            Console.WriteLine("Result: {0}", s)
        End Sub

        Sub FooNoParens2
            Dim s As String = "test"
            Console.WriteLine("Result: {0}", s)
        End Sub
```

ideally this should solve the coverage problem

costin-zaharia-sonarsource: Nice, That's what I was missing. I've updated the code.

PR #2768, opened 2019-11-06
S2068 should consider NetworkCredential and PasswordDeriveBytes

### LINKED ISSUE #2754: S2068 should consider NetworkCredential and PasswordDeriveBytes
S2068 shoud raise an issue when the "password" parameter of NetworkCredential and PasswordDeriveBytes is hard-coded.

NetworkCredential: https://docs.microsoft.com/en-us/dotnet/api/system.net.networkcredential
PasswordDeriveBytes: https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.passwordderivebytes

APIs to consider when "password" is hard-coded:
* NetworkCredential(String, String)
* NetworkCredential(String, String, String)
* PasswordDeriveBytes(String, Byte[])
* PasswordDeriveBytes(String, Byte[], CspParameters)
* PasswordDeriveBytes(String, Byte[], String, Int32)
* PasswordDeriveBytes(String, Byte[], String, Int32, CspParameters)

Microsoft Note:
> Never hard-code a password within your source code. Hard coded passwords can be retrieved from an assembly using the Ildasm.exe (IL Disassembler) tool, a hex editor, or by simply opening up the assembly in a text editor like notepad.exe.

christophe-zurn-sonarsource:
Fixes #2754


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
```diff
@@ -1,85 +1,108 @@
...
-                (syntaxNode.Left as IdentifierNameSyntax)?.Identifier.ValueText;
-
-            protected override bool IsAssignedWithStringLiteral(AssignmentExpressionSyntax syntaxNode, SemanticModel semanticModel) =>
-                syntaxNode.IsKind(SyntaxKind.SimpleAssignmentExpression) &&
-                syntaxNode.Left.IsKnownType(KnownType.System_String, semanticModel) &&
-                syntaxNode.Right.IsKind(SyntaxKind.StringLiteralExpression);
-        }
-    }
-}
+﻿/*
```
andrei-epure-sonarsource: line endings changed, could you revert please? (I cannot tell if there's a change or just rename)

(it is a nuissance, we'll have to adopt some common practice in the dotnet area and document it and enforce it with tooling, because currently the codebase is split evenly between unix and windows line endings...)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
```diff
@@ -130,6 +130,7 @@ public void AllParameterizedRules_AreDisabledByDefault()
                 .Where(RuleFinder.IsParameterized)
                 .Select(type => (DiagnosticAnalyzer)Activator.CreateInstance(type))
                 .SelectMany(analyzer => analyzer.SupportedDiagnostics)
+                .Where(analyzer => !IsSecurityHotspot(analyzer))
```
andrei-epure-sonarsource: `AllParameterizedRules_AreDisabledByDefault` - I am confused - does `Parameterized` refer to `WithNotConfigurable`?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
```diff
@@ -196,16 +197,16 @@ public void AllRules_SonarWayTagPresenceMatchesIsEnabledByDefault()
...
-                else if (IsSecurityHotspot(analyzer))
+                if (IsSecurityHotspot(analyzer))
                 {
                     // Security hotspots are enabled by default, but they will report issues only
                     // when their ID is contained in SonarLint.xml
                     analyzer.IsEnabledByDefault.Should().BeTrue($"{analyzer.Id} should be enabled by default");
                 }
+                else if (parameterizedAnalyzers.Contains(analyzer))
+                {
+                    analyzer.IsEnabledByDefault.Should().BeFalse($"{analyzer.Id} has parameters and should be disabled by default");
```
andrei-epure-sonarsource: we don't have any SonarWay rule which is parameterized?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
andrei-epure-sonarsource: or maybe just github cannot tell the diff.. I now look at the first commit

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
andrei-epure-sonarsource: I am a bit confused,  maybe a comment will help

christophe-zurn-sonarsource: There are parametrized rules that are in Sonar way (S3776) , but it was decided to not enable them by default even when they are in Sonar way (see #1274). I am not sure the easoning behind this, but probably FP raised for such cases.
Here, this is the first time we have a rule beeing a hotspot and that can be parametrized. I chose to go with the hotspot behaviour in such cases (activated by default), since there is technically no FP for hotspots

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
christophe-zurn-sonarsource: No, here `Parameterized` refer to a rule that has a `RuleParameter` aatribute, so that has a SonarQube parameter

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
christophe-zurn-sonarsource: I will add a link to the issue in the code as a comment

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
christophe-zurn-sonarsource: I tried changing line ending, but no matter what I put as line ending, github still marks the whole file as changed unfortunately

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs
```diff
@@ -27,7 +30,35 @@ public void Test()
             string myPassword1 = null;
```
pavel-mikula-sonarsource: Please add theese tests to reflect new regex. Both lines behaves the same in production, but the second test will fail with current code.

            foo = "Server=localhost; Database=Main; User=Admin; Password=TopSecret;";       //Noncompliant
            foo = "Server=localhost; Database=Main; User=Admin; Password   =   TopSecret;"; //Noncompliant

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.vb
```diff
@@ -19,9 +22,36 @@ Namespace Tests.Diagnostics
             Dim myPassword1 As String = Nothing
```
pavel-mikula-sonarsource: Please add theese tests to reflect new regex. Both lines behaves the same in production, but the second test will fail with current code.

            foo = "Server=localhost; Database=Main; User=Admin; Password=TopSecret;;       'Noncompliant
            foo = "Server=localhost; Database=Main; User=Admin; Password   =   TopSecret;; 'Noncompliant

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
```diff
@@ -58,17 +60,47 @@ public string CredentialWords
             }
         }
```
pavel-mikula-sonarsource: What's \b (backspace character \u0008) check good for?
I dont's see a reason for \S either. SQL password could start with underscore and that would be matched as white space.
And there's missing check for spaces before the delimiter [:=]. 
I think that the regex should look like this:   "(?<password>{0})\s*[:=]"

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs
```diff
@@ -27,7 +30,35 @@ public void Test()
...
             string myPassword2 = "";
             string myPassword3 = "        ";
-            string myPassword4 = @"foo"; // Noncompliant {{'password' detected in this expression, review this potentially hardcoded credential.}}
+            string myPassword4 = @"foo"; // Noncompliant
+        }
+
```
pavel-mikula-sonarsource: I suggest to add
private const string Secret= "LoremIpsum";
and
new NetworkCredential("username", Secret); // Noncompliant
to demonstrate that .IsConstant(..) is working.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.vb
```diff
@@ -19,9 +22,36 @@ Namespace Tests.Diagnostics
...
             Dim myPassword2 As String = ""
             Dim myPassword3 As String = "   "
-            Dim myPassword4 As String = "foo" 'Noncompliant {{'password' detected in this expression, review this potentially hardcoded credential.}}
+            Dim myPassword4 As String = "foo" 'Noncompliant
 
         End Sub
+
```
pavel-mikula-sonarsource: I suggest to add
Private Const Secret As String = "LoremIpsum"
and
passwordDeriveBytes = new NetworkCredential("username", Secret) // Noncompliant
to demonstrate that .IsConstant(..) is working.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/DiagnosticDescriptorBuilder.cs
```diff
@@ -107,6 +107,9 @@ private static string[] BuildUtilityCustomTags(SourceScope sourceScope)
                 .ToArray();
         }
 
+        /**
+         * Indicates that the roslyn diagnostic cannot be suppressed, filtered or have its severity changed.
+         */
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: It should. The same way it should prevent to activate the rule when it is not enabled (Sonarlint context) to avoid too many FP

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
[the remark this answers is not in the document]
christophe-zurn-sonarsource: Indeed, nice catch. I've created a new ticket for fixing this: #2783

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs
christophe-zurn-sonarsource: Ticket created to fix this: #2783

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.vb
christophe-zurn-sonarsource: Ticket created to fix this: #2783

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
christophe-zurn-sonarsource: Good point. I've created a new ticket to fix this: #2783 
Note that the `\S` part after the `[:=]` delimiter was to prevent FP in the cases `"Server=localhost; Database=Main; User=Admin; Password= " + loadPassword()`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.vb
christophe-zurn-sonarsource: Example added in unit test

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs
christophe-zurn-sonarsource: Example added in unit test

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
christophe-zurn-sonarsource: Good point. I've created a new ticket to fix this: #2783 
Note that the `\S` part after the `[:=]` delimiter was to prevent FP in the cases `"Server=localhost; Database=Main; User=Admin; Password= " + loadPassword()`

Issue #2606, opened 2019-08-26
Fix S4004 FP: Should not raise when enforced by interface or base class


Corniel:
### Description
As classes that have to implement writable collection properties have no influence on changing the contract S4004 ( Collection properties should be read-only) should not be raised. Related to #2344.

### Expected behavior
``` C#
public class S4004Interface : IS4004
{
    public IDictionary<object, object> Items { get; set; }// Compliant, enforced by interface
}
public interface IS4004
{
    IDictionary<object, object> Items { get; set; } // non-compliant
}

public class S4004Abstract : S4004Base
{
    public override IDictionary<object, object> Items { get; set; } // Compliant, enforced by base
}
public abstract class S4004Base
{
    public abstract IDictionary<object, object> Items { get; set; } // non-compliant
}
```

### Related information
* SonarC# Version 7.16

PR #2770, opened 2019-11-06
Add repro for #2606

costin-zaharia-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CollectionPropertiesShouldBeReadOnly.cs
```diff
@@ -1,71 +1,91 @@
-using System.Collections;
```
andrei-epure-sonarsource: there's some line endings problem here

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CollectionPropertiesShouldBeReadOnly.cs
```diff
@@ -1,71 +1,91 @@
...
+        public string MessageBody { get; set; }
+
+        [DataMember]
+        public Dictionary<string, string> Properties { get; set; }
+
+        [System.Runtime.Serialization.DataMember]
+        public List<int> Values { get; set; }
+    }
+
+    public abstract class S4004Base
```
andrei-epure-sonarsource: it would be nice to reference the ticket which causes this

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CollectionPropertiesShouldBeReadOnly.cs
```diff
@@ -1,71 +1,91 @@
...
+    {
+        public abstract IDictionary<object, object> Items { get; set; } // Noncompliant
+    }
+
+    public class S4004Abstract : S4004Base
```
andrei-epure-sonarsource: it would be nice to reference the ticket from which we found out this

PR #2778, opened 2019-11-07
Add reproducers for FP

christophe-zurn-sonarsource:
(Commits should not be squashed)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteCollectionElements.vb
```diff
@@ -62,6 +62,11 @@ Namespace Tests.TestCases
       list(0) = 1 ' Noncompliant
     End Sub
 
+	Private Sub ListAddFP(ByVal list As List(Of String))
```
andrei-epure-sonarsource: nitpick tab instead of space

andrei-epure-sonarsource:
hm, I squashed commits, sorry (but kept the message inside)

PR #2785, opened 2019-11-13
Rule S4830: Server certificates should be verified during SSL/TLS connections (#2755)

### LINKED ISSUE #2755: Rule S4830: Server certificates should be verified during SSL/TLS connections
Implements https://jira.sonarsource.com/browse/RSPEC-4830

C# Code Sample: https://jira.sonarsource.com/browse/RSPEC-5582

pavel-mikula-sonarsource:
New rule.
Fixes #2755


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,310 @@
...
+        {
+            var ret = ImmutableArray.CreateBuilder<Location>();
+            ISymbol MS;
+            switch (expression)
+            {
+                case IdentifierNameSyntax identifier:
+                    MS = c.Context.SemanticModel.GetSymbolInfo(identifier).Symbol;
+                    if (MS != null)
+                    {
+                        foreach (var syntax in MS.DeclaringSyntaxReferences.Select(x => x.GetSyntax()))
```
christophe-zurn-sonarsource: I am not sure to understand why this foreach loop is necessary here. Replacing the for-loop with
```
if (MS != null && MS.DeclaringSyntaxReferences.Length == 1)
{
    ret.AddRange(IdentifierLocations(c, MS.DeclaringSyntaxReferences[0].GetSyntax()));
}
```
does not change the unit test results. Is it possible to have a case where we have more than 1 declaring syntax node in our case ?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,310 @@
...
+                case InvocationExpressionSyntax invocation:
+                    MS = c.Context.SemanticModel.GetSymbolInfo(invocation).Symbol;
+                    if (MS != null)
+                    {
+                        foreach (var syntax in MS.DeclaringSyntaxReferences.Select(x => x.GetSyntax() as MethodDeclarationSyntax).Where(x => x != null))
+                        {
+                            c.VisitedMethods.Add(syntax);
+                            ret.AddRange(InvocationLocations(c, syntax));
+                        }
+                    }
```
christophe-zurn-sonarsource: Same question/remark as on line 124: if I replace the for-loop, I don't see any change in unit tests:
```
                            if (MS != null && MS.DeclaringSyntaxReferences.Length == 1 && MS.DeclaringSyntaxReferences[0].GetSyntax() is MethodDeclarationSyntax syntax)
                            {
                                c.VisitedMethods.Add(syntax);
                                ret.AddRange(InvocationLocations(c, syntax));
                            }
```
So I wonder if the `foreach` is truly necessary, as I don't believe we can create a case where the declaration syntax has several nodes.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,310 @@
...
+                .Where(x => !IsVisited(c, x));      //Ignore all return statements with recursive call. Result depends on returns that could return compilant validator.
+            return MultiExpressionSublocations(c, returnExpressionSublocationsList);
+        }
+
+        private static bool IsVisited(InspectionContext c, ExpressionSyntax expression)
+        {
+            if (expression is InvocationExpressionSyntax invocation)
+            {
+                var MS = c.Context.SemanticModel.GetSymbolInfo(invocation).Symbol;
+                return MS != null && MS.DeclaringSyntaxReferences.Length != 0 && MS.DeclaringSyntaxReferences.Select(x => x.GetSyntax() as MethodDeclarationSyntax).Any(x => x != null && c.VisitedMethods.Contains(x));
```
christophe-zurn-sonarsource: _minor_: we tend to use `!xxx.IsEmpty` syntax for readibility instead of `xxx.Length != 0`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,310 @@
...
+            {
+                return ImmutableArray<Location>.Empty;
+            }
+            return exprSublocationsList.SelectMany(x => x).ToImmutableArray();      //Else every return statement is noncompliant
+        }
+
+        private static ImmutableArray<Location> CallStackSublocations(InspectionContext c, ExpressionSyntax expression)
+        {
+            var lst = ArgumentLocations(c, expression);
+            if (lst.Length != 0)        //There's noncompilant issue in this chain
```
christophe-zurn-sonarsource: _minor_: we tend to use `!xxx.IsEmpty` syntax for readibility instead of `xxx.Length != 0`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,411 @@
...
+            CreateRQ().ServerCertificateValidationCallback += (sender, certificate, chain, SslPolicyErrors) => false;
+            CreateRQ().ServerCertificateValidationCallback += (sender, certificate, chain, SslPolicyErrors) => certificate.Subject == "Test";
+
+            //Without variable
+            //Secondary@+1
+            CreateRQ().ServerCertificateValidationCallback += (sender, certificate, chain, SslPolicyErrors) => true;    //Noncompliant
+
+            //Other occurances
+            var rq = CreateRQ();
+            rq.ServerCertificateValidationCallback += InvalidValidation;                           //Noncompliant
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I would like to keep the block-style indent of comments in order to have the source text easy to read.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,411 @@
...
+            //var ws = new System.Net.WebSockets.ClientWebSocket();
+            //ws.Options.RemoteCertificateValidationCallback += InvalidValidation;
+
+            //Do not test this one. It's .NET Standard 2.1 target only. It shoudl work since we're hunting RemoteCertificateValidationCallback and method signature
+            //var sslOpts = new System.Net.Security.SslClientAuthentication();
+            //Security.SslClientAuthenticationOptions.RemoteCertificateValidationCallback;
+        }
+
+        void MultipleHandlers() {
+            ServicePointManager.ServerCertificateValidationCallback += CompilantValidation;
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I've fixed all occurrences. There were some more in the comments of the rule as well. It's somehow hard-wired in my keyboard typing.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
christophe-zurn-sonarsource: To avoid unexpected false positives, I would actually only consider the cases the we are sure of. 
For instance, taking the xml as example, if we have a rule raising an issue when all child nodes are true, in the case of partial definition at multiple locations:
```
<!-- Partial def 1 -->
<Notes>
    <Note1>true<Note1>
<Notes>

<!-- Partial def 2 -->
<Notes>
    <Note2>false<Note2>
<Notes>
```
Current rule would raise an issue here, as it found locations where all child nodes were true.
So out of precaution, I would only handle the case where we know that there is a single definition location, as we know what to expect then

christophe-zurn-sonarsource:
Added a few last comments related to minor styling issues that I missed on the first look. Then we will be good to go :)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpEquivalenceChecker.cs
```diff
@@ -37,4 +38,18 @@ public static bool AreEquivalent(SyntaxList<SyntaxNode> nodeList1, SyntaxList<Sy
                 (n1, n2) => SyntaxFactory.AreEquivalent(n1, n2));
         }
     }
+
+    internal class CSharpSyntaxNodeEqualityComparer<T> : IEqualityComparer<T> where T : SyntaxNode
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Very good point. It actually didn't behave as intended.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,306 @@
...
+            return MultiExpressionSublocations(c, returnExpressionSublocationsList);
+        }
+
+        private static bool IsVisited(InspectionContext c, ExpressionSyntax expression)
+        {
+            if (expression is InvocationExpressionSyntax invocation)
+            {
+                var symbol = c.Context.SemanticModel.GetSymbolInfo(invocation).Symbol;
+                return symbol != null && !symbol.DeclaringSyntaxReferences.IsEmpty
+                    && symbol.DeclaringSyntaxReferences.Select(x => x.GetSyntax() as MethodDeclarationSyntax).Any(x => x != null && c.VisitedMethods.Contains(x));
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: Ok. Fine by me

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,299 @@
...
+                    }
+                    break;
+                case ParenthesizedLambdaExpressionSyntax lambda:
+                    if ((lambda.Body as LiteralExpressionSyntax)?.Kind() == SyntaxKind.TrueLiteralExpression)
+                    {
+                        ret.Add(lambda.Body.GetLocation());   //Code was found guilty for lambda (...) => true
+                    }
+                    else if (lambda.Body is BlockSyntax block)
+                    {
+                        ret.AddRange(BlockLocations(block));
```
christophe-zurn-sonarsource: I just checked the coverage on [SonarCloud](https://sonarcloud.io/code?id=sonaranalyzer-dotnet&line=130&pullRequest=2785&selected=sonaranalyzer-dotnet%3Asrc%2FSonarAnalyzer.CSharp%2FRules%2FCertificateValidationCheck.cs) and this line is actually not covered.
For completeness sake, it would be nice to have one additional test case with lambda having a block body:
```
            CreateRQ().ServerCertificateValidationCallback += (sender, certificate, chain, SslPolicyErrors) =>
            {
                return true;
            };

            CreateRQ().ServerCertificateValidationCallback += (sender, certificate, chain, SslPolicyErrors) =>
            {
                return false;
            };
```

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/AbstractMethodParameterLookup.cs
```diff
@@ -70,6 +70,20 @@ public bool TryGetParameterSymbol(TArgumentSyntax argument, out IParameterSymbol
             return true;
         }
 
+        public bool TryGetSymbolParameter(IParameterSymbol parameter, out TArgumentSyntax argument)
```
christophe-zurn-sonarsource: This method could also have a dedicated unit test for testing its actual behaviour

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -135,13 +132,10 @@ private static ImmutableArray<Location> ArgumentLocations(InspectionContext c, E
                     break;
                 case InvocationExpressionSyntax invocation:
                     var invSymbol = c.Context.SemanticModel.GetSymbolInfo(invocation).Symbol;
-                    if (invSymbol != null)
+                    if (invSymbol != null && invSymbol.DeclaringSyntaxReferences.Length==1 && invSymbol.DeclaringSyntaxReferences.Single().GetSyntax() is MethodDeclarationSyntax syntax)
                     {
-                        foreach (var syntax in invSymbol.DeclaringSyntaxReferences.Select(x => x.GetSyntax() as MethodDeclarationSyntax).Where(x => x != null))
-                        {
                             c.VisitedMethods.Add(syntax);
                             ret.AddRange(InvocationLocations(c, syntax));
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Some copy/paste scenarios or line moving scenarios do not raise automatic formatting properly. That's why I remember Ctrl+K, C shortcut.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/MethodParameterLookupTest.cs
```diff
@@ -0,0 +1,262 @@
...
+            Assert.IsNotNull(mainMethod);   
+
+            foreach (var invocation in mainMethod.DescendantNodes().OfType<CSharpSyntax.InvocationExpressionSyntax>())
+            {
+                var method = compiler.SemanticModel.GetSymbolInfo(invocation).Symbol as IMethodSymbol;
+                var lookup = new CSharpMethodParameterLookup(invocation.ArgumentList, method);
+                var context = new CSharpInspectionContext();
+                context.Init(compiler, lookup, method, invocation.ArgumentList.Arguments.ToArray(), specialArgument, specialParameter);
+                InspectLookupValues(context);
+            }
```
christophe-zurn-sonarsource: It is really hard to read and understand what the test actually does, as it has logic regarding expected values spread across several helper methods, and the for-loop makes it hard to know what we are testing at one point.
It would be easier to have one helper method doing the all the logic and one doing the assertions on what we expect.

I would have a much easier time reading something similar to:
```
var doSomethingFirstInvocation = GetInvocationExpression(mainMethod, "DoSomething")[0];
assertParameterNames(doSomethingFirstInvocation, "a", "b", "c");
assertParameterValues(doSomethingFirstInvocation, 42, true, null, "Lorem");

var doSomethingSecondInvocation = GetInvocationExpression(mainMethod, "DoSomething")[1];
assertParameterNames(doSomethingSecondInvocation , "a", "b", "c");
assertParameterValues(doSomethingSecondInvocation , 42, true, null, "Lorem");
```

Which would also allow to test different values for the parameters, as at the moment each parameter has a static mapping between parameter name and parameter value.

sonaranalyzer-dotnet/.sonarlint/SonarAnalyzer.slconfig
```diff
@@ -9,7 +9,7 @@
   "Profiles": {
     "CSharp": {
       "ProfileKey": "AVxYKDdrFTbgxqUNcakm",
-      "ProfileTimestamp": "2019-05-07T17:01:40+02:00"
+      "ProfileTimestamp": "2019-10-29T09:27:27+01:00"
```
christophe-zurn-sonarsource: Changes in this file are unrelated to the current PR. If this was intended, I think it deserves a separate PR

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/MethodParameterLookupTest.cs
```diff
@@ -0,0 +1,290 @@
...
+        {
+        }
+
+        void SpecialMethod(int specialParameter)
+        {
+        }
+    }
+}";
+            var c = new CSharpInspection(Source);
+            c.Inspect(0, "DoNothing", new { });
```
christophe-zurn-sonarsource: I would suggest to rename the `Inspect` method to something more explicit, like `HasExpectedParameterMappings(...` to be clear at first glance on what it does

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,299 @@
...
+            //Handling of += delegate syntax
+            context.RegisterSyntaxNodeActionInNonGenerated(c => CheckAddHandlerSyntax(c), SyntaxKind.AddAssignmentExpression);
+
+            //Handling of constructor parameter syntax (SslStream)
+            context.RegisterSyntaxNodeActionInNonGenerated(c => CheckConstructorParameterSyntax(c), SyntaxKind.ObjectCreationExpression);
+        }
+
+        private void CheckAddHandlerSyntax(SyntaxNodeAnalysisContext c)
+        {
+            var leftIdentifier = ((AssignmentExpressionSyntax)c.Node).Left.DescendantNodesAndSelf().OfType<IdentifierNameSyntax>().LastOrDefault();
```
andrei-epure-sonarsource: minor: you can do the cast `(AssignmentExpressionSyntax)c.Node` once and extract it in a variable

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,299 @@
...
+        private void CheckConstructorParameterSyntax(SyntaxNodeAnalysisContext c)
+        {
+            if (c.SemanticModel.GetSymbolInfo(c.Node).Symbol is IMethodSymbol ctor)
+            {
+                CSharpMethodParameterLookup methodParamLookup = null;       //Cache, there might be more of them
+                foreach (var param in ctor.Parameters.Where(x => IsValidationDelegateType(x.Type)))
+                {
+                    methodParamLookup = methodParamLookup ?? new CSharpMethodParameterLookup((c.Node as ObjectCreationExpressionSyntax).ArgumentList, ctor);
+                    if (methodParamLookup.TryGetSymbolParameter(param, out var argument))
+                    { //For Lambda expression extract location of the parentheses only to separate them from secondary location of "true"
```
andrei-epure-sonarsource: nitpick: we normally put comments on the next line (not on the same line with `{`)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,299 @@
...
+                    }
+                }
+            }
+        }
+
+        private void TryReportLocations(InspectionContext c, Location primaryLocation, ExpressionSyntax expression)
+        {
+            var locations = ArgumentLocations(c, expression);
+            if (!locations.IsEmpty)
+            {   //Report both, assignemnt as well as all implementation occurances
```
andrei-epure-sonarsource: nitpick: same here, to be consistent with the codebase the comment should be on next line

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,299 @@
...
+        }
+
+        private static ImmutableArray<Location> ArgumentLocations(InspectionContext c, ExpressionSyntax expression)
+        {
+            var ret = ImmutableArray.CreateBuilder<Location>();
+            switch (expression)
+            {
+                case IdentifierNameSyntax identifier:
+                    var identSymbol = c.Context.SemanticModel.GetSymbolInfo(identifier).Symbol;
+                    if (identSymbol != null && identSymbol.DeclaringSyntaxReferences.Length == 1)
```
andrei-epure-sonarsource: _(subjective)_ you could skip the need of a `var` declaration by doing in the if
```
if (c.Context.SemanticModel.GetSymbolInfo(identifier).Symbol is ISymbol identSymbol &&
  identSymbol.DeclaringSyntaxReferences.Length == 1)
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,299 @@
...
+                        ret.Add(lambda.Body.GetLocation());   //Code was found guilty for lambda (...) => true
+                    }
+                    else if (lambda.Body is BlockSyntax block)
+                    {
+                        ret.AddRange(BlockLocations(block));
+                    }
+                    break;
+                case InvocationExpressionSyntax invocation:
+                    var invSymbol = c.Context.SemanticModel.GetSymbolInfo(invocation).Symbol;
+                    if (invSymbol != null && invSymbol.DeclaringSyntaxReferences.Length == 1 && invSymbol.DeclaringSyntaxReferences.Single().GetSyntax() is MethodDeclarationSyntax syntax)
```
andrei-epure-sonarsource: _subjective_ same here, I find it more expressive to avoid local vars and null-checks if not necessary

```
if (c.Context.SemanticModel.GetSymbolInfo(invocation).Symbol is ISymbol invSymbol && ...)

```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,299 @@
...
+            }
+            return lst;
+        }
+
+        private static ImmutableArray<Location> BlockLocations(BlockSyntax block)
+        {
+            var ret = ImmutableArray.CreateBuilder<Location>();
+            if (block != null)
+            {
+                //ToDo: VB.NET vs. return by assign to function name
```
andrei-epure-sonarsource: we need to get alligned between `TODO` and `ToDo` - normally in the codebase it's `TODO` :)

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/UsingCookiesBase.cs
```diff
@@ -73,15 +73,15 @@ protected override void Initialize(SonarAnalysisContext context)
             InvocationTracker.Track(context,
                 InvocationTracker.MatchMethod(
                     new MemberDescriptor(KnownType.Microsoft_AspNetCore_Http_IResponseCookies, "Append")));
-
+
```
andrei-epure-sonarsource: nitpick: trailing whitespace

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/UsingCookiesBase.cs
```diff
@@ -73,15 +73,15 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            
             InvocationTracker.Track(context,
                 InvocationTracker.MatchMethod(
                     new MemberDescriptor(KnownType.System_Collections_Generic_IDictionary_TKey_TValue, "Add"),
                     new MemberDescriptor(KnownType.System_Collections_Generic_IDictionary_TKey_TValue_VB, "Add")),
                 InvocationTracker.ArgumentAtIndexEquals(0, "Set-Cookie"),
                 InvocationTracker.MethodHasParameters(2),
                 IsIHeadersDictionary());
-
+
```
andrei-epure-sonarsource: nitpick: trailing whitespace

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
andrei-epure-sonarsource: no need to change now, just to keep in mind for the coding style discussion

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
pavel-mikula-sonarsource: Christope has the same note and it has been discussed [here](https://github.com/SonarSource/sonar-dotnet/pull/2785#discussion_r346245941)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
pavel-mikula-sonarsource: Same as [above](https://github.com/SonarSource/sonar-dotnet/pull/2785#discussion_r346245941)

PR #2789, opened 2019-11-13
Fix integration tests to run with VS2019 / MSBuild 16 (#2489)

### LINKED ISSUE #2489: Fix integration tests to run with VS2019 / MSBuild 16
### Description

When running the ITs having only VS2019 installed, it is failing during the build of a project with a generic S4MSB error message.

### Repro steps

Uninstall all VS versions. Install only VS2019. Build `sonar-dotnet` and run the integration tests.

### Expected behavior

ITs run and pass successfully on the `master` branch.

### Actual behavior

Fails with the error:

### Known workarounds

Install VS2017 and change the path to point to the old version of MSBuild.

pavel-mikula-sonarsource:
Fixes #2489

### DIFF
--- a/scripts/build/build-utils.ps1
+++ b/scripts/build/build-utils.ps1
@@ -15,8 +15,8 @@ function Get-MsBuildPath([ValidateSet("14.0", "15.0", "16.0", "Current")][string
 
     #If MSBUILD_PATH environment variable is found, the version is not checked and the value of input parameter is ignored.
     Write-Host "Trying to find 'msbuild.exe' using 'MSBUILD_PATH' environment variable"
-    $msbuildEnv = "MSBUILD_PATH"
-    $msbuildPath = [environment]::GetEnvironmentVariable($msbuildEnv, "Process")
+    $msbuildPathEnvVar = "MSBUILD_PATH"
+    $msbuildPath = [environment]::GetEnvironmentVariable($msbuildPathEnvVar, "Process")
 
     if (!$msbuildPath) {
         Write-Host "Environment variable ${msbuildEnv} not found"
@@ -27,7 +27,7 @@ function Get-MsBuildPath([ValidateSet("14.0", "15.0", "16.0", "Current")][string
         # Test if vswhere.exe is in your path. Download from: https://github.com/Microsoft/vswhere/releases
         $msbuildPath = Exec { & (Get-VsWherePath) -latest -requires Microsoft.Component.MSBuild -find MSBuild\**\Bin\MSBuild.exe } | Select-Object -First 1
         if ($msbuildPath) {
-            [environment]::SetEnvironmentVariable($msbuildEnv, $msbuildPath)
+            [environment]::SetEnvironmentVariable($msbuildPathEnvVar, $msbuildPath)
         }
     }
 
--- a/scripts/build/build-utils.ps1
+++ b/scripts/build/build-utils.ps1
@@ -8,36 +8,35 @@ function Get-VsWherePath {
     return Get-ExecutablePath -name "vswhere.exe" -envVar "VSWHERE_PATH"
 }
 
-function Get-MsBuildPath([ValidateSet("14.0", "15.0")][string]$msbuildVersion) {
+function Get-MsBuildPath([ValidateSet("14.0", "15.0", "16.0", "Current")][string]$msbuildVersion) {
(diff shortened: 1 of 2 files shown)


scripts/build/build-utils.ps1
```diff
@@ -8,36 +8,35 @@ function Get-VsWherePath {
     return Get-ExecutablePath -name "vswhere.exe" -envVar "VSWHERE_PATH"
 }
 
-function Get-MsBuildPath([ValidateSet("14.0", "15.0")][string]$msbuildVersion) {
+function Get-MsBuildPath([ValidateSet("14.0", "15.0", "16.0", "Current")][string]$msbuildVersion) {
```
andrei-epure-sonarsource: Given the [official docs](https://docs.microsoft.com/en-us/visualstudio/msbuild/whats-new-msbuild-16-0?view=vs-2019#changed-path) say that MSBuild 16 is located under _Current_ , is there a reason to have `16.0` there?

scripts/build/build-utils.ps1
```diff
@@ -8,36 +8,35 @@ function Get-VsWherePath {
...
     if ($msbuildVersion -eq "14.0") {
         return Get-ExecutablePath -name "msbuild.exe" -envVar "MSBUILD_PATH"
     }
 
-    Write-Host "Trying to find 'msbuild.exe 15' using 'MSBUILD_PATH' environment variable"
-    $msbuild15Env = "MSBUILD_PATH"
-    $msbuild15Path = [environment]::GetEnvironmentVariable($msbuild15Env, "Process")
+    #If MSBUILD_PATH environment variable is found, the version is not checked and the value of input parameter is ignored.
+    Write-Host "Trying to find 'msbuild.exe' using 'MSBUILD_PATH' environment variable"
+    $msbuildEnv = "MSBUILD_PATH"
```
andrei-epure-sonarsource: would `$msbuildPathEnvVar` be a better name?

because just seeing `$msbuildEnv` I was thinking at "msbuild environment" (like in "MSBuild 16 environment"), whereas this is a constant name for a env variable

andrei-epure-sonarsource: applies to all places where `16.0` is used

scripts/build/build-utils.ps1
andrei-epure-sonarsource: applies to all places where `16.0` is used

andrei-epure-sonarsource:
Also, I guess this should fix #2489, so you can add "Fixes #2489" in the PR description to auto-link

scripts/build/build-utils.ps1
pavel-mikula-sonarsource: One day, there will be 17.0 located in "Current". That's why I used it everywhere like this to follow original pattern of "14.0", "15.0".
From semantic point of view, someone may like to be explicit about version 16 (and stay there when 17 is released) and someone may just want the current, no matter what it means.

andrei-epure-sonarsource: did you test the code paths with the `16.0` folder?

I believe you should document (in comments in the script header could be a choice) a setting which involves manually creating a folder on the drive

because using the default installation will never create the `16.0` folder, right? people will have to manually create a folder and copy paste binaries?

scripts/build/build-utils.ps1
```diff
@@ -8,36 +8,35 @@ function Get-VsWherePath {
...
-        # All subsequent builds after this command will use MSBuild 15!
+        # Sets the path to MSBuild into an the MSBUILD_PATH environment variable
+        # All subsequent builds after this command will use that MSBuild!
         # Test if vswhere.exe is in your path. Download from: https://github.com/Microsoft/vswhere/releases
-        $path = Exec { & (Get-VsWherePath) -version "[15.0, 16.0)" -products * -requires Microsoft.Component.MSBuild `
-            -property installationPath } | Select-Object -First 1
-        if ($path) {
-            $msbuild15Path = Join-Path $path "MSBuild\15.0\Bin\MSBuild.exe"
-            [environment]::SetEnvironmentVariable($msbuild15Env, $msbuild15Path)
+        $msbuildPath = Exec { & (Get-VsWherePath) -latest -requires Microsoft.Component.MSBuild -find MSBuild\**\Bin\MSBuild.exe } | Select-Object -First 1
```
andrei-epure-sonarsource: this will always take the latest installed version

PR #2797, opened 2019-11-18
Enforce Linux line endings ("lf") for code files (C# and VB)

costin-zaharia-sonarsource:


andrei-epure-sonarsource:
before merging it - @christophe-zurn-sonarsource @pavel-mikula-sonarsource  - do you think this will interact with some of the files you are working on? to avoid conflicts before merging some of the open PRs?

christophe-zurn-sonarsource:
@andrei-epure-sonarsource It might, but it will be the case for any ongoing open PRs while we haven't rebased everything on this new commit. So either we wait for all PR to be merged, or we merge it and adapt open PR to be compliant with new attribute.
IMO we should merge it, otherwise we will delay this PR indefinitely

PR #2798, opened 2019-11-18
Fix S1128 FP: documentation tags are not taken into account

### LINKED ISSUE #2694: Fix S1128 FP: documentation tags are not taken into account
### Description

[Rule S1128: Unused "using" should be removed](https://rules.sonarsource.com/csharp/RSPEC-1128)

When checking which namespaces are used, documentation tags like `<see cref='SomeClass' />` are not taken into account.

### Repro steps
``` C#
using System; // Compliant, used by cref.

{
    public enum S1128
    {
        /// <summary><see cref="DateTime"/></summary>
        DateTimeValue,
    }
}
```

### Known workarounds

``` C#
{
    public enum S1128
    {
        /// <summary><see cref="System.DateTime"/></summary>
        DateTimeValue,
    }
}
```

### Related information
* SonarC# Version 7.17.0.9346

christophe-zurn-sonarsource:
Fixes #2694


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnnecessaryUsings.cs
```diff
@@ -88,6 +90,28 @@ namespace MyNamespace3
     class Ns3_0 { }
 }
 
+namespace ReferencesInsideDocumentationTags
+{
+    using System.ComponentModel;
+
+    /// <summary><see cref="Win32Exception"/></summary>
```
pavel-mikula-sonarsource: Just update a test a little to demonstrate that it really iterates all of them in single statement as well:
```suggestion
    /// <summary>There's <see cref="Win32Exception"/> or <see cref="WarningException"/>.</summary>
```

christophe-zurn-sonarsource: I added another class instead, as `WarningException` is from the same namespace as the other, so it would be compliant even if it stopped at the first found class.

Issue #2795, opened 2019-11-18
[VB.NET] Rule S4830: Server certificates should be verified during SSL/TLS connections


pavel-mikula-sonarsource:
Implements https://jira.sonarsource.com/browse/RSPEC-4830 for VB.NET

C# Code Sample: https://jira.sonarsource.com/browse/RSPEC-5582

Based on #2755

Issue #2784, opened 2019-11-13
Fix the `IssueLocationCollector` for rules unit tests


christophe-zurn-sonarsource:
In the rules unit tests, any line containing `Noncompliant` will expect an issue from the a rule, no matter whether it is inside a comment or not. Example:
```
public void MyNoncompliantMethod()
{
  // doSomethind
}
```
In the above code, an issue is expected on the line of the method name as it contains "Noncompliant" in the name.
The `ISSUE_LOCATION_PATTERN` should be changed from `COMMENT_PATTERN + "*" + ...` to `COMMENT_PATTERN + "+" + ...`.

Furtheremore, messages displayed are often unclear:
- when a Secondary location is involved, it is hard to know that it was actually a secondary that was expected
- when unexpected issues occur in the unit test, only the first line is mentioned in the error message, whereas all the lines could be displayed (as it is already done for missing expected issues message)
- compilation error messages are in the same format as nonexpected issues message

Finally, as a secondary location may appear as secondary several times for different main issues, it would be nice to have an easy way to write it, such as (example, format to be discussed): `var myStr = "secondaryIssueLine" // Secondary:8x`

pavel-mikula-sonarsource:
Due to VB.NET Pretty Listing config:

Please update the logic, that any commented line containing only ^ and white space is considered as location marker. Even when it does not start at the beginning of the line.

pavel-mikula-sonarsource:
Situation with two secondary locations does not evaluate the positions of ^^^^ exactly. It seems that partial mach is enough.
See `C:\Projects\sonar-dotnet\sonaranalyzer-dotnet\tests\SonarAnalyzer.UnitTest\TestCases\CertificateValidationCheck.cs` and `ConstructorArguments()` method.
When markers are broken in way that they partially overlaps expected locations, it is evaluated as valid.

pavel-mikula-sonarsource:
VB specific situation - initial '^ should be tolerated as '^ OR ^^
```
    dict.Add(0, 0) ' Secondary
    '^^^^^^^^^^^^^
```

PR #2803, opened 2019-11-20
[VB.NET] Rule S4830: Server certificates should be verified during SSL/TLS connections

pavel-mikula-sonarsource:
Add rule S4830 for VB.NET #2795


pavel-mikula-sonarsource:
Please run all local ITS for S4830 to verify new project setup please.

christophe-zurn-sonarsource:
> Please run all local ITS for S4830 to verify new project setup please.

Done. All green on my side 👍

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs
```diff
@@ -0,0 +1,298 @@
...
+                    && parameters[2].Type.Is(KnownType.System_Security_Cryptography_X509Certificates_X509Chain)
+                    && parameters[3].Type.Is(KnownType.System_Net_Security_SslPolicyErrors)
+                    && namedSymbol.DelegateInvokeMethod.ReturnType.Is(KnownType.System_Boolean);
+            }
+            return false;
+        }
+
+        private ImmutableArray<Location> ArgumentLocations(InspectionContext c, TExpressionSyntax expression)
+        {
+            switch (TryExtractAddressOfOperand(expression))
```
christophe-zurn-sonarsource: I suggest to rename `TryExtractAddressOfOperand` as it is very VB.NET specific. Something like `GetMethodReferenceNode` (name subject to change) would make sense in both C# & VB contexts

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs
```diff
@@ -0,0 +1,298 @@
...
+            }
+            return false;
+        }
+
+        private ImmutableArray<Location> MultiExpressionSublocations(InspectionContext c, IEnumerable<TExpressionSyntax> expressions)
+        {
+            var exprSublocationsList = expressions.Distinct(CreateNodeEqualityComparer())
+                .Select(x => CallStackSublocations(c, x))
+                .ToArray();
+            if (exprSublocationsList.Any(x => x.IsEmpty))   //If there's at leat one concurrent expression, that returns compliant delegate, than there's some logic and this scope is compliant
```
christophe-zurn-sonarsource: minor typo
```suggestion
            if (exprSublocationsList.Any(x => x.IsEmpty))   //If there's at leat one concurrent expression, that returns compliant delegate, then there's some logic and this scope is compliant
```

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs
```diff
@@ -0,0 +1,298 @@
...
+
+        protected ImmutableArray<Location> ParamLocations(InspectionContext c, TParameterSyntax param)
+        {
+            var ret = ImmutableArray.CreateBuilder<Location>();
+            var containingMethodDeclaration = param.FirstAncestorOrSelf<TMethodSyntax>();
+            if (!c.VisitedMethods.Contains(containingMethodDeclaration))
+            {
+                c.VisitedMethods.Add(containingMethodDeclaration);
+                var containingMethod = c.Context.SemanticModel.GetDeclaredSymbol(containingMethodDeclaration) as IMethodSymbol;
+                var paramSymbol = containingMethod.Parameters.Single(x => x.Name == IdentifierText(param));
```
christophe-zurn-sonarsource: You can extract `IdentifierText(param)` outside the iteration as it will always have the same value

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs
```diff
@@ -0,0 +1,298 @@
...
+            var parentScope = LocalVariableScope(variable);
+            if (parentScope != null)
+            {
+                allAssignedExpressions.AddRange(parentScope.DescendantNodes().OfType<TAssignmentExpressionSyntax>()
+                    .Select(x =>
+                    {
+                        SplitAssignment(x, out var leftIdentifier, out var right);
+                        return new { leftIdentifier, right };
+                    })
+                    .Where( x => x.leftIdentifier != null && IdentifierText(x.leftIdentifier) == IdentifierText(variable))
```
christophe-zurn-sonarsource: You can extract `IdentifierText(variable)` outside the iteration as it will always have the same value

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs
```diff
@@ -0,0 +1,298 @@
...
+    public abstract class CertificateValidationCheckBase<TMethodSyntax, TArgumentSyntax, TExpressionSyntax, TIdentifierNameSyntax, TAssignmentExpressionSyntax, TInvocationExpressionSyntax, TParameterSyntax, TVariableSyntax, TLambdaSyntax> : SonarDiagnosticAnalyzer
+        where TMethodSyntax : SyntaxNode
+        where TArgumentSyntax : SyntaxNode
+        where TExpressionSyntax : SyntaxNode
+        where TIdentifierNameSyntax : SyntaxNode
+        where TAssignmentExpressionSyntax : SyntaxNode
+        where TInvocationExpressionSyntax : SyntaxNode
+        where TParameterSyntax : SyntaxNode
+        where TVariableSyntax : SyntaxNode
+        where TLambdaSyntax : SyntaxNode
```
christophe-zurn-sonarsource: There are actually only 2 methods relying all these generic types to be specifc. I suggest to have a single generic `TSyntaxNode` with `where TSyntaxNode : SyntaxNode`, and override the 2 methods left that rely on specific types: `ArgumentLocations` and `IdentifierLocations`, with the logic inside `ArgumentLocations` refactored into helper methods.
WDYT ?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/KnownType.cs
```diff
@@ -1,511 +1,512 @@
...
+        internal static readonly KnownType Microsoft_VisualStudio_TestTools_UnitTesting_TestClassAttribute = new KnownType("Microsoft.VisualStudio.TestTools.UnitTesting.TestClassAttribute");
+        internal static readonly KnownType Microsoft_VisualStudio_TestTools_UnitTesting_TestMethodAttribute = new KnownType("Microsoft.VisualStudio.TestTools.UnitTesting.TestMethodAttribute");
+        internal static readonly KnownType Microsoft_VisualStudio_TestTools_UnitTesting_DataTestMethodAttribute = new KnownType("Microsoft.VisualStudio.TestTools.UnitTesting.DataTestMethodAttribute");
+        internal static readonly KnownType Microsoft_VisualStudio_TestTools_UnitTesting_WorkItemAttribute = new KnownType("Microsoft.VisualStudio.TestTools.UnitTesting.WorkItemAttribute");
+        internal static readonly KnownType Microsoft_VisualStudio_TestTools_UnitTesting_AssemblyCleanupAttribute = new KnownType("Microsoft.VisualStudio.TestTools.UnitTesting.AssemblyCleanupAttribute");
+        internal static readonly K
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: It's here #2812

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CertificateValidationCheck.vb
```diff
@@ -0,0 +1,485 @@
...
+            'VB Specific syntax with return variable, codepaths are not inspected
+            CreateRQ.ServerCertificateValidationCallback = AddressOf FalseNegativeVBSpecific
+            CreateRQ.ServerCertificateValidationCallback = FindFalseNegativeVBSpecific()
+        End Sub
+
+        Private Sub DirectAddHandlers()
+            'Inline version
+            'Secondary@+1
+            CreateRQ().ServerCertificateValidationCallback = Function(sender, certificate, chain, SslPolicyErrors) True    'Noncompliant  {{Enable server certificate validation on this SSL/TLS connection}}
+            '          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                         ^^^^
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I know, it's for the future. It is impossible to keep it at the beginning of the line as it gets automatically aligned with the block. IssueLocationCollector should be updated in #2784 to handle this scenario.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/CertificateValidationCheckTest.cs
```diff
@@ -30,12 +30,23 @@ public class CertificateValidationCheckTest
...
                 additionalReferences : FrameworkMetadataReference.SystemNetHttp.Concat(FrameworkMetadataReference.Netstandard)
                 );
         }
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void CertificateValidationCheck_VB()
+        {
+            Verifier.VerifyAnalyzer(@"TestCases\CertificateValidationCheck.vb",
+                new SonarAnalyzer.Rules.VisualBasic.CertificateValidationCheck(),
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I would keep original version as long as there's only single usage. Two or more will deserve alias just as in `EquivalenceCheckerTest.cs`

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/CertificateValidationCheck.cs
```diff
@@ -0,0 +1,187 @@
...
+
+        internal override KnownType GenericDelegateType()
+        {
+            return KnownType.System_Func_T1_T2_T3_T4_TResult_VB;
+        }
+
+        protected override Location ArgumentLocation(ArgumentSyntax argument)
+        {
+            //For Lambda expression extract location of the parentheses only to separate them from secondary location of "true"
+            return ((argument.GetExpression() is LambdaExpressionSyntax lambda) ? (SyntaxNode)lambda.SubOrFunctionHeader.ParameterList : argument).GetLocation();
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Good point. I've added markers into C# and VB.NET. But it's not working currently due to #2784

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs
pavel-mikula-sonarsource: I would be fine with `IdentifierLocations`. But `ArgumentLocations` contains the core of the algorithm and contains heavy logic. I would rather not to couple this piece of code. When having this and `IdentifierLocations`, it creates two usage places to justify this design.

I consider this a common problem when writing common C# and VB.NET rule. I plan to address in the future update of C# vs. VB.NET rule unification.

PR #2810, opened 2019-11-22
Update *.cs src files to have unix style line ending

costin-zaharia-sonarsource:


costin-zaharia-sonarsource:
I've updated also those.

PR #2818, opened 2019-11-22
Fix S1128 FP: using System.Linq query syntax missing cases

### LINKED ISSUE #2758: S1128 False Positive when using System.Linq
### Description

Visual Studio 2019 had SonarLint automatically upgrade to 4.13.0.11687. Warning S1128 (remove unused usings) false positive with `using System.Linq` and there's a Linq query in the code. It will raise a warning to remove the statement, resulting in a compilation error if removed.

### Repro steps

See the code

```csharp
using System.Linq;

[Test]
public void TestCase {
  List section = new List<string>() {
    "key";
  }
  var copiedKeys = from item in array select item.Key;
  Assert.That(copiedKeys, Has.Member("key"));
}

```

### Expected behavior

That SonarLint is able to know there is a Linq query, and that it does not warn about removing `System.Linq`.

### Actual behavior

The warning S1128 is shown, when it shouldn't be, and it is removed in a bulk change.

### Known workarounds

None

### Related information

* SonarC# Version: Using SonarLint 4.13.0.11687
* Visual Studio Version 2019 version 16.3.6

christophe-zurn-sonarsource:
Fixes #2758

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnnecessaryUsings.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnnecessaryUsings.cs
@@ -146,7 +146,7 @@ namespace LinqQuery1
     {
         public void DoLinqQuery(List<string> myList)
         {
-            var linkQuery = from item in myList select item.GetType();
+            var query = from item in myList select item.GetType();
         }
     }
 }
@@ -158,7 +158,22 @@ namespace LinqQuery2
     {
         public void DoLinqQuery(List<string> myList)
         {
-            var linkQuery = from item in myList select item.GetType();
+            var query = from item in myList select item.GetType();
+        }
+    }
+}
+
+namespace LinqQuery3.System.Linq { }
+
+namespace LinqQuery3
+{
+    using System.Linq; // Noncompliant - This is 'LinqQuery3.System.Linq' whose import is indeed unnecessary
+    using global::System.Linq;
+    class LinqQueryUser
+    {
+        public void DoLinqQuery(List<string> myList)
+        {
+            var query = from item in myList select item.GetType();
         }
     }
 }
@@ -176,7 +191,7 @@ namespace System
     {
         public void DoLinqQuery(List<string> myList)
         {
-            var linkQuery = from item in myList select item.GetType();
+            var query = from item in myList select item.GetType();
         }
     }
 }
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnnecessaryUsings.cs
```diff
@@ -151,6 +151,36 @@ public void DoLinqQuery(List<string> myList)
...
+    using Linq; // Compliant - statement is needed for query syntax
+    class LinqQueryUser
+    {
+        public void DoLinqQuery(List<string> myList)
+        {
+            var linkQuery = from item in myList select item.GetType();
+        }
+    }
+}
+
```
pavel-mikula-sonarsource: Please add test like this to be bullet-proof in the future:

```
namespace Custom.System.Linq
{
    class Coincidence
    {
    }
}

namespace Custom
{
    using System.Linq;  //Noncompliant, it is from Custom.System.Linq and it is not used.
    using global::System.Linq;

    class Usage
    {
        public void DoLinq(List<string> myList)
        {
            var query = from item in myList select item.GetType();
        }
    }
}

```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnnecessaryUsings.cs
```diff
@@ -151,6 +151,36 @@ public void DoLinqQuery(List<string> myList)
...
 }
 
+namespace LinqQuery2
+{
+    using global::System.Linq; // Compliant - statement is needed for query syntax
+    class LinqQueryUser
+    {
+        public void DoLinqQuery(List<string> myList)
+        {
+            var linkQuery = from item in myList select item.GetType();
```
pavel-mikula-sonarsource: Little typo here and on other places.
```suggestion
            var linqQuery = from item in myList select item.GetType();
```
Since Q in LINQ stands for Query, I would suggest just `var query = ...`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnnecessaryUsings.cs
christophe-zurn-sonarsource: test added

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnnecessaryUsings.cs
christophe-zurn-sonarsource: Update name to `query` and `DoQuery` for most. However, I left the namespace names as they are, in order to quickly understand what theses tests are about, even if the name is slightly redundant

PR #2827, opened 2019-11-25
Cbde raise multiple issuekeys

geoffray-adde-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CbdeHandlerRule.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(S2583DiagnosticId)]
+    [Rule(S3949DiagnosticId)]
+    public sealed class CbdeHandlerRule : SonarDiagnosticAnalyzer
+    {
+        CbdeHandler cbdeHandler = new CbdeHandler();
+        private const string S2583DiagnosticId = "S2583";
```
loic-joly-sonarsource: This boilerplate can probably be reduced, but it's good enough for now, let's wait until we have more rules...

geoffray-adde-sonarsource: agreed

PR #2831, opened 2019-11-25
Add linux executable of CBDE to the DotNet plugin

loic-joly-sonarsource:
Refactor the code of DownloadCBDE.targets because I could not see how to make CallTarget work with batched targets, and having a batched target to download CBDE for each OS seemed a nice thing.

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
@@ -9,79 +9,93 @@
   
   -->
   
+    <ItemGroup>
+    <CBDEUrls Include="http://repox.jfrog.io/repox/sonarsource-public-builds/org/sonarsource/cbde/windows/cbde-windows.zip" >
+      <SubFolder>windows\</SubFolder>
+      <MainExe>dotnet-symbolic-execution.exe</MainExe>
+      <ZipFileName>cbde-windows.zip</ZipFileName>
+    </CBDEUrls>
+    <CBDEUrls Include="http://repox.jfrog.io/repox/sonarsource-public-builds/org/sonarsource/cbde/linux/cbde-linux.zip" >
+      <SubFolder>linux\</SubFolder>
+      <MainExe>dotnet-symbolic-execution</MainExe>
+      <ZipFileName>cbde-linux.zip</ZipFileName>
+    </CBDEUrls>
+  </ItemGroup>
+
   <PropertyGroup>
     <CBDEDownloadUrl>http://repox.jfrog.io/repox/sonarsource-public-builds/org/sonarsource/cbde/windows/cbde-windows.zip</CBDEDownloadUrl>
-    <RelativeCBDEFolder>CBDE\windows\</RelativeCBDEFolder>
+    <RelativeCBDEFolder>CBDE\</RelativeCBDEFolder>
     <CBDEFolder>$(MSBuildProjectDirectory)\$(RelativeCBDEFolder)</CBDEFolder>
-    <CBDEZipFileName>cbde-windows.zip</CBDEZipFileName>
-    <CBDEZipFilePath>$(CBDEFolder)$(CBDEZipFileName)</CBDEZipFilePath>
-  </PropertyGroup>
+   </PropertyGroup>
 
   <!-- Specify them here in case they are already present. If they are not, they will be added by the task that fetches the files -->
   <ItemGroup>
-      <Content Include="$(RelativeCBDEFolder)\**\*.exe" >
-        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      </Content>
-      <Content Include="$(RelativeCBDEFolder)\**\*.txt" >
-        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      </Content>
+    <Content Include="$(RelativeCBDEFolder)\windows\**\*.exe" >
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="$(RelativeCBDEFolder)\windows\**\*.txt" >
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="$(RelativeCBDEFolder)\linux\**" Exclude="$(RelativeCBDEFolder)\linux\**\*.zip">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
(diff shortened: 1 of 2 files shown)


loic-joly-sonarsource:
@duncanp-sonar : Since you helped us previously on this subject, I assigned you as a reviewer. If you don't have time/fun doing it, let me know, and I'll find someone else.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
```diff
@@ -9,79 +9,93 @@
...
-
-  <Target Name="DownloadFromRepox">
-    
-    <Message Importance="high" Text="CBDE: Deleting any existing files..." />
+    <Message Importance="high" Text="%(CBDEUrls.SubFolder)/%(CBDEUrls.MainExe): binaries exist? $(CBDEFilesExist)" />
+    <Message Importance="high" Text="%(CBDEUrls.SubFolder)/%(CBDEUrls.MainExe): Updating CBDE binaries" Condition="$(UpdateCBDE)=='true'" />
+    <Message Importance="high" Text="CBDE: Deleting any existing files..." Condition="$(ShouldDownloadCBDE)=='true'"/>
 
     <ItemGroup>
       <ExistingCBDEFiles Include="$(CBDEFolder)\*.*" />
```
duncanp-sonar: Shouldn't this `Include` also specify the `%(CBDEUrls.SubFolder)` to check?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
```diff
@@ -9,79 +9,93 @@
...
     <PropertyGroup>
-      <CBDEFilesExist>$([System.IO.File]::Exists('$(CBDEFolder)dotnet-symbolic-execution.exe'))</CBDEFilesExist>
+      <PowerShellExe Condition="'$(PowerShellExe)'==''">%WINDIR%\System32\WindowsPowerShell\v1.0\powershell.exe</PowerShellExe>
+      <UnzipScript>$(MSBuildProjectDirectory)\..\..\..\Scripts\unzip.ps1</UnzipScript>
+      <DownloadScript>$(MSBuildProjectDirectory)\..\..\..\Scripts\download.ps1</DownloadScript>
+    </PropertyGroup>
+    
+    <PropertyGroup>
+      <CBDEFilesExist>$([System.IO.File]::Exists('$(CBDEFolder)%(CBDEUrls.SubFolder)%(CBDEUrls.MainExe)'))</CBDEFilesExist>
+      <!--<CBDEZipFilePath>$(CBDEFolder)%(CBDEUrls.ZipFileName)</CBDEZipFilePath>-->
```
duncanp-sonar: Commented out line can be removed

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
```diff
@@ -9,79 +9,93 @@
...
-      Command="$(PowerShellExe) -NonInteractive -ExecutionPolicy Unrestricted -File $(DownloadScript) -url $(CBDEDownloadUrl) -outfolder $(CBDEFolder) -outfile $(CBDEZipFileName)">
-      <Output TaskParameter="ConsoleOutput" PropertyName="OutputOfExec" />
+    <Message Importance="high" Text="CBDE: Begin downloading CBDE zip..." Condition="$(ShouldDownloadCBDE)=='true'"/>
+    <Exec ConsoleToMsBuild="true" Condition="$(ShouldDownloadCBDE)=='true'"
+      Command="$(PowerShellExe) -NonInteractive -ExecutionPolicy Unrestricted -File $(DownloadScript) -url @(CBDEUrls) -outfolder $(CBDEFolder)%(CBDEUrls.SubFolder) -outfile %(CBDEUrls.ZipFileName)">
+      <Output TaskParameter="ConsoleOutput" PropertyName="OutputOfExec"/>
     </Exec>
-    <Message Importance="low" Text="Output of powershell download: $(OutputOfExec)" />
-    <Message Importance="high" Text="CBDE: Finished downloading CBDE binaries: $(CBDEZipFilePath)" />
+    <Message Importance="low" Text="Output of powershell download: $(OutputOfExec)"  Condition="$(ShouldDownloadCBDE)=='true'"/>
```
duncanp-sonar: It should be possible to use batching to avoid repeating this condition everywhere, but batching can be a pain to get working properly so if this works I'd leave it as it is.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
```diff
@@ -9,79 +9,93 @@
...
+      Command="$(PowerShellExe) -NonInteractive -ExecutionPolicy Unrestricted -File $(UnzipScript) -zipfile $(CBDEFolder)%(CBDEUrls.SubFolder)%(CBDEUrls.ZipFileName) -outpath $(CBDEFolder)%(CBDEUrls.SubFolder)">
       <Output TaskParameter="ConsoleOutput" PropertyName="OutputOfExec" />
     </Exec>
-    <Message Importance="low" Text="Output of powershell unzip: $(OutputOfExec)" />
+    <Message Importance="low" Text="Output of powershell unzip: $(OutputOfExec)" Condition="$(ShouldDownloadCBDE)=='true'"/>
 
-    <Message Importance="high" Text="CBDE: Finished unzipping CBDE binaries" />
+    <Message Importance="high" Text="CBDE: Finished unzipping CBDE binaries" Condition="$(ShouldDownloadCBDE)=='true'"/>
+
+    <!-- Specifies the CBDE files that should be included in the project and copied to the output directory -->
```
duncanp-sonar: This block was repeated in the prior version so it's not any worse than it was before, but it would be nice if it didn't have to be.
You could try a quick test of creating a new target containing just this `ItemGroup`, with the target specifying `AfterTargets="FetchCBDEBinaries"` and `BeforeTargets="BeforeBuild"`, but if you can't get it work easily I'd leave it as it is.

loic-joly-sonarsource: It tried it, but if I understand you correctly, this target would have to be conditional, with the condition depending on the metadata of an item. And this is not allowed...

duncanp-sonar: It doesn't need to be conditional - the same lines are included unconditionally in 33-41.
The idea would be to have the ItemGroup in a target so that it is defined only once, and then make sure the target is run after the "download binaries" target.

PR #2237, opened 2019-01-15
Fix S4143: AD0001 NullReferenceException

Evangelink:
Fix #2235

PR #2836, opened 2019-11-26
S4143 False positives with `List<T>.Add` method calls

### LINKED ISSUE #2674: S4143 False positives with `List<T>.Add` method calls
### Description

Using the SonarAnalyzer.CSharp nuget, S4143 flags issues for List<String> (and probably any lists)

### Repro steps

Using the code:
```
List<string> S4143 = new List<string>();
S4143.Add("SomeText");
S4143.Add("\n");
S4143.Add("SomeText");
S4143.Add("\n");
```

The last 2 Adds are showing with warnings for S4143, as they are Adds they're extending the list rather than overwriting any existing entries.

### Expected behavior

The above code shouldn't generate a warning.

### Actual behavior

S4143 generates warnings.

### Known workarounds

Set the warning to info or suppress.

### Related information

* SonarC# Version: 7.16.0.8981
* Visual Studio Version: 2017, 15.9.16

pavel-mikula-sonarsource:
Fixes #2674

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteCollectionElements.vb
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteCollectionElements.vb
@@ -69,10 +69,10 @@ Namespace Tests.TestCases
             list.Add("MyText")
         End Sub
 
-        Private Sub ICollectionAdd(list As ICollection(Of String))
+        Private Sub ICollectionAdd(collection As ICollection(Of String))
             '#2674 List.Add method should not raise any issue when used with same elements
-            list.Add("MyText")
-            list.Add("MyText")
+            collection.Add("MyText")
+            collection.Add("MyText")
         End Sub
 
         Private Sub SameIndexSpacedOut(ByVal names() As String)
@@ -109,9 +109,9 @@ Namespace Tests.TestCases
         End Sub
 
         Private Sub IDictionaryAddMixedCase(ByVal dict As IDictionary(Of Integer, Integer))
-            dict.Add(0, 0) ' Secondary
+            dict.add(0, 0) ' Secondary
 '           ^^^^^^^^^^^^^^
-            dict.Add(0, 1) ' Noncompliant
+            dict.ADD(0, 1) ' Noncompliant
 '           ^^^^^^^^^^^^^^
         End Sub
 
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteCollectionElements.vb
```diff
@@ -4,214 +4,232 @@ Imports System.Linq
...
+            dict.add(0, 1) ' Noncompliant
+'           ^^^^^^^^^^^^^^
+        End Sub
+
+        Private Sub IDictionaryAddMixedCase(ByVal dict As IDictionary(Of Integer, Integer))
+            dict.Add(0, 0) ' Secondary
+'           ^^^^^^^^^^^^^^
+            dict.Add(0, 1) ' Noncompliant
+'           ^^^^^^^^^^^^^^
+        End Sub
```
christophe-zurn-sonarsource: This test used to check whether the rule would also work when casing is different than CamelCase. The formatting here removes the original purpose of the test. It should remain as it was originally:
```
      dict.add(0, 0) ' Secondary
'     ^^^^^^^^^^^^^^
      dict.ADD(0, 1) ' Noncompliant
'     ^^^^^^^^^^^^^^
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteCollectionElements.vb
```diff
@@ -4,214 +4,232 @@ Imports System.Linq
...
+
+    End Class
 
     'See https://github.com/SonarSource/sonar-dotnet/issues/2235
     Class NullReferenceReproducer
+
         Sub FooBar()
-            Bar ' AD0001 NullReferenceException in GetFirstArgumentExpression
+            Bar() ' AD0001 NullReferenceException in GetFirstArgumentExpression
```
christophe-zurn-sonarsource: Formatting also removed the purpose of the test here. The method call without parenthesis was added to make sure that rule does not throw `NullReferenceException` when method invocation is done without argument. See PR #2237

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteCollectionElements.vb
pavel-mikula-sonarsource: I've actually fix this already, but haven't noticed that I got autoformatted again :(

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteCollectionElements.vb
pavel-mikula-sonarsource: Same as above

Issue #2834, opened 2019-11-26
Migrate CI to Azure Pipelines


andrei-epure-sonarsource:
### Description

We'll migrate the CI from the internal on-premise Jenkins instance to Azure Pipelines.

```
+---------+         +-----------+          +----------+                   
|Net build|         | MVN build |--------->| MVN UTs  |--------+          
|+ publish|-------->| + publish |-----+    | + SC QG  |        |          
+---------+         +-----------+     |    +----------+        v          
 |   |             +-----------+      |    +---------+         |---------+
 |   |             | Net UTs   |      |--->| MVN ITs | ------> | PROMOTE |
 |   +------------>| + SC QG   |           +---------+         +---------+
 |                 +-----------+--------------------------------^    ^    
 |                +-----------+                                      |    
 +--------------->|  Net ITs  |--------------------------------------+    
                  +-----------+                                           
```

**Dotnet** (@costin-zaharia-sonarsource)
- [x] Make dotnet build work
- [x] Use cache for nuget dependencies
- [x] Run UTs w/ code coverage
- [x] Integrate with sonarcloud for the .net analyzer - for PR
- [ ] Integrate with sonarcloud for the .net analyzer - for Master
- Integrate with sonarcloud for the .net analyzer - for Feature Branch - **treated as PR**
- [x] Test various scenarios for PR integration with sonarcloud - low ccov, issues - #2835
- [x] Set the correct version which includes the build number
- [x] Create nuget packages
~Sign the nuget packages~ #2844
- [x] Make the nuget packages available as build artifacts
- [x] Run the ruling ITs (analyzer only)

**Java** (@andrei-epure-sonarsource)
- [x] Java build for PR
- [x] Java UTs and ccov
- [x] Integrate with sonarcloud for the Java plugins - for PR
- [ ] Integrate with sonarcloud for the Java plugins - for Master
- Integrate with sonarcloud for the Java plugins - for Feature Branch  - **treated as PR**
- [x] Test various scenarios for PR integration with sonarcloud - low ccov, issues
- [x] Push the jars to the private repox
- [x] Use correct build number for private repox (@tomverin )
- [x] Download settings.xml
- [x] Run the Java ITs (Orchestrator)

**Web calls** (@tomverin)
~~After Dotnet and Java ITs, promote the nuget packages to the public repox~~ (no need to have the nugets on repox, having them as build artifacts on Azure Pipelines is fine)
[...]

PR #2837, opened 2019-11-26
[WIP] Add mvn build to sonar-dotnet

andrei-epure-sonarsource:
Relates to #2834


azure-pipelines.yml
```diff
@@ -10,7 +10,7 @@ pool:
   vmImage: 'windows-latest'
 
 variables:
-#  - group: sonarsource-build-variables
+  - group: sonarsource-build-variables
```
andrei-epure-sonarsource: to make this work, I had to check "Allow access to all pipelines" in Pipelines > Library for the variable group (and also for `maven-settings.xml`) - these are being used in templates, and there's some scope differences there, I believe

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/CoverageReportImportSensorTest.java
```diff
@@ -153,6 +155,10 @@ public void execute_coverage_no_main_file() throws IOException {
     context.fileSystem().add(new TestInputFileBuilder("foo", "Foo.cs").setLanguage("cs")
       .setType(Type.TEST).build());
 
+    File f = new File(fooPath);
+    assertThat(f.exists()).isTrue();
```
andrei-epure-sonarsource: the file does not get created on the Azure Pipelines VM and this is why the UT fails

its/src/test/java/com/sonar/it/shared/TestUtils.java
```diff
@@ -80,7 +81,7 @@ public static Location getPluginLocation (String pluginName) {
   public static Build<ScannerForMSBuild> newScanner(Path projectDir) {
     // We need to set the fallback version to run from inside the IDE when the property isn't set
     return ScannerForMSBuild.create(projectDir.toFile())
-      .setScannerVersion(System.getProperty("scannerMsbuild.version", "4.6.1.2049"));
+      .setScannerVersion(System.getProperty("scannerMsbuild.version", "4.8.0.12008"));
```
andrei-epure-sonarsource: I am not sure this is used anywhere... because the one in the pm.xml is actually used!

sonar-csharp-plugin/pom.xml
```diff
@@ -247,7 +247,7 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <maxsize>4400000</maxsize>
```
andrei-epure-sonarsource: should revert, I think I had local problems when building debug version?

sonar-vbnet-plugin/pom.xml
```diff
@@ -247,7 +247,7 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <maxsize>3800000</maxsize>
+                  <maxsize>4800000</maxsize>
```
andrei-epure-sonarsource: should revert

PR #2838, opened 2019-11-26
Fix the `IssueLocationCollector` for rules unit tests

### LINKED ISSUE #2672: Unit tests should fail on unexpected "Noncompliant" comments
Unit tests are based on this kind of comments:
```cs
  foo(); // Noncompliant {{This method is dangerous}}
```
But humans can write them with errors, and in this case, the test should fail:
```cs
  foo(); // Noncompliant {This method is dangerous}
  foo(); // Noncompliant {"This method is dangerous"}
```
One problem to enforce Noncompliant text format is the fee comments at the end. In this case one solution can be to enforce a specific 'free comment separator' like ' - ':
```cs
  foo(); // Noncompliant - false-positive, limitation of our analyzer
```

### LINKED ISSUE #2784: Fix the `IssueLocationCollector` for rules unit tests

christophe-zurn-sonarsource:
Fixes #2784
Fixes #2672


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -224,13 +310,28 @@ private void EnsureOnlyOneElement(IEnumerable<IssueLocation> preciseLocationsOnS
...
+            }
+        }
+
+        private static void ThrowUnexpectedPreciseLocationCount(int  count, int line)
+        {
+            var message = $"Expecting only one precise location per line, found {count} on line {line}. " +
+                @"If you want to specify more than one precise location per line you need to omit the Noncompliant comment:
 internal class MyClass : IInterface1 // there should be no Noncompliant comment
 ^^^^^^^ {{Do not create internal classes.}}
                          ^^^^^^^^^^^ @-1 {{IInterface1 IInterface1 is bad for your health.}}";
```
pavel-mikula-sonarsource: Just a typo
```suggestion
                         ^^^^^^^^^^^ @-1 {{IInterface1 is bad for your health.}}";
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -224,13 +310,28 @@ private void EnsureOnlyOneElement(IEnumerable<IssueLocation> preciseLocationsOnS
...
+            if (remainingLine.Contains("{"))
+            {
+                Execute.Assertion.FailWith("Unexpected '{{' found on line: {0}. Either correctly use the '{{{{message}}}}' " +
+                    "format or remove the curly brace on the line of the expected issue", line.LineNumber);
+            }
+        }
+
+        private static void ThrowUnexpectedPreciseLocationCount(int  count, int line)
+        {
+            var message = $"Expecting only one precise location per line, found {count} on line {line}. " +
```
pavel-mikula-sonarsource: There's probably missing a test for this exception, or I haven't found it.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CertificateValidationCheck.vb
```diff
@@ -21,9 +21,9 @@ Namespace Tests.TestCases
 
         Private Sub DirectAddHandlers()
             'Inline version
-            'Secondary@+1
-            CreateRQ().ServerCertificateValidationCallback = Function(sender, certificate, chain, SslPolicyErrors) True         'Noncompliant  {{Enable server certificate validation on this SSL/TLS connection}}
-            '          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                         ^^^^
+            CreateRQ().ServerCertificateValidationCallback = Function(sender, certificate, chain, SslPolicyErrors) True
+            '          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ {{Enable server certificate validation on this SSL/TLS connection}}
+            '                                                                                                      ^^^^ Secondary@-1
```
pavel-mikula-sonarsource: I've added typo situation like this:
```
        static bool Sec(object sender, X509Certificate certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors)        {
            return true;    //Secondary [x1, x2]
        }
        void Check()        {
            CreateRQ().ServerCertificateValidationCallback += Sec;    //Noncompliant [x2]
            CreateRQ().ServerCertificateValidationCallback += Sec;    //Noncompliant [x2]
        }
```
and I got `Unexpected secondary issue [x2] on line 14 with message ''.` error message. 
I would expect [x1] to be unexpected in the message. Aren't there swapped variables somewhere?

pavel-mikula-sonarsource: Having all the IDs as [x2] passes fine. Is this desired behavior?
```
        static bool Sec(object sender, X509Certificate certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors)        {
            return true;    //Secondary [x2, x2]
        }
        void Check()        {
            CreateRQ().ServerCertificateValidationCallback += Sec;    //Noncompliant [x2]
            CreateRQ().ServerCertificateValidationCallback += Sec;    //Noncompliant [x2]
        }
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -30,23 +30,83 @@
...
+    /// Note that most of the previous patterns can be used together when needed.
+    /// <code>
+    ///     private void MyMethod() // Noncompliant@+1 ^4#7 [MyIssueId] {{Remove this unused private method}}
+    /// </code>
+    /// <code>
+    /// </summary>
     public class IssueLocationCollector
     {
+        private const string EXACT_COLUMN_PATTERN = @"\s*(\^(?<columnStart>[0-9]+)#(?<length>[0-9]+))?";
         private const string OFFSET_PATTERN = @"(?:@(?<sign>[\+|-]?)(?<offset>[0-9]+))?";
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: I think you are write, this seems like a historical error when switching to another version of the regex

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -30,23 +30,83 @@
...
         private const string ISSUE_IDS_PATTERN = @"\s*(\[(?<issueIds>.*)\])*";
         private const string MESSAGE_PATTERN = @"\s*(\{\{(?<message>.*)\}\})*";
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: I don't think it was desired to allow for empty message, as it is optional anyways. I don't see any problem to allow `{` and `}` inside the message itself, as it could be valid to have such characters

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -30,23 +30,83 @@
...
         private const string TYPE_PATTERN = @"(?<issueType>Noncompliant|Secondary)";
         private const string BUILD_ERROR_PATTERN = @"Error";
-        private const string PRECISE_LOCATION_PATTERN = @"\s*(?<position>\^+)\s*";
+        private const string PRECISE_LOCATION_PATTERN = @"\s*(?<position>\^+)(\s+(?<position>\^+))*\s*";
         private const string NO_PRECISE_LOCATION_PATTERN = @"\s*(?<!\^+\s{1})";
         private const string COMMENT_PATTERN = @"(?<comment>\/\/|\')";
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: I think this change would be too much for what it would bring. Updating it means creating an instance depending on the underlying language we are currently analyzing, which imply updating all the places an instance is currently created. Furthermore, it would not completely fix the issue , as having this in C# would still yield a problem.
```
var myString ="Well, //Noncompliant ?";
```
Since the code being analyzed in our unit test rules is decided by us, I don't think we will encounter such a corner case, or a workaround would be simpler

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -30,23 +30,83 @@
...
 
         internal const string ISSUE_LOCATION_PATTERN =
-            COMMENT_PATTERN + "*" + NO_PRECISE_LOCATION_PATTERN + TYPE_PATTERN + OFFSET_PATTERN + ISSUE_IDS_PATTERN + MESSAGE_PATTERN;
+            COMMENT_PATTERN + "+" + NO_PRECISE_LOCATION_PATTERN + TYPE_PATTERN + OFFSET_PATTERN + EXACT_COLUMN_PATTERN + ISSUE_IDS_PATTERN + MESSAGE_PATTERN;
         private const string PRECISE_ISSUE_LOCATION_PATTERN =
-            @"^" + COMMENT_PATTERN + PRECISE_LOCATION_PATTERN + TYPE_PATTERN + "*" + OFFSET_PATTERN + ISSUE_IDS_PATTERN + MESSAGE_PATTERN + "$";
+            COMMENT_PATTERN + PRECISE_LOCATION_PATTERN + TYPE_PATTERN + "?" + OFFSET_PATTERN + ISSUE_IDS_PATTERN + MESSAGE_PATTERN + "$";
         internal const string BUILD_ERROR_LOCATION_PATTERN =
-            COMMENT_PATTERN + NO_PRECISE_LOCATION_PATTERN + BUILD_ERROR_PATTERN + OFFSET_PATTERN + ISSUE_IDS_PATTERN;
+            COMMENT_PATTERN + NO_PRECISE_LOCATION_PATTERN + BUILD_ERROR_PATTERN + OFFSET_PATTERN + EXACT_COLUMN_PATTERN + ISSUE_IDS_PATTERN;
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: I partially agree with you on this point. I agree that it makes sense to rename fields here to comply to PascalCase convention. However, I would still keep `Pattern` suffix, as what is defined here are the string pattern for the regex, not the regex themselves.
I suggest to keep the patterns name relatively the same, but to update the 3 main "patterns" into 3 main regex, that way the difference between the main and partial patterns are more obvious, and we can also precompile the regex in order to optimize rules unit tests performances

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -224,13 +310,28 @@ private void EnsureOnlyOneElement(IEnumerable<IssueLocation> preciseLocationsOnS
...
+                ThrowUnexpectedPreciseLocationCount(preciseLocationsOnSameLine.Count(), preciseLocationsOnSameLine.First().LineNumber);
+            }
+        }
+
+        private static void EnsureNoRemainingCurlyBrace(TextLine line, Match match)
+        {
+            var remainingLine = line.ToString().Substring(match.Index + match.Length);
+            if (remainingLine.Contains("{"))
+            {
+                Execute.Assertion.FailWith("Unexpected '{{' found on line: {0}. Either correctly use the '{{{{message}}}}' " +
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: There is a test in `DiagnosticVerifierTest#UnexpectedRemainingCurlyBrace` checking for this message name.
Formatter used to print string here uses the special `{X}` format for additional parameter. So, the `{` character needs to be escaped by using another `{{` character, hence why the message looks strange, but the actual printed message will be `Unexpected '{' found on line: 5. Either correctly use the '{{message}}' format or remove the curly brace on the line of the expected issue`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -177,6 +243,26 @@ private IEnumerable<IssueLocation> CreateIssueLocations(Match match, int lineNum
             return position.Success ? (int?)position.Length : null;
         }
 
+        private static int? GetColumnStart(Match match)
+        {
+            var columnStart = match.Groups["columnStart"];
+            if (columnStart.Success)
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: Can we discuss about this point ? I'm not sure we will improve readability by adding an helper method that does 2 different things here

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CertificateValidationCheck.vb
christophe-zurn-sonarsource: This seems weird. I would say that 2 primary issues should not have the same id, otherwise it looses the purpose of assigning ids to issues

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: LGTM

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Oh, now I see it. The first two are not parsing the value.
In that case: Can you update those two new methods to use ternary operator as the two old does? That will save 8 lines of code :)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CertificateValidationCheck.vb
christophe-zurn-sonarsource: Actually the message is OK here. Since there is twice [x2] primary raising the issue, there is an additional secondary issue on that line with id [x2] that was unexpected.
As mentioned in the other comment, it should not be allowed to have 2 primary with same id anyway.
But in the following case:
```
        static bool Sec(object sender, X509Certificate certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors)        {
            return true;    //Secondary [x1, x2, x3]
        }
        void Check()        {
            CreateRQ().ServerCertificateValidationCallback += Sec;    //Noncompliant [x2]
            CreateRQ().ServerCertificateValidationCallback += Sec;    //Noncompliant [x3]
        }
```
then we would have a message for expected but not raised issue. I just checked and I ve improved that message so that it now mentions the type and id of the expected issue that was not raised.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -30,23 +30,83 @@
...
     {
+        private const string EXACT_COLUMN_PATTERN = @"\s*(\^(?<columnStart>[0-9]+)#(?<length>[0-9]+))?";
         private const string OFFSET_PATTERN = @"(?:@(?<sign>[\+|-]?)(?<offset>[0-9]+))?";
         private const string ISSUE_IDS_PATTERN = @"\s*(\[(?<issueIds>.*)\])*";
         private const string MESSAGE_PATTERN = @"\s*(\{\{(?<message>.*)\}\})*";
         private const string TYPE_PATTERN = @"(?<issueType>Noncompliant|Secondary)";
         private const string BUILD_ERROR_PATTERN = @"Error";
-        private const string PRECISE_LOCATION_PATTERN = @"\s*(?<position>\^+)\s*";
+        private const string PRECISE_LOCATION_PATTERN = @"\s*(?<position>\^+)(\s+(?<position>\^+))*\s*";
         private const string NO_PRECISE_LOCATION_PATTERN = @"\s*(?<!\^+\s{1})";
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Update:
This is still valid:
Semantic issue: NO_..._PATTERN should not actively match \s* for it's parents. I suggest to add \s* at the beginning of BUILD_ERROR_PATTERN and TYPE_PATTERN to address this issue. And to stick to the pattern where everyone is responsible for it's own leading white space.
```suggestion
        private const string NO_PRECISE_LOCATION_PATTERN = @"(?<!\s*\^+\s{1})";
```
Pattern itself is useful.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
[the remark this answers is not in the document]
christophe-zurn-sonarsource: I'm not sure it would actually happen with code review to have such comments in unit tests, but ok, to be on the safe side, I added a check for this 👍

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -30,23 +30,83 @@
...
-        private const string COMMENT_PATTERN = @"(?<comment>\/\/|\')";
-
-        internal const string ISSUE_LOCATION_PATTERN =
-            COMMENT_PATTERN + "*" + NO_PRECISE_LOCATION_PATTERN + TYPE_PATTERN + OFFSET_PATTERN + ISSUE_IDS_PATTERN + MESSAGE_PATTERN;
-        private const string PRECISE_ISSUE_LOCATION_PATTERN =
-            @"^" + COMMENT_PATTERN + PRECISE_LOCATION_PATTERN + TYPE_PATTERN + "*" + OFFSET_PATTERN + ISSUE_IDS_PATTERN + MESSAGE_PATTERN + "$";
-        internal const string BUILD_ERROR_LOCATION_PATTERN =
-            COMMENT_PATTERN + NO_PRECISE_LOCATION_PATTERN + BUILD_ERROR_PATTERN + OFFSET_PATTERN + ISSUE_IDS_PATTERN;
+        private const string CommentPattern = "(?<comment>//|')";
+        private const string PrecisePositionPattern = @"\s*(?<position>\^+)(\s+(?<invalid>\^+))*\s*";
```
pavel-mikula-sonarsource: The last \s* will be useless, everyone should be responsible for it's own leading white space.
```suggestion
        private const string PrecisePositionPattern = @"\s*(?<position>\^+)(\s+(?<invalid>\^+))*";
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -93,19 +92,18 @@ private static void CompareActualToExpected(IEnumerable<Diagnostic> diagnostics,
...
+                        issueId);
                 }
             }
 
             if (expectedIssues.Count != 0)
             {
-                Execute.Assertion.FailWith($"Issue expected but not raised on line(s) " +
-                    $"{string.Join(",", expectedIssues.Select(i => i.LineNumber))}.");
+                var expectedIssuesDescription = expectedIssues.Select(i => $"{Environment.NewLine}Line: {i.LineNumber}, Type: {IssueType(i.IsPrimary)}, Id: '{i.IssueId}'");
+                Execute.Assertion.FailWith($"Issue(s) expected but not raised on line(s):{expectedIssuesDescription.JoinStr("")}");
```
pavel-mikula-sonarsource: Creating empty strings is not good practice for memory reasons. null is cheap and will work just as good.
```suggestion
                Execute.Assertion.FailWith($"Issue(s) expected but not raised on line(s):{expectedIssuesDescription.JoinStr(null)}");
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -105,41 +175,36 @@ public static bool IsNotIssueLocationLine(string line)
...
             return Enumerable.Empty<IssueLocation>();
         }
 
-        internal /*for testing*/ IEnumerable<IssueLocation> GetPreciseIssueLocations(TextLine line)
+        private IEnumerable<IssueLocation> GetLocations(TextLine line, Regex rx)
         {
             var lineText = line.ToString();
 
-            var match = Regex.Match(lineText, PRECISE_ISSUE_LOCATION_PATTERN);
+            var match = rx.Match(lineText);
```
pavel-mikula-sonarsource: Is the lading empty line needed? `lineText` is used only once, I would remove the variable and put `.ToString()` directly into `.Match(...)`.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -105,41 +175,36 @@ public static bool IsNotIssueLocationLine(string line)
...
+            return GetLocations(line, RxBuildError);
         }
 
-        private IEnumerable<IssueLocation> GetLocations(TextLine line, string pattern)
+        internal /*for testing*/ IEnumerable<IssueLocation> GetPreciseIssueLocations(TextLine line)
         {
             var lineText = line.ToString();
 
-            var match = Regex.Match(lineText, pattern);
+            var match = RxPreciseLocation.Match(lineText);
```
pavel-mikula-sonarsource: Is the lading empty line needed? `lineText` is used only once, I would remove the variable and put `.ToString()` directly into `.Match(...)`.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -208,29 +291,50 @@ private IEnumerable<string> GetIssueIds(Match match)
 
         private int GetOffset(Match match)
         {
-            var sign = match.Groups["sign"];
             var offset = match.Groups["offset"];
-            if (!sign.Success && !offset.Success)
+            return offset.Success ? int.Parse(offset.Value) : 0;
```
pavel-mikula-sonarsource: `GetIssueIds` above: from `[abc, def, g.h.i]` it will return `abc, def, g` and strip the `.h.i` off. 
Why? Is it really needed? If yes, it should be documented in at the beginning of the file.

pavel-mikula-sonarsource: It would be nice to have `GetOffset` above `GetIssueIds` to have the similar methods together.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -208,29 +291,50 @@ private IEnumerable<string> GetIssueIds(Match match)
...
+        }
+
+        private static void EnsureNoDuplicatedPrimaryIds(IList<IIssueLocation> mergedLocations)
+        {
+            var duplicateLocationsIds = mergedLocations
+                            .Where(location => location.IsPrimary)
```
pavel-mikula-sonarsource: This should be little cheaper. Two where's looks suspicious when the second one does not touch the grouped values.
```suggestion
                            .Where(x => x.IsPrimary && x.Key != null)
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -208,29 +291,50 @@ private IEnumerable<string> GetIssueIds(Match match)
...
+                            .GroupBy(x => x.IssueId)
+                            .Where(x => x.Key != null)
+                            .FirstOrDefault(group => group.Count() > 1);
+            if (duplicateLocationsIds != null)
             {
-                return 0;
+                var duplicatedIdLines = duplicateLocationsIds.Select(issueLocation => issueLocation.LineNumber).JoinStr(",");
```
pavel-mikula-sonarsource: I would place a space there for nice looking error message.
```suggestion
                var duplicatedIdLines = duplicateLocationsIds.Select(issueLocation => issueLocation.LineNumber).JoinStr(", ");
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -295,6 +310,16 @@ private int GetOffset(Match match)
             return offset.Success ? int.Parse(offset.Value) : 0;
         }
 
+        private static void EnsureNoInvalidFormat(TextLine line)
+        {
+            var lineText = line.ToString();
+            if (RxInvalidType.IsMatch(lineText) || RxInvalidxPreciseLocation.IsMatch(lineText))
+            {
+                throw new InvalidOperationException($@"Line {line.LineNumber} looks like it contains comment RegEx for noncompliant code, but it is not recognized as one of the expected RegEx.
```
pavel-mikula-sonarsource: I would remove the first word `RegEx` from the message and replace the second one with `Pattern`
It's our internal tooling and it's not related to unit test itself. Can be confusing.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -30,23 +30,83 @@
...
-        internal const string ISSUE_LOCATION_PATTERN =
-            COMMENT_PATTERN + "*" + NO_PRECISE_LOCATION_PATTERN + TYPE_PATTERN + OFFSET_PATTERN + ISSUE_IDS_PATTERN + MESSAGE_PATTERN;
-        private const string PRECISE_ISSUE_LOCATION_PATTERN =
-            @"^" + COMMENT_PATTERN + PRECISE_LOCATION_PATTERN + TYPE_PATTERN + "*" + OFFSET_PATTERN + ISSUE_IDS_PATTERN + MESSAGE_PATTERN + "$";
-        internal const string BUILD_ERROR_LOCATION_PATTERN =
-            COMMENT_PATTERN + NO_PRECISE_LOCATION_PATTERN + BUILD_ERROR_PATTERN + OFFSET_PATTERN + ISSUE_IDS_PATTERN;
+        private const string CommentPattern = "(?<comment>//|')";
+        private const string PrecisePositionPattern = @"\s*(?<position>\^+)(\s+(?<invalid>\^+))*\s*";
+        private const string NoPrecisePositionPattern = @"\s*(?<!\^+\s{1})";
+        private const string IssueTypePattern = "(?<issueType>Noncompliant|Secondary)";
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: This was addressed in ec0decaffb5fa8979b3c02900787fafe124a5c83

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -30,23 +30,83 @@
...
+        private const string ErrorTypePattern = "Error";
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: This was adressed in ec0decaffb5fa8979b3c02900787fafe124a5c83

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -30,23 +30,83 @@
...
-
-        internal const string ISSUE_LOCATION_PATTERN =
-            COMMENT_PATTERN + "*" + NO_PRECISE_LOCATION_PATTERN + TYPE_PATTERN + OFFSET_PATTERN + ISSUE_IDS_PATTERN + MESSAGE_PATTERN;
-        private const string PRECISE_ISSUE_LOCATION_PATTERN =
-            @"^" + COMMENT_PATTERN + PRECISE_LOCATION_PATTERN + TYPE_PATTERN + "*" + OFFSET_PATTERN + ISSUE_IDS_PATTERN + MESSAGE_PATTERN + "$";
-        internal const string BUILD_ERROR_LOCATION_PATTERN =
-            COMMENT_PATTERN + NO_PRECISE_LOCATION_PATTERN + BUILD_ERROR_PATTERN + OFFSET_PATTERN + ISSUE_IDS_PATTERN;
+        private const string CommentPattern = "(?<comment>//|')";
+        private const string PrecisePositionPattern = @"\s*(?<position>\^+)(\s+(?<invalid>\^+))*\s*";
+        private const string NoPrecisePositionPattern = @"\s*(?<!\^+\s{1})";
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: This was addressed in ec0decaffb5fa8979b3c02900787fafe124a5c83

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
christophe-zurn-sonarsource: This should actually not create any new empty string, as empty literals will reuse the interned empty string.
I'd much prefer to keep `expectedIssuesDescription.JoinStr("")` as I find it much more explicit that we are joining using an empty string

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
christophe-zurn-sonarsource: Removing it did not affect anything apart a unit test relying on it, so I removed it as it does not seem necessary

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
pavel-mikula-sonarsource: The interning of he empty string is what I wanted to avoid.
null and "" looks the same to me when the overall result cannot be null.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IssueLocationCollector.cs
```diff
@@ -30,52 +30,117 @@
...
-
-        internal const string ISSUE_LOCATION_PATTERN =
-            COMMENT_PATTERN + "*" + NO_PRECISE_LOCATION_PATTERN + TYPE_PATTERN + OFFSET_PATTERN + ISSUE_IDS_PATTERN + MESSAGE_PATTERN;
-        private const string PRECISE_ISSUE_LOCATION_PATTERN =
-            @"^" + COMMENT_PATTERN + PRECISE_LOCATION_PATTERN + TYPE_PATTERN + "*" + OFFSET_PATTERN + ISSUE_IDS_PATTERN + MESSAGE_PATTERN + "$";
-        internal const string BUILD_ERROR_LOCATION_PATTERN =
-            COMMENT_PATTERN + NO_PRECISE_LOCATION_PATTERN + BUILD_ERROR_PATTERN + OFFSET_PATTERN + ISSUE_IDS_PATTERN;
+        private const string CommentPattern = "(?<comment>//|')";
+        private const string PrecisePositionPattern = @"\s*(?<position>\^+)(\s+(?<invalid>\^+))*";
+        private const string NoPrecisePositionPattern = @"(?<!\s*\^+\s{1})";
```
pavel-mikula-sonarsource: Little cleanup, `{1}` is not needed
```suggestion
        private const string NoPrecisePositionPattern = @"(?<!\s*\^+\s)";
```

christophe-zurn-sonarsource:
Failing unit tests is related to: #2856

PR #2842, opened 2019-11-27
Update orchestrator to 3.27.0.2172

### LINKED ISSUE #2617: update orchestrator to 3.27.0.2172
### Description

See email from @jacek-poreda-sonarsource in which he announced the latest release (_"Release of Orchestrator 3.27.0.2172"_)

In order to update your libraries please use latest sonar-ws API for instance: https://mvnrepository.com/artifact/org.sonarsource.sonarqube/sonar-ws/7.9.1

Use `WsClientFactories` to create new clients. See SonarSource/sonar-enterprise#1936 for sample

### Related information

* SonarC# Version 7.16

christophe-zurn-sonarsource:
Fixes #2617
Use of deprecated methods was already removed in PR #2486
This version update is just a formality, as the removed API were already not used anymore by our codebase

### DIFF
--- a/its/pom.xml
+++ b/its/pom.xml
@@ -41,7 +41,7 @@
     <dependency>
       <groupId>org.sonarsource.sonarqube</groupId>
       <artifactId>sonar-ws</artifactId>
-      <version>7.9.1</version>
+      <version>7.9</version>
     </dependency>
     <dependency>
       <groupId>org.sonarsource.ucfg</groupId>
--- a/its/pom.xml
+++ b/its/pom.xml
@@ -29,7 +29,7 @@
     <dependency>
       <groupId>org.sonarsource.orchestrator</groupId>
       <artifactId>sonar-orchestrator</artifactId>
-      <version>3.26.0.2111</version>
+      <version>3.27.0.2172</version>
     </dependency>
     <dependency>
       <groupId>org.sonarsource.scanner.msbuild</groupId>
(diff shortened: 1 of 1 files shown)


its/pom.xml
```diff
@@ -41,7 +41,7 @@
     <dependency>
       <groupId>org.sonarsource.sonarqube</groupId>
       <artifactId>sonar-ws</artifactId>
-      <version>7.9</version>
+      <version>7.9.1</version>
```
andrei-epure-sonarsource: is there a difference between the 2 versions? I wouldn't update this w/o a need...

I know it might not related to the `minimal version` specified [in this retro](https://discuss.sonarsource.com/t/learn-from-failure-sonarpython-1-15-1-bugfix-release/3291), but the conclusion from that retro
> Therefore, as a rule of thumb, when upgrading to a new LTS version the minimal SQ version of a plugin, it should be upgraded to the “root” version of this LTS and not a patched version.
> As technical release was done but not published publicly, decision was made to do a bugfix to downgrade minimal requirement to 7.9 instead of 7.9.1.

leads me to suggest to keep this version to the "root" version unless a good reason

Having `7.9` for minimal version of SQ and using `7.9.1` for the WS version can lead to confusion

WDYT?

christophe-zurn-sonarsource: Ok. Makes sense to avoid confusion for now, since we don't actually need to update version in the ITs. But it might be needed if/when SQ introduces new APIs that the analyzer wants to use

andrei-epure-sonarsource: in that case, the SQ minimum version will have to be updated as well OR the plugin will have to deal with the two different APIs (to support LTS)

PR #2848, opened 2019-11-28
Fix S4136 FP: should not raise when static methods are grouped…

### LINKED ISSUE #2776: Fix S4136 FP: should not raise when static methods are grouped together
### Description

[from community](https://community.sonarsource.com/t/false-positive-for-s4136/15787)

If we group static methods together, it should not be a code smell.

Currently, it raises an issue - see below

### Repro steps

```
    public class StaticMethodsTogether
    {
        public static void MethodA() // Noncompliant FP - static methods are grouped together, it's ok
        {
            // do nothing
        }

        public static void MethodB()
        {
            // do nothing
        }

        public void MethodA(int i)
        {
            // do nothing
        }
    }
```

### Known workarounds

Group static and non-static overloads together

```
    public class StaticMethodsTogether
    {
        public static void MethodA()
        {
            // do nothing
        }
        public void MethodA(int i)
        {
            // do nothing
        }
        public static void MethodB()
        {
            // do nothing
        }


    }
```

### Related information

* SonarC# Version 8.0

pavel-mikula-sonarsource:
Fixes #2776

### DIFF
--- a/sonaranalyzer-dotnet/its/expected/Automapper/AutoMapper-025B9F93-7953-454B-AF29-E6A55B2AA58F-S4136.json
+++ b/sonaranalyzer-dotnet/its/expected/Automapper/AutoMapper-025B9F93-7953-454B-AF29-E6A55B2AA58F-S4136.json
@@ -2,6 +2,30 @@
 "issues":  [
 {
 "id":  "S4136",
+"message":  "All 'AutoMapperConfigurationException' method overloads should be adjacent.",
+"location":  [
+{
+"uri":  "sources\Automapper\src\AutoMapper\AutoMapperConfigurationException.cs",
+"region":  {
+"startLine":  27,
+"startColumn":  16,
+"endLine":  27,
+"endColumn":  48
+}
+},
+{
+"uri":  "sources\Automapper\src\AutoMapper\AutoMapperConfigurationException.cs",
+"region":  {
+"startLine":  37,
+"startColumn":  16,
+"endLine":  37,
+"endColumn":  48
+}
+}
+]
+},
+{
+"id":  "S4136",
 "message":  "All 'MapFrom' method overloads should be adjacent.",
 "location":  [
 {
--- a/sonaranalyzer-dotnet/its/expected/Automapper/AutoMapper-025B9F93-7953-454B-AF29-E6A55B2AA58F-S4136.json
+++ b/sonaranalyzer-dotnet/its/expected/Automapper/AutoMapper-025B9F93-7953-454B-AF29-E6A55B2AA58F-S4136.json
@@ -2,30 +2,6 @@
 "issues":  [
 {
 "id":  "S4136",
-"message":  "All 'AutoMapperConfigurationException' method overloads should be adjacent.",
-"location":  [
-{
-"uri":  "sources\Automapper\src\AutoMapper\AutoMapperConfigurationException.cs",
-"region":  {
-"startLine":  27,
-"startColumn":  16,
-"endLine":  27,
-"endColumn":  48
-}
-},
-{
-"uri":  "sources\Automapper\src\AutoMapper\AutoMapperConfigurationException.cs",
-"region":  {
-"startLine":  37,
-"startColumn":  16,
-"endLine":  37,
-"endColumn":  48
-}
-}
-]
-},
-{
-"id":  "S4136",
 "message":  "All 'MapFrom' method overloads should be adjacent.",
 "location":  [
 {
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
@@ -19,6 +19,7 @@
  */
 
 using System.Collections.Immutable;
+using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -61,6 +62,15 @@ namespace SonarAnalyzer.Rules.CSharp
                 return methodDeclaration.ExplicitInterfaceSpecifier == null;
             }
             return true;
+        }
+
+        protected override bool IsStatic(MemberDeclarationSyntax member)
+        {
+            if (member is BaseMethodDeclarationSyntax declaration) //Method or Constructor
+            {
+                return declaration.Modifiers.Any(x => x.Kind() == SyntaxKind.StaticKeyword);
+            }
+            return false;
         }
         
         protected override void Initialize(SonarAnalysisContext context)
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
@@ -66,7 +66,8 @@ namespace SonarAnalyzer.Rules.CSharp
 
         protected override bool IsStatic(MemberDeclarationSyntax member)
         {
-            if (member is BaseMethodDeclarationSyntax declaration) //Method or Constructor
+            //Method or Constructor
+            if (member is BaseMethodDeclarationSyntax declaration) 
             {
                 return declaration.Modifiers.Any(x => x.Kind() == SyntaxKind.StaticKeyword);
             }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
@@ -32,7 +32,6 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class MethodOverloadsShouldBeGrouped : MethodOverloadsShouldBeGroupedBase<MemberDeclarationSyntax>
     {
-
         public MethodOverloadsShouldBeGrouped() : base(RspecStrings.ResourceManager) { }
 
         protected override MemberInfo CreateMemberInfo(SyntaxNodeAnalysisContext c, MemberDeclarationSyntax member)
@@ -51,14 +50,13 @@ namespace SonarAnalyzer.Rules.CSharp
             }
             return null;
         }
-        
+
         protected override void Initialize(SonarAnalysisContext context)
         {
             context.RegisterSyntaxNodeActionInNonGenerated(c =>
             {
                 var typeDeclaration = (TypeDeclarationSyntax)c.Node;
                 CheckMembers(c, typeDeclaration.Members);
-                    
             },
             SyntaxKind.ClassDeclaration,
             SyntaxKind.InterfaceDeclaration,
@@ -74,10 +72,7 @@ namespace SonarAnalyzer.Rules.CSharp
             return true;
         }
 
-        private bool IsStatic(BaseMethodDeclarationSyntax declaration)
-        {
-            return declaration.Modifiers.Any(x => x.Kind() == SyntaxKind.StaticKeyword);
-        }
+        private bool IsStatic(BaseMethodDeclarationSyntax declaration) => declaration.Modifiers.Any(x => x.Kind() == SyntaxKind.StaticKeyword);
 
     }
 }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
@@ -18,7 +18,6 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -33,57 +32,26 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class MethodOverloadsShouldBeGrouped : MethodOverloadsShouldBeGroupedBase<MemberDeclarationSyntax>
     {
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+        public MethodOverloadsShouldBeGrouped() : base(RspecStrings.ResourceManager) { }
 
-        protected override DiagnosticDescriptor Rule { get; } = rule;
-
-        protected override bool IsCaseSensitive => true;
-
-        protected override SyntaxToken? GetNameSyntaxNode(MemberDeclarationSyntax member)
+        protected override MemberInfo CreateMemberInfo(SyntaxNodeAnalysisContext c, MemberDeclarationSyntax member)
         {
-            if (member is ConstructorDeclarationSyntax constructorDeclaration)
+            if (!IsValidMemberForOverload(member))
             {
-                return constructorDeclaration.Identifier;
+                return null;
             }
-            else if (member is MethodDeclarationSyntax methodDeclaration)
+            if (member is ConstructorDeclarationSyntax constructor)
             {
-                return methodDeclaration.Identifier;
+                return new MemberInfo(c, member, constructor.Identifier, IsStatic(constructor), false, true);
+            }
+            else if (member is MethodDeclarationSyntax method)
+            {
+                return new MemberInfo(c, member, method.Identifier, IsStatic(method), method.Modifiers.Any(x => x.Kind() == SyntaxKind.AbstractKeyword), true);
             }
             return null;
         }
         
-        protected override bool IsValidMemberForOverload(MemberDeclarationSyntax member)
-        {
-            if (member is MethodDeclarationSyntax methodDeclaration)
-            {
-                return methodDeclaration.ExplicitInterfaceSpecifier == null;
-            }
-            return true;
-        }
-
-        protected override bool IsStatic(MemberDeclarationSyntax member)
-        {
-            //Method or Constructor
-            if (member is BaseMethodDeclarationSyntax declaration) 
-            {
-                return declaration.Modifiers.Any(x => x.Kind() == SyntaxKind.StaticKeyword);
-            }
-            return false;
-        }
-
-        protected override bool IsAbstract(MemberDeclarationSyntax member)
-        {
-            //Method only. Constructor cannot be abstract
-            if (member is MethodDeclarationSyntax declaration)
-            {
-                return declaration.Modifiers.Any(x => x.Kind() == SyntaxKind.AbstractKeyword);
-            }
-            return false;
-        }
-
         protected override void Initialize(SonarAnalysisContext context)
         {
             context.RegisterSyntaxNodeActionInNonGenerated(c =>
@@ -96,5 +64,20 @@ namespace SonarAnalyzer.Rules.CSharp
             SyntaxKind.InterfaceDeclaration,
             SyntaxKind.StructDeclaration);
         }
+
+        private bool IsValidMemberForOverload(MemberDeclarationSyntax member)
+        {
+            if (member is MethodDeclarationSyntax methodDeclaration)
+            {
+                return methodDeclaration.ExplicitInterfaceSpecifier == null;
+            }
+            return true;
+        }
+
+        private bool IsStatic(BaseMethodDeclarationSyntax declaration)
+        {
+            return declaration.Modifiers.Any(x => x.Kind() == SyntaxKind.StaticKeyword);
+        }
+
     }
 }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
@@ -73,7 +73,17 @@ namespace SonarAnalyzer.Rules.CSharp
             }
             return false;
         }
-        
+
+        protected override bool IsAbstract(MemberDeclarationSyntax member)
+        {
+            //Method only. Constructor cannot be abstract
+            if (member is MethodDeclarationSyntax declaration)
+            {
+                return declaration.Modifiers.Any(x => x.Kind() == SyntaxKind.AbstractKeyword);
+            }
+            return false;
+        }
+
         protected override void Initialize(SonarAnalysisContext context)
         {
             context.RegisterSyntaxNodeActionInNonGenerated(c =>
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/EnumerableExtensions.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/EnumerableExtensions.cs
@@ -77,7 +77,7 @@ namespace SonarAnalyzer.Helpers
         {
             return enumerable.Where(x => x.HasValue).Select(x => x.Value);
         }
-        
+
         /// <summary>
         /// Applies a specified function to the corresponding elements of two sequences,
         /// producing a sequence of the results. If the collections have different length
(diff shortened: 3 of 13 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -151,19 +151,39 @@ class J
     // https://github.com/SonarSource/sonar-dotnet/issues/2776
     public class StaticMethodsTogether
     {
```
andrei-epure-sonarsource: would it be useful to also add tests with fields and properties?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodOverloadsShouldBeGrouped.vb
```diff
@@ -134,14 +134,37 @@ Namespace Tests.TestCases
 
     ' https://github.com/SonarSource/sonar-dotnet/issues/2776
     Public Class StaticMethodsTogether
```
andrei-epure-sonarsource: would it be useful to add `Func` tests as well?

andrei-epure-sonarsource:
also, QA is failing....

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -81,28 +83,29 @@ protected void CheckMembers(SyntaxNodeAnalysisContext c, IEnumerable<TMemberDecl
...
             var misplacedOverloads = new Dictionary<string, List<TMemberDeclarationSyntax>>();
-            string previousMemberName = null;
+            string previousKey = null, key;
             foreach (var member in members)
             {
                 if (GetMethodName(member, IsCaseSensitive) is string methodName
                     && IsValidMemberForOverload(member))
                 {
-                    if (misplacedOverloads.TryGetValue(methodName, out var currentList))
+                    key = methodName + "/" + IsStatic(member);  //#4136 Should not raise when static methods are grouped together
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: The key generation itself is little weird on first look and not so clear to understand. So it deserves a comment. Adding quick # reference is free and effective way to explain all the details about it. If any part goes out of date, code and comment (even without # ) will need update.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -81,28 +83,29 @@ protected void CheckMembers(SyntaxNodeAnalysisContext c, IEnumerable<TMemberDecl
         protected IDictionary<string, List<TMemberDeclarationSyntax>> GetMisplacedOverloads(IEnumerable<TMemberDeclarationSyntax> members)
         {
             var misplacedOverloads = new Dictionary<string, List<TMemberDeclarationSyntax>>();
-            string previousMemberName = null;
+            string previousKey = null, key;
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I would stay with the `key`. The only purpose of it is beeing a key for that dictionary and nothing else. No information can be (easily a reasonably) extracted from it.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodOverloadsShouldBeGrouped.cs
pavel-mikula-sonarsource: The rule is not implemented for fields and properties due to RSPEC:
> Method overloads should be grouped together

Interleaving by fields is tested above.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodOverloadsShouldBeGrouped.vb
pavel-mikula-sonarsource: I'm sorry, I don't understand this point. Do you mean like this?
`public static int FunctionA() => 0;`

andrei-epure-sonarsource: yes, `Function` in addition to `Sub`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -62,6 +63,15 @@ protected override bool IsValidMemberForOverload(MemberDeclarationSyntax member)
             }
             return true;
         }
+
+        protected override bool IsStatic(MemberDeclarationSyntax member)
+        {
+            if (member is BaseMethodDeclarationSyntax declaration) //Method or Constructor
```
andrei-epure-sonarsource: [minor] we all agreed to follow Microsoft coding conventions unless we explicitly defined other conventions. 

> We’re all aligned to follow the Microsoft convention
([notes](https://docs.google.com/document/d/1W40QrhWslS-8dxx7Fu316qv6cLy1si9XQS3RgyVyDQE/edit)) 

Please read [Commenting Conventions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions#commenting-conventions) (which are aligned with existing SonarSource conventions) and put the comment on a separate line, with a space after the `//`

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: fair point , this info is indeed useful

[minor] please put the comment on a separate line with a space after `//` (to follow Microsoft conventions as decided within the team)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
andrei-epure-sonarsource: [minor] comment on separate line with space after `//` (Microsoft convention followed in team)

pavel-mikula-sonarsource:
One more issue: Shouldn't there be the same tweak for `abstract` as well?

I've just got the same when `abstract` is defined above constructor and it's overload is somewhere in the code below.

andrei-epure-sonarsource:
@pavel-mikula-sonarsource - indeed I think it makes sense to have the same exception for abstract methods grouped together

pavel-mikula-sonarsource:
Code has been updated and it's ready for review. This update killed even more noise in the ITs than the IsStatic.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -82,6 +82,25 @@ private bool IsStaticStatement(MethodBaseSyntax statement)
         {
             return statement.DescendantTokens().Any(x => x.Kind() == SyntaxKind.SharedKeyword);
         }
+        
+        protected override bool IsAbstract(StatementSyntax member)
```
christophe-zurn-sonarsource: According to what we decided as coding style, these 2 methods are not in the apporpriate place due to their visibility accessors

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -90,11 +82,12 @@ protected void CheckMembers(SyntaxNodeAnalysisContext c, IEnumerable<TMemberDecl
                 if (GetMethodName(member, IsCaseSensitive) is string methodName
                     && IsValidMemberForOverload(member))
                 {
-                    // #4136 Should not raise when static methods are grouped together
-                    key = methodName + "/" + IsStatic(member);  
+                    // Groups that should be together are defined by accessibility, abstract, static and member name #4136
+                    var accessibility = c.SemanticModel.GetDeclaredSymbol(member)?.DeclaredAccessibility.ToString();
+                    key = $"{accessibility}/{IsStatic(member)}/{IsAbstract(member)}/{methodName}";
```
christophe-zurn-sonarsource: I believe it would be better now to use a dedicated utility class than a simple string as key for grouping misplaced overloads, as it is now comprised of many different attributes.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -49,38 +49,30 @@ protected StringComparison CaseSensitivity
...
         protected abstract bool IsStatic(TMemberDeclarationSyntax member);
 
+        protected abstract bool IsAbstract(TMemberDeclarationSyntax member);
+
         protected void CheckMembers(SyntaxNodeAnalysisContext c, IEnumerable<TMemberDeclarationSyntax> members)
         {
-            var misplacedOverloadsMapping = GetMisplacedOverloads(members);
-            foreach (var misplacedOverloadsEntry in misplacedOverloadsMapping)
+            var misplacedOverloads = GetMisplacedOverloads(c, members);
+            foreach (var misplacedMethods in misplacedOverloads.Values.Where(x => x.Count > 1))
```
christophe-zurn-sonarsource: I think the filtering `misplacedOverloads.Values.Where(x => x.Count > 1))` can already be done in the `GetMisplacedOverloads` method, as don't actually use the dictionary key here anyway

andrei-epure-sonarsource: in addition, maybe use `X.Any()` instead of `x.Count > 1` ?

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -82,6 +82,25 @@ private bool IsStaticStatement(MethodBaseSyntax statement)
...
+                return IsMustInheritStatement(methodDeclaration.SubOrFunctionStatement);
+            }
+            else if (member is MethodStatementSyntax methodStatement)
+            {
+                return IsMustInheritStatement(methodStatement);
+            }
+            return false;
+        }
+
+        private bool IsMustInheritStatement(MethodBaseSyntax statement)
```
andrei-epure-sonarsource: [minor] this can be an arrow function

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -49,38 +49,30 @@ protected StringComparison CaseSensitivity
...
-                                properties: secondaryLocations.ToProperties(),
-                                messageArgs: GetMethodName(firstMethodSignature, true)));
-
-                    }
-                }
+                var firstMethod = misplacedMethods.First();
+                var secondaryLocations = misplacedMethods
+                    .Skip(1)
+                    .Select(member => GetNameSyntaxNode(member))
+                    .Where(nameSyntax => nameSyntax != null)
```
andrei-epure-sonarsource: we have an extension method `WhereNotNull`

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -49,38 +49,30 @@ protected StringComparison CaseSensitivity
...
+                    .Select(nameSyntax => new SecondaryLocation(nameSyntax?.GetLocation(), "Non-adjacent overload"));
```
andrei-epure-sonarsource: does it make sense to create `null` locations? if `nameSyntax` is null, the result will be null.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -49,38 +49,30 @@ protected StringComparison CaseSensitivity
...
+                c.ReportDiagnosticWhenActive(
+                    Diagnostic.Create(
+                        descriptor: Rule,
+                        location: GetNameSyntaxNode(firstMethod)?.GetLocation(),
```
andrei-epure-sonarsource: the same here - what if the result of `GetNameSyntaxNode` is `null`?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
pavel-mikula-sonarsource: I consider the `key` as trivial "hash" for grouping. IEqualityComparer and GetHashCode (with the same concatenation anyway) seems like a lot of extra code and state handling in order to do the same thing anyway.

What I actually don't like in this line is too much concatenation with the same separator. That could be addressed like this:
`key = new object[] { accessibility, IsStatic(member), IsAbstract(member), methodName }.JoinStr("/", x => x?.ToString());`

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
pavel-mikula-sonarsource: `X.Any()` is for `x.Count <>0`. But `GetMisplacedOverloads` actually adds first item as legal primary location and every other item as illegal location. Only those with count>1 are illegal then.

I've updated the `GetMisplacedOverloads` to do this check as Chrisophe suggested, because it makes much more sense. In old version even well placed where in the collection.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
pavel-mikula-sonarsource: The existing one did not work for struct?, I've added an overload.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
pavel-mikula-sonarsource: No, it doesn't. nameSyntax was nullable in old version and it couldn't get there anyway because of the null check. New WhereNotNull deals with that.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
pavel-mikula-sonarsource: I changed my mind. In bigger picture, it indeed is better to create dedicated class.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -96,5 +64,20 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            if (member is MethodDeclarationSyntax methodDeclaration)
+            {
+                return methodDeclaration.ExplicitInterfaceSpecifier == null;
+            }
+            return true;
+        }
+
+        private bool IsStatic(BaseMethodDeclarationSyntax declaration)
+        {
+            return declaration.Modifiers.Any(x => x.Kind() == SyntaxKind.StaticKeyword);
```
andrei-epure-sonarsource: _minor_ can be an arrow function

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/EnumerableExtensions.cs
```diff
@@ -72,6 +72,12 @@ public static IEnumerable<T> WhereNotNull<T>(this IEnumerable<T> enumerable)
             return enumerable.Where(e => e != null);
         }
 
+        public static IEnumerable<T> WhereNotNull<T>(this IEnumerable<Nullable<T>> enumerable)
+                where T : struct
+        {
+            return enumerable.Where(x => x.HasValue).Select(x => x.Value);
```
andrei-epure-sonarsource: _minor_ can be an arrow function

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -24,99 +24,111 @@
...
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 
 namespace SonarAnalyzer.Rules
 {
     public abstract class MethodOverloadsShouldBeGroupedBase<TMemberDeclarationSyntax> : SonarDiagnosticAnalyzer
         where TMemberDeclarationSyntax : SyntaxNode
     {
+
```
andrei-epure-sonarsource: _nitpick_ is this space needed?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -24,99 +24,111 @@
...
+            private readonly bool isCaseSensitive;
+
+            public MemberInfo(SyntaxNodeAnalysisContext context, TMemberDeclarationSyntax member, SyntaxToken nameSyntax, bool isStatic, bool isAbstract, bool isCaseSensitive)
+            {
+                Member = member;
+                Accessibility = context.SemanticModel.GetDeclaredSymbol(member)?.DeclaredAccessibility.ToString();
+                NameSyntax = nameSyntax;
+                IsStatic = isStatic;
+                IsAbstract = isAbstract;
+                this.isCaseSensitive = isCaseSensitive;
```
andrei-epure-sonarsource: _minor_ trailing spaces here (there's a VS pluging Whitespace Visualizer that helps with that)

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -24,99 +24,111 @@
...
+            {
+                Member = member;
+                Accessibility = context.SemanticModel.GetDeclaredSymbol(member)?.DeclaredAccessibility.ToString();
+                NameSyntax = nameSyntax;
+                IsStatic = isStatic;
+                IsAbstract = isAbstract;
+                this.isCaseSensitive = isCaseSensitive;                
+            }
+            
+            public override bool Equals(object obj)
```
andrei-epure-sonarsource: nice, I like this much better! 👍

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -113,5 +63,10 @@ protected override void Initialize(SonarAnalysisContext context)
             SyntaxKind.InterfaceBlock,
             SyntaxKind.StructureBlock);
         }
+
+        private bool IsStaticStatement(MethodBaseSyntax statement) => statement.DescendantTokens().Any(x => x.Kind() == SyntaxKind.SharedKeyword);
+
+        private bool IsMustInheritStatement(MethodBaseSyntax statement) => statement.DescendantTokens().Any(x => x.Kind() == SyntaxKind.MustOverrideKeyword);
```
andrei-epure-sonarsource: _nitpick_ - would it be clearer to name this `IsAbstract` ? afterall, semantically it would be the same, right? it would make the code more uniform and readable

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -33,57 +32,26 @@ namespace SonarAnalyzer.Rules.CSharp
...
+            else if (member is MethodDeclarationSyntax method)
             {
-                return declaration.Modifiers.Any(x => x.Kind() == SyntaxKind.AbstractKeyword);
+                return new MemberInfo(c, member, method.Identifier, IsStatic(method), method.Modifiers.Any(x => x.Kind() == SyntaxKind.AbstractKeyword), true);
             }
-            return false;
+            return null;
         }
-
+
```
andrei-epure-sonarsource: minor trailing spaces

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -24,99 +24,111 @@
...
+                    && NameSyntax.ValueText.Equals(info.NameSyntax.ValueText, isCaseSensitive && info.isCaseSensitive ? StringComparison.InvariantCulture : StringComparison.InvariantCultureIgnoreCase)
+                    && IsStatic == info.IsStatic
+                    && IsAbstract == info.IsAbstract;
             }
-            else
+
+            public override int GetHashCode()
             {
-                return nameSyntaxNode?.ValueText?.ToLowerInvariant();
+                return NameSyntax.ValueText.ToLowerInvariant().GetHashCode();
```
christophe-zurn-sonarsource: It seems strange to have the hashcode being generated from a single field, when the `Equals` method is using 4 of them

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/EnumerableExtensions.cs
pavel-mikula-sonarsource: I find it difficult to read with the `where` clause for optical reasons.
But I've applied this on the JoinStr below that I've recently added.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
pavel-mikula-sonarsource: No, it's not. I just like it :)

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
pavel-mikula-sonarsource: I wanted them to end up in same bucket and let Equals to deal with the rest. I didn't want to duplicate the full Equals/Grouping logic here with that ugly and long concatenation for generating the key.

christophe-zurn-sonarsource:
LGTM !
Can you please rebase your branch on current master, to make sure the build and all ITs are green

pavel-mikula-sonarsource:
@andrei-epure-sonarsource 
I had to rebase in order to fix the UT nuget failure #2856. 
I'm rebasing using Team Explorer (fetch/pull on master, rebase onto master, sync). 

Can you please resolve the resolved issues? I'd like to be sure that I haven't missed/skipped anything.

andrei-epure-sonarsource:
@pavel-mikula-sonarsource the changes LGMT from my side and once you fix the commit history this PR is good to be merged

sonaranalyzer-dotnet/its/expected/Automapper/AutoMapper-025B9F93-7953-454B-AF29-E6A55B2AA58F-S4136.json
```diff
@@ -2,6 +2,30 @@
 "issues":  [
 {
 "id":  "S4136",
+"message":  "All 'AutoMapperConfigurationException' method overloads should be adjacent.",
```
christophe-zurn-sonarsource: These new issues in the IT actually look a bit suspicious. I think it's a new kind of FP

christophe-zurn-sonarsource:
While rebasing on master, we noticed suspicious new issues in ITs. This should be investigated before merging

PR #2851, opened 2019-11-29
Fix AbstractMethodParameterLookup.TryGetArgumentSyntax to handle param/ParamArray properly

### LINKED ISSUE #2815: Fix AbstractMethodParameterLookup.TryGetArgumentSyntax to handle param/ParamArray properly
### Description

`AbstractMethodparameterLookup.TryGetSymbolParameter` and `AbstractMethodparameterLookup.TryGetArgumentSyntax` do not handle param/ParamArray situation correctly. Only first argument is returned, but array should be returned instead.

### Solution

Existing methods should return array to all callers. 
Renamed methods should return single value only when symbol is not param/ParamArray and symbol type is checked by the caller.

pavel-mikula-sonarsource:
Fixes #2815

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/MethodParameterLookupTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/MethodParameterLookupTest.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Linq;
 using System.Collections;
 using FluentAssertions;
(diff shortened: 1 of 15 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/MethodParameterLookupTest.cs
```diff
@@ -36,7 +35,6 @@ namespace SonarAnalyzer.UnitTest.Helpers
     public class MethodParameterLookupTest
     {
 
-
         [TestMethod]
         public void TestMethodParameterLookup_CS()
```
christophe-zurn-sonarsource: Could you please add a test, checking that `TryGetNonParamsSyntax` will correctly throw an exception when used on an actual params parameter symbol.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/MethodParameterLookupTest.cs
```diff
@@ -198,15 +198,24 @@ private void InspectTryGetSymbolParametr(AbstractMethodParameterLookup<TArgument
                 }
```
christophe-zurn-sonarsource: Error message on line 196 is not in snyc with method naming anymore

PR #2853, opened 2019-12-02
README.md update with description of issues.md

pavel-mikula-sonarsource:


docs/issues.md
```diff
@@ -0,0 +1,18 @@
+# Issue tracking
+
+All issues for SonarC\# and SonarVB are tracked using [GitHub Issues](https://github.com/SonarSource/sonar-dotnet/issues).
+
+You're welcome to submit any issue that you've found using these plugins. Please provide minimal reproducible example for each case.
```
andrei-epure-sonarsource:
```suggestion
You're welcome to submit any issue that you've found using the C# and VB.NET code analyzers (as [standalone nugets](https://www.nuget.org/packages/SonarAnalyzer.CSharp/), inside [SonarLint for Visual Studio](https://www.sonarlint.org/visualstudio/), as [SonarQube](https://www.sonarqube.org/) plugins or in [SonarCloud](https://sonarcloud.io/about) analysis).

Please provide a minimal reproducible example in each case.
```

andrei-epure-sonarsource: (note: in the above change suggestions there are two sides
- adding the promotion :) and mentioning the code analyzers rather than the plugins
- small grammar fix "Please provide **a** minimal reproducible example"

docs/issues.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Labels
+
+Labels are divided into several categories. Single issue can contain:
```
andrei-epure-sonarsource:
```suggestion
Labels are divided into several categories. One issue can be labelled with:
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UsingCookiesTest.cs
```diff
@@ -33,11 +33,11 @@ namespace SonarAnalyzer.UnitTest.Rules
     [TestClass]
     public class UsingCookies
     {
-        private static IEnumerable<MetadataReference> GetAdditionalReferences_NetCore() =>
+        private static IEnumerable<MetadataReference> GetAdditionalReferences_NetCore(string packageVersion) =>
```
andrei-epure-sonarsource: not related change

PR #2858, opened 2019-12-04
Add coding style README

christophe-zurn-sonarsource:


CODING_STYLE.md
```diff
@@ -0,0 +1,39 @@
...
+## Class Members
+
+Members and types should always have the lowest possible visibility.
+
+Ordering of class members should be the following:
+1. Constants
+2. Fields
+3. Abstract members
+4. Constructors
+5. Methods and Properties
```
andrei-epure-sonarsource: I would say
5. Properties
6. Methods
7. Private nested classes

CODING_STYLE.md
```diff
@@ -0,0 +1,39 @@
...
+6. Private nested classes
+
+Furthermore, each of these categories should be ordered from higher to lower accessibility level (public, internal, protected, private).
+
+## FIXME and ToDo
+
+`FIXME` should only be used in the code during development as a temporary reminder that there is still work to
```
andrei-epure-sonarsource: _nitpick_ you could use two bullet points for `FIXME` and `ToDo`

CODING_STYLE.md
```diff
@@ -0,0 +1,39 @@
...
+be done here. A `FIXME` should never appear in a PR or be merged into master.
+`ToDo` can be used to mark part of the code that will need to be updated at a later time. It can be used to
```
andrei-epure-sonarsource: I would also add something like `ideally, a ToDo comment should be followed by an issue number (what needs to be done should be in the github issues)`

CODING_STYLE.md
```diff
@@ -0,0 +1,39 @@
+# Coding Style
```
pavel-mikula-sonarsource: I suggest tomove this file to `/docs/` dir. It's a good practice (in OS as well as DEV) to have the root as clean as possible for maintanability.

I'd like to move the CODE_OF_CONDUCT and CONTRIBUTING there as well, but I'm not sure if it is possible due to external references.

CODING_STYLE.md
```diff
@@ -0,0 +1,39 @@
...
+
+When contributing to the project, and if otherwise not mentioned in this document, our coding conventions
+follow the Microsoft [C# Coding Conventions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions)
+and standard [Naming Guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines).
+
+## Class Members
+
+Members and types should always have the lowest possible visibility.
+
+Ordering of class members should be the following:
```
pavel-mikula-sonarsource: Using [live preview](https://markdownlivepreview.com/), this section does not work and get's collapsed to single paragraph. There should be empty line before the first item.

CODING_STYLE.md
```diff
@@ -0,0 +1,39 @@
...
+1. Constants
+2. Fields
```
pavel-mikula-sonarsource: I suggest to use build-in `1.` auto increment functionality for every item for better maintainability.
```
1. Constants
1. Fields
1. Abstract members
1. Constructors
1. Methods and Properties
1. Private nested classes

```

CODING_STYLE.md
pavel-mikula-sonarsource: There should be an empty line before `ToDo` in order to separate paragraph. Current version gets collapsed into one long paragraph.

CODING_STYLE.md
pavel-mikula-sonarsource: I'd like to be explicit about the double colon, because `FIXME This is not finished` does work as well and it's not nice/consistent. I would keep the title without it, but the text full of it.
```suggestion
`FIXME:` should only be used in the code during development as a temporary reminder that there is still work to
```

CODING_STYLE.md
```diff
@@ -0,0 +1,39 @@
...
+## Regions
+
+Generally, as we do not want to have classes that are too long, regions are not necessary and should be avoided.
+It can still be used when and where it makes sense. For instance, when a class having a specific purpose is
+implementing generic interfaces (such as `IComparable`, `IDisposable`), it can make sense to have regions 
+for the implementation of these interfaces.
+
+## VB.NET Specifics
+
+Empty lines should be used between blocks and regions to improve readability.
```
pavel-mikula-sonarsource: As well as around Namespace/Class/End Class/End Namespace statements.

CODING_STYLE.md
```diff
@@ -0,0 +1,39 @@
+# Coding Style
+
+## General
+
+When contributing to the project, and if otherwise not mentioned in this document, our coding conventions
```
pavel-mikula-sonarsource: I've proposed discussion about on-the-line comments into How we work not scheduled discussion. I'd like to allow it (and therefore mention it here). I found them practical and they appear a lot in test cases by-design (Noncompliant, Secondary).

Should we have a quick team discussion first? Or are we going to update that document later (little overhead)?

CODING_STYLE.md
christophe-zurn-sonarsource: Also added bullet points as suggested by Andrei

CODING_STYLE.md
christophe-zurn-sonarsource: I'd rather not to be that strict here. I personally never put colons in FIXME comments and I think being that strict for code that is under work and should theoretically not be reviewed and merged to master is a bit overkill

PR #2875, opened 2019-12-10
Add code analyzers for roslyn API usage

andrei-epure-sonarsource:
Add `Microsoft.CodeAnalysis.Analyzers` - to detect correct usage of the Roslyn APIs

You can see the rules it has here https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md


andrei-epure-sonarsource:
actually currently the results are quite noisy , but we should discuss it

PR #2876, opened 2019-12-10
Reduce code duplication in tests by extracting createStartStep

costin-zaharia-sonarsource:


its/src/test/java/com/sonar/it/csharp/AutoGeneratedTest.java
```diff
@@ -95,11 +94,7 @@ private void analyzeCoverageTestProject(String projectName, boolean analyzeGener
 
       Path projectDir = Tests.projectDir(temp, projectName);
 
-      ScannerForMSBuild beginStep = TestUtils.newScanner(projectDir)
-        .addArgument("begin")
-        .setProjectKey(projectName)
-        .setProjectName(projectName)
-        .setProjectVersion("1.0");
+      ScannerForMSBuild beginStep = TestUtils.createStartStep(projectName, projectDir);
```
andrei-epure-sonarsource: I suggest to use `createBeginStep` to keep S4MSB terminology

its/src/test/java/com/sonar/it/shared/TestUtils.java
```diff
@@ -82,7 +82,27 @@ public static ScannerForMSBuild newEndStep(Path projectDir) {
...
+      .setProjectName(projectName)
+      .setProjectVersion("1.0")
+      .setProperty("sonar.projectBaseDir", getProjectBaseDir(projectDir, subProjectName));
+  }
+
+  private static String getProjectBaseDir(Path projectDir, String subProjectName){
+    if (subProjectName.isEmpty()) {
+      return projectDir.toString();
+    }
+    return projectDir.toString() + File.separator + subProjectName;
```
andrei-epure-sonarsource: _nitpick (feel free to ignore)_: a more elegant solution would be to call `projectDir.resolve(subProjectName)` and `toString` the result, see https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#resolve-java.lang.String-

costin-zaharia-sonarsource: That's actually great. I was looking for something similar but I didn't knew how to do this in Java.

PR #2879, opened 2019-12-10
Coding style - Comments on lines with code

pavel-mikula-sonarsource:


docs/coding-style.md
```diff
@@ -22,6 +22,12 @@ Ordering of class members should be the following:
 
 Furthermore, each of these categories should be ordered from higher to lower accessibility level (public, internal, protected, private).
 
+## Comments
+
+* Comments should be on separate lines.
+
+* Comments on the same line with code are acceptable for short lines of code and short comments.
```
christophe-zurn-sonarsource: These 2 statements are directly contradictory, which makes it strange for a guide. I'd use a `generally` (`Comments should generally ...`), and also mention the fact that comments should be avoided as much as possible by using meaningful variable and method names

PR #2884, opened 2019-12-11
Update RSPEC metadata

pavel-mikula-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.Utilities/Rules.Description/S4426.html
```diff
@@ -1,7 +1,84 @@
...
+      <td> x </td>
+      <td> x </td>
+      <td> x </td>
+      <td> AES-128, 192, 256<br><br>This rule will not raise issues for ciphers that are considered weak (no matter the key size) like
+      <code>DES</code>, <code>Blowfish</code>.<br><br>h3. Notation<br>* DSA ([Digital Signature Algorithm</td>
+      <td>https://en.wikipedia.org/wiki/Digital_Signature_Algorithm]): <code>p</code> is key length and <code>q</code> the modulus length<br>* EC
+      ([Elliptic-curve</td>
+      <td>https://en.wikipedia.org/wiki/Elliptic-curve_cryptography])<br></td>
+      <td></td>
+      <td>EC parameters</td>
```
christophe-zurn-sonarsource: Changes in this file should be reverted, as the html generation is wrong here. What's more, new description does not fit current implementation anyways

sonaranalyzer-dotnet/rspec/cs/S4426_c#.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Cryptographic keys should not be too short",
+  "title": "Cryptographic keys should be robust",
```
christophe-zurn-sonarsource: Since we are reverting changes related to this rule, I think we can also revert the title change here, as this change should be done when updating the implementation of the rule

sonaranalyzer-dotnet/src/SonarAnalyzer.Utilities/Rules.Description/S1479_cs.html
```diff
@@ -1,5 +1,61 @@
...
+            case 'j':
+                return 2;
+            case 'k':
+            case 'l':
+            case 'm':
+            case 'n':
+            case 'o':
+                return 3;
+            default:
+                return 4;
```
christophe-zurn-sonarsource: This noncompliant example is actually wrong, as the rule has an exception that "fall-through" cases are ignored, this would actually not raise an issue

sonaranalyzer-dotnet/rspec/cs/S1479_c#.html
```diff
@@ -1,5 +1,61 @@
...
+            case 'i':
+            case 'j':
+                return 2;
+            case 'k':
+            case 'l':
+            case 'm':
+            case 'n':
+            case 'o':
+                return 3;
+            default:
```
christophe-zurn-sonarsource: This noncompliant example is actually wrong, as the rule has an exception that "fall-through" cases are ignored, this would actually not raise an issue

pavel-mikula-sonarsource:
Please take a look at my RSPEC update as well to check if it's OK. [RSPEC-1479/VB.NET](https://jira.sonarsource.com/browse/RSPEC-4850)

sonaranalyzer-dotnet/rspec/cs/S1479_c#.html
```diff
@@ -1,5 +1,55 @@
-<p>When <code>switch</code> statements have large sets of <code>case</code> clauses, it is usually an attempt to map two sets of data. A real map
-structure would be more readable and maintainable, and should be used instead.</p>
+<p>When <code>switch</code> statements have large sets of case clauses, it is usually an attempt to map two sets of data. A <code>Dictionary</code>
```
andrei-epure-sonarsource: FYI @pavel-mikula-sonarsource  @christophe-zurn-sonarsource  there's an ongoing discussion around this RSPEC on the community forum https://community.sonarsource.com/t/c-switch-too-many-case-clauses-better-suggested-fix/16993/2

From what I can tell (also see Loic's last comment/whisper), there isn't a clear agreement on what it should become.

However, I believe this form is better than before. Before it was mentioning a map (Java terms), now it's mentioning the C#-specific `Dictionary` and adding a use-case with it.

pavel-mikula-sonarsource: As long as the RSPEC matches the current rule implementation, it should be fine for _this_ release.

andrei-epure-sonarsource: I agree

christophe-zurn-sonarsource: Indeed, new description is more detailed and adds code example. Even though the discussion around it is still ongoing, there is no issue in updating current description, since this is already the behaviour of the rule.

PR #2885, opened 2019-12-11
Use modern MsBuild features and remove mention of Linux …

loic-joly-sonarsource:

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
@@ -20,13 +20,6 @@
       <MainExe>dotnet-symbolic-execution.exe</MainExe>
       <ZipFileName>cbde-windows.zip</ZipFileName>
     </CBDEUrls>
-    <!-- Only Windows is supported for now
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/DownloadCBDE.targets
```diff
@@ -74,20 +64,12 @@
     <Delete Files="@(ExistingCBDEFiles)" Condition="$(ShouldDownloadCBDE)=='true'"/>
 
     <Message Importance="high" Text="CBDE: Begin downloading CBDE zip at @(CBDEUrls)..." Condition="$(ShouldDownloadCBDE)=='true'"/>
-    <Exec ConsoleToMsBuild="true" Condition="$(ShouldDownloadCBDE)=='true'"
-      Command="$(PowerShellExe) -NonInteractive -ExecutionPolicy Unrestricted -File $(DownloadScript) -url @(CBDEUrls) -outfolder $(CBDECurrentFolder) -outfile %(CBDEUrls.ZipFileName)">
```
andrei-epure-sonarsource: you should also remove the `UnzipScript` and `DownloadScript` (lines 45-46) and remove the `unzip.ps1` and `download.ps1` scripts from the disk, right?

loic-joly-sonarsource: Right, I just forgot :)

PR #2890, opened 2019-12-13
CFG Tests import from Sonar-Security

pavel-mikula-sonarsource:


christophe-zurn-sonarsource:
This PR should not be merged on master, but on the feature branch instead

PR #2903, opened 2019-12-13
CBDE: log execution only when the env variable SONAR_DOTNET_LOG_CBDE is defined

geoffray-adde-sonarsource:
Fixes #


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -111,28 +128,29 @@ private static void Initialize()
...
+            Log($"After initialize ({watch.ElapsedMilliseconds} ms)");
         }
 
         private void RegisterMlirAndCbdeInOneStep(SonarAnalysisContext context)
         {
             context.RegisterCompilationAction(
                 c =>
                 {
-                    if(!shouldRunInContext(c))
+                    emitLog = Environment.GetEnvironmentVariables().Contains("SONAR_DOTNET_LOG_CBDE");
```
andrei-epure-sonarsource: small suggestion: put _internal_ in the name - `SONAR_DOTNET_INTERNAL_LOG_CBDE` , to make it clear it's for our internal usage

PR #2905, opened 2019-12-13
Rule S1172: Add C# 8 unit tests for Non-SE CFG rule MethodParameterUnused

### LINKED ISSUE #2901: Rule S1172: Add C# 8 unit tests for Non-SE CFG rule MethodParameterUnused
Subtask of #2849

pavel-mikula-sonarsource:
Fixes #2901


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodParameterUnused.AfterCSharp8.cs
```diff
@@ -0,0 +1,53 @@
+﻿namespace Tests.TestCases
```
christophe-zurn-sonarsource: I think we could add a test with switch expression and tuple deconstruction to ensure C#8 syntax works correctly for theses syntaxes on this rule too

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/LiveVariableAnalysis/CSharp/CSharpLiveVariableAnalysis.cs
```diff
@@ -71,8 +72,11 @@ public sealed class CSharpLiveVariableAnalysis : AbstractLiveVariableAnalysis
                         break;
 
                     case SyntaxKind.VariableDeclarator:
-                        ProcessVariableDeclarator((VariableDeclaratorSyntax)instruction, assignedInBlock,
-                            usedInBlock);
+                        ProcessVariableDeclarator((VariableDeclaratorSyntax)instruction, assignedInBlock, usedInBlock);
+                        break;
+
+                    case SyntaxKind.InvocationExpression:
+                        ProcessLocalFunction(instruction, assignedInBlock, usedInBlock);
```
christophe-zurn-sonarsource: You should keep in a state somewhere the local functions you have already visited, in order to avoid stack overflow in cases of recursive local functions. For instance, the following code will throw `StackOverflowException`:
```
        public int DoSomething(int a)
        {
            static int BadIncB(int x, int seed)
            {
                seed = 1;
                if (x > 1)
                {
                    return BadIncB(x - 1, seed);
                }
                return x + seed;
            }

            return BadIncB(a, 42);
        }
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/LiveVariableAnalysis/CSharp/CSharpLiveVariableAnalysis.cs
```diff
@@ -187,6 +191,28 @@ public sealed class CSharpLiveVariableAnalysis : AbstractLiveVariableAnalysis
             }
         }
 
+        private void ProcessLocalFunction(SyntaxNode instruction, HashSet<ISymbol> assignedInBlock,
```
christophe-zurn-sonarsource: Although there are unfortunately no direct unit test at the moment for LVA, I believe it would be a good thing to test just these changes in a dedicated UT. You could use something similar to what is in `ExplodedGraphTest`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/MethodParameterUnusedTest.cs
```diff
@@ -51,12 +51,22 @@ public void MethodParameterUnused_CodeFix_CS()
         [TestCategory("Rule")]
         public void MethodParameterUnused_FromCSharp7_CS()
         {
-            Verifier.VerifyNoIssueReported(@"TestCases\MethodParameterUnused.CSharp7.cs",
+            Verifier.VerifyNoIssueReported(@"TestCases\MethodParameterUnused.AfterCSharp7.cs",
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I wasn't sure about this since I've seen `ParseOptionsHelper.CSharp7` as well.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodParameterUnused.AfterCSharp8.cs
pavel-mikula-sonarsource: I'll skip Tuple deconstruction after offline discussion.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/LiveVariableAnalysis/CSharp/CSharpLiveVariableAnalysis.cs
pavel-mikula-sonarsource: Good catch!

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/LiveVariableAnalysis/LiveVariableAnalysisTests.cs
```diff
@@ -0,0 +1,148 @@
+﻿extern alias csharp;
```
christophe-zurn-sonarsource: We usually put such `extern alias` statements after the license header, at the start of the using directives

pavel-mikula-sonarsource: I've updated that for `ExplodedGraphTests.cs` as well = the source of the copy/paste :)

PR #2907, opened 2019-12-16
Add tests with default interface members

costin-zaharia-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodOverrideAddsParams.Fixed.cs
```diff
@@ -38,4 +38,18 @@ abstract class Derived : Base
...
+    public interface IFirst
+    {
+        public virtual void Method(int[] numbers)
+        {
+        }
+    }
+
+    public class Second : IFirst
+    {
+        public void Method(params int[] numbers) // Compliant - The default interface method can be accessed only after cast.
```
andrei-epure-sonarsource: do you mean the `base` interface? (why default?)

costin-zaharia-sonarsource: I was referring to the language feature `Default Interface Methods` but indeed this can be a bit confusing.

PR #2909, opened 2019-12-16
Add tests with null coalescing assignment

costin-zaharia-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ExpressionComplexity.cs
```diff
@@ -17,6 +17,10 @@ public bool M()
 
             var d1 = true && false && true && false && true && true; // Noncompliant
 
+            bool? v1 = null, v2 = null, v3 = null, v4 = null, v5 = null;
```
andrei-epure-sonarsource: should you add tests to CognitiveComplexity.cs and FunctionComplexity.cs as well?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CognitiveComplexity.cs
```diff
@@ -148,6 +148,13 @@ class MethodsComplexity
             }
         }
 
+        void NullCoalescenceAssignment() // Compliant - Null coalescence assignment is not considered
```
andrei-epure-sonarsource: rather FN than Compliant?

costin-zaharia-sonarsource: Indeed a FN

PR #2910, opened 2019-12-16
Add tests with alternative interpolated string syntax 

costin-zaharia-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/ControlCharacterInStringTest.cs
```diff
@@ -1,4 +1,4 @@
-/*
+﻿/*
```
andrei-epure-sonarsource: nitpick revert this change?

PR #2917, opened 2019-12-17
Add tests with static local functions

costin-zaharia-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/MethodsShouldNotHaveIdenticalImplementationsTest.cs
```diff
@@ -19,8 +19,12 @@
  */
 
 extern alias csharp;
+extern alias vbnet;
+
 using Microsoft.VisualStudio.TestTools.UnitTesting;
-using csharp::SonarAnalyzer.Rules.CSharp;
+using CSharp = csharp::SonarAnalyzer.Rules.CSharp;
```
andrei-epure-sonarsource: nitpick please put aliases together

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/VariableShadowsField.cs
```diff
@@ -76,4 +76,37 @@ public static VariableShadowsField build(int MyField)
...
+            bool isUsed = true;
+
+            void doSomething()
+            {
+                bool isUsed = true; // Compliant - FN: should check also function variable shadowing?
+            }
+
+            static void doMore()
+            {
+                bool isUsed = true; // Compliant - FN: should check also function variable shadowing?
```
andrei-epure-sonarsource: if it's static, it won't capture the context, so in this case it should be ok, I believe... right?

costin-zaharia-sonarsource: Indeed the context is not captured but it can still be misleading. Part of the rule specifies: 

> shadowing a variable declared in an outer scope can strongly impact the readability, and therefore the maintainability, of a piece of code.

From another point of view since the context is not shared don't know if this really qualifies as shadowing :)

I will remove the comment (leave only `// Compliant`) and we will decide how to proceed when we fix the other case.

andrei-epure-sonarsource: I guess shadowing refers to hiding what could otherwise be accessible.

Indeed it would be misleading, but the message of the rule woulnd't really apply, as it wouldn't truly be shadowing

PR #2918, opened 2019-12-17
Add tests with disposable ref structs and using declarations

costin-zaharia-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ImplementIDisposableCorrectly.cs
```diff
@@ -243,6 +243,13 @@ public class Foo5 : Foo2
             // Do not forget to call base
             base.Dispose(disposing);
     }
+
+    public ref struct Struct
```
andrei-epure-sonarsource: nitpick: just for completeness, I would add `public void Dispose(bool disposing)` to document behavior - I believe this should **not** be permitted, as the pattern is parameterless Dispose call, to be consumable by the `using` statement?

costin-zaharia-sonarsource: Indeed, also `Dispose(bool disposing)` wouldn't make sense since for structs you cannot have inheritance.

PR #2919, opened 2019-12-18
Add tests with switch expressions

costin-zaharia-sonarsource:

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/AllBranchesShouldNotHaveSameImplementation.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/AllBranchesShouldNotHaveSameImplementation.cs
@@ -145,7 +145,7 @@ namespace Tests.Diagnostics
         {
         }
 
-        public int Test(string type)
+        public int SwitchExpressionNoncompliant(string type)
         {
             return type switch // Compliant - FN: all branches call the same function
             {
@@ -155,6 +155,30 @@ namespace Tests.Diagnostics
             };
         }
 
+        public int SwitchExpressionNested(string type)
+        {
+            return type switch
+            {
+                "a" => GetNumber(),
+                _ => type switch // Compliant - FN: all branches call the same function
+                {
+                        "b" => GetNumber(),
+                        "c" => GetNumber(),
+                        _ => GetNumber()
+                    }
+            };
+        }
+
         public int GetNumber() => 42;
+
+        public int SwitchExpressionCompliant(string type)
+        {
+            return type switch // Compliant
+            {
+                "a" => 42,
+                "b" => type.Length,
+                _ => GetNumber()
+            };
+        }
     }
 }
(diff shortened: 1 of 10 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/AllBranchesShouldNotHaveSameImplementation.cs
```diff
@@ -144,5 +144,17 @@ private void DoSomething()
         private void DoSomethingElse()
         {
         }
+
+        public int Test(string type)
+        {
+            return type switch // Compliant - FN: all branches call the same function
+            {
+                "a" => GetNumber(),
```
andrei-epure-sonarsource: would it be useful to add
- a test with nested identical switch expressions?
- compliant tests for switch expressions

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SwitchSectionShouldNotHaveTooManyStatements_CustomValue.cs
```diff
@@ -22,5 +22,21 @@ public Program(int myVariable)
                     break;
             }
         }
+
+        public int Test(string type)
+        {
+            return type switch // Compliant - FN
+            {
+                _ => 1
+                       + 2
```
andrei-epure-sonarsource: this is not really multiple statements, but a big expression inside - I think we don't have a rule for that ("this statement has too many expressions in it"), and it's out of scope anyway

so in this case, the rule does not really apply to switch expressions (given they cannot have statements inside)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SwitchShouldNotBeNested.cs
```diff
@@ -42,6 +42,19 @@ public Program(char choice, int value)
                     Console.WriteLine("Other");
                     break;
             }
+
+            string first = "foo", second = "bar";
+
+            var result = first switch
+            {
+                "a" => second switch // Compliant - FN: should check also switch expressions?
```
andrei-epure-sonarsource: I'm not sure it should apply to switch expressions... at least MS gives example w/ nested switch expressions on their docs site, like a pattern that can be used. 

We can discuss whether it's a nice pattern or not, but nested switch expressions are less bad then nested switch statements (where you can have break, early return, multiple statements per case). Switch expressions have a single return.

@nicolas-harraudeau-sonarsource - WDYT?

costin-zaharia-sonarsource: Indeed, nested expressions are hard to read but not as bad as the statements. 

I've removed the "FN" comment but I've let the test to document the current behavior. We can decide later if we want to extend the rule to cover also this.

PR #2925, opened 2019-12-20
Rule S3900: Update SE to the new CFG and add C# 8 unit tests

### LINKED ISSUE #2897: Rule S3900: Update SE to the new CFG and add C# 8 unit tests PublicMethodArgumentsShouldBeCheckedForNull
Subtask of #2849

costin-zaharia-sonarsource:
Fixes #2897

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -806,11 +806,14 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
         {
             var currentArmBlock = currentBlock;
 
-            foreach (var arm in switchExpressionSyntax.Arms.Reverse())
+            for (var index = switchExpressionSyntax.Arms.Count - 1; index >= 0; index--)
             {
+                var arm = switchExpressionSyntax.Arms[index];
+                var isLast = index == switchExpressionSyntax.Arms.Count - 1;
+
                 var armBlock = BuildExpression(arm.Expression, CreateBlock(currentBlock));
 
-                currentArmBlock = BuildArmBranch(arm, armBlock, currentArmBlock);
+                currentArmBlock = BuildArmBranch(arm, armBlock, currentArmBlock, isLast);
 
                 currentArmBlock = BuildExpression(switchExpressionSyntax.GoverningExpression, currentArmBlock);
             }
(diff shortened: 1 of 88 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PublicMethodArgumentsShouldBeCheckedForNull.cs
```diff
@@ -310,3 +315,48 @@ public override bool Equals(object obj)
...
     }
 }
+
+namespace CSharp8
+{
+    public class NullCoalescenceAssignment
+    {
+        public void NullCoalescenceAssignment_NotNull(string s1)
+        {
+            s1 ??= "N/A";
```
andrei-epure-sonarsource: I guess an interesting test would be
```
            if (o1 != null)
            {
				o1 ??= null;
                o1.ToString(); // Compliant, we know o1 is not null
            }

            if (o1 == null)
            {
				o1 ??= "foo";
                o1.ToString(); // Compliant, assigned foo
            }
```

costin-zaharia-sonarsource: Nice! 😄

sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -524,8 +524,6 @@ private Block BuildPatternExpression(PatternSyntaxWrapper patternSyntaxWrapper,
             }
             else if (DiscardPatternSyntaxWrapper.IsInstance(patternSyntaxWrapper))
             {
-                currentBlock.ReversedInstructions.Add(patternSyntaxWrapper);
-
```
costin-zaharia-sonarsource: Removed since it's not needed as a CFG instruction.

sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -813,9 +811,11 @@ private Block BuildSwitchExpression(SwitchExpressionSyntaxWrapper switchExpressi
                 var armBlock = BuildExpression(arm.Expression, CreateBlock(currentBlock));
 
                 currentArmBlock = BuildArmBranch(arm, armBlock, currentArmBlock);
+
+                currentArmBlock = BuildExpression(switchExpressionSyntax.GoverningExpression, currentArmBlock);
```
costin-zaharia-sonarsource: In order to be consistent and to ease the exploded graph analysis I've added the switch governing expression on each arm.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
```diff
@@ -524,6 +529,14 @@ protected override void VisitInstruction(ExplodedGraphNode node)
                     // Do nothing
                     break;
 
+                case SyntaxKindEx.DiscardPattern:
+                    break;
+
```
christophe-zurn-sonarsource: If discard pattern is not added as instruction in the CFG, then we could remove this part.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
```diff
@@ -520,10 +525,20 @@ protected override void VisitInstruction(ExplodedGraphNode node)
...
                     // The 0 in 'case 0 when ...'
                     // Do nothing
                     break;
 
+                case SyntaxKindEx.DeclarationExpression:
+                    // e.g.: dictionary.TryGetValue(key, out TValue value);
+                    // Do nothing
+                    break;
+
+                case SyntaxKindEx.TupleExpression:
```
christophe-zurn-sonarsource: Last note: I would expect to see the changes in `CSharpExplodedGraph` to be reflected in `ExplodedGraphTests`. As currently, we only check that these syntaxes do not fail for rules we expect, but we don't check that the exploded graph behaves as expected

PR #2934, opened 2019-12-26
Rule S3966: Support C#8 CFG

### LINKED ISSUE #2896: Rule S3966: Update SE to the new CFG and add C# 8 unit tests ObjectsShouldNotBeDisposedMoreThanOnce
Subtask of #2849

christophe-zurn-sonarsource:
Fixes #2896


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
```diff
@@ -142,10 +142,36 @@ public override ProgramState PreProcessUsingStatement(ProgramPoint programPoint,
 
             public override ProgramState PreProcessInstruction(ProgramPoint programPoint, ProgramState programState)
             {
+                var instruction = programPoint.Block.Instructions[programPoint.Offset];
+                if (instruction is InvocationExpressionSyntax invocation)
+                {
+                    return VisitInvocationExpression(invocation, programState);
+                }
+                else if (instruction is VariableDeclaratorSyntax declarator &&
```
costin-zaharia-sonarsource: In this case else is not needed due the return statement from the first if.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ObjectsShouldNotBeDisposedMoreThanOnce.cs
```diff
@@ -142,10 +142,36 @@ public override ProgramState PreProcessUsingStatement(ProgramPoint programPoint,
...
+                    declarator.Parent?.Parent is LocalDeclarationStatementSyntax declaration &&
+                    declaration.UsingKeyword() != null &&
+                    declaration.UsingKeyword().Kind() != SyntaxKind.None)
```
costin-zaharia-sonarsource: `declaration.UsingKeyword() != null` is always true since the SyntaxToken is a struct. Could we use `declaration.IsKind(SyntaxKind.UsingKeyword)` instead?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/AbstractExplodedGraph.cs
```diff
@@ -274,6 +275,15 @@ protected virtual void VisitSimpleBlock(SimpleBlock block, ExplodedGraphNode nod
 
         protected ProgramState CleanStateAfterBlock(ProgramState programState, Block block)
         {
+            // We do not clean state when entering a using block statement, as some variables declared with "using" keywords
+            // may still be implictly disposed later, even if they are not referenced anymore. State will still be cleaned
```
costin-zaharia-sonarsource: Small typo, it think it was meant to be "implicitly"

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/AbstractExplodedGraph.cs
```diff
@@ -274,6 +275,15 @@ protected virtual void VisitSimpleBlock(SimpleBlock block, ExplodedGraphNode nod
...
+            // in the "VisitSimpleBlock" of UsingEndBlock
+            if (block is JumpBlock jumpBlock &&
+                jumpBlock.SuccessorBlock is UsingEndBlock)
+            {
```
costin-zaharia-sonarsource: Should not differentiate here between the using statement and local declaration with using?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -935,5 +935,93 @@ public void Test(string s)
 
             explodedGraph.Walk();
         }
+
+        [TestMethod]
+        [TestCategory("Symbolic execution")]
+        public void ExplodedGraph_DefaultLiteral()
+        {
+            var testInput = "var i = default(int); int j = default; System.IO.File k = default;";
```
costin-zaharia-sonarsource: pedantic: can be a constant

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -935,5 +935,93 @@ public void Test(string s)
...
+            var method = ControlFlowGraphTest.CompileWithMethodBody(string.Format(TestInput, testInput), "Bar", out var semanticModel);
+            var methodSymbol = semanticModel.GetDeclaredSymbol(method);
+            var varDeclarators = method.DescendantNodes().OfType<VariableDeclaratorSyntax>();
+            var iSymbol = semanticModel.GetDeclaredSymbol(varDeclarators.First(d => d.Identifier.ToString() == "i"));
```
costin-zaharia-sonarsource: I think that we could extract a method, for symbol retrieving, based on it's name, to reduce a bit the code duplication.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -935,5 +935,93 @@ public void Test(string s)
...
+            var cfg = CSharpControlFlowGraph.Create(method.Body, semanticModel);
+            var lva = CSharpLiveVariableAnalysis.Analyze(cfg, methodSymbol, semanticModel);
+            var explodedGraph = new CSharpExplodedGraph(cfg, methodSymbol, semanticModel, lva);
+
+            var numberOfProcessedInstructions = 0;
+            explodedGraph.InstructionProcessed +=
+                (sender, args) =>
+                {
+                    numberOfProcessedInstructions++;
+                    if (args.Instruction.ToString() == "i = default(int)")
```
costin-zaharia-sonarsource: - we transform the instruction to string multiple times, an option would be to do it once and use the value in the if cases.
- we could also use a switch since (I find the switch more readable)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -935,5 +935,93 @@ public void Test(string s)
...
+                    // Stack is clean after assignment
+                    if (args.Instruction.ToString() == "(object d, object e) = (1, null)")
+                    {
+                        args.ProgramState.HasValue.Should().BeFalse();
+                    }
+                };
+
+            explodedGraph.Walk();
+            numberOfProcessedInstructions.Should().Be(7);
+        }
```
costin-zaharia-sonarsource: I think it would be great to have a test also for using declarations and one for statements.

christophe-zurn-sonarsource: I added a test for local using statement, but in the `ControlFlowGraphTest` class, as it is not really related to the eploded graph. Other tests for usings are already there

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
christophe-zurn-sonarsource: Added a variable to hold the string value of the instruction. I'd rather no use switch here when there is only a few (3) cases.
Same as for the other test, as I find the comments are necessarys, and adding them to a switch expression makes it a bit less readable in my opinion. (But this one I don't mind updating if you think it better as a switch)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
christophe-zurn-sonarsource: Indeed, but since all other `testInput` variables in this test file are declared the same way, I ll leave it as is for consistency

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/AbstractExplodedGraph.cs
christophe-zurn-sonarsource: Local declaration with using will not create a `UsingEndBlock` at all, the way the CFG is built was not modified. Here, only the rule was updated to look for local using declarations to for "dispose" calls or its equivalent

PR #2940, opened 2020-01-02
Rule S1944: Enable and add C# 8 unit tests

### LINKED ISSUE #2894: Rule S1944: Update SE to the new CFG and add C# 8 unit tests InvalidCastToInterface
Subtask of #2849

christophe-zurn-sonarsource:
Fixes #2894

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InvalidCastToInterface.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InvalidCastToInterface.cs
@@ -175,3 +175,57 @@ namespace Tests.Diagnostics
         }
     }
 }
+
+namespace Tests.Diagnostics
+{
+    namespace CSharp8
+    {
+        public class NullCoalescenceAssignment
+        {
+            public void NullCoalescenceAssignment_Null()
+            {
+                int? i1 = null;
+                i1 ??= (int)i1; // Noncompliant {{Nullable is known to be empty, this cast throws an exception.}}
+            }
+
+            public void NullCoalescenceAssignment_NotNull()
+            {
+                int? i1 = 1;
+                i1 ??= (int)i1;
+            }
+        }
+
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InvalidCastToInterface.cs
```diff
@@ -175,3 +175,92 @@ public void Method<T>(T generic) where T : new()
...
+
+            public void NullCoalescenceAssignment_NotNull()
+            {
+                int? i1 = 1;
+                i1 ??= (int)i1;
+            }
+        }
+
+        public class SwitchExpressions
+        {
```
pavel-mikula-sonarsource: I suggest to add pair of tests like this to illustrate FN behaviour as documented in new issue here #2949 
```
            public int CorrectCheck(int? val)
            {
                return val.HasValue switch
                {
                    true => (int)val,
                    _ => 0
                };
            }

            public int WrongCheck_Noncompliant(int? val)
            {
                return val.HasValue switch
                {
                    false => (int)val,  // FN
                    _ => 0
                };
            }
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InvalidCastToInterface.cs
```diff
@@ -175,3 +175,92 @@ public void Method<T>(T generic) where T : new()
         }
     }
 }
+
+namespace Tests.Diagnostics
+{
+    namespace CSharp8
+    {
+        public class NullCoalescenceAssignment
```
costin-zaharia-sonarsource: I think it would be nice to have one more test to check the result of the coalesce assignment.
e.g. 
```
	int? i = null;
	
	i ??= null;
	var r1 = (int) i; // Noncompliant
	
	i ??= 1;
	var r2 = (int) i; // Compliant
```

Issue #2891, opened 2019-12-13
Rule S2583, S2589: Update SE to the new CFG and add C# 8 unit tests ConditionEvaluatesToConstant


pavel-mikula-sonarsource:
Subtask of #2849

Issue #2849, opened 2019-11-28
Add C# 8 unit tests to the CFG-based rules


andrei-epure-sonarsource:
### Description

Part of https://jira.sonarsource.com/browse/MMF-1808

This is a big task (which can be split into multiple issues) which requires:

- changes to the unit test framework to support C# 8 syntax w/ netcore3 and/or netframework4.8 dependencies
- adding unit tests with relevant C# 8 syntax for the following rules which are using the Control Flow Graph

Symbolic execution rules: 
1. - ConditionEvaluatesToConstant / S2583 , S2589
2. - EmptyCollectionsShouldNotBeEnumerated / S4158
3. - EmptyNullableValueAccess / S3655
4. - InvalidCastToInterface / S1944
5. - NullPointerDereference / S2259
6. - ObjectsShouldNotBeDisposedMoreThanOnce / S3966
7. - PublicMethodArgumentsShouldBeCheckedForNull / S3900

Non-SE CFG rules:

1. - DeadStores / S1854
2. - InfiniteRecursion / S2190
3. - MemberInitializerRedundant / S3604
4. - MethodParameterUnused / S1172
5. - RedundantJumpStatement / S3626

pavel-mikula-sonarsource:
I'm closing this based on offline discussion with @andrei-epure-sonarsource. We've agreed that this issue is not needed anymore. It has been divided into partial issues referenced above. C#8 support in unit test framework was done as well.

PR #2941, opened 2020-01-02
ExplodedGraphTests refactoring

pavel-mikula-sonarsource:
Related to #2891 and all other SE issues from #2849

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
@@ -669,9 +669,9 @@ namespace Namespace
                     }
                 };
 
-            context.WalkWithExitBlocks(2, 9);
+            context.WalkWithExitBlocks(9, 2);
 
-            instructionsInspected.Should().Be(4);
+            instructionsInspected.Should().Be(5);
         }
 
         [TestMethod]
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
@@ -595,7 +595,6 @@ namespace Namespace
             numberOfValidatedInstructions.Should().Be(1);
         }
 
-        [Ignore("Temporary disabled due to SwitchArm CFG")] //FIXME: Activate after CFG switch expression stack fix. There are 2 exit blocks reached, but there should be only one (right?)
         [TestMethod]
         [TestCategory("Symbolic execution")]
         public void ExplodedGraph_SwitchWithRecursivePatternVisit()
@@ -670,7 +669,7 @@ namespace Namespace
                     }
                 };
 
-            context.Walk(9);
+            context.WalkWithExitBlocks(2, 9);
 
             instructionsInspected.Should().Be(4);
         }
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
@@ -551,7 +551,6 @@ namespace Namespace
             context.WalkWithInstructions(5);
         }
 
-        [Ignore("Temporary disabled due to SwitchArm CFG")] //FIXME: Activate after CFG switch expression stack fix
         [TestMethod]
         [TestCategory("Symbolic execution")]
         public void ExplodedGraph_SwitchExpression_SimpleExpression()
@@ -569,7 +568,7 @@ namespace Namespace
                 }
             };
 
-            context.WalkWithInstructions(0);
+            context.WalkWithInstructions(14);
 
             numberOfValidatedInstructions.Should().Be(1);
         }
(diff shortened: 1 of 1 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -222,39 +159,21 @@ public void ExplodedGraph_SingleBranchVisited_If()
...
+            const string testInput = "var a = false; if (a && !a) { a = true; }";
+            var context = new ExplodedGraphContext(testInput);
+            var aSymbol = context.GetSymbol("a");
+            context.EG.InstructionProcessed +=
                 (sender, args) =>
                 {
-                    numberOfProcessedInstructions++;
                     if (args.Instruction.ToString() == "a = !true")
                     {
                         args.ProgramState.GetSymbolValue(aSymbol).Should().Be(SymbolicValue.False); // Roslyn is clever !true has const value.
```
christophe-zurn-sonarsource: Are we actually visiting this check here ? Looking at the input code now, the instruction `a = !true` is not present there

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -265,47 +184,27 @@ public void ExplodedGraph_SingleBranchVisited_And()
                     }
                 };
 
-            explodedGraph.Walk();
-
-            explorationEnded.Should().BeTrue();
-            numberOfExitBlockReached.Should().Be(1);
+            context.Walk(3); // Only EG final status check
```
christophe-zurn-sonarsource: Why is this comment necessary here ? Isn't it the case for all (except 1) tests ?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -339,117 +238,71 @@ public void ExplodedGraph_BothBranchesVisited()
                     }
                 };
 
-            explodedGraph.Walk();
+            // Number of Exit blocks:
```
christophe-zurn-sonarsource: Minor clarification:
```suggestion
            // Number of exit block is still 1 in this case:
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -525,90 +366,60 @@ public void ExplodedGraph_NonDecisionMakingAssignments()
...
                     {
                         args.ProgramState.GetSymbolValue(propertySymbol).Should().BeNull();
                     }
                 };
 
-            explodedGraph.Walk();
-
-            explorationEnded.Should().BeTrue();
-            numberOfExitBlockReached.Should().Be(1);
+            context.Walk(3);     // Only EG final status check
```
christophe-zurn-sonarsource: Comment not needed here wither

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -56,27 +57,13 @@ public void Bar(bool inParameter, out bool outParameter)
...
-            var numberOfExitBlockReached = 0;
-            explodedGraph.ExitBlockReached += (sender, args) => { numberOfExitBlockReached++; };
-
-            var numberOfProcessedInstructions = 0;
-            explodedGraph.InstructionProcessed +=
+            const string testInput = "var a = true; var b = false; b = !b; a = (b);";
+            var context = new ExplodedGraphContext(testInput);
+            var aSymbol = context.GetSymbol("a");
+            var bSymbol = context.GetSymbol("b");
+            context.EG.InstructionProcessed +=
```
christophe-zurn-sonarsource: I would prefer to keep `ExplodedGraph` instead of `EG` for readability inside the context class. WDYT ?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -97,50 +84,28 @@ public void ExplodedGraph_SequentialInput()
                     }
                 };
 
-            explodedGraph.Walk();
-
-            explorationEnded.Should().BeTrue();
-            numberOfProcessedInstructions.Should().Be(9);
-            numberOfExitBlockReached.Should().Be(1);
+            context.Walk(9);
```
christophe-zurn-sonarsource: I think it would be better to separate the `Walk`step from the assertion step, that way it is clear what we do in each method:
```
context.Walk();

context.HasWalked(9);
```
I would find the above easier to read, and it would not need the comments `Special case with manual checks` as it would be clearly separated

pavel-mikula-sonarsource: There are 3 more checks in the `Walk` itself. And I wanted to enforce all these checks for every test (except special cases) to maximize coverage. Newly added tests tended to miss some of those checks.

I'd prefer just to rename the method for something more explicit. Like `WalkWithInstructions(9)` as there already exists `WalkWithExitBlocks`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
pavel-mikula-sonarsource: Good catch! It has been there longer than 2 years already 😄

christophe-zurn-sonarsource:
You can rebase on current feature bracnh now. It should slightly impact test results, as you mentioned earlier

andrei-epure-sonarsource:
nice! 😄

PR #2943, opened 2020-01-02
Update CFG for switch expressions

### LINKED ISSUE #2938: Fix CFG generation for the last branch of switch arm
### Description

The generated CFG for the last branch of the switch arm needs to be improved: currently there are 2 simple nodes created for the last branch but only one is needed and this makes the symbolic execution difficult.

e.g.
```
public int Test(string str)
{
    return str switch { null => 1, """" => 2, _ => 3};
}
```

In this example the node with "str" instruction is redundant and needs to be removed.

image

costin-zaharia-sonarsource:
- last arm should have only one node instead of two

Fixes #2938

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -815,7 +815,10 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
 
                 currentArmBlock = BuildArmBranch(arm, armBlock, currentArmBlock, isLast);
 
-                currentArmBlock = BuildExpression(switchExpressionSyntax.GoverningExpression, currentArmBlock);
+                if (!isLast)
+                {
+                    currentArmBlock = BuildExpression(switchExpressionSyntax.GoverningExpression, currentArmBlock);
+                }
             }
 
             return currentArmBlock;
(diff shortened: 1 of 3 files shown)


pavel-mikula-sonarsource:
This PR fixes issues I've run into with switch expressions CFG.

sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -815,7 +815,10 @@ private Block BuildSwitchExpression(SwitchExpressionSyntaxWrapper switchExpressi
 
                 currentArmBlock = BuildArmBranch(arm, armBlock, currentArmBlock, isLast);
 
-                currentArmBlock = BuildExpression(switchExpressionSyntax.GoverningExpression, currentArmBlock);
+                if (!isLast)
+                {
+                    currentArmBlock = BuildExpression(switchExpressionSyntax.GoverningExpression, currentArmBlock);
+                }
```
christophe-zurn-sonarsource: Could we extract the last arm outside the loop ? That would avoid the need for the `isLast` boolean completely and simplify the loop. WDYT ?

costin-zaharia-sonarsource: If I understood correctly I don't think that that it's possible since we want to add the expression for all the arms but the last.

christophe-zurn-sonarsource: It could be possible, but indeed, I m not sure we would win much readability actually, so forget about my comment

PR #2945, opened 2020-01-03
Add C# 8 unit tests for EmptyCollectionsShouldNotBeEnumerated

### LINKED ISSUE #2892: Rule S4158: Update SE to the new CFG and add C# 8 unit tests EmptyCollectionsShouldNotBeEnumerated
Subtask of #2849

costin-zaharia-sonarsource:
Fixes #2892

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/EmptyCollectionsShouldNotBeEnumerated.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/EmptyCollectionsShouldNotBeEnumerated.cs
@@ -809,7 +809,7 @@ namespace CSharp8
     {
         private static bool Predicate(int i) => true;
 
-        public bool Test(int type)
+        public bool InsideSwitch(int type)
         {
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/EmptyCollectionsShouldNotBeEnumerated.cs
```diff
@@ -792,3 +792,82 @@ public bool TryAdd(TKey key, TValue value)
...
+
+        public void UsingSwitchResult(bool cond)
+        {
+            var list = cond switch
+            {
+                _ => new List<int>()
+            };
+
+            list.Clear(); // Noncompliant
+        }
```
christophe-zurn-sonarsource: I think it would be nice to add a compliant example when there are multiple arms:
```
        public void UsingSwitchResult_Compliant(bool cond)
        {
            IList<int> list = cond switch
            {
                true => new int[] { 5 },
                _ => new List<int>()
            };

            list.Clear();
        }
```

PR #2948, opened 2020-01-03
Rule S2259: Enable and add C# 8 unit tests

christophe-zurn-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereferenceCSharp8.cs
```diff
@@ -0,0 +1,139 @@
...
+    {
+        public void Nullable_In_Arm_Noncompliant(string s)
+        {
+            var result = s switch
+            {
+                null => s.ToString(), // Noncompliant
+                _ => s.ToString()
+            };
+        }
+
```
pavel-mikula-sonarsource: I suggest to add FN test for those omnipresent switch expressions:
```
public void Nullable_In_Arm_FN(string s)
{
    var result = s switch
    {
        "" => s.ToString(),
        _ => s.ToString()   // FN SwitchArms are not constrained
    };
}
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereferenceCSharp8.cs
```diff
@@ -0,0 +1,148 @@
...
+            string result = val switch
+            {
+                1 => "1",
+                2 => "2",
+                _ => "Neither 1 or 2"
+            };
+            result.ToString();
+        }
+
+        public void Nullable_In_Arm(string? s)
```
pavel-mikula-sonarsource: Is that `?` needed here? I would make sense for S1944, but I think that in this rule, it should work fine without it.
```suggestion
        public void Nullable_In_Arm(string s)
```

PR #2950, opened 2020-01-03
Rule S2583, S2589: add C# 8 unit tests

### LINKED ISSUE #2891: Rule S2583, S2589: Update SE to the new CFG and add C# 8 unit tests ConditionEvaluatesToConstant

pavel-mikula-sonarsource:
Fixes #2891

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
@@ -1,0 +1,129 @@
+﻿using System;
+
+namespace Tests.Diagnostics
+{
+    public interface IWithDefaultImplementation
+    {
+        void DoSomething();
+
+        //Default interface methods
+        void DoNothing()
+        {
+            var a = false;
+            if (a)  // Noncompliant
+            {       // Secondary
+                DoSomething(); // never executed
+            }
+        }
+    }
+
+    public class CSharp8
+    {
+        int SwitchExpression()
+        {
+            var a = false;
+            return a switch { true => 0, false => 1 };  // FN - switch arms are not inspected
+        }
+
+        int SwitchExpression_Discard()
+        {
+            var a = false;
+            return a switch { true => 0, _ => 1 };  // FN - switch arms are not inspected
+        }
+
+        int SwitchExpression_Results()
+        {
+            bool a;
+
+            a = false switch { _ => false };
+            if (a)  // Noncompliant
+            {       // Secondary
+                return 42; // never executed
+            }
+
+            a = false switch { false => false, _ => false };
+            if (a)  // Noncompliant
+            {       // Secondary
+                return 42; // never executed
+            }
+
+            a = false switch { true => true, _ => false };
+            if (a)  // FN - switch arms are not constrained
+            {
+                return 42; // never executed
+            }
+
+            a = 0 switch { 1 => true, _ => false };
+            if (a)  // FN - switch arms are not constrained
+            {
+                return 42; // never executed
+            }
+            return 0;
+        }
+
+        int UsingDeclaration_Null()
+        {
+            using System.IO.MemoryStream ms = null;
+            if (ms != null) // Noncompliant
+            {               // Secondary
+                return 1;
+            }
+            return 0;
+        }
+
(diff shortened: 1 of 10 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
```diff
@@ -0,0 +1,129 @@
+﻿using System;
+
+namespace Tests.Diagnostics
+{
+    public interface IWithDefaultImplementation
+    {
+        void DoSomething();
+
+        //Default interface methods
```
costin-zaharia-sonarsource: pedantic: I think the comment is redundant and can be removed

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+    }
+
+    public class CSharp8
+    {
+        int SwitchExpression()
+        {
+            var a = false;
+            return a switch { true => 0, false => 1 };  // FN - switch arms are not inspected
```
costin-zaharia-sonarsource: side notes:
- the switch arms are inspected but the only constrain we currently add is for null
- we have the same behavior also for switch statements so if we want to support this it would make sense to handle them together

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            a = false switch { false => false, _ => false };
+            if (a)  // Noncompliant
+            {       // Secondary
+                return 42; // never executed
+            }
+
+            a = false switch { true => true, _ => false };
+            if (a)  // FN - switch arms are not constrained
```
costin-zaharia-sonarsource: In the example above the arms are considered and here not? 😄 

I think this case is different from the others since we miss to add a constraint for the governing expression (not switch arms like it's mentioned)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
```diff
@@ -0,0 +1,129 @@
...
+            {
+                var a = false;
+                if (a)  // FN Static local functions are not inspected in CFG
+                {
+                    return 0; // never executed
+                }
+                return 1;
+            }
+
+            if (ReturnFalse())  // FN - content and resutl value of local static funcition is not inspected
```
costin-zaharia-sonarsource: pedantic: typos

costin-zaharia-sonarsource:
We have also a related FP which is waiting for the CFG update 😄  https://github.com/SonarSource/sonar-dotnet/issues/2496#issuecomment-516840623. 

I think we can add a unit test to check if this was indeed fixed and if not to document the behavior.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
pavel-mikula-sonarsource: In the example above, the arms produce the same value. So the result of the test does not depend on constraints (not now, nor in the future).

I would keep the comment like this. We don't need to document too much details in here. And it will get deleted when the CFG is fixed anyway.

costin-zaharia-sonarsource: Maybe I miss something but shouldn't the false negative be on the switch expression line and not here?

pavel-mikula-sonarsource: You're right. It should be on both.

PR #2954, opened 2020-01-06
Fix S2479 FP: add support for verbatim interpolated string

### LINKED ISSUE #2773: Fix S2479 FP: add support for verbatim interpolated string
### Description

from [community](https://community.sonarsource.com/t/verbatim-string-with-interpolation-causing-false-positives-with-rule-c-s2479/16026)

We do not support verbatim interpolated strings (C#7 syntax)

### Repro steps

```
// FP below
string testB = $@"asdasdasd
        asasdasd{testA}asdasd
        asdasd";
```

### Related information

* SonarC# Version 8.0
* MSBuild 16

andrei-epure-sonarsource:
Fixes #2773

note: to see the test file, you need to open it locally in a text editor, because it's considered as binary by VS and github


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ControlCharacterInString.cs
```diff
@@ -113,10 +113,16 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            // simple verbatim - the node is StringLiteralExpression, the first token is verbatim
             if (c.Node.GetFirstToken().IsVerbatimStringLiteral())
             {
                 return;
             }
+            // interpolated verbatim - the node is InterpolatedStringText, the parent is InterpolatedStringExpression
+            if (c.Node.Parent.GetFirstToken().IsKind(SyntaxKind.InterpolatedVerbatimStringStartToken))
+            {
+                return;
+            }
```
costin-zaharia-sonarsource: I think that extracting a method with the checks (e.g. `ShouldIgnore(SyntaxNode)`) will make the code a bit more readable. 

Also, instead of using comments I would create 2 methods `IsVerbatimString` and `IsInterpolatedVerbatimString`.

Issue #2922, opened 2019-12-20
Fix S3949 failures on test projects


andrei-epure-sonarsource:
### Description

There are some failing tests for the CBDE integration (rule S3649) - related to #619 

The failing projects are benchmark-dot-net, nodatime, ravendb and servuo

```
    CSC : error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.CbdeHandlerRule' threw an exception of type 'SonarAnalyzer.CBDE.CbdeException' with message 'Top level error in CBDE handling: CBDE external process reported an error'. [C:\Project\samples\BenchmarkDotNet.Samples\BenchmarkDotNet.Samples.csproj]
    CSC : error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.CbdeHandlerRule' threw an exception of type 'SonarAnalyzer.CBDE.CbdeException' with message 'Top level error in CBDE handling: CBDE external process reported an error'. [C:\Project\samples\BenchmarkDotNet.Samples\BenchmarkDotNet.Samples.csproj]
    CSC : error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.CbdeHandlerRule' threw an exception of type 'SonarAnalyzer.CBDE.CbdeException' with message 'Top level error in CBDE handling: CBDE external process reported an error'. [C:\Project\samples\BenchmarkDotNet.Samples\BenchmarkDotNet.Samples.csproj]
```
```
CSC : error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.CbdeHandlerRule' threw an exception of type 'SonarAnalyzer.CBDE.CbdeException' with message 'Top level error in CBDE handling: Object reference not set to an instance of an object.'. [C:\Project\src\NodaTime.Tools.DumpTimeZoneInfo\NodaTime.Tools.DumpTimeZoneInfo.csproj]
CSC : error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.CbdeHandlerRule' threw an exception of type 'SonarAnalyzer.CBDE.CbdeException' with message 'Top level error in CBDE handling: Object reference not set to an instance of an object.'. [C:\Project\src\NodaTime.Tools.Common\NodaTime.Tools.Common.csproj]
[...]

PR #2955, opened 2020-01-06
Replace Json.Net with Utf8Json...

andrei-epure-sonarsource:
...in the attempt to overcome the runtime binding problems we have on netcore projects/environments

side effects:
- Utf8Json is faster than Json.Net according to benchmarks on multiple blogs
- the side of the jar slightly decreased

Related to #2922


andrei-epure-sonarsource:
not relevant anymore

PR #2956, opened 2020-01-07
Fixes issues related to execution CBDE on .NET Core

loic-joly-sonarsource:
Related to #2922 
It might be interesting not to squash those commit since they are related to different issues discovered in the same investigation.


andrei-epure-sonarsource:
you'll need to remove the newtonsoft json DLL copying from the scripts (see https://github.com/SonarSource/sonar-dotnet/pull/2955/commits/5be98144527b45717dcf8a66dfd595e68e4cdd29)

-  sonaranalyzer-dotnet/src/SonarAnalyzer.Vsix/SonarAnalyzer.CSharp.nuspec
- sonaranalyzer-dotnet/src/SonarAnalyzer.Vsix/tools-cs/install.ps1 
-  sonaranalyzer-dotnet/src/SonarAnalyzer.Vsix/tools-cs/uninstall.ps1

andrei-epure-sonarsource:
I believe these changes are in #2962, right?

loic-joly-sonarsource:
> I believe these changes are in #2962, right?

Yes, they are. Since #2962 required to create a dummy exchange file, I did not want to do it in json, and based it on top of this branch.

PR #2962, opened 2020-01-08
Fix crashes; and more tests

loic-joly-sonarsource:
The goal is to increase the coverage of CbdeHandler.


sonaranalyzer-dotnet/tests/CBDE/CBDEFails/CBDEFails.csproj
```diff
@@ -0,0 +1,53 @@
...
+    <PlatformTarget>AnyCPU</PlatformTarget>
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>..\bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="System" />
```
andrei-epure-sonarsource: probably you don't need all references

sonaranalyzer-dotnet/tests/CBDE/CBDESucceedsWithIncorrectResults/CBDESucceedsWithIncorrectResults.csproj
```diff
@@ -0,0 +1,59 @@
...
+    <OutputPath>..\bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
```
andrei-epure-sonarsource: probably you don't need all references

sonaranalyzer-dotnet/tests/CBDE/CBDEArguments/CBDEArguments.csproj
```diff
@@ -0,0 +1,48 @@
...
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
```
andrei-epure-sonarsource: probably you don't need all references

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -23,23 +23,32 @@
...
 using System.Text;
 using System.Threading;
+using System.Xml.Linq;
+using System.Xml;
 
 namespace SonarAnalyzer.CBDE
 {
     public class CbdeHandler
     {
+        public enum MockType
```
andrei-epure-sonarsource: I believe we don't need this
- we can just have a constructor w/o the CBDE binary path, and then the `Initialize` does the magic- just like until now
- and we can have a constructor that gets an already created binaryPath which is passed from the tests

I would avoid any sort of complex mocking logic inside the production code

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/CBDE/CbdeHandlerTest.cs
```diff
@@ -1,23 +1,28 @@
...
- *
- * You should have received a copy of the GNU Lesser General Public License
- * along with this program; if not, write to the Free Software Foundation,
- * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
- */
+﻿extern alias csharp;
 
+using System.IO;
+using System.Text.RegularExpressions;
+/*
```
andrei-epure-sonarsource: the header is off

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -23,26 +23,26 @@
...
+using System.Xml.Linq;
+using System.Xml;
 
 namespace SonarAnalyzer.CBDE
 {
     public class CbdeHandler
     {
         private const int processStatPeriodMs = 1000;
-        private const string cbdeJsonOutputFileName = "cbdeSEout.json";
+        private const string cbdeJsonOutputFileName = "cbdeSEout.xml";
```
andrei-epure-sonarsource: not Json

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/CBDE/CbdeHandler.cs
```diff
@@ -127,26 +144,29 @@ private void PerformanceLog(string s)
 
         private static void Initialize()
         {
-            cbdeBinaryPath = Path.Combine(Path.GetTempPath(), $"CBDE_{Process.GetCurrentProcess().Id}");
-            Directory.CreateDirectory(cbdeBinaryPath);
-            lock (logFileLock)
+            if (extractedCbdeBinaryPath == null)
```
andrei-epure-sonarsource: I think that's not necessary anymore

PR #2964, opened 2020-01-08
Merge master into feature/csharp-extract-cfg

costin-zaharia-sonarsource:


pavel-mikula-sonarsource:
The PR itself should be merged into `feature/cshar-extrat-cfg` and not `master`, right?

christophe-zurn-sonarsource:
To be consistant with 5c6c5010aebc222e24a3d55743cd2e8db775831e, we should also add `<LangVersion>8.0</LangVersion>` to `SonarAnalyzer.CFG.csproj` in the merge PR

costin-zaharia-sonarsource:
> Can/Should we omit these files from merging to keep the commit/file history clean?
@pavel-mikula-sonarsource: I think that then the title of the commit will be misleading. I prefer to merge the files only with minimum required changes to fix the conflicts. The changes you mentioned (when done) will be later merged here if needed.

> To be consistant with 5c6c501, we should also add <LangVersion>8.0</LangVersion> to SonarAnalyzer.CFG.csproj in the merge PR

I'm thinking to do the change in another commit since it's not actually part of the merge. I want to keep the merge focused only on porting the existing changes from master to the feature branch. What do you think?

costin-zaharia-sonarsource:
I've added #2967 for C# 8 language change

PR #2968, opened 2020-01-08
Merge master into feature/csharp-extract-cfg

costin-zaharia-sonarsource:


costin-zaharia-sonarsource:
There seems to be conflicts between **master** and **feature/csharp-extract-cfg** even after 5cc4e6cfeb5de261295bdcc75392f337cb212f4d was merged and this makes me think that something went wrong with the merge. 

The conflicts can be seen at the bottom of the pull request page: https://github.com/SonarSource/sonar-dotnet/pull/2963

Due to this I've created this new PR without any other changes than an empty **merge** commit. It is empty since all the conflicts were solved by the previous PR.

PR #2971, opened 2020-01-08
Rebase csharp-extract-cfg on top of master

costin-zaharia-sonarsource:


christophe-zurn-sonarsource:
Could we rebase again this PR on latest master ?
I'm think that merging in current state will fail since there was a version change, and this change will not be reflected in the new CFG project file, which will cause the build to fail. This should be addressed before merging I believe

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
```diff
@@ -25,11 +26,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Update="ShimLayer\*.cs">
-      <SonarQubeExclude>true</SonarQubeExclude>
-    </Compile>
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
```
christophe-zurn-sonarsource: I think this was a rebasing mistake, there should not be any reference to `Newtonsoft.Json` on master anymore

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/packages.lock.json
```diff
@@ -12,6 +12,12 @@
           "Microsoft.CodeAnalysis.Workspaces.Common": "[1.3.2]"
         }
       },
+      "Newtonsoft.Json": {
```
christophe-zurn-sonarsource: packages.lock.json files will need to be regenerated without this `Newtonsoft.json` reference

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodParameterUnused.CSharp8.cs
```diff
@@ -1,31 +1,74 @@
...
+    {
+        public int DoSomething(int a)   // Compliant
         {
-            void M1(int a,
-                    int b, // Noncompliant
-                    int c,
-                    int d) // Noncompliant
-                    {
-                        var result = a + c;
-                    }
```
christophe-zurn-sonarsource: Could we keep 1 of these tests from master to show that we correctly raise the issue when several noncompliant parameters are found on the same method ?

PR #3015, opened 2020-01-10
Rule S4144 - update to support static local functions

### LINKED ISSUE #2976: Rule S4144 - update to support static local functions
## Description

costin-zaharia-sonarsource:
Fixes #2976


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/TypeDeclarationSyntaxExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+        public static IEnumerable<IMethodDeclaration> GetMethodDeclarations(this TypeDeclarationSyntax typeDeclaration) =>
+            typeDeclaration
+                .Members
+                .OfType<MethodDeclarationSyntax>()
+                .SelectMany(method => GetLocalFunctions(method).Union(new List<IMethodDeclaration> { new MethodDeclarationSyntaxAdapter(method)}));
+
+        private static IEnumerable<LocalFunctionStatementAdapter> GetLocalFunctions(MethodDeclarationSyntax methodDeclaration)
+            => methodDeclaration.DescendantNodes()
+                .Where(member => member.IsKind(SyntaxKindEx.LocalFunctionStatement))
+                .Select(member => new LocalFunctionStatementAdapter((LocalFunctionStatementSyntaxWrapper)member));
```
christophe-zurn-sonarsource: Is the null check inside the `MethodDeclarationSyntaxAdapter` and `LocalFunctionStatementAdapter` constructor necessary ?
If any of these 2 syntaxes are null, a `NullReferenceException` will be thrown as `methodDeclaration.DescendantNodes()` is called before the `MethodDeclarationSyntaxAdapter` constructor, and `member.IsKind(...)` is called before the `new LocalFunctionStatementAdapter((LocalFunctionStatementSyntaxWrapper)member))` call anyways

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/TypeDeclarationSyntaxExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+using SonarAnalyzer.ShimLayer.CSharp;
+
+namespace SonarAnalyzer.Helpers
+{
+    public static class TypeDeclarationSyntaxExtensions
+    {
+        /// <summary>
+        /// Returns a union of all the methods and local functions from a given type declaration.
+        /// </summary>
+        public static IEnumerable<IMethodDeclaration> GetMethodDeclarations(this TypeDeclarationSyntax typeDeclaration) =>
```
christophe-zurn-sonarsource: It might me overkill, but It would be nice to have unit test ensuring this new method helper logic works as intended

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/TypeDeclarationSyntaxExtensions.cs
costin-zaharia-sonarsource: I've updated both adapters to be private nested classes and since they are used only here I can safely remove the null checks.

costin-zaharia-sonarsource:
> I guess the goal here is to reuse this adapter logic of IMethodDeclaration to all other local function that need to be updated, right ?

Indeed, since there is no common base class or interface for **LocalFunctionStatementSyntaxWrapper** and **MethodDeclarationSyntax** we can use this interface and the related adapters and I think this is generic enough to be used in other places where an abstraction is needed.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/TypeDeclarationSyntaxExtensions.cs
```diff
@@ -0,0 +1,74 @@
+/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2019 SonarSource SA
```
christophe-zurn-sonarsource: I've just merged the license header update PR. We should now use `Copyright (C) 2015-2020 SonarSource SA` for the 3 new files in this PR

PR #3019, opened 2020-01-10
Happy New Year 2020!

### LINKED ISSUE #2952: Update license headers for 2020
Happy new year!

We need to update the license headers (for both C# and Java code)

pavel-mikula-sonarsource:
Fixes #2952


pavel-mikula-sonarsource:
I've restored the comments and replaced with fake message and original number of lines. Original version was FP when looking for outdated headers.

Issue #3007, opened 2020-01-09
Fix S1451 - this rule is not triggered on our project on sonarcloud


andrei-epure-sonarsource:
### Description

Rule S1451 (Track license header comment) does not get triggered on our codebase - for example it should trigger on https://github.com/SonarSource/sonar-dotnet/blob/fb779a247f4e9e00b7268edd7f764214768f8dc6/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/TooManyParametersBase.cs#L3 because it does not have the correct year

We have the SonarLint.xml file https://github.com/SonarSource/sonar-dotnet/blob/019284e8065c177ac4773d9e954eb04cb869c27e/sonaranalyzer-dotnet/.sonarlint/SonarLint.xml#L17 which should help with that and apply locally the changes

andrei-epure-sonarsource:
I am closing this issue because S1451 is not in SonarWay and currently we are using the SonarWay Quality Profile on SonarCloud.

We'll create a new QP which inherits from SonarWay and include S1451 in it

PR #3022, opened 2020-01-13
Update sonarlint metafiles; fix path to SonarLint.xml in csproj files

andrei-epure-sonarsource:
Related to #3007 

This contains changes:
- autogenerated files from binding SonarLint to SonarCloud (we have new rules in SonarWay since the last update in May 2019)
- fixing the path to the local SonarLint.xml

I tested the usage of SonarLint.xml in projects with a local SQ instance with S1451 enables in the QP. It works 😄

### DIFF
--- a/sonaranalyzer-dotnet/.sonarlint/SonarAnalyzer.slconfig
+++ b/sonaranalyzer-dotnet/.sonarlint/SonarAnalyzer.slconfig
@@ -9,7 +9,7 @@
   "Profiles": {
     "CSharp": {
       "ProfileKey": "AVxYKDdrFTbgxqUNcakm",
-      "ProfileTimestamp": "2019-05-07T17:01:40+02:00"
+      "ProfileTimestamp": "2020-01-10T11:36:33+01:00"
     }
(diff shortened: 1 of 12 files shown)


costin-zaharia-sonarsource:
As far as I can tell both **S2278** and **S3693** were removed from the analysis because they are deprecated.

andrei-epure-sonarsource:
actually it seems the correct rule parameter didn't get applied in the S4MSB analysis... 

it works only inside VS ... (as in the `.sonarlint/SonarLint.xml` file is correctly used by Visual Studio)

andrei-epure-sonarsource:
and this is actually because our https://github.com/SonarSource/sonar-dotnet/blob/7f54c38bf941bdcee3f31915aba81fff56e8e53a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/PropertiesHelper.cs#L44 takes the first SonarLint.xml file it finds

to fix this, we should allow local checked in configuration ... and we need to clarify if that's what we want

Issue #2936, opened 2019-12-27
Extend CSharpExplodedGraph with support for DeclarationExpressionSyntax


costin-zaharia-sonarsource:
### Description

Currently the `CSharpExplodedGraph` is ignoring `DeclarationExpressionSyntax` nodes and there are no symbolic values created for the declared variables.

For example, in case of: `dictionary.TryGetValue(key, out var value);`, the variable **value** should be handled and added to the list of values.

Additionally we should add constraints for the new symbolic value.

PR #3023, opened 2020-01-13
Push SymbolicValue for DeclarationExpression

costin-zaharia-sonarsource:
- this is a temporary fix until #2936 is done

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/EmptyCollectionsShouldNotBeEnumerated.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/EmptyCollectionsShouldNotBeEnumerated.cs
@@ -28,6 +28,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.ShimLayer.CSharp;
 using SonarAnalyzer.SymbolicExecution;
 using SonarAnalyzer.SymbolicExecution.Constraints;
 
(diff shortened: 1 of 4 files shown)


costin-zaharia-sonarsource:
It turned out that also "Method(out var _)" and "Method(out _)" were not handled correctly. I've changed the code to not retrieve symbols if the argument is a **DiscardDesignation**.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/EmptyCollectionsShouldNotBeEnumerated.cs
```diff
@@ -353,6 +355,18 @@ private CollectionCapacityConstraint GetInitializerConstraint(InitializerExpress
...
+            {
+                // e.g. Method(out var _)
+                if (DeclarationExpressionSyntaxWrapper.IsInstance(argumentSyntax.Expression) &&
+                    ((DeclarationExpressionSyntaxWrapper)argumentSyntax.Expression).Designation.SyntaxNode.IsKind(SyntaxKindEx.DiscardDesignation))
+                {
+                    return false;
+                }
+
+                return true;
+            }
```
christophe-zurn-sonarsource: (personal preference) I'd use the sorthand version instead:
```
private static bool HasAssociatedSymbolicValue(ArgumentSyntax argumentSyntax) =>
    !DeclarationExpressionSyntaxWrapper.IsInstance(argumentSyntax.Expression) ||
      !((DeclarationExpressionSyntaxWrapper)argumentSyntax.Expression).Designation.SyntaxNode.IsKind(SyntaxKindEx.DiscardDesignation)  
```

costin-zaharia-sonarsource: If it's fine with you I would leave it as it is since I find it a little bit harder to read due to the negation.

PR #3027, opened 2020-01-13
Rule S100 - update to support static local functions

### LINKED ISSUE #2984: Rule S100 - update to support static local functions
### Description

christophe-zurn-sonarsource:
Fixes #2984


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs
```diff
@@ -85,6 +87,14 @@ protected override void Initialize(SonarAnalysisContext context)
                 },
                 SyntaxKind.MethodDeclaration);
 
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
+                {
+                    var declaration = (LocalFunctionStatementSyntaxWrapper)c.Node;
+                    CheckMemberName(declaration, declaration.Identifier, c);
+                },
+                SyntaxKindEx.LocalFunctionStatement);
```
costin-zaharia-sonarsource: In order to remove the code duplication we could merge the 3 calls of RegisterSyntaxNodeActionInNonGenerated and extract a function which will return the identifier based on the node type.

PR #3036, opened 2020-01-14
Add support for tuple assignment

costin-zaharia-sonarsource:

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
@@ -564,6 +564,8 @@ namespace SonarAnalyzer.SymbolicExecution
                     var variableSymbol = SemanticModel.GetDeclaredSymbol(variable);
                     var variableSymbolicValue = SymbolicValue.Create(variableSymbol.GetSymbolType());
 
+                    // No symbolic values are created and pushed to the stack since they are not currently used.
+                    // See also VisitSimpleAssignment where ParenthesizedVariableDesignationSyntaxWrapper are handled.
                     programState = programState.StoreSymbolicValue(variableSymbol, variableSymbolicValue);
                 }
 
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
@@ -555,6 +555,21 @@ namespace SonarAnalyzer.SymbolicExecution
         {
             // DeclarationExpression is not yet fully supported: https://github.com/SonarSource/sonar-dotnet/issues/2936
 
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
```diff
@@ -1139,8 +1154,19 @@ private ProgramState VisitOpAssignment(AssignmentExpressionSyntax assignment, Pr
 
         private ProgramState VisitSimpleAssignment(AssignmentExpressionSyntax assignment, ProgramState programState)
         {
+            if (DeclarationExpressionSyntaxWrapper.IsInstance(assignment.Left))
```
andrei-epure-sonarsource: are there other places where `AssignmentExpressionSyntax` gets processed and a similar logic should be emplyed?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
```diff
@@ -555,6 +555,21 @@ private ProgramState VisitDeclarationExpression(DeclarationExpressionSyntaxWrapp
         {
             // DeclarationExpression is not yet fully supported: https://github.com/SonarSource/sonar-dotnet/issues/2936
 
+            if (ParenthesizedVariableDesignationSyntaxWrapper.IsInstance(wrapper.Designation))
+            {
+                var designation = (ParenthesizedVariableDesignationSyntaxWrapper)wrapper.Designation;
+
+                foreach (var variable in designation.Variables)
```
andrei-epure-sonarsource: I would add a comment with `<cref>` to point to `VisitSimpleAssignment` - saying that we don't push to the stack because we don't consume them later on

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
costin-zaharia-sonarsource: A `AssignmentExpressionSyntax` is used in the following cases:
```
SyntaxKind.SubtractAssignmentExpression:
SyntaxKind.AddAssignmentExpression:
SyntaxKind.DivideAssignmentExpression:
SyntaxKind.MultiplyAssignmentExpression:
SyntaxKind.ModuloAssignmentExpression:
SyntaxKind.LeftShiftAssignmentExpression:
SyntaxKind.RightShiftAssignmentExpression:
SyntaxKind.OrAssignmentExpression
SyntaxKind.AndAssignmentExpression
SyntaxKind.ExclusiveOrAssignmentExpression
```
but a tuple cannot be part of these.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
costin-zaharia-sonarsource: I've added a comment but I didn't used `<cref>` since it's meant for use at the method level and the comment makes sense only for this particular case.

PR #3040, opened 2020-01-15
No need to add symbolic values for DiscardDesignation

costin-zaharia-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -976,6 +976,60 @@ public void Main(Loader loader)
...
+
+    public class Test
+    {
+        public void Main(Loader loader)
+        {
+            var (projectInstance, _) = loader;
+        }
+    }
+}";
+            var context = new ExplodedGraphContext(TestHelper.Compile(testInput));
```
christophe-zurn-sonarsource: Note that for simplicity you could replace the full test input by only:
```
            const string testInput = "var (projectInstance, _) = loader;";
            var context = new ExplodedGraphContext(testInput);
```
as there is no semantic check for the compilation

costin-zaharia-sonarsource: Nice! It simplifies the code quite a bit.

PR #3043, opened 2020-01-15
Create SymbolicValue for DiscardDesignation

costin-zaharia-sonarsource:
- in order to be able to correctly handle the method invocations which
have as argument a DiscardDesignation
- update CFG to generate instructions for Dictionary
- add test

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -471,12 +471,7 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
                     return BuildJumpToExitStatement(throwExpression, currentBlock, throwExpression.Expression);
 
                 case SyntaxKindEx.DeclarationExpression:
-                    // we ignore 'discard' parameters
-                    var declaration = (DeclarationExpressionSyntaxWrapper)expression;
-                    if (!declaration.Designation.SyntaxNode.IsKind(SyntaxKindEx.DiscardDesignation))
-                    {
-                        currentBlock.ReversedInstructions.Add(expression);
-                    }
+                    currentBlock.ReversedInstructions.Add(expression);
                     return currentBlock;
 
                 case SyntaxKindEx.RefExpression:
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
@@ -562,7 +562,7 @@ namespace SonarAnalyzer.SymbolicExecution
 
             if (DiscardDesignationSyntaxWrapper.IsInstance(wrapper.Designation))
             {
-                return VisitDiscardDesignationSyntax((DiscardDesignationSyntaxWrapper)wrapper.Designation, programState);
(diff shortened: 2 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -471,12 +471,7 @@ private Block BuildExpression(ExpressionSyntax expression, Block currentBlock)
                     return BuildJumpToExitStatement(throwExpression, currentBlock, throwExpression.Expression);
 
                 case SyntaxKindEx.DeclarationExpression:
-                    // we ignore 'discard' parameters
-                    var declaration = (DeclarationExpressionSyntaxWrapper)expression;
-                    if (!declaration.Designation.SyntaxNode.IsKind(SyntaxKindEx.DiscardDesignation))
-                    {
-                        currentBlock.ReversedInstructions.Add(expression);
-                    }
```
costin-zaharia-sonarsource: This will affect also **sonar-security**. I will do a separate commit for handling the update.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
```diff
@@ -569,6 +574,15 @@ private ProgramState VisitDeclarationExpression(DeclarationExpressionSyntaxWrapp
             return programState;
         }
 
+        private static ProgramState VisitDiscardDesignationSyntax(DiscardDesignationSyntaxWrapper wrapper, ProgramState programState)
+        {
+            // In case of discard designation we push a new symbolic value with no constrains.
```
andrei-epure-sonarsource:
```suggestion
            // In case of discard designation we push a new symbolic value with no constraints.
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -990,6 +990,52 @@ public void ExplodedGraph_TupleExpressionsWithDiscardDeconstruct()
...
+            context.ExplodedGraph.InstructionProcessed +=
+                (sender, args) =>
+                {
+                    var instruction = args.Instruction.ToString();
+
+                    switch (instruction)
+                    {
+                        case "obj":
+                            args.ProgramState.GetSymbolValue(resultSymbol).Should().BeNull();
+                            args.ProgramState.ExpressionStack.Peek().ToString().Should().Be("SV_2"); // obj => SV_2
```
andrei-epure-sonarsource: what is SV_2 , SV_2.Read?

costin-zaharia-sonarsource: SV_2 is the "obj" identifier and "SV_2.Read" is "obj.Read". I've put some comments at the end of the line but since the line is quite long they are not easily visible on github.

costin-zaharia-sonarsource: I had to remove them in the end since the symbolic value counter is static and we will get different values depending on the order of the tests.

andrei-epure-sonarsource: > I've put some comments at the end of the line
well... I thought we said we wouldn't do that on long lines :)

PR #3044, opened 2020-01-15
Update S1301 with support for `switch` expressions

### LINKED ISSUE #2994: Rule S1301 - support switch expressions, Property patterns, Tuple patterns, Positional patterns
### Description
https://jira.sonarsource.com/browse/RSPEC-1301
Add tests with  switch expressions, Property patterns, Tuple patterns, Positional patterns and fix implementation

andrei-epure-sonarsource:
Fixes #2994

### DIFF
--- a/sonaranalyzer-dotnet/rspec/cs/S1301_c#.html
+++ b/sonaranalyzer-dotnet/rspec/cs/S1301_c#.html
@@ -1,5 +1,5 @@
-<p><code>switch</code> statements are useful when there are many different cases depending on the value of the same expression.</p>
-<p>For just one or two cases however, the code will be more readable with <code>if</code> statements.</p>
+<p><code>switch</code> statements and expressions are useful when there are many different cases depending on the value of the same expression.</p>
+<p>For just one or two cases however, the code will be more readable using <code>if</code> statements or the ternary operator instead.</p>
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/rspec/cs/S1301_c#.html
```diff
@@ -22,5 +29,20 @@ <h2>Compliant Solution</h2>
 {
   doSomethingElse();
 }
+
+var foo = variable == 0
+  ? doSomething()
+  : doSomethingElse();
+
+</pre>
+<h2>Exceptions</h2>
```
nicolas-harraudeau-sonarsource: Adding an exception makes the rule description more complex. How about simplifying the description above so that we don't need to add the exception?
Example
```
<p><code>switch</code> statements and expressions are useful when there are many different cases depending on the value of the same expression.</p>
<p>When a switch statement is simple enough to be replaced with a single `if`, `if .. else` or ternary operation, one of these more readable forms should be used instead.</p>
```

andrei-epure-sonarsource: thanks I updated https://jira.sonarsource.com/browse/RSPEC-2828

PR #3046, opened 2020-01-16
Rule S2997 - support Disposable ref structs and using declarati…

### LINKED ISSUE #3002: Rule S2997 - support Disposable ref structs and using declarations
### Description

add tests and update implementation

christophe-zurn-sonarsource:
Fixes #3002

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableReturnedFromUsing.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableReturnedFromUsing.cs
@@ -50,10 +50,10 @@ namespace SonarAnalyzer.Rules.CSharp
                 {
                     var usingStatement = (UsingStatementSyntax) c.Node;
                     var declaration = usingStatement.Declaration;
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableReturnedFromUsing.cs
```diff
@@ -57,47 +58,67 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            }
+
+            var returnedSymbols = GetReturnedSymbols(body, c.SemanticModel);
+            returnedSymbols.IntersectWith(declaredSymbols);
+
+            if (returnedSymbols.Any())
+            {
+                c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, usingKeyword.GetLocation(),
+                    string.Join(", ",
+                        returnedSymbols.Select(s => $"'{s.Name}'").OrderBy(s => s))));
```
andrei-epure-sonarsource: could you add a test where this concatenation is used? like more symbols in the issue message?

PR #3054, opened 2020-01-17
Rule S3263 - update to support default interface members

### LINKED ISSUE #2992: Rule S3263 - update to support default interface members
### Description

Add tests and fix implementation

andrei-epure-sonarsource:
Fixes #2992


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/StaticFieldInitializerOrder.cs
```diff
@@ -49,7 +49,8 @@ protected override void Initialize(SonarAnalysisContext context)
                 {
                     var fieldDeclaration = (FieldDeclarationSyntax)c.Node;
                     var variables = fieldDeclaration.Declaration.Variables;
-                    var classDeclaration = fieldDeclaration.FirstAncestorOrSelf<ClassDeclarationSyntax>();
+                    var typeDeclaration = fieldDeclaration.FirstAncestorOrSelf<TypeDeclarationSyntax>(
```
andrei-epure-sonarsource: the relevant change is this line

all the other changes in the file are refactorings
- change name from *Class* to *Type*
- use `TypeDeclarationSyntax` instead of `ClassDeclarationSyntax`

PR #3059, opened 2020-01-20
Rule S2326 - update to support static local functions

### LINKED ISSUE #2977: Rule S2326 - update to support static local functions
### Description

https://jira.sonarsource.com/browse/RSPEC-2326

costin-zaharia-sonarsource:
Fixes #2977

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterUnused.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterUnused.cs
@@ -27,6 +27,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.ShimLayer.CSharp;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
@@ -34,9 +35,8 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class GenericTypeParameterUnused : SonarDiagnosticAnalyzer
     {
-        internal const string DiagnosticId = "S2326";
+        private const string DiagnosticId = "S2326";
         private const string MessageFormat = "'{0}' is not used in the {1}.";
-
 
         private static readonly DiagnosticDescriptor rule =
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
@@ -52,6 +52,7 @@ namespace SonarAnalyzer.Rules.CSharp
                     {
                         var methodDeclaration = c.Node as MethodDeclarationSyntax;
                         var classDeclaration = c.Node as ClassDeclarationSyntax;
+                        var localFunctionDeclaration = LocalFunctionStatementSyntaxWrapper.IsInstance(c.Node) ? (LocalFunctionStatementSyntaxWrapper) c.Node : default;
 
                         if (methodDeclaration != null &&
                             !IsMethodCandidate(methodDeclaration, c.SemanticModel))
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterUnused.cs
```diff
@@ -52,6 +52,7 @@ protected override void Initialize(SonarAnalysisContext context)
                     {
                         var methodDeclaration = c.Node as MethodDeclarationSyntax;
```
andrei-epure-sonarsource: I find this code quite confusing, I think it's a good time to refactor.
Lines 53-72 have a lot of local variables which are not used after line 73: `methodDeclaration`, `classDeclaration` and now `localFunctionDeclaration`. They're only useful to get the `helper` which is later used.
So maybe it's a good hint to extract a method to get the `helper` based on the `c.Node`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterUnused.cs
```diff
@@ -65,7 +66,10 @@ protected override void Initialize(SonarAnalysisContext context)
                             return;
                         }
 
-                        var helper = GetTypeParameterHelper(methodDeclaration, classDeclaration);
+                        // For local methods the only valid modifiers are async, static and unsafe
+                        // https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions#local-function-syntax
+
+                        var helper = GetTypeParameterHelper(methodDeclaration, classDeclaration, localFunctionDeclaration);
```
andrei-epure-sonarsource: maybe rename `helper` to `typeParameterHelper` or `typeParameterInformation` ?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/GenericTypeParameterUnused.cs
```diff
@@ -122,17 +123,28 @@ public class WithLocalFunctions
     {
         public void Method()
         {
-            DoMoreStuff<int, string>("value");
+            void LocalFunctionNoTypeParameter() { }
 
-            T3 DoStuff<T, T1, T3>(params T3[] o) // Compliant - FN: local functions are not verified
+            void LocalFunctionUnusedParameter<T>() { } // Noncompliant
+
+            T LocalFunctionWithUsedTypeParameter<T>(T p) => p;
```
andrei-epure-sonarsource: from what I can see, all compliant tests have the type parameter as both return type and parameters type

it would be useful to have a test with only return
```
T LocalFunctionWithUsedTypeParameter<T>(object o) => (T)o;
```
and a test with only parameter
```
void LocalFunctionWithUsedTypeParameter<T>(T o) => o;
```

and also a test with many parameters and different combinations that might "break" the algorithm
```
static void StaticLocalFunctionWithUnusedTypeParameter<T1, T2, T3, T4, T5>(T1, T5, T5, T1) // Noncompliant - should mention 2 parameters which are not used : T2, T3 and T4

static void StaticLocalFunctionWithUnusedTypeParameter<T1, T2, T3, T4>(T4, T3, T2, T1)
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterUnused.cs
costin-zaharia-sonarsource: Yes, the naming (already existing) is really unfortunate. I didn't want to change it to ease the code review.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterUnused.cs
costin-zaharia-sonarsource: Have you checked all commits? This was handled already in 2a2c938441beb24aa54be4973f4c9a246e1d1073.

PR #3062, opened 2020-01-20
Rule S2952 - support Disposable ref structs

### LINKED ISSUE #2998: Rule S2952 - support Disposable ref structs
### Description

add tests and update implementation

https://jira.sonarsource.com/browse/RSPEC-2952

andrei-epure-sonarsource:
Fixes #2998

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpCompilationHelper.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpCompilationHelper.cs
@@ -26,7 +26,7 @@ namespace SonarAnalyzer.Helpers
     internal static class CSharpCompilationHelper
     {
         internal static bool IsAtLeastLanguageVersion(this Compilation compilation, LanguageVersion languageVersion) =>
-            compilation.GetLanguageVersion().CompareTo(languageVersion) >= 0;
+            compilation.GetLanguageVersion().IsAtLeast(languageVersion);
 
         internal static LanguageVersion GetLanguageVersion(this Compilation compilation) => ((CSharpCompilation) compilation).LanguageVersion;
     }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpCompilationHelper.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpCompilationHelper.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
@@ -23,9 +23,11 @@ using Microsoft.CodeAnalysis.CSharp;
 
 namespace SonarAnalyzer.Helpers
 {
-    public static class CSharpCompilationHelper
+    internal static class CSharpCompilationHelper
     {
-        public static bool IsAtLeastLanguageVersion(this Compilation compilation, LanguageVersion languageVersion) =>
-            (compilation as CSharpCompilation)?.LanguageVersion.CompareTo(languageVersion) >= 0;
+        internal static bool IsAtLeastLanguageVersion(this Compilation compilation, LanguageVersion languageVersion) =>
+            compilation.GetLanguageVersion().CompareTo(languageVersion) >= 0;
+
+        internal static LanguageVersion GetLanguageVersion(this Compilation compilation) => ((CSharpCompilation) compilation).LanguageVersion;
     }
 }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/ITypeSymbolExtensions.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/ITypeSymbolExtensions.cs
@@ -28,8 +28,8 @@ namespace SonarAnalyzer.Helpers
 {
     internal static class ITypeSymbolExtensions
     {
-        internal static bool IsDisposableRefStruct(this ITypeSymbol symbol, Compilation compilation) =>
-            compilation.IsAtLeastLanguageVersion(LanguageVersionEx.CSharp8) &&
+        internal static bool IsDisposableRefStruct(this ITypeSymbol symbol, LanguageVersion languageVersion) =>
+            languageVersion.IsAtLeast(LanguageVersionEx.CSharp8) &&
             IsRefStruct(symbol) &&
             symbol.GetMembers("Dispose").Any(
                 s => s is IMethodSymbol disposeMethod &&
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/ITypeSymbolExtensions.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/ITypeSymbolExtensions.cs
@@ -1,0 +1,45 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2020 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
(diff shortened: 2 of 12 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposeFromDispose.cs
```diff
@@ -84,17 +92,32 @@ private static bool IsDisposeMethodCalled(InvocationExpressionSyntax invocation,
                 methodSymbol.Equals(methodSymbol.ContainingType.FindImplementationForInterfaceMember(disposeMethod));
         }
 
-        private static bool IsDisposableField(ExpressionSyntax expression, SemanticModel semanticModel)
```
andrei-epure-sonarsource: this method was previously verifying both that it's a disposable field and that the containing type implemented IDisposable, so I split it

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposeFromDispose.cs
```diff
@@ -84,17 +97,18 @@ private static bool IsDisposeMethodCalled(InvocationExpressionSyntax invocation,
...
-        private static bool IsMethodMatchingDisposeMethodName(IMethodSymbol enclosingMethodSymbol)
-        {
-            return enclosingMethodSymbol.Name == DisposeMethodName ||
-                enclosingMethodSymbol.ExplicitInterfaceImplementations.Any() && enclosingMethodSymbol.Name == DisposeMethodExplicitName;
-        }
+        private static bool ImplementsDisposable(INamedTypeSymbol containingType) =>
+            containingType.Implements(KnownType.System_IDisposable);
+
+        private static bool IsDisposableRefStruct(Compilation compilation, INamedTypeSymbol containingType) =>
+            compilation.IsAtLeastLanguageVersion(LanguageVersionEx.CSharp8) &&
```
andrei-epure-sonarsource: this doesn't test if it's a struct

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposeFromDispose.cs
```diff
@@ -50,28 +52,39 @@ protected override void Initialize(SonarAnalysisContext context)
...
+
+        private static bool IsCalledInsideDispose(InvocationExpressionSyntax invocation, SemanticModel semanticModel) =>
+            semanticModel.GetEnclosingSymbol(invocation.SpanStart) is IMethodSymbol enclosingMethodSymbol &&
+            IsMethodMatchingDisposeMethodName(enclosingMethodSymbol);
+
+        private Location GetLocation(InvocationExpressionSyntax invocation) =>
+            // We already did the type check before
+            ((MemberAccessExpressionSyntax)invocation.Expression).Name.GetLocation();
+
         private static bool IsDisposeMethodCalled(InvocationExpressionSyntax invocation, SemanticModel semanticModel)
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: yes, only in C# 8
I'll look to see if we can get that from the symbol (if it's not an implicit declaration of the struct)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/ITypeSymbolExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace SonarAnalyzer.Helpers
+{
+    internal static class ITypeSymbolExtensions
+    {
+        internal static bool IsDisposableRefStruct(this ITypeSymbol symbol) =>
+            symbol != null &&
```
christophe-zurn-sonarsource: _minor_: this null check is not needed since we are calling `IsRefStruct(symbol)` right after and the null check is also done there

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposeFromDispose.cs
```diff
@@ -93,7 +92,10 @@ private static bool IsDisposeMethodCalled(InvocationExpressionSyntax invocation,
 
             var disposeMethod = DisposeNotImplementingDispose.GetDisposeMethod(semanticModel.Compilation);
             return disposeMethod != null &&
-                methodSymbol.Equals(methodSymbol.ContainingType.FindImplementationForInterfaceMember(disposeMethod));
+            (
+                methodSymbol.Equals(methodSymbol.ContainingType.FindImplementationForInterfaceMember(disposeMethod)) ||
+                methodSymbol.ContainingType.IsDisposableRefStruct()
```
christophe-zurn-sonarsource: We should check for the method symbol signature here too in case of ref structs, otherwise we will get FP:
```
    public ref struct MyTestDisposableRefStruct
    {
        public void Dispose(){}
        public void AnotherMethod(){}
    }

    public ref struct RefStructResourceHolder
    {
        // ref structs can only be fields in other ref structs
        private MyTestDisposableRefStruct foo;

        public void CleanUp()
        {
            this.foo.AnotherMethod(); // FP
        }

        public void Dispose()
        {}
    }
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/ITypeSymbolExtensions.cs
andrei-epure-sonarsource: good point!

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposeFromDispose.cs
andrei-epure-sonarsource: you saved the day!

andrei-epure-sonarsource:
@christophe-zurn-sonarsource  - I added 3 commits since your review (I had to rebase and it's not easy to follow the history now)

andrei-epure-sonarsource:
@christophe-zurn-sonarsource actually it still does not work with
```
    public ref struct AnotherDisposableRefStruct
    {
        public void Dispose() { }
        public void Dispose(bool x) { }
        public void Dispose(string y) { }
        public void AnotherMethod() { }
    }

    public ref struct AnotherDisposableRefStructHolder1
    {
        private AnotherDisposableRefStruct foo;

        public void Cleanup()
        {
            this.foo.Dispose(); // Noncompliant
        }

        public void Dispose()
        {
            foo.Dispose(true);
        }
    }

    public ref struct AnotherDisposableRefStructHolder2
    {
        private AnotherDisposableRefStruct foo;

        public void Cleanup()
        {
            this.foo.AnotherMethod(); // FP
        }

        public void Dispose()
        {
        }
    }

    public ref struct AnotherDisposableRefStructHolder3
    {
        private AnotherDisposableRefStruct foo;

        public void Cleanup()
        {
            this.foo.Dispose(true); // FP
        }

        public void Dispose()
        {
        }
    }

```

andrei-epure-sonarsource:
Since your review, these commits have been added:
f2badc71b Address CR and fix introduced FP
2c629915f Avoid passing Compilation just for LanguageVersion
53880dca4 More cleanup
69868588e Fix remaining FPs

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DisposeFromDispose.cs
```diff
@@ -104,11 +104,231 @@ public void CloseResource()
...
+        }
+    }
+
+
+
+    public ref struct AnotherDisposableRefStruct
+    {
+        public void Dispose() { }
+        public void Dispose(bool x) { }
+        public void Dispose(string y) { }
```
andrei-epure-sonarsource: I didn't add tests for `public void Dispose<T>() { }` (with different type parameter arity) , but I think that's fine, I wouldn't go through another iteration

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpCompilationHelper.cs
```diff
@@ -23,9 +23,11 @@
 
 namespace SonarAnalyzer.Helpers
 {
-    public static class CSharpCompilationHelper
+    internal static class CSharpCompilationHelper
     {
-        public static bool IsAtLeastLanguageVersion(this Compilation compilation, LanguageVersion languageVersion) =>
-            (compilation as CSharpCompilation)?.LanguageVersion.CompareTo(languageVersion) >= 0;
+        internal static bool IsAtLeastLanguageVersion(this Compilation compilation, LanguageVersion languageVersion) =>
+            compilation.GetLanguageVersion().CompareTo(languageVersion) >= 0;
```
christophe-zurn-sonarsource: You could reuse the `IsAtLeast` extension method you declared in `LanguageVersionExtensions` here

PR #3068, opened 2020-01-27
Rule S2857 - update to support interpolated string syntax

### LINKED ISSUE #2986: Rule S2857 - update to support interpolated string syntax
### Description

`$""`, `$@""` and `@$""` are not yet handled

____________________________________________________________

LATER EDIT:
Looking better at the rule, indeed the scope is outside of the "support for C#8" features.

The idea would be to support interpolated strings, and not verbatim interpolated strings - because a verbatim string contains newlines inside and I think it's overkill to analyze all SQL verbatim strings. And anyway, it's quite a big change.

I see two separate opportunities to improve the rule:

- Support interpolated strings
- Support verbatim strings (simple and interpolated) - this would mean to parse the verbatim string and analyze the newlines inside it, as well as prepared queries.

christophe-zurn-sonarsource:
Fixes #2986

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SqlKeywordsDelimitedBySpace.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SqlKeywordsDelimitedBySpace.cs
@@ -133,13 +133,11 @@ namespace SonarAnalyzer.Rules.CSharp
             {
                 if (node.IsKind(SyntaxKind.AddExpression) &&
                     // we do the analysis only if it's a SQL keyword on the left
-                    node.Left is LiteralExpressionSyntax leftSide &&
-                    node.Right is LiteralExpressionSyntax rightSide &&
-                    leftSide.IsKind(SyntaxKind.StringLiteralExpression) &&
-                    rightSide.IsKind(SyntaxKind.StringLiteralExpression) &&
-                    StartsWithSqlKeyword(leftSide.Token.ValueText.Trim()))
+                    TryGetStringWrapper(node.Left, out var leftSide) &&
+                    TryGetStringWrapper(node.Right, out var rightSide) &&
+                    StartsWithSqlKeyword(leftSide.Text.Trim()))
                 {
-                    var strings = new List<LiteralExpressionSyntax>();
+                    var strings = new List<StringWrapper>();
                     strings.Add(leftSide);
                     strings.Add(rightSide);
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SqlKeywordsDelimitedBySpace.cs
```diff
@@ -157,41 +155,58 @@ public override void VisitBinaryExpression(BinaryExpressionSyntax node)
...
             }
 
+            private static bool TryGetStringWrapper(ExpressionSyntax expression, out StringWrapper stringWrapper)
+            {
+                if (expression is LiteralExpressionSyntax literal && literal.IsKind(SyntaxKind.StringLiteralExpression))
+                {
+                    stringWrapper = new StringWrapper(literal, literal.Token.ValueText);
+                    return true;
+                }
+                else if (expression is InterpolatedStringExpressionSyntax interpolatedString)
```
costin-zaharia-sonarsource: I think we can remove the else here since we do a direct return in the first if.

An option would be to use a switch here since we already have three branches but this is just a mater of preferences.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SqlKeywordsDelimitedBySpace.cs
```diff
@@ -208,10 +223,22 @@ private static bool AddStringsToList(BinaryExpressionSyntax node, List<LiteralEx
...
             /**
-             * The '@' symbol is used for named parameters.
+             * The '@' symbol is used for named parameters. The '{' and '}' symbols are used in string interpolations.
              * We ignore other non-alphanumeric characters (e.g. '>','=') to avoid false positives.
              */
-            private static bool IsAlphaNumericOrAt(char c) => char.IsLetterOrDigit(c) || c == '@';
+            private static bool IsAlphaNumericOrAt(char c) => char.IsLetterOrDigit(c) || c == '@' || c == '{' || c == '}';
+        }
+
+        internal class StringWrapper
```
costin-zaharia-sonarsource: Any reason not to make this class private?

christophe-zurn-sonarsource: No, indeed. Updated class to private

PR #3069, opened 2020-01-28
Add constraints for values added by DeclarationExpression

### LINKED ISSUE #2936: Extend CSharpExplodedGraph with support for DeclarationExpressionSyntax

costin-zaharia-sonarsource:
Fixes #2936

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
@@ -504,7 +504,7 @@ namespace Namespace
 {
   public class DeclarationStatement
   {
-    public int Main(Dictionary<string, string> dictionary, string key)
+    public int Main(Dictionary<string, int> dictionary, string key)
     {
         dictionary.TryGetValue(key, out var value);
     }
@@ -512,8 +512,7 @@ namespace Namespace
 }";
             var context = new ExplodedGraphContext(TestHelper.Compile(testInput));
             var dictionarySymbol = context.SemanticModel.GetDeclaredSymbol(context.MainMethod.ParameterList.Parameters.First());
-            var valueDeclaration = context.MainMethod.DescendantNodes().OfType<DeclarationExpressionSyntax>().First();
-            var valueSymbol = context.SemanticModel.GetDeclaredSymbol(valueDeclaration);
+            var valueSymbol = context.GetSymbol("value", ExplodedGraphContext.SymbolType.Declaration);
             context.ExplodedGraph.InstructionProcessed +=
                 (sender, args) =>
                 {
@@ -538,15 +537,19 @@ namespace Namespace
                         case "var value":
                             args.ProgramState.GetSymbolValue(dictionarySymbol).Should().NotBeNull();
                             dictionarySymbol.HasConstraint(ObjectConstraint.NotNull, args.ProgramState).Should().BeTrue();
+
+                            args.ProgramState.GetSymbolValue(valueSymbol).Should().NotBeNull();
+                            valueSymbol.HasConstraint(ObjectConstraint.NotNull, args.ProgramState).Should().BeTrue();
                             break;
 
                         case "dictionary.TryGetValue(key, out var value)":
                             args.ProgramState.GetSymbolValue(dictionarySymbol).Should().NotBeNull();
                             dictionarySymbol.HasConstraint(ObjectConstraint.NotNull, args.ProgramState).Should().BeTrue();
 
-                            // Currently the DeclarationExpressionSyntax are ignored so the "value" variable is not
-                            // https://github.com/SonarSource/sonar-dotnet/issues/2936
-                            args.ProgramState.GetSymbolValue(valueSymbol).Should().BeNull();
+                            args.ProgramState.GetSymbolValue(valueSymbol).Should().NotBeNull();
+                            valueSymbol.HasConstraint(ObjectConstraint.NotNull, args.ProgramState).Should().BeTrue();
+
+                            args.ProgramState.HasValue.Should().BeFalse();
                             break;
                     }
                 };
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -495,7 +495,7 @@ public void ExplodedGraph_StaticLocalFunctions()
 
         [TestMethod]
         [TestCategory("Symbolic execution")]
-        public void ExplodedGraph_DeclarationStatementVisit()
+        public void ExplodedGraph_DeclarationExpressionVisit_AsOutParameter()
```
andrei-epure-sonarsource: I think the name should mention that this is testing a value type declaration

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -504,16 +504,15 @@ namespace Namespace
 {
   public class DeclarationStatement
   {
-    public int Main(Dictionary<string, string> dictionary, string key)
+    public int Main(Dictionary<string, int> dictionary, string key)
```
andrei-epure-sonarsource: do we still have a UT for reference type declarations? because I see here only value-type ones

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -538,22 +537,58 @@ public int Main(Dictionary<string, string> dictionary, string key)
...
             context.WalkWithInstructions(5);
         }
 
+        [TestMethod]
+        [TestCategory("Symbolic execution")]
+        public void ExplodedGraph_DeclarationExpressionVisit_AsOutParameter_Discard()
+        {
+            const string testInput = "string key = null; Dictionary<string, int> dictionary = new Dictionary<string, int>(); dictionary.TryGetValue(key, out var _); dictionary.TryGetValue(key, out _);";
+            var context = new ExplodedGraphContext(testInput);
+            var dictionarySymbol = context.SemanticModel.GetDeclaredSymbol(context.MainMethod.ParameterList.Parameters.First());
```
christophe-zurn-sonarsource: The `dictionarySymbol` here is not the actual disctionary symbol. The default main has a `bool` as parameter.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -538,22 +537,58 @@ public int Main(Dictionary<string, string> dictionary, string key)
...
+            context.ExplodedGraph.InstructionProcessed +=
+                (sender, args) =>
+                {
+                    var instruction = args.Instruction.ToString();
+                    switch (instruction)
+                    {
+                        case "dictionary.TryGetValue(key, out var _)":
+                            args.ProgramState.GetSymbolValue(dictionarySymbol).Should().NotBeNull();
+                            dictionarySymbol.HasConstraint(ObjectConstraint.NotNull, args.ProgramState).Should().BeTrue();
```
christophe-zurn-sonarsource: Not sure what this is testing here as this is not related to out parameter or discard designation

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -538,22 +537,58 @@ public int Main(Dictionary<string, string> dictionary, string key)
...
+
+                            args.ProgramState.HasValue.Should().BeFalse();
+                            break;
+
+                        case "dictionary.TryGetValue(key, out _)":
+                            args.ProgramState.GetSymbolValue(dictionarySymbol).Should().NotBeNull();
+                            dictionarySymbol.HasConstraint(ObjectConstraint.NotNull, args.ProgramState).Should().BeTrue();
+
+                            args.ProgramState.HasValue.Should().BeFalse();
```
christophe-zurn-sonarsource: Before checking that the program state has empty stack here, it would be nice to check that it has value on the stack after having evaluated the `key` parameter

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
```diff
@@ -538,22 +537,58 @@ public int Main(Dictionary<string, string> dictionary, string key)
...
+                            break;
+                    }
+                };
+
+            context.WalkWithInstructions(14);
```
christophe-zurn-sonarsource: It has already happen in the past that some cases of the switch above are not explored: it would be nice to add a counter that increments for each of the cases, so that we know for sure all of them were explored.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
costin-zaharia-sonarsource: No, indeed. I've added a new test to assert that we don't add the **NotNull** constraint for reference types.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
costin-zaharia-sonarsource: 😄 very good catch. It was a copy paste error which I've fixed now.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SymbolicExecution/ExplodedGraphTests.cs
costin-zaharia-sonarsource: Indeed this was not meant to test the parameter but the actual dictionary constraints.

PR #3078, opened 2020-01-29
Ignore RecursivePatterns inside IsPattern

### LINKED ISSUE #3077: CSharpExplodedGraph should not throw when handling Recursive Patterns in IsPattern
### Description

Running symbolic execution for RecursivePatterns inside IsPattern results in an error.

### Repro steps

```var x = address is Address { State: ""WA"" } addr;```

### Expected behavior

No exception should be thrown.

### Actual behavior

The following exception is thrown:
```
System.NotSupportedException: IsPatternExpression
    at SonarAnalyzer.SymbolicExecution.CSharpExplodedGraph.VisitInstruction(ExplodedGraphNode node) in C:\src\sonar-dotnet\sonaranalyzer-dotnet\src\SonarAnalyzer.CSharp\SymbolicExecution\CSharp\CSharpExplodedGraph.cs:line 509
   at SonarAnalyzer.SymbolicExecution.AbstractExplodedGraph.Walk() in C:\src\sonar-dotnet\sonaranalyzer-dotnet\src\SonarAnalyzer.CSharp\SymbolicExecution\Common\AbstractExplodedGraph.cs:line 111
```

### PR DESCRIPTION
**Pull request by costin-zaharia-sonarsource**
- during symbolic execution
- proper handling will be done by #2937

Fixes #3077

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
@@ -506,8 +506,7 @@ namespace SonarAnalyzer.SymbolicExecution
                     }
                     else if (RecursivePatternSyntaxWrapper.IsInstance(isPatternExpression.Pattern))
                     {
-                        var recursivePattern = (RecursivePatternSyntaxWrapper)isPatternExpression.Pattern;
-                        VisitRecursivePattern(recursivePattern);
+                        newProgramState = VisitRecursivePattern((RecursivePatternSyntaxWrapper)isPatternExpression.Pattern, newProgramState);
                     }
                     else
                     {
@@ -560,10 +559,11 @@ namespace SonarAnalyzer.SymbolicExecution
             EnqueueNewNode(newProgramPoint, newProgramState);
         }
 
-        private static void VisitRecursivePattern(RecursivePatternSyntaxWrapper recursivePattern)
+        private static ProgramState VisitRecursivePattern(RecursivePatternSyntaxWrapper recursivePattern, ProgramState programState)
         {
             // Currently RecursivePatterns are not considered during the symbolic execution
             // https://github.com/SonarSource/sonar-dotnet/issues/2937
+            return programState;
         }
 
         private ProgramState VisitDeclarationExpression(DeclarationExpressionSyntaxWrapper wrapper, ProgramState programState)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs [hunk 1]
```diff
@@ -519,12 +524,15 @@ protected override void VisitInstruction(ExplodedGraphNode node)
...
-                case SyntaxKindEx.RecursivePattern: // https://github.com/SonarSource/sonar-dotnet/issues/2937
                 case SyntaxKindEx.ConstantPattern:
                     // The 0 in 'case 0 when ...'
                     // Do nothing
                     break;
 
+                case SyntaxKindEx.RecursivePattern:
+                    VisitRecursivePattern((RecursivePatternSyntaxWrapper)instruction);
+                    break;
+
```
christophe-zurn-sonarsource: `VisitRecursivePattern` should not appear for this instruction: the goal here is to handle the recusrive pattern the same way `DeclarationPattern` is currently handled for `IsPattern`. That means that we ignore the instruction itself (for `DeclarationPattern` it does not appear as instruction in CFG), and the pattern is handled at the higher level: inside `IsPatternExpression` (as updated above), and inside `SwitchExpressionArm` for switch

costin-zaharia-sonarsource: Good point. I've removed the call from here and updated the `VisitSwitchExpressionArmBinaryBranch` to call `VisitRecursivePattern`.
It does nothing for now but at least it will make sure that we will not update just a case and forget about the other.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
```diff
@@ -551,6 +560,12 @@ protected override void VisitInstruction(ExplodedGraphNode node)
             EnqueueNewNode(newProgramPoint, newProgramState);
         }
 
+        private static void VisitRecursivePattern(RecursivePatternSyntaxWrapper recursivePattern)
```
christophe-zurn-sonarsource: Although not really important at the moment anyway, since it does nothing, but the method should theoretically take a program state and return a program state, otherwise even if some constraints are set inside this method, it will never be reflected in the following SE paths.
But as it does nothing at the moment anyway, we can leave it as is if you want

costin-zaharia-sonarsource: I think it makes sense to have the correct format. I've update the code.

Issue #3076, opened 2020-01-28
Improve LiveOut of LVA for SE


pavel-mikula-sonarsource:
### Description

Tracking of alive variables should be improved for SE and it's state clearing. Due to the shape of CFG for `??` operator the operand symbolic values are not tracked as alive out of the block. This removes them from the program state and prevents inspecting their state.

### Repro steps
```
string notNull = "";
string ret;
ret = notNull ?? "N/A";
```

When `ExplodedGraph.InstructionProcessed` event is triggered the last SimpleBlock with single instruction `return isNotNull ?? isNull;`, there is no state persisted about isNull nor isNotNull. 

### CFG shape

> digraph "" {
> 0 [shape=record label="{BINARY:CoalesceExpression|\"\"|notNull = \"\"|ret|notNull}"]
> 0 -> 1 [label="True"]
> 0 -> 2 [label="False"]
> 1 [shape=record label="{SIMPLE|\"N/A\"}"]
> 1 -> 2
> 2 [shape=record label="{SIMPLE|ret = notNull ?? \"N/A\"}"]
> 2 -> 3
> 3 [shape=record label="{EXIT}"]
> }

### Actual behavior

At the end of the binary branch block `AbstractExplodedGraph.CleanStateAfterBlock` is called. `lva.GetLiveOut` is empty so symbolic values and their constraints are removed.

This is probably caused by `CSharpLiveVariableAnalysis.ProcessBlockInternal` not processing all inner parts of instructions but only few selected types. And `usedInBlock` is not filled properly here.

### Expected behavior

Constraints should be available until the last moment of usage of a variable. Either by updating GetLiveOut behavior if DeadStore rule is not affected or by adding another list of still-alive variables.

### Known workarounds

_Ugly:_ If the variable is used one more time AFTER inspected instruction in regular way, LVA persists it's value as expected.

pavel-mikula-sonarsource:
This is no longer reproducible with S2583 due to updated implementation.

LVA issue is still valid if inspection of used variable is needed in ProcessInstruction phase, but there's no point of fixing this now.

PR #3079, opened 2020-01-29
S2583: NullCoalesceOperator ??= for ConditionEvaluatesToConstant

### LINKED ISSUE #3066: S2583: NullCoalesceOperator ??= for ConditionEvaluatesToConstant
### Description

Add support for `??=` operator to rule S2583 and S2589 and update RSPEC .

This point was removed from S3240 specification in favor of these rules:

> * removing the {{??}} when the left operand is known to be {{null}}, or known to be definitely not {{null}}.

It should be applied to SE rules and extended to ??= as well.

pavel-mikula-sonarsource:
Fixes #3066

Please review RSPEC text update as well.

I've closed the spin-off LVA #3076 because it's not relevant anymore.

### DIFF
--- a/sonaranalyzer-dotnet/rspec/cs/S2583_c#.html
+++ b/sonaranalyzer-dotnet/rspec/cs/S2583_c#.html
@@ -23,6 +23,29 @@ public void Sample(bool b)
     var res = d ?? "value"; // Noncompliant, d is always not null, "value" is never used
 }
 </pre>
+<h2>Compliant Solution</h2>
+<pre>
+public void Sample(bool b)
+{
+    bool a = false;
+    if (Foo(a))
+    {
+        DoSomething();
+    }
+
+    if (b)
+    {
+        DoSomething();
+    }
+    else
+    {
+        DoSomethingElse();
+    }
+
+    var d = "xxx";
+    var res = d;
+}
+</pre>
 <h2>Exceptions</h2>
 <p>This rule will not raise an issue in either of these cases:</p>
 <ul>
--- a/sonaranalyzer-dotnet/rspec/cs/S2583_c#.html
+++ b/sonaranalyzer-dotnet/rspec/cs/S2583_c#.html
@@ -2,26 +2,26 @@
 be used in production.</p>
 <h2>Noncompliant Code Example</h2>
 <pre>
-a = false;
-if (a) // Noncompliant
+public void Sample(bool b)
 {
-  DoSomething(); // never executed
-}
+    bool a = false;
+    if (a) // Noncompliant
+    {
+        DoSomething(); // never executed
+    }
 
-if (!a || b) // Noncompliant; "!a" is always "true", "b" is never evaluated
-{
-  DoSomething();
-}
-else
-{
-  DoSomethingElse(); // never executed
-}
+    if (!a || b) // Noncompliant; "!a" is always "true", "b" is never evaluated
+    {
+        DoSomething();
+    }
+    else
+    {
+        DoSomethingElse(); // never executed
+    }
 
-c = null;
-d = "value";
-c ?? "value"; // Noncompliant, c is always null
-d ?? "value"; // Noncompliant, d is always not null, "value" is never used
-x ?? c; // Noncompliant, c is always null
+    var d = "xxx";
+    var res = d ?? "value"; // Noncompliant, d is always not null, "value" is never used
+}
 </pre>
 <h2>Exceptions</h2>
 <p>This rule will not raise an issue in either of these cases:</p>
--- a/sonaranalyzer-dotnet/rspec/cs/S2583_c#.html
+++ b/sonaranalyzer-dotnet/rspec/cs/S2583_c#.html
@@ -16,6 +16,12 @@ else
 {
   DoSomethingElse(); // never executed
 }
+
+c = null;
+d = "value";
+c ?? "value"; // Noncompliant, c is always null
+d ?? "value"; // Noncompliant, d is always not null, "value" is never used
+x ?? c; // Noncompliant, c is always null
 </pre>
 <h2>Exceptions</h2>
 <p>This rule will not raise an issue in either of these cases:</p>
--- a/sonaranalyzer-dotnet/rspec/cs/S2589_c#.html
+++ b/sonaranalyzer-dotnet/rspec/cs/S2589_c#.html
@@ -20,27 +20,33 @@ public void Sample(bool b, bool c, string s)
         DoSomething();
     }
 
-    string d = null, res;
-    res = s ?? d; // Noncompliant, d is always null
-    res = d ?? "value"; // Noncompliant, d is always null
+    string d = null;
+    var v1 = d ?? "value"; // Noncompliant, d is always null
+    var v2 = s ?? d; // Noncompliant, d is always null and produced value is always equal to s. Condition to check value of s is gratuitous.
 }
 </pre>
 <h2>Compliant Solution</h2>
 <pre>
-a = true;
-if (Foo(a))
+public void Sample(bool b, bool c, string s)
 {
-  DoSomething();
-}
+    var a = true;
+    if (Foo(a))
+    {
+        DoSomething();
+    }
 
-if (b)
-{
-  DoSomething();
-}
+    if (b)
+    {
+        DoSomething();
+    }
 
-if (c)
-{
-  DoSomething();
+    if (c)
+    {
+        DoSomething();
+    }
+
+    var v1 = "value";
+    var v2 = s;
 }
 </pre>
 <h2>See</h2>
--- a/sonaranalyzer-dotnet/rspec/cs/S2589_c#.html
+++ b/sonaranalyzer-dotnet/rspec/cs/S2589_c#.html
@@ -2,20 +2,27 @@
 because it does not match the programmer's intent, then it's a bug and the expression should be fixed.</p>
 <h2>Noncompliant Code Example</h2>
 <pre>
-a = true;
-if (a) // Noncompliant
+public void Sample(bool b, bool c, string s)
 {
-  DoSomething();
-}
+    var a = true;
+    if (a) // Noncompliant
+    {
+        DoSomething();
+    }
 
-if (b &amp;&amp; a) // Noncompliant; "a" is always "true"
-{
-  DoSomething();
-}
+    if (b &amp;&amp; a) // Noncompliant; "a" is always "true"
+    {
+        DoSomething();
+    }
 
-if (c || !a) // Noncompliant; "!a" is always "false"
-{
-  DoSomething();
+    if (c || !a) // Noncompliant; "!a" is always "false"
+    {
+        DoSomething();
+    }
+
+    string d = null, res;
+    res = s ?? d; // Noncompliant, d is always null
+    res = d ?? "value"; // Noncompliant, d is always null
 }
 </pre>
 <h2>Compliant Solution</h2>
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
@@ -92,18 +92,20 @@ namespace SonarAnalyzer.Rules.CSharp
             SyntaxKind.FalseLiteralExpression
         };
 
+        private const string MessageFormat = "{0}";
+
         private const string S2583DiagnosticId = "S2583"; // Bug
         private const string S2583MessageFormatBool = "Change this condition so that it does not always evaluate to '{0}'; some subsequent code is never executed.";
-        private const string S2583MessageFormatNull = "Change this expression which always evaluates to '{0}'.";
+        private const string S2583MessageNotNull = "Change this expression which always evaluates to 'not null'; some subsequent code is never executed.";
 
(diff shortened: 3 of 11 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
```diff
@@ -330,5 +350,49 @@ bool IsBooleanLiteral(SyntaxNode syntaxNode)
...
+            }
+            var rightOperands = args.Instruction.DescendantNodesAndSelf()
+                .Select(x => (x is BinaryExpressionSyntax binary && binary.IsKind(SyntaxKind.CoalesceExpression) ? binary.Right : null)
+                          ?? (x is AssignmentExpressionSyntax assign && assign.IsKind(SyntaxKindEx.CoalesceAssignmentExpression) ? assign.Right : null))
+                .WhereNotNull().ToArray();
+            foreach (var right in rightOperands)
+            {
+                //It's too late for left operands of ??= or x=x??, symbolic value was already set
+                ProcessOperand(right, false);
+            }
```
christophe-zurn-sonarsource: Why do we need to look at the descendants here ? Can't we just check the current instruction ? since we will evaluate all instructions, it seems many syntax nodes will be reevaluated many times currently

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp6.cs
```diff
@@ -2562,6 +2562,61 @@ public string Method8(string message)
...
+            ret = "Lorem " + (isNull ?? a) + " ipsum"; // Noncompliant
+
+            //Right operand: isNull is known to be null, therefore ?? is useless
+            ret = a ?? null;    // Noncompliant
+            ret = a ?? isNull;  // Noncompliant {{Change this expression which always evaluates to 'null'.}}
+//                     ^^^^^^
+
+            //Combo/Fatality
+            ret = notNull ?? isNull;    //Noncompliant [LeftA, RightA]
+            ret = isNull ?? null;       //Noncompliant [LeftB, RightB]
```
christophe-zurn-sonarsource: I think these cases are out of scope of the rule. The rule is only about conditions. Although the right hand side is indeed useless, it doesn't really correspond what the rule is supposed to check

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
```diff
@@ -85,18 +93,17 @@ public sealed class ConditionEvaluatesToConstant : SonarDiagnosticAnalyzer
         };
 
         private const string S2583DiagnosticId = "S2583"; // Bug
-        private const string S2583MessageFormat = "Change this condition so that it does not always evaluate to '{0}'; some subsequent code is never executed.";
+        private const string S2583MessageFormatBool = "Change this condition so that it does not always evaluate to '{0}'; some subsequent code is never executed.";
+        private const string S2583MessageFormatNull = "Change this expression which always evaluates to '{0}'.";
```
christophe-zurn-sonarsource: I'd keep the `; some subsequent code is never executed.` part of the message, as the rule S2583 is really about unreachable code

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
```diff
@@ -132,6 +145,16 @@ private static void CheckForRedundantConditions(CSharpExplodedGraph explodedGrap
                         .Except(conditionTrue)
                         .Where(c => !IsInsideCatchOrFinallyBlock(c))
                         .Select(node => GetDiagnostics(node, false)))
+                    .Union(isNull
+                        .Except(isUnknown)
+                        .Except(isNotNull)
+                        .Where(c => !IsInsideCatchOrFinallyBlock(c))
+                        .Select(node => Diagnostic.Create(s2583Null, node.GetLocation(), messageArgs: "null")))
```
christophe-zurn-sonarsource: For this case, I'd raise the `S2589` version of the check, as the right hand side of the code will always be executed, so the condition is useless. This falls in the code smell rule case

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
```diff
@@ -85,18 +93,17 @@ public sealed class ConditionEvaluatesToConstant : SonarDiagnosticAnalyzer
...
         private const string S2589DiagnosticId = "S2589"; // Code smell
         private const string S2589MessageFormat = "Change this condition so that it does not always evaluate to '{0}'.";
 
-        private static readonly DiagnosticDescriptor s2583 =
-            DiagnosticDescriptorBuilder.GetDescriptor(S2583DiagnosticId, S2583MessageFormat, RspecStrings.ResourceManager);
-
-        private static readonly DiagnosticDescriptor s2589 =
-            DiagnosticDescriptorBuilder.GetDescriptor(S2589DiagnosticId, S2589MessageFormat, RspecStrings.ResourceManager);
+        private static readonly DiagnosticDescriptor s2583Bool = DiagnosticDescriptorBuilder.GetDescriptor(S2583DiagnosticId, S2583MessageFormatBool, RspecStrings.ResourceManager);
+        private static readonly DiagnosticDescriptor s2583Null = DiagnosticDescriptorBuilder.GetDescriptor(S2583DiagnosticId, S2583MessageFormatNull, RspecStrings.ResourceManager);
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: We need to fix this one as well `ConsumeValueTaskCorrectly.cs` - that's what I've found as example.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
pavel-mikula-sonarsource: That doesn't work for `ret = a ?? null`, because instruction itself is AssignmentExpression here and there's no separate instruction for "a ?? null" in the CFG. I can check AssignmentExpression but I'm afraid that `??` can be subexpression of any kind of expression here.
DescendantNodesAndSelf will produce a lot of reevaluations and I'm hoping for small chunks here.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp6.cs
pavel-mikula-sonarsource: These two cases are only for documentation purposes to see that it will be raised twice on the same line. Both left and right side are useless in this case.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
pavel-mikula-sonarsource: The rule title is `Conditionally executed blocks should be reachable`. Shouldn't we update it to `Conditionally executed code should be reachable`? In order to cover expressions with ?? and ??= as well as no-block single liners:
```
if(false)
  DoSomething();
```
Blocks seems to restrictive, not broad enough. But the rename would affect other languages as well - can we do it?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp6.cs
christophe-zurn-sonarsource: Yes. My point was that the fact that the right hand side is useless is not related to this rule, the lines (2593, 2594)
```
ret = a ?? null;
ret = a ?? isNull;
```
should not raise an issue in my POV, as here, the fact that the right hand side is always null is not related to a condition, which seems to me out of scope of the rule.
Same for right hand side of:
```
ret = notNull ?? isNull;
ret = isNull ?? null;
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
christophe-zurn-sonarsource: Makes sense to me

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp6.cs
pavel-mikula-sonarsource: So should I drop the check? Or do we have a better *EvaluatesToConstant rule to use?
I agree that this check does not fulfill the root purpose of the rule. It's collateral additional value only.

sonaranalyzer-dotnet/rspec/cs/S2589_c#.html
```diff
@@ -2,20 +2,27 @@
...
-{
-  DoSomething();
+    if (c || !a) // Noncompliant; "!a" is always "false"
+    {
+        DoSomething();
+    }
+
+    string d = null, res;
+    res = s ?? d; // Noncompliant, d is always null
+    res = d ?? "value"; // Noncompliant, d is always null
```
pavel-mikula-sonarsource: Do not merge yet. I need to update Compliant version.

sonaranalyzer-dotnet/rspec/cs/S2583_c#.html
```diff
@@ -2,19 +2,25 @@
...
+    {
+        DoSomething();
+    }
+    else
+    {
+        DoSomethingElse(); // never executed
+    }
+
+    var d = "xxx";
+    var res = d ?? "value"; // Noncompliant, d is always not null, "value" is never used
```
pavel-mikula-sonarsource: Do not merge yet. Compliant version was missing in RSPEC.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp6.cs
pavel-mikula-sonarsource: CC @andrei-epure-sonarsource 
Conclusion after discussion with @nicolas-harraudeau-sonarsource is to keep this feature. 
Nicolas pointed out a new POV on this case:
```
ret = a ?? null;
```
the condition `a==null` is **the** useless condition here. The final product will always be the same (value of `a`) no matter value comes in. Therefore it is a "condition" issue and it fits the rule title and description.

I will update RSPEC noncompliant example to make it clear.

sonaranalyzer-dotnet/rspec/cs/S2589_c#.html
pavel-mikula-sonarsource: Done

sonaranalyzer-dotnet/rspec/cs/S2583_c#.html
pavel-mikula-sonarsource: Done

pavel-mikula-sonarsource:
Regarding the `?? isNull` situation, I'was trying to make message and marked location clear and consistent. But I left it in the original way:
```
var d = null;
var v2 = s ?? d; {{Change this expression which always evaluates to "null".}}
              ^
```
It would be nice to mark the _condition_, but in every case it's not clear that 'd' is the root cause of this issue.

> Change this condition so that it does not always evaluate to 'FIXME: s? left operand? How to simply explain d?

```
var v2 = s ?? d;
         ^
var v2 = s ?? d;
         ^^^^
var v2 = s ?? d;
         ^^^^^^
```

sonaranalyzer-dotnet/rspec/cs/S2589_c#.html
```diff
@@ -2,38 +2,51 @@
...
-{
-  DoSomething();
+    if (c || !a) // Noncompliant; "!a" is always "false"
+    {
+        DoSomething();
+    }
+
+    string d = null;
+    var v1 = d ?? "value"; // Noncompliant, d is always null
+    var v2 = s ?? d; // Noncompliant, d is always null and produced value is always equal to s. Condition to check value of s is gratuitous.
```
andrei-epure-sonarsource: should use the definite article "the produced value"

andrei-epure-sonarsource: "The condition to check the value of s is gratuitous." or "Checking the value of s is gratuitous."

pavel-mikula-sonarsource: RSPEC is updated. Files will be updated as part of release process.

andrei-epure-sonarsource: that was quick :)

PR #3080, opened 2020-01-29
Use Environment.OSVersion to verify the OS type

### LINKED ISSUE #3048: S3949 throws AD0001 in VS 2015/MSBuild with .NET Framework v4.6.1
### Description

The CBDE rule S3949 is throwing when run inside VS2015 or building using MSBuild if only .NET Framework 4.6.1 is installed:

#### Error in SLVS2015
image

#### Error from MSBuild (without using the Scanner for MSBuild)
image

### Investigation
The rule is not expected to be run in VS or a "normal" MSBuild build (i.e. should not run unless running under the _Scanner for MSBuild_).

However, some of the code in the rule has to run to do the "should I run?" detection. The line that is causing the problem seems to be the call to `RuntimeInformation.IsOSPlatform` (it's defined in the assembly that can't be found).

```
        protected sealed override void Initialize(SonarAnalysisContext context)
        {
            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
            {
                new CbdeHandler(context, OnCbdeIssue, ShouldRunCbdeInContext, () => { return WorkDirectoryBasePath; },
                    testCbdeBinaryPath, onCbdeExecution);
            }
        }
```

It looks like that API first shipped in .NET Framework v4.7.1 ([see apisof.net](https://apisof.net/catalog/System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(OSPlatform)))

### Repro steps
The issue was discovered when testing a pre-release version of SLVS that embedded SonarC# v8.2, but it can easily be reproduced just be referencing the analyzer NuGet without SLVS being installed.

1. Create a C# project in VS2015
2. Add a reference to the SonarC# v8.2 NuGet package

### Expected behavior

CBDE rule does not produce an error or return any issues.

### Actual behavior

CBDE rule throws AD0001

```
Warning	AD0001	Analyzer 'SonarAnalyzer.Rules.CSharp.CbdeHandlerRule' threw an exception of type 'System.IO.FileNotFoundException' with message 'Could not load file or assembly 'System.Runtime.InteropServices.RuntimeInformation, Version=4.0.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.'.
```

### Known workarounds

Manually disable the rule in a ruleset.

Or do nothing. There are no adverse side-effects on VS or the other analyzer rules, other than creating a poor impression for the user.

### Related information

* SonarC# Version: v8.2
* Visual Studio Version: VS2015.3

costin-zaharia-sonarsource:
- RuntimeInformation.IsOSPlatform is only supported starting with
.NET Framework 4.7.1 and failed on Visual Studio 2015

- Environment.OSVersion is supported starting with .Net Framework 1.1
and .Net Core 2.0

Fixes #3048

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CbdeHandlerRule.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CbdeHandlerRule.cs
@@ -28,7 +28,6 @@ using SonarAnalyzer.Common;
 using System.Collections.Immutable;
 using System.Linq;
 using System.IO;
-using System.Runtime.InteropServices;
 using System.Diagnostics.CodeAnalysis;
 
 [assembly: InternalsVisibleTo("SonarAnalyzer.UnitTest" + Signing.InternalsVisibleToPublicKey)]
@@ -68,11 +67,12 @@ namespace SonarAnalyzer.Rules.CSharp
             return new CbdeHandlerRule(true, testCbdeBinaryPath, onCbdeExecution);
         }
 
-        protected sealed override void Initialize(SonarAnalysisContext context)
+        protected override void Initialize(SonarAnalysisContext context)
         {
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            // The available platform ids are documented here: https://docs.microsoft.com/en-us/dotnet/api/system.platformid?view=netframework-4.8
+            if (Environment.OSVersion.Platform == PlatformID.Win32NT)
             {
-                new CbdeHandler(context, OnCbdeIssue, ShouldRunCbdeInContext, () => { return WorkDirectoryBasePath; },
+                new CbdeHandler(context, OnCbdeIssue, ShouldRunCbdeInContext, () => WorkDirectoryBasePath,
                     testCbdeBinaryPath, onCbdeExecution);
             }
         }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
@@ -12,6 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="1.3.2" />
+    <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="3.1.1" />
     <PackageReference Include="System.Collections.Immutable" Version="1.1.37" />
     <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
   </ItemGroup>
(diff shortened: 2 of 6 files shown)


costin-zaharia-sonarsource:
I've tested this fix with the following:
- Windows 7: .Net Framework 4.6.1
- Windows 10: .Net Framework 4.8, .Net Core 2.1 and.Net Core 3.1
- macOS Catalina: .Net Core 2.1 and.Net Core 3.1
- Ubuntu: .Net Core 2.1 and.Net Core 3.1

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
```diff
@@ -12,6 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="1.3.2" />
+    <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="3.1.1" />
     <PackageReference Include="System.Collections.Immutable" Version="1.1.37" />
     <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
```
duncanp-sonar: Is the `System.Runtime.InteropServices.RuntimeInformation` package still required? I assume not (it was the one that was added previously that cause the bug).

costin-zaharia-sonarsource: Indeed it is no longer needed. I created a new PR for this: #3085.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CbdeHandlerRule.cs
```diff
@@ -68,11 +67,12 @@ internal static CbdeHandlerRule MakeUnitTestInstance(string testCbdeBinaryPath,
             return new CbdeHandlerRule(true, testCbdeBinaryPath, onCbdeExecution);
         }
 
-        protected sealed override void Initialize(SonarAnalysisContext context)
+        protected override void Initialize(SonarAnalysisContext context)
         {
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            // The available platform ids are documented here: https://docs.microsoft.com/en-us/dotnet/api/system.platformid?view=netframework-4.8
+            if (Environment.OSVersion.Platform == PlatformID.Win32NT)
```
duncanp-sonar: In the notes for #3048, the CFamily team said they tried to use `Environment.OSVersion.Platform` but it "didn't work on .NET Core". Perhaps they tried a different configuration where it failed? Or were they just using a version of .NET Core older than v2.0?

costin-zaharia-sonarsource: I'm not sure why it didn't work for them. I've asked @loic-joly-sonarsource for more details since this seems to be supported starting with .Net Core 2: https://docs.microsoft.com/en-us/dotnet/api/system.environment.osversion?view=netframework-4.8

PR #3081, opened 2020-01-29
Update documentation files

### LINKED ISSUE #3008: Update rspec2resx for .NET 4.8 SDK and *.MD files
### Update rspec2resx

- [x] Add support for .NET 4.8 SDK path to the script.
- [x] Update error message according to VS installer package name

### Update contributing-analyzer.md

- [x] Remove "Version 16.4.1" from VS 2019 name, this particular version is not necessary and link is universal
- [x] Add .NET 4.8 targeting pack
- [x] Rename all mentioned items according to VS installer UI (i.e. replace `.NET Core SDK 2.1.402`)
- [x] Make an items a sublist
- [x] Sort all items according to VS installer UI
- [x] Add note to `ORCHESTRATOR_URL` that it really has to be an URL (i.e. `file:///c:/something/orchestrator.properties`)

### Update rspec/README.md

- [x] Add note about root directory that script must be started from
- [x] Update all examples to copy/paste-able forms (add path prefix)

### Update docs/issues.md

- [x] Add close issue procedure opened by users if the issue is older than 3 months and we could not properly label it (if we asked more details and there was no answer)
- [x] Add the process that we have to consider new rules, or rule improvements, forum

### CODE_OF_CONDUCT.md  

- [x] Replace dotnet@sonarsource.com with other PoC. Community forum?

pavel-mikula-sonarsource:
Fixes #3008

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -53,7 +53,7 @@ further defined and clarified by project maintainers.
 ## Enforcement
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
-reported by contacting the project team at [dotnet@sonarsource.com]. All
+reported by contacting the project team at [Community Forum](https://community.sonarsource.com/). All
 complaints will be reviewed and investigated and will result in a response that
 is deemed necessary and appropriate to the circumstances. The project team is
 obligated to maintain confidentiality with regard to the reporter of an incident.
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ vulnerabilities and smells in your code.
 
 ## Have question or feedback?
 
-* [Contact us on Community Forum](https://community.sonarsource.com/) to provide feedback, request new rule or feature.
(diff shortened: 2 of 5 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -53,7 +53,7 @@ further defined and clarified by project maintainers.
 ## Enforcement
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
-reported by contacting the project team at [dotnet@sonarsource.com]. All
+reported by contacting the project team at [Community Forum](https://community.sonarsource.com/). All
```
andrei-epure-sonarsource: hmmm indeed I don't know about this one...
could we just align with Sonar Java regarding this? https://github.com/SonarSource/sonar-java/blob/master/CONTRIBUTING.md

do we need this?

pavel-mikula-sonarsource: Do we need to replace dotnet@? I think yes. This PR is baby step to address this.

Do we need Code of Conduct itself? I don't know. It sounds like a legal question to me => out of scope of this issue.
I personally don't see a problem if we drop the whole file. But there might be a reason why it was added.

PR #3085, opened 2020-01-29
Uninstall System.Runtime.InteropServices.RuntimeInformation

costin-zaharia-sonarsource:
- not needed


costin-zaharia-sonarsource:
> Actually, it looks like the `Microsoft.DotNet.PlatformAbstractions` is not used either. Is that correct?

Indeed. In the end I've used `Environment.OSVersion.Platform` since it's in the framework and seems to be supported by .Net Core 2: https://docs.microsoft.com/en-us/dotnet/api/system.environment.osversion?view=netframework-4.8

andrei-epure-sonarsource:
indeed, `Environment.OSVersion.Platform` was not supported in .NET Core 1.1, but that reached it's end-of-support on 2019-06-27 (see https://dotnet.microsoft.com/download/dotnet-core)

Issue #3086, opened 2020-01-30
S3241 UnusedReturnValue throws AD0001 System.ArgumentException


andrei-epure-sonarsource:
### Description

Dnspy https://cirrus-ci.com/task/5443779553394688 

```
   CSC : error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.UnusedReturnValue' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'. [C:\Project\Extensions\dnSpy.Debugger\dnSpy.Debugger.DotNet.Mono\dnSpy.Debugger.DotNet.Mono.csproj]
```

Nhibernate https://cirrus-ci.com/task/4986382716239872

```
   CSC : error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.UnusedReturnValue' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'. [C:\Project\src\NHibernate\NHibernate.csproj]
    CSC : error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.UnusedReturnValue' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'. [C:\Project\src\NHibernate\NHibernate.csproj]
    CSC : error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.UnusedReturnValue' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'. [C:\Project\src\NHibernate\NHibernate.csproj]
        13325 Warning(s)
```

PR #3087, opened 2020-01-30
Add reproducer for S3241 'AD0001' error

christophe-zurn-sonarsource:
related to #3086


andrei-epure-sonarsource:
(waiting for master to be green)

Issue #3055, opened 2020-01-17
tests fail on system with non-english culture


lg2de:
### Description

Getting lastest master version the unit tests does not complete successfully on my PC.
They fail due to different culture settings (German).

### Repro steps

1. Switch (Windows) language to different language, e.g. German.
2. Compile the solution and run the tests.

### Expected behavior

All tests are run fine on every (Windows) system.

### Actual behavior

Some tests fail. Here is one example:

```
Test Name:	BuildError
Test FullName:	SonarAnalyzer.UnitTest.SonarAnalyzer.UnitTest.TestFramework.DiagnosticVerifierTest.BuildError
Test Source:	[...]\sonaranalyzer-dotnet\tests\SonarAnalyzer.UnitTest\TestFramework\DiagnosticVerifierTest.cs : line 178
Test Outcome:	Failed
Test Duration:	0:00:00

Test Name:	BuildError
Test Outcome:	Failed
Result StackTrace:	
bei FluentAssertions.Execution.LateBoundTestFramework.Throw(String message)
   bei FluentAssertions.Execution.TestFrameworkProvider.Throw(String message)
   bei FluentAssertions.Execution.CollectingAssertionStrategy.ThrowIfAny(IDictionary`2 context)
   bei FluentAssertions.Execution.AssertionScope.Dispose()
   bei FluentAssertions.Specialized.ExceptionAssertions`1.ExceptionMessageAssertion.Execute(IEnumerable`1 messages, String expectation, String because, Object[] becauseArgs)
   bei FluentAssertions.Specialized.ExceptionAssertions`1.WithMessage(String expectedWildcardPattern, String because, Object[] becauseArgs)
   bei SonarAnalyzer.UnitTest.TestFramework.DiagnosticVerifierTest.BuildError() in [...]\SonarAnalyzer.UnitTest\TestFramework\DiagnosticVerifierTest.cs:Zeile 188.
Result Message:	
Expected exception message to match the equivalent of 
"Unexpected build error [CS1514]: { expected on line 2", but 
"Unexpected build error [CS1514]: { erwartet. on line 2" does not.
Result StandardOutput:	
Actual diagnostics: 3
  Id: CS1514, Line: 2, [46, 46]
  Id: CS1513, Line: 2, [46, 46]
  Id: CS1022, Line: 5, [0, 1]
```

Currently the following tests fail:
[...]

PR #3099, opened 2020-02-03
Make tests pass on non-english culture

andrei-epure-sonarsource:
@lg2de opened a fix for #3055 on #3098 (Thanks!)

this PR is needed to run the CI tests


lg2de:
Wondering why this PR is needed to validate CI.
Do you have any explanation, @andrei-epure-sonarsource ?

andrei-epure-sonarsource:
historically we were only trigerring QA (which are more expensive checks) on PRs done by sonarsourcers. I'm not sure if it's still the case now that we've migrated to Azure Pipelines, but I saw your CI fail so I imagined it would be the same reason (although [the error](https://sonarsource.visualstudio.com/DotNetTeam%20Project/_build/results?buildId=14669&view=logs&jobId=5dae5997-de96-559e-c3ae-b3120e3536d3) is a bit cryptic - `No service connection found with identifier TaskInputs[TaskInputs['connectedServiceNameSelector']].`)

Issue #2783, opened 2019-11-12
Fix S2068 Regular Expression creation from parameters


christophe-zurn-sonarsource:
If the `CredentialWords` parameter is set to a invald regex expression, such as `PASS)WORD`, rule instanciation will fail and might cause issues for user analysis.
To avoid such an issue, we should make sure that the regex creation does not fail. For instance, by escaping its ivalid characters `this.splitCredentialWords.Select(x => Regex.Escape(x))`. A warning that the parameter is an invalid Regex could also be displayed to the user.
Furthermore, the current regular expression does not catch all cases. Both the following code examples should be noncompliant:
C#
```
foo = "Server=localhost; Database=Main; User=Admin; Password=TopSecret;";       //Noncompliant
foo = "Server=localhost; Database=Main; User=Admin; Password   =   TopSecret;"; //Noncompliant
```
VBNET
```
foo = "Server=localhost; Database=Main; User=Admin; Password=TopSecret;;       'Noncompliant
foo = "Server=localhost; Database=Main; User=Admin; Password   =   TopSecret;; 'Noncompliant
```

New version could be `(?{0})\s*[:=]` (to be validated)

pavel-mikula-sonarsource:
As mentioned in #2768, to avoid FP in 

> "Server=localhost; Database=Main; User=Admin; Password= " + loadPassword()

 the regex should look like `(?{0})\s*[:=]\s*\S`

pavel-mikula-sonarsource:
This was fixed in #3114

PR #3101, opened 2020-02-04
Rule S2068: ensure the custom parameters are set

### LINKED ISSUE #2913: (Regression) Rule 2068: customized parameters not taken into account
On SonarQube 7.9
Test case: https://github.com/eric-therond-sonarsource/testcsharp/blob/master/Program.cs

With sonar-csharp-plugin-8.2.0.13504.jar:
1) A new default quality profile is setted with this parameterized rule:
sonarqube2
2) But the rule is not raised on lines 17 & 18:
sonarqube1

Expected behaviour like for sonar-csharp-plugin-7.15.0.8572.jar:
sonarqube3

costin-zaharia-sonarsource:
Fixes #2913


andrei-epure-sonarsource:
@costin-zaharia-sonarsource could you also add a regression test for it, either in this PR (ideally) or in a separate one? see https://github.com/SonarSource/sonar-dotnet/blob/cc164f4b085cd025f6931157d99a8e0f8ed61562/its/src/test/java/com/sonar/it/csharp/AutoGeneratedTest.java#L84 as a model - although it's changing a project property rather than a rule parameter via the API - otherwise you could put a SonarLint.xml file and link it to the solution

You could also take a look at #2783 as it's in the same context

costin-zaharia-sonarsource:
> https://github.com/SonarSource/sonar-dotnet/blob/cc164f4b085cd025f6931157d99a8e0f8ed61562/its/src/test/java/com/sonar/it/csharp/AutoGeneratedTest.java#L84
> 
> as a model - although it's changing a project property rather than a rule parameter via the API - otherwise you could put a SonarLint.xml file and link it to the solution

As far as I can tell the API does not support rule parameter customization. I will try the solution with linking the SonarLint.xml.

its/projects/DefaultRuleParametersCanBeCustomized/DefaultRuleParametersCanBeCustomized/DefaultRuleParametersCanBeCustomized.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Credentials
+{
+    public class Credentials
+    {
+        private const string Username = "admin";
+        private const string Password = "Password123";
+        private const string Usario = "admin";
```
andrei-epure-sonarsource: the test is a bit misleading - I expect 'Usario' is 'username' and an issue is raised on it because 'usario' is in given as a name for credential, together with "password, pwd, senha" (senha=password in portuguese, apparently).

The problem is that an issue is raised on line 7 and not on line 5 and this leads to some cognitive overload when reading the test - why is the rule raising for portuguese username but not for english username?

So I would keep just 'senha' and keep the test easy to follow (avoid raising on 'username' and 'usario')

Issue #3110, opened 2020-02-11
S2583 FP: Pattern matching generates wrong bool constraint


H4pH:
### Description

False positive with pattern matching and bool value

### Repro steps

With this code, rule throw "Change this condition so that it does not always evaluate to 'true'; some subsequent code is never executed.". Value is `object`, and valueBool can be `true` or `false`, so conditional expression also can be `true` or `false`
```csharp
if (value == null || !(value is bool valueBool))
	return;

if (valueBool)
{
	// some code
}
else
{
	// some code
}
```

### Expected behavior

Don't throw

### Related information

* SonarC# Version 8.1
* Visual Studio Version 2019 
* MSBuild version 16
* If running through the Scanner for MSBuild, its version 4.7.1.2311-netcoreapp2.0

pavel-mikula-sonarsource:
Hi @H4pH ,
Thank you for reporting this issue. We've confirmed it as FP on the latest SonarC# 8.3 as well.

PR #3111, opened 2020-02-11
Add repro for S2583 FP #3110

pavel-mikula-sonarsource:
Repro for #3110


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp7.cs
```diff
@@ -243,4 +243,24 @@ public void Write()
...
+
+    // See https://github.com/SonarSource/sonar-dotnet/issues/3110
+    public class Repro_3110
+    {
+        public int DoSomething(object value)
+        {
+            if (!(value is bool valueBool))
+                return -1;
+
+            if (valueBool)  // Noncompliant, FN - valueBool can be true or false after this line
```
andrei-epure-sonarsource:
```suggestion
            if (valueBool)  // Noncompliant, FP - valueBool can be true or false
```

Issue #2933, opened 2019-12-26
Add support for TupleExpressions in CSharpExplodedGraph


costin-zaharia-sonarsource:
### Description

We need to add support for handling TupleExpressions in CSharpExplodedGraph.

### Repro steps

Currently the TupleExpressions are just ignored when processing code like:
```
public class Address
{
    public string Name { get; }

    public string State { get; }

    public void Deconstruct(out string name, out string state) =>
        (name, state) = (Name, State);
}
```

### Additional notes

Once the tuples are supported by CFG and semantic execution we need to update the DeadStores rule (S1854) and remove the check added by #3113.

andrei-epure-sonarsource:
closing as this will be fixed with the new Roslyn CFG

pavel-mikula-sonarsource:
@andrei-epure-sonarsource I'd prefer to keep this documented as opened issue until we support it.

andrei-epure-sonarsource:
@pavel-mikula-sonarsource can we close this now?

PR #3113, opened 2020-02-12
Avoid false positives while handling tuples

### LINKED ISSUE #3094: S1854 False Positive with usage inside tuples
In the following code, SonarAnalyzer.CSharp 8.3.0.14607 flags `S1854: Remove this useless assignment to local variable 'r'.` Is this a false positive?

Version 8.2 didn't flag this code.

```
namespace S1854_False_Positive
{
    public static class Class
    {
        public static (int r, int g, int b) M(string text)
        {
            int r = int.Parse(text.Substring(0, 3));
            int g = int.Parse(text.Substring(4, 3));
            int b = int.Parse(text.Substring(8, 3));
            return (r, g, b);
        }
    }
}
```

costin-zaharia-sonarsource:
Since the tuples are not yet fully supported by both CFG and symbolic
execution (#2933), **S1854** will ignore the methods which are containing them.

Fixes #3094

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
@@ -869,9 +869,13 @@ namespace Tests.Diagnostics
     {
         public static (int foo, int bar) M(string text)
         {
-            int b = int.Parse(text); // Noncompliant -- false positive
+            int b = int.Parse(text);
             return (1, b);
         }
-    }
 
+        public void UnusedTuple_FalseNegative()
+        {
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
```diff
@@ -869,9 +869,13 @@ public class TupleReturn
...
         {
-            int b = int.Parse(text); // Noncompliant -- false positive
+            int b = int.Parse(text);
             return (1, b);
         }
-    }
 
+        public void UnusedTuple_FalseNegative()
+        {
+            (int x, int y) t = (1, 2); // Compliant - FN, tuples are not yet covered: https://github.com/SonarSource/sonar-dotnet/issues/2933
```
andrei-epure-sonarsource: probably it's worth mentioning in the ticket that once that ticket for CSharpExplodedGraph will be handled, this workaround should be removed

PR #3114, opened 2020-02-12
Rule S2068: add exceptions for query parameters

### LINKED ISSUE #2870: Rule S2068: filter database query parameters

The following use cases show legitimate use of parameters in a database query.

```csharp
string query1 = "password=?"; // Compliant
string query2 = "password=:password"; // Compliant
string query3 = "password=:param"; // Compliant
string query4 = "password='"+pwd+"'"; // Compliant
string query5 = "password={0}"; // Compliant
```

costin-zaharia-sonarsource:
Fixes #2870

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
@@ -56,7 +56,7 @@ namespace SonarAnalyzer.Rules
                     .Select(Regex.Escape)
                     .ToList();
 
-                this.passwordValuePattern = new Regex(string.Format(@"\b(?<password>{0})\s*[:=]\s*(?<suffix>.*)$",
+                this.passwordValuePattern = new Regex(string.Format(@"\b(?<password>{0})\s*[:=]\s*(?<suffix>.+)$",
                     string.Join("|", this.splitCredentialWords)), RegexOptions.Compiled | RegexOptions.IgnoreCase);
             }
         }
@@ -148,18 +148,10 @@ namespace SonarAnalyzer.Rules
                     .Intersect(analyzer.splitCredentialWords)
                     .ToHashSet(StringComparer.OrdinalIgnoreCase);
 
-                var matches = analyzer.passwordValuePattern.Matches(variableValue);
-                foreach (Match match in matches)
+                var match = analyzer.passwordValuePattern.Match(variableValue);
+                if (match.Success && !IsValidParameter(match.Groups["suffix"].Value))
                 {
-                    if (!match.Success)
-                    {
-                        continue;
-                    }
-
-                    if (!IsValidParameter(match.Groups["suffix"].Value))
-                    {
-                        credentialWordsFound.Add(match.Groups["password"].Value);
-                    }
+                    credentialWordsFound.Add(match.Groups["password"].Value);
                 }
 
                 // Rule was initially implemented with everything lower (which is wrong) so we have to force lower
@@ -170,10 +162,12 @@ namespace SonarAnalyzer.Rules
             private bool IsValidParameter(string suffix)
             {
                 var candidateParameter = suffix
-                    .Split(new[] {';'}, StringSplitOptions.RemoveEmptyEntries)
-                    .FirstOrDefault();
+                    .Split(';')
+                    .First()
+                    .Trim();
 
-                return candidateParameter == null || validParameterPattern.IsMatch(suffix);
+                return string.IsNullOrWhiteSpace(candidateParameter) ||
+                       validParameterPattern.IsMatch(candidateParameter);
             }
         }
     }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
@@ -55,7 +55,7 @@ namespace SonarAnalyzer.Rules
                     .Select(x => x.Trim())
                     .ToList();
 
-                this.passwordValuePattern = new Regex(string.Format(@"\b(?<password>{0})\s*[:=]\s*(?<suffix>.+)$",
+                this.passwordValuePattern = new Regex(string.Format(@"\b(?<credential>{0})\s*[:=]\s*(?<suffix>.+)$",
                     string.Join("|", this.splitCredentialWords.Select(Regex.Escape))), RegexOptions.Compiled | RegexOptions.IgnoreCase);
             }
         }
@@ -100,7 +100,7 @@ namespace SonarAnalyzer.Rules
         protected abstract class CredentialWordsFinderBase<TSyntaxNode>
              where TSyntaxNode : SyntaxNode
         {
-            private readonly Regex validParameterPattern = new Regex(@"^\?|:\w+|\{\d+[^}]*\}$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+            private readonly Regex validCredentialPattern = new Regex(@"^\?|:\w+|\{\d+[^}]*\}$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
             private readonly DoNotHardcodeCredentialsBase<TSyntaxKind> analyzer;
 
             protected CredentialWordsFinderBase(DoNotHardcodeCredentialsBase<TSyntaxKind> analyzer)
@@ -148,9 +148,9 @@ namespace SonarAnalyzer.Rules
                     .ToHashSet(StringComparer.OrdinalIgnoreCase);
 
                 var match = analyzer.passwordValuePattern.Match(variableValue);
-                if (match.Success && !IsValidParameter(match.Groups["suffix"].Value))
+                if (match.Success && !IsValidCredential(match.Groups["suffix"].Value))
                 {
-                    credentialWordsFound.Add(match.Groups["password"].Value);
+                    credentialWordsFound.Add(match.Groups["credential"].Value);
                 }
 
                 // Rule was initially implemented with everything lower (which is wrong) so we have to force lower
@@ -158,12 +158,12 @@ namespace SonarAnalyzer.Rules
                 return credentialWordsFound.Select(x => x.ToLowerInvariant());
             }
 
-            private bool IsValidParameter(string suffix)
+            private bool IsValidCredential(string suffix)
             {
-                var candidateParameter = suffix.Split(';').First().Trim();
+                var candidateCredential = suffix.Split(';').First().Trim();
 
-                return string.IsNullOrWhiteSpace(candidateParameter) ||
-                       validParameterPattern.IsMatch(candidateParameter);
+                return string.IsNullOrWhiteSpace(candidateCredential) ||
+                       this.validCredentialPattern.IsMatch(candidateCredential);
             }
         }
     }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
@@ -54,7 +54,7 @@ namespace SonarAnalyzer.Rules
                     .Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
                     .Select(x => x.Trim())
                     .ToList();
-                this.passwordValuePattern = new Regex(string.Format(@"\b(?<password>{0})\b[:=]\S",
+                this.passwordValuePattern = new Regex(string.Format(@"\b(?<password>{0})\b[:=](?!\?|:|\{{[0-9]*\}})\S",
                     string.Join("|", this.splitCredentialWords)), RegexOptions.Compiled | RegexOptions.IgnoreCase);
             }
         }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
@@ -53,11 +53,10 @@ namespace SonarAnalyzer.Rules
                 this.splitCredentialWords = value.ToUpperInvariant()
                     .Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
                     .Select(x => x.Trim())
-                    .Select(Regex.Escape)
                     .ToList();
 
                 this.passwordValuePattern = new Regex(string.Format(@"\b(?<password>{0})\s*[:=]\s*(?<suffix>.+)$",
-                    string.Join("|", this.splitCredentialWords)), RegexOptions.Compiled | RegexOptions.IgnoreCase);
+                    string.Join("|", this.splitCredentialWords.Select(Regex.Escape))), RegexOptions.Compiled | RegexOptions.IgnoreCase);
             }
         }
 
@@ -161,10 +160,7 @@ namespace SonarAnalyzer.Rules
 
             private bool IsValidParameter(string suffix)
             {
-                var candidateParameter = suffix
-                    .Split(';')
-                    .First()
-                    .Trim();
+                var candidateParameter = suffix.Split(';').First().Trim();
 
                 return string.IsNullOrWhiteSpace(candidateParameter) ||
                        validParameterPattern.IsMatch(candidateParameter);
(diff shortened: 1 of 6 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
```diff
@@ -54,7 +54,7 @@ public string CredentialWords
                     .Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
                     .Select(x => x.Trim())
                     .ToList();
-                this.passwordValuePattern = new Regex(string.Format(@"\b(?<password>{0})\b[:=]\S",
+                this.passwordValuePattern = new Regex(string.Format(@"\b(?<password>{0})\b[:=](?!\?|:|\{{[0-9]*\}})\S",
```
pavel-mikula-sonarsource: The new regex will be more difficult to build. Currently there are some FN:

- Passwords cannot START with `?`, but text like `?whaaat?` is SQL hardcoded password.
- Passwords cannot START with `:`, but text like  `:something:` is SQL hardcoded password. I'm not sure if `:password` and `:param` are the only two compliant examples, or if it's `:\w+`. @pierre-loup-tristant-sonarsource ?
- The `{0}` format string does not consider explicit format `{0:yyMMdd}`. I would suggest `\{\d+[^}]*\}`

pavel-mikula-sonarsource: I don't understand this part:
```
\{{[0-9]*\}}
```
To break it into pieces, I see:
```
\{         Escaped character { - OK

{          Start of Quantifier? The quantifier format should be like {1}, {1,2} or {1,}. Maybe it's some escape sequence I don't know, can you provide documentation for that?
  [0-9]*   Zero or more numbers. I prefer \d for readability. I think "zero" is mistake, it should be at least one, right? So \d+ instead.
  \}       Escaped character } - why is it inside the { } Quantifier section? First character was outside.
}          End of Quantifier? Same question as Start of Quantifier.
```

pavel-mikula-sonarsource:
Negative zero-width lookahead assertions are difficult to understand and make them right. It might be better to get the password (the rest of the string) out as explicit capture and inspect it's value in the code.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
costin-zaharia-sonarsource: In this case (`password={0}`), `{0}` can be any number, or, as you already mentioned, an explicit format.

costin-zaharia-sonarsource: Just checked with @pierre-loup-tristant-sonarsource and `:param` can actually be any word.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
```diff
@@ -148,13 +151,30 @@ private IEnumerable<string> FindCredentialWords(string variableName, string vari
                 var matches = analyzer.passwordValuePattern.Matches(variableValue);
                 foreach (Match match in matches)
                 {
-                    credentialWordsFound.Add(match.Groups["password"].Value);
+                    if (!match.Success)
```
pavel-mikula-sonarsource: `if(!Success) continue` followed by single if statement in the loop seems difficult to understand from cyclomatic complexity POV. I suggest to change it to
`if (match.Success && !IsValidParameter(match.Groups["suffix"].Value))`

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
```diff
@@ -148,13 +151,30 @@ private IEnumerable<string> FindCredentialWords(string variableName, string vari
                 var matches = analyzer.passwordValuePattern.Matches(variableValue);
                 foreach (Match match in matches)
```
pavel-mikula-sonarsource: Can we add test to prove this `foreach` usefull? That would require multiple passwords in single string situation and that does not make sense and is not handled by current regex.
I guess we can't due to current regex.
So I suggest to use `.Match` and use values directly. Check for `match.Success` should be enough.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
```diff
@@ -148,13 +151,30 @@ private IEnumerable<string> FindCredentialWords(string variableName, string vari
...
                 // Rule was initially implemented with everything lower (which is wrong) so we have to force lower
                 // before reporting to avoid new issues to appear on SQ/SC.
                 return credentialWordsFound.Select(x => x.ToLowerInvariant());
             }
+
+            private bool IsValidParameter(string suffix)
+            {
+                var candidateParameter = suffix
+                    .Split(new[] {';'}, StringSplitOptions.RemoveEmptyEntries)
+                    .FirstOrDefault();
```
pavel-mikula-sonarsource: This connection string `password=;user=xxx` will produce `;user=xxx` suffix. Using `RemoveEmptyEntries` will remove the first empty split result (right?) and cause FP.
I suggest to add more connection strings to the tests.
I suggest to use `.Split(';').First().Trim()` and remove the check for null from next command.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
```diff
@@ -53,8 +53,10 @@ public string CredentialWords
                 this.splitCredentialWords = value.ToUpperInvariant()
                     .Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
                     .Select(x => x.Trim())
+                    .Select(Regex.Escape)
```
pavel-mikula-sonarsource: Can we add configuration test to prove that we'll not fail with bad config value?

pavel-mikula-sonarsource: `FindCredentialWords` is using `splitCredentialWords` without regex to generate `credentialWordsFound`. We should move the `.Select(Regex.Escape)` to the next statement with RegEx creation for consistent behavior.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs
```diff
@@ -63,6 +63,16 @@ public void StandardAPI(SecureString secureString, string nonHardcodedPassword,
             new PasswordDeriveBytes("hardcoded", byteArray, "strHashName", 1); // Noncompliant
             new PasswordDeriveBytes("hardcoded", byteArray, "strHashName", 1, cspParams); // Noncompliant
         }
+
+        public void CompliantParameterUse(string pwd)
```
pavel-mikula-sonarsource: We should add similar test to VB test file as well.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs
```diff
@@ -63,6 +63,16 @@ public void StandardAPI(SecureString secureString, string nonHardcodedPassword,
...
+        {
+            string query1 = "password=?";
+            string query2 = "password=:password";
+            string query3 = "password=:param";
+            string query4 = "password='"+pwd+"'";
+            string query5 = "password={0}";
+            string query6 = @"Server=myServerName\myInstanceName;Database=myDataBase;Password=:myPassword;User Id=:username;";
+        }
```
pavel-mikula-sonarsource: I suggest to add (crazy) test like this to document FN and explain the semicolon handling and :param being valid
`"password=':crazy;secret';user=xxx"`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs
```diff
@@ -63,6 +63,16 @@ public void StandardAPI(SecureString secureString, string nonHardcodedPassword,
             new PasswordDeriveBytes("hardcoded", byteArray, "strHashName", 1); // Noncompliant
             new PasswordDeriveBytes("hardcoded", byteArray, "strHashName", 1, cspParams); // Noncompliant
         }
+
+        public void CompliantParameterUse(string pwd)
+        {
+            string query1 = "password=?";
+            string query2 = "password=:password";
+            string query3 = "password=:param";
+            string query4 = "password='"+pwd+"'";
```
pavel-mikula-sonarsource: The way this is handled produces FN in similar cases. We should add test to document the FN and explain why.
`"password=hardcoded;user='" + login + "'"`

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
```diff
@@ -148,13 +151,30 @@ private IEnumerable<string> FindCredentialWords(string variableName, string vari
...
                 return credentialWordsFound.Select(x => x.ToLowerInvariant());
             }
+
+            private bool IsValidParameter(string suffix)
+            {
+                var candidateParameter = suffix
+                    .Split(new[] {';'}, StringSplitOptions.RemoveEmptyEntries)
+                    .FirstOrDefault();
+
+                return candidateParameter == null || validParameterPattern.IsMatch(suffix);
```
pavel-mikula-sonarsource: `candidateParameter` is checked for null, but `suffix` is checked against regex. This looks like typo.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
costin-zaharia-sonarsource: This is quite a corner case since the escaped characters cannot be part of the identifier name. I would let this as it is since it's a little bit easier to read.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs
costin-zaharia-sonarsource: No really related to this commit but indeed since only `LiteralExpressionSyntax` nodes are covered, we should document this FN with an unit test.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
pavel-mikula-sonarsource: It is a corner case, but I don't like it for general principle. The variable `splitCredentialWords` should contain words, otherwise it has bad name and that's dangerous. Moving .Select() to statement below is not difficult to read and it's clear sanitizer for the regex.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DoNotHardcodeCredentialsTest.cs
```diff
@@ -43,7 +43,7 @@ public void DoNotHardcodeCredentials_CS_DefaultValues()
         public void DoNotHardcodeCredentials_CS_CustomValues()
         {
             Verifier.VerifyAnalyzer(@"TestCases\DoNotHardcodeCredentials_CustomValues.cs",
-                new CSharp.DoNotHardcodeCredentials(AnalyzerConfiguration.AlwaysEnabled) { CredentialWords = "kode,facal-faire" });
+                new CSharp.DoNotHardcodeCredentials(AnalyzerConfiguration.AlwaysEnabled) { CredentialWords = @"kode,facal-faire,\*+?|{}[]()^$.# " });
```
pavel-mikula-sonarsource: There should be a one test with `\*+?|{}[]()^$.#` value in `DoNotHardcodeCredentials_CustomValues.cs`. In best case, demonstrating {{error message}}.

pavel-mikula-sonarsource: The regex is not valid, but {}, [] and () are quite close to valid syntax. I suggest to swap them to be absolutely invalid RegExes on first sight.
```suggestion
                new CSharp.DoNotHardcodeCredentials(AnalyzerConfiguration.AlwaysEnabled) { CredentialWords = @"kode,facal-faire,\*+?|}{][)(^$.# " });
```

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
```diff
@@ -155,6 +158,17 @@ private IEnumerable<string> FindCredentialWords(string variableName, string vari
                 // before reporting to avoid new issues to appear on SQ/SC.
                 return credentialWordsFound.Select(x => x.ToLowerInvariant());
             }
+
+            private bool IsValidParameter(string suffix)
+            {
+                var candidateParameter = suffix
```
pavel-mikula-sonarsource: _nitpicking_: Commands are super-short in this case. I would put them on single line.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs
pavel-mikula-sonarsource: Clean as you code :)

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
costin-zaharia-sonarsource: I've moved it to the other line but the characters we are escaping are not words and will not remain words if we are not escaping them.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
```diff
@@ -155,6 +157,14 @@ private IEnumerable<string> FindCredentialWords(string variableName, string vari
                 // before reporting to avoid new issues to appear on SQ/SC.
                 return credentialWordsFound.Select(x => x.ToLowerInvariant());
             }
+
+            private bool IsValidParameter(string suffix)
+            {
+                var candidateParameter = suffix.Split(';').First().Trim();
```
pavel-mikula-sonarsource: Is `candidateParameter` a good name? I would suggest to use `password` to see what is known so far. It will be parameter after the regex check.

costin-zaharia-sonarsource: Indeed it was a bit strange. I've rename it to credential to match the "CredentialWords" property (since the user can customize in the configuration the CredentialWords)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_CustomValues.cs
```diff
@@ -34,4 +34,12 @@ public void StandardAPI(SecureString secureString, string nonHardcodedPassword,
             new PasswordDeriveBytes("hardcoded", byteArray, cspParams); // Noncompliant
         }
     }
+
+    public class FalseNegatives
+    {
+        public void Method()
+        {
+            string x = @"\*+?|}{][)(^$.# =something"; // FN
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: It's the `\b` at the start of the regex. Fixing this will trigger some false positives so I prefer to keep it like this for now. I think it's a good compromise since this is really a corner case.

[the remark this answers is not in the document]
pavel-mikula-sonarsource: I don't want to change the regex. It's just not clear why it is a FN. So adding some letter `x` at the beginning of the config value will work this around?
It would be nice to document this as FN with `*` config value like `string y = @"*=something"; // FN keyword must start with letter`

PR #3136, opened 2020-02-19
Fix S4136 FP: Allow grouped interface implementations

### LINKED ISSUE #3035: Fix S4136 FP: allow grouped interface implementations
### Description

When S4136 is active, it collides with the default cleanup of ReSharper, that preferes grouping interface implementations before method overloads.

Could this be included in S4136 as well, so grouping by interface implementation takes precedence over method name?

### Repro steps

Small code stump to replicate the behavior of the code that is reported as a violation to S4136:

```
interface IValidater<T> { bool Validate(T obj); bool CanValidate(T obj); }
class ValidateHelper : IValidater<string>, IValidater<int>
{
  bool Validate(string obj) => false;
  bool CanValidate(string obj) => false;
  
  bool Validate(int obj) => true;
  bool CanValidate(int obj) => true;
}
```

### Expected behavior

S4136 respects the above sorting of method names, by respecting that grouping of methods via the interfaces they implement take precedence over methods with the same name.

### Actual behavior

S4136 reports an error.

### Known workarounds

None that I know of, except disabling the rule.

### Related information

* SonarC# Version
   8.1 (build 13383)
   4.8.0

pavel-mikula-sonarsource:
Fixes #3035


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -274,4 +279,68 @@ protected override void DoSomething(bool b)
...
+        }
+        public void Aaa(string x) { }
+        public string this[string index] { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
+        public string Ppp { get; set; }
+        public void Zzz(string x) { }
+    }
+
+    public class ImplicitInterfaceWithEventHandler : ITest<int>
+    {
+        public void Zzz(string x) { } // Noncompliant
```
costin-zaharia-sonarsource: I think we might have a corner case here. If I insert `public void Aaa(string x) { }` right at the beginning of the class (before Zzz), it will be compliant but this is unexpected since it's not grouped with it's overload.

Another note related to this is that if we treat the methods like **Zzz** as noncompliant, the user will have to move them between the interface members in order to get rid of the warning.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -274,4 +279,68 @@ protected override void DoSomething(bool b)
...
+    }
+
+    public interface ICancel
+    {
+        void Cancel();
+        void Cancel(bool b);
+        void Renew();
+    }
+
+    public class InterfaceImplementationMethodsTogether : ICancel
```
costin-zaharia-sonarsource: I've noticed that the ctor is ignored if it separates the overloads:
```
    public class InterfaceImplementationMethodsTogether : ICancel
    {
        public void Cancel() { } // Noncompliant, it should be adjecent inside same interface implementation group

        public InterfaceImplementationMethodsTogether() { }

        public void Cancel(bool b) { } // Secondary

        public void Renew() { }
    }
```
Isn't this a false negative?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -48,10 +48,10 @@ protected MethodOverloadsShouldBeGroupedBase(System.Resources.ResourceManager re
 
         protected void CheckMembers(SyntaxNodeAnalysisContext c, IEnumerable<TMemberDeclarationSyntax> members)
         {
-            foreach (var misplacedNameSyntaxes in GetMisplacedOverloads(c, members))
+            foreach (var misplaced in GetMisplacedOverloads(c, members))
```
costin-zaharia-sonarsource: pedantic: it would be nice to rename `misplaced` with `misplacedOverload` since it make the code more clear

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -274,4 +279,68 @@ protected override void DoSomething(bool b)
 
     }
 
+    public class InterfaceImplementationTogether : IEqualityComparer<int>, IEqualityComparer<string>
```
costin-zaharia-sonarsource: I think it would be great to add a few additional tests with:
- a class which implements interfaces with at least a common method (e.g IFirst with M1() and M2() and ISecond with M2() and M3())
- a class which implements interfaces with default members

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodOverloadsShouldBeGrouped.cs
pavel-mikula-sonarsource: Both situation are expected and actually documented by this `ImplicitInterfaceWithEventHandler` case. This is FP corner case documented by this comment 
`// C# events are not recognized as parts of interface group, grouping is splitted here`

C# event `Eee` is a corner case, it's not recognized as part of interface group.  And this have some consequences:
- `Zzz(int x)` is not considered as grouped with interface (it's cut off by the `Eee`). So it's standalone and `Zzz(string x)` **is** considered as Noncompliant because of this.
- Inserting `Aaa(string x)` at the beginning is Compliant. Because `Aaa(int x)` is grouped with another interface member `Ppp`. So `Aaa(int x)` **is** grouped by interface and `Aaa(string x)` can be wherever else. This is actually prove that we don't force users to move members between interface members :)

Do you think I should add more comments to document this?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodOverloadsShouldBeGrouped.cs
pavel-mikula-sonarsource: Looks fine to me :) Issues are reported.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodOverloadsShouldBeGrouped.cs
pavel-mikula-sonarsource: What should it prove?
Implicit implementation will create method `M1`, `M2` and `M3`. Nothing interesting to raise.
Explicit implementation will be ignored. Ignoring explicit interfaces is documented in the test on line 117.
C#8 default member implementation? That should not interact with this rule. It will either not be implemented => not in syntax => does nothing. Or be implemented as interface and that's already tested.

costin-zaharia-sonarsource: I was thinking at this case:
```
    public interface ISecond
    {
        void A();
        void Cancel();
        void B();
    }

    public class ImplementsBoth : ICancel, ISecond
    {
        public void Cancel(bool b) { }
        public void Renew() { }

        public void A() { }
        public void Cancel() { } // Noncompliant
        public void B() { }
    }
```

pavel-mikula-sonarsource: That's a good catch. It did not behave as expected.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodOverloadsShouldBeGrouped.cs
costin-zaharia-sonarsource: Now I understand why it behaves this way. I've missed the group split part. No need for additional comments.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -89,11 +90,70 @@ protected List<SyntaxToken>[] GetMisplacedOverloads(SyntaxNodeAnalysisContext c,
...
+            {
+                if(groupedByInterface.TryGetValue(member, out var interfaces))
+                {
+                    return interfaces.Length==1 && others.Any(other => FindInterfaces(c.SemanticModel, other.Member).Contains(interfaces.Single()));
+                }
+                return true; // Not member of an interface => process
+            }
+        }
+
+        private Dictionary<TMemberDeclarationSyntax, ImmutableArray<INamedTypeSymbol>> GroupedByInterface(SyntaxNodeAnalysisContext c, IEnumerable<TMemberDeclarationSyntax> members)
```
costin-zaharia-sonarsource: GetMembersGroupedByInterface? 😄

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -89,11 +90,70 @@ protected List<SyntaxToken>[] GetMisplacedOverloads(SyntaxNodeAnalysisContext c,
                 }
             }
             return misplacedOverloads.Values.Where(x => x.Count > 1).ToArray();
+
+            bool ProcessByInterface(TMemberDeclarationSyntax member, List<MemberInfo> others)
```
costin-zaharia-sonarsource: IsMisplaced?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -62,24 +62,25 @@ protected void CheckMembers(SyntaxNodeAnalysisContext c, IEnumerable<TMemberDecl
             }
         }
 
-        protected List<SyntaxToken>[] GetMisplacedOverloads(SyntaxNodeAnalysisContext c, IEnumerable<TMemberDeclarationSyntax> members)
+        protected List<MemberInfo>[] GetMisplacedOverloads(SyntaxNodeAnalysisContext c, IEnumerable<TMemberDeclarationSyntax> members)
         {
-            var misplacedOverloads = new Dictionary<MemberInfo, List<SyntaxToken>>();
+            var misplacedOverloads = new Dictionary<MemberInfo, List<MemberInfo>>();
+            var groupedByInterface = GroupedByInterface(c, members);
```
costin-zaharia-sonarsource: membersGroupedByInterface?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -89,11 +90,70 @@ protected List<SyntaxToken>[] GetMisplacedOverloads(SyntaxNodeAnalysisContext c,
...
+                        ret.Add(previous, previousInterfaces);
+                    }
+                }
+                previousInterfaces = currentInterfaces;
+                previous = member;
+            }
+            return ret;
+        }
+
+        private ImmutableArray<INamedTypeSymbol> FindInterfaces(SemanticModel semanticModel, TMemberDeclarationSyntax member)
```
costin-zaharia-sonarsource: It would be great to add some comments since it will take a while if in 3 months we will have to update this.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -89,11 +90,70 @@ protected List<SyntaxToken>[] GetMisplacedOverloads(SyntaxNodeAnalysisContext c,
...
+                    if (@interface.GetMembers().Any(x => symbol.ContainingType.FindImplementationForInterfaceMember(x) == symbol))
+                    {
+                        ret.Add(@interface);
+                    }
+                }
+            }
+            return ret.ToImmutableArray();
         }
 
+        private IEnumerable<ISymbol> ExplicitInterfaces(ISymbol symbol) =>
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: ~~Get~~`ExplicitInterfaceImplementations` to follow Roslyn convention :)

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
pavel-mikula-sonarsource: Almost 😄
~~Get~~MembersGroupedByInterface?

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
pavel-mikula-sonarsource: If IsMisplaced or not is complex question answered by the entire `foreach`, variable state and bunch of `if`. This answers question if it ShouldBeProcessed based on two different interface criteria (not only interface grouping).
`ShouldProcess` ?

costin-zaharia-sonarsource: IsMisplacedCandidate? I think that Process is too generic and it doesn't provide additional value

Issue #3137, opened 2020-02-19
Fix S1481 FP: "var using" declarations should not raise issue


costin-zaharia-sonarsource:
### Description

Rule S1481 raises a false positive when the variable is declared inside a using declaration.

### Repro steps

```
void Foo(string path)
{
    using var stream = File.Create(path);
}
```

### Expected behavior

There should be no issue raised.

### Actual behavior

The issue is raised for the variable **stream**

### Known workarounds

```
using(var t = new TestTimer()) // t never used, but compliant.
{
  //...
}
```

but this will raise another issue if the using body is empty.

### Related information

* SonarC# Version 8.3

costin-zaharia-sonarsource:
I'm closing this issue since it would be easier to use `File.Create(path).Dispose();` instead and this will not need a new variable.

albankimor:
Hello,

I encounter the same false positive issue with the following code :

`using var _ = _logger.BeginScope("XXX");`

I think this issue should be reopenned and the rule S1481 should be adapted for the C# 8 syntax.

Thank you for advance

pavel-mikula-sonarsource:
`using var _ = new ...` is not a discard pattern, but `_` variable that could be eventually used. So we should implement this as originally reported by @costin-zaharia-sonarsource

GeeWee:
While `using var _ = new ...` is not an actual discard pattern - I'd like to think the intent is very clear anyways - and that it should not be marked as an error.

PR #3138, opened 2020-02-19
Document false positive for S1481

costin-zaharia-sonarsource:
- related to #3137


pavel-mikula-sonarsource:
I don't see #3137 as FP. Variable `stream` is indeed unused in the code. It's just misused to dispose the object in obfuscated and confusing way.

Removing the variable `System.IO.File.Create(path);` should trigger S2930 instead. Here we might have a FN actually.

Proper way of handling this would be `System.IO.File.Create(path).Dispose();`

It actually could have functional impact, like using lock on the file for operations done in `Foo()`, but that would be a poor design as well.

costin-zaharia-sonarsource:
> Proper way of handling this would be `System.IO.File.Create(path).Dispose();`

Indeed, this would be easier to do that and it will not introduce an unused variable.

PR #3142, opened 2020-02-19
Fix S4004 FP: Should not raise on [Serializable] objects

### LINKED ISSUE #2762: Fix S4004 FP: Should not raise on [Serializable] objects
### Description
Classes designed for serialization with the `[Serializable]` attribute, require public setters for all there properties, also those who are advised not to be so by S4004. Therefor, S4004 should not apply on `[Serializable]` classes, and this should be explained on the help page.

### Repro steps
``` C#
[Serializable]
public class S4004Serializable
{
    public List<int> Values { get; set; } // Compliant, [Serializable] requires a public setter.
}
```
### Related information
* SonarC# Version 8.0.0

pavel-mikula-sonarsource:
Fixes #2762

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CollectionPropertiesShouldBeReadOnly.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CollectionPropertiesShouldBeReadOnly.cs
@@ -1,3 +1,4 @@
+﻿using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Runtime.Serialization;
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CollectionPropertiesShouldBeReadOnly.cs
```diff
@@ -69,6 +70,15 @@ public class Message
         public List<int> Values { get; set; }
     }
 
+    // Ignore collections marked with Serializable attribute: https://github.com/SonarSource/sonar-dotnet/issues/2762
```
costin-zaharia-sonarsource: I think it would be great to update also the [RSPEC](https://jira.sonarsource.com/browse/RSPEC-4004) and include the exception.

pavel-mikula-sonarsource: Good point! I'll do that in #3146

PR #3147, opened 2020-02-20
Fix S4275: FP when the value is encapsulated in another class

### LINKED ISSUE #2867: Fix S4275: FP when the value is encapsulated in another class
### Description

When the field represents a class which encapsulates the actual value returned from the property, S4275 is incorrectly raised.

### Repro steps

Consider this code:
```
public class Sample
{
    private class ValueWrapper<T>
    {
        public T Value { get; set; }
        // ...
    }

    private readonly ValueWrapper<double> _someMember = new ValueWrapper<double>();

    public double SomeMember
    {
        get => _someMember.Value;
        set => _someMember.Value = value;
    }
}
```

### Expected behavior

S4275 should not be raised because the setter _does_ access the corresponding field.

### Actual behavior

S4275 is raised on the setter.

### Related information

* SonarLint Version 4.14.0.12332
* Visual Studio Version 16.4.0

pavel-mikula-sonarsource:
Fixes #2867

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
@@ -54,7 +54,9 @@ namespace SonarAnalyzer.Rules.CSharp
                 FieldData? foundField = null;
                 if (node is AssignmentExpressionSyntax assignment && node.IsKind(SyntaxKind.SimpleAssignmentExpression))
                 {
-                    foundField = ExtractFieldFromExpression(AccessorKind.Setter, assignment.Left, compilation);
+                    foundField = assignment.Left.DescendantNodesAndSelf().OfType<ExpressionSyntax>()
+                        .Select(x => ExtractFieldFromExpression(AccessorKind.Setter, x, compilation))
+                        .FirstOrDefault(x => x != null);
                 }
                 else if (node is ArgumentSyntax argument && IsRefOrOut(argument.RefOrOutKeyword))
                 {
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -54,7 +54,9 @@ protected override IEnumerable<FieldData> FindFieldAssignments(IPropertySymbol p
                 FieldData? foundField = null;
                 if (node is AssignmentExpressionSyntax assignment && node.IsKind(SyntaxKind.SimpleAssignmentExpression))
                 {
-                    foundField = ExtractFieldFromExpression(AccessorKind.Setter, assignment.Left, compilation);
+                    foundField = assignment.Left.DescendantNodesAndSelf().OfType<ExpressionSyntax>()
+                        .Select(x => ExtractFieldFromExpression(AccessorKind.Setter, x, compilation))
```
costin-zaharia-sonarsource: pedantic: "x" does not provide any helpful info, I think that using "expression" would provide a bit of context

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -57,7 +57,9 @@ protected override IEnumerable<FieldData> FindFieldAssignments(IPropertySymbol p
                 FieldData? foundField = null;
                 if (node is AssignmentStatementSyntax assignment && assignment.IsKind(SyntaxKind.SimpleAssignmentStatement))
                 {
-                    foundField = ExtractFieldFromExpression(AccessorKind.Setter, assignment.Left, compilation);
+                    foundField = assignment.Left.DescendantNodesAndSelf().OfType<ExpressionSyntax>()
```
costin-zaharia-sonarsource: The code is similar with the one for C#. Can we extract a few generic methods in order to enable reuse?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
pavel-mikula-sonarsource: I don't want to change this. It's lambda, so single letter variable is fine. `x` is by convention the most used one (I've found 63 other files in project with `x =>`). 
It comes from `.OfType<ExpressionSyntax>()` so in this case, it's very clear and explicit what `x` is.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/PropertiesAccessCorrectField.cs
pavel-mikula-sonarsource: I didn't like the old version of the rule and considered full refactoring. But there are too many language dependent types => too much work for this sprint. So I went for keeping the original scheme of the rule.
Turning VB into low hanging fruit is a bigger topic.

PR #3151, opened 2020-02-20
Fix FPs regarding XXE rule inside using

### LINKED ISSUE #3150: Fix FPs regarding XXE rule inside using
for .NET FW 3.5 or 4 where the constructor is vulnerable

```
        public void WithUsingStatement(FileStream fs)
        {
            using (XmlTextReader r = new XmlTextReader(fs, XmlNodeType.Element, null)) // FP!!!
            {
                r.XmlResolver = null; // no DTD resolving
            }

        }
```

andrei-epure-sonarsource:
Fixes #3150

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpObjectInitializationTracker.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpObjectInitializationTracker.cs
@@ -212,8 +212,11 @@ namespace SonarAnalyzer.SyntaxTrackers
 
             var variableSymbol = GetAssignedVariableSymbol(objectCreation, semanticModel);
 
+            var nextStatements = GetNextStatements(statement);
+            var innerStatements = GetInnerStatements(statement);
+
             return variableSymbol != null
-                && GetNextStatements(statement)
+                && nextStatements.Union(innerStatements)
                     .OfType<ExpressionStatementSyntax>()
                     .Select(x => x.Expression)
                     .OfType<AssignmentExpressionSyntax>()
@@ -254,5 +257,8 @@ namespace SonarAnalyzer.SyntaxTrackers
 
         private static IEnumerable<StatementSyntax> GetNextStatements(StatementSyntax statement) =>
             statement.Parent.ChildNodes().OfType<StatementSyntax>().SkipWhile(x => x != statement).Skip(1);
+
+        private static IEnumerable<StatementSyntax> GetInnerStatements(StatementSyntax statement) =>
+            statement.DescendantNodes().OfType<StatementSyntax>();
     }
 }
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpObjectInitializationTracker.cs
```diff
@@ -212,8 +212,11 @@ private bool IsLaterAssignedWithAllowedValue(ObjectCreationExpressionSyntax obje
 
             var variableSymbol = GetAssignedVariableSymbol(objectCreation, semanticModel);
 
+            var nextStatements = GetNextStatements(statement);
```
andrei-epure-sonarsource: `IsLaterAssignedWithAllowedValue` is used to exclude FPs, so including more statements to check can only lead to less noise - therefore it should be safe

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpObjectInitializationTracker.cs
```diff
@@ -212,8 +212,11 @@ private bool IsLaterAssignedWithAllowedValue(ObjectCreationExpressionSyntax obje
...
+            var innerStatements = GetInnerStatements(statement);
+
             return variableSymbol != null
-                && GetNextStatements(statement)
+                && nextStatements.Union(innerStatements)
                     .OfType<ExpressionStatementSyntax>()
                     .Select(x => x.Expression)
                     .OfType<AssignmentExpressionSyntax>()
```
costin-zaharia-sonarsource: If I'm not missing something we can simplify the code a bit by replacing `GetNextStatements` and `GetInnerStatements` with:
```
private static IEnumerable<AssignmentExpressionSyntax> GetAssignmentExpressions(StatementSyntax statement) =>
    statement.Parent.DescendantNodes().OfType<AssignmentExpressionSyntax>();
```

then we could replace the return with: 
`return variableSymbol != null && GetAssignmentExpressions(statement).Any(TrackedPropertySetWithAllowedValue);`

andrei-epure-sonarsource: your approach changes the functionality (because it takes all the nodes, not only the ones in the same block and the "inner block" of the using/if in the samples)

for example, we currently have tests 
```
            var reader = new XmlTextReader(Url); // Noncompliant
            if (foo)
            {
                reader.ProhibitDtd = true; // this is set conditionally, so not enough
            }

```

and this will not raise anymore (the idea is that after the if, the `reader` object is still vulnerable)

also, there are failing tests 

```
            var c2 = new HttpCookie("c"); // Noncompliant, HttpOnly is set conditionally
            if (foo)
            {
                c2.HttpOnly = true;
            }

```

and I don't want to change the functionality of the security hotspots that are using this base class

PR #3153, opened 2020-02-21
Fix S4275 FP when method is used inside getter or setter

### LINKED ISSUE #2435: Fix S4275 FP when method is used inside getter or setter
### Description

When the field is returned or set using a method call, the rule raises an issue.

The problem is that we don't really support inter-procedural analysis, so this FP is non-trivial to fix.

### Repro steps

```
    public class CrossProcedural
    {
        private int myValue;
        public int MyValue
        {
            // below there are two False Positives
            get => this.GetMyValue(); // Noncompliant {{Refactor this getter so that it actually refers to the field 'myValue'.}}
            set => this.SetMyValue(value); // Noncompliant {{Refactor this setter so that it actually refers to the field 'myValue'.}}
        }

        private int GetMyValue() => this.myValue;
        private void SetMyValue(int value) => this.myValue = value;
    }
```

### Related information

* SonarC# Version 7.14

pavel-mikula-sonarsource:
Fixes #2435

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
@@ -19,7 +19,6 @@
  */
 
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -35,11 +34,7 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class PropertiesAccessCorrectField : PropertiesAccessCorrectFieldBase
     {
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
-
-        protected override DiagnosticDescriptor Rule => rule;
+        public PropertiesAccessCorrectField() : base(RspecStrings.ResourceManager) { }
 
         protected override IEnumerable<FieldData> FindFieldAssignments(IPropertySymbol property, Compilation compilation)
         {
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
@@ -45,41 +45,17 @@ namespace SonarAnalyzer.Rules.CSharp
 
             // we only keep information for the first location of the symbol
             var assignments = new Dictionary<IFieldSymbol, FieldData>();
-            FillAssignments(setter, true);
+            FillAssignments(assignments, compilation, setter, true);
 
             // If there're no candidate variables, we'll try to inspect one local method invocation with value as argument
             if (assignments.Count == 0
-                && (setter.ExpressionBody()?.Expression ?? SingleInvocation(setter.Body)) is ExpressionSyntax expression
+                && (setter.ExpressionBody()?.Expression ?? SingleInvocation(setter.Body)) is { } expression
                 && FindInvokedMethod(compilation, property.ContainingType, expression) is MethodDeclarationSyntax invokedMethod)
             {
-                FillAssignments(invokedMethod, false);
+                FillAssignments(assignments, compilation, invokedMethod, false);
             }
 
             return assignments.Values;
-
-            bool IsRefOrOut(SyntaxToken node) => node.IsKind(SyntaxKind.RefKeyword) || node.IsKind(SyntaxKind.OutKeyword);
-
-            void FillAssignments(SyntaxNode root, bool useFieldLocation)
-            {
-                foreach (var node in root.DescendantNodes())
-                {
-                    FieldData? foundField = null;
-                    if (node is AssignmentExpressionSyntax assignment && node.IsKind(SyntaxKind.SimpleAssignmentExpression))
-                    {
-                        foundField = assignment.Left.DescendantNodesAndSelf().OfType<ExpressionSyntax>()
-                            .Select(x => ExtractFieldFromExpression(AccessorKind.Setter, x, compilation, useFieldLocation))
-                            .FirstOrDefault(x => x != null);
-                    }
-                    else if (node is ArgumentSyntax argument && IsRefOrOut(argument.RefOrOutKeyword))
-                    {
-                        foundField = ExtractFieldFromExpression(AccessorKind.Setter, argument.Expression, compilation, useFieldLocation);
-                    }
-                    if (foundField.HasValue && !assignments.ContainsKey(foundField.Value.Field))
-                    {
-                        assignments.Add(foundField.Value.Field, foundField.Value);
-                    }
-                }
-            }
         }
 
         protected override IEnumerable<FieldData> FindFieldReads(IPropertySymbol property, Compilation compilation)
@@ -142,13 +118,35 @@ namespace SonarAnalyzer.Rules.CSharp
             (property.GetMethod?.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax() is AccessorDeclarationSyntax getter &&
             getter.DescendantNodes().Any());
 
-        private ExpressionSyntax SingleReturn(BlockSyntax body)
+        private static void FillAssignments(Dictionary<IFieldSymbol, FieldData> assignments, Compilation compilation, SyntaxNode root, bool useFieldLocation)
+        {
+            foreach (var node in root.DescendantNodes())
+            {
+                FieldData? foundField = null;
+                if (node is AssignmentExpressionSyntax assignment && node.IsKind(SyntaxKind.SimpleAssignmentExpression))
+                {
+                    foundField = assignment.Left.DescendantNodesAndSelf().OfType<ExpressionSyntax>()
+                        .Select(x => ExtractFieldFromExpression(AccessorKind.Setter, x, compilation, useFieldLocation))
+                        .FirstOrDefault(x => x != null);
+                }
+                else if (node is ArgumentSyntax argument && argument.RefOrOutKeyword.IsAnyKind(SyntaxKind.RefKeyword,SyntaxKind.OutKeyword))
+                {
+                    foundField = ExtractFieldFromExpression(AccessorKind.Setter, argument.Expression, compilation, useFieldLocation);
+                }
+                if (foundField.HasValue && !assignments.ContainsKey(foundField.Value.Field))
+                {
+                    assignments.Add(foundField.Value.Field, foundField.Value);
+                }
+            }
+        }
+
+        private static ExpressionSyntax SingleReturn(BlockSyntax body)
         {
             var returns = body.DescendantNodes().OfType<ReturnStatementSyntax>().ToArray();
             return returns.Length == 1 ? returns.Single().Expression : null;
         }
 
-        private ExpressionSyntax SingleInvocation(BlockSyntax body)
+        private static ExpressionSyntax SingleInvocation(BlockSyntax body)
         {
             var expressions = body.DescendantNodes().OfType<InvocationExpressionSyntax>().Select(x => x.Expression).ToArray();
             if (expressions.Length == 1)
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
@@ -27,6 +27,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.ShimLayer.CSharp;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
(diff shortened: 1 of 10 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -114,7 +139,28 @@ protected override bool ShouldIgnoreAccessor(IMethodSymbol accessorMethod)
...
             getter.Parent.DescendantNodes().Any());
 
-        private static FieldData? ExtractFieldFromRefArgument(ArgumentSyntax argument, Compilation compilation)
+        private ExpressionSyntax SingleReturn(StatementSyntax body)
+        {
+            var returns = body.Parent.DescendantNodes().OfType<ReturnStatementSyntax>().ToArray();
+            return returns.Length == 1 ? returns.Single().Expression : null;
+        }
+
+        private ExpressionSyntax SingleInvocation(StatementSyntax body)
```
costin-zaharia-sonarsource: can be static

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -114,7 +139,28 @@ protected override bool ShouldIgnoreAccessor(IMethodSymbol accessorMethod)
             (property.GetMethod?.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax() is AccessorStatementSyntax getter &&
             getter.Parent.DescendantNodes().Any());
 
-        private static FieldData? ExtractFieldFromRefArgument(ArgumentSyntax argument, Compilation compilation)
+        private ExpressionSyntax SingleReturn(StatementSyntax body)
```
costin-zaharia-sonarsource: can be static

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PropertiesAccessCorrectField.cs
```diff
@@ -554,4 +540,116 @@ public double SomeMember
...
+            {
+                SetByBody(value);
+            }
+        }
+
+        public int BodyValue_ // With "this."
+        {
+            get
+            {
+                return this.GetByBody();
```
costin-zaharia-sonarsource: I think we should add a test also with more than one instruction

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -50,27 +50,39 @@ protected override IEnumerable<FieldData> FindFieldAssignments(IPropertySymbol p
 
             // we only keep information for the first location of the symbol
             var assignments = new Dictionary<IFieldSymbol, FieldData>();
+            FillAssignments(setter, true);
+            // If there're no candidate variables, we'll try to inspect one local method invocation with value as argument
+            if (assignments.Count == 0
+                && SingleInvocation(setter) is ExpressionSyntax expression
```
costin-zaharia-sonarsource: SingleInvocation always returns ExpressionSyntax. Can use `{ }` if you want to check for null and assign a name to the return value.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/PropertiesAccessCorrectFieldBase.cs
```diff
@@ -112,7 +112,7 @@ protected void CheckExpectedFieldIsUsed(IMethodSymbol methodSymbol, IFieldSymbol
             {
                 Location location = null;
                 string accessorType = null;
-                if (fields.Count() == 1)
+                if (fields.Count(x => x.UseFieldLocation) == 1)
```
costin-zaharia-sonarsource: fields parameter is iterated here more than once. We should use a List instead of IEnumerable<>.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/PropertiesAccessCorrectFieldBase.cs
```diff
@@ -145,6 +145,16 @@ protected IList<PropertyData> CollectPropertyData(IEnumerable<IPropertySymbol> p
             return allPropertyData;
         }
 
+        protected SyntaxNode FindInvokedMethod(Compilation compilation, INamedTypeSymbol containingType, SyntaxNode expression) =>
```
costin-zaharia-sonarsource: can be static

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/PropertiesAccessCorrectFieldBase.cs
```diff
@@ -145,6 +145,16 @@ protected IList<PropertyData> CollectPropertyData(IEnumerable<IPropertySymbol> p
...
+            compilation.GetSemanticModel(expression.SyntaxTree) is SemanticModel semanticModel
+            && semanticModel.GetSymbolInfo(expression).Symbol is ISymbol invocationSymbol
+            && invocationSymbol.ContainingType == containingType
+            && invocationSymbol.DeclaringSyntaxReferences.Length == 1
+            && invocationSymbol.DeclaringSyntaxReferences.Single().GetSyntax() is SyntaxNode invokedMethod
```
costin-zaharia-sonarsource: We can use object pattern `is { }` instead of `is SemanticModel`, `is ISymbol` and `is SyntaxNode`. Specifying the same type as the method return type is misleading.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -49,27 +50,41 @@ protected override IEnumerable<FieldData> FindFieldAssignments(IPropertySymbol p
 
             // we only keep information for the first location of the symbol
             var assignments = new Dictionary<IFieldSymbol, FieldData>();
-            foreach (var node in setter.DescendantNodes())
+            FillAssignments(setter, true);
+
+            // If there're no candidate variables, we'll try to inspect one local method invocation with value as argument
+            if (assignments.Count == 0
+                && (setter.ExpressionBody()?.Expression ?? SingleInvocation(setter.Body)) is ExpressionSyntax expression
```
costin-zaharia-sonarsource: Can use object pattern (`is { }`) instead of `is ExpressionSyntax `

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -49,27 +50,41 @@ protected override IEnumerable<FieldData> FindFieldAssignments(IPropertySymbol p
...
-                    assignments.Add(foundField.Value.Field, foundField.Value);
-                }
+                FillAssignments(invokedMethod, false);
             }
+
             return assignments.Values;
 
             bool IsRefOrOut(SyntaxToken node) => node.IsKind(SyntaxKind.RefKeyword) || node.IsKind(SyntaxKind.OutKeyword);
+
+            void FillAssignments(SyntaxNode root, bool useFieldLocation)
```
costin-zaharia-sonarsource: I think it would be better to extract this outside of this method. The method is big enough already and like this it hides the dependencies (compilation, assignments, IsRefOrOut)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -119,9 +147,28 @@ protected override bool ShouldIgnoreAccessor(IMethodSymbol accessorMethod)
             (property.GetMethod?.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax() is AccessorDeclarationSyntax getter &&
             getter.DescendantNodes().Any());
 
-        private static FieldData? ExtractFieldFromExpression(AccessorKind accessorKind,
-            ExpressionSyntax expression,
-            Compilation compilation)
+        private ExpressionSyntax SingleReturn(BlockSyntax body)
```
costin-zaharia-sonarsource: can be static

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -119,9 +147,28 @@ protected override bool ShouldIgnoreAccessor(IMethodSymbol accessorMethod)
...
+        {
+            var returns = body.DescendantNodes().OfType<ReturnStatementSyntax>().ToArray();
+            return returns.Length == 1 ? returns.Single().Expression : null;
+        }
+
+        private ExpressionSyntax SingleInvocation(BlockSyntax body)
```
costin-zaharia-sonarsource: can be static

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -135,14 +182,14 @@ protected override bool ShouldIgnoreAccessor(IMethodSymbol accessorMethod)
             if (strippedExpression is IdentifierNameSyntax &&
                 semanticModel.GetSymbolInfo(strippedExpression).Symbol is IFieldSymbol field)
             {
-                return new FieldData(accessorKind, field, strippedExpression);
+                return new FieldData(accessorKind, field, strippedExpression, useFieldLocation);
             }
             // Check for "this.foo"
             else if (strippedExpression is MemberAccessExpressionSyntax member &&
```
costin-zaharia-sonarsource: The else is not really needed since the first if returns directly.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PropertiesAccessCorrectField.cs
```diff
@@ -554,4 +540,116 @@ public double SomeMember
...
+        private int tooNested;
+        private int tooComplex;
+
+        public int ExpressionValue
+        {
+            get => GetByExpression();
+            set => SetByExpression(value);
+        }
+
+        public int ExpressionValue_ // With "this."
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Yes, _ are ignored. See documentation of `PropertyToFieldMatcher` class in `PropertiesAccessCorrectFieldBase.cs`. I believe that test case `NonCompliant_UnderscoresInNamesAndCasing` on line 45 covers this.

Purpose of this _ here was to reduce scaffolding noise in order to test "this." usage (it requires different handling). I'll update the RSPEC.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
pavel-mikula-sonarsource: I prefer to keep it for readability. It is else-if logic.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -124,15 +170,13 @@ protected override bool ShouldIgnoreAccessor(IMethodSymbol accessorMethod)
...
             }
             return null;
         }
 
-        private static FieldData? ExtractFieldFromExpression(AccessorKind accessorKind,
-            ExpressionSyntax expression,
-            Compilation compilation)
+        private static FieldData? ExtractFieldFromExpression(AccessorKind accessorKind, ExpressionSyntax expression, Compilation compilation, bool useFieldLocation)
         {
             var semanticModel = compilation.GetSemanticModel(expression.SyntaxTree);
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: It's also used by `IsFieldOrWithEvents` local function in the first if branch.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -145,16 +189,16 @@ protected override bool ShouldIgnoreAccessor(IMethodSymbol accessorMethod)
             // Check for direct field access: "Foo"
             if (strippedExpression is IdentifierNameSyntax && IsFieldOrWithEvents(out var directSymbol))
             {
-                return new FieldData(accessorKind, directSymbol, strippedExpression);
+                return new FieldData(accessorKind, directSymbol, strippedExpression, useFieldLocation);
             }
             else
             {
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I prefer to keep it for readability. It is else-if logic.

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -145,16 +189,16 @@ protected override bool ShouldIgnoreAccessor(IMethodSymbol accessorMethod)
...
                 // Check for "Me.Foo"
                 if (strippedExpression is MemberAccessExpressionSyntax member &&
-                    member.Expression is MeExpressionSyntax thisExpression &&
-                    IsFieldOrWithEvents(out var meSymbol))
+                    member.Expression is MeExpressionSyntax &&
+                    semanticModel.GetSymbolInfo(strippedExpression).Symbol is IFieldSymbol field2)
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Probably refactoring artifact. Fixed.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/PropertiesAccessCorrectFieldBase.cs
```diff
@@ -145,34 +171,23 @@ protected IList<PropertyData> CollectPropertyData(IEnumerable<IPropertySymbol> p
...
-         * - directly via an assignment
-         * - indirectly, when passed as 'out' or 'ref' parameter
-         */
-        protected abstract IEnumerable<FieldData> FindFieldAssignments(IPropertySymbol property, Compilation compilation);
-        protected abstract IEnumerable<FieldData> FindFieldReads(IPropertySymbol property, Compilation compilation);
-        protected abstract bool ImplementsExplicitGetterOrSetter(IPropertySymbol property);
-        protected abstract bool ShouldIgnoreAccessor(IMethodSymbol accessorMethod);
-
-
         protected struct PropertyData
```
costin-zaharia-sonarsource: not really related to this commit but I noticed that we could make this: "private readonly struct" and it would be nice.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/PropertiesAccessCorrectFieldBase.cs
```diff
@@ -145,34 +171,23 @@ protected IList<PropertyData> CollectPropertyData(IEnumerable<IPropertySymbol> p
...
         {
             public PropertyData(IPropertySymbol propertySymbol, IEnumerable<FieldData> read, IEnumerable<FieldData> updated,
                 bool ignoreGetter, bool ignoreSetter)
             {
                 PropertySymbol = propertySymbol;
-                ReadFields = read;
-                UpdatedFields = updated;
+                ReadFields = read.ToImmutableArray();
```
costin-zaharia-sonarsource: This is nice but it would make sense in this case that both FindFieldAssignments and FindFieldReads to return ImmutableArray<T>. What do you think?

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/PropertiesAccessCorrectField.cs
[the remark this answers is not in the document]
costin-zaharia-sonarsource: true, that's one reason why I prefer to extract local functions when the methods get big; it make the code hard to read :)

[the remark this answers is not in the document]
pavel-mikula-sonarsource: It's not hard to read/understand. It's hard to review 😄

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/PropertiesAccessCorrectFieldBase.cs
pavel-mikula-sonarsource: I considered that and decided to go with this version to remove unneeded duplication. It's PropertyData that really needs ImmutableArray, so it make sense to me that this is the single point responsible of doing so.

chucklu:
@pavel-mikula-sonarsource  We are using Version 8.4.2 (build 36762) of sonarqube,  and the same problem still exists.

pavel-mikula-sonarsource:
Hi @chucklu, can you please create a new issue with a reproducer and CSharp analyzer plugin version used? This is just a PR that was merged and closed.

There should be analyzer for C# 8.9 embedded in SQ 8.4.2.

chucklu:
@pavel-mikula-sonarsource  Is it possible to reopen https://github.com/SonarSource/sonar-dotnet/issues/2435, since it's the same issue.

pavel-mikula-sonarsource:
@chucklu We don't reopen issues released to milestones. Please create a new one, because your issue doesn't seem to be relevant to SonarC# Version 7.14 as reported with #2435

PR #3168, opened 2020-02-28
Performance improvement for GenericTypeParameterUnused

costin-zaharia-sonarsource:
- use structs
- filter the identifiers by name before asking for SymbolInfo

This reduces GenericTypeParameterUnused rule execution time
(using AnalyzerTool) for Entity Framework Core project from
~1315636ms to ~4531ms.

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterUnused.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterUnused.cs
@@ -65,7 +65,7 @@ namespace SonarAnalyzer.Rules.CSharp
                         var declarations = declarationSymbol.DeclaringSyntaxReferences
                             .Select(reference => reference.GetSyntax());
 
-                        var typeParameterNames = helper.Parameters.Parameters.Select(typeParameter => typeParameter.Identifier.Text).ToList();
+                        var typeParameterNames = helper.Parameters.Parameters.Select(typeParameter => typeParameter.Identifier.Text).ToArray();
 
                         var usedTypeParameters = GetUsedTypeParameters(declarations, typeParameterNames, c, analysisContext.Compilation);
 
@@ -136,7 +136,7 @@ namespace SonarAnalyzer.Rules.CSharp
                 methodSymbol.IsChangeable();
         }
 
-        private static List<string> GetUsedTypeParameters(IEnumerable<SyntaxNode> declarations, List<string> typeParameterNames,
+        private static List<string> GetUsedTypeParameters(IEnumerable<SyntaxNode> declarations, string[] typeParameterNames,
             SyntaxNodeAnalysisContext localContext,
             Compilation compilation)
         {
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterUnused.cs
```diff
@@ -135,14 +136,15 @@ private static bool IsMethodCandidate(MethodDeclarationSyntax methodDeclaration,
...
-        private static List<string> GetUsedTypeParameters(IEnumerable<SyntaxNode> declarations,
+        private static List<string> GetUsedTypeParameters(IEnumerable<SyntaxNode> declarations, List<string> typeParameterNames,
             SyntaxNodeAnalysisContext localContext,
             Compilation compilation)
         {
             return declarations
                 .SelectMany(declaration => declaration.DescendantNodes())
                 .OfType<IdentifierNameSyntax>()
                 .Where(identifier => !(identifier.Parent is TypeParameterConstraintClauseSyntax))
+                .Where(identifier => typeParameterNames.Contains(identifier.Identifier.ValueText))
```
andrei-epure-sonarsource: given that `typeParameterNames` is used only for search, in a loop, would an `ISet` be better than `List`?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/GenericTypeParameterUnused.cs
```diff
@@ -57,18 +57,19 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                        if (helper.Parameters == null || helper.Parameters.Parameters.Count == 0)
                         {
                             return;
                         }
 
                         var declarations = declarationSymbol.DeclaringSyntaxReferences
                             .Select(reference => reference.GetSyntax());
 
-                        var usedTypeParameters = GetUsedTypeParameters(declarations, c, analysisContext.Compilation);
+                        var typeParameterNames = helper.Parameters.Parameters.Select(typeParameter => typeParameter.Identifier.Text).ToList();
```
andrei-epure-sonarsource: maybe make it a set (see below comment)?

Corniel: It depends on the epected size: https://stackoverflow.com/questions/150750/hashset-vs-list-performance
If a set is not the best way to go (collection to small), consider an array as it read-only (right?).

costin-zaharia-sonarsource: Looking at the graph (https://stackoverflow.com/a/10762995), in the vast majority of cases we should be fine with a list since usually the methods have just a few type parameters.

I've replaced the list with an array but I expect that the optimizations done at this level would be rather small.

andrei-epure-sonarsource: ok, basically my comment was not appropiate at small hash sizes given  that the list is actually used for type parameters

my bad :)

Corniel: array v.s. list would indeed make no (real) difference, but I think you should minimize the overhead if you just need an internal read-only collection. :)

costin-zaharia-sonarsource: @andrei-epure-sonarsource: I think the idea is very good and that it's important to keep this in mind and make conscious decisions when picking collections so I'm very glad you asked (it's so easy to forget and select the list by default)

@Corniel: indeed, and this will make a difference in the end if used consistently :)

PR #3172, opened 2020-03-03
Rule S2068: support url userinfo component

### LINKED ISSUE #2865: Rule S2068: support url userinfo component

The URL user info component can contain a hardcoded password:

```csharp
string url1 = "scheme://user:azerty123@domain.com"; // Sensitive
string url2 = "scheme://user:@domain.com"; // Compliant
string url3 = "scheme://user@domain.com:80"; // Compliant
string url4 = "scheme://user@domain.com"; // Compliant
string url5 = "scheme://domain.com/user:azerty123"; // Compliant
```

Exception: no issue should be raised if user and password part of the userinfo component are the same:

```
string url1 = "scheme://admin:admin@domain.com"; // Compliant
```

pavel-mikula-sonarsource:
Fixes #2865

Reported messages were update according to current RSPEC.

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
@@ -20,6 +20,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 using System.Text.RegularExpressions;
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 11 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
```diff
@@ -118,20 +124,23 @@ protected CredentialWordsFinderBase(DoNotHardcodeCredentialsBase<TSyntaxKind> an
...
                     var variableName = GetVariableName(declarator);
                     var variableValue = GetAssignedValue(declarator);
 
                     var bannedWords = FindCredentialWords(variableName, variableValue);
                     if (bannedWords.Any())
                     {
-                        context.ReportDiagnosticWhenActive(
-                            Diagnostic.Create(rule, declarator.GetLocation(), string.Join(", ", bannedWords)));
+                        context.ReportDiagnosticWhenActive(Diagnostic.Create(rule, declarator.GetLocation()
+                            , string.Format(MessageFormatCredential, bannedWords.JoinStr(", ", x => $@"""{x}"""))));
```
christophe-zurn-sonarsource: _minor_ coma is after new line

christophe-zurn-sonarsource: Could we extract the `"` character inside the message format to use `x => $"{x}"` here instead.
Or remove the `@` string from the message format since it is not currently needed

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
```diff
@@ -33,17 +34,21 @@ public abstract class DoNotHardcodeCredentialsBase<TSyntaxKind> : ParameterLoadi
...
+        private const string MessageFormat = "{0}";
+        private const string MessageFormatCredential = @"{0} detected here, make sure this is not a hard-coded credential.";
+        private const string MessageUriUserInfo = "Review this hard-coded URL, which may contain a credential.";
+        private const string DefaultCredentialWords = "password, passwd, pwd, passphrase";
 
+        protected readonly DiagnosticDescriptor rule;
         private string credentialWords;
         private IEnumerable<string> splitCredentialWords;
         private Regex passwordValuePattern;
+        private readonly Regex uriUserInfoPattern = new Regex(@"\w+:\/\/(?<Login>[^:]+):(?<Password>[^@]+)@", RegexOptions.Compiled);
```
christophe-zurn-sonarsource: I think we coukld move this regex insided the nested class, next to the existing `validCredentialPattern` regex, as it is only used there.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
pavel-mikula-sonarsource: I've moved it into message. It will produce `"a, b, c" detected here` instead of `"a", "b", "c" detected here`. It's not so nice, but it matches the RSPEC message.

Issue #3125, opened 2020-02-17
Rule S2755: XML parsers should not be vulnerable to XXE attacks - XmlReader and XPathNavigator


andrei-epure-sonarsource:
[MMF-1946](https://jira.sonarsource.com/browse/MMF-1946)

Improve #3096 by adding support for XmlReader and XPathNavigator

XML Parser | Safe by default?
-- | --
XmlReader | Yes
XPathNavigator |  
...prior to 4.5.2 | No
...in versions 4.5.2+ | Yes

PR #3176, opened 2020-03-04
Implement S2755 for XmlReader

costin-zaharia-sonarsource:
First part of: #3125


costin-zaharia-sonarsource:
The 5 code smells reported by sonarcloud are all in the test cases directory and I need to use the obsolete property in order to test the rule behavior on .Net 3.5

sonaranalyzer-dotnet/its/sources/ManuallyAddedNoncompliantIssues/NetFramework35/XmlReaderTest.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Xml;
+
+namespace NetFramework35
+{
+    public class XmlReaderTest
+    {
+        public void XmlReader_EnableProhibitDtdAndSetResolver()
+        {
+            var settings = new XmlReaderSettings {ProhibitDtd = false, XmlResolver = new XmlUrlResolver()};
```
andrei-epure-sonarsource: you should also use in ITs 
```
    XmlReaderSettings settings = new XmlReaderSettings();
    settings.ProhibitDtd = false;
    settings.XmlResolver = new XmlSafeResolver();
```

andrei-epure-sonarsource: (initialization after the object creation)
(safe resolver as well)

andrei-epure-sonarsource: I've seen now that it's covered in the UTs

you can still add in one of the ITs

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/XmlExternalEntityShouldNotBeParsed_XmlReader_Net452.cs
```diff
@@ -0,0 +1,114 @@
+﻿using System.Xml;
+
+namespace SonarAnalyzer.UnitTest.TestCases
+{
+    public class XmlReaderUnsafe
```
andrei-epure-sonarsource: in [Microsoft's guide](https://docs.microsoft.com/en-us/archive/msdn-magazine/2009/november/xml-denial-of-service-attacks-and-defenses#defending-against-external-entity-attacks) which is referenced in the [OWASP guide](https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#net), they tell how to enable safe XXE processing

search for "Figure 6 Applying Safer Entity Parsing Settings to XmlDocument"

```
void processXml(string xml)
{
    MemoryStream stream =
        new MemoryStream(Encoding.Default.GetBytes(xml));
    XmlReaderSettings settings = new XmlReaderSettings();

    // allow entity parsing but do so more safely
    settings.ProhibitDtd = false;
    settings.MaxCharactersFromEntities = 1024;
    settings.XmlResolver = new XmlSafeResolver();

    XmlReader reader = XmlReader.Create(stream, settings);
    XmlDocument doc = new XmlDocument();
    doc.Load(reader);
}
```
*if possible* it would be cool to avoid this known FP

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -0,0 +1,106 @@
...
+            var assignments = invocation.FirstAncestorOrSelf<MethodDeclarationSyntax>()
+                .DescendantNodes()
+                .OfType<AssignmentExpressionSyntax>()
+                .Where(assignment => assignment.Left is MemberAccessExpressionSyntax memberAccess &&
+                                     IsXmlReaderSettings(memberAccess.Expression, semanticModel) &&
+                                     settings.Equals(semanticModel.GetSymbolInfo(memberAccess.Expression).Symbol));
+
+            foreach (var assignment in objectCreationAssignments.Union(assignments))
+            {
+                if (assignment.Left.NameIs("ProhibitDtd") ||
```
andrei-epure-sonarsource: is it possible to move the assignment name checks before the `semanticModel.GetSymbolInfo` call?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -279,19 +279,6 @@ public static bool IsStringEmpty(this ExpressionSyntax expression, SemanticModel
         public static Location FindIdentifierLocation(this BaseMethodDeclarationSyntax methodDeclaration) =>
             GetIdentifierOrDefault(methodDeclaration)?.GetLocation();
 
-        public static bool IsCatchingAllExceptions(this CatchClauseSyntax catchClause)
```
andrei-epure-sonarsource: this wasn't used?

costin-zaharia-sonarsource: Indeed, dead code. 

Don't know yet what's the best approach when removing these. A separate PR or mix  with the PR in progress?

sonaranalyzer-dotnet/its/sources/ManuallyAddedNoncompliantIssues/NetFramework35/XmlReaderTest.cs
costin-zaharia-sonarsource: Indeed it is covered in the unit tests but I've added now the case in the UTs

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/XmlExternalEntityShouldNotBeParsed_XmlReader_Net452.cs
costin-zaharia-sonarsource: We don't have a false positive here since we check the exact type of the XmlResolver, similar with what has been done for the other XXE rule types.

The blacklist contains only XmlUrlResolver and XmlPreloadedResolver for now.

I've added a test case though to document the behavior.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
costin-zaharia-sonarsource: I've actually did this at first but the problem is that this check needs to be done only for assignments and not for the object creation so the code complexity increases a bit to handle the checks. 

In the end I've decided to do it like this since we do the call to GetSymbolInfo only for the assignments done on XmlReaderSettings so there is a very low chance to do unnecessary calls.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
andrei-epure-sonarsource: ideally it would be a separate commit, to keep the history clean.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
andrei-epure-sonarsource: so `semanticModel.GetTypeInfo` is not as expensive of `semanticModel.GetSymbolInfo` ?

andrei-epure-sonarsource: you could extract `assignment.Left.Name` (whatever the name is) in a local variable and just do string comparisons to avoid calling a method each time

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -0,0 +1,106 @@
...
+                    assignment.Left.NameIs("DtdProcessing"))
+                {
+                    unsafeDtdProcessing = IsXmlResolverDtdProcessingUnsafe(assignment, semanticModel);
+                }
+
+                if (assignment.Left.NameIs("XmlResolver"))
```
andrei-epure-sonarsource: this should be `else if`, right?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
costin-zaharia-sonarsource: Yes, as far as I can tell GetTypeInfo is at least twice as fast:

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18363
Intel Core i5-5575R CPU 2.80GHz (Broadwell), 1 CPU, 4 logical and 4 physical cores
  [Host]     : .NET Framework 4.8 (4.8.4121.0), X86 LegacyJIT
  DefaultJob : .NET Framework 4.8 (4.8.4121.0), X86 LegacyJIT


|                   Method |       Mean |    Error |   StdDev | Ratio | RatioSD |
|------------------------- |-----------:|---------:|---------:|------:|--------:|
|      GetTypeWithTypeInfo |   677.5 us | 13.30 us | 14.23 us |  1.00 |    0.00 |
| GetTypeWithGetSymbolInfo | 1,462.3 us | 29.00 us | 32.23 us |  2.16 |    0.06 |

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/XmlExternalEntityShouldNotBeParsed_XmlReader_Net452.cs
```diff
@@ -0,0 +1,137 @@
+﻿using System;
+using System.IO;
+using System.Text;
+using System.Xml;
+
+namespace SonarAnalyzer.UnitTest.TestCases
```
andrei-epure-sonarsource: in one of these files, it would be good to add some test cases similar to what I did in the `VariousUnsafeCombinations` (well, what would make sense for this rule)

**1**. document what's happening when the XmlReaderSettings is just a parameter of a factory method
---> what happens if nothing is set on the settings
---> what happens when something is set (so settings are initialized w/o constructor)

**2**. from [the docs](https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmlreader?view=netframework-4.8#creating-an-xml-reader), what happens when you use the XmlReader `Create` constructor called from one of the inheriting classes, such as
```
            XmlNodeReader.Create("", settings, "resources/");  // returns XmlReader
```
or 
```
            XmlTextReader.Create("", settings, "resources/"); // returns XmlReader
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/XmlExternalEntityShouldNotBeParsed_XmlReader_Net35.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System.Xml;
+
+namespace SonarAnalyzer.UnitTest.TestCases
+{
+    public class XmlExternalEntityShouldNotBeParsed_XmlReader_Net35
```
andrei-epure-sonarsource: please also add the case with no settings passed to the constructor

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/XmlExternalEntityShouldNotBeParsed_XmlReader_BeforeNet452.cs
```diff
@@ -0,0 +1,146 @@
+﻿using System.Xml;
```
andrei-epure-sonarsource: I wouldn't rename it BeforeNet452 because Net35 is before 4.5.2 as well

maybe use the same convention as for XmlDocument - Net4

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -356,22 +343,19 @@ public static SimpleNameSyntax GetIdentifier(this ExpressionSyntax expression)
...
 
+        public static string GetName(this ExpressionSyntax expression) =>
+            expression switch
+            {
+                MemberAccessExpressionSyntax memberAccess => memberAccess.Name.Identifier.ValueText,
+                IdentifierNameSyntax identifierName => identifierName.Identifier.ValueText,
+                _ => string.Empty
+            };
+
         public static bool NameIs(this MemberAccessExpressionSyntax memberAccess, string name) =>
```
andrei-epure-sonarsource: can you drop this and only use `NameIs(this ExpressionSyntax)` ?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -343,20 +343,19 @@ public static SimpleNameSyntax GetIdentifier(this ExpressionSyntax expression)
...
 
-            return invocationName.Equals(name, StringComparison.InvariantCulture);
-        }
+        public static bool NameIs(this MemberAccessExpressionSyntax memberAccess, string name) =>
+            memberAccess.Name.Identifier.ValueText.Equals(name, StringComparison.InvariantCulture);
+
+        public static bool NameIs(this ExpressionSyntax expression, string name) =>
+            expression.GetName().Equals(name, StringComparison.InvariantCulture);
 
         public static bool NameIs(this IdentifierNameSyntax identifierName, string name)
```
andrei-epure-sonarsource: can you drop this as well?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/InvocationExpressionSyntaxExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace SonarAnalyzer.Helpers
+{
+    internal static class InvocationExpressionSyntaxExtensions
+    {
+        internal static bool IsMemberAccessOnKnownType(this InvocationExpressionSyntax invocation, string identifierName, KnownType knownType, SemanticModel semanticModel) =>
```
andrei-epure-sonarsource: currently `InvocationExpressionSyntax` extensions are in 
https://github.com/SonarSource/sonar-dotnet/blob/30e88d1dd135b2da11af74acfbc16e12be33322e/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs#L63

and 
https://github.com/SonarSource/sonar-dotnet/blob/30e88d1dd135b2da11af74acfbc16e12be33322e/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpOverloadHelper.cs#L32

With this change you are adding another class. While it's going in the right direction - do you want to make a refactoring PR immediately after this one, to not leave things as they are?

andrei-epure-sonarsource: (I don't expect any action on this PR)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/InvocationExpressionSyntaxExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+            invocation.Expression is MemberAccessExpressionSyntax memberAccessExpressionSyntax &&
+            memberAccessExpressionSyntax.Name?.Identifier.ValueText == identifierName &&
```
andrei-epure-sonarsource: use `GetName()` extension?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        public XmlReaderSettingsValidator(SemanticModel semanticModel)
+        {
+            this.semanticModel = semanticModel;
+        }
+
+        public bool IsUnsafe(InvocationExpressionSyntax invocation, ISymbol settings, NetFrameworkVersion version)
+        {
+            var unsafeDtdProcessing = false;
+            var unsafeResolver = version == NetFrameworkVersion.Probably35 || version == NetFrameworkVersion.Between4And451;
```
andrei-epure-sonarsource: I would move `version == NetFrameworkVersion.Probably35 || version == NetFrameworkVersion.Between4And451;` in a private `IsUnsafeVersion()` method to improve the readability of this method

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -0,0 +1,106 @@
...
+    internal class XmlReaderSettingsValidator
+    {
+        private readonly SemanticModel semanticModel;
+
+        public XmlReaderSettingsValidator(SemanticModel semanticModel)
+        {
+            this.semanticModel = semanticModel;
+        }
+
+        public bool IsUnsafe(InvocationExpressionSyntax invocation, ISymbol settings, NetFrameworkVersion version)
```
andrei-epure-sonarsource: could you add a comment , saying that `settings` should be XmlReaderSettings?

and why is this passed, as well? why do you need a symbol settings when you are looking at some invocation - what's the link between them? what should this invocation be?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -0,0 +1,106 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.XXE
+{
+    internal class XmlReaderSettingsValidator
+    {
+        private readonly SemanticModel semanticModel;
+
+        public XmlReaderSettingsValidator(SemanticModel semanticModel)
```
andrei-epure-sonarsource: you can pass the `NetFrameworkVersion version` to the constructor and have a field `unsafeResolver`, it would clarify a bit the class

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Linq;
+using System.Xml;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.XXE
+{
+    internal class XmlReaderSettingsValidator
```
andrei-epure-sonarsource: it would really help to have a high-level overview of this class in a class comment, (e.g. saying in which version of .NET it's safe by default or not)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        public bool IsUnsafe(InvocationExpressionSyntax invocation, ISymbol settings, NetFrameworkVersion version)
+        {
+            var unsafeDtdProcessing = false;
+            var unsafeResolver = version == NetFrameworkVersion.Probably35 || version == NetFrameworkVersion.Between4And451;
+
+            var objectCreation = GetObjectCreation(settings, invocation, semanticModel);
+            var objectCreationAssignments = objectCreation.GetInitializerExpressions().OfType<AssignmentExpressionSyntax>();
+
+            var assignments = invocation.FirstAncestorOrSelf<MethodDeclarationSyntax>()
```
andrei-epure-sonarsource: maybe rename to `memberAccessAssignments` to clearly differentiate from `objectCreationAssignments`? 

or some better name, I don't know

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        private static IEnumerable<SyntaxNode> GetDescendantNodes(Location location, InvocationExpressionSyntax invocation) =>
+            location.SourceTree?.GetRoot()?.FindNode(location.SourceSpan).DescendantNodes() ??
+            invocation.FirstAncestorOrSelf<MethodDeclarationSyntax>()?.DescendantNodes() ??
+            Enumerable.Empty<SyntaxNode>();
+
+        private static bool IsXmlResolverDtdProcessingUnsafe(AssignmentExpressionSyntax assignment, SemanticModel semanticModel) =>
+            semanticModel.GetConstantValue(assignment.Right).Value switch
+            {
+                false => true,
```
andrei-epure-sonarsource: small comment above that this `false` is for `ProhibitDtd`?

`false =>  true` is inherently difficult to grasp :)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                var name = assignment.Left.GetName();
+
+                if (name =="ProhibitDtd" || name == "DtdProcessing")
+                {
+                    unsafeDtdProcessing = IsXmlResolverDtdProcessingUnsafe(assignment, semanticModel);
+                }
+                else if (assignment.Left.NameIs("XmlResolver"))
+                {
+                    unsafeResolver = IsXmlResolverAssignmentUnsafe(assignment, semanticModel);
```
andrei-epure-sonarsource: until now `unsafeResolver` has actually been `unsafeResolverByDefault`, and now you're overwriting it's value with `unsafeResolverByAssignment`

would it be easier to understand if here it would be:

```
unsafeResolver = IsXmlResolverAssignmentUnsafe(assignment, semanticModel) || unsafeResolverByDefault
```
?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        private static bool IsXmlResolverAssignmentUnsafe(AssignmentExpressionSyntax assignment, SemanticModel semanticModel)
+        {
+            if (assignment.Right.IsKind(SyntaxKind.NullLiteralExpression))
+            {
+                return false;
+            }
+
+            var symbol = semanticModel.GetSymbolInfo(assignment.Right).Symbol;
+            var type = symbol.GetSymbolType() ?? symbol.ContainingType;
```
andrei-epure-sonarsource: I don't understand this line - is it the type of the symbol or the type of the containing type?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XmlExternalEntityShouldNotBeParsed.cs
```diff
@@ -88,9 +91,31 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            var invocation = (InvocationExpressionSyntax)context.Node;
+            if (!invocation.IsMemberAccessOnKnownType("Create", KnownType.System_Xml_XmlReader, context.SemanticModel))
+            {
+                return;
+            }
+
+            var settings = invocation.GetArgumentsOfKnownType(KnownType.System_Xml_XmlReaderSettings, context.SemanticModel).FirstOrDefault();
+            if (settings == null)
+            {
+                return;
```
andrei-epure-sonarsource: I would add a small comment `safe by default`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
costin-zaharia-sonarsource: I've moved the `version` parameter to the constructor and added a field `isXmlResolverSafeByDefault` which is initialized in ctor.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
costin-zaharia-sonarsource: Good point. I've added a comment.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
costin-zaharia-sonarsource: `propertyAssignments`?

first we get the assignments done during the object creation (initialization) and next the property sets

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
costin-zaharia-sonarsource: I've changed the initialization to be `var unsafeResolver = isXmlResolverSafeByDefault;` and now I think it's already easier to understand.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
costin-zaharia-sonarsource: 😄 indeed

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
costin-zaharia-sonarsource: Indeed this is not straight forward. If XmlResolver is set during constructor initialization, `GetSymbolType()` will return null so we have to rely on ContainingType to get the type.

I will add a comment.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/XmlExternalEntityShouldNotBeParsed_XmlReader_Net452.cs
costin-zaharia-sonarsource: Good point I've also found a bug with this because fields were not correctly handled :). 
I've added now a few additional tests similar with what has been done for XmlDocument and XmlTextReader.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -28,39 +29,54 @@
...
 {
+    /// <summary>
+    /// This class is responsible to check if a XmlReaderSettings node is vulnerable to XXE attacks.
+    ///
+    /// By default the class is safe:
+    ///  - before .Net 4.5.2 it has XmlResolver set to an instance of XmlUrlResolver which is not secure but since has ProhibitDtd set to false it is safe to use
+    ///  - starting with .Net 4.5.2 XmlResolver is set to null, ProhibitDtd (now obsolete) set to false and DtdProcessing set to ignore
+    ///
+    /// If the properties are modified, in order to be secure, we have to check that either ProhibitDtd is set to false, DtdProcessing set to Ignore or
+    /// XmlResolver is set to null or is different than XmlUrlResolver and XmlPreloadedResolver.
```
andrei-epure-sonarsource:
```suggestion
    /// the internal XmlResolver is secure.
```

andrei-epure-sonarsource: the actual checks may change in the future (maybe instead of using a blacklist we'll use a whitelist) and can be verified in the code itself.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -28,39 +29,54 @@
 
 namespace SonarAnalyzer.Rules.XXE
 {
+    /// <summary>
+    /// This class is responsible to check if a XmlReaderSettings node is vulnerable to XXE attacks.
+    ///
+    /// By default the class is safe:
+    ///  - before .Net 4.5.2 it has XmlResolver set to an instance of XmlUrlResolver which is not secure but since has ProhibitDtd set to false it is safe to use
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: I hope by default `ProhibitDtd` is set to true. If it's set to false it means it's not secure.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -28,39 +29,54 @@
 
 namespace SonarAnalyzer.Rules.XXE
 {
+    /// <summary>
+    /// This class is responsible to check if a XmlReaderSettings node is vulnerable to XXE attacks.
+    ///
+    /// By default the class is safe:
```
andrei-epure-sonarsource:
```suggestion
    /// By default the XmlReaderSettings is safe:
```

andrei-epure-sonarsource: "This class" at line 33 has a different meaning than "the class" at line 35. Better be explicit.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -28,39 +29,54 @@
...
+        /// <summary>
+        /// Checks if a method invocation (e.g. XmlReader.Create) receives a secure instance of XmlReaderSettings.
+        /// </summary>
+        /// <param name="invocation">A method invocation syntax node (e.g. XmlReader.Create).</param>
+        /// <param name="settings">The symbol of the XmlReaderSettings node received as parameter. This is used to check
+        /// if certain properties (ProhibitDtd, DtdProcessing or XmlUrlResolver) were modified for the given symbol.</param>
+        /// <returns>true, if the XmlReaderSettings is secure; otherwise false.</returns>
+        public bool IsUnsafe(InvocationExpressionSyntax invocation, ISymbol settings)
         {
             var unsafeDtdProcessing = false;
```
andrei-epure-sonarsource: I would add a comment above `unsafeDtdProcessing` to say:
```
// By default ProhibitDtd is 'true' and DtdProcessing is 'ignore'
```
There is no risk that the default values will change, so the comment will likely just stay like.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -77,15 +106,27 @@ public bool IsUnsafe(InvocationExpressionSyntax invocation, ISymbol settings, Ne
...
+                       invocation.SyntaxTree.GetRoot();
+
+            return root.DescendantNodes();
+        }
 
         private static bool IsXmlResolverDtdProcessingUnsafe(AssignmentExpressionSyntax assignment, SemanticModel semanticModel) =>
             semanticModel.GetConstantValue(assignment.Right).Value switch
             {
-                false => true,
+                false => true, // If ProhibitDtd is set to false the settings will be secure (there will be no parsing)
```
andrei-epure-sonarsource:
```suggestion
                false => true, // If ProhibitDtd is set to false the settings will be unsafe (parsing is allowed)
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
```diff
@@ -31,6 +31,7 @@
 
   <ItemGroup>
     <Compile Remove="TestCases\**\*" />
+    <Compile Include="TestCases\XmlExternalEntityShouldNotBeParsed_XmlReader_Net35.cs" />
```
andrei-epure-sonarsource: remove this?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/InvocationExpressionSyntaxExtensions.cs
andrei-epure-sonarsource: ping @costin-zaharia-sonarsource

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -28,39 +29,54 @@
 
 namespace SonarAnalyzer.Rules.XXE
 {
+    /// <summary>
+    /// This class is responsible to check if a XmlReaderSettings node is vulnerable to XXE attacks.
+    ///
+    /// By default the class is safe:
+    ///  - before .Net 4.5.2 it has XmlResolver set to an instance of XmlUrlResolver which is not secure but since has ProhibitDtd set to false it is safe to use
+    ///  - starting with .Net 4.5.2 XmlResolver is set to null, ProhibitDtd (now obsolete) set to false and DtdProcessing set to ignore
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Indeed it's true :)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
costin-zaharia-sonarsource: It seems I've messed up all the comments 😄

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -55,7 +55,7 @@ public XmlReaderSettingsValidator(SemanticModel semanticModel, NetFrameworkVersi
         /// <param name="invocation">A method invocation syntax node (e.g. XmlReader.Create).</param>
         /// <param name="settings">The symbol of the XmlReaderSettings node received as parameter. This is used to check
         /// if certain properties (ProhibitDtd, DtdProcessing or XmlUrlResolver) were modified for the given symbol.</param>
-        /// <returns>true, if the XmlReaderSettings is secure; otherwise false.</returns>
+        /// <returns>true, if the XmlReaderSettings is not secure; otherwise false.</returns>
```
andrei-epure-sonarsource: in all honesty you can just remove this line, the method name is clear enough

PR #3177, opened 2020-03-04
Rule S2068: filter string literal that contains the wordlist item

### LINKED ISSUE #2868: Rule S2068: filter string literal that contains the wordlist item

The implementation of this rule highly rely on symbol names matching wordlist items to raise issues. The downside of this is that it raises many FPs when constants are used to avoid duplicated stings:

```csharp
string password = "Password"; // Compliant
string pwd = "pwd"; // Compliant

public class S2068 {

  private const string PASSWORD = "Password"; // Compliant
  private const string PASSWORD_INPUT = "[id='password']"; // Compliant
  private const string PASSWORD_PROPERTY = "custom.password"; // Compliant
  private const string TRUSTSTORE_PASSWORD = "trustStorePassword"; // Compliant
  private const string CONNECTION_PASSWORD = "connection.password"; // Compliant
  private const string RESET_PASSWORD = "/users/resetUserPassword"; // Compliant

// ...
}
```

In most string constants use cases the wordlist item is present in both the symbol name and the string value.
The new approach is to avoid raising issues when **the wordlist item is present in both symbol name and literal string value**.
Exception made for the following use cases that are still True Positives:

```csharp
string params = "user=admin&password=Password123"; // Sensitive
string sqlserver = "pgsql:host=localhost port=5432 dbname=test user=postgres password=postgres"; // Sensitive
```

pavel-mikula-sonarsource:
Fixes #2868

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
@@ -151,7 +151,7 @@ namespace SonarAnalyzer.Rules
                     return Enumerable.Empty<string>();
                 }
 
-                var credentialWordsFound = (variableName ==null || variableName.Any(x => char.IsLower(x)) ? variableName : variableName.ToLower()) // Prepare "PASSWORD" for SplitCamelCaseToWords()
+                var credentialWordsFound = variableName
                     .SplitCamelCaseToWords()
                     .Intersect(analyzer.splitCredentialWords)
                     .ToHashSet(StringComparer.OrdinalIgnoreCase);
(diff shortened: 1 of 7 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
```diff
@@ -151,11 +151,17 @@ private IEnumerable<string> FindCredentialWords(string variableName, string vari
                     return Enumerable.Empty<string>();
                 }
 
-                var credentialWordsFound = variableName
+                var credentialWordsFound = (variableName ==null || variableName.Any(x => char.IsLower(x)) ? variableName : variableName.ToLower()) // Prepare "PASSWORD" for SplitCamelCaseToWords()
```
christophe-zurn-sonarsource: I think this fix deserves a separate ticket: fixing camelcase split for fully uppercase words should be directly done inside the `SplitCamelCaseToWords` method IMO.
But since it would impact other part of the project, I suggest to do it in a separate commit.
WDYT ?

pavel-mikula-sonarsource:
SplitCamelCaseToWords is addressed in PR #3188 that needs to be merged first.

pavel-mikula-sonarsource:
Rebasing issue was fixed. There was duplicate commit adding duplicate expected issue as in #3188  PR.

PR #3188, opened 2020-03-09
SplitCamelCaseToWords handles UPPERCASE sequences

pavel-mikula-sonarsource:
Strings like `HTML`  or `PEHeader` were handled per upper case letter and splitted into `h, t, m, l` or `p, e, header`.

This causes some problems with upper-case only names like `PASSWORD` or `URL`.

New implementation is aware of upper case sequences and can handle situations like `ThisIsSMTPName` to `this, is, smtp, name`

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/Helpers/StringExtensions.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/Helpers/StringExtensions.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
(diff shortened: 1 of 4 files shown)


pavel-mikula-sonarsource:
SC QG will fail due to coverage generation issue. Might be worked around by two ugly ways:
- adding fake dummy method into source file.
- marking with ExcludeFromCodeCoverageAttribute.
Both are bad ideas.

christophe-zurn-sonarsource:
It seems the coverage is not correctly reported for iterator methods. The [following class](https://sonarcloud.io/code?id=sonaranalyzer-dotnet&selected=sonaranalyzer-dotnet%3Asonaranalyzer-dotnet%2Fsrc%2FSonarAnalyzer.Common%2FHelpers%2FSymbolHelper.cs) also has a `GetAllNamedTypes` iterator method, which is clearly covered by `Symbol_GetAllNamedTypes_Namespace` of `SymbolHelperTest`. However, no coverage is reported either... It seems a bug, but I cannot find anything current on miscrosoft side. It seems big enough that something should be there...

sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/Helpers/StringExtensions.cs
```diff
@@ -23,49 +23,60 @@
...
         /// bin2hex => bin hex
-        /// HTML => h t m l
-        /// PEHeader => p e header
+        /// HTML => html
+        /// SOME_value => some value
+        /// PEHeader => pe header
         /// </summary>
         /// <param name="name">A string containing words.</param>
         /// <returns>A list of words (all lowercase) contained in the string.</returns>
         public static IEnumerable<string> SplitCamelCaseToWords(this string name)
```
christophe-zurn-sonarsource: Although we cannot do anything regarding coverage, I think we could improve the duplication.
I looked at the code, and this method is only used [in a single place](https://github.com/SonarSource/sonar-dotnet/blob/1946c1b77e124446a38f301bb7fd10f67a0f0fce/sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/ControlFlowGraph/CfgSerializer.cs#L128) in the CFG assembly (it is not used either in the security assembly, as it is present there as another duplicated method...).

I'd suggest to remove the implementation from CFG dll and use a simplified version inside the `WriteNode` method given that the context in which it is used is limited.

WDYT ?

pavel-mikula-sonarsource: I don't like the duplication, but I replaced it anyway. It will be easier to decouple the helpers layer in the future if we keep it as close as possible.

All helpers in CFG are duplicates by-design, for now.

christophe-zurn-sonarsource: Ok. Let's keept it like this for now, and indeed see if/when we will have to decouple some logic between these projects

Issue #2537, opened 2019-08-05
Rule S4433 LDAP: connections should be authenticated (false positive)


gfox1984:
### Description

When the authentication type is set to a secure values using the ternary operator, Sonar triggers a false positive.

### Repro steps

```
AuthenticationTypes authTypes = condition ? AuthenticationTypes.SecureSocketsLayer : AuthenticationTypes.Secure;
DirectoryEntry entry = new DirectoryEntry(connectionString, null, null, authTypes); 
```

### Expected behavior

> This rule raises an issue when an LDAP connection is created with AuthenticationTypes.Anonymous or AuthenticationTypes.None.

### Actual behavior

An issue is raised despite not using AuthenticationTypes.Anonymous or AuthenticationTypes.None.

### Related information

* SonarC# 7.15.0.8572

andrei-epure-sonarsource:
I was wrong when labeling `taint-analysis`, this is actually a vulnerability rule from sonar-dotnet, see [LdapConnectionShouldBeSecure ](https://github.com/SonarSource/sonar-dotnet/blob/3d6d2303b7afbd40bcac8aff1a48b4f8bc8a8d48/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/LdapConnectionShouldBeSecure.cs)

PR #3197, opened 2020-03-10
Fix S4433 FP

costin-zaharia-sonarsource:
Fixes: #2537

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/LdapConnectionShouldBeSecure.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/LdapConnectionShouldBeSecure.cs
@@ -53,28 +53,37 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static bool IsAllowedValue(object constantValue) =>
             constantValue is int integerValue &&
-            IsValidAuthenticationType(integerValue);
+            !IsUnsafe(integerValue);
 
         private static readonly ImmutableArray<KnownType> TrackedTypes = ImmutableArray.Create(KnownType.System_DirectoryServices_DirectoryEntry);
 
-        private static bool IsAllowedObject(ISymbol symbol, SyntaxNode node, SemanticModel semanticModel)
+        /// <summary>
+        /// Verifies if the AuthenticationType parameter used on constructor invocation is secure.
+        /// </summary>
+        /// <param name="authTypeSymbol">The symbol associated with the expression used to populate the AuthenticationType parameter</param>
+        /// <param name="authTypeExpression">Expression used to populate the AuthenticationType parameter</param>
+        private static bool IsAllowedObject(ISymbol authTypeSymbol, ExpressionSyntax authTypeExpression, SemanticModel semanticModel)
         {
-            if (!symbol.GetSymbolType().Is(KnownType.System_DirectoryServices_AuthenticationTypes))
+            if (!authTypeSymbol.GetSymbolType().Is(KnownType.System_DirectoryServices_AuthenticationTypes))
             {
                 return false;
             }
 
-            // In order to correctly detect all the cases we will need to refactor the rule and use symbolic execution.
+            // In order to correctly detect all the cases we will need to refactor the rule to use symbolic execution.
             // Until this is done, we can reduce the number of false positives by checking if the variable has one of the following
             // values assigned in the parent scope: AuthenticationTypes.None, AuthenticationTypes.Anonymous, default or default(AuthenticationTypes).
-            var root = node.FirstAncestorOrSelf<MethodDeclarationSyntax>() ??
-                       node.FirstAncestorOrSelf<ClassDeclarationSyntax>() ??
-                       node.FirstAncestorOrSelf<StructDeclarationSyntax>() ??
-                       node.FirstAncestorOrSelf<InterfaceDeclarationSyntax>() ??
-                       node.SyntaxTree.GetRoot();
+            var root = authTypeExpression.FirstAncestorOrSelf<MethodDeclarationSyntax>() ??
+                       authTypeExpression.FirstAncestorOrSelf<ClassDeclarationSyntax>() ??
+                       authTypeExpression.FirstAncestorOrSelf<StructDeclarationSyntax>() ??
+                       authTypeExpression.FirstAncestorOrSelf<InterfaceDeclarationSyntax>() ??
+                       authTypeExpression.SyntaxTree.GetRoot();
 
-            return !HasUnsafeDeclaration(root, symbol, semanticModel) &&
-                   !HasUnsafeAssignment(root, symbol, semanticModel);
+            // The CSharpObjectInitializationTracker is verifying only if AuthenticationType property or DirectoryEntry ctor
+            // was invoked with a valid value, which can be evaluated as a constant.
+            // If the parameter is a variable, in order to reduce FP, we have to check if an unsafe value was set during
+            // variable declaration or later by assignment.
+            return !HasUnsafeDeclaration(root, authTypeSymbol, semanticModel) &&
+                   !HasUnsafeAssignment(root, authTypeSymbol, semanticModel);
         }
 
         private static bool HasUnsafeAssignment(SyntaxNode root, ISymbol symbol, SemanticModel semanticModel) =>
@@ -102,9 +111,9 @@ namespace SonarAnalyzer.Rules.CSharp
             semanticModel.GetConstantValue(node) is {} constantValue &&
             constantValue.HasValue &&
             constantValue.Value is int authType &&
-            !IsValidAuthenticationType(authType);
+            IsUnsafe(authType);
 
-        private static bool IsValidAuthenticationType(int authType) =>
-            authType != AuthenticationTypesNone && authType != AuthenticationTypesAnonymous;
+        private static bool IsUnsafe(int authType) =>
+            authType == AuthenticationTypesNone || authType == AuthenticationTypesAnonymous;
     }
 }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/LdapConnectionShouldBeSecure.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/LdapConnectionShouldBeSecure.cs
@@ -19,7 +19,9 @@
  */
 
 using System.Collections.Immutable;
+using System.Linq;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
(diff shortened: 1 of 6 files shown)


andrei-epure-sonarsource:
related to #2537

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XmlExternalEntityShouldNotBeParsed.cs
```diff
@@ -216,7 +216,7 @@ private static bool IsAllowedValueForXmlTextReader(object constantValue)
                     symbol.Kind == SymbolKind.Method &&
                     symbol.ContainingType.GetSymbolType().IsAny(UnsafeXmlResolvers);
 
-            private static bool IsAllowedObject(ISymbol symbol) =>
+            private static bool IsAllowedObject(ISymbol symbol, SyntaxNode node, SemanticModel model) =>
```
andrei-epure-sonarsource: I'd comment that we don't use `node` and `model` on purpose (to avoid confusion in the future)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/LdapConnectionShouldBeSecure.cs
```diff
@@ -129,4 +129,64 @@ void Cases()
             Property6 = new DirectoryEntry(); // Compliant
         }
     }
+
+    public class AllTypes
```
andrei-epure-sonarsource: can you also add test cases (unless I miss them) for initializations inside
- constructor
- inside getter / setter (for properties we only have `private DirectoryEntry Property3 { get; set; } = new DirectoryEntry()...`)
- struct
- interface default method

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/LdapConnectionShouldBeSecure.cs
```diff
@@ -43,14 +47,64 @@ public sealed class LdapConnectionShouldBeSecure : ObjectShouldBeInitializedCorr
...
 
         private static bool IsAllowedValue(object constantValue) =>
             constantValue is int integerValue &&
-            (integerValue & AuthenticationTypes_Secure) > 0; // The expected value is a bit from a Flags enum
-
+            IsValidAuthenticationType(integerValue);
 
         private static readonly ImmutableArray<KnownType> TrackedTypes = ImmutableArray.Create(KnownType.System_DirectoryServices_DirectoryEntry);
+
+        private static bool IsAllowedObject(ISymbol symbol, SyntaxNode node, SemanticModel semanticModel)
```
andrei-epure-sonarsource: could you add a comment for this rule what is the symbol, what is the node ?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/LdapConnectionShouldBeSecure.cs
```diff
@@ -43,14 +47,64 @@ public sealed class LdapConnectionShouldBeSecure : ObjectShouldBeInitializedCorr
...
+            root.DescendantNodes()
+                .OfType<VariableDeclaratorSyntax>()
+                .Where(variableDeclarator => VariableDeclaratorHasSymbol(variableDeclarator, symbol, semanticModel))
+                .Any(variableDeclarator => HasUnsafeConstantValue(variableDeclarator.Initializer.Value, semanticModel));
+
+        private static bool VariableDeclaratorHasSymbol(VariableDeclaratorSyntax variableDeclarator, ISymbol symbol, SemanticModel semanticModel) =>
+            variableDeclarator.Identifier.ValueText == symbol.Name &&
+            symbol.Equals(semanticModel.GetDeclaredSymbol(variableDeclarator));
+
+        private static bool HasUnsafeConstantValue(SyntaxNode node, SemanticModel semanticModel) =>
```
andrei-epure-sonarsource: the logic is a bit hard to follow
- HasUnsafeConstantValue - negative in name
- `!IsValidAuthenticationType(authType)` - negation with positive name
- `IsValidAuthenticationType` contains negations inside `authType != AuthenticationTypesNone && authType != AuthenticationTypesAnonymous`

negative logic is harder to follow

if would be easier to read 
```
private static bool HasUnsafeConstantValue(SyntaxNode node, SemanticModel semanticModel) =>
 // ...
 // IsUnsafeAuthenticationType(authType);	 
```

and then the only negation would be in the names , and would be consistent

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/LdapConnectionShouldBeSecure.cs
```diff
@@ -43,14 +47,64 @@ public sealed class LdapConnectionShouldBeSecure : ObjectShouldBeInitializedCorr
...
+            // In order to correctly detect all the cases we will need to refactor the rule and use symbolic execution.
+            // Until this is done, we can reduce the number of false positives by checking if the variable has one of the following
+            // values assigned in the parent scope: AuthenticationTypes.None, AuthenticationTypes.Anonymous, default or default(AuthenticationTypes).
+            var root = node.FirstAncestorOrSelf<MethodDeclarationSyntax>() ??
+                       node.FirstAncestorOrSelf<ClassDeclarationSyntax>() ??
+                       node.FirstAncestorOrSelf<StructDeclarationSyntax>() ??
+                       node.FirstAncestorOrSelf<InterfaceDeclarationSyntax>() ??
+                       node.SyntaxTree.GetRoot();
+
+            return !HasUnsafeDeclaration(root, symbol, semanticModel) &&
```
andrei-epure-sonarsource: it would be nice to add a comment specifying what these two checks do, the scenario(s), because it's not obvious and the methods themselves don't help either

and basically explain how it's complementary to what the ObjectInitializationTracker verifies

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/LdapConnectionShouldBeSecure.cs
```diff
@@ -129,4 +129,64 @@ void Cases()
...
+        {
+            var authType = condition ? AuthenticationTypes.SecureSocketsLayer : AuthenticationTypes.Secure;
+            var entry = new DirectoryEntry("", null, null, authType);
+        }
+
+        public void WithTernaryOperatorUnsecure(bool condition)
+        {
+            var authType = condition ? AuthenticationTypes.None : AuthenticationTypes.Anonymous;
+            // Symbolic execution would a better fit to increase precision
+            var entry = new DirectoryEntry("", null, null, authType); // Compliant - FN
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: as discussed, I will not open a new issue since this is by design (the rule does not use symbolic execution) and this kind of false negatives is expected.
Indeed we need to make this more visible to the AppSec team but this will be handled separately.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/LdapConnectionShouldBeSecure.cs
```diff
@@ -53,28 +53,37 @@ public sealed class LdapConnectionShouldBeSecure : ObjectShouldBeInitializedCorr
...
-
-            return !HasUnsafeDeclaration(root, symbol, semanticModel) &&
-                   !HasUnsafeAssignment(root, symbol, semanticModel);
+            var root = authTypeExpression.FirstAncestorOrSelf<MethodDeclarationSyntax>() ??
+                       authTypeExpression.FirstAncestorOrSelf<ClassDeclarationSyntax>() ??
+                       authTypeExpression.FirstAncestorOrSelf<StructDeclarationSyntax>() ??
+                       authTypeExpression.FirstAncestorOrSelf<InterfaceDeclarationSyntax>() ??
+                       authTypeExpression.SyntaxTree.GetRoot();
+
+            // The CSharpObjectInitializationTracker is verifying only if AuthenticationType property or DirectoryEntry ctor
```
andrei-epure-sonarsource:
```suggestion
            // The CSharpObjectInitializationTracker is verifying only if the AuthenticationType property or DirectoryEntry ctor
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/LdapConnectionShouldBeSecure.cs
```diff
@@ -53,28 +53,37 @@ public sealed class LdapConnectionShouldBeSecure : ObjectShouldBeInitializedCorr
...
+            // was invoked with a valid value, which can be evaluated as a constant.
```
andrei-epure-sonarsource:
```suggestion
            // was invoked with a valid constant value.
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/LdapConnectionShouldBeSecure.cs
```diff
@@ -53,28 +53,37 @@ public sealed class LdapConnectionShouldBeSecure : ObjectShouldBeInitializedCorr
...
+            // If the parameter is a variable, in order to reduce FP, we have to check if an unsafe value was set during
```
andrei-epure-sonarsource:
```suggestion
            // When the initialization is done with a variable, we check if an unsafe value was set during
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XmlExternalEntityShouldNotBeParsed.cs
```diff
@@ -173,15 +174,15 @@ public static (CSharpObjectInitializationTracker xmlDocumentTracker, CSharpObjec
                     isAllowedConstantValue: constantValue => false,
                     trackedTypes: XmlDocumentTrackedTypes,
                     isTrackedPropertyName: propertyName => "XmlResolver" == propertyName,
-                    isAllowedObject: IsAllowedObject,
+                    isAllowedObject: (symbol, _, __) => IsAllowedObject(symbol),
```
andrei-epure-sonarsource: is it me or are there two underscores `__` ?
is it intentional?
(also see line 185)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpObjectInitializationTracker.cs
```diff
@@ -56,9 +56,9 @@ public class CSharpObjectInitializationTracker
         private readonly Predicate<object> isAllowedConstantValue;
 
         /// <summary>
-        /// Given the symbol of an object, analyzed syntaxNode and the semantic model, returns true if it is allowed.
+        /// Given the symbol of an object, expresion used to populate the value and the semantic model, returns true if it is allowed.
```
andrei-epure-sonarsource:
```suggestion
        /// Given the symbol of an object, the expression used to populate the value and the semantic model, returns true if it is allowed.
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/XmlExternalEntityShouldNotBeParsed.cs
costin-zaharia-sonarsource: Indeed there are two underscores and it was intentional. AFAICT it's not possible to use _ for both parameters. The compiler will complain that the parameter name is a duplicate.

PR #3199, opened 2020-03-11
Update SonarLint profiles after activating more rules

andrei-epure-sonarsource:


pavel-mikula-sonarsource:
@andrei-epure-sonarsource There's one problem with this - for whatever reason, this file was not updated `sonaranalyzer-dotnet\.sonarlint\SonarLint.xml` by connected mode so SL4VS asks to update headers to 2019 😄 
Can you please check the configuration (Issues does not appear in SC), so it might be SL4VS issue?

PR #3202, opened 2020-03-11
Rule S5659 JWT should be signed and verified with strong cipher algorithms

### LINKED ISSUE #2866: Rule S5659 JWT should be signed and verified with strong cipher algorithms
Implements [RSPEC-5659](https://jira.sonarsource.com/browse/RSPEC-5659)

Code examples: https://github.com/SonarSource/security-expected-issues/blob/master/dotnet/wip/basic-rules/RSPEC_5659/Rule5659.cs

pavel-mikula-sonarsource:
Fixes #2866


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SyntaxTrackers/InvocationTracker.cs
```diff
@@ -116,6 +116,10 @@ bool IsTrackedMethod(SyntaxNode invocation, SemanticModel semanticModel)
                 context.MethodSymbol.Value != null &&
                 context.MethodSymbol.Value.ReturnType.DerivesFrom(returnType);
 
+        public InvocationCondition InvalidBuilderInitialization<TInvocationSyntax>(BuilderPatternCondition<TInvocationSyntax> condition)
```
andrei-epure-sonarsource: `IsInvalidBuilderInitialization` ? (to use a verb for the method name)

pavel-mikula-sonarsource:
QG is failing due to C# vs. VB.NET duplication - there's same syntax for different namespaces. After discussion with @andrei-epure-sonarsource , this is reoccurring problem in our code base and I'll ask should a Champion for merge.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpBuilderPatternCondition.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        public CSharpBuilderPatternCondition(bool constructorIsSafe, params BuilderPatternDescriptor<InvocationExpressionSyntax>[] descriptors) : base(constructorIsSafe, descriptors) { }
+
+        protected override SyntaxNode RemoveParentheses(SyntaxNode node) =>
+            node.RemoveParentheses();
+
+        protected override SyntaxNode TopMostContainingMethod(SyntaxNode node) =>
+            node.GetTopMostContainingMethod();
+
+        protected override SyntaxNode InvocationExpression(InvocationExpressionSyntax node, out string identifierName)
```
costin-zaharia-sonarsource: I would rename this `GetExpression`. It receives an invocation as parameter so the invocation from name is redundant.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpBuilderPatternCondition.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
```
costin-zaharia-sonarsource: This method is doing 2 different things and returns 2 results: the invocation expression and the identifier name. Would be great to separate the responsibilities.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpBuilderPatternCondition.cs
```diff
@@ -0,0 +1,96 @@
...
+            identifierName = node.Expression.GetIdentifier()?.Identifier.ValueText;
```
costin-zaharia-sonarsource: `node.Expression.GetIdentifier()?.Identifier.ValueText` can be replaced with `node.Expression.GetName()`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpBuilderPatternCondition.cs
```diff
@@ -0,0 +1,96 @@
...
+            return node.Expression;
+        }
+
+        protected override bool IsMemberAccess(SyntaxNode node, out SyntaxNode expression)
```
costin-zaharia-sonarsource: Can be renamed to `TryGetMemberAccessExpression`. Using a try prefix is a common naming pattern for this kind of methods which try to get a result and return a boolean value.

Also the out parameter name, expression, is confusing since both `MemberAccessExpressionSyntax` and  `MemberAccessExpressionSyntax` are expressions and as a caller you don't know, without looking at the implementation, what you get.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpBuilderPatternCondition.cs
```diff
@@ -0,0 +1,96 @@
...
+    public class CSharpBuilderPatternCondition : BuilderPatternCondition<InvocationExpressionSyntax>
+    {
+
+        public CSharpBuilderPatternCondition(bool constructorIsSafe, params BuilderPatternDescriptor<InvocationExpressionSyntax>[] descriptors) : base(constructorIsSafe, descriptors) { }
+
+        protected override SyntaxNode RemoveParentheses(SyntaxNode node) =>
+            node.RemoveParentheses();
+
+        protected override SyntaxNode TopMostContainingMethod(SyntaxNode node) =>
+            node.GetTopMostContainingMethod();
```
costin-zaharia-sonarsource: I find `GetTopMostContainingMethod` much better than `TopMostContainingMethod`.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/SyntaxTrackers/InvocationTracker.cs
```diff
@@ -116,6 +116,10 @@ bool IsTrackedMethod(SyntaxNode invocation, SemanticModel semanticModel)
                 context.MethodSymbol.Value != null &&
                 context.MethodSymbol.Value.ReturnType.DerivesFrom(returnType);
 
+        public InvocationCondition IsInvalidBuilderInitialization<TInvocationSyntax>(BuilderPatternCondition<TInvocationSyntax> condition)
+            where TInvocationSyntax : SyntaxNode =>
+            context => condition.InvalidBuilderInitialization(context);
```
costin-zaharia-sonarsource: `context => condition.InvalidBuilderInitialization(context);` can be replaced with `condition.InvalidBuilderInitialization;`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpBuilderPatternCondition.cs
```diff
@@ -0,0 +1,96 @@
...
+            return false;
+        }
+
+        protected override bool IsAssignmentToIdentifier(SyntaxNode node, string identifierName, out SyntaxNode rightExpression)
+        {
+            if (node is ExpressionStatementSyntax statement)
+            {
+                node = statement.Expression;
+            }
+            if (node is AssignmentExpressionSyntax assignment && assignment.Left.GetIdentifier()?.Identifier.ValueText == identifierName)
```
costin-zaharia-sonarsource: `assignment.Left.GetIdentifier()?.Identifier.ValueText == identifierName` can be replaced with `assignment.Left.NameIs(identifierName)`

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/JwtSignedBase.cs
```diff
@@ -0,0 +1,79 @@
...
+                    InvocationTracker.MethodHasParameters(1)
+                    ));
+
+            InvocationTracker.Track(context,
+                InvocationTracker.MatchMethod(new MemberDescriptor(KnownType.JWT_Builder_JwtBuilder, "Decode")),
+                InvocationTracker.IsInvalidBuilderInitialization(BuilderPattern())
+                );
+        }
+
+        protected BuilderPatternDescriptor<TInvocationSyntax>[] JwtBuilderDescriptors(Func<InvocationContext, TInvocationSyntax, bool> singleArgumentIsNotFalseLiteral)
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: It's not used in this rule, but it's there for reusability purposes. Future usages might need to know what `context.Invocation` started this tracking to make correct decision.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpBuilderPatternCondition.cs
pavel-mikula-sonarsource: I don't like "Get" prefix, but I don't have any better idea here 🙂

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/JwtSignedBase.cs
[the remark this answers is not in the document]
costin-zaharia-sonarsource: This is just an assumption currently we don't know yet for sure if this will be needed or not. I suggest to remove this now and add it later when needed.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpBuilderPatternCondition.cs
pavel-mikula-sonarsource: I've renamed the argument to `memberAccessExpression`. 
I agree with the TryGet naming convention in general, but primary branching condition of these 5 functions is to check the type. No matter what is extracted as secondary output value. TryGet should return false for null, but these functions does not do/check/care about that.

I see that this is not 100% perfect solution, but since we don't have that C# vs. VB.NET layer yet, I think this is good for what we have now.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpBuilderPatternCondition.cs
pavel-mikula-sonarsource: Only in this case, because the original extension does already contains redundant and unneeded Get prefix. The `Get` prefix is redundant to the `= ... ()` syntax 🙂

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/BuilderPatternDescriptorTest.cs
```diff
@@ -0,0 +1,95 @@
...
+{
+    [TestClass]
+    public class BuilderPatternDescriptorTest
+    {
+        [TestMethod]
+        public void IsMatch()
+        {
+            var context = CreateContext();
+            InvocationCondition trueCondition = (context) => true;
+            InvocationCondition falseCondition = (context) => false;
```
costin-zaharia-sonarsource: pedantic: parentheses are redundant here

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/BuilderPatternDescriptorTest.cs
```diff
@@ -0,0 +1,95 @@
...
+            var context = CreateContext();
+
+            var descriptor = new BuilderPatternDescriptor<InvocationExpressionSyntax>((context, invocation) => context != null && invocation != null);
+            descriptor.IsValid(null, null).Should().BeFalse();
+            descriptor.IsValid(context, null).Should().BeFalse();
+            descriptor.IsValid(null, context.Invocation as InvocationExpressionSyntax).Should().BeFalse();
+            descriptor.IsValid(context, context.Invocation as InvocationExpressionSyntax).Should().BeTrue();
+        }
+
+        private InvocationContext CreateContext()
```
costin-zaharia-sonarsource: can be static

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/SyntaxHelperTest.cs
```diff
@@ -0,0 +1,99 @@
...
+            toString.NameIs("tostring").Should().BeTrue();
+            toString.NameIs("test").Should().BeFalse();
+            toString.NameIs("").Should().BeFalse();
+            toString.NameIs(null).Should().BeFalse();
+        }
+
+        private SyntaxNode[] Parse_CS(string source) =>
+            Microsoft.CodeAnalysis.CSharp.CSharpSyntaxTree.ParseText(source).GetRoot().DescendantNodes().ToArray();
+
+        private SyntaxNode[] Parse_VB(string source) =>
```
costin-zaharia-sonarsource: The methods Parse_CS and Parse_VB can be made static.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SyntaxTrackers/BuilderPatternConditionTest.cs
```diff
@@ -0,0 +1,221 @@
...
+
+            var condition7 = new CSharpBuilderPatternCondition(true, bbbValidator, cccInvalidator);
+            condition7.InvalidBuilderInitialization(context_CS).Should().BeTrue(); // Valid constructor, validated by Bbb, invalidated by Ccc
+
+            var condition8 = new CSharpBuilderPatternCondition(true, cccInvalidator, bbbValidator); // Configuration order has no effect
+            condition8.InvalidBuilderInitialization(context_CS).Should().BeTrue(); // Valid constructor, validated by Bbb, invalidated by Ccc
+
+            var dddContext = new InvocationContext(FindMethodInvocation_CS(context_CS.Invocation.SyntaxTree, "Ddd"), "Ddd", context_CS.SemanticModel);
+            var condition9 = new CSharpBuilderPatternCondition(false);
+            condition9.InvalidBuilderInitialization(dddContext).Should().BeFalse(); // Invalid constructor is not tracked throught array propagation
```
costin-zaharia-sonarsource: pedantic: typo at "throught"

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SyntaxTrackers/BuilderPatternConditionTest.cs
```diff
@@ -0,0 +1,221 @@
...
+
+            var condition7 = new VisualBasicBuilderPatternCondition(true, bbbValidator, cccInvalidator);
+            condition7.InvalidBuilderInitialization(context_VB).Should().BeTrue(); // Valid constructor, validated by Bbb, invalidated by Ccc
+
+            var condition8 = new VisualBasicBuilderPatternCondition(true, cccInvalidator, bbbValidator); // Configuration order has no effect
+            condition8.InvalidBuilderInitialization(context_VB).Should().BeTrue(); // Valid constructor, validated by Bbb, invalidated by Ccc
+
+            var dddContext = new InvocationContext(FindMethodInvocation_VB(context_VB.Invocation.SyntaxTree, "Ddd"), "Ddd", context_VB.SemanticModel);
+            var condition9 = new VisualBasicBuilderPatternCondition(false);
+            condition9.InvalidBuilderInitialization(dddContext).Should().BeFalse(); // Invalid constructor is not tracked throught array propagation
```
costin-zaharia-sonarsource: pedantic: typo also here

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/JwtSigned.cs
```diff
@@ -0,0 +1,243 @@
...
+using System.Collections.Generic;
+using JWT;
+using JWT.Algorithms;
+using JWT.Builder;
+
+namespace Tests.Diagnostics
+{
+    class Program
+    {
+        const string secret = "GQDstcKsx0NHjPOuXOYg5MbeJ1XT0uFiwDVvVBrk";
```
costin-zaharia-sonarsource: Would be nice to add a few more tests in:
- constructor
- nested function
- inside lambda
- inside property
- inside interface with default members
...

pavel-mikula-sonarsource: I've added FP and FN for local function. Placing code into different wrapping syntax does not bring value, before rule does not depend on parent wrapper and we don't usually do that in that case.

PR #3203, opened 2020-03-11
Fix S2068 FN: Raise issue for literals passed as arguments

### LINKED ISSUE #3175: Fix S2068 FN: Raise issue for literals passed as arguments
### Description

Issue is not raised when hardcoded password is used directly as argument.

### Repro steps

```
Using Conn As SqlConnection = OpenConn("Server=localhost; Database=Test; User=SA; Password=Secret123") ' Noncompliant
End Using
Using Conn As New SqlConnection("Server=localhost; Database=Test; User=SA; Password=Secret123") ' Noncompliant
End Using
```
### Other

We can fix other FNs when registering directly for string literal values as well:

```
Dim fn1 As String = "scheme://user:azerty123@" & Domain  ' Compliant FN, concatenated strings are not supported

Dim query2 As String = "password=hardcoded;user='" + user + "'" ' False Negative - Only LiteralExpressionSyntax nodes are covered
```

pavel-mikula-sonarsource:
Fixes #3175

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
@@ -96,6 +96,12 @@ namespace SonarAnalyzer.Rules.CSharp
                 syntaxNode.Right.IsKind(SyntaxKind.StringLiteralExpression);
         }
 
+        /// <summary>
+        /// This finder checks all string literal in the code, except VariableDeclarator and SimpleAssignmentExpression. These two have their own
+        /// finders with precise logic and variable name checking.
+        /// This class inspects all other standalone string literals for values considered as hardcoded passwords (in connection strings)
+        /// based on same rules as in VariableDeclarationBannedWordsFinder and AssignmentExpressionBannedWordsFinder.
+        /// </summary>
         private class StringLiteralBannedWordsFinder : CredentialWordsFinderBase<LiteralExpressionSyntax>
         {
             public StringLiteralBannedWordsFinder(DoNotHardcodeCredentialsBase<SyntaxKind> analyzer) : base(analyzer) { }
@@ -103,6 +109,9 @@ namespace SonarAnalyzer.Rules.CSharp
             protected override string GetAssignedValue(LiteralExpressionSyntax syntaxNode) =>
                 syntaxNode.GetStringValue();
 
+            // We don't have a variable for cases that this finder should handle.  Cases with variable name are
+            // handled by VariableDeclarationBannedWordsFinder and AssignmentExpressionBannedWordsFinder
+            // Returning null is safe here, it will not be considered as a value.
             protected override string GetVariableName(LiteralExpressionSyntax syntaxNode) =>
                 null;
 
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
@@ -133,6 +133,7 @@ namespace SonarAnalyzer.Rules.CSharp
                             break;
                     }
                 }
+                // We want to handle all other literals (property initializers, return statement and return values from lambdas, arrow functions, ...)
                 return true;
             }
         }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
@@ -57,6 +57,10 @@ namespace SonarAnalyzer.Rules.CSharp
                     c.RegisterSyntaxNodeActionInNonGenerated(
                         new AssignmentExpressionBannedWordsFinder(this).GetAnalysisAction(rule),
                         SyntaxKind.SimpleAssignmentExpression);
+
+                    c.RegisterSyntaxNodeActionInNonGenerated(
+                        new StringLiteralBannedWordsFinder(this).GetAnalysisAction(rule),
+                        SyntaxKind.StringLiteralExpression);
                 });
         }
 
@@ -90,6 +94,41 @@ namespace SonarAnalyzer.Rules.CSharp
                 syntaxNode.IsKind(SyntaxKind.SimpleAssignmentExpression) &&
                 syntaxNode.Left.IsKnownType(KnownType.System_String, semanticModel) &&
                 syntaxNode.Right.IsKind(SyntaxKind.StringLiteralExpression);
+        }
+
+        private class StringLiteralBannedWordsFinder : CredentialWordsFinderBase<LiteralExpressionSyntax>
+        {
+            public StringLiteralBannedWordsFinder(DoNotHardcodeCredentialsBase<SyntaxKind> analyzer) : base(analyzer) { }
+
+            protected override string GetAssignedValue(LiteralExpressionSyntax syntaxNode) =>
+                syntaxNode.GetStringValue();
+
+            protected override string GetVariableName(LiteralExpressionSyntax syntaxNode) =>
+                null;
+
+            protected override bool IsAssignedWithStringLiteral(LiteralExpressionSyntax syntaxNode, SemanticModel semanticModel) =>
+                syntaxNode.IsKind(SyntaxKind.StringLiteralExpression) && !IsHandledByOtherFinder(syntaxNode.GetTopMostContainingMethod(), syntaxNode);
+
+            private static bool IsHandledByOtherFinder(SyntaxNode method, SyntaxNode current)
+            {
+                while (current != null && current != method)
+                {
+                    switch (current.Kind())
+                    {
+                        case SyntaxKind.VariableDeclarator:
+                        case SyntaxKind.SimpleAssignmentExpression:
+                            return true;
+                        case SyntaxKind.InvocationExpression:
+                        case SyntaxKind.Argument:
+                        case SyntaxKind.AddExpression: // String concatenation is not supported by other finders
+                            return false;
+                        default:
+                            current = current.Parent;
+                            break;
+                    }
+                }
+                return false;
+            }
         }
     }
 }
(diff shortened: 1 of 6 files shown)


pavel-mikula-sonarsource:
QG is failing due to know problem with C# vs. VB.NET code duplication

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
```diff
@@ -40,10 +40,8 @@ public sealed class DoNotHardcodeCredentials : DoNotHardcodeCredentialsBase<Synt
             PropertyAccessTracker = new CSharpPropertyAccessTracker(analyzerConfiguration, rule);
         }
 
-        protected override void Initialize(ParameterLoadingAnalysisContext context)
+        protected override void InitializeSyntax(ParameterLoadingAnalysisContext context)
```
costin-zaharia-sonarsource: I think we can rename `InitializeSyntax` with `RegisterSyntaxNodeActions` or something similar.

pavel-mikula-sonarsource: It's doing what `Initialize` method usually does. Following chain of `ParameterLoadingDiagnosticAnalyzer.Initialize` that prepares parameters and `DoNotHardcodeCredentialsBase.Initialize` that initializes trackers. I think InitializeSyntax is a good name here.

costin-zaharia-sonarsource: It's the word "syntax" which seems confusing to me. The method doesn't act on a syntax, it's registering  some callbacks on a context.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
```diff
@@ -93,5 +95,40 @@ private class AssignmentExpressionBannedWordsFinder : CredentialWordsFinderBase<
...
+            public StringLiteralBannedWordsFinder(DoNotHardcodeCredentialsBase<SyntaxKind> analyzer) : base(analyzer) { }
+
+            protected override string GetAssignedValue(LiteralExpressionSyntax syntaxNode) =>
+                syntaxNode.GetStringValue();
+
+            protected override string GetVariableName(LiteralExpressionSyntax syntaxNode) =>
+                null;
+
+            protected override bool IsAssignedWithStringLiteral(LiteralExpressionSyntax syntaxNode, SemanticModel semanticModel) =>
+                syntaxNode.IsKind(SyntaxKind.StringLiteralExpression) && !IsHandledByOtherFinder(syntaxNode.GetTopMostContainingMethod(), syntaxNode);
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: And this one know that he should not handle what the others were doing to avoid collisions.
What I don't like is that the evaluation is done inside function called `IsAssignedWithStringLiteral`, but I didn't want to add another method just for this case since this one is called just where it's needed.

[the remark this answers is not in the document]
costin-zaharia-sonarsource: > And this one know that he should not handle what the others were doing to avoid collisions.

This is exactly the problem I see. It shouldn't be it's responsibility to know what others are doing. They should be able to tell.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
```diff
@@ -93,5 +95,40 @@ private class AssignmentExpressionBannedWordsFinder : CredentialWordsFinderBase<
...
+                        case SyntaxKind.InvocationExpression:
+                        case SyntaxKind.Argument:
+                        case SyntaxKind.AddExpression: // String concatenation is not supported by other finders
+                            return false;
+                        default:
+                            current = current.Parent;
+                            break;
+                    }
+                }
+                return false;
```
andrei-epure-sonarsource: we're going to return `false` anyway... I understand the early return above, the problem is that it seems we won't raise issues only for `SyntaxKind.InvocationExpression`, `Argument` and `AddExpression`

do we know, then, all the cases where we will raise issues?

I would make this function differently: `ShouldHandle`
- returns `true` if it's one of the cases we know we want to handle
- `false` otherwise

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs
```diff
@@ -11,7 +11,7 @@ class Program
 
         private const string secret = "constantValue";
 
-        public void Test()
+        public void Test(string user)
         {
             string password = @"foo"; // Noncompliant {{"password" detected here, make sure this is not a hard-coded credential.}}
```
andrei-epure-sonarsource: do we have tests with 
```
var something = @"
foo
bar
password
";
```

or combinations with 

`var something = (a = "foo") + (b = "bar");`

and other weird cases like this?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
pavel-mikula-sonarsource: There are many different situations where string can appear in source code. Therefore can appear compiled inside assembly violating the S2068. We don't know list of all possible scenarios.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
```diff
@@ -93,5 +95,46 @@ private class AssignmentExpressionBannedWordsFinder : CredentialWordsFinderBase<
...
+
+        private class StringLiteralBannedWordsFinder : CredentialWordsFinderBase<LiteralExpressionSyntax>
+        {
+            public StringLiteralBannedWordsFinder(DoNotHardcodeCredentialsBase<SyntaxKind> analyzer) : base(analyzer) { }
+
+            protected override string GetAssignedValue(LiteralExpressionSyntax syntaxNode) =>
+                syntaxNode.GetStringValue();
+
+            protected override string GetVariableName(LiteralExpressionSyntax syntaxNode) =>
+                null;
```
andrei-epure-sonarsource: this seems like an abuse of the base class

does it mean it will trigger on 

```
foo("password");
```
?

pavel-mikula-sonarsource: There's a lot of logic needed from base class to unify behavior. Thanks to using this base class, the result will behave consistently (and not raise) for 
```C#
var x="password";
foo("password");
```

andreiepure: ah ok...
could you add a comment specifying why the Variable name is `null` and what this finder actually does?

do you have test cases with just "password"?
```
foo("password");
s += "password";

string property message => "Please enter password.";
foo("password=");
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs
```diff
@@ -22,6 +22,10 @@ public void Test(string user)
             string pwdPassword = "a"; // Noncompliant {{"pwd, password" detected here, make sure this is not a hard-coded credential.}}
 
             string foo2 = @"Password=123"; // Noncompliant
+            string multiline = // Noncompliant
```
andreiepure: you only added true positives. how about
```
string message = @"
Please enter your password below:
foo
bar
";
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs
```diff
@@ -33,6 +37,9 @@ public void Test(string user)
             foo = "foo=1password=1";
             foo = ""; // Compliant
 
+            var something1 = (foo = "foo") + (bar = "bar");
+            var something2 = (foo = "foo") + (bar = "password=123"); // Noncompliant
```
andreiepure:
```
var something2 = (foo = "foo") + (bar = "password"); // ?
```

andreiepure: var something2 = (foo = "foo") + (bar = "123=password"); // ?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
```diff
@@ -93,5 +95,47 @@ private class AssignmentExpressionBannedWordsFinder : CredentialWordsFinderBase<
                 syntaxNode.Left.IsKnownType(KnownType.System_String, semanticModel) &&
                 syntaxNode.Right.IsKind(SyntaxKind.StringLiteralExpression);
         }
+
+        private class StringLiteralBannedWordsFinder : CredentialWordsFinderBase<LiteralExpressionSyntax>
```
andrei-epure-sonarsource: could you add a high level comment about this class, as it's a special usage of the CredentialWordsFinderBase?
- it should clarify why `GetVariableName` returns `null`
- it should say that it looks inside strings for a specific pattern - what pattern?

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/DoNotHardcodeCredentials.cs
```diff
@@ -98,5 +100,41 @@ private class AssignmentExpressionBannedWordsFinder : CredentialWordsFinderBase<
...
+            protected override string GetAssignedValue(LiteralExpressionSyntax syntaxNode) =>
+                syntaxNode.GetStringValue();
+
+            protected override string GetVariableName(LiteralExpressionSyntax syntaxNode) =>
+                null;
+
+            protected override bool IsAssignedWithStringLiteral(LiteralExpressionSyntax syntaxNode, SemanticModel semanticModel) =>
+                syntaxNode.IsKind(SyntaxKind.StringLiteralExpression) && !IsHandledByOtherFinder(syntaxNode.GetTopMostContainingMethod(), syntaxNode);
+
+            private static bool IsHandledByOtherFinder(SyntaxNode method, SyntaxNode current)
```
costin-zaharia-sonarsource: we should rename also this one

pavel-mikula-sonarsource:
@andrei-epure-sonarsource something1,2,3 does not have a VB.NET equivalent, I considered that while adding new tests :)

PR #3206, opened 2020-03-12
Fix S4462 FP when using nameof()

### LINKED ISSUE #2794: S4462 False Positive
### Description

The statement below give a false positive of `S4462`

### Repro steps

Use somewhere in code: `nameof(Task<object>.Result)` 

### Expected behavior

S4462 should not create an warning.

### Actual behavior

A warning S4462 is created.

### Related information

* SonarC# Version: SonarAnalyzer.CSharp 8.0.0.9566
* Visual Studio Version VS2019 16.3.9

costin-zaharia-sonarsource:
Fixes #2794

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
@@ -146,7 +146,7 @@ namespace SonarAnalyzer.Helpers
             }
         }
 
-        public static bool IsInNameofCall(this ExpressionSyntax expression, SemanticModel semanticModel)
+        public static bool IsInNameOfArgument(this ExpressionSyntax expression, SemanticModel semanticModel)
         {
             var parentInvocation = expression.FirstAncestorOrSelf<InvocationExpressionSyntax>();
             return parentInvocation != null && parentInvocation.IsNameof(semanticModel);
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
@@ -146,14 +146,11 @@ namespace SonarAnalyzer.Helpers
             }
         }
 
-        public static bool IsInNameofCall(this ExpressionSyntax expression, SemanticModel semanticModel) =>
-            expression.Parent != null &&
-            expression.Parent.IsKind(SyntaxKind.Argument) &&
-            expression.Parent.Parent != null &&
-            expression.Parent.Parent.IsKind(SyntaxKind.ArgumentList) &&
-            expression.Parent.Parent.Parent != null &&
-            expression.Parent.Parent.Parent.IsKind(SyntaxKind.InvocationExpression) &&
-            ((InvocationExpressionSyntax)expression.Parent.Parent.Parent).IsNameof(semanticModel);
+        public static bool IsInNameofCall(this ExpressionSyntax expression, SemanticModel semanticModel)
+        {
+            var parentInvocation = expression.FirstAncestorOrSelf<InvocationExpressionSyntax>();
+            return parentInvocation != null && parentInvocation.IsNameof(semanticModel);
+        }
 
         public static bool IsNameof(this InvocationExpressionSyntax expression, SemanticModel semanticModel)
         {
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
@@ -152,17 +152,11 @@ namespace SonarAnalyzer.Helpers
             return parentInvocation != null && parentInvocation.IsNameof(semanticModel);
         }
 
-        public static bool IsNameof(this InvocationExpressionSyntax expression, SemanticModel semanticModel)
-        {
-            if (expression == null ||
-                !expression.Expression.IsKind(SyntaxKind.IdentifierName) ||
-                semanticModel.GetSymbolInfo(expression).Symbol?.Kind == SymbolKind.Method)
-            {
-                return false;
-            }
-
-            return ((IdentifierNameSyntax)expression.Expression).Identifier.ToString() == NameOfKeywordText;
-        }
+        public static bool IsNameof(this InvocationExpressionSyntax expression, SemanticModel semanticModel) =>
+            expression != null &&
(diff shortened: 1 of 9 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -98,5 +98,11 @@ public Task<string> Run(Task<string> task)
                 return completedTask.Result; // Noncompliant FP, task is already completed at this point.
             });
         }
+
```
pavel-mikula-sonarsource: Please add this line into `SleepAsync()` to be sure to handle it correctly for the other circumstances
`var name = nameof(Thread.Sleep);`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -146,14 +146,11 @@ private static bool IsOn(this ExpressionSyntax expression, SyntaxKind onKind)
...
 
-        public static bool IsInNameofCall(this ExpressionSyntax expression, SemanticModel semanticModel) =>
-            expression.Parent != null &&
-            expression.Parent.IsKind(SyntaxKind.Argument) &&
-            expression.Parent.Parent != null &&
-            expression.Parent.Parent.IsKind(SyntaxKind.ArgumentList) &&
-            expression.Parent.Parent.Parent != null &&
-            expression.Parent.Parent.Parent.IsKind(SyntaxKind.InvocationExpression) &&
-            ((InvocationExpressionSyntax)expression.Parent.Parent.Parent).IsNameof(semanticModel);
+        public static bool IsInNameofCall(this ExpressionSyntax expression, SemanticModel semanticModel)
```
pavel-mikula-sonarsource: It would be nice to add tests for helpers when we touch them, because they are covered only as secondary effect and it's not documented what they are intended to do.
Unfortunately for this file, I added `SyntaxHelperTest.cs` file in the JWT PR that is not merged yet, so it would be a hassle to synchronize/rebare&rerebase it 😒

pavel-mikula-sonarsource: I like the code reduction, but I'm not very happy with the old `IsInNameofCall` name.  I suggest to rename it to `IsNameofArgument`. 
WDYT?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -71,17 +71,13 @@ public sealed class CallToAsyncMethodShouldNotBeBlocking : SonarDiagnosticAnalyz
                 ["Sleep"] = new[] { "Thread.Sleep", "await Task.Delay" },
             };
 
-        protected override void Initialize(SonarAnalysisContext context)
-        {
-            context.RegisterSyntaxNodeActionInNonGenerated(
-                ReportOnViolation,
-                SyntaxKind.SimpleMemberAccessExpression);
-        }
+        protected override void Initialize(SonarAnalysisContext context) =>
```
pavel-mikula-sonarsource: Regarding `InvalidMemberAccess` vs `MemberNameToMessageArguments`, I don't like that there are two duplicate sets of initialization that hopes to contain same keys. 
Can you please either:
- sort the members to be in the same alphabetical order and put the declarations together (moving `TaskThreadPoolCalls` before or after)
or
- create some local Config class that would contain all information needed for single method name. And initialize it something like 
`["GetResult"] = new MethodConfig("Task.GetAwaiter.GetResult", "await", KnownType.System_Runtime_CompilerServices_TaskAwaiter, KnownType.System_Runtime_CompilerServices_TaskAwaiter_TResult)`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -100,6 +96,11 @@ private static void ReportOnViolation(SyntaxNodeAnalysisContext context)
                 return;
             }
 
+            if (simpleMemberAccess.IsInNameofCall(context.SemanticModel))
```
pavel-mikula-sonarsource: Can we move this before `possibleMemberAccesses` to have a tiny performance optimization? IsIsNameofCall is cheap, getting symbol info is more expensive.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
costin-zaharia-sonarsource: It's a good point. I will add some tests after both commits are merged to avoid the merge pain.

costin-zaharia-sonarsource: IsInNameOfArgument? since it can be just a part of the nameof() argument

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
costin-zaharia-sonarsource: Moved MemberNameToMessageArguments up and did the ordering but I would avoid spreading this commit even more. I suggest to start with baby steps.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
costin-zaharia-sonarsource: IsInNameofCall needs the semantic model to retrieve the symbols so it's actually the same thing

pavel-mikula-sonarsource: Maybe in another PR: IsNameof can be improved (changed to `=>` as single bool expression) and take semanticModel performance penalty AFTER `Identifier.ToString() == NameOfKeywordText` check. 

In this PR: Moving the condition upwards will prepare this rule for this small performance improvement.

costin-zaharia-sonarsource: I've changed the `IsNameof` method and moved the `IsInNameofCall` call up.

PR #3207, opened 2020-03-12
Fix S3263 FP: Static initialization is not detected in struct

### LINKED ISSUE #3187: Fix S3263 FP: Static initialization is not detected in struct
### Description
Static fields should appear in the order they must be initialized, but for array (and list) initialization, this is not detected correctly. Note, the noncompliant is already detected correctly, but might be a good test case when changing the behaviour of the rule.

### Repro steps
``` C#
public struct S3263
{
    public static readonly S3263 Zero;
    public static readonly S3263 V42 = new S3263(42);

    public static readonly IReadOnlyCollection<S3263> All = new[] { Zero, V42, }; // compliant, both values have been initialized.

    public static readonly IReadOnlyCollection<S3263> Extra = new[] { Zero, V007, V42, }; // noncompliant, V007 has not been initalized.

    public static readonly S3263 V007 = new S3263(7);

    public S3263(int i) => Value = i;

    public int Value { get; }
}
```

### Related information
* SonarC# 8.4

pavel-mikula-sonarsource:
Fixes #3187

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StaticFieldInitializerOrder.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StaticFieldInitializerOrder.cs
@@ -12,14 +12,17 @@ namespace Tests.TestCases
         public static string s2 = new string('x', Const2);
         public static string s3 = new string('x', Y); // Noncompliant
 
-        public static Action<int> A = (i) => { var x = i + StaticFieldInitializerOrder.Y; }; // Okay??? Might or might not. For no we don't report on it
+        public static Action<int> A = (i) => { var x = i + StaticFieldInitializerOrder.Y; }; // Okay??? Might or might not. For now we don't report on it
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StaticFieldInitializerOrder.cs
```diff
@@ -52,6 +52,31 @@ public partial interface IStaticFieldsOrder
         public const int Const2 = 5;
     }
 
+    public partial struct StaticFieldsOrderStruct
+    {
+        public static string s1 = new string('x', Const); // Compliant. Const do not suffer from initialization order fiasco.
+        public static string s2 = new string('x', Const2);
+        public static string s3 = new string('x', Y); // Noncompliant
+
+        public static Action<int> A = (i) => { var x = i + StaticFieldInitializerOrder.Y; }; // Okay??? Might or might not. For no we don't report on it
```
costin-zaharia-sonarsource: I think it would be nice to have coverage in test cases also with array initialization as Corniel mentioned.

PR #3209, opened 2020-03-12
Fix S4462 FP: Result in ContinueWith

### LINKED ISSUE #2413: False Positive on S4462 when using ContinueWith()
### Description
When using ` System.Threading.Tasks.Task<TResult>.ContinueWith()` there is no way to avoid the usage of `Task.Result`. The whole purpose of `ContinueWith()` is to act on the outcome of the `Task`.

### Repro steps
``` C#
using System;
using System.Threading;
using System.Threading.Tasks;

{
    public class S4462
    {
        public Task<S4462> Run(Func<Task<S4462>> continuation, CancellationToken token)
        {
            return continuation().ContinueWith((action) =>
            {
                return action.Result; // Compliant, ContinueWith() enforces to do so.
            });
        }
    }
}
```
### Expected behavior
I would expect that this usage is compliant. You could argue that you would have another rule that discourages the usage of  `ContinueWith()` completely, but that is a totally different debate.

costin-zaharia-sonarsource:
Fixes #2413


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -36,6 +36,8 @@ public sealed class CallToAsyncMethodShouldNotBeBlocking : SonarDiagnosticAnalyz
     {
         private const string DiagnosticId = "S4462";
         private const string MessageFormat = "Replace this use of '{0}' with '{1}'.";
+        private const string ResultName = "Result";
```
pavel-mikula-sonarsource: It would be nice to use constant for `SleepName` as well. It's also used in code separately.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -96,9 +98,10 @@ private static void ReportOnViolation(SyntaxNodeAnalysisContext context)
                 return;
             }
 
-            if (simpleMemberAccess.IsInNameofCall(context.SemanticModel))
+            if (IsResultInContinueWithCall(memberAccessNameName, simpleMemberAccess) ||
```
pavel-mikula-sonarsource: Can we move this above `possibleMemberAccesses` to gain little performance improvement? `GetSymbolInfo` is more expensive.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -122,6 +125,17 @@ private static void ReportOnViolation(SyntaxNodeAnalysisContext context)
...
         }
 
+        private static bool IsResultInContinueWithCall(string memberAccessName, SyntaxNode syntaxNode)
+        {
+            if (memberAccessName != ResultName)
+            {
+                return false;
+            }
+
+            var parentInvocation = syntaxNode.FirstAncestorOrSelf<InvocationExpressionSyntax>();
```
pavel-mikula-sonarsource: I'm afraid that this will stop on first wrapping Invocation and cause FNs like this
``` C#
return continuation().ContinueWith((action) =>
{
    var t = action.Result.Run(null, token); // This is now OK
    return t.Result; // This should still be noncompliant
});
```
or FPs when there would be wrapper like `CheckResult(action.Result)` inside the call

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -91,11 +91,12 @@ private Task SleepTask()
             Thread.Sleep(10);
         }
 
+        // See https://github.com/SonarSource/sonar-dotnet/issues/2413
         public Task<string> Run(Task<string> task)
         {
             return task.ContinueWith(completedTask =>
             {
-                return completedTask.Result; // Noncompliant FP, task is already completed at this point.
+                return completedTask.Result; // Compliant, task is already completed at this point.
```
pavel-mikula-sonarsource: We should fix or document FN like this:
``` C#
Action<Task<S4462>> arg = (action) =>
{
    var ret = action.Result;
};
continuation().ContinueWith(arg);
```

costin-zaharia-sonarsource: I've documented the FP.

costin-zaharia-sonarsource:
@pavel-mikula-sonarsource 
> Underlying issue seems to be more complicated to me here.

Indeed, but this is not a semantic execution rule. Judging by the number of false positives we removed from akka.net this is way better than before.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
costin-zaharia-sonarsource: That is a very strange case when a task has an action as a result.

pavel-mikula-sonarsource: My concern is about scenario, when `ContinueWith` calls any other async operation, retrieving it's task and calling that secondary `task.Result` => violating the rule.

pavel-mikula-sonarsource: The implementation will not work for wrapper inside call:
```C#
return task.ContinueWith(completedTask =>
{
    return string.Format("Result: {0}", completedTask.Result);
});
```

pavel-mikula-sonarsource: Better finding of `ContinueWith`, extracting it's argument name and validating it against the `variableName `that` .Result` was invoked on should do the trick. Or checking symbol type to be IParameterSymbol (or how is it called) with small performance penalty for symbol extraction

costin-zaharia-sonarsource:
> Thanx for fixing this.

Hi @Corniel, thank you for letting us know about the issue!

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
costin-zaharia-sonarsource: I understand now what you mean and the example makes sense. I've updated the code to look up until it finds a ContinueWith method with the right argument name. Shadowing will give FN but I think that's ok for now.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -117,6 +121,16 @@ private static void ReportOnViolation(SyntaxNodeAnalysisContext context)
                 messageArgs: MemberNameToMessageArguments[memberAccessNameName]));
         }
 
+        private static bool IsResultInContinueWithCall(string memberAccessName, MemberAccessExpressionSyntax memberAccess) =>
+            memberAccessName == ResultName &&
+            memberAccess.Expression is IdentifierNameSyntax identifierNameSyntax &&
+            memberAccess.FirstAncestorOrSelf<InvocationExpressionSyntax>(invocation => IsContinueWithCallWithArgumentName(invocation, identifierNameSyntax.GetName())) != null;
```
pavel-mikula-sonarsource: I suggest to extract result of `GetName()` to variable. Invocation chain like `A().B().C().D()` could create long ancestor chain to iterate.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -117,6 +121,16 @@ private static void ReportOnViolation(SyntaxNodeAnalysisContext context)
...
+
+        private static bool IsContinueWithCallWithArgumentName(InvocationExpressionSyntax invocation, string argumentName) =>
+            invocation.Expression.NameIs(ContinueWithName) &&
+            invocation.ArgumentList.Arguments.Any(argument => argument.Expression is SimpleLambdaExpressionSyntax lambda &&
```
pavel-mikula-sonarsource: We need to check `ParenthesizedLambdaExpression` as well. As I recently did this useless parenthesis:
``` C#
return task.ContinueWith((completedTask) => ...
```
Please add test as well.
It would be actually a nice C#-specific code quality rule idea to remove the unneeded parenthesis 😄

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -92,11 +93,30 @@ private Task SleepTask()
...
 
+        public void Run(Task<int> task)
+        {
+            Action<Task<int>> arg = (action) =>
+            {
+                var ret = action.Result; // Noncompliant FP, we do not track actions which are used on ContinueWith
+            };
+
+            task.ContinueWith(arg);
+        }
```
pavel-mikula-sonarsource: Please document this FP, `.Result` is safe here, but we don't track it:
```C#
var safeTask = Task.FromResult(42);
var a = safeTask.Result; // Noncompliant FP, we don't track source of the task

```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -92,11 +93,30 @@ private Task SleepTask()
...
+
+        // See https://github.com/SonarSource/sonar-dotnet/issues/2413
         public Task<string> Run(Task<string> task)
         {
             return task.ContinueWith(completedTask =>
             {
-                return completedTask.Result; // Noncompliant FP, task is already completed at this point.
+                return completedTask.Result; // Compliant, task is already completed at this point.
```
pavel-mikula-sonarsource: IsContinueWithCallWithArgumentName is great. Please add test to demonstrate it 🙂 
```C#
Task<int> anotherTask = null; // Pretend to compute something
var b = anotherTask.Result; // Noncompliant, this task is not safe inside ContinueWith
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
costin-zaharia-sonarsource: AFAICT GetName() is called on identifierNameSyntax and we don't have a long chain.

As it used in only one place I prefer to keep this as it is. It also keeps the code simpler.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
costin-zaharia-sonarsource: Good point. I will add a test for that. Also I think it would be a nice rule.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
pavel-mikula-sonarsource: Clarification of my previous statement from offline discussion:
If someone does `A(B(C(D(task.Result))))` then then `GetName` would be called in a loop instead of once.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -117,6 +121,25 @@ private static void ReportOnViolation(SyntaxNodeAnalysisContext context)
...
+            memberAccess.FirstAncestorOrSelf<InvocationExpressionSyntax>(invocation => IsContinueWithCallWithArgumentName(invocation, identifierName)) != null;
+
+        private static bool IsContinueWithCallWithArgumentName(InvocationExpressionSyntax invocation, string argumentName) =>
+            invocation.Expression.NameIs(ContinueWithName) &&
+            invocation.ArgumentList.Arguments.Any(argument => IsLambdaExpressionWithArgumentName(argument.Expression, argumentName));
+
+        private static bool IsLambdaExpressionWithArgumentName(ExpressionSyntax expression, string argumentName) =>
+            expression switch
+            {
+                SimpleLambdaExpressionSyntax lambda => lambda.Parameter.Identifier.ValueText == argumentName,
```
pavel-mikula-sonarsource: _Minor:_ I would rename `lambda` to `simpleLambda` for consistency.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -117,6 +121,25 @@ private static void ReportOnViolation(SyntaxNodeAnalysisContext context)
...
+                ParenthesizedLambdaExpressionSyntax parenthesizedLambda =>
```
pavel-mikula-sonarsource: I was wondering if the `Any` is really needed and it is. There are so many overloads of `ContinueWith`, I suggest to add some of them to tests. We already have overload for `Func<Task<T>, TResult>`.
Please add tests
```
// Action<Task<T>>
var a= task.ContinueWith(completedTask =>
{
    var result = completedTask.Result; // Compliant, task is already completed at this point.
});
// Action<Task<T>, object>
var b = task.ContinueWith((completedTask, state) =>
{
    var result = completedTask.Result; // Compliant, task is already completed at this point.
}, null);
// Func<Task<T>, object, TResult>
var c = task.ContinueWith((completedTask, state) =>
{
    return completedTask.Result; // Compliant, task is already completed at this point.
}, null);
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
costin-zaharia-sonarsource: In this case it will just make the line longer. We have all the context in the same line.

PR #3214, opened 2020-03-13
Display correct message when hardcoded password is detected

costin-zaharia-sonarsource:

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
@@ -41,7 +41,6 @@ namespace SonarAnalyzer.Rules
         private const string DefaultCredentialWords = "password, passwd, pwd, passphrase";
 
         protected readonly DiagnosticDescriptor rule;
-        protected readonly DiagnosticDescriptor hardCodedPasswordRule;
         private string credentialWords;
         private IEnumerable<string> splitCredentialWords;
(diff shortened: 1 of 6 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Hotspots/DoNotHardcodeCredentialsBase.cs
```diff
@@ -71,6 +73,7 @@ public string CredentialWords
         protected DoNotHardcodeCredentialsBase(System.Resources.ResourceManager rspecResources, IAnalyzerConfiguration analyzerConfiguration)
         {
             this.rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, rspecResources).WithNotConfigurable();
```
andrei-epure-sonarsource: is `rule` used anymore? the naming is confusing - I see this class only handles S2068, does it handle 2 different rules?

should you update the VB .NET rule too?

andrei-epure-sonarsource: ok... please use better names
`rule` and `hardCodedPasswordRule` both refer to "hard coded passsword rule"

please use `ruleForBaseClass` and `ruleForTrackers` if that's what they do

or some good names

costin-zaharia-sonarsource: In the end I've made the trackers configurable and we don't need the second rule. It seems to be a better solution.

PR #3216, opened 2020-03-13
Add test with previous FP to avoid regressions

costin-zaharia-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs
```diff
@@ -8,6 +8,7 @@ namespace Tests.Diagnostics
     class Program
     {
         public const string DBConnectionString = "Server=localhost; Database=Test; User=SA; Password=Secret123";    // Noncompliant
+        public const string EditPasswordPageUrlToken = "{Account.EditPassword.PageUrl}"; // Compliant
```
costin-zaharia-sonarsource: The FN was previously fixed but I've added a test case to avoid regressions.

andrei-epure-sonarsource: I guess it's FP

costin-zaharia-sonarsource: FP indeed 😄

costin-zaharia-sonarsource: at least the PR title is correct this time 😄

costin-zaharia-sonarsource: and now I see that you've change it

PR #3217, opened 2020-03-13
Using $(NugetPackageRoot) instead of $(NUGET_PACKAGES)

Corniel:
The short version is that you can not rely on $(NUGET_PACKAGES) to be configured, but you can on $(NugetPackageRoot) . By changing this (for building with ProtoBuf) i could compile (again) locally.

See: https://github.com/NuGet/Home/issues/6301#issuecomment-350939781


Corniel:
@andrei-epure-sonarsource can you look at this?

andrei-epure-sonarsource:
hey, @Corniel , thanks for this. 

indeed currently , as [it is specified in the readme](https://github.com/SonarSource/sonar-dotnet/blob/master/docs/contributing-analyzer.md#developing-with-visual-studio-2019), we need to manually set two env vars for the build + integration tests to work

image

in the thread you gave me I see [this comment](https://github.com/NuGet/Home/issues/6301#issuecomment-403856377)  and [another one](https://github.com/NuGet/Home/issues/6301#issuecomment-454711301) saying it's not very consistent.

Have you had problems with that in your experience?

Other than that, using `$(NugetPackageRoot)` is definitely a better idea

Corniel:
Without is does not work locally. (Unless I'll set the path). So yes.

andrei-epure-sonarsource:
close this in favor of #3226

PR #3218, opened 2020-03-13
Fix FP on S3717 with throw expressions

### LINKED ISSUE #2503: Rule S3717: NotImplementedException not recognized lambda notation
### Description
Throw of `NotImplementedException` is not recognized when the lambda expression syntax is used.

### Repro steps
``` C#
public class S3717
{
    public void NotImplemented() => throw new NotImplementedException(); // not-compliant
}
```
### Related information

* SonarC# Version: 7.15.0.8572

Corniel:
Fixes #2503 

Due to the use of an old version of Microsoft.CodeAnalysis.Charp 1.3.2 (backwards compatebility I asume), it was a bit less straight forward.

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TrackNotImplementedException.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TrackNotImplementedException.cs
@@ -56,19 +56,12 @@ namespace SonarAnalyzer.Rules.CSharp
                 SyntaxKind.ThrowStatement);
 
             context.RegisterSyntaxNodeActionInNonGenerated(
-                c =>
-                {
-                    var creationExpression = (ObjectCreationExpressionSyntax)c.Node;
-                    var throwExpression = creationExpression.Parent;
-
-                    if (throwExpression.Kind() != SyntaxKindEx.ThrowExpression)
-                    {
-                        return;
-                    }
-
-                    ReportDiagnostic(c, creationExpression, throwExpression);
-                },
-                SyntaxKind.ObjectCreationExpression);
+                 c =>
+                 {
+                     var throwExpression = (ThrowExpressionSyntaxWrapper)c.Node;
+                     ReportDiagnostic(c, throwExpression.Expression, throwExpression);
+                 },
+                 SyntaxKindEx.ThrowExpression);
         }
 
         private static void ReportDiagnostic(SyntaxNodeAnalysisContext c, ExpressionSyntax newExceptionExpression, SyntaxNode throwExpression)
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TrackNotImplementedException.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TrackNotImplementedException.cs
@@ -25,6 +25,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.ShimLayer.CSharp;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
@@ -32,14 +33,6 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class TrackNotImplementedException : SonarDiagnosticAnalyzer
     {
-        /// <remarks>
-        /// The ThrowExpressionSyntax and corresponding type, are not avialable
-        /// in the used version of Microsoft.CodeAnalysis.Charp 1.3.2.
-        ///
-        /// See: https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.syntax.throwexpressionsyntax?view=roslyn-dotnet
-        /// </remarks>
-        private const SyntaxKind ThrowExpression = (SyntaxKind)9052;
-
         internal const string DiagnosticId = "S3717";
         private const string MessageFormat = "Implement this method or throw 'NotSupportedException' instead.";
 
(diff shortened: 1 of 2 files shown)


Corniel:
@andrei-epure-sonarsource can you look at this?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TrackNotImplementedException.cs
```diff
@@ -32,6 +32,14 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class TrackNotImplementedException : SonarDiagnosticAnalyzer
     {
+        /// <remarks>
+        /// The ThrowExpressionSyntax and corresponding type, are not avialable
+        /// in the used version of Microsoft.CodeAnalysis.Charp 1.3.2.
+        ///
+        /// See: https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.syntax.throwexpressionsyntax?view=roslyn-dotnet
+        /// </remarks>
+        private const SyntaxKind ThrowExpression = (SyntaxKind)9052;
```
andrei-epure-sonarsource: Please use [SyntaxKindEx.ThrowExpression](https://github.com/SonarSource/sonar-dotnet/blob/8.5.0.15942/sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/ShimLayer/SyntaxKindEx.cs#L47) - we use the `ShimLayer` which is borrowed from StyleCopAnalyzers to support C#7 + syntaxes

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TrackNotImplementedException.cs
```diff
@@ -56,6 +64,20 @@ protected override void Initialize(SonarAnalysisContext context)
...
                 },
                 SyntaxKind.ThrowStatement);
+        
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
+                {
+                    var creationStatement = (ObjectCreationExpressionSyntax)c.Node;
+                    var throwStatement = creationStatement.Parent;
+
+                    if (throwStatement.Kind() == ThrowExpression &&
```
andrei-epure-sonarsource: you can use 
```
throwStatement.IsKind(SyntaxKindEx.ThrowExpression)
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TrackNotImplementedException.cs
```diff
@@ -56,6 +64,20 @@ protected override void Initialize(SonarAnalysisContext context)
                     }
                 },
                 SyntaxKind.ThrowStatement);
```
andrei-epure-sonarsource: you can add here a second SyntaxKind which is `SyntaxKindEx.ThrowExpression` , and have a method to extract what you need and then avoid duplication for checking the type against the KnownType

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TrackNotImplementedException.cs
andrei-epure-sonarsource: well... it's an expression and not a statement

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TrackNotImplementedException.cs
Corniel: I was not aware, and like the suggestion.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TrackNotImplementedException.cs
```diff
@@ -50,12 +51,32 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                        return;
                     }
+
+                    ReportDiagnostic(c, creationExpression, throwExpression);
                 },
-                SyntaxKind.ThrowStatement);
+                SyntaxKind.ObjectCreationExpression);
+        }
+
+        private static void ReportDiagnostic(SyntaxNodeAnalysisContext c, ExpressionSyntax expression, SyntaxNode syntax)
```
andrei-epure-sonarsource: I suggest to improve the names of the parameters:
- s/expression/newExceptionExpression
- s/syntax/throwExpression

(even if `throwExpression` can actually be a statement, this name is better than `syntax`)

andrei-epure-sonarsource:
we will merge this PR after implementing https://github.com/SonarSource/sonar-dotnet/issues/3231 to be able to trigger QA via comments on external PRs

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TrackNotImplementedException.cs
```diff
@@ -50,12 +51,32 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                    if (throwExpression.Kind() != SyntaxKindEx.ThrowExpression)
                     {
-                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, throwStatement.GetLocation()));
+                        return;
                     }
+
+                    ReportDiagnostic(c, creationExpression, throwExpression);
                 },
-                SyntaxKind.ThrowStatement);
+                SyntaxKind.ObjectCreationExpression);
```
andrei-epure-sonarsource: coming back on this one, @Corniel - why do you have to enter this callback on `ObjectCreationExpressionSyntax` and not on `SyntaxKindEx.ThrowExpression` ?

it looks cumbersome to always enter this callback for all object creation expression, then look at the `Parent` - is the TypeInfo only at the `creationExpression` level and not at the `throwExpression` level?

even so, it looks more intuitive to enter the callback only for `SyntaxKindEx.ThrowExpression` and then go inside it and get the `ObjectCreationExpression` from it to get the type..

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TrackNotImplementedException.cs
```diff
@@ -29,8 +29,10 @@ void FooBar2()
 //          ^^^^^^^^^
         }
 
-        void NotImplemented() =>
-            throw new NotImplementedException(); // FN
+        void ByExpression() => throw new NotImplementedException(); // Noncompliant
```
andrei-epure-sonarsource: also, please also add some more UTs with throw expressions , not only from arrow methods

e.g.
```
public bool Foo(int i)
{
  return i == 0 ?? throw new NotImplementedException();
}
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TrackNotImplementedException.cs
Corniel: @andrei-epure-sonarsource Because the type you get back is a type that is not in our version. And the base class that is has no Expression. I considered using `dynamic`, but that is never used yet, and then I had to add an dependency on MS.CSharp. Also using reflection felt like the wrong way to go.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TrackNotImplementedException.cs
Corniel: Added two test cases.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TrackNotImplementedException.cs
```diff
@@ -32,6 +32,15 @@ void FooBar2()
         void ByExpression() => throw new NotImplementedException(); // Noncompliant
 //                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+        int ConditionalEpxression(int i)
```
andrei-epure-sonarsource:
```suggestion
        int ConditionalExpression(int i)
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TrackNotImplementedException.cs
andrei-epure-sonarsource: tests are green with this version
```
                c =>
                {
                    var throwExpression = (ThrowExpressionSyntaxWrapper) c.Node;
                    ReportDiagnostic(c, throwExpression.Expression, throwExpression);
                },
                SyntaxKindEx.ThrowExpression);

```

and then I think you can simplify a lot the code , have a single callback

Corniel: Ah, you already created some nice reusable reflection based code.

andrei-epure-sonarsource: it's the Lightup layer in StyleCop.Analyzers - https://github.com/DotNetAnalyzers/StyleCopAnalyzers/tree/master/StyleCop.Analyzers/StyleCop.Analyzers/Lightup

we contributed to it, but credits go mainly to the StyleCop maintainers

PR #3227, opened 2020-03-17
Fix S3937 FP and other improvements on number pattern checks

### LINKED ISSUE #3165: FP S3937: Hexadecimal patterns are not handled properly
### Description
For hexadecimal defined numbers (int and long), groups of 3 are okay, but 2 and 4 are not. I think that is not what should be the case.

### Expected behavior
``` C#
public static class S3937
{
    public static readonly long Hexa2 = 0xFF_FF_FF_FF; // compliant
    public static readonly long Hexa3 = 0x121_1F3_456; // non-compliant, hexadecimals should have groups of 2 or 4 
    public static readonly long Hexa4 = 0xF_FFFF_FFFF; // compliant
}
```
### Already okay
The binary defined work like I expected.
``` C#
public static class S3937
{
    public static readonly long Bin2 = 0b1_10_11_10_11; // compliant
    public static readonly long Bin3 = 0b0_111_111_111; // non-compliant, binaries should have groups of 2 or 4 
    public static readonly long Bin4 = 0b111_1101_1011; // compliant
}
```
* SonarC# Version 8.3

Corniel:
Fixes #3165

Take binary and hexadecimal prefixes into account.

All groups should have the same size with these exceptions:
* first group before the dot is allowed to be shorter.
* last group after the dot is allowed to be shorter.
* It is allowed to only have groups before the dot.
* the first group after the dot (if multiple) should be one shorter than the group size

Specially the last one was missing.
``` C#
var amout = 1_123_123.23_123_12m; 
```

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
@@ -90,8 +90,8 @@ namespace SonarAnalyzer.Rules.CSharp
                     return true;
                 }
 
-                // all should have the same size, except the first.
-                if (groups.Skip(1).Any(g => g != size))
+                // all should have the same size, except the first, and the size should at least be 2.
+                if (size < 2 || groups.Skip(1).Any(g => g != size))
                 {
                     return true;
                 }
@@ -108,9 +108,6 @@ namespace SonarAnalyzer.Rules.CSharp
             {
                 return false;
             }
-
-            // The first decimal group should be one smaller.
-            decimals[0]++;
 
             size = size == NotFound ? decimals[0] : size;
 
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
@@ -69,15 +69,15 @@ namespace SonarAnalyzer.Rules.CSharp
         /// <remarks>internal for test purposes.</remarks>
         internal static bool HasIrregularPattern(string numericToken)
         {
-            var splitted = StripNumericPreAndSuffix(numericToken).Split(Dot);
+            var split = StripNumericPreAndSuffix(numericToken).Split(Dot);
 
             // ignore multiple dots.
-            if (splitted.Length > 2)
+            if (split.Length > 2)
             {
                 return false;
             }
 
-            var groups = splitted[0].Split(Underscore).Select(s => s.Length).ToArray();
+            var groups = split[0].Split(Underscore).Select(s => s.Length).ToArray();
 
             var size = NotFound;
 
@@ -98,12 +98,12 @@ namespace SonarAnalyzer.Rules.CSharp
                 }
             }
 
-            if (splitted.Length == 1)
+            if (split.Length == 1)
             {
                 return false;
             }
 
-            var decimals = splitted[1].Split(Underscore).Select(s => s.Length).ToArray();
+            var decimals = split[1].Split(Underscore).Select(s => s.Length).ToArray();
 
             if (decimals.Length == 1)
             {
@@ -120,29 +120,35 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static string StripNumericPreAndSuffix(string numericToken)
         {
-            var start = 0;
             var length = numericToken.Length;
 
-            // hexadecimal en binary prefixes
+            // hexadecimal and binary prefixes (0xFFFF_23_AB, 0b1110_1101)
             if (numericToken.StartsWith("0x", StringComparison.InvariantCultureIgnoreCase) ||
                 numericToken.StartsWith("0b", StringComparison.InvariantCultureIgnoreCase))
             {
-                start = 2;
-                length -= 2;
-            }
-            // UL and LU suffix.
-            else if (numericToken.EndsWith("UL", StringComparison.OrdinalIgnoreCase) ||
-                numericToken.EndsWith("LU", StringComparison.OrdinalIgnoreCase))
-            {
-                length -= 2;
-            }
-            // single suffixes
-            else if ("LDFUMldfum".IndexOf(numericToken[numericToken.Length - 1]) != NotFound)
-            {
-                length--;
+                return numericToken.Substring(2, length - 2);
             }
 
-            return numericToken.Substring(start, length);
+            // Scientific notation (1.23E8)
+            var exponentMarker = numericToken.IndexOf("E", StringComparison.InvariantCultureIgnoreCase);
+            if (exponentMarker != NotFound)
+            {
+                return numericToken.Substring(0, exponentMarker);
+            }
+
+            // UL and LU suffix.
+            if (numericToken.EndsWith("UL", StringComparison.OrdinalIgnoreCase) ||
+                numericToken.EndsWith("LU", StringComparison.OrdinalIgnoreCase))
+            {
+                return numericToken.Substring(0, length - 2);
+            }
+            // single suffixes
+            if ("LDFUMldfum".IndexOf(numericToken[numericToken.Length - 1]) != NotFound)
+            {
+                return numericToken.Substring(0, length - 1);
+            }
+
+            return numericToken;
         }
     }
 }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
@@ -66,27 +66,9 @@ namespace SonarAnalyzer.Rules.CSharp
                 SyntaxKind.NumericLiteralExpression);
         }
 
-        public static bool HasIrregularPattern(string numericToken)
+        internal static bool HasIrregularPattern(string numericToken)
         {
-            // hexadecimal en binary prefixes
-            var start = numericToken.StartsWith("0x", StringComparison.InvariantCultureIgnoreCase)
-                || numericToken.StartsWith("0b", StringComparison.InvariantCultureIgnoreCase) ? 2 : 0;
-
-            var length = numericToken.Length - start;
-
-            // UL suffix.
-            if (numericToken.EndsWith("UL", StringComparison.OrdinalIgnoreCase))
-            {
-                length -= 2;
-            }
-            // single suffixes
-            else if ("LDFUMldfum".IndexOf(numericToken[numericToken.Length - 1]) != NotFound)
-            {
-                length--;
-            }
-
-            var stripped = numericToken.Substring(start, length);
-            var splitted = stripped.Split(Dot);
+            var splitted = StripNumericPreAndSuffix(numericToken).Split(Dot);
 
             // ignore multiple dots.
             if (splitted.Length > 2)
@@ -132,19 +114,31 @@ namespace SonarAnalyzer.Rules.CSharp
 
             size = size == NotFound ? decimals[0] : size;
 
-            // the last should not be bigger.
-            if (decimals.Last() > size)
-            {
-                return true;
-            }
-
+            // the last should not be bigger
             // all should have the same size, except the last.
-            if (decimals.Take(decimals.Length - 1).Any(g => g != size))
-            {
-                return true;
-            }
+            return (decimals.Last() > size)
+                || (decimals.Take(decimals.Length - 1).Any(g => g != size));
+        }
 
-            return false;
+        private static string StripNumericPreAndSuffix(string numericToken)
+        {
+            // hexadecimal en binary prefixes
+            var start = numericToken.StartsWith("0x", StringComparison.InvariantCultureIgnoreCase)
+                || numericToken.StartsWith("0b", StringComparison.InvariantCultureIgnoreCase) ? 2 : 0;
+
+            var length = numericToken.Length - start;
+
+            // UL suffix.
+            if (numericToken.EndsWith("UL", StringComparison.OrdinalIgnoreCase))
+            {
+                length -= 2;
+            }
+            // single suffixes
+            else if ("LDFUMldfum".IndexOf(numericToken[numericToken.Length - 1]) != NotFound)
+            {
+                length--;
+            }
+            return numericToken.Substring(start, length);
         }
     }
 }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
@@ -137,10 +137,12 @@ namespace SonarAnalyzer.Rules.CSharp
                 return false;
             }
 
-            // the second group is assumed to be leading.
+            // the first group is allowed to contain less digits that the other ones,
+            // so take the expected length from the second group.
             var groupLength = groupLengths[1];
 
-            // first should not be bigger, and the size per group should be at least 2.
+            // we consider groups of 1 digit irregular.
+            // first should not be bigger.
             if (groupLength < 2 || groupLengths[0] > groupLength)
             {
                 return true;
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
@@ -66,6 +66,7 @@ namespace SonarAnalyzer.Rules.CSharp
                 SyntaxKind.NumericLiteralExpression);
         }
 
+        /// <remarks>internal for test purposes.</remarks>
         internal static bool HasIrregularPattern(string numericToken)
         {
             var splitted = StripNumericPreAndSuffix(numericToken).Split(Dot);
@@ -119,14 +120,18 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static string StripNumericPreAndSuffix(string numericToken)
         {
+            var start = 0;
+            var length = numericToken.Length;
+
             // hexadecimal en binary prefixes
-            var start = numericToken.StartsWith("0x", StringComparison.InvariantCultureIgnoreCase)
-                || numericToken.StartsWith("0b", StringComparison.InvariantCultureIgnoreCase) ? 2 : 0;
-
-            var length = numericToken.Length - start;
-
+            if (numericToken.StartsWith("0x", StringComparison.InvariantCultureIgnoreCase) ||
+                numericToken.StartsWith("0b", StringComparison.InvariantCultureIgnoreCase))
+            {
+                start = 2;
+                length -= 2;
+            }
             // UL suffix.
-            if (numericToken.EndsWith("UL", StringComparison.OrdinalIgnoreCase))
+            else if (numericToken.EndsWith("UL", StringComparison.OrdinalIgnoreCase))
             {
                 length -= 2;
             }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
@@ -77,45 +77,24 @@ namespace SonarAnalyzer.Rules.CSharp
                 return false;
             }
 
-            var groups = split[0].Split(Underscore).Select(s => s.Length).ToArray();
+            var groupLengthsLeftFromDot = split[0].Split(Underscore).Select(g => g.Length).ToArray();
 
-            var size = NotFound;
-
-            if (groups.Length > 1)
+            if (HasIrregularGroupLengths(groupLengthsLeftFromDot))
             {
-                size = groups[1];
-
-                // first should not be bigger.
-                if (groups[0] > size)
-                {
-                    return true;
-                }
-
-                // all should have the same size, except the first, and the size should at least be 2.
-                if (size < 2 || groups.Skip(1).Any(g => g != size))
-                {
-                    return true;
-                }
+                return true;
             }
 
+            // no dot, so done.
             if (split.Length == 1)
             {
                 return false;
             }
 
-            var decimals = split[1].Split(Underscore).Select(s => s.Length).ToArray();
+            // reverse, as for right from the dot, the last (instead of the first)
+            // group length is allowed to be shorter than the group length.
+            var groupLengthsRightFromDot = split[1].Split(Underscore).Select(g => g.Length).Reverse().ToArray();
 
-            if (decimals.Length == 1)
-            {
-                return false;
-            }
-
-            size = size == NotFound ? decimals[0] : size;
-
-            // the last should not be bigger
-            // all should have the same size, except the last.
-            return (decimals.Last() > size)
-                || (decimals.Take(decimals.Length - 1).Any(g => g != size));
+            return HasIrregularGroupLengths(groupLengthsRightFromDot);
         }
 
         private static string StripNumericPreAndSuffix(string numericToken)
@@ -149,6 +128,25 @@ namespace SonarAnalyzer.Rules.CSharp
             }
 
             return numericToken;
+        }
+
+        private static bool HasIrregularGroupLengths(int[] groupLengths)
+        {
+            if (groupLengths.Length < 2)
+            {
+                return false;
+            }
+
+            // the second group is assumed to be leading.
+            var groupLength = groupLengths[1];
+
+            // first should not be bigger, and the size per group should be at least 2.
+            if (groupLength < 2 || groupLengths[0] > groupLength)
+            {
+                return true;
+            }
+
+            return groupLengths.Skip(1).Any(l => l != groupLength);
         }
     }
 }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
@@ -130,8 +130,9 @@ namespace SonarAnalyzer.Rules.CSharp
                 start = 2;
                 length -= 2;
             }
-            // UL suffix.
-            else if (numericToken.EndsWith("UL", StringComparison.OrdinalIgnoreCase))
+            // UL and LU suffix.
+            else if (numericToken.EndsWith("UL", StringComparison.OrdinalIgnoreCase) ||
+                numericToken.EndsWith("LU", StringComparison.OrdinalIgnoreCase))
             {
                 length -= 2;
             }
@@ -140,6 +141,7 @@ namespace SonarAnalyzer.Rules.CSharp
             {
                 length--;
             }
+
             return numericToken.Substring(start, length);
         }
     }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
@@ -38,12 +38,13 @@ namespace SonarAnalyzer.Rules.CSharp
         internal const string DiagnosticId = "S3937";
         private const string MessageFormat = "Review this number; its irregular pattern indicates an error.";
 
+        private const char Underscore = '_';
+        private const char Dot = '.';
+        private const int NotFound = -1;
+
         private static readonly DiagnosticDescriptor rule =
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
-
-        private static readonly ImmutableHashSet<char> NumericTypeSuffix =
-            ImmutableHashSet.Create('L', 'D', 'F', 'U', 'M');
 
         protected override void Initialize(SonarAnalysisContext context)
         {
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,93 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            // UL suffix.
+            if (numericToken.EndsWith("UL", StringComparison.OrdinalIgnoreCase))
+            {
+                length -= 2;
+            }
+            // single suffixes
+            else if ("LDFUMldfum".IndexOf(numericToken[numericToken.Length - 1]) != NotFound)
+            {
+                length--;
+            }
```
christophe-zurn-sonarsource: This part does not solve the issue that there will be FP for hexadecimal numbers finishing with `D` or `F`.
Ex:
```
var a = 0xFF_FF_1F; // FP
var a = 0xFF_FF_1D; // FP
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,93 @@ protected override void Initialize(SonarAnalysisContext context)
...
+
+            var decimals = splitted[1].Split(Underscore).Select(s => s.Length).ToArray();
+
+            if (decimals.Length == 1)
+            {
+                return false;
+            }
+
+            // The first decimal group should be one smaller.
+            decimals[0]++;
```
christophe-zurn-sonarsource: Why this change in the rule behaviour? I'm not saying this is necessarily wrong, but if we change that, it first should be updated in the rule description

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,93 @@ protected override void Initialize(SonarAnalysisContext context)
...
-            if (NumericTypeSuffix.Contains(char.ToUpperInvariant(numberLiteral[numberLiteral.Length - 1])))
+            // all should have the same size, except the last.
+            if (decimals.Take(decimals.Length - 1).Any(g => g != size))
             {
-                return numberLiteral.Substring(0, numberLiteral.Length - 1);
+                return true;
             }
 
-            return numberLiteral;
+            return false;
```
christophe-zurn-sonarsource: this part can be simplified with:
```
return decimals.Last() > size || decimals.Take(decimals.Length - 1).Any(g => g != size);
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,93 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                    if (HasIrregularPattern(literal.Token.Text))
                     {
                         c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, literal.GetLocation()));
                     }
                 },
                 SyntaxKind.NumericLiteralExpression);
         }
 
-        private static string ClearNumberTypeSuffix(string numberLiteral)
+        public static bool HasIrregularPattern(string numericToken)
```
christophe-zurn-sonarsource: I feel this method is getting quite big. I would have kept a separate utility method to retrieve the actual content of the numeric token, to have that logic separated from the pattern check

christophe-zurn-sonarsource: Method visibility could be kept `internal` as we only want to access it from outside for testing purposes

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
Corniel: Well, given the discussion I was under the impression that we where not going for groups of 3 for decimals and groups of 2, 4, or 8 for hex and binary. I can live with both, but there were at least some in consistencies (and bugs), that had to be fixed. If there be found agreement on the allowed group sizes, its is easy to add.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
Corniel: Shorter !- simpler.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
Corniel: I think is the only right implementation. Given this changes, some descriptions have to be updated anyway.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
Corniel: If you want to do something with the decimal/hexa/binary info, that is becoming less optimal.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
christophe-zurn-sonarsource: Yes, I still need to discuss further with another colleague regarding the rule but for now let's assume we don't want a change in the behaviour, as this could be done in a separate PR if needed.

The false positive in the rule I'm talking about is not directly linked to the number of elements of such groups: it is related to the fact that hexedecimal numbers finishing with a `F` or `D` do not behave as expected.

Current behaviour is the following:
```
var a = 0xFF_FF_F1; // Ok - no issue raised, groups of 2
var b = 0xFFF_FFF_FF1; // Ok - no issue raised, groups of 3
var c = 0xFFFF_FFFF_FFF1; // Ok - no issue raised, groups of 4

var d = 0xFF_FF_FF; // KO - issue raised but should not - last group is thought to be of size 1 as the last 'F' of the group is considered as float number suffix - FP
var e = 0xFFF_FFF_FFD; // KO - issue raised but should not - last group is thought to be of size 2 as the last 'D' of the group is considered as double number suffix - FP
var f = 0xFFFF_FFFF_FFFF; // KO - issue raised but should not - FP

var g = 0xFF_FF_FFF; // No issue raised, but it should - grouping is not consistant
var h = 0xFFF_FFF_FFFD; // No issue raised, but it should - grouping is not consistant
var i = 0xFFFF_FFFF_FFFFF; // No issue raised, but it should - grouping is not consistant
```
This is the most important problem with the rule at the moment

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
christophe-zurn-sonarsource: Could you direct me to some documentation regarding this ? I cannot find anything on the grouping of the fractional part to justify this change

Corniel: @christophe-zurn-sonarsource No, I can't. :(

I'm pretty sure I've been educated that way, but I can't find any source that says I'm right.

it is probably only true if the group size is 3 anyway. What I remember is that the first decimal are 1/10, then 1/100, so it makes sense to add a _ between 1/100 and 1/1000, just like between 1000 and 100.

But I agree, that without proof, we should not go for it.

The only thing I found (that also slightly disagrees) is this:
https://math.stackexchange.com/questions/182775/convention-of-digit-grouping-after-decimal-point/182776#182776

What suggests that the group size of your decimal part can be different (although it still should be consequent).

andrei-epure-sonarsource:
tests are failing on this one (see [here](https://sonarsource.visualstudio.com/DotNetTeam%20Project/_build/results?buildId=16459&view=logs&j=22b521bd-d457-5294-620e-7d91dbfb4c56&t=74d1794f-b183-5917-cd98-3eaf49a73ed6&l=1981)), you should rebase on master

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
Corniel: I've changed the behaviour. Also I added that the group size should be at least, two, as I think that makes sense. ;)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
Corniel: @christophe-zurn-sonarsource 

The false positives have been fixed, and a minium group size of 2 is also taken into account.

andrei-epure-sonarsource:
as long as the build is broken, we cannot move forward, @Corniel

Corniel:
@ andrei-epure-sonarsource Any clue what is going wrong? maybe a new rebase?

andrei-epure-sonarsource:
not sure @Corniel , I think one of us mentioned a failing test similar to what's failing on this PR.

we're working on a different project this sprint so most likely we'll investigate it a bit later on

enjoy the long weekend!

Corniel:
@andrei-epure-sonarsource Same to you guys!

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/NumberPatternShouldBeRegularTest.cs
```diff
@@ -46,5 +46,37 @@ public void NumberPatternShouldBeRegular_FromCSharp7()
...
+
+        [DataTestMethod]
+        [TestCategory("Rule")]
+        [DataRow(".123_123_123_1", "No group before the dot")]
+        [DataRow("123", "No group character")]
+        [DataRow("1_123_123UL", "With UL suffix")]
+        [DataRow("1_123_123L", "With L suffix")]
+        [DataRow("1.1.1", "Two dots")]
+        [DataRow("0b1010_1010", "With binary prefix")]
+        [DataRow("0xFF_FF_12", "With hexadecimal prefix")]
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: `0xFFFF_FFFF_FFFF` is considered irregular because you strip the `F` suffix

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/NumberPatternShouldBeRegularTest.cs
```diff
@@ -46,5 +46,37 @@ public void NumberPatternShouldBeRegular_FromCSharp7()
...
         }
+
+        [DataTestMethod]
+        [TestCategory("Rule")]
+        [DataRow("1_1_1", "Group size of 1")]
+        [DataRow("123_12", "First group is bigger then the second")]
+        [DataRow("1234_123_123", "First group is bigger then the others")]
+        [DataRow("123_123_12_123", "Another group has a different size")]
+        [DataRow("1_123.123_1234", "Last decimal group is bigger")]
+        [DataRow(".123_123_1234", "No group before the dot")]
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: `0xFF_FF_FFF` is considered regular because you strip the `F` suffix

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,84 @@ protected override void Initialize(SonarAnalysisContext context)
...
+
+            var length = numericToken.Length - start;
+
+            // UL suffix.
+            if (numericToken.EndsWith("UL", StringComparison.OrdinalIgnoreCase))
+            {
+                length -= 2;
+            }
+            // single suffixes
+            else if ("LDFUMldfum".IndexOf(numericToken[numericToken.Length - 1]) != NotFound)
```
andrei-epure-sonarsource, line 134: you should only strip the `F` and `D` suffix for non-binary, non-hexa numbers as they are hexa "digits"

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,84 @@ protected override void Initialize(SonarAnalysisContext context)
...
+        private static string StripNumericPreAndSuffix(string numericToken)
+        {
+            // hexadecimal en binary prefixes
+            var start = numericToken.StartsWith("0x", StringComparison.InvariantCultureIgnoreCase)
+                || numericToken.StartsWith("0b", StringComparison.InvariantCultureIgnoreCase) ? 2 : 0;
+
+            var length = numericToken.Length - start;
+
+            // UL suffix.
+            if (numericToken.EndsWith("UL", StringComparison.OrdinalIgnoreCase))
```
andrei-epure-sonarsource, line 129: it can also end with "LU" ([docs](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types))

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/NumberPatternShouldBeRegularTest.cs
```diff
@@ -46,5 +46,37 @@ public void NumberPatternShouldBeRegular_FromCSharp7()
...
+        public void HasIrregularPattern(string numericToken, string message)
+        {
+            Assert.IsTrue(NumberPatternShouldBeRegular.HasIrregularPattern(numericToken), message);
+        }
+
+        [DataTestMethod]
+        [TestCategory("Rule")]
+        [DataRow(".123_123_123_1", "No group before the dot")]
+        [DataRow("123", "No group character")]
+        [DataRow("1_123_123UL", "With UL suffix")]
```
andrei-epure-sonarsource, line 67: also add with "LU" suffix as it's legal, also to test case insensitivity
```
1_123_123LU
1_123_123lU
1_123_123Lu
1_1234_1234lu
1_12_13ul
```
and with [floating point suffixes](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types) (the below numbers are from the MS docs)

```
1.0
3D
0.42e2
3_000.5F
134.45E-2f
3_000.5m
1.5E6m
2.1m
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,84 @@ protected override void Initialize(SonarAnalysisContext context)
...
                     }
                 },
                 SyntaxKind.NumericLiteralExpression);
         }
 
-        private static string ClearNumberTypeSuffix(string numberLiteral)
+        internal static bool HasIrregularPattern(string numericToken)
         {
-            if (numberLiteral.EndsWith("UL", StringComparison.OrdinalIgnoreCase))
+            var splitted = StripNumericPreAndSuffix(numericToken).Split(Dot);
```
andrei-epure-sonarsource, line 71: please use `split` (the participle and past tense forms are the same as the present form)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,84 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            }
+
+            if (splitted.Length == 1)
             {
-                return numberLiteral.Substring(0, numberLiteral.Length - 2);
+                return false;
             }
 
-            if (NumericTypeSuffix.Contains(char.ToUpperInvariant(numberLiteral[numberLiteral.Length - 1])))
+            var decimals = splitted[1].Split(Underscore).Select(s => s.Length).ToArray();
```
andrei-epure-sonarsource, line 105: do you mean `digits`?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,84 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            if (decimals.Length == 1)
             {
-                return numberLiteral.Substring(0, numberLiteral.Length - 1);
+                return false;
             }
 
-            return numberLiteral;
+            size = size == NotFound ? decimals[0] : size;
+
+            // the last should not be bigger
```
andrei-epure-sonarsource, line 114: by "last" do you mean first? you mean that `0xF_FF_FF` is ok even if the first group has a different number of digits, right?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
andrei-epure-sonarsource: also, how do we treat 'e' - what's after 'e' should be like a suffix, right?

```
1.5E6m
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NumberPatternShouldBeRegular.cs
```diff
@@ -46,7 +46,7 @@ public void NonCompliantCases()
             int million = 1_000_00_000;  // Noncompliant {{Review this number; its irregular pattern indicates an error.}}
 //                        ^^^^^^^^^^^^
 
-            var x = 1_234.56_78; // Noncompliant
```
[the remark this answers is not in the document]
Corniel: 1_234.56_78 According to the sources I found, there is no reason why the decimals should have the same group size, as the-non-decimals. So it should be compliant.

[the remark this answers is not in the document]
andrei-epure-sonarsource: ok, sounds good to me, so please leave the test as `// Compliant` 😄 

the main point here is _please don't remove an existing test case_

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/NumberPatternShouldBeRegularTest.cs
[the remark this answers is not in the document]
Corniel: You're right. Fixed!

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
Corniel: I Fully agree. Will add that too!

Corniel: Hmm, seconds thoughts:
1.234_567E2 Is obviously valid
but:
1.234_56E2
Is that wrong? should it be 0.123.456E3

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NumberPatternShouldBeRegular.cs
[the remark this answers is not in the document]
Corniel: @andrei-epure-sonarsource IK get your point, but the thing is: It was compliant now, and therefore also on a different place. I changed that test to become non-compliant and left that at the same spot. So what do you suggest?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
Corniel: yep

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NumberPatternShouldBeRegular.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: my apologies, I missed the `double groupsizes = 1_234.56_78;`

andrei-epure-sonarsource:
there's a [failed UT](https://sonarsource.visualstudio.com/DotNetTeam%20Project/_build/results?buildId=18110&view=logs&j=23c99647-d7e1-5b83-b5fb-f7e97f8a9ce7&t=3ae69f40-e780-58ec-7b2f-ab41fb46316e&l=49) 

>   X NumberPatternShouldBeRegular_FromCSharp7 [29ms]
  Error Message:
   Test method SonarAnalyzer.UnitTest.Rules.NumberPatternShouldBeRegularTest.NumberPatternShouldBeRegular_FromCSharp7 threw exception: 
SonarAnalyzer.UnitTest.TestFramework.UnexpectedDiagnosticException: Unexpected primary issue on line 15, span (14,32)-(14,43) with message 'Review this number; its irregular pattern indicates an error.'.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,97 @@ protected override void Initialize(SonarAnalysisContext context)
...
+        {
+            var split = StripNumericPreAndSuffix(numericToken).Split(Dot);
+
+            // ignore multiple dots.
+            if (split.Length > 2)
+            {
+                return false;
+            }
+
+            var groups = split[0].Split(Underscore).Select(s => s.Length).ToArray();
```
andrei-epure-sonarsource, line 80: I suggest to rename to `groupLengths`
- this isn't "groups of digits", but the size of each groups
- it doesn't refer to all groups, but only to the ones on the left of the 'dot' (if the 'dot' is present) - *please add a comment above to mention something like* `// we are interested only in what is on the left side of the 'Dot', if present`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,97 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                size = groups[1];
+
+                // first should not be bigger.
+                if (groups[0] > size)
+                {
+                    return true;
+                }
+
+                // all should have the same size, except the first, and the size should at least be 2.
+                if (size < 2 || groups.Skip(1).Any(g => g != size))
```
andrei-epure-sonarsource, line 95: instead of `g` maybe use `l` (for `lenght`)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,97 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                    return true;
+                }
+            }
+
+            if (split.Length == 1)
+            {
+                return false;
+            }
+
+            var decimals = split[1].Split(Underscore).Select(s => s.Length).ToArray();
```
andrei-epure-sonarsource, line 106: a [decimal](https://dictionary.cambridge.org/dictionary/english/decimal) is 
> relating to or expressed in a system of counting based on the number ten

please rename to [digit ](https://en.wiktionary.org/wiki/digit#Noun), as it can be in hexa as well

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,97 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                }
+
+                // all should have the same size, except the first, and the size should at least be 2.
+                if (size < 2 || groups.Skip(1).Any(g => g != size))
+                {
+                    return true;
+                }
+            }
+
+            if (split.Length == 1)
```
andrei-epure-sonarsource, line 101: please do this in the same `if` with `split.Length > 2` , immediately after the split

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -57,42 +58,97 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            {
+                return false;
+            }
+
+            var decimals = split[1].Split(Underscore).Select(s => s.Length).ToArray();
+
+            if (decimals.Length == 1)
```
andrei-epure-sonarsource, line 108: I'me really confused here - haven't you done this check already, if the group has 1 digit?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/NumberPatternShouldBeRegularTest.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: can you also add `0xFFFF_FFFF_FFFF` please as a regular format?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
andrei-epure-sonarsource: also, as above - this if `digitGroupSizes` rather than the digits themselves

andrei-epure-sonarsource: and again this is a bit weird - why are you splitting based on Undescore here and at line 80 as well?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
Corniel: Changed this code to make the intent clear. Thanks for the indebt code-reviews. It has been a long time since my code has been checked that well!

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
Corniel: Done.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -150,5 +129,24 @@ private static string StripNumericPreAndSuffix(string numericToken)
...
+        {
+            if (groupLengths.Length < 2)
+            {
+                return false;
+            }
+
+            // the second group is assumed to be leading.
+            var groupLength = groupLengths[1];
+
+            // first should not be bigger, and the size per group should be at least 2.
```
andrei-epure-sonarsource, line 143: this comment is a bit harder to read because you say "first should not be bigger" but that's the second expression in the `OR` expression

```suggestion
            // we consider groups of 1 digit irregular
            // the first group should not be bigger than the rest
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
```diff
@@ -150,5 +129,24 @@ private static string StripNumericPreAndSuffix(string numericToken)
...
         }
+
+        private static bool HasIrregularGroupLengths(int[] groupLengths)
+        {
+            if (groupLengths.Length < 2)
+            {
+                return false;
+            }
+
+            // the second group is assumed to be leading.
```
andrei-epure-sonarsource, line 140: I don't understand what "leading" means?

```suggestion
            // the first group is allowed to contain less digits that the other ones, so take the "regular" length from the second group
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NumberPatternShouldBeRegular.cs
```diff
@@ -47,7 +47,7 @@ public void NonCompliantCases()
             int million = 1_000_00_000;  // Noncompliant {{Review this number; its irregular pattern indicates an error.}}
 //                        ^^^^^^^^^^^^
 
-            var x = 1_234.56_78_1; // Noncompliant
+            var x = 1_234.56_78_123; // Noncompliant
```
andrei-epure-sonarsource, line 51: do we have any tests with `____`? it would be nice to have something like `0xFF___FF___FF` - I believe it's Noncompliant due to line 144 in `NumberPatternShouldBeRegular.cs`

and I think it's fine to be noncompliant (even if the pattern is regular, but using multiple underscores should usually be a mistake)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/NumberPatternShouldBeRegular.cs
andrei-epure-sonarsource: I'm glad you appreciate the comments and thanks for the patience :)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/NumberPatternShouldBeRegular.cs
andrei-epure-sonarsource: and also with `0xFF________FF___FF`

this program runs:
```
var x = 0xFF________FF___FF;
Console.WriteLine(x);
```

Corniel: Added the tests, and took the comment suggestions.

PR #3229, opened 2020-03-18
Add branch coverage support for OpenCover

### LINKED ISSUE #1606: OpenCover branchCoverage is measured but SonarQube "Uncovered Conditions" is missing
### Description

In my current setup code coverage appears to be properly measured and submitted to SonarQube, as it gets properly displayed. However, although my "opencover.xml" includes branchCoverage numbers, the SonarQube server is not showing anything for branch coverage. For overall file statistics "Uncovered Conditions" is a dash (–) and for new code "Conditions to Cover on New Code" and "Uncovered Conditions on New Code" are both 0.

### Repro steps

Followed the official documentation for setting up my build analysis. I'm using a PowerShell script with roughly the following relevant commands:

`
   $SonarQubeExePath begin /k:my_proj /n:my_proj /v:"$NewAssemblyFileVersion" /d:sonar.scm.provider=hg /d:sonar.cs.opencover.reportsPaths="opencover.xml" /d:sonar.cs.vstest.reportsPaths="TestResults/*.trx" /d:sonar.exclusions="Content/**/*.css,Scripts/**/*.js,Scripts/**/*.ts,GlobalSuppressions.cs" /d:sonar.cpd.exclusions="Content/**/*.css,Scripts/**/*.js,Scripts/**/*.ts,GlobalSuppressions.cs,Properties/AssemblyInfo.cs,App_Start/RouteConfig.cs"
`
and
`
$OpenCoverExePath -output:opencover.xml -register:user -target:"$VSTestExePath" -targetargs:"/Logger:trx $TestDLL /TestAdapterPath:$TestDLLPath" -filter:'+[*]* -[Proj1Name]* -[Proj2Name]* -[Proj3Name]*'
`

At the end of the analysis I see the following:

```
INFO: Parsing the OpenCover report C:\src\my_proj\.\opencover.xml
INFO: Adding this code coverage report to the cache for later reuse: C:\src\my_proj\.\opencover.xml
INFO: Sensor C# Tests Coverage Report Import [csharp] (done) | time=304ms
```

### Expected behavior

Branch coverage numbers are not missing or zero.

### Actual behavior

Branch coverage numbers are missing or zero.

### Known workarounds

N/A

### Related information

* SonarC# Version: 7.1 (build 5212)
* Scanner for MSBuild Version: 4.2.0.1214
* SonarQube version: 6.7.1.35068
* Usage: Standalone from PowerShell script

costin-zaharia-sonarsource:
Fixes #1606

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchCoverage.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchCoverage.java
@@ -20,6 +20,8 @@
 
 package org.sonar.plugins.dotnet.tests;
 
+import java.util.Objects;
+
 // This class is responsible to keep SQ/SC metrics related to branch coverage
 class BranchCoverage {
   private int line;
@@ -42,5 +44,32 @@ class BranchCoverage {
 
   public int getCoveredConditions() {
     return coveredConditions;
+  }
+
+  public void add(int conditions, int coveredConditions){
+    this.conditions += conditions;
+    this.coveredConditions += coveredConditions;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    BranchCoverage other = (BranchCoverage) o;
+
+    return line == other.line &&
+      conditions == other.conditions &&
+      coveredConditions == other.coveredConditions;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(line, conditions, coveredConditions);
   }
 }
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchCoverage.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchCoverage.java
@@ -20,12 +20,13 @@
 
 package org.sonar.plugins.dotnet.tests;
 
-class LineBranchCoverage {
+// This class is responsible to keep SQ/SC metrics related to branch coverage
+class BranchCoverage {
   private int line;
   private int conditions;
   private int coveredConditions;
 
-  public LineBranchCoverage(int line, int conditions, int coveredConditions){
+  public BranchCoverage(int line, int conditions, int coveredConditions){
     this.line = line;
     this.conditions = conditions;
     this.coveredConditions = coveredConditions;
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/Coverage.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/Coverage.java
@@ -19,8 +19,10 @@
  */
 package org.sonar.plugins.dotnet.tests;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -30,6 +32,7 @@ public class Coverage {
   private static final int GROW_FACTOR = 2;
   private static final int SPECIAL_HITS_NON_EXECUTABLE = -1;
   private final Map<String, int[]> hitsByLineAndFile = new HashMap<>();
+  private final Map<String, List<LineBranchCoverage>> branchCoverageByFile = new HashMap<>();
 
   void addHits(String file, int line, int hits) {
     int[] oldHitsByLine = hitsByLineAndFile.get(file);
@@ -57,6 +60,11 @@ public class Coverage {
     oldHitsByLine[i] += hits;
   }
 
+  public void addLineBranchCoverage(String file, LineBranchCoverage lineBranchCoverage){
+    List<LineBranchCoverage> linesBranchCoverage = branchCoverageByFile.computeIfAbsent(file, k -> new ArrayList<>());
+    linesBranchCoverage.add(lineBranchCoverage);
+  }
+
   public Set<String> files() {
     return hitsByLineAndFile.keySet();
   }
@@ -77,7 +85,16 @@ public class Coverage {
     return result;
   }
 
+  List<LineBranchCoverage> getLinesBranchCoverage(String file){
+    return branchCoverageByFile.getOrDefault(file, new ArrayList<>());
+  }
+
   void mergeWith(Coverage otherCoverage) {
+    mergeLineHits(otherCoverage);
+    mergeBranchCoverage(otherCoverage);
+  }
+
+  private void mergeLineHits(Coverage otherCoverage){
     Map<String, int[]> other = otherCoverage.hitsByLineAndFile;
 
     for (Map.Entry<String, int[]> entry : other.entrySet()) {
@@ -90,4 +107,15 @@ public class Coverage {
     }
   }
 
+  private void mergeBranchCoverage(Coverage otherCoverage){
+    Map<String, List<LineBranchCoverage>> otherBranchCoverageByFile = otherCoverage.branchCoverageByFile;
+
+    for (Map.Entry<String, List<LineBranchCoverage>> entry: otherBranchCoverageByFile.entrySet()){
+      String file = entry.getKey();
+
+      for (LineBranchCoverage lineBranchCoverage: entry.getValue()) {
+        addLineBranchCoverage(file, lineBranchCoverage);
+      }
+    }
+  }
 }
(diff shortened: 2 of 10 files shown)


sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
```diff
@@ -108,6 +108,29 @@ public void valid() throws Exception {
...
+
+    assertThat(coverage.files()).containsOnly(filePath);
+    assertThat(coverage.hits(filePath))
+      .hasSize(3)
+      .contains(
+        Assertions.entry(6, 1),
+        Assertions.entry(7, 1),
+        Assertions.entry(14, 1));
+
+    assertThat(coverage.getLinesBranchCoverage(filePath).size()).isEqualTo(1);
```
andrei-epure-sonarsource: could you also check the branch coverage at line 2433 in the test XML?
- you'll be able to verify that all get parsed
- you'll be able to verify the first, middle and last entry to make sure they get read correctly

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/CoverageReportImportSensor.java
```diff
@@ -157,6 +140,32 @@ void analyze(SensorContext context, Coverage coverage) {
...
+      newCoverage.lineHits(entry.getKey(), entry.getValue());
+    }
+
+    for (LineBranchCoverage lineBranchCoverage : coverage.getLinesBranchCoverage(filePath)){
+      fileHasCoverage = true;
+      newCoverage.conditions(lineBranchCoverage.getLine(), lineBranchCoverage.getConditions(), lineBranchCoverage.getCoveredConditions());
+    }
+    newCoverage.save();
+
+    if (fileHasCoverage) {
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: In my opinion this will make the code more complicated since we introduce a second boolean but we will not take distinct decisions based on them.

sonar-dotnet-shared-library/src/test/resources/opencover/coverage_branches.xml
```diff
@@ -1,4 +1,49 @@
...
+            Assert.That(sut.Compute(0, 0, 0).Equals(1));
+        }
+    }
+}
+
+OpenCover version 4.7.922.0
+
+Commands used:
+msbuild BranchCoveragePoc.sln
+OpenCover.Console.exe -target:"c:\Program Files\dotnet\dotnet.exe" -targetargs:"test" -output:coverage.xml -register:user -filter:"+[BranchCoveragePoc*]* -[BranchCoveragePoc.Tests]*" -oldStyle
```
andrei-epure-sonarsource: regarding `-oldStyle` - is this necessary?
if so, after doing the release we must update https://community.sonarsource.com/t/coverage-test-data-generate-reports-for-c-vb-net/9871

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
```diff
@@ -123,12 +123,12 @@ public void branchCoverage() throws Exception {
         Assertions.entry(7, 1),
         Assertions.entry(14, 1));
 
-    assertThat(coverage.getLinesBranchCoverage(filePath).size()).isEqualTo(1);
+    assertThat(coverage.getBranchCoverage(filePath).size()).isEqualTo(1);
```
andrei-epure-sonarsource: as discussed offline, I suggest to add a more complex test with more than 1 branch coverage entry

sonar-dotnet-shared-library/src/test/resources/opencover/coverage_branches.xml
andrei-epure-sonarsource: and this should be taken into consideration when doing the tests with coverlet

costin-zaharia-sonarsource: No, this is not needed and it's should not affect us. I removed it since it can be misleading.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
costin-zaharia-sonarsource: I've added a few more cases and with this I've found and fixed a corner case when multiple branch points are present at the same line

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/Coverage.java
```diff
@@ -57,6 +61,22 @@ void addHits(String file, int line, int hits) {
...
   }
 
+  public void addBranchCoverage(String file, BranchCoverage branchCoverage){
+    List<BranchCoverage> branchCoverages = branchCoverageByFile.computeIfAbsent(file, k -> new ArrayList<>());
+
+    // If there are multiple branch coverage entries per line these need to be merged; otherwise SQ/SC will display only
+    // one of them.
+    Optional<BranchCoverage> existingCoverage = branchCoverages.stream()
+      .filter(coverage -> coverage.getLine() == branchCoverage.getLine())
+      .findAny();
```
andrei-epure-sonarsource: this `findAny` means you get one result ... but from what I understand, you can have more than one already, right?

or you should not because here you are always merging?

sonar-dotnet-shared-library/src/test/resources/opencover/coverage_branches.xml
```diff
@@ -0,0 +1,2745 @@
...
+    public class Calculator
+    {
+        public int Compute(int x, int y, int z)
+        {
+            if (x == 0 || y < 0)
+                Console.WriteLine(x);
+
+            _ = y == 0 || z == 0;
+            _ = x == y || y == z;
+            _ = y == 0 || z == 0; _ = x == y || y == z;
```
andrei-epure-sonarsource: sorry to be annoying - does it make sense to add a line with 3 conditions per line, to make sure that case is treated ok as well?

andrei-epure-sonarsource: I mean 3 different entries...

maybe it doesn't make sense... I'm not sure

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/Coverage.java
costin-zaharia-sonarsource: There can be only one due to merging.

sonar-dotnet-shared-library/src/test/resources/opencover/coverage_branches.xml
costin-zaharia-sonarsource: Sure, I will add a test case with this in the Coverage.java since there it's easier to maintain.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchCoverage.java
```diff
@@ -0,0 +1,75 @@
...
+    this.coveredConditions += coveredConditions;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+
+    if (o == null || getClass() != o.getClass()) {
```
andrei-epure-sonarsource: also, would you like to add a UT for this class to ensure full code coverage? [(see sonarcloud)](https://sonarcloud.io/component_measures?id=org.sonarsource.dotnet%3Asonar-csharp&metric=new_uncovered_conditions&pullRequest=3229&selected=org.sonarsource.dotnet%3Asonar-csharp%3Asonar-dotnet-shared-library%2Fsrc%2Fmain%2Fjava%2Forg%2Fsonar%2Fplugins%2Fdotnet%2Ftests%2FBranchCoverage.java&view=list)

I think it's a low hanging fruit

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/BranchCoverageTest.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  @Test
+  public void givenEqualBranchCoverageEqualsReturnsTrue(){
+    assertThat(new BranchCoverage(3, 2, 1).equals(new BranchCoverage(3, 2, 1))).isTrue();
+  }
+
+  @Test
+  public void givenLineConditionsAndCoveredConditionsHashCodeConsidersAll(){
+    assertThat(new BranchCoverage(3, 2, 1).hashCode()).isEqualTo(32737);
```
andrei-epure-sonarsource: I think it's better to compare with 
```
Objects.hash(3, 2, 1);
```

because you're now actually testing the output of this `hash` function to be `32737`

andrei-epure-sonarsource: that may change with a future version of the JDK, dunno

costin-zaharia-sonarsource: I also thought about this and I wasn't sure what the best approach is. 

I usually don't like compute results in the unit tests, replicating the code, but, in this case, we just want to be sure that `Objects.hash` was called with the right parameters so I guess it's fine.

Issue #3231, opened 2020-03-18
Update Azure Pipelines to better support external contributions


andrei-epure-sonarsource:
Currently, there is a bug in the burgr notification script which prevents CI from running on external PRs. See [this job](https://sonarsource.visualstudio.com/DotNetTeam%20Project/_build/results?buildId=16014&view=logs&j=bc173ff5-2a20-5490-94eb-dabbe703399b&t=a8226cc3-8c0d-5f42-f8db-be2b38113537) failing with `400 event not found` - this is because instead of passing the PR number, we pass the branch name, which doesn't exist in the repo (the branch is on an external fork).

Anyway, we don't want CI to run automatically on external PRs (secrets could be exposed from a malicious PR; bots could open PRs and DOS the available VMs; etc).

The plan is to:

- [x] 1. disable CI for PRs coming from forks - [docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#validate-contributions-from-forks)

- [x] 2. SonarSourcers will be able to manually trigger the CI via a comment on the PR, after we have reviewed the PR and made sure no vulnerabilities are present - [docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#comment-triggers)

andrei-epure-sonarsource:
I am closing this because we can now trigger our CI on PRs from forks, with the comment
```
/AzurePipelines run Sonar.Net
```

PR #3234, opened 2020-03-18
CI - dump environment variables in separate stage

andrei-epure-sonarsource:
This is a step to better understand the [Azure Pipelines predefined variables](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml).

It takes 30 seconds to dump them and it's done in parallel.

related to #3231

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,41 +21,9 @@ resources:
     - repository: commonTemplates
       type: git
       name: pipelines-yaml-templates
-      ref:  refs/tags/v1.0.6
+      ref:  refs/heads/andrei/dump
 
 stages:
-- stage: 'var_dump'
-  displayName: Print environment
-  jobs:
-    - job: printPredefinedEnvVar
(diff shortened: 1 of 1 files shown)


costin-zaharia-sonarsource:
> It takes 30 seconds to dump them and it's done in parallel.

Any reason not to have this as a stage inside the `dotnetBuildjob`? 

Spawning a new node for the variable dump is not ideal since we have a limited number of nodes and this will delay other concurrent builds.

andrei-epure-sonarsource:
@costin-zaharia-sonarsource because the dotnet job is using the template from the common jobs. So if something fails in the common template stages (which are run at the beginning, before anything we run), the dump won't appear.

andrei-epure-sonarsource:
@costin-zaharia-sonarsource but thinking about it better, this logging should be in the common template.

not everything there is `server` jobs, so I will try that and come back here after

PR #3236, opened 2020-03-19
S3444 Add missing name in the message for indexers

### LINKED ISSUE #3225: S3444 Add missing name in the message for indexers
### Description

I am getting a warning S3444 on my compound interface `IDynamicList<T>` that implements `IList<T>` and `IReadOnlyList<T>`. The message is not actionable, the member name is missing:

> Rename or add member '' to this interface to resolve ambiguities.

### Repro steps

```csharp
public interface IDynamicList<T> : IList<T>, IReadOnlyList<T>
{
}
```

### Expected behavior

The message should always provide a name of the conflicting interface member. I believe it is the indexer in this case.

### Actual behavior

The conflicting interface member name is blank in the message. I believe it is the indexer. Adding a new indexer stops the warning:

```csharp
public interface IDynamicList<T> : IList<T>, IReadOnlyList<T>
{
    new T this[int index] { get; set; }
}
```

### Related information

* SonarAnalyzer.CSharp version 8.5.0.15942
* VS2019 version 16.4.6
* MSBuild version 16.4.0, dotnet SDK version 3.1.102
* Windows 10, build 1909

[Rule S3444 description](https://jira.sonarsource.com/browse/SLVS-673)

pavel-mikula-sonarsource:
Fixes #3225

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InheritedCollidingInterfaceMembers.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InheritedCollidingInterfaceMembers.cs
@@ -131,14 +131,15 @@ namespace SonarAnalyzer.Rules.CSharp
             var names = collidingMembers
                 .Take(MaxMemberDisplayCount)
                 .Select(member => GetMemberDisplayName(member, spanStart, semanticModel))
+                .Distinct()
                 .ToList();
 
-            if (collidingMembers.Count == 1)
+            if (names.Count == 1)
             {
                 return names[0];
             }
 
-            if (collidingMembers.Count == 2)
+            if (names.Count == 2)
             {
                 return $"{names[0]} and {names[1]}";
             }
@@ -156,6 +157,12 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static string GetMemberDisplayName(IMethodSymbol method, int spanStart, SemanticModel semanticModel)
         {
+            if(method.AssociatedSymbol is IPropertySymbol property && property.IsIndexer)
+            {
+                var text = property.ToMinimalDisplayString(semanticModel, spanStart, SymbolDisplayFormat.CSharpShortErrorMessageFormat);
+                return $"'{text}'";
+            }
+
             var parts = method.ToMinimalDisplayParts(semanticModel, spanStart, SymbolDisplayFormat.CSharpShortErrorMessageFormat)
                 .SkipWhile(part => !PartKindsToStartWith.Contains(part.Kind))
                 .ToList();
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InheritedCollidingInterfaceMembers.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InheritedCollidingInterfaceMembers.cs
@@ -125,7 +125,7 @@ namespace SonarAnalyzer.Rules.CSharp
             return methodSymbol2.MethodKind != MethodKind.EventRemove;
         }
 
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InheritedCollidingInterfaceMembers.cs
```diff
@@ -131,14 +131,15 @@ private static bool IsNotEventRemoveAccessor(IMethodSymbol methodSymbol2)
             var names = collidingMembers
```
costin-zaharia-sonarsource: I think it makes sense to address also the sonar source code smell and use IEnumerable: https://sonarcloud.io/project/issues?id=sonaranalyzer-dotnet&issues=AXDyLJsTbpzdS9P4et1k&open=AXDyLJsTbpzdS9P4et1k&pullRequest=3236

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InheritedCollidingInterfaceMembers.cs
```diff
@@ -131,14 +131,15 @@ private static bool IsNotEventRemoveAccessor(IMethodSymbol methodSymbol2)
...
                 .Take(MaxMemberDisplayCount)
                 .Select(member => GetMemberDisplayName(member, spanStart, semanticModel))
+                .Distinct()
                 .ToList();
 
-            if (collidingMembers.Count == 1)
```
costin-zaharia-sonarsource: I think a switch would be easier to read (we have already 3 cases):
```
switch (names.Count)
{
    case 1:
        return names[0];

    case 2:
        return $"{names[0]} and {names[1]}";

    default:
        names.Add("...");
        return string.Join(", ", names);
}
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InheritedCollidingInterfaceMembers.AfterCSharp8.cs
```diff
@@ -214,7 +214,7 @@ public interface IIndexer1
 
     public interface IIndexer2
     {
-        public string this[int i] // Rule is raised but the message is not clear
+        public string this[int i] // Indexer does not have a name, signature is used instead
         {
             get { return ""; } // Noncomliant
```
costin-zaharia-sonarsource: I know this is old but that's a very interesting way to hide an issue 😄. (typo `Noncompliant`)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InheritedCollidingInterfaceMembers.cs
```diff
@@ -84,4 +84,10 @@ public interface IMultilayer : IReadOnlyDictionary<string, object>, IDummy // Co
     public interface IDummy
     {
     }
+
+    // https://github.com/SonarSource/sonar-dotnet/issues/3225
+    // Issue is raised correctly, but message does not contain name, because indexers doesn't have an identifier
```
costin-zaharia-sonarsource: I would remove this comment (// Issue is raised correctly, but message does not contain name, because indexers doesn't have an identifier) it is no longer valid and does not help.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InheritedCollidingInterfaceMembers.AfterCSharp8.cs
pavel-mikula-sonarsource: We should consider it as design-pattern 😄
Issue should be raised on line 226, so I'll just remove it.

PR #3238, opened 2020-03-19
Simple pipeline for external checks

andrei-epure-sonarsource:
related to #3231

### DIFF
--- a/azure-pipelines-external.yml
+++ b/azure-pipelines-external.yml
@@ -1,8 +1,3 @@
-pr:
-  branches:
-    include:
-    - '*'  # must quote since "*" is a YAML reserved character; we want a string
-
 pool:
   vmImage: 'windows-latest'
 
--- a/azure-pipelines-external.yml
+++ b/azure-pipelines-external.yml
@@ -1,6 +1,8 @@
 trigger: none
 pr:
-  - master
+  branches:
+    include:
+    - '*'  # must quote since "*" is a YAML reserved character; we want a string
 
 pool:
   vmImage: 'windows-latest'
--- a/azure-pipelines-external.yml
+++ b/azure-pipelines-external.yml
@@ -1,9 +1,3 @@
-trigger: none
-pr:
-  branches:
-    include:
-    - '*'  # must quote since "*" is a YAML reserved character; we want a string
-
 pool:
   vmImage: 'windows-latest'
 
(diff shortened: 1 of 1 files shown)


azure-pipelines-external.yml
```diff
@@ -0,0 +1,150 @@
...
+          organization: 'sonarsource'
+          scannerMode: 'MSBuild'
+          projectKey: 'sonaranalyzer-dotnet'
+          projectName: 'SonarAnalyzer .Net'
+          projectVersion: '$(SONAR_PROJECT_VERSION)'
+          extraProperties: |
+            sonar.analysis.buildNumber=$(Build.BuildId)
+            sonar.analysis.pipeline=$(Build.BuildId)
+            sonar.analysis.sha1=$(Build.SourceVersion)
+            sonar.analysis.repository=$(Build.Repository.ID)
```
costin-zaharia-sonarsource: I think that we can remove this since the pipeline will be triggered only for PRs.

azure-pipelines-external.yml
```diff
@@ -0,0 +1,150 @@
...
+
+variables:
+  - group: sonar-dotnet-variables
+  - group: sonarsource-build-variables
+
+stages:
+- stage: buildAndTests
+  displayName: Build and run tests
+  jobs:
+    - job: runUnitTestsJob
```
costin-zaharia-sonarsource: This will run only .Net build, unit tests and integration tests. I don't see a reason for someone from community to change the plugin but we need to make this visible for the hole team that we won't accept changes for the java code or at least the changes are not validated.

andrei-epure-sonarsource: the idea is to run the full job as well before merging (w/o running the full job, we won't be able to merge anyway as Java ITs are required)

azure-pipelines-external.yml
```diff
@@ -0,0 +1,150 @@
...
+
+      - task: SonarCloudAnalyze@1
+        displayName: 'Run Code Analysis - End Step'
+
+      - task: SonarCloudPublish@1
+        displayName: 'Run Code Analysis - Publish QG'
+        inputs:
+          pollingTimeoutSec: '300'
+
+    - job: runIntegrationTestsJob
```
costin-zaharia-sonarsource: In this case we have just 2 jobs which need to run in a specific order since `runIntegrationTestsJob` depends on `runUnitTestsJob`. I think we should refactor this to have only one job and remove the tasks PublishPipelineArtifact and DownloadPipelineArtifact. 

I think that overall it will speed up the build and make things simpler.

azure-pipelines-external.yml
costin-zaharia-sonarsource: Yes, that's true.

andrei-epure-sonarsource:
it doesn't work on fork PRs... https://github.com/SonarSource/sonar-dotnet/pull/3237

PR #3241, opened 2020-03-20
Compute branch coverage with OpenCover

### LINKED ISSUE #3240: Use OpenCover to compute branch coverage and upload to sonarqube

costin-zaharia-sonarsource:
Fixes #3240

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -188,6 +188,7 @@ stages:
           projectName: 'SonarAnalyzer .Net'
           projectVersion: '$(SONAR_PROJECT_VERSION)'
           extraProperties: |
+            sonar.cs.opencover.reportsPaths="$(Build.SourcesDirectory)/**/coverage.xml"
             sonar.analysis.buildNumber=$(Build.BuildId)
             sonar.analysis.pipeline=$(Build.BuildId)
             sonar.analysis.sha1=$(System.PullRequest.SourceCommitId)
@@ -205,6 +206,7 @@ stages:
           projectName: 'SonarAnalyzer .Net'
           projectVersion: '$(SONAR_PROJECT_VERSION)'
           extraProperties: |
+            sonar.cs.opencover.reportsPaths="$(Build.SourcesDirectory)/**/coverage.xml"
             sonar.analysis.buildNumber=$(Build.BuildId)
             sonar.analysis.pipeline=$(Build.BuildId)
             sonar.analysis.sha1=$(Build.SourceVersion)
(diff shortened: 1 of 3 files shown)


costin-zaharia-sonarsource:
Since OpenCover does not have an Azure task I had to add the binaries to the repository.

andrei-epure-sonarsource:
> 
> 
> Since OpenCover does not have an Azure task I had to add the binaries to the repository.

can you instead [download the nuget](https://www.nuget.org/packages/OpenCover/) and unzip it with an msbuild script (see [unzip task](https://docs.microsoft.com/en-us/visualstudio/msbuild/unzip-task?view=vs-2019) and [download file task](https://docs.microsoft.com/en-us/visualstudio/msbuild/downloadfile-task?view=vs-2019)), and then reference the executable from e.g. `opencover.4.7.922\tools` ?

andrei-epure-sonarsource:
from the msbuild script you can probably also setup env vars with the path to the opencover executable, and even execute the tool itself to get the coverage

costin-zaharia-sonarsource:
> I left a comment asking if this can be done by downloading the nuget

I didn't thought at downloading the nuget, I will try

costin-zaharia-sonarsource:
> from the msbuild script you can probably also setup env vars with the path to the opencover executable, and even execute the tool itself to get the coverage

In order to keep things as simple as possible I've added the `OpenCover` as a dependency of the test project and used the `$(NUGET_PACKAGES)` variable.

azure-pipelines.yml
```diff
@@ -218,18 +220,12 @@ stages:
...
-            !**\obj\**
-          runInParallel: true
-          runTestsInIsolation: true
-          codeCoverageEnabled: true
-          platform: '$(buildPlatform)'
-          configuration: '$(buildConfiguration)'
-          rerunFailedTests: false
+      - powershell: |
+          cd sonaranalyzer-dotnet\tests
+          $(NUGET_PACKAGES)\opencover\4.7.922\tools\OpenCover.Console.exe -target:"C:/Program Files/dotnet/dotnet.exe" -targetargs:"test SonarAnalyzer.UnitTest\bin\$(BuildConfiguration)\net48\SonarAnalyzer.UnitTest.dll" -filter:"+[SonarAnalyzer.CFG]* +[SonarAnalyzer]* +[SonarAnalyzer.CSharp]* +[SonarAnalyzer.Utilities]* +[SonarAnalyzer.VisualBasic]*" -output:coverage.xml -oldStyle -register:Path32
```
andrei-epure-sonarsource: you may want to use `NugetPackageRoot` 

see https://github.com/SonarSource/sonar-dotnet/pull/3226 and https://github.com/NuGet/Home/issues/6301#issuecomment-350939781

costin-zaharia-sonarsource: Isn't this strictly related to MsBuild?

andrei-epure-sonarsource: Hmm I believe you are correct. That would mean we need to put it back in our Readme?

costin-zaharia-sonarsource:
I don't think so. This is used only by the build script.

azure-pipelines.yml
costin-zaharia-sonarsource: I don't think we need to add it in the Readme. This is used only by the build script.

PR #3243, opened 2020-03-23
Add FP repro for S1481

pavel-mikula-sonarsource:
Reproduces #3137

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/VariableUnused.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/VariableUnused.cs
@@ -96,4 +96,21 @@ namespace Tests.Diagnostics
             }
         }
     }
+
+    //https://github.com/SonarSource/sonar-dotnet/issues/3137
+    public class Repro_3137
+    {
+        public void GoGoGo(Logger log)
+        {
+            using var _ = log.BeginScope("XXX"); // Noncompliant FP, discard pattern cannot be used
+        }
+
+        public class Logger
+        {
+            public IDisposable BeginScope(string scope)
+            {
+                return null;
+            }
+        }
+    }
 }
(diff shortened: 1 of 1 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/VariableUnused.cs
```diff
@@ -96,4 +96,21 @@ static void Bar(List<int> list)
...
+    public class Repro_3137
+    {
+        public void GoGoGo(Logger log)
+        {
+            using var _ = log.BeginScope("XXX"); // Noncompliant FP, discard pattern cannot be used
+        }
+
+        public class Logger
+        {
+            public IDisposable BeginScope(string scope)
```
costin-zaharia-sonarsource: Have you changed you mind about https://github.com/SonarSource/sonar-dotnet/pull/3138#issuecomment-588835700?

If yes, we have already a PR for this: https://github.com/SonarSource/sonar-dotnet/pull/3138/files

costin-zaharia-sonarsource: In order to keep the repro as simple as possible we could remove the Logger class and use an already existing one (e.g. `File.Create(path)`)

pavel-mikula-sonarsource: Discard pattern is a special case here.
`BeginScope` explains underlying purpose of this syntax (with implicit EndScope usage). It demonstrates that the IDisposable should be kept after the statement to the end of scope. Contrary the file creation operation. I've added one more line to demonstrate the difference to be safe.

costin-zaharia-sonarsource: `File.Create` is just an example of a method which returns `IDisposable` and it could be replaced by anything. In my opinion it would be better to keep things simple and document on what syntax we raise and in what cases we not. Anyway I think this is not that important so it's up to you.

PR #3251, opened 2020-03-24
VS coverage - treat multiple coverage per line as branch coverage

### LINKED ISSUE #2622: vstest coverage does not distinguish between getters and setters in SQ/SC
### Description

SonarQube does not distinguish between getters and setters in abbreviated form for code coverage.

### Repro steps

1. Create a class with a simple property like `public int Timeout { get; set; }`.

2. Create a test and access the getter **or** setter.

3. Run the test and create a coverage file.

4. Upload the result to SonarQube.

### Expected behavior

SonarQube shows whether getter or setter has been tested.

### Actual behavior

SonarQube displays the property as **fully tested**, the code coverage file instead distinguish between getter and setter. Here is the example from the code coverage file (VS Test with Microsoft.CodeCoverage).

```xml
<function id="8449" token="0x6000007" name="get_Timeout()" namespace="Blank" type_name="SimpleModel" block_coverage="0.00" line_coverage="0.00" blocks_covered="0" blocks_not_covered="1" lines_covered="0" lines_partially_covered="0" lines_not_covered="1">
  <ranges>
    <range source_id="0" covered="no" start_line="63" start_column="26" end_line="63" end_column="30" />
  </ranges>
</function>
<function id="8457" token="0x6000008" name="set_Timeout(int)" namespace="Blank" type_name="SimpleModel" block_coverage="100.00" line_coverage="100.00" blocks_covered="1" blocks_not_covered="0" lines_covered="1" lines_partially_covered="0" lines_not_covered="0">
  <ranges>
    <range source_id="0" covered="yes" start_line="63" start_column="31" end_line="63" end_column="35" />
  </ranges>
</function>
```
Here is an example how SonarQube displays the coverage compared to Visual Studio:

SonarQube-vs-Visual-Studio

### Known workarounds

Not known.

### Related information

* Sonar Version: 7.7.0.23042
* SonarC# Version: 7.11.0.8083
* Visual Studio Version: 16.2.3
* If running through the Scanner for MSBuild, its version: 4.6.0.1930

andrei-epure-sonarsource:
Fixes #2622

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchCoverage.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchCoverage.java
@@ -28,7 +28,7 @@ class BranchCoverage {
   private int conditions;
   private int coveredConditions;
 
-  public BranchCoverage(int line, int conditions, int coveredConditions){
+  BranchCoverage(int line, int conditions, int coveredConditions){
     this.line = line;
     this.conditions = conditions;
     this.coveredConditions = coveredConditions;
@@ -42,7 +42,7 @@ class BranchCoverage {
     return conditions;
   }
 
-  public int getCoveredConditions() {
+  int getCoveredConditions() {
     return coveredConditions;
   }
 
@@ -71,5 +71,11 @@ class BranchCoverage {
   @Override
   public int hashCode() {
     return Objects.hash(line, conditions, coveredConditions);
+  }
+
+  @Override
+  public String toString() {
+    return "Branch coverage [line=" + line + ", conditions=" + conditions
+      + ", coveredConditionas=" + coveredConditions + "]";
   }
 }
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchCoverage.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchCoverage.java
@@ -76,6 +76,6 @@ class BranchCoverage {
   @Override
   public String toString() {
     return "Branch coverage [line=" + line + ", conditions=" + conditions
-      + ", coveredConditionas=" + coveredConditions + "]";
+      + ", coveredConditions=" + coveredConditions + "]";
   }
 }
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParser.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParser.java
@@ -134,8 +134,7 @@ public class VisualStudioCoverageXmlReportParser implements CoverageParser {
       }
 
       Map<Integer, List<Boolean>> fileCoverage = coveragePerLinePerFile.get(id);
-      if (!fileCoverage.isEmpty())
-      {
+      if (!fileCoverage.isEmpty()) {
         LOG.trace("Found coverage information about '{}' lines for file id '{}' , path '{}'",
           fileCoverage.size(), id, canonicalPath);
       }
@@ -158,8 +157,7 @@ public class VisualStudioCoverageXmlReportParser implements CoverageParser {
         int visits = 0;
         int entryCount = 0;
         // process line coverage
-        for (Boolean value : coverageValues)
-        {
+        for (Boolean value : coverageValues) {
           int visit = value ? 1 : 0;
           coverage.addHits(canonicalFilePath, lineId, visit);
           visits += visit;
@@ -167,8 +165,7 @@ public class VisualStudioCoverageXmlReportParser implements CoverageParser {
         }
         // where we have both covered and uncovered, use branch coverage
         // e.g. auto-implemented property with getter and setter
-        if (entryCount > 1)
-        {
+        if (entryCount > 1) {
           // this is not really branch coverage, but it's better than nothing
           coverage.addBranchCoverage(canonicalFilePath, new BranchCoverage(lineId, entryCount, visits));
         }
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParser.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParser.java
@@ -132,8 +132,7 @@ public class VisualStudioCoverageXmlReportParser implements CoverageParser {
       }
 
       LineCoverage fileCoverage = lineCoveragePerFile.get(id);
-      if (fileCoverage != null)
-      {
+      if (fileCoverage != null) {
         fileCoverage.transferData(coverage, canonicalPath);
       }
     }
@@ -169,10 +168,8 @@ public class VisualStudioCoverageXmlReportParser implements CoverageParser {
      * @param canonicalFilePath - the path of the file for which this class contains coverage data.
      */
     void transferData(Coverage coverage, String canonicalFilePath) {
-      if (!coveragePerLine.isEmpty()) {
-        LOG.trace("Found coverage information about '{}' lines for file id '{}' , path '{}'",
-          coveragePerLine.size(), fileId, canonicalFilePath);
-      }
+      LOG.trace("Found coverage information about '{}' lines for file id '{}' , path '{}'",
+        coveragePerLine.size(), fileId, canonicalFilePath);
 
       for (Map.Entry<Integer, List<Boolean>> lineCoverage : coveragePerLine.entrySet()) {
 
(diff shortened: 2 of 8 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchCoverage.java
```diff
@@ -72,4 +72,10 @@ public boolean equals(Object o) {
   public int hashCode() {
     return Objects.hash(line, conditions, coveredConditions);
   }
+
+  @Override
+  public String toString() {
+    return "Branch coverage [line=" + line + ", conditions=" + conditions
+      + ", coveredConditionas=" + coveredConditions + "]";
```
costin-zaharia-sonarsource: typo at `coveredConditionas`

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParser.java
```diff
@@ -34,22 +34,27 @@
...
 
     private final File file;
-    private final Map<Integer, List<Integer>> coveredLines = new HashMap<>();
-    private final Map<Integer, List<Integer>> uncoveredLines = new HashMap<>();
+    /**
+     * The outer map key is the file ID.
+     * The inner map key is the line number.
+     * The inner map values are a list of Boolean - true when the line is covered, false when it is not.
+     */
+    private final Map<Integer, Map<Integer, List<Boolean>>> coveragePerLinePerFile = new HashMap<>();
```
costin-zaharia-sonarsource: An option would be to create a new class LineCoverage and have a Map<Integer, List<LineCoverage>>. I think this will make the code a bit more easy to read, will allow us to move some behavior in the LineCoverage class (like aggregating the line values) and to reuse the code when we add the same behaviour or OpenCover.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParser.java
```diff
@@ -119,27 +133,51 @@ private void handleSourceFileTag(XmlParserHelper xmlParserHelper) {
...
-
-        for (Integer line : coveredLines.get(id)) {
-          coverage.addHits(canonicalPath, line, 1);
-        }
+      Map<Integer, List<Boolean>> fileCoverage = coveragePerLinePerFile.get(id);
+      if (!fileCoverage.isEmpty())
+      {
+        LOG.trace("Found coverage information about '{}' lines for file id '{}' , path '{}'",
+          fileCoverage.size(), id, canonicalPath);
       }
```
costin-zaharia-sonarsource: Since the trace log is about line coverage we could move the log in `processLineCoverage`

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParserTest.java
```diff
@@ -95,12 +96,74 @@ public void valid_with_correct_file_language() throws Exception {
```
costin-zaharia-sonarsource: I think it would be great to add a test also with multiple lines, some lines with only Get or Set covered and some with both.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParserTest.java
```diff
@@ -95,12 +96,74 @@ public void valid_with_correct_file_language() throws Exception {
...
+    );
+
+    assertThat(coverage.hits(filePath))
+      .hasSize(1)
+      .contains(
+        Assertions.entry(11, 1));
+
+    List<BranchCoverage> branchCoverages = coverage.getBranchCoverage(filePath);
+    assertThat(branchCoverages).hasSize(1);
+    assertThat(branchCoverages.get(0)).isEqualTo(new BranchCoverage(11, 2, 1));
```
costin-zaharia-sonarsource: We can replace these two lines with:
`assertThat(branchCoverages).containsExactly(new BranchCoverage(11, 2, 1));`

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParser.java
andrei-epure-sonarsource: I made a private class because 
- a universal solution should live in the `Coverage` class, and I don't want to come up with this in this PR
- the current solution is tied to the VS coverage XML (e.g. the path to the file lives in a separate section from the actual coverage information)

andrei-epure-sonarsource:
@costin-zaharia-sonarsource 
Overall, this is how things look like in the IDE:
GetSet_VS

And this is how they look in SonarQube:

_Note that while Visual Studio highlights what's partially covered in `ArrowMethod` (without mentioning what is covered and what not inside the partially covered code), we cannot so we show full coverage_

GetSet_SQ2


_The coverage tool ignores unreachable (dead) code, so we do as well_

GetSet_SQ4

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParserTest.java
costin-zaharia-sonarsource: minor: I think you forgot about this comment.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParserTest.java
```diff
@@ -95,12 +96,134 @@ public void valid_with_correct_file_language() throws Exception {
...
+    );
+
+    assertThat(coverage.hits(filePath))
+      .hasSize(1)
+      .contains(
+        Assertions.entry(11, 2));
+
+    List<BranchCoverage> branchCoverages = coverage.getBranchCoverage(filePath);
+    assertThat(branchCoverages).hasSize(1);
+    assertThat(branchCoverages.get(0)).isEqualTo(new BranchCoverage(11, 6, 2));
```
costin-zaharia-sonarsource: minor: We can use also here `containsExactly`

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParserTest.java
```diff
@@ -142,19 +139,18 @@ public void valid_with_multiple_getter_setter_per_line() throws Exception {
...
       .hasSize(1)
-      .contains(
-        Assertions.entry(11, 2));
+      .containsOnly(Assertions.entry(11, 2));
 
-    List<BranchCoverage> branchCoverages = coverage.getBranchCoverage(filePath);
-    assertThat(branchCoverages).hasSize(1);
-    assertThat(branchCoverages.get(0)).isEqualTo(new BranchCoverage(11, 6, 2));
+    assertThat(coverage.getBranchCoverage(filePath))
+      .hasSize(1)
```
costin-zaharia-sonarsource: I think that we don't need `hasSize` if we call `containsExactly`

costin-zaharia-sonarsource:
Just noticed that there are also 2 code smells to address.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParser.java
```diff
@@ -119,27 +131,67 @@ private void handleSourceFileTag(XmlParserHelper xmlParserHelper) {
...
+      coveragePerLine.get(lineId).add(isCovered);
     }
 
+    /**
+     * Transfers the coverage data to the Coverage object.
+     * @param coverage - the object which centralizes coverage information.
+     * @param canonicalFilePath - the path of the file for which this class contains coverage data.
+     */
+    void transferData(Coverage coverage, String canonicalFilePath) {
+      if (!coveragePerLine.isEmpty()) {
```
costin-zaharia-sonarsource: I was looking at the branch coverage report and not all the branches are covered. I think it's not possible to have this empty since we either add a entry or throw. WDYT?

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParserTest.java
andrei-epure-sonarsource: indeed, for small collections.

for big collections (or collections with more than 1 item), if it fails it can mean:
- either the number is different
- or one element is different

and then the dev must do a manual diff between the outputted contents of the expected and actual collection, which is painful

andrei-epure-sonarsource: I only removed `hasSize(1)`, as I said for more than one element failing due to size is helpful

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParser.java
andrei-epure-sonarsource: good point.
this is why full code coverage is important :)
I added a test with empty ranges and treating a NRE case

there would still be a case where the `source_files` is empty, but I don't think that's possible... but having an empty list of ranges for an existing file seems more plausible and I would like to be on the safe side

PR #3260, opened 2020-03-25
Do not fail the build when importing unit test reports for solutions with VB and C# projects

### LINKED ISSUE #3109: Import of unit test reports for a mixed C# - VB.NET solution is failing
When trying to import unit test reports in a solution containing both C# and VB.NET code, the import will fail and stop the analysis.

**Steps to Reproduce:**
 1. Create a simple multi language solution with tests.
    Ex:
```
        /MySimpleSolution
            /MySimpleCSharpProj
                /Program.cs
            /MySimpleCSharpProj.Test
                /UnitTest1.cs
            /MySimpleVBNetProj
                /Program.vb
            /MySimpleVBNet.Test
                /UnitTest1.vb
```
 2. Start step of analysis with unit test parameters `/d:sonar.cs.nunit.reportsPaths=CSharpResult.xml` `/d:sonar.vbnet.nunit.reportsPaths=VBNetResult.xml`
 3. Build solution
 4. Run the unit tests and save the results (can be in the same or different reports, analysis will fail in both cases):
    `nunit3-console.exe .\MySimpleCSharpProj.Test\bin\Debug\net461\MySimpleCSharpProj.Test.dll --result CSharpResult.xml`
    `nunit3-console.exe .\MySimpleVBNet.Test\bin\Debug\net461\MySimpleVBNet.Test.dll --result VBNetResult.xml`
 5. Run end step of analysis => will fail with the following error:
```
java.lang.UnsupportedOperationException: Can not add the same measure twice on [key=CSharpVBNetUnitTests]: DefaultMeasure[component=[key=CSharpVBNetUnitTests],metric=Metric[id=<null>,key=tests,description=Number of unit tests,type=INT,direction=1,domain=Coverage,name=Unit Tests,qualitative=false,userManaged=false,enabled=true,worstValue=<null>,bestValue=<null>,optimizedBestValue=false,hidden=false,deleteHistoricalData=false,decimalScale=<null>],value=1,fromCore=false,storage=org.sonar.scanner.sensor.DefaultSensorStorage@176f7f3b,saved=false]
	at org.sonar.scanner.sensor.DefaultSensorStorage.saveMeasure(DefaultSensorStorage.java:177)
	at org.sonar.scanner.sensor.DefaultSensorStorage.store(DefaultSensorStorage.java:137)
	at org.sonar.api.batch.sensor.measure.internal.DefaultMeasure.doSave(DefaultMeasure.java:96)
	at org.sonar.api.batch.sensor.internal.DefaultStorable.save(DefaultStorable.java:45)
	at org.sonar.plugins.dotnet.tests.UnitTestResultsImportSensor.analyze(UnitTestResultsImportSensor.java:88)
	at org.sonar.plugins.dotnet.tests.UnitTestResultsImportSensor.execute(UnitTestResultsImportSensor.java:74)
	at org.sonar.scanner.sensor.AbstractSensorWrapper.analyse(AbstractSensorWrapper.java:48)
	at org.sonar.scanner.sensor.ModuleSensorsExecutor.execute(ModuleSensorsExecutor.java:85)
	at org.sonar.scanner.sensor.ModuleSensorsExecutor.execute(ModuleSensorsExecutor.java:62)
	at org.sonar.scanner.scan.ModuleScanContainer.doAfterStart(ModuleScanContainer.java:82)
	at org.sonar.core.platform.ComponentContainer.startComponents(ComponentContainer.java:136)
```

**Context of failure:**
The `UnitTestResultsImportSensor` is trying to add global metrics (number of tests, number of test failures, test errors, ...) at the module level. Modules have been deprecated for a while now. So when both C# and VB.NET sensors try to register these metrics, they register it both at the project (fallback from modules) level. However, since we cannot register metrics on the same component (project in this case) more than once, the second sensor fails and crashes the analysis.
To avoid this issue, metrics should either be reported at the project level for all C# and VB.NET tests, or be reported on their corresponding folders.
Note that even if an exception is raised at this stage, this should not stop and fail the analysis, but merely report an error in the logs that test metrics could not be imported.

costin-zaharia-sonarsource:
Fixes #3109

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensor.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensor.java
@@ -25,6 +25,7 @@ import org.sonar.api.batch.sensor.Sensor;
 import org.sonar.api.batch.sensor.SensorContext;
 import org.sonar.api.batch.sensor.SensorDescriptor;
 import org.sonar.api.measures.CoreMetrics;
+import org.sonar.api.notifications.AnalysisWarnings;
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 import org.sonarsource.dotnet.shared.plugins.DotNetPluginMetadata;
@@ -34,25 +35,27 @@ import org.sonarsource.dotnet.shared.plugins.DotNetPluginMetadata;
  */
 public class UnitTestResultsImportSensor implements Sensor {
 
-  private static final Logger LOG = Loggers.get(CoverageReportImportSensor.class);
+  private static final Logger LOG = Loggers.get(UnitTestResultsImportSensor.class);
 
   private final WildcardPatternFileProvider wildcardPatternFileProvider = new WildcardPatternFileProvider(new File("."), File.separator);
   private final UnitTestResultsAggregator unitTestResultsAggregator;
   private final ProjectDefinition projectDef;
   private final String languageKey;
   private final String languageName;
+  private final AnalysisWarnings analysisWarnings;
 
   public UnitTestResultsImportSensor(UnitTestResultsAggregator unitTestResultsAggregator, ProjectDefinition projectDef,
-                                     DotNetPluginMetadata pluginMetadata) {
-    this(unitTestResultsAggregator, projectDef, pluginMetadata.languageKey(), pluginMetadata.languageName());
+                                     DotNetPluginMetadata pluginMetadata, AnalysisWarnings analysisWarnings) {
+    this(unitTestResultsAggregator, projectDef, pluginMetadata.languageKey(), pluginMetadata.languageName(), analysisWarnings);
   }
 
   public UnitTestResultsImportSensor(UnitTestResultsAggregator unitTestResultsAggregator, ProjectDefinition projectDef,
-    String languageKey, String languageName) {
+    String languageKey, String languageName, AnalysisWarnings analysisWarnings) {
     this.unitTestResultsAggregator = unitTestResultsAggregator;
     this.projectDef = projectDef;
     this.languageKey = languageKey;
     this.languageName = languageName;
+    this.analysisWarnings = analysisWarnings;
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensorTest.java
```diff
@@ -152,4 +157,22 @@ public void describe_is_global_and_only_on_language() {
...
+    SensorContextTester context = SensorContextTester.create(temp.newFolder());
+    UnitTestResults results = mock(UnitTestResults.class);
+
+    when(unitTestResultsAggregator.aggregate(Mockito.any(WildcardPatternFileProvider.class), Mockito.any(UnitTestResults.class)))
+      .thenReturn(results);
+
+    new UnitTestResultsImportSensor(unitTestResultsAggregator, ProjectDefinition.create(), "cs", "C#")
+      .analyze(context, results);
+
+    new UnitTestResultsImportSensor(unitTestResultsAggregator, ProjectDefinition.create(), "vb", "VB.Net")
```
christophe-zurn-sonarsource: _minor_: the actual language name used by the plugin for vb is `"VB.NET"` in uppercase, although it does not change anything for tests

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensor.java
```diff
@@ -76,37 +76,40 @@ public void execute(SensorContext context) {
   }
 
   void analyze(SensorContext context, UnitTestResults unitTestResults) {
-    UnitTestResults aggregatedResults = unitTestResultsAggregator.aggregate(wildcardPatternFileProvider, unitTestResults);
+    try {
+      UnitTestResults aggregatedResults = unitTestResultsAggregator.aggregate(wildcardPatternFileProvider, unitTestResults);
```
christophe-zurn-sonarsource: To avoid having a big block of code inside a try-catch statement, I'd suggest to keep the previous method as a helper ex: `saveTestMetrics` method, and just wrap that call inside the try-catch

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensor.java
```diff
@@ -76,37 +76,40 @@ public void execute(SensorContext context) {
...
+      Long executionTime = aggregatedResults.executionTime();
+      if (executionTime != null) {
+        context.<Long>newMeasure()
+          .forMetric(CoreMetrics.TEST_EXECUTION_TIME)
+          .on(context.project())
+          .withValue(executionTime)
+          .save();
+      }
+    } catch (Exception e) {
+      LOG.warn("Could not import unit test report: '{}'", e.getMessage());
```
christophe-zurn-sonarsource: To avoid confusion and be clear to the user, I would also add this warning message to the SonarQube UI warnings. (Using the `AnalysisWarnings` interface). So users will know right away that something did not work 100%

PR #3263, opened 2020-03-26
OpenCover - treat getter and setter on same line coverage

### LINKED ISSUE #3252: OpenCover - improve getter/setter partial coverage support
See #2622 - the same problem stands for OpenCover.

We should treat partial coverage of an auto-implemented property as branch coverage, due to the current SonarQube/SonarCloud limitations (of not being able to highlight what is being covered on a specific line).

```
public int Property { get; set;}
```

andrei-epure-sonarsource:
Fixes #3252


sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -119,6 +128,13 @@ private void handleSegmentPointTag(XmlParserHelper xmlParserHelper) {
           }
 
           coverage.addHits(identifiedFile, line, vc);
+
+          if (line == endLine) {
+            String elementIdentifier = String.format("%d-%d-%d-%d", line, endLine, startColumn, endColumn);
```
andrei-epure-sonarsource: here I use `elementIdentifier` and in the LineComplexCoverage class I use `methodName` - probably `elementIdentifier` is better, right?

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -131,4 +147,79 @@ private void handleSegmentPointTag(XmlParserHelper xmlParserHelper) {
...
+   * (e.g. an auto-implemented property can have a covered getter and un-covered setter)
+   *
+   * So this class holds information about methods which are declared on a single line.
+   */
+  private static class LineComplexCoverage {
+
+    String fileId;
+
+    // the key is the line ID
+    // the value is a list of method hits
```
andrei-epure-sonarsource: out of date comment

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
```diff
@@ -134,12 +134,61 @@ public void branchCoverage() throws Exception {
     assertThat(coverage.getBranchCoverage(filePath).get(0)).isEqualTo(new BranchCoverage(9, 4 , 2));
     assertThat(coverage.getBranchCoverage(filePath).get(1)).isEqualTo(new BranchCoverage(12, 2 , 1));
     assertThat(coverage.getBranchCoverage(filePath).get(2)).isEqualTo(new BranchCoverage(13, 2, 1));
-    assertThat(coverage.getBranchCoverage(filePath).get(3)).isEqualTo(new BranchCoverage(14, 4, 2));
+    assertThat(coverage.getBranchCoverage(filePath).get(3)).isEqualTo(new BranchCoverage(14, 6, 4));
```
andrei-epure-sonarsource: hard to identify the line in the test case

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
```diff
@@ -134,12 +134,61 @@ public void branchCoverage() throws Exception {
...
+
+    assertThat(coverage.files()).containsOnly(
+      filePath,
+      new File("GetSet\\FooCallsBar.cs").getCanonicalPath(),
+      new File("GetSetTests\\BarTests.cs").getCanonicalPath()
+    );
+    assertThat(coverage.hits(filePath))
+      .hasSize(10)
+      .contains(
+        // 2 hits from tests for Bar, 1 hit from tests for FooCallsBar
```
andrei-epure-sonarsource: this is a difference as oposed to vstest - opencover is quite accurrate

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
costin-zaharia-sonarsource: Indeed, `metodName` is confusing since it's not really a method name 😄. In the context of `OpenCover` I think that this is actually a `segmentPointRange` or `segmentPointIdentifier`.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -131,4 +147,79 @@ private void handleSegmentPointTag(XmlParserHelper xmlParserHelper) {
...
+        // e.g. auto-implemented property with getter and setter
+        if (methodHits.size() > 1) {
+          int coveredMethods = methodHits.countMethodsWithCoverage();
+          // this is not really branch coverage, but it's better than nothing
+          coverage.addBranchCoverage(fileId, new BranchCoverage(lineId, methodHits.size(), coveredMethods));
+        }
+      }
+    }
+
+    private static class MethodHits {
```
costin-zaharia-sonarsource: I think `SegmentPointHits` would be a better name

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -131,4 +147,79 @@ private void handleSegmentPointTag(XmlParserHelper xmlParserHelper) {
 
   }
 
+  /**
+   * This class is useful when on a single line there are multiple methods. Some may be covered, some not
```
costin-zaharia-sonarsource: I think this message is misleading. This class actually aggregates all the overlapping segments for a line together with their hits.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
```diff
@@ -134,12 +134,61 @@ public void branchCoverage() throws Exception {
     assertThat(coverage.getBranchCoverage(filePath).get(0)).isEqualTo(new BranchCoverage(9, 4 , 2));
     assertThat(coverage.getBranchCoverage(filePath).get(1)).isEqualTo(new BranchCoverage(12, 2 , 1));
     assertThat(coverage.getBranchCoverage(filePath).get(2)).isEqualTo(new BranchCoverage(13, 2, 1));
-    assertThat(coverage.getBranchCoverage(filePath).get(3)).isEqualTo(new BranchCoverage(14, 4, 2));
+    assertThat(coverage.getBranchCoverage(filePath).get(3)).isEqualTo(new BranchCoverage(14, 6, 4));
     assertThat(coverage.getBranchCoverage(filePath).get(4)).isEqualTo(new BranchCoverage(18, 6, 3));
   }
 
+    @Test
+  public void branchCoverage_getter_setter_complex_case() throws Exception {
```
costin-zaharia-sonarsource: We could use overlapping_segments instead of complex, WDYT?

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
andrei-epure-sonarsource: I guess it must be this
```
_ = y == 0 || z == 0; _ = x == y || y == z;
```

2 times (2 equalities + 1 OR) = 6 branches - sounds good

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -131,4 +147,79 @@ private void handleSegmentPointTag(XmlParserHelper xmlParserHelper) {
...
+      private MethodHits() {
+        methodHits = new HashMap<>();
+      }
+
+      int size() {
+        return methodHits.size();
+      }
+
+      int countMethodsWithCoverage() {
+        int number = 0;
```
andrei-epure-sonarsource: this can be 
```
        hitsPerSegment.values().stream().filter(hits -> hits > 0).count()
```

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
andrei-epure-sonarsource: I think it's `sequence`

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
andrei-epure-sonarsource: I will use "multiple_sequence_points_per_line" because the sequence points are not overlapping (they are on separate positions)

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -117,108 +118,110 @@ private void handleSegmentPointTag(XmlParserHelper xmlParserHelper) {
...
           }
         } else {
           LOG.debug("Skipping the fileId '{}', line '{}', vc '{}' because file '{}'" +
               " is not indexed or does not have the supported language.",
-            fileId, line, vc, identifiedFile);
+            fileId, line, visitCount, filePath);
         }
       } else {
-        LOG.debug("OpenCover parser: the fileId '{}' key is not contained in files", fileId);
+        LOG.debug("OpenCover parser: the filePath '{}' key is not contained in files", fileId);
```
andrei-epure-sonarsource: it seems we don't have a test for this

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -117,108 +118,110 @@ private void handleSegmentPointTag(XmlParserHelper xmlParserHelper) {
...
+        sequencePointCoveragePerLine.size(), filePath);
+
+      for (Map.Entry<Integer, SequencePointsCoverage> lineSequencePointCoverage : sequencePointCoveragePerLine.entrySet()) {
+
+        Integer lineId = lineSequencePointCoverage.getKey();
+        SequencePointsCoverage sequencePointsCoverage = lineSequencePointCoverage.getValue();
+
+        // where there are multiple sequence points on a line, use branch coverage
+        if (sequencePointsCoverage.size() > 1) {
+          int coveredSequencePointsOnLine = (int) sequencePointsCoverage.countCoveredSequencePoints();
```
andrei-epure-sonarsource: I don't think it's likely to have sequence points with more that Int.MAX_VALUE hits

but maybe I should do a test to fail nicely?

andrei-epure-sonarsource: or just change the value in `Coverage` to use long, unless the SQ API requires int

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
```diff
@@ -185,6 +185,14 @@ public void branchCoverage_getter_setter_complex_case() throws Exception {
           // line 21: first line inside BodyMethod - 3 statements (what is after 'goto' is ignored)
           new BranchCoverage(21, 3, 3)
         );
+
+      List<String> traceLogs = logTester.logs(LoggerLevel.TRACE);
+      assertThat(traceLogs).hasSize(35);
+
+      String logLine = traceLogs.get(32);
+      assertThat(logLine)
+        .startsWith("Found coverage information about '10' lines having multiple sequence points for file '")
```
andrei-epure-sonarsource: actually... it's about sequence points on a single line, that do not span on multiple lines

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
andrei-epure-sonarsource: I will just leave it as is

PR #3265, opened 2020-03-27
Deprecate NCover3 support

### LINKED ISSUE #3223: Deprecate NCover3 code coverage support
WHY we want to do this - in https://jira.sonarsource.com/browse/MMF-1997

In short, NCover3 is a legacy tool which is not supported anymore by NCover LLC.

Even more, at this moment, NCover Desktop offers support for VS 2010-2015, and it doesn't seem they have plans to improve that.

We will announce this deprecation on our [community forum](community.sonarsource.com/), and unless there will be strong push-back from the community and customers, we will drop support for NCover3 in the next LTS of SonarQube.

andrei-epure-sonarsource:
Fixes #3223

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/NCover3ReportParser.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/NCover3ReportParser.java
@@ -24,18 +24,24 @@ import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Predicate;
+import org.sonar.api.notifications.AnalysisWarnings;
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 
+/**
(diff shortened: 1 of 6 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/NCover3ReportParser.java
```diff
@@ -56,6 +57,8 @@ public void accept(File file, Coverage coverage) {
     }
 
     public void parse() {
+      LOG.warn("NCover3 coverage import is deprecated since version 8.6 of the plugin. " +
```
costin-zaharia-sonarsource: An idea would be to display this also on SQ/SC (using AnalysisWarnings) to increase visibility.

andrei-epure-sonarsource:
I'll need to add tests to `src/test/java/org/sonarsource/dotnet/shared/plugins/CodeCoverageProviderTest.java`

clean as you code...

andrei-epure-sonarsource:
manual testing done
image

PR #3267, opened 2020-03-27
Refactor DotCover and OpenCover coverage parsers to reuse code

costin-zaharia-sonarsource:

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePointCollector.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePointCollector.java
@@ -28,6 +28,9 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+// This class is responsible for publishing branch coverage for all the sequence points defined on the same line
+// since these can be only partially covered by tests.
+// e.g. properties with get and set defined on the same line
 public class SequencePointCollector {
   private static final Logger LOG = Loggers.get(SequencePointCollector.class);
   private Set<SequencePoint> sequencePoints = new HashSet<>();
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePointCollector.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePointCollector.java
@@ -1,0 +1,65 @@
+/*
+ * SonarSource :: .NET :: Shared library
+ * Copyright (C) 2014-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.dotnet.tests;
+
+import org.sonar.api.utils.log.Logger;
+import org.sonar.api.utils.log.Loggers;
+
+import java.util.HashSet;
+import java.util.List;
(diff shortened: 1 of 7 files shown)


sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
```diff
@@ -189,10 +189,11 @@ public void branchCoverage_getter_setter_multiple_sequence_points_per_line() thr
       List<String> traceLogs = logTester.logs(LoggerLevel.TRACE);
       assertThat(traceLogs).hasSize(35);
 
-      String logLine = traceLogs.get(32);
-      assertThat(logLine)
-        .startsWith("Found coverage information about '10' lines having single-line sequence points for file '")
-        .endsWith("Bar.cs'");
+
+      assertThat(traceLogs)
```
costin-zaharia-sonarsource: I've modified this since the order changed. I've made it now to not depend on order.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/SequencePointCollectorTest.java
```diff
@@ -0,0 +1,125 @@
...
+import org.mockito.Mockito;
+
+import static org.mockito.Mockito.*;
+
+public class SequencePointCollectorTest {
+  @Test
+  public void givenEmptyListOfPoints_publishCoverage_doesNotAddBranchCoverage() {
+    Coverage coverage = mock(Coverage.class);
+
+    SequencePointCollector sut = new SequencePointCollector();
```
costin-zaharia-sonarsource: sut = system under test (https://martinfowler.com/articles/mocksArentStubs.html)

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePointCollector.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public void add(SequencePoint sequencePoint){
+    // we only want to count covered sequences when these are on the same line
+    if (sequencePoint.getStartLine() == sequencePoint.getLineEnd()) {
+      sequencePoints.add(sequencePoint);
+    }
+  }
+
+  public void publishCoverage(Coverage coverage){
+    Map<String, List<SequencePoint>> sequencePointsPerFile = sequencePoints.stream().collect(Collectors.groupingBy(SequencePoint::getFilePath));
```
andrei-epure-sonarsource: I think it will be more readable if you iterate in a functional manner with `stream()` - to avoid the long definitions which are not bringing value

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/SequencePointCollectorTest.java
andrei-epure-sonarsource: it's the first time I see it, you could extend the name and use `systemUnderTest`

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePointCollector.java
costin-zaharia-sonarsource: Sorry, but I don't understand what you mean :) can you give an example?

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/SequencePointCollectorTest.java
costin-zaharia-sonarsource: The acronym is largely used and I hope it will enter soon in our vocabulary. If you have a strong opinion, I can change it, but I prefer to keep it like this since it's simple and it keeps the code lines short.

andrei-epure-sonarsource: that's fine, no strong opinion here

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePointCollector.java
andrei-epure-sonarsource: this method body can consist only of
```
sequencePoints.stream()
  .collect(Collectors.groupingBy(SequencePoint::getFilePath))
  forEach(fileSequencePoints -> publishCoverage(coverage, fileSequencePoints.getKey(), fileSequencePoints.getValue()));
```

costin-zaharia-sonarsource: good idea. done.

PR #3269, opened 2020-03-31
Fix S101 FP: IEnumerableExtensionsTests should not be reported

### LINKED ISSUE #3268: Fix S101 FP: IEnumerableExtensionsTests should not be reported
### Description

I've created a new class called IEnumerableExtensionsTests so I can add unit tests to my static class IEnumerableExtensions, containing some extensions methods to the IEnumerable interface.

Weirdly, my static class doesn't trigger this warning.

IEnumerableExtensionsTests: Trigger the warning (public class)
IEnumerableExtensions: Does not trigger the warning (internal static class)
Rule Id: S101

### Repro steps

```csharp
namespace Mynamespace.Extensions
{
    using System.Linq;

    internal static class IEnumerableExtensions
    {
        public static bool SafeAny<T>(this IEnumerable<T> enumerable)
        {
            return enumerable != null && enumerable.Any();
        }
    }
}
```

```csharp
namespace Mynamespace.Tests.Extensions
{
    using System.Linq;
    using Xunit;

    public class IEnumerableExtensionsTests
    {
        [Fact]
        public void SafeAny_MultipleElements_ReturnsTrue()
        {
            // Arrange
            var integers = GetDummyIEnumerable(numberOfElements: 3);

            // Act
            var result = integers.SafeAny();

            // Assert
            Assert.True(result);
        }

        private IEnumerable<int> GetDummyIEnumerable(int numberOfElements)
        {
            for (int i = 0; i < numberOfElements; i++)
            {
                yield return i;
            }
        }

        // ...
    }
}
```

### Expected behavior
I would expect the same behavior for both classes, at least. However, I believe that classes starting with 'I' from interface shouldn't trigger this warning.

### Actual behavior
IEnumerableExtensionsTests triggers the warning while IEnumerableExtensions doesn't.

### Related information
Visual Studio 2019 v16.4.5
SonarLint for Visual Studio 2019 v4.20.0.16339
.NET Core 3.1
Windows 10

pavel-mikula-sonarsource:
Fixes #3268

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
@@ -108,6 +108,17 @@ namespace SonarAnalyzer.Rules.CSharp
                     TypeKindNameMapping[typeDeclaration.Kind()],
                     identifier.ValueText, MessageFormatUnderscore));
                 return;
+            }
+
+            if (typeDeclaration is ClassDeclarationSyntax)
+            {
+                var className = typeDeclaration.Identifier.ValueText;
+                if (className != "ITest"
+                    && className != "ITests"
+                    && (className.EndsWith("Test") || className.EndsWith("Tests")))
+                {
+                    return;
+                }
             }
 
             if (symbol.DeclaringSyntaxReferences.Length > 1 &&
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs
@@ -110,15 +110,9 @@ namespace SonarAnalyzer.Rules.CSharp
                 return;
             }
 
-            if (typeDeclaration is ClassDeclarationSyntax)
+            if (typeDeclaration is ClassDeclarationSyntax && IsTestClassName(typeDeclaration.Identifier.ValueText))
             {
-                var className = typeDeclaration.Identifier.ValueText;
-                if (className != "ITest"
-                    && className != "ITests"
-                    && (className.EndsWith("Test") || className.EndsWith("Tests")))
-                {
-                    return;
-                }
+                return;
             }
 
             if (symbol.DeclaringSyntaxReferences.Length > 1 &&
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ClassName.cs
```diff
@@ -9,6 +9,11 @@ static class IEnumerableExtensions // Compliant
     {
     }
 
+    // https://github.com/SonarSource/sonar-dotnet/issues/3268
+    class IEnumerableExtensionsTests // Noncompliant FP, Classes ending with "Test" or "Tests" should not be reported
```
andrei-epure-sonarsource, line 13: it seems to me the fix for this is low-hanging, so I suggest to make it in this PR when getting some time. WDYT?

https://github.com/SonarSource/sonar-dotnet/blob/master/sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/Rules/Naming/ClassName.cs#L53

andrei-epure-sonarsource: also, this FP applies to both C# and VB.NET, right?

pavel-mikula-sonarsource: No, it's not related to VB.NET. VB.NET has very simplified implementation (regex-check-only) with no other logic for now.

pavel-mikula-sonarsource: It is low hanging indeed, just new if..return somewhere here in C# rule:
[https://github.com/SonarSource/sonar-dotnet/blob/3be496ff5f8d01ca638eee1a8d4411dc8c55439d/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs#L112](https://github.com/SonarSource/sonar-dotnet/blob/3be496ff5f8d01ca638eee1a8d4411dc8c55439d/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs#L112)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs
```diff
@@ -110,6 +110,17 @@ private static void CheckTypeName(SyntaxNodeAnalysisContext context)
                 return;
             }
 
+            if (typeDeclaration is ClassDeclarationSyntax)
+            {
+                var className = typeDeclaration.Identifier.ValueText;
+                if (className != "ITest"
```
andrei-epure-sonarsource, line 116: nitpick - I think the checks can be in a private  method / local function to make this `if` easier to read

Issue #3271, opened 2020-03-31
Fix S3353 FP: With Tuple Assignment


pavel-mikula-sonarsource:
Side effect of #3239, rule S3353 is also raised

```C#
var tmp = 0;
var flag = true;
(flag, tmp) = (false, 5);
```

andrei-epure-sonarsource:
is there a way to figure out what other rules could have been affected by this?

pavel-mikula-sonarsource:
AFAIK there's only peach and the hard way (manually try it rule by rule).

pavel-mikula-sonarsource:
Another example reported in #3661: S3353 raises a false positive with implicit conversion from a constant value.

### Repro steps

```C#
public struct MyStruct<T>
{
    public static implicit operator MyStruct<T>(T value) { return default; }
}

public static class Example
{
    public static MyStruct<int> Get()
    {
        MyStruct<int> x = 1;
        return x;
    }
}
```

We should check if the type can be declared with `const` modifier.

PR #3272, opened 2020-03-31
Add FP repro for S3353

pavel-mikula-sonarsource:
Reproduces #3271
Should be merged after 8.6 release.


pavel-mikula-sonarsource:
Burgr notification fails

PR #3274, opened 2020-04-01
Add FP repro for S3353

pavel-mikula-sonarsource:
Reproduces #3271
Should be merged after 8.6 release.


pavel-mikula-sonarsource:
Burgr notification is failing, but let's do the review process and deal with tooling issue separately.

See note from Tom (you're cc in Slack thread)
> in the case of a PR you should not have the "branch" field in the burgr payload but a "pr_number" field with the number of the PR as value, in other build engines like travis or cirrus we detect when we are in a PR and modify this field accordingly

PR #3278, opened 2020-04-02
Update RSPEC

pavel-mikula-sonarsource:


pavel-mikula-sonarsource:
Java ITs failed due to S4564. CWE and OWASP tag were removed. Investingating...

pavel-mikula-sonarsource:
S4564 is deprecated by S5753. 
That removed tags from S4564, rule-api updated JSON and Java IT failed, because it's used in the test.
S5753 is not implemented yet so JSON should not be changed yet. We'll need to update Java test `org.sonar.plugins.csharp.CSharpSonarRulesDefinitionTest.test_security_standards_with_vulnerability` as well when implementing S5753.

PR #3281, opened 2020-04-02
Add GITHUB_BRANCH and GIT_SHA1 env var

### LINKED ISSUE #3280: Fix azure pipeline burgr notifications
In order to have the burgr notifications working properly we need to update [azure_pipelines.yml](https://github.com/SonarSource/sonar-dotnet/blob/fa5a94d/azure-pipelines.yml#L297) to populate also `GITHUB_BRANCH` and `GIT_SHA1` environment variables.

GITHUB_BRANCH: $(Build.SourceBranchName)
GIT_SHA1: $(Build.SourceVersion)

costin-zaharia-sonarsource:
- in order to have the burgr notifications working

Fixes #3280

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -296,6 +296,8 @@ stages:
         env:
           ARTIFACTORY_DEPLOY_USERNAME: $(ARTIFACTORY_DEPLOY_USERNAME)
           ARTIFACTORY_DEPLOY_PASSWORD: $(ARTIFACTORY_DEPLOY_PASSWORD)
+          GITHUB_BRANCH: $(Build.SourceBranchName)
+          GIT_SHA1: $(Build.SourceVersion)
         inputs:
           goals: 'deploy'
           options: >-


azure-pipelines.yml
```diff
@@ -296,6 +296,8 @@ stages:
         env:
           ARTIFACTORY_DEPLOY_USERNAME: $(ARTIFACTORY_DEPLOY_USERNAME)
           ARTIFACTORY_DEPLOY_PASSWORD: $(ARTIFACTORY_DEPLOY_PASSWORD)
+          GITHUB_BRANCH: $(Build.SourceBranchName)
```
andrei-epure-sonarsource, line 299: so until now we didn't have burgr notifications at all for the Java plugin? 🤔

andrei-epure-sonarsource: do you know why the `burgr` notifications are working in other tasks?

maybe because they're done by `- template: stage-with-burgr-notifications.yml@commonTemplates` and this one is done by the `mvn` build?

costin-zaharia-sonarsource: If I understood correctly from @tomverin this affects only the display of the release decoration in burgr and it's related to the new release workflow.

We need to put the env variables here so they will get into buildinfo.json from [artifactory](https://repox.jfrog.io/repox/webapp/#/builds/sonar-dotnet/16497/1585731153189/json/org.sonarsource.dotnet:sonar-csharp-plugin:8.6.0.16497) (see Build Info JSON tab) and will be later used by the release workflow.

PR #3284, opened 2020-04-03
Fix S1905 FP with `OfType` and nullable types

### LINKED ISSUE #3273: Fix S1905 FP with `OfType` and reference and nullable value types
### Description

[S1905 ](https://jira.sonarsource.com/browse/RSPEC-1905) raises FPs when analyzing C# 8 nullable types code.

reported by Miroslav Jeras on our [community forum](https://community.sonarsource.com/t/s1905-false-positives-generated-by-oftype-in-nullability-scenarios/22279/2)

### Repro steps

```
    public class CastOnNullable
    {
        public static IEnumerable<string> UsefulCast()
        {
            var nullableStrings = new string?[] { "one", "two", null, "three" };
            return nullableStrings.OfType<string>(); // Noncompliant FP - filters out the null
        }
     }
```

### Related information

* C#/VB.NET Plugins Version 8.5
* MSBuild / dotnet version - MSBuild 16 is needed for C# 8

andrei-epure-sonarsource:
Fix  #3273


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RedundantCast.CSharp8.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Tests.Diagnostics
+{
+    public class CastOnNullable
+    {
+        public static IEnumerable<string> UsefulCast()
+        {
+            var nullableStrings = new string?[] { "one", "two", null, "three" };
```
pavel-mikula-sonarsource, line 11: Using C# 6 with regular strings produces the same result - skipping `null` value. I think it should be FP for `string` as well (when it's FP for `string?`).
```C#
var normalStrings = new string[] { "one", "two", null, "three" };
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RedundantCast.CSharp8.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Tests.Diagnostics
+{
+    public class CastOnNullable
```
pavel-mikula-sonarsource, line 7: As a reproducer, it would be nice to have link to the Issue here.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RedundantCast.CSharp8.cs
andrei-epure-sonarsource: that's a very good point!

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantCast.cs
```diff
@@ -106,19 +106,24 @@ private static void CheckExtensionMethodInvocation(SyntaxNodeAnalysisContext con
...
-                !returnType.TypeArguments.Any())
+            var castType = returnType.TypeArguments.First();
+            if (methodSymbol.Name == "OfType"
+                && (castType.IsReferenceType || castType.Name == "Nullable"))
             {
                 return;
             }
 
-            var castType = returnType.TypeArguments.First();
+            var elementType = GetElementType(invocation, methodSymbol, context.SemanticModel);
```
andrei-epure-sonarsource, line 122: as a side effect, less semantic model queries will be done

sonaranalyzer-dotnet/its/expected/Nancy/Nancy.Demo.Hosting.Aspnet-{E127FED3-01C0-41BA-BF83-D8DCDD827D6A}-S1905.json
```diff
@@ -1,17 +0,0 @@
...
-"region":  {
-"startLine":  38,
-"startColumn":  101,
-"endLine":  38,
-"endColumn":  131
-}
-}
-}
-]
-}
```
costin-zaharia-sonarsource, line 17: Wasn't this a true positive?
```
.WithModel(routeCacheProvider.GetCache().RetrieveMetadata<MyUberRouteMetadata>().OfType<MyUberRouteMetadata>())
```
In this case `RetrieveMetadata<MyUberRouteMetadata>` will return `IEnumerable<MyUberRouteMetadata>`

andrei-epure-sonarsource: If `MyUberRouteMetadata` is a reference type and it's not a struct, than `OfType<MyUberRouteMetadata>` filters out `null` values

andrei-epure-sonarsource: and it's a class
```
public class MyUberRouteMetadata
```

costin-zaharia-sonarsource: Ok, since the IEnumerable could contain null values and there is no way to know for sure if these are present or not, we prefer to check this only for value types (https://github.com/SonarSource/sonar-dotnet/issues/3273#issuecomment-656010838)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantCast.cs
```diff
@@ -106,19 +106,24 @@ private static void CheckExtensionMethodInvocation(SyntaxNodeAnalysisContext con
                 return;
             }
 
-            var elementType = GetElementType(invocation, methodSymbol, context.SemanticModel);
-            if (elementType == null)
+            if (!(methodSymbol.ReturnType is INamedTypeSymbol returnType) ||
+                !returnType.TypeArguments.Any())
```
costin-zaharia-sonarsource, line 110: Extracting a method with this condition (`ReturnsGenericType`) would make the code easier to read.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantCast.cs
```diff
@@ -106,19 +106,24 @@ private static void CheckExtensionMethodInvocation(SyntaxNodeAnalysisContext con
...
             {
                 return;
             }
 
-            if (!(methodSymbol.ReturnType is INamedTypeSymbol returnType) ||
-                !returnType.TypeArguments.Any())
+            var castType = returnType.TypeArguments.First();
+            if (methodSymbol.Name == "OfType"
+                && (castType.IsReferenceType || castType.Name == "Nullable"))
```
costin-zaharia-sonarsource, line 117: Extracting a method with this condition (`IsOfTypeCallOnValueType`) will make the code easier to read.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantCast.cs
costin-zaharia-sonarsource: Indeed, it is better to call this here but on the other hand we do get symbols for all the method invocations when we could filter them first by name.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantCast.cs
andrei-epure-sonarsource: well, it's not on a value type, it's on a nullable type (either nullable value type or reference type).

Do you really think such a method would make the code more readable? `IsOfTypeOnNullableType`?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RedundantCast.cs
```diff
@@ -42,5 +42,30 @@ public void N(int[,] numbers)
...
+        public void OfTypeWithReferenceTypes()
+        {
+            var stringArrayWithNull = new string[] { "one", "two", null, "three" };
+            var filteredStringArray = stringArrayWithNull.OfType<string>(); // Compliant, has 3 items, without 'null'
+
+            var enumerableOfObjects = new object[] { 1, "one", null };
+            var filteredObjectArray = enumerableOfObjects.OfType<object>(); // Compliant, has 2 items, without 'null'
+
+            var enumerableOfNullableInt = new int?[] { 1, 2, null };
+            var filteredInts = enumerableOfNullableInt.OfType<int?>(); // Compliant, has 2 items, without 'null'
```
costin-zaharia-sonarsource, line 56: For the sake of completeness we can add also a FN  with an array of strings containing only non null values.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantCast.cs
costin-zaharia-sonarsource: Indeed we can extract just `castType.IsReferenceType || castType.Name == "Nullable"` as `IsNullableType`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RedundantCast.cs
costin-zaharia-sonarsource: Yes, but it will document the fact that we don't check for nullable or value types.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantCast.cs
andrei-epure-sonarsource: I chose `CanHaveNullValue` because `IsNullableType` is ambiguous (`castType.Name == "Nullable"` would suffice)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/RedundantCast.cs
```diff
@@ -99,35 +99,44 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                            returnType.ToMinimalDisplayString(context.SemanticModel, invocation.SpanStart)));
+                    }
+                }
             }
         }
 
+        /// If the invocation one of the <see cref="CastIEnumerableMethods"/> extensions, returns the method symbol.
+        private static IMethodSymbol GetEnumerableExtensionSymbol(InvocationExpressionSyntax invocation, SemanticModel semanticModel) =>
+            invocation.GetMethodCallIdentifier() is { } methodName
+            && CastIEnumerableMethods.Contains(methodName.ValueText)
```
costin-zaharia-sonarsource, line 127: Nice!

PR #3300, opened 2020-04-24
Fix branch coverage when a file is tested by multiple projects

### LINKED ISSUE #3296: Fix Code coverage import from multiple test projects
See [this community thread](https://community.sonarsource.com/t/sonarsource-reports-invalid-code-coverage-when-using-opencover/23357) and [this blog post](https://r-vm.com/sonarsource-showing-too-many-conditions-in-code-coverage-with-multiple-dotnetcore-test-projects) and also [another community topic](https://community.sonarsource.com/t/getter-setter-properties-said-not-covered-with-c-analyzer-8-6/22851/5)

We need to first reproduce the issue:
- create a solution with multiple test projects
- analyze the code

And then fix the root cause

costin-zaharia-sonarsource:
Fixes #3296

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePoint.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePoint.java
@@ -21,18 +21,22 @@ package org.sonar.plugins.dotnet.tests;
 
 public class SequencePoint {
   private final int lineStart;
+  private final int columnStart;
   private final int lineEnd;
+  private final int columnEnd;
   private final int hits;
   private String filePath;
 
-  SequencePoint(String filePath, int lineStart, int lineEnd, int hits) {
-    this(lineStart, lineEnd, hits);
+  SequencePoint(String filePath, int lineStart, int columnStart, int lineEnd, int columnEnd, int hits) {
+    this(lineStart, columnStart, lineEnd, columnEnd, hits);
     this.filePath = filePath;
   }
 
-  SequencePoint(int lineStart, int lineEnd, int hits){
+  SequencePoint(int lineStart, int columnStart, int lineEnd, int columnEnd, int hits) {
     this.lineStart = lineStart;
+    this.columnStart = columnStart;
     this.lineEnd = lineEnd;
+    this.columnEnd = columnEnd;
     this.hits = hits;
   }
 
@@ -50,5 +54,9 @@ public class SequencePoint {
 
   public int getHits() {
     return hits;
+  }
+
+  public String getCoverageKey() {
+    return String.format("%d-%d-%d-%d", lineStart, columnStart, lineEnd, columnEnd);
   }
 }
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePoint.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePoint.java
@@ -28,11 +28,7 @@ public class SequencePoint {
   private String filePath;
 
   SequencePoint(String filePath, int lineStart, int columnStart, int lineEnd, int columnEnd, int hits) {
-    this(lineStart, columnStart, lineEnd, columnEnd, hits);
     this.filePath = filePath;
-  }
-
-  SequencePoint(int lineStart, int columnStart, int lineEnd, int columnEnd, int hits) {
     this.lineStart = lineStart;
     this.columnStart = columnStart;
     this.lineEnd = lineEnd;
@@ -56,7 +52,7 @@ public class SequencePoint {
     return hits;
   }
 
-  public String getCoverageKey() {
-    return String.format("%d-%d-%d-%d", lineStart, columnStart, lineEnd, columnEnd);
+  public String getUniqueIdentifier() {
+    return String.format("%s-%d-%d-%d-%d", filePath, lineStart, columnStart, lineEnd, columnEnd);
   }
 }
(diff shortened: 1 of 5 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePointCollector.java
```diff
@@ -59,9 +59,17 @@ private static void publishCoverage(Coverage coverage, String filePath, List<Seq
         int line = lineSequencePoints.getKey();
 
         List<SequencePoint> linePoints = lineSequencePoints.getValue();
-        int coveredPoints = (int)linePoints.stream().filter(point -> point.getHits() > 0).count();
 
-        coverage.addBranchCoverage(filePath, new BranchCoverage(line, linePoints.size(), coveredPoints));
+        Map<String, List<SequencePoint>> groupsByCoverageKey = linePoints.stream().collect(Collectors.groupingBy(SequencePoint::getCoverageKey));
```
andrei-epure-sonarsource, line 63: I would a comment above `// the key is the SequencePoint unique identifier, the value is a list of SequencePoint information coming from different sources (e.g. different coverage reports for the same file)`

I cannot think of a better name at this point

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePoint.java
```diff
@@ -51,4 +55,8 @@ public int getLineEnd() {
   public int getHits() {
     return hits;
   }
+
+  public String getCoverageKey() {
```
andrei-epure-sonarsource, line 59: I would name it `getUniqueIdentifier` to make it obvious that it identifies the SequencePoint

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePoint.java
```diff
@@ -51,4 +55,8 @@ public int getLineEnd() {
...
+    return String.format("%d-%d-%d-%d", lineStart, columnStart, lineEnd, columnEnd);
```
andrei-epure-sonarsource, line 60: maybe include `filePath` as well?

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePointCollector.java
```diff
@@ -59,9 +59,17 @@ private static void publishCoverage(Coverage coverage, String filePath, List<Seq
...
+
+        int coveredPoints = 0;
+        for (Map.Entry<String, List<SequencePoint>> group: groupsByCoverageKey.entrySet()){
```
andrei-epure-sonarsource, line 66: maybe rename `group` to `sequencePointEntries` - it's basically the various entries for the same sequence point

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePointCollector.java
andrei-epure-sonarsource: maybe `uniqueSequencePointInfo`

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePointCollector.java
```diff
@@ -59,9 +59,17 @@ private static void publishCoverage(Coverage coverage, String filePath, List<Seq
         int line = lineSequencePoints.getKey();
 
         List<SequencePoint> linePoints = lineSequencePoints.getValue();
-        int coveredPoints = (int)linePoints.stream().filter(point -> point.getHits() > 0).count();
 
-        coverage.addBranchCoverage(filePath, new BranchCoverage(line, linePoints.size(), coveredPoints));
+        Map<String, List<SequencePoint>> groupsByCoverageKey = linePoints.stream().collect(Collectors.groupingBy(SequencePoint::getCoverageKey));
+
+        int coveredPoints = 0;
```
andrei-epure-sonarsource, line 65: maybe `coveredSequencePoints` to keep the same naming

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/SequencePointCollectorTest.java
```diff
@@ -105,18 +105,119 @@ public void givenMultiplePointsPerLineInDifferentFiles_publishCoverage_addsBranc
...
+    sut.add(new SequencePoint(firstPath, 1, 2, 1, 4, 0));
+
+    sut.publishCoverage(coverage);
+
+    verify(coverage, Mockito.times(1))
+      .addBranchCoverage(eq(firstPath), argThat(branchCoverage -> IsMatch(branchCoverage, 1, 1, 1)));
+  }
+
+  @Test
+  public void givenMultiplePointsForTheSameCodeWithoutHits_publishCoverage_doesNotSumConditionsMultipleTimes() {
```
andrei-epure-sonarsource, line 155: test name should end with `doesNotSumConditions` I think (at all)

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/SequencePointCollector.java
costin-zaharia-sonarsource: I think in this case we can let it as it is, since it's used close to declaration and keeps the lines short.

costin-zaharia-sonarsource:
Closing this for now since it solves the problem only partially

PR #3302, opened 2020-04-27
Fix OpenCover branch coverage import by using BranchPoints

costin-zaharia-sonarsource:
When a code file is touched by unit tests from different test projects,
it will appear multiple times in the open cover reports. Since different
tests can cover different branches we need to correctly identify the
covered branches and count them only once.

SequencePoint does not have enough information to allow this
and because of that we have to rely on BranchPoints.

Fixes branch coverage import for #3296

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchPoint.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchPoint.java
@@ -25,13 +25,15 @@ class BranchPoint {
   private final int offsetEnd;
   private final int startLine;
   private final int hits;
+  private final int path;
   private final String filePath;
 
-  public BranchPoint(String filePath, int startLine, int offset, int offsetEnd, int hits) {
+  public BranchPoint(String filePath, int startLine, int offset, int offsetEnd, int path, int hits) {
     this.filePath = filePath;
     this.startLine = startLine;
     this.offset = offset;
     this.offsetEnd = offsetEnd;
+    this.path = path;
     this.hits = hits;
   }
 
@@ -48,6 +50,6 @@ class BranchPoint {
   }
 
   public String getUniqueKey() {
-    return String.format("%s-%d-%d-%d", filePath, startLine, offset, offsetEnd);
+    return String.format("%s-%d-%d-%d-%d", filePath, startLine, offset, offsetEnd, path);
   }
 }
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchPoint.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchPoint.java
@@ -25,6 +25,7 @@ class BranchPoint {
   private final int offsetEnd;
   private final int startLine;
   private final int hits;
+  // identifier for code path
   private final int path;
   private final String filePath;
 
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchPoint.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchPoint.java
@@ -1,0 +1,53 @@
+/*
+ * SonarSource :: .NET :: Shared library
+ * Copyright (C) 2014-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package org.sonar.plugins.dotnet.tests;
+
+class BranchPoint {
+  private final int offset;
+  private final int offsetEnd;
+  private final int startLine;
+  private final int hits;
+  private final String filePath;
+
+  public BranchPoint(String filePath, int startLine, int offset, int offsetEnd, int hits) {
+    this.filePath = filePath;
+    this.startLine = startLine;
+    this.offset = offset;
+    this.offsetEnd = offsetEnd;
+    this.hits = hits;
+  }
+
+  public int getStartLine() {
+    return startLine;
+  }
+
+  public int getHits() {
+    return hits;
+  }
+
+  public String getFilePath() {
+    return filePath;
+  }
+
+  public String getUniqueKey() {
+    return String.format("%s-%d-%d-%d", filePath, startLine, offset, offsetEnd);
+  }
+}
--- a/sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/CoverageTest.java
+++ b/sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/CoverageTest.java
@@ -70,7 +70,7 @@ public class CoverageTest {
   }
 
   @Test
-  public void testBranchCoverage(){
+  public void testBranchCoverageBySequencePoint(){
     Coverage first = new Coverage();
 
     BranchCoverage fooFirstLine = new BranchCoverage(1, 6, 3);
@@ -83,8 +83,8 @@ public class CoverageTest {
     first.addBranchCoverage("foo.txt", fooThirdLine);
     first.addBranchCoverage("bar.txt", barFirstLine);
 
-    assertThat(first.getBranchCoverage("foo.txt")).containsExactly(fooFirstLine, fooThirdLine);
-    assertThat(first.getBranchCoverage("bar.txt")).containsExactly(barFirstLine);
+    assertThat(first.getBranchCoverageBySequencePoints("foo.txt")).containsExactly(fooFirstLine, fooThirdLine);
+    assertThat(first.getBranchCoverageBySequencePoints("bar.txt")).containsExactly(barFirstLine);
 
     Coverage second = new Coverage();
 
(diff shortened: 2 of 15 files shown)


sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
```diff
@@ -129,16 +130,56 @@ public void branchCoverage() throws Exception {
         Assertions.entry(18, 1),
         Assertions.entry(25, 1));
 
-    assertThat(coverage.getBranchCoverage(filePath).size()).isEqualTo(5);
+    assertThat(coverage.getBranchCoverage(filePath))
+      .hasSize(5)
+      .contains(
+        new BranchCoverage(9, 4 , 2),
```
andrei-epure-sonarsource, line 136: for the record, the test file is [here](https://github.com/SonarSource/sonar-dotnet/blob/aa05edb2b401dacac9cc1423422fcc8729623f13/sonar-dotnet-shared-library/src/test/resources/opencover/coverage_branches.xml)

this puzzles me, line 9 is:
```
  if (x == 0 || y < 0)
```

From the ccov report:
```
<BranchPoint vc="0" uspid="11" ordinal="0" offset="2" sl="9" path="0" offsetend="4" fileid="1" />
<BranchPoint vc="1" uspid="12" ordinal="1" offset="2" sl="9" path="1" offsetend="10" fileid="1" />
<BranchPoint vc="1" uspid="13" ordinal="2" offset="13" sl="9" path="0" offsetend="15" fileid="1" />
<BranchPoint vc="0" uspid="14" ordinal="3" offset="13" sl="9" path="1" offsetend="22" fileid="1" />
```

The 4 conditions, I imagine:
- offset 2-4 - the IF - why not executed? because not all branches are executed?
- offset 2-10 - the IF and `x==0` - executed as first instruction in the OR
- offset 13-15 - the OR itself - executed with short-circuit
- offset 13-22 - the OR and `y < 10` - not executed because of short-circuit

I think we parse it well, however it's slightly confusing.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
```diff
@@ -129,16 +130,56 @@ public void branchCoverage() throws Exception {
...
+        new BranchCoverage(12, 2 , 1),
```
andrei-epure-sonarsource, line 137: For the record,line 12 should be
```
            _ = y == 0 || z == 0;
```
and given the tests is invoked with `x = 0, y = 0, z = 0`, then the boolean short-circuit means that `z==0` is not executed at all -> CORRECT

andrei-epure-sonarsource: same for line 13

andrei-epure-sonarsource: same idea for line 14

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
```diff
@@ -129,16 +130,56 @@ public void branchCoverage() throws Exception {
...
+        new BranchCoverage(13, 2, 1),
+        new BranchCoverage(14, 4, 2),
+        new BranchCoverage(18, 6, 3));
```
andrei-epure-sonarsource, line 140: for the record, line 18 is
```
            return x < 2
                ? y < 3
                    ? z < 1
                        ? 1
                        : 2
                    : 3
                : 4;
```

3 out of 6 executed as all have value `0` - CORRECT

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
```diff
@@ -129,16 +130,56 @@ public void branchCoverage() throws Exception {
...
+
+    assertThat(coverage.getBranchCoverage(filePath)).containsOnly(new BranchCoverage(5, 2, 2));
+  }
+
+  @Test
+  public void branchCoverage_codeFile_unsupportedFile() throws Exception {
+    Coverage coverage = new Coverage();
+    String filePath = new File("BranchCoverage3296\\Code\\ValueProvider.cs").getCanonicalPath();
+
+    new OpenCoverReportParser(alwaysFalse).accept(new File("src/test/resources/opencover/code_tested_by_multiple_projects.xml"), coverage);
```
andrei-epure-sonarsource, line 159: I would add a comment above like `// Notice we pass "alwaysFalse" as a predicate`

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/CoverageTest.java
```diff
@@ -103,21 +103,128 @@ public void testBranchCoverageWithMultipleEntriesOnTheSameLine(){
...
+  }
+
+  @Test
+  public void givenSingleBranchPoint_getBranchCoverage_returnsEmpty() {
+    final String filePath = "filePath";
+
+    Coverage sut = new Coverage();
+    sut.add(new BranchPoint(filePath, 1, 2, 3, 4));
+
+    assertThat(sut.getBranchCoverage(filePath)).isEmpty();
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: given I don't initially get it, I think it's good to add a comment to explain this in the test - it will help maintain the codebase

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
andrei-epure-sonarsource: (it wouldn't hurt to add this as a comment - we will rarely change these tests and when we will, it will ease understanding them)

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
andrei-epure-sonarsource: this is basically the only difference between this and the above test and it's easy to miss

andrei-epure-sonarsource:
also, not related to this PR, but we should add tests for dotCover and VS Coverage with multiple test projects

sonar-dotnet-shared-library/src/test/resources/opencover/invalid_file_id.xml
```diff
@@ -0,0 +1,4107 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+Below is the code for which this coverage file has been created,
+together with the commands to generate the file and the tools versions.
+
+Note: the file paths were manually changed inside the report to be relative and
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: But there is a small mention in the note: "the file id was modified to be invalid". 
I will put the changes as a list to make it a bit more visible.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
costin-zaharia-sonarsource: I've added the code as comment before each BranchCoverage entry in test. I know that OpenCover computes the conditions by looking at the generated IL but I'm not sure how and I don't want to make the wrong assumptions.

costin-zaharia-sonarsource: I think that the cases are x== 0, x != 0, y <0  and y>0 and it matches the number of branch points...

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
costin-zaharia-sonarsource: In this case the decompiled code is `int num1 = y == 0 ? 1 : (z == 0 ? 1 : 0);`
but the branches seem to be:
- y == 0 || z == 0 -> 1
- y != 0 && z != 0 -> 0

Fun fact: I would expect this, in the initial form (_ = y == 0 || z == 0;) to be a boolean not an int.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchPoint.java
```diff
@@ -25,13 +25,15 @@
   private final int offsetEnd;
   private final int startLine;
   private final int hits;
+  private final int path;
```
andrei-epure-sonarsource, line 28: it would be nice to add a comment for `path`

andrei-epure-sonarsource: (especially as there's no obvious link between the `BranchCoverage` class and OpenCover reports)

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/CoverageTest.java
```diff
@@ -161,47 +161,83 @@ public void givenSingleBranchPointPerLine_getBranchCoverage_returnsEmpty() {
...
+
+    // Identical branch points are merged
+    sut.add(new BranchPoint(filePath, 1, 1, 2, 0, 1));
+    sut.add(new BranchPoint(filePath, 1, 1, 2, 0, 1));
+
+    // Branch points with different line are not merged
+    sut.add(new BranchPoint(filePath, 2, 1, 2, 0, 1));
+    sut.add(new BranchPoint(filePath, 3, 1, 2, 0, 1));
+
+    // Branch points with different offset are not merged
```
andrei-epure-sonarsource, line 184: uhm, I see below that lines `4` , `5` and `6` do get merged (and the comment says "are not merged")

andrei-epure-sonarsource: `// Branch points with different offset are correctly aggregated`

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/BranchPoint.java
andrei-epure-sonarsource: `// Identifier for code path`

PR #3303, opened 2020-04-27
Remove custom coverage for dotCover, OpenCover, VS ccov

### LINKED ISSUE #3296: Fix Code coverage import from multiple test projects

costin-zaharia-sonarsource:
There are multiple problems with it:
- it is a work around meant to solve the coverage of properties
with get and set on the same line but we received feedback from
community that this is not wanted in all the cases (e.g. properties
of serialized types)
- it is not computed correctly if a code file is covered by tests from
different projects
- it can be confusing since it will contradict the information from
the coverage file generated by the tool

If we decide in the future that we still want to have this, there should
be an option to enable/disable this (probably disabled by default)

Fixes #3296


andrei-epure-sonarsource:
I mean the UTs for OpenCover, dotCover and VS cover reports

costin-zaharia-sonarsource:
> I wouldn't remove the tests, I would just update them
> 
> The tests are valuable to see the behavior, and for example for VS ccoverage, you haven't reverted [all the changes](https://github.com/SonarSource/sonar-dotnet/pull/3251/files#diff-733c1184b558489d17812fa57e057ef3) we did in the branch coverage sprint, so having the tests would still be useful to see the behavior

I've added back the vstest tests and reverted the VisualStudioCoverageXmlReportParser completely. For dotCover I've just removed the calls done to publish coverage.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/Coverage.java
```diff
@@ -34,7 +33,6 @@
   private static final int GROW_FACTOR = 2;
   private static final int SPECIAL_HITS_NON_EXECUTABLE = -1;
   private final Map<String, int[]> hitsByLineAndFile = new HashMap<>();
-  private final Map<String, List<BranchCoverage>> branchCoverageByFile = new HashMap<>();
   private final List<BranchPoint> branchPoints = new ArrayList<>();
```
andrei-epure-sonarsource, line 36: is this needed?

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/CoverageReportImportSensorTest.java
```diff
@@ -136,8 +136,6 @@ public void analyze() throws Exception {
     SensorContextTester context = computeCoverageMeasures(false);
     assertThat(context.lineHits("foo:Foo.cs", 2)).isEqualTo(1);
     assertThat(context.lineHits("foo:Foo.cs", 4)).isEqualTo(0);
-    assertThat(context.coveredConditions("foo:Foo.cs", 1)).isEqualTo(2);
```
andrei-epure-sonarsource, line 139: do you have to remove these lines, or can you just update them?

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/Coverage.java
costin-zaharia-sonarsource: I'm not sure I understand what you mean. In this case the variable was not used anymore since we no longer push branch coverage this way so it can be removed.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/CoverageReportImportSensorTest.java
costin-zaharia-sonarsource: I've added the lines back and updated the expected values to null

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/Coverage.java
andrei-epure-sonarsource: exactly - so I think you can remove what's not used anymore, unless you reckon it's going to be used in the next PR

andrei-epure-sonarsource: my comment is on this line
```
   private final List<BranchPoint> branchPoints = new ArrayList<>(); 
```

costin-zaharia-sonarsource: We discussed directly, but I will add a comment also here to not forget about it. This is needed in order to properly compute the imported branch coverage.

andrei-epure-sonarsource:
there was probably a glitch with the github notification, but all jobs passed and binaries got updated ... so I merged the PR

PR #3315, opened 2020-05-01
Update rspec for S3011

### LINKED ISSUE #3285: Change issue S3011 type to code smell
OOP visibility/accessibility is likely more a code quality subject than security thus [S3011](https://jira.sonarsource.com/browse/RSPEC-3011) should live as a code smell

Related to:
https://github.com/SonarSource/sonar-dotnet/issues/1983

costin-zaharia-sonarsource:
This will update the rule type to Code Smell and severity to Major.

Fixes #3285


costin-zaharia-sonarsource:
> `CsRuleTypeMapping` and `VbRuleTypeMapping` should be updated to reflect new rule type for S3011.
> I am really surprised also that tests `RuleTypeTests.DetectRuleTypeChanges_CS` and `RuleTypeTests.DetectRuleTypeChanges_VB` didn't fail. They are supposed to serve as failsafes, which they didn't in that case

Nice catch. I think something is wrong now with our build because both these tests are failing on my local machine but but the CI build doesn't.

christophe-zurn-sonarsource:
PR LGTM, but it needs to be rebased on newest master to ensure everything works as expected

PR #3317, opened 2020-05-01
Add assembly name to S3904, S3990, and S3992 messages

### LINKED ISSUE #3219: Add assembly name to S3904 message
### Description

As reported in this internal [discuss thread](https://discuss.sonarsource.com/t/could-s3904-be-more-helpful/4320/4), assembly name is not part of the reported message.

That's fine when analyzing the project and seeing build output per project.
That's not fine when issues are collected for SQ/SC for solutions with large number of projects.

Message should be updated to
` Provide an 'AssemblyVersion' attribute for assembly 'xxx'.`

christophe-zurn-sonarsource:
Fixes #3219

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MarkAssemblyWithAssemblyVersionAttribute.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MarkAssemblyWithAssemblyVersionAttribute.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
@@ -31,7 +31,7 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class MarkAssemblyWithAssemblyVersionAttribute : MarkAssemblyWithAttributeBase
     {
         internal const string DiagnosticId = "S3904";
-        private const string MessageFormat = "Provide an 'AssemblyVersion' attribute for this assembly.";
+        private const string MessageFormat = "Provide an 'AssemblyVersion' attribute for assembly '{0}'.";
 
         private static readonly DiagnosticDescriptor rule =
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MarkAssemblyWithAttributeBase.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MarkAssemblyWithAttributeBase.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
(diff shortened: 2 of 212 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MarkAssemblyWithAssemblyVersionAttribute.cs
```diff
@@ -31,7 +31,7 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class MarkAssemblyWithAssemblyVersionAttribute : MarkAssemblyWithAttributeBase
```
costin-zaharia-sonarsource, line 31: I've noticed that there is also `MarkAssemblyWithComVisibleAttribute` and `MarkAssemblyWithClsCompliantAttribute` which refer to the assembly with this "Mark **this** assembly with ".

What to you think about updating also the remaining rules?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MarkAssemblyWithAttributeBase.cs
```diff
@@ -39,7 +39,7 @@ protected sealed override void Initialize(SonarAnalysisContext context)
                             var requiredAttributeFound = cc.Compilation.Assembly.GetAttributes(AttributeToFind).Any();
                             if (!requiredAttributeFound)
                             {
-                                cc.ReportDiagnosticWhenActive(Diagnostic.Create(SupportedDiagnostics[0], null));
+                                cc.ReportDiagnosticWhenActive(Diagnostic.Create(SupportedDiagnostics[0], null, cc.Compilation.AssemblyName));
```
costin-zaharia-sonarsource, line 42: According to [doc](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.compilation.assemblyname?view=roslyn-dotnet), `Compilation.AssemblyName` can be also null. 

Could we handle also this case?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MarkAssemblyWithAssemblyVersionAttribute.cs
christophe-zurn-sonarsource: I thought about it, but wasn't sure whether to do it in this PR or not. But I guess it makes sense. I updated PR to also update these rules message

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MarkAssemblyWithAttributeBase.cs
christophe-zurn-sonarsource: I don't think I can force the assemblyName to null in our test framework. If we can, I don't know how, and I'm not sure it's worth investing time into this, since this parameter is used in a formatting string, so even if it is null, the formatted string will show empty string `''` instead of the assembly name, which IMO is acceptable given the rare case (if any) it should happen

pavel-mikula-sonarsource:
Message parameter of RSPECs in Jira should be updated as well.

PR #3319, opened 2020-05-01
Ensure that OpenCover returns the exit code from test runner

costin-zaharia-sonarsource:

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -222,7 +222,7 @@ stages:
 
       - powershell: |
           cd sonaranalyzer-dotnet\tests
-          $(NUGET_PACKAGES)\opencover\4.7.922\tools\OpenCover.Console.exe -target:"C:/Program Files/dotnet/dotnet.exe" -targetargs:"test SonarAnalyzer.UnitTest\bin\$(BuildConfiguration)\net48\SonarAnalyzer.UnitTest.dll" -filter:"+[SonarAnalyzer.CFG]* +[SonarAnalyzer]* +[SonarAnalyzer.CSharp]* +[SonarAnalyzer.Utilities]* +[SonarAnalyzer.VisualBasic]*" -output:coverage.xml -oldStyle -register:Path32
+          $(NUGET_PACKAGES)\opencover\4.7.922\tools\OpenCover.Console.exe -target:"C:/Program Files/dotnet/dotnet.exe" -targetargs:"test SonarAnalyzer.UnitTest\bin\$(BuildConfiguration)\net48\SonarAnalyzer.UnitTest.dll" -returntargetcode -filter:"+[SonarAnalyzer.CFG]* +[SonarAnalyzer]* +[SonarAnalyzer.CSharp]* +[SonarAnalyzer.Utilities]* +[SonarAnalyzer.VisualBasic]*" -output:coverage.xml -oldStyle -register:Path32
           cd ..
           cd ..
         displayName: 'Run .Net unit tests'
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/XmlExternalEntityShouldNotBeParsedTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/XmlExternalEntityShouldNotBeParsedTest.cs
@@ -99,8 +99,8 @@ namespace SonarAnalyzer.UnitTest.Rules
             Verifier.VerifyAnalyzer(
                 new[]
                 {
-                    @"TestCases\XmlExternalEntityShouldNotBeParsed_XmlReader_ParameterProvider.cs",
-                    @"TestCases\XmlExternalEntityShouldNotBeParsed_XmlReader_ExternalParameter.cs"
+                    @"TestCases\XmlExternalEntityShouldNotBeParsed_XmlReader_ExternalParameter.cs",
+                    @"TestCases\XmlExternalEntityShouldNotBeParsed_XmlReader_ParameterProvider.cs"
                 },
                 new XmlExternalEntityShouldNotBeParsed(GetVersionProviderMock(NetFrameworkVersion.After452)),
                 additionalReferences: FrameworkMetadataReference.SystemXml);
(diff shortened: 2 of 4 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/XmlExternalEntityShouldNotBeParsedTest.cs
```diff
@@ -99,8 +99,8 @@ public void XmlExternalEntityShouldNotBeParsed_NoCrashOnExternalParameterUse()
             Verifier.VerifyAnalyzer(
                 new[]
                 {
-                    @"TestCases\XmlExternalEntityShouldNotBeParsed_XmlReader_ParameterProvider.cs",
-                    @"TestCases\XmlExternalEntityShouldNotBeParsed_XmlReader_ExternalParameter.cs"
+                    @"TestCases\XmlExternalEntityShouldNotBeParsed_XmlReader_ExternalParameter.cs",
+                    @"TestCases\XmlExternalEntityShouldNotBeParsed_XmlReader_ParameterProvider.cs"
```
costin-zaharia-sonarsource, line 103: As far as I can tell we have a bug in the `DiagnosticVerifier`. It seems to check the expected issues only in the first file.

azure-pipelines.yml
```diff
@@ -222,7 +222,7 @@ stages:
 
       - powershell: |
           cd sonaranalyzer-dotnet\tests
-          $(NUGET_PACKAGES)\opencover\4.7.922\tools\OpenCover.Console.exe -target:"C:/Program Files/dotnet/dotnet.exe" -targetargs:"test SonarAnalyzer.UnitTest\bin\$(BuildConfiguration)\net48\SonarAnalyzer.UnitTest.dll" -filter:"+[SonarAnalyzer.CFG]* +[SonarAnalyzer]* +[SonarAnalyzer.CSharp]* +[SonarAnalyzer.Utilities]* +[SonarAnalyzer.VisualBasic]*" -output:coverage.xml -oldStyle -register:Path32
+          $(NUGET_PACKAGES)\opencover\4.7.922\tools\OpenCover.Console.exe -target:"C:/Program Files/dotnet/dotnet.exe" -targetargs:"test SonarAnalyzer.UnitTest\bin\$(BuildConfiguration)\net48\SonarAnalyzer.UnitTest.dll" -returntargetcode -filter:"+[SonarAnalyzer.CFG]* +[SonarAnalyzer]* +[SonarAnalyzer.CSharp]* +[SonarAnalyzer.Utilities]* +[SonarAnalyzer.VisualBasic]*" -output:coverage.xml -oldStyle -register:Path32
```
costin-zaharia-sonarsource, line 225: This is the actual fix. OpenCover will return the exit code from the target tool only if `-returntargetcode` parameter is provided.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodParameterUnused.CSharp8.cs
```diff
@@ -94,7 +94,7 @@ public void DoSomething(int a, int b)
     // https://github.com/SonarSource/sonar-dotnet/issues/3255
```
christophe-zurn-sonarsource, line 94: Is this repro still valid since the FP appears to be wrong ? Should the corresponding issue be closed since the FP is not happening anymore ? If yes, do we know when this was fixed

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/XmlExternalEntityShouldNotBeParsedTest.cs
christophe-zurn-sonarsource: Not sure I understand this comment related to this change. And if we have a bug, do we have ticket to track it ?

costin-zaharia-sonarsource: I took a better look this morning and I don't think that it's a bug anymore, just a partial implementation :).

Currently the location does dot know about the file: https://github.com/SonarSource/sonar-dotnet/blob/master/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/IIssueLocation.cs

and the **DiagnosticVerifier** will check only in the first file for issue locations: https://github.com/SonarSource/sonar-dotnet/blob/master/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs#L56

Please note that the first syntax tree (the one that is skipped) is for "ExtraEmptyFile.g.cs" which is added to the project by default: https://github.com/SonarSource/sonar-dotnet/blob/master/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs#L81

Probably checking for issues in multiple files was never needed.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodParameterUnused.CSharp8.cs
costin-zaharia-sonarsource: The issue was added recently and I think that we didn't see that the reproducer didn't worked due to the build which was always passing.

I've talked with @pavel-mikula-sonarsource on Friday (since he added the reproducer) and he will take a look and close the issue or ask for more details in the community thread to update it.

christophe-zurn-sonarsource: Ok, thanks !

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/XmlExternalEntityShouldNotBeParsedTest.cs
christophe-zurn-sonarsource: Ok. Makes sense now.
I think it might be worth it to add a comment for method declaration `public static void VerifyAnalyzer(IEnumerable<string> paths,` making it explicit that only the 1st file will be checked by the verifier. After looking in the code, I indeed remember now that these extra paths are used for additional file dependency. Since it's not something that we use that often, making it clear might help in the future

Issue #3304, opened 2020-04-27
Fix S1854 FP: ref keyword on a local variable


volgunin:
### Description

When a local variable with ref keyword is used, S1854 is raised.

### Repro steps

```
private void Test(int[] values)
{
    ref int value = ref values[0];
    value = default;
}
```

### Expected behavior

No warning.

### Actual behavior

warning S1854: Remove this useless assignment to local variable 'value'.

### Related information

SonarAnalyzer.CSharp.8.6.0.16497
VisualStudio 2019 16.5.4

PR #3323, opened 2020-05-05
Add FP repro for S1854

pavel-mikula-sonarsource:
Reproduces #3304

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
@@ -899,7 +899,7 @@ namespace Tests.Diagnostics
     {
         public void WithRefKeyword(int[] values)
         {
-            ref int value = ref values[0]; // Noncompliant FP because there exists write that is propagated
+            ref int value = ref values[0]; // Noncompliant FP because `value` keeps the reference to `values[0]`, and below `default` is actually assigned to `values[0]`
             value = default; // Noncompliant FP because it's ref variable and value is propagated somewhere
         }
     }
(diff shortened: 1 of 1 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
```diff
@@ -892,6 +892,15 @@ string Local()
...
             }
         }
+    }
 
+    // https://github.com/SonarSource/sonar-dotnet/issues/3304
+    public class Repro_3304
+    {
+        public void WithRefKeyword(int[] values)
+        {
+            ref int value = ref values[0]; // Noncompliant FP because there exists write that is propagated
```
andrei-epure-sonarsource, line 902:
```suggestion
            ref int value = ref values[0]; // Noncompliant FP because `value` keeps the reference to `values[0]`, and below `default` is actually assigned to `values[0]`
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
```diff
@@ -892,6 +892,15 @@ string Local()
...
+            value = default; // Noncompliant FP because it's ref variable and value is propagated somewhere
```
andrei-epure-sonarsource, line 903:
```suggestion
            value = default; // Noncompliant FP (see above comment)
```

pavel-mikula-sonarsource: I'd keep the original text, because there are two different FPs with two different reasoning behind them. I took some time to understand the difference so I'd like to help our future ourselves.

PR #3326, opened 2020-05-06
Allow import of internal issues with invalid location

### LINKED ISSUE #2484: Roslyn external issues can contain invalid locations (again)
We are testing SonarQube 7.9 with SonarC# 7.15 (build 8572) and a StyleCop.Analyzers plugin (which we build using the SonarQube Roslyn SDK).

Rule [SA1629](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1629.md) (“Documentation text should end with period”) leads to an error, the scan is not imported and the build job fails.

Example:
```Console
ERROR: 43 is not a valid line offset for pointer. File Collections/IStatisticHeap.cs has 42 character(s) at line 72
```

The analyzer reports the issue one position after the last character of that line. In Visual Studio a squiggly line is shown right next to the last character of that line, however, SonarQube right now considers this an invalid line offset.

This issue was already reported and fixed with #2192, but it looks like it's back in SonarC# 7.15.

(See also [this post](https://community.sonarsource.com/t/stylecopanalyzers-c-rule-sa1629-leads-to-not-a-valid-line-offset-for-pointer-and-breaks-import-of-scan-results/11876) in the SonarSource community).

costin-zaharia-sonarsource:
Issues from third-party Roslyn analyzers (including Roslyn analyzers provided by Microsoft) are included in the MSBuild output and imported by default into SonarQube.

In this case they are considered external issues and the Roslyn importer is very forgiving regarding the location:
- if the column is invalid it will use only the line
- if both column and line are invalid, it will report the issue on the file.

On the other hand, if the [SDK for SonarQube Roslyn Analyzer Plugins](https://github.com/SonarSource/sonarqube-roslyn-sdk) is used to create a plugin for the analyzers, then the issues are treated as internal and the importer will not fallback to line or file level.

This commit is changing that and allows the import of internal issues (excepting the ones from `csharpsquid` and `vbnet` repositories) with invalid location.

Fixes #2484

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/SarifParserCallbackImpl.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/SarifParserCallbackImpl.java
@@ -162,10 +162,10 @@ public class SarifParserCallbackImpl implements SarifParserCallback {
 
   private void createExternalIssue(InputFile inputFile, String ruleId, @Nullable String level, Location primaryLocation, Collection<Location> secondaryLocations) {
     NewExternalIssue newIssue = newExternalIssue(ruleId);
-    newIssue.at(createPrimaryLocation(inputFile, primaryLocation, newIssue::newLocation, true));
+    newIssue.at(createPrimaryLocation(inputFile, primaryLocation, newIssue::newLocation));
     setExternalIssueSeverityAndType(ruleId, level, newIssue);
 
-    populateSecondaryLocations(secondaryLocations, newIssue::newLocation, newIssue::addLocation, true);
+    populateSecondaryLocations(secondaryLocations, newIssue::newLocation, newIssue::addLocation);
 
     newIssue.save();
   }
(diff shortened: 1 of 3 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/SarifParserCallbackImpl.java
```diff
@@ -235,11 +232,6 @@ private static NewIssueLocation createIssueLocation(InputFile inputFile, Locatio
...
     } catch (IllegalArgumentException ex1) {
 
-      if (!isLocationResilient) {
-        // Our rules should fail if they report on an invalid location
-        throw ex1;
-      }
-
       try {
         // Precise location failed, now try the line...
         newIssueLocation = newIssueLocation.at(inputFile.selectLine(location.getStartLine()));
```
christophe-zurn-sonarsource, line 237: It might be good to add some `debug` level log inside the catch clause. It might help investigation in the future if a user has issues reported not on a precise location

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/SarifParserCallbackImpl.java
```diff
@@ -235,11 +232,6 @@ private static NewIssueLocation createIssueLocation(InputFile inputFile, Locatio
 
     } catch (IllegalArgumentException ex1) {
 
-      if (!isLocationResilient) {
-        // Our rules should fail if they report on an invalid location
-        throw ex1;
-      }
-
```
christophe-zurn-sonarsource, line 242: I think it'd be good to still throw when it's our own rules (from `csharpsquid` repo), since its ours, we know that their location should be correct, and it's nice to fail in that case, to detect as early as possible any issue in how we're handling location

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/SarifParserCallbackImpl.java
```diff
@@ -192,12 +194,14 @@ private NewExternalIssue newExternalIssue(String ruleId) {
   }
 
   private void createIssue(InputFile inputFile, String ruleId, Location primaryLocation, Collection<Location> secondaryLocations, String repositoryKey) {
+    boolean isOwnRepository = isOwnRepository(repositoryKey);
```
christophe-zurn-sonarsource, line 197: _minor_ : for clarity, I'd suggest to rename `own`, although I leave it to you if you think it's clear enough
```suggestion
    boolean isSonarsourceRepository = isSonarsourceRepository(repositoryKey);
```

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/SarifParserCallbackImpl.java
```diff
@@ -53,6 +54,7 @@
   private static final Logger LOG = Loggers.get(SarifParserCallbackImpl.class);
 
   private static final String EXTERNAL_ENGINE_ID = "roslyn";
+  private static final ImmutableList<String> OWN_REPOSITORIES = ImmutableList.of("csharpsquid", "vbnet");
```
christophe-zurn-sonarsource, line 57: _minor_ : as in c#, it's better to use the top level interface for declaration instead of the implementing class
```suggestion
  private static final List<String> OWN_REPOSITORIES = ImmutableList.of("csharpsquid", "vbnet");
```

costin-zaharia-sonarsource: I had to change this in the end to `List<String> OWN_REPOSITORIES =  Arrays.asList("csharpsquid", "vbnet")` because the integration tests were failing due to `java.lang.NoClassDefFoundError: org/sonar/api/internal/google/common/collect/ImmutableList`.

PR #3337, opened 2020-05-12
S3453: Fix FP with SafeHandle

### LINKED ISSUE #3329: S3453: FP with SafeHandle
### Description

When creating a custom SafeHandle, it is possible to only have a private constructor. This SafeHandle still can be instantiated by the PInvoke Marshaler, as shown in the snippet below.

### Repro steps

```csharp
public class MySafeHandle : SafeHandleZeroOrMinusOneIsInvalid
{
    private MySafeHandle() : base(true)
    {
    }

    protected override bool ReleaseHandle()
    {
    }
}

public static class NativeMethods
{
    [DllImport("Some.dll")]
    private static extern MySafeHandle CreateSafeHandle();
}
```

### Expected behavior

No warning.

### Actual behavior

Warning on `MySafeHandle`: S3453: This class can't be instantiated; make its constructor 'public'.

### Related information

* C# Plugins Version 8.6.1.17183
* Visual Studio Version 16.5.4
* dotnet version .NET Core 3.1
* Operating System Windows 10

ThorstenReichert:
Fixes #3329

Added check for classes derived from SafeHandle to not flag S3453.

Note: I based this PR on the Repro cases to be added with #3336. Let me know if that's not desirable and I will remove the test cases from this PR again.

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ClassNotInstantiatableBase.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ClassNotInstantiatableBase.cs
@@ -91,9 +91,9 @@ namespace SonarAnalyzer.Rules
             && !HasNonPrivateConstructor(constructors)
             && constructors.All(c => !c.GetAttributes().Any());
 
-        protected static bool DerivesFromSafeHandle(INamedTypeSymbol namedType)
+        protected static bool DerivesFromSafeHandle(ITypeSymbol typeSymbol)
         {
-            return namedType.DerivesFrom(KnownType.System_Runtime_InteropServices_SafeHandle);
+            return typeSymbol.DerivesFrom(KnownType.System_Runtime_InteropServices_SafeHandle);
         }
     }
 }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ClassNotInstantiatableBase.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ClassNotInstantiatableBase.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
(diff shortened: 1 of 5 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ClassNotInstantiatable.vb
```diff
@@ -99,4 +99,23 @@ Namespace Tests.Diagnostics
         Private Sub New()
         End Sub
     End Class
+
+    ' https//github.com/SonarSource/sonar-dotnet/issues/3329
+    Public Class Repro_3329
```
pavel-mikula-sonarsource, line 104: Preserve the original comment here please to explain why is this case compliant (except the Non+FP part)
```suggestion
    Public Class Repro_3329 `Compliant, instance will be created by PInvoke of DllImport
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ClassNotInstantiatable.cs
```diff
@@ -94,4 +95,13 @@ private WithAttribute2()
         {
         }
     }
+
+    // https://github.com/SonarSource/sonar-dotnet/issues/3329
+    public class Repro_3329 : System.Runtime.InteropServices.SafeHandle
```
pavel-mikula-sonarsource, line 100: Preserve the original comment here please to explain why is this case compliant (except the Non+FP part)
```suggestion
    public class Repro_3329 : System.Runtime.InteropServices.SafeHandle 'Compliant, instance will be created by PInvoke of DllImport
```

pavel-mikula-sonarsource:
Ok, I see what happened. My repro-commit was not on master yet, so we have a merge conflict now.

Please rebase on current master and that will solve it.

ThorstenReichert:
@pavel-mikula-sonarsource, I think the commit history looks fine now. Comments have been added as per suggestion.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/ClassNotInstantiatableBase.cs
```diff
@@ -90,5 +90,10 @@ protected static bool IsAnyConstructorCalled
             constructors.Any()
             && !HasNonPrivateConstructor(constructors)
             && constructors.All(c => !c.GetAttributes().Any());
+
+        protected static bool DerivesFromSafeHandle(INamedTypeSymbol namedType)
+        {
+            return namedType.DerivesFrom(KnownType.System_Runtime_InteropServices_SafeHandle);
```
pavel-mikula-sonarsource, line 96:
```suggestion
        protected static bool DerivesFromSafeHandle(ITypeSymbol typeSymbol)
        {
            return typeSymbol.DerivesFrom(KnownType.System_Runtime_InteropServices_SafeHandle);
```

PR #3338, opened 2020-05-13
Reuse symbolic execution logic in EmptyNullableValueAccess and ObjectsShouldNotBeDisposedMoreThanOnce

costin-zaharia-sonarsource:
This commit introduces `SymbolicExecutionRunner` which allows the reuse of the generated exploded graph and walking it one single time for all the symbolic execution rules.

As a proof of concept, `SymbolicExecutionRunner` is now executing the `EmptyNullableValueAccess` and `ObjectsShouldNotBeDisposedMoreThanOnce` analyzers which are refactored to implement `ISymbolicExecutionAnalyzer` instead of `SonarDiagnosticAnalyzer`.

With this refactoring the overall analysis time drops ~6%.

Fixes: #3341

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/EmptyNullableValueAccess.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/EmptyNullableValueAccess.cs
@@ -21,63 +21,30 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
-using Microsoft.CodeAnalysis.Diagnostics;
-using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.Rules.SymbolicExecution;
 using SonarAnalyzer.SymbolicExecution;
 using SonarAnalyzer.SymbolicExecution.Constraints;
 using SonarAnalyzer.SymbolicExecution.SymbolicValues;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
-    [DiagnosticAnalyzer(LanguageNames.CSharp)]
-    [Rule(DiagnosticId)]
-    public sealed class EmptyNullableValueAccess : SonarDiagnosticAnalyzer
+    internal sealed class EmptyNullableValueAccess : ISymbolicExecutionAnalyzer
     {
         internal const string DiagnosticId = "S3655";
         private const string MessageFormat = "'{0}' is null on at least one execution path.";
-
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
-
         private const string ValueLiteral = "Value";
         private const string HasValueLiteral = "HasValue";
 
-        protected override void Initialize(SonarAnalysisContext context)
-        {
-            context.RegisterExplodedGraphBasedAnalysis((e, c) => CheckEmptyNullableAccess(e, c));
-        }
+        private static readonly DiagnosticDescriptor rule =
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        public IEnumerable<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
 
-        private static void CheckEmptyNullableAccess(CSharpExplodedGraph explodedGraph, SyntaxNodeAnalysisContext context)
-        {
-            var nullPointerCheck = new NullValueAccessedCheck(explodedGraph);
-            explodedGraph.AddExplodedGraphCheck(nullPointerCheck);
-
-            var nullIdentifiers = new HashSet<IdentifierNameSyntax>();
-
-            void nullValueAccessedHandler(object sender, MemberAccessedEventArgs args) => nullIdentifiers.Add(args.Identifier);
-
-            nullPointerCheck.ValuePropertyAccessed += nullValueAccessedHandler;
-
-            try
-            {
-                explodedGraph.Walk();
-            }
-            finally
-            {
-                nullPointerCheck.ValuePropertyAccessed -= nullValueAccessedHandler;
-            }
-
-            foreach (var nullIdentifier in nullIdentifiers)
-            {
-                context.ReportDiagnosticWhenActive(Diagnostic.Create(rule, nullIdentifier.Parent.GetLocation(), nullIdentifier.Identifier.ValueText));
-            }
-        }
+        public ISymbolicExecutionAnalysisContext AddChecks(CSharpExplodedGraph explodedGraph) => new AnalysisContext(explodedGraph);
 
         internal sealed class NullValueAccessedCheck : ExplodedGraphCheck
         {
@@ -153,7 +120,7 @@ namespace SonarAnalyzer.Rules.CSharp
             }
         }
 
-        internal sealed class HasValueAccessSymbolicValue : MemberAccessSymbolicValue
+        private sealed class HasValueAccessSymbolicValue : MemberAccessSymbolicValue
         {
             public HasValueAccessSymbolicValue(SymbolicValue nullable)
                 : base(nullable, HasValueLiteral)
@@ -173,6 +140,27 @@ namespace SonarAnalyzer.Rules.CSharp
 
                 return MemberExpression.TrySetConstraint(nullabilityConstraint, programState);
             }
+        }
+
+        private class AnalysisContext : ISymbolicExecutionAnalysisContext
+        {
+            private readonly HashSet<IdentifierNameSyntax> nullIdentifiers = new HashSet<IdentifierNameSyntax>();
+            private readonly NullValueAccessedCheck nullPointerCheck;
+
+            public IEnumerable<Diagnostic> GetDiagnostics() =>
+                this.nullIdentifiers.Select(nullIdentifier => Diagnostic.Create(rule, nullIdentifier.Parent.GetLocation(), nullIdentifier.Identifier.ValueText));
+
+            public AnalysisContext(CSharpExplodedGraph explodedGraph)
+            {
+                this.nullPointerCheck = new NullValueAccessedCheck(explodedGraph);
+                this.nullPointerCheck.ValuePropertyAccessed += AddIdentifier;
+
+                explodedGraph.AddExplodedGraphCheck(this.nullPointerCheck);
+            }
+
+            private void AddIdentifier(object sender, MemberAccessedEventArgs args) => this.nullIdentifiers.Add(args.Identifier);
+
+            public void Dispose() => this.nullPointerCheck.ValuePropertyAccessed -= AddIdentifier;
         }
     }
 }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/EmptyNullableValueAccess.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/EmptyNullableValueAccess.cs
@@ -142,7 +142,7 @@ namespace SonarAnalyzer.Rules.CSharp
             }
         }
 
-        private class AnalysisContext : ISymbolicExecutionAnalysisContext
+        private sealed class AnalysisContext : ISymbolicExecutionAnalysisContext
         {
             private readonly HashSet<IdentifierNameSyntax> nullIdentifiers = new HashSet<IdentifierNameSyntax>();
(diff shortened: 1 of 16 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
```diff
@@ -0,0 +1,77 @@
...
+using SonarAnalyzer.SymbolicExecution;
+
+namespace SonarAnalyzer.Rules.SymbolicExecution
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(EmptyNullableValueAccess.DiagnosticId)]
+    [Rule(ObjectsShouldNotBeDisposedMoreThanOnce.DiagnosticId)]
+    public sealed class SymbolicExecutionRunner : SonarDiagnosticAnalyzer
+    {
+        private readonly SymbolicExecutionAnalyzerFactory symbolicExecutionAnalyzerFactory = new SymbolicExecutionAnalyzerFactory();
```
andrei-epure-sonarsource, line 39: the rules are registered here at lines 35-36 and also in `SymbolicExecutionAnalyzerFactory`

to improve maintainability, I suggest to keep the rules only in this class and pass them to the constructor of `SymbolicExecutionAnalyzerFactory` - and maybe rename that class to something to suggest that is checks if the rules are enabled or not - maybe `EnabledRuleFilter` (it's not really used as a factory)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        private void Analyze(CSharpExplodedGraph explodedGraph, SyntaxNodeAnalysisContext context)
+        {
+            var analyzerContexts = InitializeAnalyzers(explodedGraph, context).ToList();
+
+            try
+            {
+                explodedGraph.Walk();
+            }
+            finally
```
andrei-epure-sonarsource, line 61: why are you using finally here?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/SymbolicExecution/SymbolicExecutionAnalyzerFactoryTest.cs
```diff
@@ -0,0 +1,109 @@
...
+
+namespace SonarAnalyzer.UnitTest.Rules.SymbolicExecution
+{
+    [TestClass]
+    public class SymbolicExecutionAnalyzerFactoryTest
+    {
+        [TestMethod]
+        public void SupportedDiagnostics_ReturnsSymbolicExecutionRuleDescriptors()
+        {
+            // Currently only EmptyNullableValueAccess was updated to use the new analyzer.
```
andrei-epure-sonarsource, line 39: remove comment?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/SymbolicExecution/SymbolicExecutionAnalyzerFactoryTest.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            var sut = new SymbolicExecutionAnalyzerFactory();
+            var supportedDiagnostics = sut.SupportedDiagnostics.Select(descriptor => descriptor.Id).ToList();
+
+            CollectionAssert.AreEquivalent(supportedDiagnostics, new[] {"S3655", "S3966"});
```
andrei-epure-sonarsource, line 44: maybe store in class constants the rule ids (S3655) with the name of the actual rule? it would make the tests more readable

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/EmptyNullableValueAccess.cs
```diff
@@ -174,5 +141,26 @@ public override IEnumerable<ProgramState> TrySetConstraint(SymbolicValueConstrai
                 return MemberExpression.TrySetConstraint(nullabilityConstraint, programState);
             }
         }
+
+        private sealed class AnalysisContext : ISymbolicExecutionAnalysisContext
```
andrei-epure-sonarsource, line 145: you should move this class declaration above `HasValueAccessSymbolicValue` and `NullValueAccessedCheck`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/EmptyNullableValueAccess.cs
```diff
@@ -174,5 +141,26 @@ public override IEnumerable<ProgramState> TrySetConstraint(SymbolicValueConstrai
...
+        {
+            private readonly HashSet<IdentifierNameSyntax> nullIdentifiers = new HashSet<IdentifierNameSyntax>();
+            private readonly NullValueAccessedCheck nullPointerCheck;
+
+            public IEnumerable<Diagnostic> GetDiagnostics() =>
+                this.nullIdentifiers.Select(nullIdentifier => Diagnostic.Create(rule, nullIdentifier.Parent.GetLocation(), nullIdentifier.Identifier.ValueText));
+
+            public AnalysisContext(CSharpExplodedGraph explodedGraph)
+            {
+                this.nullPointerCheck = new NullValueAccessedCheck(explodedGraph);
```
andrei-epure-sonarsource, line 155: nitpick: do you need `this` here?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
andrei-epure-sonarsource: actually I saw this is how it was used in the old rules, so it's good to keep the behavior of eating analysis exceptions

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionAnalyzerFactory.cs
```diff
@@ -0,0 +1,57 @@
...
+        public IEnumerable<ISymbolicExecutionAnalyzer> GetEnabledAnalyzers(SyntaxNodeAnalysisContext context) =>
+            // Enabled analyzers can be changed at runtime in the IDE or by using connected mode in SonarLint and because of this
+            // they cannot be cached.
+            this.analyzers.Where(analyzer => HasEnabledDiagnostics(analyzer, context.Compilation.Options));
+
+        private static bool HasEnabledDiagnostics(ISymbolicExecutionAnalyzer analyzer, CompilationOptions options) =>
+            analyzer.SupportedDiagnostics.Any(diagnosticDescriptor => IsEnabled(options, diagnosticDescriptor));
+
+        private static bool IsEnabled(CompilationOptions options, DiagnosticDescriptor diagnosticDescriptor) =>
+            diagnosticDescriptor.GetEffectiveSeverity(options) != ReportDiagnostic.Suppress;
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Yes, I've tested it with S4MSB and I get the same results as with sonar-dotnet 8.7

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/EmptyNullableValueAccess.cs
costin-zaharia-sonarsource: No. I forgot to rebase and remove these after #3350 was merged.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
costin-zaharia-sonarsource: The goal was to decouple the `SymbolicExecutionRunner` from the rules and let the factory create the analyzers, provide the supported rules and tell which are the ones enabled.

The `Rule` attributes on top are used only by the `RuleDescriptorGenerator` to generate the html description files. I cannot move yet those since the `RuleDetailBuilder.GetAllRuleDetails` looks only at the types which implement `DiagnosticAnalyser` to retrieve the rule ids: https://github.com/SonarSource/sonar-dotnet/blob/master/sonaranalyzer-dotnet/src/SonarAnalyzer.Utilities/RuleFinder.cs#L47.

I am more inclined to extend the `RuleFinder` to check also additional types and keep the responsibilities separated. What do you think?

costin-zaharia-sonarsource: I've moved the Rule attributes in https://github.com/SonarSource/sonar-dotnet/pull/3338/commits/c79baa98e7fd7530d308ec45d9c5b36eb31a8e4d

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Constants.cs
```diff
@@ -24,5 +24,8 @@ internal static class Constants
     {
         internal const string NuGetLatestVersion = "LATEST";
         internal const string DotNetCore220Version = "2.2.0";
+
+        internal const string EmptyNullableValueAccess = "S3655";
```
andrei-epure-sonarsource, line 28: I think these can stay just inside `SymbolicExecutionAnalyzerFactoryTest.cs`, no need to make them internal

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
andrei-epure-sonarsource: sounds good

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Constants.cs
costin-zaharia-sonarsource: Indeed. We can extract them later if needed.

sonaranalyzer-dotnet/src/SonarAnalyzer.Utilities/RuleFinder.cs
```diff
@@ -78,28 +85,25 @@ public static IEnumerable<Type> GetUtilityAnalyzerTypes(AnalyzerLanguage languag
...
+            });
+        }
+
+        private static IEnumerable<string> GetLanguages(MemberInfo analyzerType)
+        {
+            var diagnosticAttribute = analyzerType.GetCustomAttributes<DiagnosticAnalyzerAttribute>().FirstOrDefault();
 
-            return language;
+            return diagnosticAttribute == null
+                ? analyzerType.GetCustomAttributes<RuleAttribute>().FirstOrDefault()?.Languages
```
andrei-epure-sonarsource, line 104: do you want to throw here? in case the language is not provided? and add tests for this case?

it should be caught by unit tests

sonaranalyzer-dotnet/src/SonarAnalyzer.Utilities/RuleFinder.cs
```diff
@@ -42,13 +44,18 @@ public class RuleFinder
 
         public RuleFinder()
         {
-            this.diagnosticAnalyzers = PackagedRuleAssemblies
+            diagnosticAnalyzers = PackagedRuleAssemblies
                 .SelectMany(assembly => assembly.GetTypes())
-                .Where(t => t.IsSubclassOf(typeof(DiagnosticAnalyzer)))
+                .Where(t => t.IsSubclassOf(typeof(DiagnosticAnalyzer)) || typeof(IRuleFactory).IsAssignableFrom(t))
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: I was expecting to return false but it seems that it returns true also for interfaces.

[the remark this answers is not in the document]
andrei-epure-sonarsource: the only downside of using `IsSubclassOf` would be if the Roslyn API will change in the future...

although we should add ITs with all versions of MSBuild and then we can catch this sort of differences in behavior. And I would keep the code more readable by using the same method.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/ResourceTests/GeneratedResourcesTest.cs
```diff
@@ -59,6 +60,30 @@ public void AnalyzersHaveCorrespondingResource_VisualBasic()
             rulesFromResources.Should().Equal(rulesFromClasses);
         }
 
+        [TestMethod]
+        public void ThereShouldBeRuleDetailsForAllCSharpRuleClasses()
```
andrei-epure-sonarsource, line 64: great!

PR #3352, opened 2020-05-18
Refactor S2583 and S2589 to reuse symbolic execution

### LINKED ISSUE #3343: Refactor ConditionEvaluatesToConstant to reuse symbolic execution
Refactor `ConditionEvaluatesToConstant` analyzer to reuse symbolic execution.

Part of https://jira.sonarsource.com/browse/MMF-2056.

costin-zaharia-sonarsource:
Fixes #3343

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
@@ -18,7 +18,6 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
@@ -26,9 +25,9 @@ using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
-using SonarAnalyzer.Common;
 using SonarAnalyzer.ControlFlowGraph;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.Rules.SymbolicExecution;
 using SonarAnalyzer.ShimLayer.CSharp;
 using SonarAnalyzer.SymbolicExecution;
 using SonarAnalyzer.SymbolicExecution.Constraints;
@@ -36,32 +35,29 @@ using CSharpExplodedGraph = SonarAnalyzer.SymbolicExecution.CSharpExplodedGraph;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
-    [DiagnosticAnalyzer(LanguageNames.CSharp)]
-    [Rule(S2583DiagnosticId)]
-    [Rule(S2589DiagnosticId)]
-    public sealed class ConditionEvaluatesToConstant : SonarDiagnosticAnalyzer
+    internal sealed class ConditionEvaluatesToConstant : ISymbolicExecutionAnalyzer
     {
-        private static readonly ISet<SyntaxKind> OmittedSyntaxKinds = new HashSet<SyntaxKind>
+        private static readonly ISet<SyntaxKind> omittedSyntaxKinds = new HashSet<SyntaxKind>
         {
             SyntaxKind.LogicalAndExpression,
             SyntaxKind.LogicalOrExpression
         };
 
-        private static readonly ISet<SyntaxKind> LoopBreakingStatements = new HashSet<SyntaxKind>
+        private static readonly ISet<SyntaxKind> loopBreakingStatements = new HashSet<SyntaxKind>
         {
             SyntaxKind.BreakStatement,
             SyntaxKind.ThrowStatement,
             SyntaxKind.ReturnStatement
         };
 
-        private static readonly ISet<SyntaxKind> LoopStatements = new HashSet<SyntaxKind>
+        private static readonly ISet<SyntaxKind> loopStatements = new HashSet<SyntaxKind>
         {
             SyntaxKind.WhileStatement,
             SyntaxKind.DoStatement,
             SyntaxKind.ForStatement
         };
 
-        private static readonly ISet<SyntaxKind> ConditionalStatements = new HashSet<SyntaxKind>
+        private static readonly ISet<SyntaxKind> conditionalStatements = new HashSet<SyntaxKind>
         {
             SyntaxKind.IfStatement,
             SyntaxKind.WhileStatement,
@@ -70,7 +66,7 @@ namespace SonarAnalyzer.Rules.CSharp
             SyntaxKind.CoalesceExpression
         };
 
-        private static readonly ISet<SyntaxKind> CoalesceExpressions = new HashSet<SyntaxKind>
+        private static readonly ISet<SyntaxKind> coalesceExpressions = new HashSet<SyntaxKind>
         {
             SyntaxKind.CoalesceExpression,
             SyntaxKindEx.CoalesceAssignmentExpression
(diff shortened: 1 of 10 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
```diff
@@ -60,16 +61,38 @@ private void Analyze(CSharpExplodedGraph explodedGraph, SyntaxNodeAnalysisContex
...
+            void ExplorationEndedHandler(object sender, EventArgs args)
+            {
+                ReportDiagnostics(analyzerContexts, context, false);
+            }
+        }
 
-                analyzerContexts.ForEach(analyzerContext => analyzerContext.Dispose());
+        private static void ReportDiagnostics(IEnumerable<ISymbolicExecutionAnalysisContext> analyzerContexts, SyntaxNodeAnalysisContext context, bool supportsPartialWalking)
+        {
+            foreach (var diagnostic in analyzerContexts
```
andrei-epure-sonarsource, line 91: hm do you know why the coverage says 3/4 conditions covered for the `foreach`? - see https://sonarcloud.io/code?id=sonaranalyzer-dotnet&pullRequest=3352&selected=sonaranalyzer-dotnet%3Asonaranalyzer-dotnet%2Fsrc%2FSonarAnalyzer.CSharp%2FRules%2FSymbolicExecution%2FSymbolicExecutionRunner.cs&line=91

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/ConditionEvaluatesToConstantTest.cs
```diff
@@ -52,9 +55,16 @@ public void ConditionEvaluatesToConstant_FromCSharp7()
...
-                new ConditionEvaluatesToConstant(),
+                GetAnalyzer(),
                 ParseOptionsHelper.FromCSharp8,
                 additionalReferences: NuGetMetadataReference.NETStandardV2_1_0);
         }
+
+        private static SonarDiagnosticAnalyzer GetAnalyzer()
+        {
+            // Symbolic execution analyzers are run by the SymbolicExecutionRunner
+            var analyzers = ImmutableArray.Create<ISymbolicExecutionAnalyzer>(new ConditionEvaluatesToConstant());
```
andrei-epure-sonarsource, line 66: nitpick: I think you can make this an arrow method and move the `ImmutableArray.Create` as parameter to the factory

`analyzers` suggests there could be more, and in this UT we don't expect more, so by having an arrow method you'd simplify things

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
```diff
@@ -151,29 +150,36 @@ void explorationEnded(object sender, EventArgs args)
...
-                explodedGraph.Walk();
+                hasYieldStatement = hasYieldStatement || IsYieldNode(args.ProgramPoint.Block);
+                CollectCoalesce(args, isNull, isNotNull, isUnknown, context.SemanticModel);
             }
-            finally
+
+            private void ConditionEvaluatedHandler(object sender, ConditionEvaluatedEventArgs args) =>
+                CollectConditions(args, conditionTrue, conditionFalse, context.SemanticModel);
+
+            private static Diagnostic GetDiagnostics(SyntaxNode constantNode, bool constantValue)
```
andrei-epure-sonarsource, line 178: nitpick: you've moved the `GetDiagnostics` method here but I don't see a big need...

it has the same modifiers as `GetUnreachableLocations` which has been left where it was

just from a file history perspective. Also: I'm fine with leaving as it is if you think it improves the readability in any way.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/ISymbolicExecutionAnalysisContext.cs
```diff
@@ -28,6 +28,8 @@ namespace SonarAnalyzer.Rules.SymbolicExecution
     // or cached nodes) and clear it and the end.
     public interface ISymbolicExecutionAnalysisContext : IDisposable
     {
+        bool SupportPartialWalk { get; }
```
andrei-epure-sonarsource, line 31: nitpick: I would use `SupportsPartialWalking` (it's a state rather than an action)

Also - can we find a better name or add a comment on the interface - what does "support partial walking mean"?

I need to look in `SymbolicExecutionRunner.cs` to tell, and it's not very clear - for those rules that support partial walking, I raise _only_ during the run and not after the full exploration, whereas for the other ones, which don't support Partial Walking, I run _only_ when the exploration ended.

So from what I see, `SupportPartialWalking` is exclusive with `ReportAtExplorationEnd`.

Whereas when I first read "Support Partial Walking", I imagined this is an additive / complementary attribute, like "I do what the other rules do AND I also do stuff with partial results".

I think we need somewhere to document this behavior.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
costin-zaharia-sonarsource: Hard to say. In the open cover report there are indeed 4 branches and one is not covered (line 91 from column 23 to 47).

```
<BranchPoints>
	<BranchPoint vc="1077" uspid="15803" ordinal="0" offset="23" sl="91" path="0" offsetend="25" fileid="1145" />
	<BranchPoint vc="0" uspid="15804" ordinal="1" offset="23" sl="91" path="1" offsetend="47" fileid="1145" />
	<BranchPoint vc="1" uspid="15805" ordinal="2" offset="58" sl="91" path="0" offsetend="60" fileid="1145" />
	<BranchPoint vc="1076" uspid="15806" ordinal="3" offset="58" sl="91" path="1" offsetend="83" fileid="1145" />
</BranchPoints>
```

This is what report generator is showing:
image

AFAICT it is related to generated code and I think it's not worth to cover just for the sake of having 100%.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
costin-zaharia-sonarsource: The problem was that the compiler will resolve `GetDiagnostics()` (the one without parameters) instead and I wanted to avoid a rename or a bigger refactoring by moving all the methods in the context. 

I've moved the method back and renamed it. I think it makes sense to stay near the other ones.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/ISymbolicExecutionAnalysisContext.cs
costin-zaharia-sonarsource: WDYT about `SupportPartialResults`? Since it's not the context responsibility to know about the tree walking/exploration.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
andrei-epure-sonarsource: no, it was just out of curiosity

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
```diff
@@ -128,7 +128,7 @@ public AnalysisContext(CSharpExplodedGraph explodedGraph, SyntaxNodeAnalysisCont
                 explodedGraph.ConditionEvaluated += ConditionEvaluatedHandler;
             }
 
-            public bool SupportPartialWalk { get; }
+            public bool SupportPartialResults { get; }
```
andrei-epure-sonarsource, line 131: sorry... it should be `SupportsPartialResults` (notice the s)

when asking the object `f.SupportsPartialResults`

`Support()` would be a command to the object (please support this)

and as a property "Support" is the plural form of the verb, we need the singular

PR #3356, opened 2020-05-19
Fix S2933 FP and related S1854, S3010, S2696, S2934 FN: Null-coalescing assignment is not recognized

### LINKED ISSUE #3339: Fix S2933 FP: Null-coalescing assignment is not recognized
Ruleset tells me to make _value readonly but this is not what I want.
The following is just an easy example.

(In my real scenario the inherited class is doing some expensive calculation with an object other than string.)

    public abstract class SampleS2933Issue
        {
            private string? _value = null;
            public string Value => _value ??= Generate();

            protected abstract string Generate();
        }

pavel-mikula-sonarsource:
Fixes #3339

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldBeReadonly.Fixed.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldBeReadonly.Fixed.cs
@@ -193,4 +193,19 @@ namespace Tests.Diagnostics
             otherField = 5;
         }
     }
+
+    // https://github.com/SonarSource/sonar-dotnet/issues/3339
+    public class NullCoalesceAssignment
+    {
+        private string value = null;
+        private readonly string notWritten = "Default"; // Fixed
+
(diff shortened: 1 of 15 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldBeReadonly.Fixed.cs
```diff
@@ -193,4 +193,19 @@ public DerivedFromSerializable()
...
         }
     }
+
+    // https://github.com/SonarSource/sonar-dotnet/issues/3339
+    public class NullCoalesceAssignment
+    {
+        private string value = null;
+        private readonly string notWritten = "Default"; // Fixed
+
+        public void DoSomething()
```
costin-zaharia-sonarsource, line 203: I would use the following snippet: 
```
public class NullCoalesceAssignment
{
    private string value = null;

    public void DoSomething()
    {
        value ??= "Empty";
    }
}
```
it reproduces the problem and it makes it obvious. Having another field which is not related to the test case can be confusing.

Issue #3346, opened 2020-05-15
Refactor NullPointerDereference to reuse symbolic execution


costin-zaharia-sonarsource:
Refactor `NullPointerDereference` analyzer to reuse symbolic execution.

Part of https://jira.sonarsource.com/browse/MMF-2056.

PR #3367, opened 2020-05-26
Split InvalidCastToInterface

costin-zaharia-sonarsource:
Since this rule combines two (distinct and complementary) implementations I've split it in two analyzers:
- one is doing semantic analysis
- the other symbolic execution

This split will ease #3346 implementation.

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InvalidCastToInterfaceRuleDetails.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InvalidCastToInterfaceRuleDetails.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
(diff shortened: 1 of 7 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -44,13 +44,16 @@ internal static class DiagnosticVerifier
             "BC36716" // VB12 does not support line continuation comments" i.e. a comment at the end of a multi-line statement.
         };
 
-        public static void Verify(Compilation compilation, DiagnosticAnalyzer diagnosticAnalyzer, CompilationErrorBehavior checkMode)
+        public static void Verify(Compilation compilation, DiagnosticAnalyzer diagnosticAnalyzer, CompilationErrorBehavior checkMode) =>
```
costin-zaharia-sonarsource, line 47: I've extended `DiagnosticVerifier.Verify` to run multiple analyzers in order to be able to keep tests for `InvalidCastToInterface` in the same file. Alternatively, if we decide this is not needed, we can split the tests and remove the changes done on `DiagnosticVerifier`.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InvalidCastToInterfaceRuleDetails.cs
```diff
@@ -0,0 +1,34 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Microsoft.CodeAnalysis;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules
+{
+    internal static class InvalidCastToInterfaceRuleDetails
```
andrei-epure-sonarsource, line 26: I would rename to `InvalidCastToInterfaceRuleConstants` (details is a bit vague)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
andrei-epure-sonarsource: I think it's good to keep the test as is (one single test for two DiagnosticAnalyzers), especially for this refactoring PR

if we'll need to modify one of the DiagnosticAnalyzers in the future , we may want to be able to test each individually

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -125,15 +132,20 @@ IEnumerable<TMessage> ReadProtobuf(string path)
 
         // This method is checking only the expected issues from the first file path provided. The rest of the paths are added to the
         // project for enabling testing of different scenarios.
-        public static void VerifyAnalyzer(IEnumerable<string> paths, SonarDiagnosticAnalyzer diagnosticAnalyzer,
+        public static void VerifyAnalyzer(IEnumerable<string> paths, SonarDiagnosticAnalyzer diagnosticAnalyzers,
```
andrei-epure-sonarsource, line 135:
```suggestion
        public static void VerifyAnalyzer(IEnumerable<string> paths, SonarDiagnosticAnalyzer diagnosticAnalyzer,
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -119,14 +122,21 @@ private static void DumpActualDiagnostics(IEnumerable<Diagnostic> diagnostics)
         }
 
         public static IEnumerable<Diagnostic> GetDiagnostics(Compilation compilation,
-            DiagnosticAnalyzer diagnosticAnalyzer, CompilationErrorBehavior checkMode,
+            DiagnosticAnalyzer diagnosticAnalyzers, CompilationErrorBehavior checkMode,
```
andrei-epure-sonarsource, line 125:
```suggestion
            DiagnosticAnalyzer diagnosticAnalyzer, CompilationErrorBehavior checkMode,
```

PR #3369, opened 2020-05-27
Stop caching rule parameters; hardening for reading the config file

### LINKED ISSUE #3361: Analyzers incorrectly cache rule parameters from the SonarLint.xml file
### Description

Rule parameters are passed to the analyzers via a file called `SonarLint.xml` using the Roslyn `AdditionalFiles` mechanism. Roslyn passes the additional files as part of the analysis context each time an analysis is triggered. The contents of the additional files can change between analysis runs.

The sonar-dotnet analyzers load the settings once per AppDomain and then ignore any settings that are supplied in subsequent analysis runs in the same AppDomain. This can lead to the wrong rule parameters being applied.

### Repro steps

1. Open a project in VS and install the SonarC# analyzer (or use SonarLint).
2. Add a `SonarLint.xml` settings with the content given below, with the `Build Action` set to `C# analyzer additional file` (or set the item type to `AdditionalFiles` if you are editing the project file directly.
3. Edit the ruleset to turn on rule S1451
4. Edit a code file to add a copyright header with the wrong year, and check the issue is raised.
4. Change the year in the copyright in the `SonarLint.xml` file so the rule should not fire, and save the file.
5. Edit the code to trigger the rule again.

### Expected behaviour
The new settings are picked up and the issue is not raised.

### Actual behaviour
The new settings are ignored until VS is restarted.

#### Repro SonarLint.xml file
```
<?xml version="1.0" encoding="UTF-8"?>
<AnalysisInput>
  <Settings/>
  <Rules>
    <Rule>
      <Key>S1451</Key>
      <Parameters>
        <Parameter>
          <Key>headerFormat</Key>
          <Value>// Copyright ACME 2020
</Value>
        </Parameter>
        <Parameter>
          <Key>isRegularExpression</Key>
          <Value>false</Value>
        </Parameter>
      </Parameters>
    </Rule>
  </Rules>
  <Files>
  </Files>
</AnalysisInput>
```

### Known workarounds

Restart VS.

### Related information

* C#/VB.NET Plugins Version: all versions with parameterised rules
* Visual Studio Version: all

### Background
The parameters are loaded and cached by the static `ParameterLoader` class that holds a static list of the parameterised analyzers it has processed. If it has already set the parameters for a particular analyzer it won't set them again - [see here](https://github.com/SonarSource/sonar-dotnet/blob/25c1ce8c65d14f2d284657a920dfe2490a3a93b7/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/ParameterLoader.cs#L88).

In VS, this means that the once the analyzers have loaded a `SonarLint.xml` file, those rule parameters will be used until VS is restarted. It doesn't matter if the user edits the settings file, switches to a solution with a different `SonarLint.xml` file, or even a different solution that does not have a `SonarLint.xml` file.

I haven't checked whether command line builds are impacted, but my guess is that they are. MSBuild can reuse nodes to build subsequent projects, which might have different (or no) `SonarLint.xml` settings.

Also, both `dotnet` and `MSBuild` have shared build server processes that keep running after a build is complete that can then be used when building other solutions.

andrei-epure-sonarsource:
Fixes #3361

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/ParameterLoader.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/ParameterLoader.cs
@@ -58,10 +58,16 @@ namespace SonarAnalyzer.Helpers
             }
 
             ImmutableList<RuleParameterValues> parameters;
-            using (var xtr = XmlReader.Create(additionalFile.Path))
+            try
             {
+                using var xtr = XmlReader.Create(additionalFile.Path);
                 var xml = XDocument.Load(xtr);
                 parameters = ParseParameters(xml);
+            }
+            catch (Exception ex) when (ex is IOException || ex is XmlException)
+            {
+                // cannot log exception
+                return;
             }
 
             var propertyParameterPairs = parameteredAnalyzer.GetType()
@@ -83,9 +89,10 @@ namespace SonarAnalyzer.Helpers
                     return;
                 }
 
-                var value = parameterValue.ParameterValue;
-                var convertedValue = ChangeParameterType(value, propertyParameterPair.Descriptor.Type);
-                propertyParameterPair.Property.SetValue(parameteredAnalyzer, convertedValue);
+                if (TryConvertToParameterType(parameterValue.ParameterValue, propertyParameterPair.Descriptor.Type, out var value))
+                {
+                    propertyParameterPair.Property.SetValue(parameteredAnalyzer, value);
+                }
             }
         }
 
@@ -127,22 +134,26 @@ namespace SonarAnalyzer.Helpers
             return builder.ToImmutable();
         }
 
-        private static object ChangeParameterType(string parameter, PropertyType type)
+        private static bool TryConvertToParameterType(string parameter, PropertyType type, out object result)
         {
             switch (type)
             {
                 case PropertyType.Text:
                 case PropertyType.String:
-                    return parameter;
+                    result = parameter;
+                    return true;
 
-                case PropertyType.Integer:
-                    return int.Parse(parameter, NumberStyles.None, CultureInfo.InvariantCulture);
+                case PropertyType.Integer when int.TryParse(parameter, NumberStyles.None, CultureInfo.InvariantCulture, out var parsedInt):
+                    result = parsedInt;
+                    return true;
 
-                case PropertyType.Boolean:
-                    return bool.Parse(parameter);
+                case PropertyType.Boolean when bool.TryParse(parameter, out var parsedBool):
+                    result = parsedBool;
+                    return true;
 
                 default:
-                    throw new NotSupportedException();
+                    result = null;
+                    return false;
             }
         }
 
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/ParameterLoader.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/ParameterLoader.cs
@@ -33,11 +33,11 @@ using SonarAnalyzer.Common;
 
 namespace SonarAnalyzer.Helpers
 {
-    public static class ParameterLoader
+    internal static class ParameterLoader
     {
-        public static readonly string ParameterConfigurationFileName = "SonarLint.xml";
+        private const string ParameterConfigurationFileName = "SonarLint.xml";
 
-        public static void SetParameterValues(ParameterLoadingDiagnosticAnalyzer parameteredAnalyzer,
+        internal static void SetParameterValues(ParameterLoadingDiagnosticAnalyzer parameteredAnalyzer,
             AnalyzerOptions options)
         {
             if (ProcessedAnalyzers.ContainsKey(parameteredAnalyzer))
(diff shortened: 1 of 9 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/ParameterLoader.cs
```diff
@@ -127,22 +134,26 @@ private static ImmutableList<RuleParameterValues> ParseParameters(XDocument xml)
             return builder.ToImmutable();
         }
 
-        private static object ChangeParameterType(string parameter, PropertyType type)
+        private static bool TryConvertToParameterType(string parameter, PropertyType type, out object result)
```
andrei-epure-sonarsource, line 132: I refactored this to fail gracefuly

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/ParameterLoader.cs
```diff
@@ -86,7 +86,7 @@ internal static class ParameterLoader
 
                 if (parameterValue == null)
                 {
-                    return;
+                    continue;
```
andrei-epure-sonarsource: the logic didn't work if the rule had 2 parameters and the sonarlint.xml contained the second parameter only - this logic was stopping too early

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/ParameterLoader.cs
```diff
@@ -86,7 +86,7 @@ internal static class ParameterLoader
 
                 if (parameterValue == null)
                 {
-                    return;
```
andrei-epure-sonarsource, line 124: the logic didn't work if the rule had 2 parameters and the sonarlint.xml contained the second parameter only - this logic was stopping too early

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/ParameterLoader.cs
```diff
@@ -79,86 +129,46 @@ private static ImmutableList<RuleParameterValues> ParseParameters(XDocument xml)
...
 
                 default:
-                    throw new NotSupportedException();
+                    result = null;
+                    return false;
             }
         }
 
-        public static bool IsSonarLintXml(string path)
+        private class RuleParameterValues
```
costin-zaharia-sonarsource, line 160: I think that we can make these sealed.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/ParameterLoader.cs
```diff
@@ -33,22 +32,73 @@
...
+            return ConfigurationFilePathMatchesExpected(path, ParameterConfigurationFileName);
+        }
 
-            public List<RuleParameterValue> ParameterValues { get; } = new List<RuleParameterValue>();
+        internal static bool ConfigurationFilePathMatchesExpected(string path, string fileName)
+        {
+            return new FileInfo(path).Name.Equals(fileName, StringComparison.OrdinalIgnoreCase);
         }
 
         private static ImmutableList<RuleParameterValues> ParseParameters(XDocument xml)
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: I changed it, although for private methods I don't think it's that important

also, for `SetParameterValues` I want to keep `ParameterLoadingDiagnosticAnalyzer` because this class should be used by that class (it's kinda tightly coupled)

andrei-epure-sonarsource:
> I've noticed that the hotspots configuration is also cached: /sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs@master#L105
> we don't care about those since they are run only by the scanner, right?

It's a good question. I think @duncanp-sonar mentioned at the .NET guild meeting that msbuild can have a behavior in which it could end up caching the wrong thing, and we may consider not caching them if we can get a reproducer of the problem. I'm not sure how to reproduce that problem though, because there should be two different SQ projects build during the same run, with 2 different QPs - one QP with enabled hotspots and one QP with disabled hotspots. I don't have any plans to open a ticket for this.

Anyhow, indeed it doesn't affect SonarLint which is the main topic of #3361.

andrei-epure-sonarsource:
@costin-zaharia-sonarsource please take another look at the updates

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/ParameterLoaderTests.cs
```diff
@@ -76,6 +76,20 @@ public void SetParameterValues_WhenGivenSonarLintFileHasStringParameterType_Only
             analyzer.FlagsEnumNamePattern.Should().Be("1"); // value from XML file
         }
 
+        [TestMethod]
+        public void SetParameterValues_WhenGivenSonarLintFileHasBooleanParameterType_OnlyOneParameter_PopulatesProperty()
+        {
+            // Arrange
+            var options = CreateOptionsWithAdditionalFile("ResourceTests\\RuleWithBooleanParameter\\SonarLint.xml");
```
costin-zaharia-sonarsource, line 83: I would rename `CreateOptionsWithAdditionalFile(string relativePath)` to `CreateOptions(string additionalFilePath)`. I like this more since it simplifies the method name and the parameter will provide all the relevant information about the context. I'm also thinking that if we add an additional parameter it's possible that the name will not make sense anymore and we will have to change it which make the code a bit harder to maintain.

costin-zaharia-sonarsource:
> @costin-zaharia-sonarsource please take another look at the updates

It looks good. Just the comment regarding the method name.

PR #3375, opened 2020-06-02
S2583: Fix FP when using the new "using statement" from C# 8

costin-zaharia-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecutionRules.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace SonarAnalyzer.UnitTest.TestCases
+{
+    public class SymbolicExecutionRules
+    {
+        internal static class Program
+        {
+            public static void Main()
+            {
+                using var resource = GetNullResource();
```
andrei-epure-sonarsource, line 11: can you also add a test with a `using` statement?

PR #3376, opened 2020-06-03
Fix S3949 FN by adding support for constants in intermediate code generation

### LINKED ISSUE #3368: Fix S3949 FN by adding support for constants in intermediate code generation
### Description

When detecting integer overflows, we currently don't support constants, such as `int.MaxValue`, and therefore don't trigger the rule in situations where it should be triggered.

### Repro steps
Related to [community thread](https://community.sonarsource.com/t/rule-s3949-fn-calculations-should-not-overflow-bug-warning-is-not-caught-by-sonar-scanner-and-it-is-not-displayed-in-the-report-on-sonarqube-8-2-version-community-edition/21883)

### Actual behavior

Looking at internal intermediate code representation, we generate: 
```
%5 = cbde.unknown : i32  // int (PredefinedType)
%6 = cbde.unknown : i32  // int.MaxValue (SimpleMemberAccessExpression)
```
While we should generate:
```
%5 = constant 2147483647 : i32
```

pavel-mikula-sonarsource:
Fixes #3368 

Please review it commit by commit, since there's small functional fix and big code cleanup.


pavel-mikula-sonarsource:
cc @loic-joly-sonarsource if you'd like to review the second commit "Fix issue"

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/CBDE/CheckMLIRWithReference.cs
```diff
@@ -302,7 +302,48 @@ private void g(int dim)
...
+}
+";
+            var expected = @"
+func @_$invalid$global$code$.Func$$() {
+  br ^bb1
+^bb1:	// pred: ^bb0
+  %c42_i32 = constant 42 : i32
+  %0 = cbde.alloca i32
+  cbde.store %c42_i32, %0 : memref<i32>
+  %1 = cbde.unknown : i32
```
loic-joly-sonarsource, line 328: Does this line matches the `int.`, or the `MaxValue` part?
Anyways, it is useless (and could be removed), but it looks harmless...

loic-joly-sonarsource:
I only reviewed the "fix issue" commit.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/CBDE/CheckMLIRWithReference.cs
pavel-mikula-sonarsource: It matches the `int.` part. Cleaning this one would be another story (not too difficult). But it would not remove the `Int32.` variant (more difficult) so I've chosen to go with "harmless" approach.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/CBDE/MlirTestUtilities.cs
```diff
@@ -71,7 +69,12 @@ public static void ValidateWithReference(string code, string expected, string te
             var expectedLines = trimmedExpected.Split('\n');
             var actualLines = trimmedActual.Split('\n');
             var maxLines = Math.Min(expectedLines.Length, actualLines.Length);
-            for (int i = 0; i < maxLines; ++i)
+            Console.WriteLine("Expected:");
+            Console.WriteLine(trimmedExpected);
+            Console.WriteLine();
+            Console.WriteLine("Actual:");
+            Console.WriteLine(trimmedActual);
```
costin-zaharia-sonarsource, line 76: I think you forgot to remove these :)

pavel-mikula-sonarsource: No, it's intentional improvement. Just like the dump of UCFGs and the "Open additional output for this result" in tests.
The utility only reports difference of the first unexpected line. That's a nightmare to debug/diagnose.

Issue #3282, opened 2020-04-02
S5773 Types allowed to be deserialized should be restricted


eric-therond-sonarsource:
Implements https://jira.sonarsource.com/browse/RSPEC-5773 (C# description [here](https://jira.sonarsource.com/browse/RSPEC-5774)). Part of MMF-2064.

For `BinaryFormatter`, `NetDataContractSerializer`, `ObjectStateFormatter`, `SoapFormatter` serializers raise when :
1.  A binder is not set before `Deserialize` method is called:
```
BinaryFormatter formatter = new BinaryFormatter(); // Noncompliant
// formatter.Binder = new CustomBinder(); 
dsc = (ExpectedSerializedClass)formatter.Deserialize(fs);
formatter.Binder = new CustomBinder(); 
```
2.  There is a `Deserialize` method call, a binder is set but the `BindToType` method of the binder is not returning at least one `throw` or `null` value:
```
sealed class CustomBinder : SerializationBinder
{
   public override Type BindToType(string assemblyName, string typeName)  // Noncompliant: secondary location
   {
      return Assembly.Load(assemblyName).GetType(typeName); // no checks is done here, always return the type
   }
}

BinaryFormatter formatter = new BinaryFormatter();
formatter.Binder = new CustomBinder(); 
dsc = (ExpectedSerializedClass)formatter.Deserialize(fs); // Noncompliant: main location
```

For `LosFormatter` serializer it should raise on the constructor with no arguments or when the first argument is set to `false`:
```
formatter1 = new LosFormatter(); // Noncompliant
formatter1 = new LosFormatter(false, "toto"); // Noncompliant
formatter1 = new LosFormatter(true, "toto"); // Compliant
```

For `JavaScriptSerializer` serializer it should raise when:
1.  `SimpleTypeResolver` is used (and there is a call to `Deserialize`):
```
JavaScriptSerializer serializer1 = new JavaScriptSerializer(new SimpleTypeResolver()); // Noncompliant: SimpleTypeResolver is unsafe, secondary location
serializer1.Deserialize<ExpectedType>(json); // Noncompliant: main location
```

2.  There is a call to `Deserialize` and a custom resolver is used but with a `ResolveType` method which is not returning at least one `throw` or `null` value:
```
public class CustomTypeResolver : JavaScriptTypeResolver
{
  public override Type ResolveType(string id)  // Noncompliant: secondary location 
  {
    return Type.GetType(id); // every type is resolved
  }
}

[...]

PR #3382, opened 2020-06-10
S5773: Deserialized types should be restricted for BinaryFormatter, NetDataContractSerializer and SoapFormatter serializers

costin-zaharia-sonarsource:
Partial implementation for #3282


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
```diff
@@ -0,0 +1,239 @@
...
+                {
+                    var typeSymbol = semanticModel.GetTypeInfo(instruction).Type;
+                    if (IsFormatterWithBinder(typeSymbol) &&
+                        !HasSafeBinderInitialization(objectCreation.Initializer))
+                    {
+                        programState = programState.SetConstraint(symbolicValue, SerializationBinderConstraint.Unsafe);
+                    }
+                }
+
+                return base.ObjectCreated(programState, symbolicValue, instruction);
```
christophe-zurn-sonarsource, line 106: `return programState` would have the exact same behaviour and would be clearer in what it does, as otherwise we could think there is some logic done by the parent class, which is not the case

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
```diff
@@ -0,0 +1,239 @@
...
+                var instruction = programPoint.Block.Instructions[programPoint.Offset];
+
+                programState = instruction switch
+                {
+                    MemberAccessExpressionSyntax memberAccess => VisitMemberAccess(memberAccess, programState),
+                    AssignmentExpressionSyntax assignmentExpressionSyntax => VisitAssignmentExpression(assignmentExpressionSyntax, programState),
+                    _ => programState
+                };
+
+                return base.PreProcessInstruction(programPoint, programState);
```
christophe-zurn-sonarsource, line 120: `return programState` would have the exact same behaviour and would be clearer in what it does, as otherwise we could think there is some logic done by the parent class, which is not the case

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
```diff
@@ -0,0 +1,239 @@
...
+                });
+
+            private static MethodDeclarationSyntax GetBindToTypeMethodDeclaration(ISymbol symbol) =>
+                symbol.DeclaringSyntaxReferences
+                    .SelectMany(GetDescendantNodes)
+                    .OfType<MethodDeclarationSyntax>()
+                    .FirstOrDefault(IsBindToType);
+
+            private static IEnumerable<SyntaxNode> GetDescendantNodes(SyntaxReference syntaxReference) =>
+                syntaxReference.SyntaxTree.GetRoot().FindNode(syntaxReference.Span).DescendantNodes();
```
christophe-zurn-sonarsource, line 215: Can't we use
```suggestion
                syntaxReference.GetSyntax().DescendantNodes();
```
instead ?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RestrictDeserializedTypes.Binders.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    internal sealed class UnsafeBinderWithOtherMethods : SerializationBinder
+    {
+        public void Accept()
+        {
+        }
+
+        public override Type BindToType(string assemblyName, string typeName) =>
+            Assembly.Load(assemblyName).GetType(typeName);
+    }
```
christophe-zurn-sonarsource, line 69: Can we add a test with a binder that has almost same signature but incorrect paramters (Ex: param 2 is int)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            private static bool IsBindToType(MethodDeclarationSyntax methodDeclaration) =>
+                methodDeclaration.Identifier.Text == "BindToType" &&
+                methodDeclaration.ReturnType.NameIs("Type") &&
+                methodDeclaration.ParameterList.Parameters.Count == BindToTypeParameterCount &&
+                methodDeclaration.ParameterList.Parameters[0].IsString() &&
+                methodDeclaration.ParameterList.Parameters[1].IsString();
```
christophe-zurn-sonarsource, line 222: I'd rather have the `BindToTypeParameterCount ` value (2) inlined here, as we are directly using list access to `[0]` and `[1]` afterwards, even if we need to mark the SQ issue as "won't fix

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RestrictDeserializedTypes.cs
```diff
@@ -0,0 +1,263 @@
...
+            var formatter4 = new BinaryFormatter();
+            formatter4.Binder = possibleNullBinder;
+            formatter4.Deserialize(stream); // Noncompliant: the binder can be null
+
+            var formatter5 = new BinaryFormatter();
+            if (condition)
+            {
+                formatter5.Binder = new SafeBinderStatementWithReturnNull();
+            }
+            formatter5.Deserialize(stream); // Noncompliant: the binder can be null
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Indeed if there are debug checks this could be a source for FP but SE brings a lot of benefits and I think it's better to stick with it.

sonaranalyzer-dotnet/its/sources/ManuallyAddedNoncompliantIssues/NetFramework48/BinaryFormatterTest.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System.IO;
+using System.Runtime.Serialization.Formatters.Binary;
+
+namespace NetFramework48
+{
+    public class BinaryFormatterTest
+    {
+        internal void BinaryFormatterDeserialize(Stream stream) =>
+            new BinaryFormatter().Deserialize(stream); // Noncompliant (S5773) {{Restrict types of objects allowed to be deserialized.}}
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Excellent question, while doing this I've found out that the setter for the `ObjectStateFormatter` binder is read-only so the binder cannot be set. I will ask @eric-therond-sonarsource how to handle this.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
costin-zaharia-sonarsource: In this case the base class does not have any processing of it's own but, in the future, if we chose  to extend it, we will have to modify all the derived classes. 

In my opinion the code is easier to maintain if we call the base class.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -339,6 +339,7 @@ expression switch
             {
                 MemberAccessExpressionSyntax memberAccess => memberAccess.Name.Identifier.ValueText,
                 IdentifierNameSyntax identifierName => identifierName.Identifier.ValueText,
+                PredefinedTypeSyntax predefinedTypeSyntax => predefinedTypeSyntax.Keyword.ToString(),
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: I thought it was needed during development but it wasn't in the end and I forgot about it.

sonaranalyzer-dotnet/its/sources/ManuallyAddedNoncompliantIssues/NetFramework48/BinaryFormatterTest.cs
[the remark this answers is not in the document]
costin-zaharia-sonarsource: I've talked with @eric-therond-sonarsource and we will remove `ObjectStateFormatter ` from the scope.

sonaranalyzer-dotnet/its/sources/ManuallyAddedNoncompliantIssues/NetFramework48/NetFramework48.csproj
```diff
@@ -35,6 +35,9 @@
   <ItemGroup>
     <Reference Include="System" />
     <Reference Include="System.Core" />
+    <Reference Include="System.Runtime.Serialization" />
+    <Reference Include="System.Runtime.Serialization.Formatters.Soap" />
+    <Reference Include="System.Web" />
```
christophe-zurn-sonarsource, line 40: Since `ObjectStateFormatterTest` was removed from the scope, I think we can remove the `System.Web` reference from the csproj file

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RestrictDeserializedTypes.cs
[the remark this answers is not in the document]
christophe-zurn-sonarsource: Ok. Fine by me. I just wanted to make sure we were aware of that :)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
christophe-zurn-sonarsource: Ok

sonaranalyzer-dotnet/its/sources/ManuallyAddedNoncompliantIssues/NetFramework48/NetFramework48.csproj
costin-zaharia-sonarsource: Indeed in this commit is not needed but I will have to remove it now and add it again in the next PR which handles [LosFormatter](https://docs.microsoft.com/en-us/dotnet/api/system.web.ui.losformatter.-ctor?redirectedfrom=MSDN&view=netframework-4.8#System_Web_UI_LosFormatter__ctor_System_Boolean_System_String_).

PR #3384, opened 2020-06-15
S5773: Add checks for JavaScriptSerializer

costin-zaharia-sonarsource:
Partial fix for: #3282

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
@@ -77,7 +77,7 @@ namespace SonarAnalyzer.Rules.CSharp
                     KnownType.System_Runtime_Serialization_NetDataContractSerializer,
                     KnownType.System_Runtime_Serialization_Formatters_Soap_SoapFormatter);
 
-            private readonly Dictionary<ITypeSymbol, bool> binderValidityMap = new Dictionary<ITypeSymbol, bool>();
+            private readonly Dictionary<ITypeSymbol, bool> symbolValidityMap = new Dictionary<ITypeSymbol, bool>();
 
             private readonly Action<SyntaxNode> addNode;
 
@@ -95,7 +95,13 @@ namespace SonarAnalyzer.Rules.CSharp
                     if (IsFormatterWithBinder(typeSymbol) &&
                         !HasSafeBinderInitialization(objectCreation.Initializer))
                     {
-                        programState = programState.SetConstraint(symbolicValue, SerializationBinderConstraint.Unsafe);
+                        programState = programState.SetConstraint(symbolicValue, SerializationConstraint.Unsafe);
+                    }
+
+                    if (IsJavaScriptSerializer(typeSymbol) &&
+                        !HasSafeResolverInitialization(objectCreation))
+                    {
+                        programState = programState.SetConstraint(symbolicValue, SerializationConstraint.Unsafe);
                     }
                 }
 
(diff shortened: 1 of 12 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
```diff
@@ -165,17 +171,45 @@ private ProgramState VisitAssignmentExpression(AssignmentExpressionSyntax assign
...
+                }
+
+                var typeSymbol = semanticModel.GetTypeInfo(resolverType.Expression).Type;
+                // If the symbol cannot be determined we consider the resolver safe to avoid FPs.
+                if (typeSymbol == null)
+                {
+                    return true;
+                }
+
+                return !typeSymbol.Is(KnownType.System_Web_Script_Serialization_SimpleTypeResolver) && IsTypeSafe(typeSymbol);
```
christophe-zurn-sonarsource, line 210:
```suggestion
                return typeSymbol == null || (!typeSymbol.Is(KnownType.System_Web_Script_Serialization_SimpleTypeResolver) && IsTypeSafe(typeSymbol));
```
Why is this logic different than for `HasSafeBinderInitialization` ? For `HasSafeBinderInitialization` if `typeSymbol` is null, method will return false.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
```diff
@@ -188,16 +222,17 @@ private bool HasSafeBinderInitialization(InitializerExpressionSyntax initializer
...
             }
 
-            private bool IsBinderSafe(ITypeSymbol symbol) =>
-                binderValidityMap.GetOrAdd(symbol, typeSymbol =>
+            private bool IsTypeSafe(ITypeSymbol symbol) =>
+                symbolValidityMap.GetOrAdd(symbol, typeSymbol =>
                 {
-                    var declaration = GetBindToTypeMethodDeclaration(typeSymbol);
+                    var declaration = GetBindToTypeMethodDeclaration(typeSymbol) ??
+                                      GetResolveTypeMethodDeclaration(typeSymbol);
```
christophe-zurn-sonarsource, line 233: Although unlikely, if we have a `JavaScriptTypeResolver` that implements `BindToType`, its the `BindToType` implementation that will be checked here instead of the `ResolveType`.
I think it's ok as the likelihood of this actually happening is very small, but it can be worth it to document as a unit test

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
```diff
@@ -40,6 +40,9 @@
     <None Include="TestCasesForRuleFailure\**\*">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Include="TestCases\RestrictDeserializedTypes.JavaScriptSerializer.cs">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
```
christophe-zurn-sonarsource, line 45: Why is this needed ? Isn't this already covered by 
```
    <None Include="TestCases\**\*">
    </None>
```
above ?

costin-zaharia-sonarsource: It's not needed. Was automatically added by the IDE and I forgot to remove it.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
costin-zaharia-sonarsource: I've added a fix and some test cases for this.

PR #3385, opened 2020-06-15
S5773: Add checks for LosFormatter

costin-zaharia-sonarsource:
Partial fix for #3282

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
@@ -105,7 +105,7 @@ namespace SonarAnalyzer.Rules.CSharp
                     }
 
                     if (IsLosFormatter(typeSymbol) &&
-                        !HasSafeConstructorInitialization(objectCreation, programState))
+                        !IsLosFormatterSafe(objectCreation, programState))
                     {
                         // For LosFormatter the rule is raised directly on the constructor.
                         addNode(objectCreation);
@@ -115,7 +115,7 @@ namespace SonarAnalyzer.Rules.CSharp
                 return base.ObjectCreated(programState, symbolicValue, instruction);
             }
 
-            private bool HasSafeConstructorInitialization(ObjectCreationExpressionSyntax objectCreation, ProgramState programState)
+            private bool IsLosFormatterSafe(ObjectCreationExpressionSyntax objectCreation, ProgramState programState)
             {
                 // The constructor is safe only if it has 2 arguments and the first argument value is true.
                 if (objectCreation.ArgumentList.Arguments.Count != 2)
@@ -123,7 +123,7 @@ namespace SonarAnalyzer.Rules.CSharp
                     return false;
                 }
 
-                var firstArgument = objectCreation.ArgumentList.Arguments[0].Expression;
+                var firstArgument = GetEnableMacArgumentSyntax(objectCreation.ArgumentList);
                 if (firstArgument.IsKind(SyntaxKind.FalseLiteralExpression))
                 {
                     return false;
@@ -146,6 +146,9 @@ namespace SonarAnalyzer.Rules.CSharp
                 return symbolicValue == null ||
                        !programState.HasConstraint(symbolicValue, BoolConstraint.False);
             }
+
+            private static ExpressionSyntax GetEnableMacArgumentSyntax(BaseArgumentListSyntax list) =>
+                (list.GetArgumentByName("enableMac") ?? list.Arguments[0]).Expression;
 
             public override ProgramState PreProcessInstruction(ProgramPoint programPoint, ProgramState programState)
             {
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
@@ -103,9 +103,48 @@ namespace SonarAnalyzer.Rules.CSharp
                     {
                         programState = programState.SetConstraint(symbolicValue, SerializationConstraint.Unsafe);
                     }
+
+                    if (IsLosFormatter(typeSymbol) &&
+                        !HasSafeConstructorInitialization(objectCreation, programState))
+                    {
+                        // For LosFormatter the rule is raised directly on the constructor.
+                        addNode(objectCreation);
+                    }
                 }
 
                 return base.ObjectCreated(programState, symbolicValue, instruction);
+            }
+
+            private bool HasSafeConstructorInitialization(ObjectCreationExpressionSyntax objectCreation, ProgramState programState)
+            {
+                // The constructor is safe only if it has 2 arguments and the first argument value is true.
+                if (objectCreation.ArgumentList.Arguments.Count != 2)
+                {
+                    return false;
+                }
+
+                var firstArgument = objectCreation.ArgumentList.Arguments[0].Expression;
+                if (firstArgument.IsKind(SyntaxKind.FalseLiteralExpression))
+                {
+                    return false;
+                }
+
+                if (firstArgument.IsKind(SyntaxKind.TrueLiteralExpression))
+                {
+                    return true;
+                }
+
+                var symbol = semanticModel.GetSymbolInfo(firstArgument).Symbol;
+                if (symbol == null)
+                {
+                    // In this case we cannot determine if the first parameter is true or false so we
+                    // assume it's true to avoid FPs.
+                    return true;
+                }
+
+                var symbolicValue = programState.GetSymbolValue(symbol);
+                return symbolicValue == null ||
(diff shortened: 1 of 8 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
```diff
@@ -103,11 +103,50 @@ public override ProgramState ObjectCreated(ProgramState programState, SymbolicVa
...
+                if (objectCreation.ArgumentList.Arguments.Count != 2)
+                {
+                    return false;
+                }
+
+                var firstArgument = objectCreation.ArgumentList.Arguments[0].Expression;
+                if (firstArgument.IsKind(SyntaxKind.FalseLiteralExpression))
+                {
+                    return false;
+                }
```
christophe-zurn-sonarsource, line 130: For readability, I think we could merge these 2 early return together

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
```diff
@@ -103,11 +103,50 @@ public override ProgramState ObjectCreated(ProgramState programState, SymbolicVa
...
+                    return true;
+                }
+
+                var symbol = semanticModel.GetSymbolInfo(firstArgument).Symbol;
+                if (symbol == null)
+                {
+                    // In this case we cannot determine if the first parameter is true or false so we
+                    // assume it's true to avoid FPs.
+                    return true;
+                }
```
christophe-zurn-sonarsource, line 143: Same here: for readability, I think we could merge these 2 early return together too

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
```diff
@@ -103,11 +103,50 @@ public override ProgramState ObjectCreated(ProgramState programState, SymbolicVa
...
 
+            private bool HasSafeConstructorInitialization(ObjectCreationExpressionSyntax objectCreation, ProgramState programState)
+            {
+                // The constructor is safe only if it has 2 arguments and the first argument value is true.
+                if (objectCreation.ArgumentList.Arguments.Count != 2)
+                {
+                    return false;
+                }
+
+                var firstArgument = objectCreation.ArgumentList.Arguments[0].Expression;
```
christophe-zurn-sonarsource, line 126: Contrary to `JavaScriptSerializer` constructor that was taking a single argument, or to formatters that had initializer, here the constructor can take 2 arguments, so it would be better to first check for named argument, in case the expressions were swapped for some reason. That way we will also cover the case:
```
new LosFormatter (macKeyModifier: new byte[0], enableMac: false)
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RestrictDeserializedTypes.LosFormatter.cs
```diff
@@ -0,0 +1,103 @@
+﻿using System;
+using System.Web.UI;
+
+namespace SonarAnalyzer.UnitTest.TestCases
+{
+    internal class RestrictDeserializedTypes
+    {
+        public void DefaultConstructor()
+        {
+            new LosFormatter(); // Noncompliant - MAC filtering should be enabled
```
christophe-zurn-sonarsource, line 10: Note: the message here will still be `Restrict types of objects allowed to be deserialized.`. Is this wanted or should we change it to `MAC filtering should be enabled` ?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
```diff
@@ -103,11 +103,50 @@ public override ProgramState ObjectCreated(ProgramState programState, SymbolicVa
...
+                    {
+                        // For LosFormatter the rule is raised directly on the constructor.
+                        addNode(objectCreation);
+                    }
                 }
 
                 return base.ObjectCreated(programState, symbolicValue, instruction);
             }
 
+            private bool HasSafeConstructorInitialization(ObjectCreationExpressionSyntax objectCreation, ProgramState programState)
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Renamed it to `IsLosFormatterSafe`.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
costin-zaharia-sonarsource: As discussed offline, the fact that `firstArgument` is used bellow it makes it harder to join the conditions in a more readable way.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
costin-zaharia-sonarsource: As discussed offline, the fact that `symbol` is used bellow it makes it harder to join the conditions in a more readable way.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
costin-zaharia-sonarsource: That's a very nice case I didn't think of. Thanks!

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RestrictDeserializedTypes.LosFormatter.cs
costin-zaharia-sonarsource: Indeed. I've checked also with @eric-therond-sonarsource and we will display a different message for `LosFormatter`: "Serialized data signature (MAC) should be verified."

costin-zaharia-sonarsource: If this is fine with you I would do this change in the following PR (the one with secondary locations) to avoid git re-base conflicts.

christophe-zurn-sonarsource: Ok

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/BaseArgumentListSyntaxExtensionsTests.cs
```diff
@@ -0,0 +1,81 @@
...
+    [TestClass]
+    public class BaseArgumentListSyntaxExtensionsTests
+    {
+        [TestMethod]
+        public void GivenEmptyList_GetArgumentByName_ReturnsNull() =>
+            CreateNamedArgumentList().GetArgumentByName("argument").Should().BeNull();
+
+        [TestMethod]
+        public void GivenListWithAnotherNamedArgument_GetArgumentByName_ReturnsNull() =>
+            CreateNamedArgumentList("p1").GetArgumentByName("p2").Should().Be(null);
```
christophe-zurn-sonarsource, line 39: For consistency, can we also have `Should().BeNull();` for `GivenListWithAnotherNamedArgument_GetArgumentByName_ReturnsNull` as it is done for `GivenEmptyList_GetArgumentByName_ReturnsNull`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/BaseArgumentListSyntaxExtensionsTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        [TestMethod]
+        public void GivenListWithNamedArgument_GetArgumentByName_ReturnsArgument() =>
+            CreateNamedArgumentList("p1").GetArgumentByName("p1").Should().Match(p => ((ArgumentSyntax)p).NameIs("p1"));
+
+        [TestMethod]
+        public void GivenListWithMultipleNamedArguments_GetArgumentByName_ReturnsArgument() =>
+            CreateNamedArgumentList("p1", "p2", "p3").GetArgumentByName("p2").Should().Match(p => ((ArgumentSyntax)p).NameIs("p2"));
```
christophe-zurn-sonarsource, line 47: Can we avoid using `NameIs` method in at least 1 of the assertion. (since the underlying logic of `GetArgumentByName` is based on `NameIs`, it deafeats the unit test purpose to test it with the same logic here)

PR #3387, opened 2020-06-17
S5773 Show secondary locations

costin-zaharia-sonarsource:
Partial implementation for #3282


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RestrictDeserializedTypes.cs
```diff
@@ -37,7 +37,7 @@ internal void BinderAsVariable(Stream stream, bool condition)
 
             var formatter2 = new BinaryFormatter();
             formatter2.Binder = unsafeBinder;
-            formatter2.Deserialize(stream); // Noncompliant: unsafe binder used
+            formatter2.Deserialize(stream); // Noncompliant [unsafeBinder1]: unsafe binder used
```
andrei-epure-sonarsource, line 40: what is `unsafeBinder1` used for?

costin-zaharia-sonarsource: This is a new syntax improvement added by @pavel-mikula-sonarsource and @christophe-zurn-sonarsource which allows to link a line of code to a specific secondary location.

sonaranalyzer-dotnet/its/expected/Nancy/Nancy-{34576216-0DCA-4B0F-A0DC-9075E75A676F}-S5773.json
```diff
@@ -7,7 +7,7 @@
 "uri":  "sources\Nancy\src\Nancy\DefaultObjectSerializer.cs",
 "region":  {
 "startLine":  54,
-"startColumn":  28,
+"startColumn":  38,
```
andrei-epure-sonarsource, line 10: shouldn't a secondary location be raised?

andrei-epure-sonarsource: or is this related to
```
                // An issue can appear multiple times for the same node while exploring the exploded graph.
                // In this case the location is added one single time to avoid duplicates.
```

but in that case, two issues should have been replaced by one, right?

costin-zaharia-sonarsource: In this case no since the secondary locations are added only for classes which derive from `SerializationBinder` or `JavaScriptTypeResolver`

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
```diff
@@ -322,5 +371,43 @@ private bool HasSafeBinderInitialization(InitializerExpressionSyntax initializer
...
+        }
+
+        private sealed class TypeDeclarationInfo
+        {
+            internal Location Location { get; }
+
+            internal bool IsSafe { get; }
+
+            public TypeDeclarationInfo(MethodDeclarationSyntax declaration)
+            {
```
andrei-epure-sonarsource, line 400: if think you can refactor this to
```
if (declaration == null)
{
  Location = null;
  IsSafe = true;
}
else
{
  Location = ..;
  IsSafe = declaration...;
}
```

otherwise, there are two null-checks which makes it harder to read

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/RestrictDeserializedTypes.JavaScriptSerializer.cs
```diff
@@ -7,7 +7,7 @@ internal class Serializer
     {
         public void SimpleTypeResolverDefaultConstructorIsNotSafe(string json)
         {
-            new JavaScriptSerializer().Deserialize<string>(json); // Noncompliant: default constructor is not safe
+            new JavaScriptSerializer().Deserialize<string>(json); // Noncompliant {{Restrict types of objects allowed to be deserialized.}}
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Yes I think it makes sense to have IT's with this. I've added two test cases with an unsafe binder and one unsafe resolver in their separate files but currently our script which checks differences will not look at secondary locations.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
```diff
@@ -77,38 +87,47 @@ private sealed class SerializationBinderCheck : ExplodedGraphCheck
...
                 if (instruction is ObjectCreationExpressionSyntax objectCreation)
                 {
                     var typeSymbol = semanticModel.GetTypeInfo(instruction).Type;
-                    if (IsFormatterWithBinder(typeSymbol) &&
-                        !HasSafeBinderInitialization(objectCreation.Initializer))
+                    if (IsFormatterWithBinder(typeSymbol))
                     {
-                        programState = programState.SetConstraint(symbolicValue, SerializationConstraint.Unsafe);
+                        var declarationInfo = GetBinderDeclaration(objectCreation.Initializer);
+                        if (!declarationInfo.IsSafe)
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: The condition is needed here to be sure we add the constraint only when the declaration is not safe. I've removed the condition from `AddSecondaryLocation` and moved it to 249.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/RestrictDeserializedTypes.cs
costin-zaharia-sonarsource: But this would mean to replace 2 lines with 10. In my opinion the code is more readable in the current form.

PR #3389, opened 2020-06-19
Rule S5766: Insecure deserialization

### LINKED ISSUE #3283: S5766 Deserializing objects without validating data is security-sensitive
Implements https://jira.sonarsource.com/browse/RSPEC-5766 (C# specific RSPEC [here](https://jira.sonarsource.com/browse/RSPEC-5767)). Part of MMF-2064.

For each `serializable` class should raise when:
1.  The class doesn't inherit from `ISerializable` or `IDeserializationCallback` and at least one constructor performs checks on its arguments
```
[Serializable]
public class Foo
{
    private int n;

    public Foo(int _n)  // Sensitive: mainlocation
    {
       if(_n > 10) { // there is some conditions on arguments
          n = 10;
       }
       else {
         n = _n;
       }
    }
```
2. if the class inherit from `ISerializable` but has a regular constructor with parameters and these last are used in conditions and the special constructor(`SerializationInfo` info, `StreamingContext` context) (used after deserialization) doesn't perform checks (conditions):
```
[Serializable]
public class Foo : ISerializable 
{
    private int n;

    public Foo(int _n)  // Sensitive: secondary location
    {
       if(_n > 10) { // there is some conditions on arguments
          n = 10;
       }
       else {
         n = _n;
       }
    }
    protected Foo(SerializationInfo info, StreamingContext context) // Sensitive: main location
    {
       n = (int)info.GetValue("n", typeof(int)); // the same validation is not performed after deserialization
     }
}
```
3. if the class inherit from `IDeserializationCallback` but has a regular constructor with parameters and these last are used in conditions and the method `IDeserializationCallback.OnDeserialization` (used after deserialization) doesn't perform checks (conditions):
```
[Serializable]
public class Foo : IDeserializationCallback
{
    private int n;

    public Foo(int _n)  // Sensitive: secondary location
    {
       if(_n > 10) { // there is some conditions on arguments
          n = 10;
       }
       else {
         n = _n;
       }
    }
    void IDeserializationCallback.OnDeserialization(Object sender)  // Sensitive: main location
    {
        
    }
```

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/dotnet/wip/basic-rules/NetCore3MVC/Controllers/RSPEC_5766

costin-zaharia-sonarsource:
Fixes #3283


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InsecureDeserialization.cs
```diff
@@ -0,0 +1,283 @@
...
+
+            private static ISymbol GetParameterSymbols(SyntaxNode parameterSyntax, SemanticModel semanticModel) =>
+                semanticModel.GetDeclaredSymbol(parameterSyntax);
+        }
+
+        /// <summary>
+        /// This walker is responsible to visit all conditional structures and check if a list of parameters
+        /// are used or not.
+        /// </summary>
+        private sealed class ConditionalsWalker : CSharpSyntaxWalker
```
christophe-zurn-sonarsource, line 208: Question regarding switch statement support also relates to `ConditionalsWalker`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InsecureDeserialization.cs
```diff
@@ -0,0 +1,339 @@
...
+        {
+        }
+    }
+
+    [Serializable]
+    public class WithoutConditionsInCtor : IDeserializationCallback
+    {
+        private string Name { get; }
+
+        public WithoutConditionsInCtor()
```
christophe-zurn-sonarsource, line 270: If we add a parameter `string name` and use `Name = name`, there will be a FP (see following comment)

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InsecureDeserialization.cs
```diff
@@ -0,0 +1,283 @@
...
+                foreach (var ctorInfo in walker.GetConstructorsInfo().Where(info => !info.IsDeserializationConstructor && info.HasConditionalConstructs))
+                {
+                    context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, ctorInfo.Declaration.GetLocation()));
+                }
+            }
+
+            if (implementsIDeserializationCallback &&
+                !OnDeserializationHasConditions(classDeclaration, context.SemanticModel))
+            {
+                foreach (var ctorInfo in walker.GetConstructorsInfo())
```
christophe-zurn-sonarsource, line 94: Shouldn't we also filter on `.Where(info => info.HasConditionalConstructs)` to raise issue ? Won't this cause FP otherwise ?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InsecureDeserialization.cs
```diff
@@ -0,0 +1,283 @@
...
+            {
+                // If there are no constructors, or if these don't have parameters, there is no validation done
+                // and the class is considered safe.
+                return;
+            }
+
+            var typeSymbol = context.SemanticModel.GetDeclaredSymbol(classDeclaration);
+            if (!HasSerializableAttribute(typeSymbol))
+            {
+                return;
```
christophe-zurn-sonarsource, line 64: For completeness sake, we could add a test case that doesn't have `Serializable` attribute to show it does not raise issue

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ImplementISerializableCorrectly.cs
```diff
@@ -231,8 +231,8 @@ private static bool IsCallingBaseConstructor(IMethodSymbol constructorSymbol)
             typeSymbol.IsPubliclyAccessible() &&
             typeSymbol.AllInterfaces.Any(IsOrImplementsISerializable);
 
-        private static bool HasSerializableAttribute(ITypeSymbol typeSymbol) =>
-            typeSymbol.GetAttributes(KnownType.System_SerializableAttribute).Any();
+        private static bool HasSerializableAttribute(ISymbol typeSymbol) =>
+            typeSymbol.HasAttribute(KnownType.System_SerializableAttribute);
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Indeed. I will do a separate PR for this.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/ISymbolExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+using System.Linq;
+using Microsoft.CodeAnalysis;
+
+namespace SonarAnalyzer.Helpers
+{
+    internal static class ISymbolExtensions
+    {
+        public static bool HasAttribute(this ISymbol symbol, KnownType type) =>
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: I think it would be better to keep this file in this PR (since this is where we want to get) and do a separate PR in which we replace all the occurrences. I will do this PR today so we will not forget about this.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/SyntaxNodeExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using SonarAnalyzer.ShimLayer.CSharp;
+
+namespace SonarAnalyzer.Helpers
+{
+    internal static class SyntaxNodeExtensions
+    {
+        public static bool ContainsConditionalConstructs(this SyntaxNode node) =>
+            node.DescendantNodes()
+                .Any(descendant => descendant.IsAnyKind(SyntaxKind.IfStatement,
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: I also thought about them but I wasn't sure. I guess we could consider them conditionals.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InsecureDeserialization.cs
```diff
@@ -0,0 +1,283 @@
...
+                    context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, ctorInfo.Declaration.GetLocation()));
+                }
+            }
+        }
+
+        private static bool OnDeserializationHasConditions(TypeDeclarationSyntax typeDeclaration, SemanticModel semanticModel) =>
+            typeDeclaration
+                .Members
+                .OfType<MethodDeclarationSyntax>()
+                .Single(methodDeclaration => IsOnDeserialization(methodDeclaration, semanticModel))
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: This does not throw but I've added test cases with abstract classes which implement `ISerializable` or `IDeserializationCallback`.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/InsecureDeserialization.cs
costin-zaharia-sonarsource: I was so sure that I've added that and I've completely forgotten 😄

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/ISymbolExtensions.cs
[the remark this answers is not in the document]
christophe-zurn-sonarsource: Ok, let's keep that as is then :+1:

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ImplementISerializableCorrectly.cs
```diff
@@ -231,7 +231,7 @@ private static bool IsCallingBaseConstructor(IMethodSymbol constructorSymbol)
             typeSymbol.IsPubliclyAccessible() &&
             typeSymbol.AllInterfaces.Any(IsOrImplementsISerializable);
 
-        private static bool HasSerializableAttribute(ITypeSymbol typeSymbol) =>
+        private static bool HasSerializableAttribute(ISymbol typeSymbol) =>
```
christophe-zurn-sonarsource, line 234: While reverting the changes in this file, this change was still kept, although it is unrelated to the PR

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InsecureDeserialization.cs
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    [Serializable]
+    public class ClassWithDefaultConstructorIsSafe
+    {
+        public string Name { get; set; }
+
+        public ClassWithDefaultConstructorIsSafe()
+        {
+            Name = string.Empty;
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: I think it was :) I've decided to keep this class and add one additional test case `ClassWithDefaultConstructorWithConditionalsIsSafe` with the scenario you mentioned. you can find it here https://github.com/SonarSource/sonar-dotnet/blob/costin/S5766/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InsecureDeserialization.cs#L28

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ImplementISerializableCorrectly.cs
costin-zaharia-sonarsource: Indeed this seems more like a random change but I think it would be just a loss of time to make this back `ITypeSymbol`, trigger the azure job and wait for it to finish 😄

PR #3392, opened 2020-06-22
S4261: should not report for SignalR hub methods

### LINKED ISSUE #3378: S4261 should not report for SignalR hub methods
Similar to solved issue #1875 (Fix S4261: rule should not report for MVC controller methods) it would be great to have SonarSource not reporting when a SignalR hub method does not have the "Async" suffix.

costin-zaharia-sonarsource:
Fixes #3378

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodShouldBeNamedAccordingToSynchronicity.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodShouldBeNamedAccordingToSynchronicity.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Threading.Tasks;
+using Microsoft.AspNet.SignalR;
 
 namespace Tests.Diagnostics
 {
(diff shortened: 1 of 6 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodShouldBeNamedAccordingToSynchronicity.cs
```diff
@@ -140,4 +141,21 @@ public class TestAttributes
         [System.ComponentModel.Browsable(true)]
         public async Task OtherAttributes() { } // Noncompliant
     }
+
+    public class MyHub : Hub
```
andrei-epure-sonarsource, line 145: maybe also add some cases from [the docs](https://docs.microsoft.com/en-us/aspnet/core/signalr/hubs?view=aspnetcore-3.1)?

```
public interface IChatClient
{
    Task ReceiveMessage(string user, string message);
    Task ReceiveMessage(string message);
}

public class StronglyTypedChatHub : Hub<IChatClient>
{
    public async Task SendMessage(string user, string message)
    {
        await Clients.All.ReceiveMessage(user, message);
    }

    public Task SendMessageToCaller(string message)
    {
        return Clients.Caller.ReceiveMessage(message);
    }

	public override async Task OnConnectedAsync()
	{
	    await base.OnConnectedAsync();
	}

	public Task ThrowException()
	{
	    throw new HubException("This error will be sent to the client!");
	}
}

```

costin-zaharia-sonarsource: I've added also the test case you mentioned with a generic Hub.

PR #3399, opened 2020-06-24
Fix S3242 FP and FN: Consider double conditional access

### LINKED ISSUE #3397: Fix S3242 FP and FN: Consider double conditional access
### Description

S3242 is raised when variable and it's property have conditional access operator.

>Consider using more general type 'System.Exception' instead of 'System.ArgumentException'.

### Repro steps

False Positive:
```
public class Repro
{
    public void Go(System.ArgumentException ex)
    {
        var length = ex?.ParamName?.Length;
        var message = ex.Message;
    }
}
```

False Negative:
```
public class Repro
{
    public void Go(System.ArgumentException ex)
    {
        var length = ex?.Message?.Length;
    }
}
```

### Expected behavior

`ParamName` is specific to `ArgumentException` and issue should not be raised.

pavel-mikula-sonarsource:
Fixes #3397

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldUseBaseTypes.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldUseBaseTypes.cs
@@ -152,15 +152,14 @@ namespace SonarAnalyzer.Rules.CSharp
                 ? subsequentConditionalAccess.Expression
                 : conditionalAccess.WhenNotNull;
 
-            if (conditionalAccessExpression is MemberBindingExpressionSyntax binding)
+            if (conditionalAccessExpression is MemberBindingExpressionSyntax binding
+                && binding.Name != null)
             {
-                return binding.Name != null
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodsShouldUseBaseTypes.cs
```diff
@@ -342,9 +346,12 @@ ITypeSymbol FindMostGeneralAccessibleClassOrSelf(ITypeSymbol typeSymbol)
                         if (DerivesOrImplementsAll(currentSymbol))
                         {
                             typeSymbol = currentSymbol;
+                            currentSymbol = currentSymbol.BaseType;
```
pavel-mikula-sonarsource, line 349: This is just a performance optimization to end the loop early. There's no need to re-check base types with `DerivesOrImplementsAll` once we're sure that the type itself doesn't conform.

pavel-mikula-sonarsource:
@costin-zaharia-sonarsource I suggest to review it per commit, since there's only refactoring in the 2nd one.

pavel-mikula-sonarsource:
I converted this PR back to draft to prevent merging. It should be merged after release and version bump

PR #3401, opened 2020-06-25
Update rspec

costin-zaharia-sonarsource:
- a few security hot spots are now deprecated


sonaranalyzer-dotnet/rspec/cs/S5773_c#.html
```diff
@@ -17,7 +17,7 @@ <h2>Noncompliant Code Example</h2>
 href="https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.formatters.binary.binaryformatter?view=netframework-4.8">BinaryFormatter</a>,
 <a
 href="https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.netdatacontractserializer?view=netframework-4.8">NetDataContractSerializer
-</a>, <a href="https://docs.microsoft.com/en-us/dotnet/api/system.web.ui.objectstateformatter?view=netframework-4.8">ObjectStateFormatter</a>, <a
```
andrei-epure-sonarsource, line 20: just to double check - is this ok to be removed? is it according to the implementation?

sonaranalyzer-dotnet/rspec/vbnet/S2077_vb.net.html
```diff
@@ -68,19 +68,10 @@ <h2>Sensitive Code Example</h2>
...
-  This feature was added in Entity Framework Core 2.0. See <a
-  href="https://docs.microsoft.com/en-us/ef/core/what-is-new/ef-core-2.0#string-interpolation-in-fromsql-and-executesqlcommand">Entity Framework's
-  documentation</a> for more information. </li>
-</ul>
+<h2>Compliant Solution</h2>
 <pre>
 Public Sub Foo(ByVal context As DbContext, ByVal value As String)
-    context.Database.ExecuteSqlCommand("SELECT * FROM mytable") ' No issue raised. The query is hard-coded. Thus no injection is possible.
-    context.Database.ExecuteSqlCommand($"SELECT * FROM mytable WHERE mycol={value}") ' No issue raised. The FormattableString is transformed into a parametrized query.
+    context.Database.ExecuteSqlCommand($"SELECT * FROM mytable WHERE mycol=@p0", param) ' Compliant, the query is parametrized
```
andrei-epure-sonarsource, line 74: is the interpolated string necessary here? it's using `@p0` rather than `{value}` as before

sonaranalyzer-dotnet/rspec/vbnet/S4507_vb.net.html
```diff
@@ -22,29 +22,38 @@ <h2>Recommended Secure Coding Practices</h2>
 <h2>Sensitive Code Example</h2>
 <p>This rule raises issues when the following .Net Core methods are called:
 <code>Microsoft.AspNetCore.Builder.IApplicationBuilder.UseDeveloperExceptionPage</code>,
-<code>Microsoft.AspNetCore.Builder.IApplicationBuilder.UseDatabaseErrorPage</code>. No Issue is raised when those calls are disabled by <code>if
```
andrei-epure-sonarsource, line 25: do we need to do a code change for `env.IsDevelopment()`?

sonaranalyzer-dotnet/rspec/cs/S5773_c#.html
costin-zaharia-sonarsource: Yes. This formatter is not using a binder. It has the property since it implements `IFormatter` but the property set will just ignore the received value.

sonaranalyzer-dotnet/rspec/vbnet/S2077_vb.net.html
costin-zaharia-sonarsource: Indeed there are multiple problems here:
- $ is not needed
- the parameter is called **value** and not **param**
- typo at **parametrized**

I've updated the docs and regenerated the rpsec.

sonaranalyzer-dotnet/rspec/vbnet/S4507_vb.net.html
costin-zaharia-sonarsource: Indeed, this was not implemented yet. I will revert the changes.

PR #3410, opened 2020-07-06
Update analyzers target to .Net Standard 2.0

costin-zaharia-sonarsource:
This commit will update to .Net Standard 2.0 the following projects:
- SonarAnalyzer.CFG
- SonarAnalyzer.Common
- SonarAnalyzer.CSharp
- SonarAnalyzer.Utilities
- SonarAnalyzer.VisualBasic

and to .Net 4.6.1 the following:
- SonarAnalyzer.RuleDescriptorGenerator
- CBDEArguments
- CBDEFails
- CBDESucceedsWithIncorrectResults
- CBDEWaitAndSucceeds

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/packages.lock.json
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/packages.lock.json
@@ -1,7 +1,7 @@
 {
   "version": 1,
   "dependencies": {
-    ".NETFramework,Version=v4.6.1": {
+    ".NETStandard,Version=v2.0": {
       "Microsoft.CodeAnalysis.CSharp.Workspaces": {
         "type": "Direct",
         "requested": "[1.3.2, )",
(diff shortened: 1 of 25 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/packages.lock.json
```diff
@@ -101,214 +100,772 @@
...
+          "System.Runtime.InteropServices": "4.1.0"
+        }
       },
       "System.Diagnostics.StackTrace": {
         "type": "Transitive",
         "resolved": "4.0.1",
-        "contentHash": "6i2EbRq0lgGfiZ+FDf0gVaw9qeEU+7IS2+wbZJmFVpvVzVOgZEt0ScZtyenuBvs6iDYbGiF51bMAa0oDP/tujQ=="
+        "contentHash": "6i2EbRq0lgGfiZ+FDf0gVaw9qeEU+7IS2+wbZJmFVpvVzVOgZEt0ScZtyenuBvs6iDYbGiF51bMAa0oDP/tujQ==",
+        "dependencies": {
+          "System.Collections.Immutable": "1.2.0",
```
andrei-epure-sonarsource, line 219: this shouldn't affect the `System.Collections.Immutable" : 1.1.37` that we specify explicitly , right?

I believe we have a regression test for vanilla .NET 4.6, right? that should cover the VS2015 update3 case

sonaranalyzer-dotnet/src/SonarAnalyzer.RuleDescriptorGenerator/Descriptor.cs.nuspec
```diff
@@ -17,7 +17,7 @@
   </metadata>
   <files>
     <!-- If these are changed than the build\release.ps1 should also be changed -->
```
andrei-epure-sonarsource, line 19: is this comment still actual? I don't see any `release.ps` in the changelist

costin-zaharia-sonarsource: I think we can remove the comment. We don't have a `release.ps1` file in the project.

sonaranalyzer-dotnet/src/SonarAnalyzer.CFG/packages.lock.json
costin-zaharia-sonarsource: This is a dependency of a dependency (`System.Diagnostics.StackTrace`). In our case, 1.1.37 version will be used since we reference it directly.

PR #3423, opened 2020-07-09
Migrate RuleDescriptorGenerator to .NetCore 3.1

costin-zaharia-sonarsource:

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.RuleDescriptorGenerator/SonarAnalyzer.RuleDescriptorGenerator.csproj
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.RuleDescriptorGenerator/SonarAnalyzer.RuleDescriptorGenerator.csproj
@@ -1,7 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net461</TargetFramework>
     <LangVersion>8.0</LangVersion>
     <AssemblyName>SonarAnalyzer.RuleDescriptorGenerator</AssemblyName>
     <RootNamespace>SonarAnalyzer.RuleDescriptorGenerator</RootNamespace>
@@ -9,6 +8,7 @@
     <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
     <RestoreLockedMode>true</RestoreLockedMode>
     <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
+    <TargetFrameworks>net48;netcoreapp3.1</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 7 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.RuleDescriptorGenerator/SonarAnalyzer.RuleDescriptorGenerator.csproj
```diff
@@ -1,14 +1,14 @@
...
   <PropertyGroup>
-    <TargetFramework>net461</TargetFramework>
     <LangVersion>8.0</LangVersion>
     <AssemblyName>SonarAnalyzer.RuleDescriptorGenerator</AssemblyName>
     <RootNamespace>SonarAnalyzer.RuleDescriptorGenerator</RootNamespace>
     <OutputType>Exe</OutputType>
     <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
     <RestoreLockedMode>true</RestoreLockedMode>
     <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
+    <TargetFrameworks>net48;netcoreapp3.1</TargetFrameworks>
```
costin-zaharia-sonarsource, line 11: We need to target both .Net Framework and .Net Core because it is referenced by tests which are also targeting .Net Framework and .Net Core. Since it's a console app we cannot target .Net Standard.

In the long term, I think it would be better to split the test project we have now and add a dedicated test project for `RuleDescriptorGenerator`.

andrei-epure-sonarsource: well, I guess those tests just test the behavior of this class, right?

> In the long term, I think it would be better to split the test project we have now and add a dedicated test project for RuleDescriptorGenerator.

+1

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/MetadataReferences/FrameworkMetadataReference.cs
```diff
@@ -19,25 +19,14 @@
...
 using Microsoft.CodeAnalysis;
 
-namespace SonarAnalyzer.UnitTest
+using static SonarAnalyzer.UnitTest.MetadataReferences.MetadataReferenceFactory;
+
+namespace SonarAnalyzer.UnitTest.MetadataReferences
 {
     internal static class FrameworkMetadataReference
     {
-        #region Helpers
```
andrei-epure-sonarsource, line 30: I guess you'll need to rebase

costin-zaharia-sonarsource: ✔️

PR #3429, opened 2020-07-10
Fix build errors when using VS2015

costin-zaharia-sonarsource:
Visual Studio 2015 does not support .Net Standard out of the box.

In order to keep compatibility we will target both .Net 4.6.1 and .Net Standard 2.0 but use only .Net 4.6.1 in plugin.


azure-pipelines.yml
```diff
@@ -57,15 +57,6 @@ stages:
...
 
-      - task: Cache@2
-        displayName: Cache NuGet packages
-        inputs:
-          key: 'nuget | "$(Agent.OS)" | **/packages.lock.json'
-          restoreKeys: |
-            nuget | "$(Agent.OS)"
-            nuget
-          path: $(NUGET_PACKAGES)
-
```
costin-zaharia-sonarsource, line 68: I've removed caching since this needs **RestoreLockedMode** and this does not work when targeting both .Net 4.6.1 and .Net Standard

sonaranalyzer-dotnet/src/SonarAnalyzer.Vsix/SonarAnalyzer.Vsix.csproj
```diff
@@ -96,12 +96,16 @@
       <Name>SonarAnalyzer.VisualBasic</Name>
       <IncludeOutputGroupsInVSIX>BuiltProjectOutputGroup</IncludeOutputGroupsInVSIX>
       <IncludeOutputGroupsInVSIXLocalOnly>DebugSymbolsProjectOutputGroup</IncludeOutputGroupsInVSIXLocalOnly>
+      <!-- https://github.com/dotnet/sdk/issues/433 -->
+      <AdditionalProperties>TargetFramework=net461</AdditionalProperties>
```
costin-zaharia-sonarsource, line 100: As described in the linked issue, multi-targeting .net core projects cannot be referenced by VSIX projects. I order to work around this we have to filter the target framework by using **AdditionalProperties**

azure-pipelines.yml
andrei-epure-sonarsource: do you think it's worth letting the azure pipelines team know of this?

sonaranalyzer-dotnet/src/SonarAnalyzer.VisualBasic/SonarAnalyzer.VisualBasic.csproj
```diff
@@ -1,13 +1,11 @@
...
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
     <LangVersion>8.0</LangVersion>
     <AssemblyName>SonarAnalyzer.VisualBasic</AssemblyName>
     <RootNamespace>SonarAnalyzer</RootNamespace>
-    <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
-    <RestoreLockedMode>true</RestoreLockedMode>
     <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
+    <TargetFrameworks>net461;netcoreapp2.0</TargetFrameworks>
```
andrei-epure-sonarsource, line 8: why not netstandard?

costin-zaharia-sonarsource: Indeed. It should be .Net Standard.

azure-pipelines.yml
costin-zaharia-sonarsource: I think nuget is the culprit here since I can reproduce this also locally. Indeed it would be nice to report an issue.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
```diff
@@ -1,13 +1,11 @@
...
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
     <LangVersion>8.0</LangVersion>
     <AssemblyName>SonarAnalyzer.CSharp</AssemblyName>
     <RootNamespace>SonarAnalyzer</RootNamespace>
-    <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
-    <RestoreLockedMode>true</RestoreLockedMode>
     <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
+    <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
```
andrei-epure-sonarsource, line 8: another nitpick observation is that we're now inconsistent where we put the TargetFrameworks - here it's at the end of the PropertyGroup, in the other csproj (like CFG or common, which haven't changed) it's at the beginning

I don't care, TBH, but it's just a comment. I don't mind leaving it as it is.

costin-zaharia-sonarsource: I've used the project config UI from Rider. Not sure it matters that much.

PR #3432, opened 2020-07-13
Add NetCore31 and Net5 ITs

### LINKED ISSUE #3414: Create integration test with C# 9

andrei-epure-sonarsource:
Fixes #3414


andrei-epure-sonarsource:
I have to investigate why it complains about the SDK, it may be the fact that it's a different pipe than the one I modified in my previous PR.

However, the ps1 scripts can be reviewed and I can apply the early comments

sonaranalyzer-dotnet/its/create-issue-reports.ps1
```diff
@@ -142,13 +143,23 @@ function New-IssueReports([string]$sarifReportPath) {
...
                     ForEach-Object { [System.Text.RegularExpressions.Regex]::Unescape($_) }  # Unescape powershell to json automatic escape
-                ) -split "`r`n"										            	         # Convert JSON to String and split lines
+                ) -split "`r`n"                                                              # Convert JSON to String and split lines
             ) | Foreach-Object { $_.TrimStart() }                                            # Remove leading spaces
-        Set-Content $path $lines
+
+        if (Test-Path $path) {
+            Set-Content $path $lines
+        } else {
+            Write-Host "The path '${path}' does not exist, will first create the file."
```
costin-zaharia-sonarsource, line 160: This is a nice change but do you know when the path does not exist? It's a bit strange that this didn't occurred until now.

sonaranalyzer-dotnet/its/regression-test.ps1
```diff
@@ -53,6 +53,39 @@ function Build-Project([string]$ProjectName, [string]$SolutionRelativePath, [int
...
+
+    New-Item -ItemType directory -Path .\output\$ProjectName | out-null
+
+    $solutionPath = Resolve-Path ".\sources\${ProjectName}\${SolutionRelativePath}"
+
+    # The PROJECT env variable is used by 'SonarAnalyzer.Testing.ImportBefore.targets'
+    Write-Debug "Setting PROJECT environment variable to '${ProjectName}'"
+    $Env:PROJECT = $ProjectName
+
+    # TODO create the global.json based on target SDK version (passed by parameter) and then clean it in this function
```
costin-zaharia-sonarsource, line 70: This To do will be done as part of this PR or should we create an issue for it?

sonaranalyzer-dotnet/its/regression-test.ps1
```diff
@@ -53,6 +53,39 @@ function Build-Project([string]$ProjectName, [string]$SolutionRelativePath, [int
...
+    $globalJsonPath = Resolve-Path ".\sources\${ProjectName}\global.json"
+    Copy-Item $globalJsonPath .
+
+    # The PROJECT env variable is used by 'SonarAnalyzer.Testing.ImportBefore.targets'
+    Write-Debug "Setting PROJECT environment variable to '${ProjectName}'"
+
+    $Env:PROJECT = $ProjectName
```
costin-zaharia-sonarsource, line 77: This seems to be done twice (here and on line 66-68).

sonaranalyzer-dotnet/its/regression-test.ps1
```diff
@@ -53,6 +53,39 @@ function Build-Project([string]$ProjectName, [string]$SolutionRelativePath, [int
...
+    dotnet restore $solutionPath
```
costin-zaharia-sonarsource, line 78: When using `dotnet` tool you don't have to call `restore` since this runs implicitly. More details here: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-build#implicit-restore

sonaranalyzer-dotnet/its/regression-test.ps1
```diff
@@ -53,6 +53,39 @@ function Build-Project([string]$ProjectName, [string]$SolutionRelativePath, [int
...
+    dotnet msbuild $solutionPath `
```
costin-zaharia-sonarsource, line 79: If we build a .Net Core project (or .Net 😄) we can call `dotnet build` command instead of `dotnet msbuild`

sonaranalyzer-dotnet/its/sources/Net5/Main.cs
```diff
@@ -0,0 +1,13 @@
...
+Console.WriteLine("Hello World!");
+MyClass.Foo();
+
+var x = new MyClass.PublicPerson { FirstName = "C#", LastName = "9" };
+Console.WriteLine(x.FirstName + x.LastName);
+
+LocalRecord y = new("1", "2");
+Console.WriteLine(y.A + y.B);
+
+public record LocalRecord(string A, string B);
```
costin-zaharia-sonarsource, line 13: Nice! 🎉

sonaranalyzer-dotnet/its/update-expected.ps1
```diff
@@ -10,6 +19,19 @@ try {
...
+        $projectPath = ".\expected\${project}"
+        if (Test-Path $projectPath) {
+            Remove-Item $projectPath  -Recurse -Force
+        }
+        Copy-Item -Path ".\actual\${project}" -Destination  ".\expected" -Force -Recurse
+        if (Test-Path .\diff) {
+            Remove-Item .\diff -Recurse -Force
+        }
+        return
+    }
```
costin-zaharia-sonarsource, line 33: We could extract 2 methods: one which updates all projects and one which updates just the specified one and call these inside the try, after validation.

sonaranalyzer-dotnet/its/create-issue-reports.ps1
andrei-epure-sonarsource: Yes, it happens when adding a new test project:
- when adding a new test project, the "expected" folder doesn't exist yet
- the `actual` folder is created in [function Initialize-ActualFolder()](https://github.com/SonarSource/sonar-dotnet/blob/a11a3dca45803ae093705ceb5de90cc28af36468/sonaranalyzer-dotnet/its/regression-test.ps1#L89) where it gets initialized with the contents of the "expected" folder
- after running the test, the existing files are overwritten with `Set-Content $path $lines` - if the `$path` does not exist, it fails. And it doesn't exist, because the "expected" folder doesn't exist, so no files could be copied.

andrei-epure-sonarsource: Actually inside `Initialize-ActualFolder` I see `# this copies no files if ruleId is not set, and all but ending with ruleId if set`.

And I tested now, and `Initialize-ActualFolder` only creates empty folders if the tests are running for all rules.

And then here in `New-IssueReports` the files get created. So here, we only need to make sure that the folder exists (because it's not created in `Initialize-ActualFolder`), and if it doesn't exist, create it.

sonaranalyzer-dotnet/its/regression-test.ps1
andrei-epure-sonarsource: I forgot about it. I will do it now.

sonaranalyzer-dotnet/its/regression-test.ps1
andrei-epure-sonarsource: (I am using `dotnet msbuild` which doesn't do all the magic of `dotnet build`)

azure-pipelines.yml
```diff
@@ -38,25 +38,6 @@ stages:
       - task: NuGetToolInstaller@1
         displayName: "Install NuGet"
 
-      - task: UseDotNet@2
```
andrei-epure-sonarsource, line 41: this was where we only build the project and don't run UTs or ITs.

azure-pipelines.yml
```diff
@@ -269,6 +250,25 @@ stages:
       - task: NuGetToolInstaller@1
         displayName: "Install NuGet"
 
+      - task: UseDotNet@2
```
andrei-epure-sonarsource, line 253: And in commit 7442d39c01f687f5ef02bfa706cbc I added `UseDotnet` under the job that was running the UTs, and now I add under the job that is running the ITs (they run in parallel)

sonaranalyzer-dotnet/its/expected/NetCore31/NetCore31--S3990.json
```diff
@@ -2,7 +2,7 @@
 "issues":  [
 {
 "id":  "S3990",
-"message":  "Provide a 'CLSCompliant' attribute for assembly 'NetCore31'.",
+"message":  "Mark this assembly with 'System.CLSCompliantAttribute'",
```
andrei-epure-sonarsource, line 5: this seems like a problem, it's a bit weird

andrei-epure-sonarsource:
@costin-zaharia-sonarsource I took the opportunity to update the regression-test.ps1 script to stop doing a silent diff, and do a restricted diff based on the parameters (until now, if you were running the tests just for one project, it would complain that there are no found issues for the rest of the project; it was doing the comparison blindly)

sonaranalyzer-dotnet/its/regression-test.ps1
```diff
@@ -186,7 +186,7 @@ function Copy-FolderRecursively($From, $To, $Include, $Exclude) {
         if (-Not (Test-Path $parent)) {
             New-Item $parent -Type Directory -Force | out-null
         }
-        Copy-Item $file -Destination $path
+        Copy-Item $file.FullName -Destination $path
```
andrei-epure-sonarsource, line 189: this is needed now - I am not sure why, but it was failing when copying the files just for one project

sonaranalyzer-dotnet/its/regression-test.ps1
```diff
@@ -211,22 +211,45 @@ function Show-DiffResults() {
...
     $errorMsg = "ERROR: There are differences between the actual and the expected issues."
-    if (!$ruleId)
+
+
+    if (!$ruleId -And !$project)
     {
-        Write-Debug "Running 'git diff' between 'actual' and 'expected'"
+        Write-Host "Will find differences for all projects, all rules."
+
+        Write-Debug "Running 'git diff' between 'actual' and 'expected'."
```
andrei-epure-sonarsource, line 220: I think I can remove this now...

sonaranalyzer-dotnet/its/regression-test.ps1
```diff
@@ -211,22 +211,45 @@ function Show-DiffResults() {
...
 
+    # do a partial diff
     New-Item ".\diff" -Type Directory | out-null
     New-Item ".\diff\actual" -Type Directory | out-null
     New-Item ".\diff\expected" -Type Directory | out-null
 
-    Copy-FolderRecursively -From .\expected -To .\diff\expected -Include "*${ruleId}.json"
-    Copy-FolderRecursively -From .\actual   -To .\diff\actual   -Include "*${ruleId}.json"
+    if (!$ruleId -And $project) {
+        Write-Host "Will find differences for '${project}', all rules."
```
andrei-epure-sonarsource, line 234: this is new - manually tested

sonaranalyzer-dotnet/its/regression-test.ps1
```diff
@@ -211,22 +211,45 @@ function Show-DiffResults() {
...
+
+        Copy-FolderRecursively -From ".\expected\${project}" -To .\diff\expected
+        Copy-FolderRecursively -From ".\actual\${project}"   -To .\diff\actual
+
+    } elseif ($ruleId -And !$project) {
+        Write-Host "Will find differences for all projects, rule ${ruleId}."
```
andrei-epure-sonarsource, line 240: this was here before

sonaranalyzer-dotnet/its/regression-test.ps1
```diff
@@ -211,22 +211,45 @@ function Show-DiffResults() {
...
+
+        Copy-FolderRecursively -From .\expected -To .\diff\expected -Include "*${ruleId}.json"
+        Copy-FolderRecursively -From .\actual   -To .\diff\actual   -Include "*${ruleId}.json"
+
+    } else {
+        Write-Host "Will find differences for '${project}', rule ${ruleId}."
```
andrei-epure-sonarsource, line 246: this is new - manually tested

PR #3434, opened 2020-07-14
S4261: FP on methods which return IAsyncEnumerable<T>

### LINKED ISSUE #3433: S4261: FP on methods which return IAsyncEnumerable<T>

costin-zaharia-sonarsource:
Fixes #3433

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNamedAccordingToSynchronicity.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNamedAccordingToSynchronicity.cs
@@ -19,7 +19,6 @@
  */
 
 using System.Collections.Immutable;
-using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -42,12 +41,13 @@ namespace SonarAnalyzer.Rules.CSharp
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
 
-        private static readonly ImmutableArray<KnownType> TaskTypes =
+        private static readonly ImmutableArray<KnownType> asyncReturnTypes =
             ImmutableArray.Create(
                 KnownType.System_Threading_Tasks_Task,
                 KnownType.System_Threading_Tasks_Task_T,
                 KnownType.System_Threading_Tasks_ValueTask, // NetCore 2.2+
-                KnownType.System_Threading_Tasks_ValueTask_TResult
+                KnownType.System_Threading_Tasks_ValueTask_TResult,
+                KnownType.System_Collections_Generic_IAsyncEnumerable_T
             );
 
         protected override void Initialize(SonarAnalysisContext context) =>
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodShouldBeNamedAccordingToSynchronicity.CSharp8.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Tests.Diagnostics
+{
+    public abstract class IEnumerableReproducer
+    {
+        protected abstract IAsyncEnumerable<int> GetFilesAsync(string folderName); // Compliant (https://github.com/SonarSource/sonar-dotnet/issues/3433)
```
andrei-epure-sonarsource, line 8: maybe also add a class that implements `IAsyncEnumerable` ?  and make sure that works?

costin-zaharia-sonarsource: Good catch. This worked fine for types which derived from tasks since `.DerivesFromAny(KnownType)` was used but `IAsyncEnumerable` is an interface..

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNamedAccordingToSynchronicity.cs
```diff
@@ -72,24 +73,37 @@ public sealed class MethodShouldBeNamedAccordingToSynchronicity : SonarDiagnosti
...
+        private static bool HasAsyncReturnType(IMethodSymbol methodSymbol)
+        {
+            var returnSymbol = (methodSymbol.ReturnType as INamedTypeSymbol)?.ConstructedFrom;
+            if (returnSymbol == null)
+            {
+                return false;
+            }
+
+            return returnSymbol.DerivesFromAny(asyncReturnTypes) ||
+                   returnSymbol.IsAny(asyncReturnInterfaces) ||
```
costin-zaharia-sonarsource, line 99: `.DerivesFromAny` checks also the equality between the type parameter and known type but, for interfaces, `.ImplementsAny` is not doing that and we have to check the equality using `IsAny`.

andrei-epure-sonarsource: hmm, this is confusing...

it would be nice to check where we're using `ImplementsAny` and find errors , but not sure we should open a ticket for it

PR #3436, opened 2020-07-14
Add option to run tests on .Net 5 (preview)

costin-zaharia-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SocketsCreation.vb
```diff
@@ -32,7 +32,7 @@ Namespace Tests.TestCases
     Public Class MySocket
         Inherits Socket
         Sub New()
-            MyBase.New(Nothing)
+            MyBase.New(CType(Nothing, SocketInformation))
```
costin-zaharia-sonarsource, line 35: Constructor overload resolution fails under .Net 5 since a new constructor was added and calling the base with Nothing will match more than one.

https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.-ctor?view=net-5.0#overloads

azure-pipelines.yml
```diff
@@ -153,11 +153,6 @@ stages:
           dotnet --version
         displayName: 'List dotnet versions and current version'
 
-      - task: NuGetCommand@2
-        displayName: "Restore NuGet packages"
-        inputs:
-          restoreSolution: '$(solution)'
-
```
costin-zaharia-sonarsource, line 160: This is done by dotnet restore.

sonaranalyzer-dotnet/NuGet.Config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
costin-zaharia-sonarsource, line 1: The NuGet config file is needed when restore runs on Azure Devops builds. The [nuget.org](nuget.org) source is not added by default when using `nuget restore` or `dotnet build/restore` commands.

sonaranalyzer-dotnet/its/ReviewDiffs/ReviewDiffs.csproj
```diff
@@ -5,6 +5,7 @@
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <LangVersion>8.0</LangVersion>
     <SonarQubeExclude>true</SonarQubeExclude>
+    <ProjectGuid>{7e3adcd6-01d7-44e2-b12f-c485c35c8a9c}</ProjectGuid>
```
costin-zaharia-sonarsource, line 8: `sonnarscanner` is complaining about missing project guids when `dotnet build` is executed.

sonaranalyzer-dotnet/NuGet.Config
andrei-epure-sonarsource: like we discussed offline, `sonar-dotnet\sonaranalyzer-dotnet\tests\SonarAnalyzer.UnitTest\NuGet.config` should be harmonized with this one

azure-pipelines.yml
```diff
@@ -213,18 +208,26 @@ stages:
...
+
+          $testProjFileName = "tests\SonarAnalyzer.UnitTest\SonarAnalyzer.UnitTest.csproj"
+
+          Write-Host "Add .Net 5 target for UnitTest project"
+          ((Get-Content -Path $testProjFileName -raw) -Replace "net48;netcoreapp3.1","net48;netcoreapp3.1;net5") | Set-Content -Path $testProjFileName
+
+          Write-Host "Build UnitTest project"
+          dotnet build $testProjFileName -c $(BuildConfiguration)
+
+          Write-Host "Run tests on .Net 4.8 and compute coverage"
```
andrei-epure-sonarsource, line 222: on second thought - this is confusing - it says running tests on .NET 4.8 but it's using core dotnet 3.1

```
Run tests on .Net 4.8 and compute coverage
Executing: C:\Program Files\dotnet\dotnet.exe

Welcome to .NET Core 3.1!
---------------------
SDK Version: 3.1.301
```

could you improve this message in the logs?

andrei-epure-sonarsource: it seems it's still there - [see lines 2933-2939](https://sonarsource.visualstudio.com/DotNetTeam%20Project/_build/results?buildId=19772&view=logs&jobId=645a1db5-d210-5b98-34db-87d2a8fd8bb0&j=23c99647-d7e1-5b83-b5fb-f7e97f8a9ce7&t=0075a4ce-4d44-5a4c-36ef-97ab11b0917b)

costin-zaharia-sonarsource: I've also noticed that. Not sure why since I've used `--nologo` and for the others it works. Not sure also how we can get rid of it.

andrei-epure-sonarsource: well meanwhile you can modify the message to say that it's testing the net48 binary (running on .net core 3.1)

PR #3438, opened 2020-07-14
Change the target framework back to .Net 4.6

costin-zaharia-sonarsource:


sonaranalyzer-dotnet/tests/CBDE/CBDEArguments/CBDEArguments.csproj
```diff
@@ -9,7 +9,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>CBDEArguments</RootNamespace>
     <AssemblyName>CBDEArguments</AssemblyName>
-    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
```
andrei-epure-sonarsource, line 12: we can ignore the tests/CBDE/* projects as they are here just to be able to re-create the executables used in our tests

costin-zaharia-sonarsource: Indeed. I've set the same version for them to be consistent but I'm still unsure what to do with them. Maybe we should migrate to .Net Core 3.1...

andrei-epure-sonarsource: yes, at least the csproj files would be smaller

Issue #3442, opened 2020-07-15
S4275 FP with WPF DependencyProperty naming pattern


andrei-epure-sonarsource:
### Description

WPF has "Dependency Properties" - see [detailed docs.](https://docs.microsoft.com/en-us/dotnet/framework/wpf/advanced/dependency-properties-overview#dependency-property-value-precedence)

> A property that is backed by the WPF property system is known as a dependency property.

And from [Dependency properties back CLR properties](https://docs.microsoft.com/en-us/dotnet/framework/wpf/advanced/dependency-properties-overview#dependency-properties-back-clr-properties)

> Dependency property: A property that is backed by a DependencyProperty.
>
> Dependency property identifier: A DependencyProperty instance, which is obtained as a return value when registering a dependency property, and then stored as a static member of a class. 

**And**:

> The naming convention of the property and its backing DependencyProperty field is important. The name of the field is always the name of the property, with the suffix Property appended. 

Thus the following pattern is needed:

```
public static readonly DependencyProperty ToolTipProperty = DependencyProperty.Register("ToolTip", typeof(string), typeof(MyObject), new PropertyMetadata(new PropertyChangedCallback(ToolTipPropertyChanged)));

public string ToolTip
{
    get
    {
        return (string)GetValue(ToolTipProperty);
    }
    set
    {
        SetValue(ToolTipProperty, value); // Noncompliant FP
    }
}
```

From community: https://community.sonarsource.com/t/c-dependencyproperty-flagged-by-s4275-getters-and-setters-should-access-the-expected-fields/26739

### Expected behavior

The rule should realize that the `DependencyProperty` is being used here, and not raise.

### Known workarounds

None.

### Related information

* C#/VB.NET Plugins Version 8.9

andrei-epure-sonarsource:
Actually I cannot reproduce with:
```
    public class Repro3442 : System.Windows.Controls.Primitives.ButtonBase
    {
        public static readonly DependencyProperty IsSpinningProperty = DependencyProperty.Register("IsSpinning", typeof(Boolean),typeof(Repro3442));
        public bool IsSpinning
        {
            get { return (bool)GetValue(IsSpinningProperty); }
            set { SetValue(IsSpinningProperty, value); }
        }
    }

```

I'm waiting for an answer from the user on the community forum. Otherwise, this should be closed.

andrei-epure-sonarsource:
[...]

Issue #3441, opened 2020-07-15
S4275 FP with private setter and collection deserialization


andrei-epure-sonarsource:
### Description

This scenario should be valid for deserialization purposes.

Coming from the community - https://community.sonarsource.com/t/s4275-getters-and-setters-should-access-the-expected-fields/24581/3

```
    public class ReproDeserializationScenario
    {
        private readonly List<int> _data = new List<int>();
        public List<int> Data
        {
            get => _data;
            // The setter in this case is only executed by a deserializer, but we don’t want to replace the field instance.
            private set // Noncompliant FP
            {
                foreach (var item in value)
                {
                    _data.Add(item);
                }
            }
        }
    }

```

### Known workarounds

Do the deserialization using the constructor.

### Related information

* C#/VB.NET Plugins Version 8.9

andrei-epure-sonarsource:
note: this is a bit of an edge case, not sure it's worth adding complexity for the rule just for this

miminno:
So what's the status on this one? #3443 suggests that it's been addressed, yet we're still getting tons of these false-positives across our projects.

pavel-mikula-sonarsource:
Internal ticket [NET-1956](https://sonarsource.atlassian.net/browse/NET-1956)

[NET-1956]: https://sonarsource.atlassian.net/browse/NET-1956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3443, opened 2020-07-15
Add two repro cases for S4275

andrei-epure-sonarsource:
note that the second ticket is on hold as I cannot repro

related to #3441 and #3442


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PropertiesAccessCorrectField.cs
```diff
@@ -706,4 +708,33 @@ public int ReadOnlyProperty
...
+                foreach (var item in value)
+                {
+                    _data.Add(item);
+                }
+            }
+        }
+    }
+
+    // https://github.com/SonarSource/sonar-dotnet/issues/3442
+    public class Repro3442 : System.Windows.Controls.Primitives.ButtonBase
```
costin-zaharia-sonarsource, line 731: pedantic: not really a repro since it does the opposite, documents that the issue is not reproducible for dependency properties

Issue #3426, opened 2020-07-09
Add WinForms support in unit tests when targeting .Net Core


costin-zaharia-sonarsource:
Some tests, like `DoNotCallExitMethodsTest` reference `WinForms` types and currently cannot be run while targeting .Net Core.

andrei-epure-sonarsource:
Also see `PropertiesAccessCorrectFieldTest`

pavel-mikula-sonarsource:
Also `MarkWindowsFormsMainWithStaThread` for noncompliant example of TopLevelStatements

csaba-sagi-sonarsource:
`PropertiesAccessCorrectFieldTest` needs WPF support, #4883 was created to add supprot for WPF.

pavel-mikula-sonarsource:
WPF support can be scaffolded the same way as it is in #4765

PR #3445, opened 2020-07-16
Fix build break due to new CI configuration

andrei-epure-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PropertiesAccessCorrectField.cs
```diff
@@ -727,6 +727,7 @@ public List<int> Data
         }
     }
 
+#if NETFRAMEWORK // System.Windows.Controls.Primitives.ButtonBase is in a different assembly in NETCOREAPP
```
costin-zaharia-sonarsource, line 730: We could add a link to https://github.com/SonarSource/sonar-dotnet/issues/3426 since it's windows forms related. What do you think?

PR #3447, opened 2020-07-17
Add unit test with C#9 syntax

### LINKED ISSUE #3413: Create one Unit Test with C# 9 syntax in a separate project

costin-zaharia-sonarsource:
Fixes #3413

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest.CSharp9/EmptyMethodTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest.CSharp9/EmptyMethodTest.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
@@ -18,11 +18,30 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System.Reflection;
-using System.Runtime.CompilerServices;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarAnalyzer.Rules.CSharp;
+using SonarAnalyzer.UnitTest.TestFramework;
 
-[assembly: AssemblyTitle("SonarAnalyzer.UnitTest")]
-[assembly: AssemblyProduct("SonarAnalyzer.UnitTest")]
-[assembly: AssemblyDescription("")]
+namespace SonarAnalyzer.UnitTest.CSharp9
+{
+    [TestClass]
+    public class EmptyMethodTest
+    {
+        [TestMethod]
+        public void EmptyMethod_InsideRecord()
+        {
+            const string code = @"
+namespace CSharp9Playground
+{
+    public record Person
+    {
+        public void EmptyMethod() { } // Noncompliant
+    }
+}
+";
 
-[assembly: InternalsVisibleTo("SonarAnalyzer.UnitTest.CSharp9")]
+            Verifier.VerifyCSharpAnalyzer(code, new EmptyMethod(), new[] { new CSharpParseOptions(LanguageVersion.CSharp9) });
+        }
+    }
+}
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest.CSharp9/SonarAnalyzer.UnitTest.CSharp9.csproj
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest.CSharp9/SonarAnalyzer.UnitTest.CSharp9.csproj
@@ -1,0 +1,35 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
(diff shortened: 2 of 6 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest.CSharp9/SonarAnalyzer.UnitTest.CSharp9.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <Reference Include="SonarAnalyzer">
+      <HintPath>..\SonarAnalyzer.UnitTest\bin\Release\netcoreapp3.1\SonarAnalyzer.dll</HintPath>
+    </Reference>
+    <Reference Include="SonarAnalyzer.CFG">
+      <HintPath>..\SonarAnalyzer.UnitTest\bin\Release\netcoreapp3.1\SonarAnalyzer.CFG.dll</HintPath>
+    </Reference>
+    <Reference Include="SonarAnalyzer.CSharp">
+      <HintPath>..\SonarAnalyzer.UnitTest\bin\Release\netcoreapp3.1\SonarAnalyzer.CSharp.dll</HintPath>
+    </Reference>
+  </ItemGroup>
```
costin-zaharia-sonarsource, line 30: In order to be able to run the analysis on C# 9 code we have to reference `Microsoft.CodeAnalysis.CSharp` version `3.7.0-4.final` which is not compatible with the version used by the `SonarAnalyzer.CSharp`.

If we try to reference the analyzer as a project, NuGet will report the following error:
```
Error	NU1107	Version conflict detected for Microsoft.CodeAnalysis.Common. Install/reference Microsoft.CodeAnalysis.Common 3.7.0-4.final directly to project SonarAnalyzer.UnitTest.CSharp9 to resolve this issue. 
 SonarAnalyzer.UnitTest.CSharp9 -> Microsoft.CodeAnalysis.CSharp 3.7.0-4.final -> Microsoft.CodeAnalysis.Common (= 3.7.0-4.final) 
 SonarAnalyzer.UnitTest.CSharp9 -> SonarAnalyzer.CSharp -> Microsoft.CodeAnalysis.CSharp.Workspaces 1.3.2 -> Microsoft.CodeAnalysis.Workspaces.Common 1.3.2 -> Microsoft.CodeAnalysis.Common (= 1.3.2).	SonarAnalyzer.UnitTest.CSharp9	C:\src\sonar-dotnet\sonaranalyzer-dotnet\tests\SonarAnalyzer.UnitTest.CSharp9\SonarAnalyzer.UnitTest.CSharp9.csproj	1	
```

However, as far as I can tell at this point, this doesn't seem to be a problem at runtime if we reference the analyzer as a file.

andrei-epure-sonarsource: > However, as far as I can tell at this point, this doesn't seem to be a problem at runtime if we reference the analyzer as a file.

what do you mean reference the analyzer as a file?

andrei-epure-sonarsource: You mean reference the DLL vs reference the project?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest.CSharp9/EmptyMethodTest.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarAnalyzer.Rules.CSharp;
+
+namespace SonarAnalyzer.UnitTest.CSharp9
+{
+    [TestClass]
```
andrei-epure-sonarsource, line 34: it would be nice to document what work would be left to make it work with our test framework (that parses the `Noncompliant` comments etc)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest.CSharp9/SonarAnalyzer.UnitTest.CSharp9.csproj
andrei-epure-sonarsource: Yes, we don't need the analyzer as a project.

However, it would be nice to know that we can use our own testing framework (that parses files for `Noncompliant` comments)

costin-zaharia-sonarsource: Ok, I've tried to avoid to reference the test project directly but as a POC I guess we can do this. 

Before adding unit tests we should extract our testing framework (`Verifier` and related classes) in a separate assembly which can be shared between test projects.

andrei-epure-sonarsource:
The technical release for 8.10 [failed](https://github.com/SonarSource/sonar-dotnet/runs/882467192?check_suite_focus=true), so we should first clarify that on Monday before merging this.

PR #3454, opened 2020-07-21
Fix net5 project to build locally

andrei-epure-sonarsource:


sonaranalyzer-dotnet/its/regression-test.ps1
```diff
@@ -53,13 +53,15 @@ function Build-Project-MSBuild([string]$ProjectName, [string]$SolutionRelativePa
...
-function Build-Project-DotnetTool([string]$ProjectName, [string]$SolutionRelativePath, [string]$dotnetVersion) {
+function Build-Project-DotnetTool([string]$ProjectName, [string]$SolutionRelativePath) {
     if ($project -And -Not ($ProjectName -eq $project)) {
         Write-Host "Build skipped: $ProjectName"
         return
     }
 
-    Write-Host "Will build dotnet project: '${ProjectName}', version ${dotnetVersion}."
+    $projectGlobalJsonPath = ".\sources\${ProjectName}\global.json"
+    $globalJsonContent = $(Get-Content $projectGlobalJsonPath)
```
pavel-mikula-sonarsource, line 63: Once we add a `dotnet` project in the future without `global.json`, this will fail with 
>Cannot find path 'C:\Projects\sonar-dotnet\sonaranalyzer-dotnet\its\sources\NetCore31\global.json' because it does not exist.

Is that intended? 

We'll always have a `global.json` for our `Net5`, `Net6`, `Net7` internal projects, but we'll also add some open source projects that might not be explicit and might want `the latest`.

sonaranalyzer-dotnet/its/regression-test.ps1
```diff
@@ -83,7 +86,7 @@ function Build-Project-DotnetTool([string]$ProjectName, [string]$SolutionRelativ
         -flp:"logFile=output\${ProjectName}.log;verbosity=d" `
     } -errorMessage "ERROR: Build FAILED."
 
-    Remove-Item $globalJsonPath
+    Remove-Item $tempGlobalJsonPath
```
pavel-mikula-sonarsource, line 89: If the PS script fails, we might end up with random file `\sonaranalyzer-dotnet\its\global.json`.

That should not be a problem, right? It should be deleted from build machine and locally we should notice that before commit. We might just wonder how it happened.

sonaranalyzer-dotnet/its/regression-test.ps1
andrei-epure-sonarsource: > Is that intended?
Yes.

We haven't added new open source projects to the ITs since I've joined...

We may/should in the future, but let's treat the case when we will actually do it.

sonaranalyzer-dotnet/its/regression-test.ps1
andrei-epure-sonarsource: Indeed. I can add the deletion in the `finally` block. Good point.

PR #3456, opened 2020-07-22
Run InsecureDeserialization only when using Scanner for MsBuild

### LINKED ISSUE #3421: S5766 should only be available when running in batch analysis mode
### Description

We are seeing many false-positives for the new S5766 rule (Deserializing objects without performing data validation is security-sensitive) and would like to disable it, but our attempts to suppress the rule in source (via `#pragma warning disable S5766`) or reconfigure it's severity in our ruleset file have no effect.

### Repro steps

1) Create an instance of S5766
2) Add `#pragma warning disable S5766` to the top of the file

### Known workarounds

None

### Related information

* C#/VB.NET Plugins Version: 8.9.0.19135
* Visual Studio Version: 16.6.3
* MSBuild / dotnet version: .NET Core SDK 3.1.102

costin-zaharia-sonarsource:
Fixes #3421


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/InsecureDeserializationTest.cs
```diff
@@ -35,6 +35,13 @@ public class InsecureDeserializationTest
             Verifier.VerifyAnalyzer(@"TestCases\InsecureDeserialization.cs",
                 new InsecureDeserialization(AnalyzerConfiguration.AlwaysEnabled),
                 ParseOptionsHelper.CSharp8);
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void InsecureDeserialization_Not_Enabled() =>
```
andrei-epure-sonarsource, line 41: can we have such a test for all hotspots? That would get the hotspots via reflection and would find the test files based on the file name and then run this?

otherwise, we'll always going to have the same problem in the future

costin-zaharia-sonarsource: I'm not sure how can we address this. It needs as an input code which breaks the rule. 

We can get indeed all the hotspots and check if the derive from `HotspotDiagnosticAnalyzer` but this is not enough since the test will need noncompliant code specific for the rule or to check if the `IsEnabled` method is called. 

An option would be to add a custom analyzer specific for our project (internal).

costin-zaharia-sonarsource: Discussed offline and we can do this based on the name convention. The test cases files are named the same as the rule class name.

sonaranalyzer-dotnet/its/sources/ManuallyAddedNoncompliantIssues/NetFramework48/NetFramework48.csproj
```diff
@@ -48,7 +48,6 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="BinaryFormatterTest.cs" />
-    <Compile Include="InsecureDeserialization.cs" />
```
andrei-epure-sonarsource, line 51: why do you want to remove it? you can leave it and replace the `Noncompliant` comment with `// we don't raise issues when not running with the S4MSB`

and maybe open a ticket to add a manual safety net for hotspots in the Java IT, where we integrate with SonarQube

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/SecurityHotspotTest.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+#if NETFRAMEWORK
+            return true;
+#else
+            // IdentityModel is not available on .Net Core
+            return analyzer.GetType().Name != "ControllingPermissions";
+#endif
+        }
+
+        private static IEnumerable<SonarDiagnosticAnalyzer> GetHotspotAnalyzers(AnalyzerLanguage language) =>
```
andrei-epure-sonarsource, line 90: nitpick: in the logical order of reading, this and `IsSecurityHotspot` should be immediately after `VerifyNoIssueReported`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/SecurityHotspotTest.cs
```diff
@@ -0,0 +1,117 @@
...
+                "UsingCookies" => "UsingCookies_Net46",
+#else
+                "ExecutingSqlQueries" => "ExecutingSqlQueries_NetCore",
+                "UsingCookies" => "UsingCookies_NetCore",
+#endif
+                _ => typeName
+            };
+        }
+
+        private static bool IsValid(DiagnosticAnalyzer analyzer)
```
andrei-epure-sonarsource, line 80: maybe `IsTestValid` ? because the rule is valid on NetCore as well, right?

maybe add a comment to say what Valid means

andrei-epure-sonarsource:
also, before merging, this PR should get a better name :)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/SecurityHotspotTest.cs
costin-zaharia-sonarsource: I've reordered the methods a bit.

PR #3457, opened 2020-07-22
Add all C# 9 features to the IT

andrei-epure-sonarsource:


andrei-epure-sonarsource:
I haven't really checked the raised issues, if they make sense. At this stage, I think it's not a good investment of our time. Anyway, we'll have them in peach and we can check them in the UI.

sonaranalyzer-dotnet/its/sources/Net5/InitOnlyProperties.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Net5
+{
+    public class InitOnlyProperties
+    {
+        private readonly string firstName;
+        public string FirstName
+        {
+            get => firstName;
+            init => firstName = (value ?? throw new ArgumentNullException(nameof(FirstName)));
+        }
```
costin-zaharia-sonarsource, line 16: I think we should add also the following cases:
```
public string LastName
{
    get => lastName;
    init => lastName = value!;
}

public string Details { get; init; }

public InitOnlyProperties()
{
    Details = string.Empty;
}
```

sonaranalyzer-dotnet/its/sources/Net5/LambdaDiscardParameters.cs
```diff
@@ -0,0 +1,37 @@
...
+            LinqQuery(items);
+
+            _ = Bar(_ => { return true; });
+
+            Func<int, string, int> func = (int _, string _) => 1;
+
+            LocalFunction(1, 1);
+
+            void LocalFunction(int _, int _2) { }
+        }
```
costin-zaharia-sonarsource, line 23: We can add a test case also with: `Func<int, Func<int, bool>> nested = _ => _ => true;`

sonaranalyzer-dotnet/its/sources/Net5/Main.cs
```diff
@@ -2,12 +2,22 @@
...
-Console.WriteLine(x.FirstName + x.LastName);
+new AttributesOnLocalFunctions().Method();
+// new CallIndirectAttribute();
+// new CovariantReturn();
+new ExtensionGetEnumerator().Method();
+new FunctionPointers().Method();
+new InitOnlyProperties { FirstName = "not null" };
+new LambdaDiscardParameters().Method();
+new NativeInts().Method();
+// ParameterValidation();
```
costin-zaharia-sonarsource, line 14: can the commented code  be removed?

sonaranalyzer-dotnet/its/sources/Net5/ModuleInitializers.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
costin-zaharia-sonarsource, line 5: pedantic:  unused usings

sonaranalyzer-dotnet/its/sources/Net5/NativeInts.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public class NativeInts
+    {
+        public void Method()
+        {
+            nint i = -1;
+            nuint i2 = 2;
+            Console.WriteLine($"Type of {i} is {typeof(nint)}."); // System.IntPtr.
+            Console.WriteLine($"Type of {i2} is {typeof(nuint)}."); // System.UIntPtr
+        }
```
costin-zaharia-sonarsource, line 13: According the the proposal

```
    enum E : nint
    {
    }
```

this should be permitted but currently it is not. Not sure if we should add a test case (it will be commented but we will not forget about this)

sonaranalyzer-dotnet/its/sources/Net5/Records.cs
```diff
@@ -0,0 +1,82 @@
...
+        public int Property => field;
+    }
+
+    public record Finalizer
+    {
+        ~Finalizer()
+        {
+        }
+    }
+
```
costin-zaharia-sonarsource, line 75: I would add one test case in which it implements an interface. I'm thinking that implementing `IDisposable` will have an impact on some of our rules.

sonaranalyzer-dotnet/its/sources/Net5/TargetTypedNew.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public class TargetTypedNew
+    {
+        private Dictionary<string, string> dict = new();
+        private List<int> list;
+
+        public void Method()
+        {
+            list = new();
+            StringBuilder sb = new();
```
costin-zaharia-sonarsource, line 15: We can add a test also with constructor parameters (e.g. `StringBuilder sb2 = new("a");`)

sonaranalyzer-dotnet/its/sources/Net5/FunctionPointers.cs
```diff
@@ -0,0 +1,15 @@
...
+    public unsafe class FunctionPointers
+    {
+        public static void Log() { }
+
+        public void Method()
+        {
+            delegate*<void> ptr1 = &FunctionPointers.Log;
+
+            // This should work but it does not
+            // void* v = &Util.Log;
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: yes; refactoring artefact , it didn't refactor the comment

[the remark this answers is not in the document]
andrei-epure-sonarsource: we'll need to better understand this feature; TBH I don't fully grasp how these are supposed to work

sonaranalyzer-dotnet/its/sources/Net5/Records.cs
costin-zaharia-sonarsource: I've also noticed that the records can be abstract and this might confuse some of our rules. We could add a small test case also with that.

Issue #3408, opened 2020-07-02
Fix S1128 FP: System namespace is needed for tuple deconstruction


RobertSimonsSMA:
### Description

The Rule S1128 marks "using System" with "Remove this unnecessary 'using'. While Deconstructor of System.Tuple in foreach loop depends on it.

### Repro steps
Analyze attached classes with rule S1128 enabled
[SonarRuleS1128AndDeconstructedTuples.zip](https://github.com/SonarSource/sonar-dotnet/files/4862943/SonarRuleS1128AndDeconstructedTuples.zip)

### Expected behavior

Since the compiler needs "using System" to find System.TupleExtensions.Deconstruct, it is expected that Rule S1128 does not mark the using as unneccessary in ClassUsingSystemNamespace.

### Actual behavior

Rule S1128 marks "using System" with "Remove this unnecessary 'using'." in ClassUsingSystemNamespace

### Known workarounds

Do not deconstruct inside the foreach statement header, but instead deconstruct inside the statement body.
E.g: 
// ..
foreach (var tuple  in inputs) {
   var (x, y) = tuple;
   // .. do stuff
}

### Related information
Reproduced using
 * Visual Studio 2019 
 * NuGet package SonarAnalyzer.CSharp 8.9.0.19135 target framework 4.7.2
 * C# Project  with target framework .net 4.7.2 and C# Version 7.3
 * Windows 10 / 1809

costin-zaharia-sonarsource:
Hi @RobertSimonsSMA,

thanks for the feedback. 

I can confirm the issue. We should not raise a warning in this case since `using System` is needed for tuple deconstruction.

PR #3465, opened 2020-07-24
S1128: Add repro for #3408

costin-zaharia-sonarsource:

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UnnecessaryUsingsTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UnnecessaryUsingsTest.cs
@@ -39,6 +39,16 @@ namespace SonarAnalyzer.UnitTest.Rules
                                     additionalReferences: GetAdditionalReferences());
 
         [TestMethod]
+        [TestCategory("Rule")]
+        public void UnnecessaryUsings_TupleDeconstruct() =>
+            Verifier.VerifyAnalyzer(new[]
+                                    {
+                                        @"TestCases\UnnecessaryUsings.TupleDeconstructConsumer.cs",
+                                        @"TestCases\UnnecessaryUsings.TupleDeconstructProvider.cs"
+                                    },
+                                    new CS.UnnecessaryUsings());
+
+        [TestMethod]
         [TestCategory("CodeFix")]
         public void UnnecessaryUsings_CodeFix() =>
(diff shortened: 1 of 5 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UnnecessaryUsingsTest.cs
```diff
@@ -38,6 +38,16 @@ public class UnnecessaryUsingsTest
                                     new CS.UnnecessaryUsings(),
                                     additionalReferences: GetAdditionalReferences());
 
+        [TestMethod]
```
pavel-mikula-sonarsource, line 41: This looks like too many new files to maintain for one simple cases to me. Can you add them to existing file and demonstrate it inside namespace using?
```C#
namespace Repro_3408_Consumer
{
    using System; // Noncompliant - FP, See: https://github.com/SonarSource/sonar-dotnet/issues/3408

    public class Repro3408
    {
        public void Consumer()
        {
            var (_, y) = Repro_3408_Provider.ServiceReturningTuples.GetPair();
        }
    }
}

namespace Repro_3408_Provider
{
    using System;

    public static class ServiceReturningTuples
    {
        public static Tuple<string, int> GetPair() => Tuple.Create("a", 1);
    }
}
```

costin-zaharia-sonarsource: Nice idea with usings in namespace. I've removed the files and added the code inline.

pavel-mikula-sonarsource: Can you remove this test and move the fragment inside `TestCases\UnnecessaryUsings.cs` file? For simplicity and to follow our usual way of adding test cases.

costin-zaharia-sonarsource: Doing this will require to change all file content to put everything what already exists in a namespace to avoid applying already existing usings. This will also require to change the code fix file accordingly.

I prefer to keep this test separated to make the problem obvious.

pavel-mikula-sonarsource:
:shipit: .. once the failed burger check works

costin-zaharia-sonarsource:
> :shipit: .. once the failed burger check works

Had to re-run the step. Works now 😄

PR #3483, opened 2020-08-10
Remove canonical path from Java coverage tests

pavel-mikula-sonarsource:


pavel-mikula-sonarsource:
To prevent tests being environment dependent a fail on some machines. It's related to the `/PAVEL~1.MIK/` issue.
This PR is a fix to make the project working as we would normally expect it. So I can work on the sprint.

PR #3484, opened 2020-08-10
Run the RuleParameterCustomizationTest integration test on CI/CD

andrei-epure-sonarsource:


its/pom.xml
```diff
@@ -41,7 +41,7 @@
     <dependency>
       <groupId>org.sonarsource.sonarqube</groupId>
       <artifactId>sonar-ws</artifactId>
-      <version>7.9</version>
```
andrei-epure-sonarsource, line 44: given we test with LATEST, there's no particular reason to stick on 7.9

moving forward, we should consider having a matrix with [LATEST, LTS], because our current ITs run only against LATEST

its/pom.xml
```diff
@@ -57,13 +57,11 @@
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
       <version>3.6.1</version>
-      <scope>test</scope>
```
andrei-epure-sonarsource, line 60: the whole project is a test project, so either we put the `test` scope to all dependencies, or to none

its/src/test/java/com/sonar/it/csharp/RuleParameterCustomizationTest.java
```diff
@@ -23,18 +23,19 @@
...
 import com.sonar.orchestrator.build.ScannerForMSBuild;
-import org.junit.*;
-import org.junit.rules.TemporaryFolder;
-
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.List;
-
-import org.sonarqube.ws.Issues.Issue;
+import org.junit.Before;
```
andrei-epure-sonarsource, line 29: I replaced ` import org.junit.*; ` with clearly defined imports

its/src/test/java/com/sonar/it/csharp/RuleParameterCustomizationTest.java
```diff
@@ -62,17 +63,16 @@ public void doNotHardcodeCredentials_defaultParameters_canBeCustomized() throws
 
     assertThat(getComponent(componentKey)).isNotNull();
 
-    List<Issue> issues = getIssues(componentKey);
+    List<Hotspot> hotspots = getHotspots(CustomParametersProjectName);
 
-    assertThat(issues.size()).isEqualTo(2);
+    assertThat(hotspots.size()).isEqualTo(1);
 
-    assertIssue(issues.get(0), 6, "Make sure hard-coded credential is safe.");
```
andrei-epure-sonarsource, line 69: this is a regression and I will open a ticket after testing locally as well

its/src/test/java/com/sonar/it/csharp/EnsureAllTestsRunTest.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    return classNamesInTestsSuite;
+  }
+
+  private List<String> getTestClassesInThisPackage(ScanResult scanResult)
+  {
+    Map<String, ClassInfo> allClasses = scanResult.getAllClassesAsMap();
+    List<String> testClassesInPackage = new ArrayList<>();
+    for (String name : allClasses.keySet())
```
andrei-epure-sonarsource, line 77: there's a code smell here telling me to use Map.Entry, but I think it's more readable like this

its/src/test/java/com/sonar/it/csharp/RuleParameterCustomizationTest.java
andrei-epure-sonarsource: https://github.com/SonarSource/sonar-dotnet/issues/3489

its/pom.xml
costin-zaharia-sonarsource: Sounds good. Took a quick look and 8.4.1.35646 is the latest public released version.

its/src/test/java/com/sonar/it/csharp/EnsureAllTestsRunTest.java
```diff
@@ -0,0 +1,87 @@
...
+  public void testClassesAreListedInTestsSuite() {
+    try (ScanResult scanResult = new ClassGraph()
+             .enableAllInfo()
+             .acceptPackages(PACKAGE_NAME)
+             .scan()) {
+
+      List<String> classNamesInTestsSuiteAnnotation = getClassNamesMentionedInSuiteClasses(scanResult);
+      String[] testClassesInPackage = getTestClassesInThisPackage(scanResult).toArray(new String[0]);
+
+      assertThat(classNamesInTestsSuiteAnnotation).containsExactlyInAnyOrder(testClassesInPackage );
```
costin-zaharia-sonarsource, line 53: pedantic: extra space near `testClassesInPackage `

its/src/test/java/com/sonar/it/csharp/EnsureAllTestsRunTest.java
```diff
@@ -0,0 +1,87 @@
...
+  private static final String PACKAGE_NAME = "com.sonar.it.csharp";
+
+  private static final String TEST_SUITE_CLASS = PACKAGE_NAME + ".Tests";
+
+  /**
+   * In order for the tests to run when executing "mvn verify", all test classes need to be mentioned in the
+   * "@SuiteClasses" annotation on {@link com.sonar.it.csharp.Tests}
+   */
+  @Test
+  public void testClassesAreListedInTestsSuite() {
```
costin-zaharia-sonarsource, line 44: Is not possible to replace the @SuiteClasses with an auto-detection mechanism in com.sonar.it.csharp.Tests?

its/src/test/java/com/sonar/it/shared/TestUtils.java
```diff
@@ -155,7 +155,11 @@ public static Integer getMeasureAsInt(Orchestrator orch, String componentKey, St
   }
 
   public static List<Issues.Issue> getIssues(Orchestrator orch, String componentKey) {
-    return newWsClient(orch).issues().search(new SearchRequest().setComponentKeys(Collections.singletonList(componentKey))).getIssuesList();
+    return newWsClient(orch).issues().search(new org.sonarqube.ws.client.issues.SearchRequest().setComponentKeys(Collections.singletonList(componentKey))).getIssuesList();
```
costin-zaharia-sonarsource, line 158: Too bad Java does not support aliases on imports.

its/src/test/java/com/sonar/it/vbnet/EnsureAllTestsRunTest.java
```diff
@@ -0,0 +1,87 @@
...
+import io.github.classgraph.ClassInfo;
+import io.github.classgraph.ScanResult;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class EnsureAllTestsRunTest {
```
costin-zaharia-sonarsource, line 33: We have the `shared` package. I think it would be great to extract a base class in order to reuse the code. Removing the duplication will make the code much easier to maintain and will avoid the cases when we update one class but not the other.

its/src/test/java/com/sonar/it/csharp/EnsureAllTestsRunTest.java
andrei-epure-sonarsource: it may be possible, but I don't find a lot of value in doing this versus the current approach.

having in mind also the effort. Because all the annotations in tests (`@Rule`, `@ClassRule`, `@Before`) work together with `@Suite`. And the question is: why spend a day to rewrite this?

costin-zaharia-sonarsource: Currently a test is needed to ensure that we don't forget to mention the test in `@SuiteClasses`. This is, as far as I can tell, a work around since it's not done automatically and we forget to add the new tests.

In my opinion there would be a big advantage in having this work out of the box (e.g. less code to maintain). I was hopping that the framework allows this but I'm not familiar with it and that's why I was asking.

If the effort to change this is too big I can understand avoiding this change.

Issue #3277, opened 2020-04-01
S4017: Don't count Nullable<T> shorthand when used as a generic type parameter


saciervo:
### Description

Rule S4017 is violated if we nest a nullable type in a generic type. The rule description states that calling a method with such a nested type argument requires complicated and confusing code. Imho, this is not the case with `Nullable<T>` and even more if the shorthand is used. E.g `decimal?` instead of `Nullable<decimal>`. 

Therefore, I would like to suggest that `Nullable<T>` is allowed as a nested generic type or maybe even better; Allow `Nullable<T>`, when the shorthand is used.

### Repro steps

```csharp
public static PropertyBuilder<decimal?> HasPrecision(this PropertyBuilder<decimal?> builder, int precision, int scale)
{
    if (builder == null)
    {
        throw new System.ArgumentNullException(nameof(builder));
    }

    return builder.HasColumnType($"decimal({precision},{scale})");
}
```

### Related information

* SonarAnalyzer.CSharp 8.4.0.15306

pavel-mikula-sonarsource:
Hi @saciervo ,

Thank you for reporting this case. I confirm this as False Positive, the rule should not raise when the `decimal?` nullable syntax is used.

PR #3485, opened 2020-08-11
S4017: Add FP repro for #3277

pavel-mikula-sonarsource:
Reproduces #3277

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotNestTypesInArguments.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotNestTypesInArguments.cs
@@ -62,4 +62,16 @@ namespace MyLibrary
             static void DoSomethingElse(ICollection<ICollection<int>> outerCollect) { } // Noncompliant
         }
     }
+
+    // https://github.com/SonarSource/sonar-dotnet/issues/3277
+    public class Repro_3277
+    {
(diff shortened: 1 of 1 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotNestTypesInArguments.cs
```diff
@@ -62,4 +62,16 @@ public void Method()
...
     }
+
+    // https://github.com/SonarSource/sonar-dotnet/issues/3277
+    public class Repro_3277
+    {
+        public void QuestionMark(List<int?> list)  // Noncompliant FP
+        {
+        }
+
+        public void Explicit(List<Nullable<int>> list) // Noncompliant, there's a nice way out of this
```
costin-zaharia-sonarsource, line 73: According to the issue this should also be a FP, right?

pavel-mikula-sonarsource: From my POV the syntax `List<Nullable<int>>` is just as any other generic `Outer<Inner<int>>` that the rule is trying to prevent. And luckily, there exist a very nice way out of this problem.

While the syntax `List<int?>` is clearly a FP due to it's semantic model. Syntactically it's not a nested generic type (even when we know that semantically it is, under the hood).

I'd keep `List<Nullable<int>>` as a True Positive.

PR #3488, opened 2020-08-11
Rename Azure Pipelines jobs

pavel-mikula-sonarsource:


pavel-mikula-sonarsource:
This GitHub page reports some of the old check to be expected, but they're not in YML file anymore (were renamed and run) so it looks like some outdated cache issue.

azure-pipelines.yml
```diff
@@ -174,7 +176,7 @@ stages:
           msbuildArgs: '/p:Sha1=$(Build.SourceVersion) /p:BranchName=$(Build.SourceBranchName) /p:BuildNumber=$(Build.BuildId) /p:BuildConfiguration=$(BuildConfiguration)'
 
       - task: SonarCloudPrepare@1
-        displayName: 'Run Code Analysis - Begin Step (Pull Request)'
+        displayName: 'Code Analysis - Begin Step (Pull Request)'
```
costin-zaharia-sonarsource, line 179: I think we can remove also the "Step" and use PR instead of Pull Request.

azure-pipelines.yml
```diff
@@ -192,7 +194,7 @@ stages:
             sonar.analysis.repository=$(Build.Repository.ID)
 
       - task: SonarCloudPrepare@1
-        displayName: 'Run Code Analysis - Begin Step (master or branch)'
+        displayName: 'Code Analysis - Begin Step (master or branch)'
```
costin-zaharia-sonarsource, line 197: `Code Analysis - Begin (master or branch)`

azure-pipelines.yml
pavel-mikula-sonarsource: I did the same for "End Step" as well to be consistent.

azure-pipelines.yml
```diff
@@ -278,13 +280,13 @@ stages:
         displayName: 'List dotnet versions and current version'
 
       - task: PowerShell@2
-        displayName: 'Run SonarAnalyzer integration tests'
+        displayName: 'Run ITs'
```
costin-zaharia-sonarsource, line 283: pedantic: forgot about run here

pavel-mikula-sonarsource: This one was intentional. We're inside `.NET ITs` job that has tasks (not visible in GitHub):

- Download binaries to test
- Install NuGet
- Install NET 5 SDK preview
- Install dotnet 3.1 LTS
- List dotnet versions and current version
- Run ITs

Considering this context, the last step should be named `Run` (or as I usually do when prototyping in non production code `GoGoGo` 🙂) because it's the actual step where the `.NET ITs` are triggered. 

I didn't want to repeat `.NET ITs` name and I didn't want to have just `Run` because it's too vague when you're looking just at the job.

pavel-mikula-sonarsource: And I didn't want to use the `regression-test.ps1` in the name as well, that seems like an implementation detail.

PR #3496, opened 2020-08-12
Remove deprecated APIs: org.sonar.api.batch.ScannerSide

### LINKED ISSUE #3473: Remove deprecated APIs: org.sonar.api.batch.ScannerSide
Remove usage of org.sonar.api.batch.ScannerSide from:

AbstractRulesDefinition, CodeCoverageProvider, CoverageAggregator, CSharpSonarRulesDefinition, DotNetPluginMetadata, PropertiesSensor, ProtobufDataImporter, ProtobufImporters, RoslynDataImporter, UnitTestResultsAggregator, UnitTestResultsProvider, VbNetSonarRulesDefinition
AbstractProjectConfiguration, AbstractSolutionConfiguration, EncodingPerFile, ReportPathCollector

andrei-epure-sonarsource:
Fixes #3473


andrei-epure-sonarsource:
Before I can consider this to be ready for review, I need to add more complex ITs that test how imports work for a solution with multiple projects, with different project-level settings (if possible) and test:
- import of metrics
- import of issues
- coverage and UT results

andrei-epure-sonarsource:
This depends on  #3506.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/PropertiesSensor.java
```diff
@@ -22,10 +22,10 @@
 import java.nio.file.Path;
 import java.util.List;
 import java.util.stream.Collectors;
-import org.sonar.api.batch.ScannerSide;
 import org.sonar.api.batch.sensor.Sensor;
 import org.sonar.api.batch.sensor.SensorContext;
 import org.sonar.api.batch.sensor.SensorDescriptor;
+import org.sonar.api.scanner.ScannerSide;
```
pavel-mikula-sonarsource, line 28: `PropertiesSensor` needs to run on each Module to feed the `reportPathCollector`.
IF this works then I don't understand WHY it works 😕 ITs are failing on timeout so I'm not sure if this works.

Do you know how this works on scanner side?

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractSolutionConfiguration.java
```diff
@@ -19,8 +19,7 @@
  */
 package org.sonarsource.dotnet.shared.plugins;
 
-import org.sonar.api.batch.InstantiationStrategy;
-import org.sonar.api.batch.ScannerSide;
+import org.sonar.api.scanner.ScannerSide;
```
pavel-mikula-sonarsource, line 22: Ah, I missed existence of this class when adding `AbstractLanguageConfiguration`. Can we remove this class and move it's content there?

I prefer `LanguageConfiguration` over `SolutionConfiguration` in this context.

pavel-mikula-sonarsource (changes requested):
We'll need to cleanup the mess I created 👼

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/PropertiesSensor.java
andrei-epure-sonarsource: ITs work locally, I tested that before sending the PR. 

I will try and do some tests

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/AbstractLanguageConfigurationTest.java
```diff
@@ -73,6 +73,36 @@ public void vulnerabilityCategories_reads_configuration() {
...
+    Configuration configuration = mock(Configuration.class);
+    when(configuration.getBoolean("sonar.cs.analyzeGeneratedCode")).thenReturn(Optional.of(false));
+
+    AbstractLanguageConfiguration config = createConfiguration(configuration);
+
+    assertThat(config.analyzeGeneratedCode()).isFalse();
+  }
+
+  @Test
+  public void analyzeGeneratedCodeIsEmpty() {
```
pavel-mikula-sonarsource, line 97: It's not clear what's expected result from the test name. Can we rename it to `analyzeGeneratedCodeIsFalseByDefault`?

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/AbstractLanguageConfigurationTest.java
```diff
@@ -73,6 +73,36 @@ public void vulnerabilityCategories_reads_configuration() {
...
+    Configuration configuration = mock(Configuration.class);
+    when(configuration.getBoolean("sonar.cs.analyzeGeneratedCode")).thenReturn(Optional.empty());
```
pavel-mikula-sonarsource, line 99: Can this line be removed for authenticity? In normal scenario, the line would be missing and reading of the configuration will deal with the `Optional` stuff.

I would not add another test.

andrei-epure-sonarsource: Well, it's a unit test, not an integration test. The part about reading the configuration is at integration level.

I will leave it as, especially as the code has inside the `orElse` part

```
    return configuration.getBoolean(AbstractPropertyDefinitions.getAnalyzeGeneratedCode(languageKey)).orElse(false);
```

If the configuration returns `empty`, the `orElse` is executed, and we need a test for that.

Issue #2763, opened 2019-11-04
Update rspec2resx.ps1 to latest .NET 4.8 SDK and clarify error message


pavel-mikula-sonarsource:
### Description

First build of sonar-dotnet and SonarAnalyzer solution fails when .NET 4.7.2 Tools nor .NET 4.6.1 SDK is installed on the computer with 
> You need to install the Windows SDK before using this script

### Repro steps

Prerequisities: Visual Studio installation is missing .NET Framework 4.7.2 SDK and .NET Framework 4.6.1 SDK. It have the default and latest .NET Framework 4.8. SDK installed.

1. Open SonarAnalyzer.sln
2. Build the projects.

### Expected behavior

Solution should build.

### Actual behavior

Build fails with misleading error message
> You need to install the Windows SDK before using this script

### Known workarounds

Install .NET Framework 4.7.2 SDK manually.

### Suggestions

- Update rspec2resx.ps1 to check for .NET Framework 4.8 SDK as well
- Update error message to notify of missing .NET Framework SDK instead of Windows SDK

pavel-mikula-sonarsource:
This was already done in #3031

PR #3500, opened 2020-08-13
Add logging to help with troubleshooting

costin-zaharia-sonarsource:

### DIFF
--- a/scripts/rspec/rspec2resx.ps1
+++ b/scripts/rspec/rspec2resx.ps1
@@ -71,9 +71,11 @@ $roslynLanguageMap = @{
 }
 
 function FindResGen(){
+    Write-Host "Finding ResGen.exe path"
     foreach ($SDK in $SupportedSDKs){
         $Ret = "${Env:ProgramFiles(x86)}\\Microsoft SDKs\\Windows\\v10.0A\\bin\\NETFX ${SDK} Tools\\ResGen.exe"
         if (Test-Path $Ret) {
+            Write-Host "Path found: $Ret"
             return $Ret
         }
     }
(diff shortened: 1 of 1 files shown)


scripts/rspec/rspec2resx.ps1
```diff
@@ -71,9 +71,11 @@ $roslynLanguageMap = @{
 }
 
 function FindResGen(){
+    Write-Host "Finding ResGen.exe path"
     foreach ($SDK in $SupportedSDKs){
         $Ret = "${Env:ProgramFiles(x86)}\\Microsoft SDKs\\Windows\\v10.0A\\bin\\NETFX ${SDK} Tools\\ResGen.exe"
```
andrei-epure-sonarsource, line 76: could you also print the folder structure inside `"${Env:ProgramFiles(x86)}\\Microsoft SDKs\\Windows\\v10.0A\\bin\\` ? 

Because #2763 says that the script doesn't work with only 4.8 installed, but I see at line 35 that `$SupportedSDKs` has
```
$SupportedSDKs = @("4.8", "4.7.2", "4.6.1")
```

so I expect that 4.8 has some different structure?

andrei-epure-sonarsource: and anyway, to see if there's a corelation between failures and machine setup

costin-zaharia-sonarsource: I doubt that there is a problem with the folder structure. I've checked on my pc and it's the following:
```
C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.7.2 Tools\ResGen.exe
C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6.1 Tools\ResGen.exe
C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\ResGen.exe
```

If there will be a problem with this, we would see the following error and will know to investigate more:
> You need to install the .NET Framework $($SupportedSDKs[0]) SDK before using this script.

I think that dumping the folder structure will pollute too much the logs. We can add this later if we notice the above message.

PR #3503, opened 2020-08-13
Make encoding and generated code paths case insensitive

### LINKED ISSUE #3501: Make encoding and generated code paths case insensitive
This test is sometimes failing on CI:

org.junit.ComparisonFailure: expected:<[1]> but was:<[6]> 	at com.sonar.it.vbnet.AutoGeneratedTest.autogenerated_code_with_analyzeGenerated_false(AutoGeneratedTest.java:56)

pavel-mikula-sonarsource:
Fixes #3501


sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFileTest.java
```diff
@@ -51,40 +50,43 @@ public void should_treat_as_match_when_roslyn_entry_missing_for_file() throws IO
...
+    assertEncodingMatch(roslynCharset, fileUri, sqCharset, true);
   }
 
-  private void assertEncodingMatch(Map<URI, Charset> encodingPerUri, URI fileUri, Charset sqCharset, boolean result) throws IOException {
+  @Test
+  public void encoding_per_file_is_not_case_sensitive() throws IOException {
+    Charset sqCharset = StandardCharsets.UTF_16;
+    Charset roslynCharset = StandardCharsets.UTF_8;
+
+    assertEncodingMatch(roslynCharset, URI.create(fileUri.toString().toUpperCase()), sqCharset, false);
```
andrei-epure-sonarsource, line 82: and also add tests that
```
assertEncodingMatch(roslynCharset, URI.create(fileUri.toString().toUpperCase()), roslynCharset, true);
assertEncodingMatch(roslynCharset, URI.create(fileUri.toString().toLowerCase()), roslynCharset, true);
```

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFileTest.java
```diff
@@ -51,40 +50,43 @@ public void should_treat_as_match_when_roslyn_entry_missing_for_file() throws IO
     Charset roslynCharset = StandardCharsets.UTF_8;
     Charset sqCharset = StandardCharsets.UTF_16;
 
-    HashMap<URI, Charset> encodingPerUri = new HashMap<>();
-    encodingPerUri.put(fileUri, roslynCharset);
-    assertEncodingMatch(encodingPerUri, URI.create("dummy"), sqCharset, true);
+    assertEncodingMatch(roslynCharset, URI.create("dummy"), sqCharset, true);
```
andrei-epure-sonarsource, line 53: I don't understand the name of this test - `when_roslyn_entry_missing_for_file`

but `encodingPerUri.put(fileUri, roslynCharset);` does contain the roslyn entry, right?

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFileTest.java
```diff
@@ -51,40 +50,43 @@ public void should_treat_as_match_when_roslyn_entry_missing_for_file() throws IO
...
 
   @Test
   public void should_treat_as_mismatch_when_roslyn_utf8_and_sq_utf16() throws IOException {
     Charset roslynCharset = StandardCharsets.UTF_8;
     Charset sqCharset = StandardCharsets.UTF_16;
 
-    HashMap<URI, Charset> encodingPerUri = new HashMap<>();
-    encodingPerUri.put(fileUri, roslynCharset);
-    assertEncodingMatch(encodingPerUri, fileUri, sqCharset, false);
+    assertEncodingMatch(roslynCharset, fileUri, sqCharset, false);
```
andrei-epure-sonarsource, line 66: hmm reading this test it's very hard to understand the difference between this and `should_treat_as_match_when_roslyn_entry_missing_for_file`

why is `URI.create("dummy")` different from `fileUri` which is `temp.newFile().toURI();` ? 

because the first does not exist on disk and the `fileUri` does exist on disk?

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFileTest.java
pavel-mikula-sonarsource: Be careful about this little difference:
```
encodingPerUri.put(this.fileUri.toString(), roslynCharset);  // Test does this
encodingPerUri.put(fileUri.toString(), roslynCharset);       // You've mentioned something like this
```
The orchestration of the test adds `this.fileUri`. That's result of `temp.newFile().toURI()`. Entry for `temp.newFile().toURI()` is added with declared encoding. 

And in this test, entry for `URI.create("dummy")` is tested. This entry does not exist, as the test name says.

I think the name of the test describes it's content well.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFileTest.java
pavel-mikula-sonarsource: I think the names describe their contents well. 
`entry_missing_for_file` - entire entry for the file itself is missing in the list. 
`encoding_missing` - the entry for the file exists in the list, but contains charset NULL

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/GeneratedFileFilterTest.java
```diff
@@ -56,44 +55,32 @@ public void setUp() {
...
+    Boolean result = filter.accept(inputFile);
 
     // Assert
     assertThat(result).isFalse();
     assertThat(logs.logs(LoggerLevel.DEBUG)).contains("Will ignore generated code");
     assertThat(logs.logs(LoggerLevel.DEBUG)).contains("Skipping auto generated file: autogenerated");
   }
 
-  @Test
-  public void accept_is_not_case_sensitive() throws IOException {
```
andrei-epure-sonarsource: why remove this?

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFile.java
```diff
@@ -39,16 +38,16 @@ public EncodingPerFile(AbstractGlobalProtobufFileProcessor globalReportProcessor
...
+    String uri = inputFile.uri().toString();
 
-    if (!globalReportProcessor.getRoslynEncodingPerUri().containsKey(inputFileUri)) {
+    if (!globalReportProcessor.getRoslynEncodingPerUri().containsKey(uri)) {
       // When there is no entry for a file, it means it was not processed by Roslyn. So we consider encoding to be ok.
       return true;
     }
 
-    Charset roslynEncoding = globalReportProcessor.getRoslynEncodingPerUri().get(inputFileUri);
+    Charset roslynEncoding = globalReportProcessor.getRoslynEncodingPerUri().get(uri); // Case-Insensitive check
```
andrei-epure-sonarsource, line 48: SC says there's a code smell here, and I tend to agree - the line is quite long, it's better to put the comment above

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractGlobalProtobufFileProcessor.java
```diff
@@ -78,17 +80,25 @@ private void processMetadataReportIfPresent(Path reportPath) {
...
       FileMetadataImporter fileMetadataImporter = new FileMetadataImporter();
       fileMetadataImporter.accept(metadataReportProtobuf);
-      this.generatedFileUris.addAll(fileMetadataImporter.getGeneratedFileUris());
-      this.roslynEncodingPerUri.putAll(fileMetadataImporter.getEncodingPerUri());
+      this.generatedFileUris.addAll(fileMetadataImporter.getGeneratedFileUris().stream().map(URI::toString).collect(Collectors.toList()));
+      for (Map.Entry<URI, Charset> entry : fileMetadataImporter.getEncodingPerUri().entrySet()) {
+        String key = entry.getKey().toString();
+        if (!this.roslynEncodingPerUri.containsKey(key)) {
+          this.roslynEncodingPerUri.put(key, entry.getValue());
+        } else if (this.roslynEncodingPerUri.get(key) != entry.getValue()) {
```
andrei-epure-sonarsource, line 88: could you do `this.roslynEncodingPerUri.get(key)` just once? store it in a var?

and maybe the same for `entry.getValue()` (you do the same for `entry.getKey()`)

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractGlobalProtobufFileProcessor.java
```diff
@@ -78,17 +80,25 @@ private void processMetadataReportIfPresent(Path reportPath) {
...
 
-  public Map<URI, Charset> getRoslynEncodingPerUri() {
+  public Map<String, Charset> getRoslynEncodingPerUri() {
     return Collections.unmodifiableMap(roslynEncodingPerUri);
   }
 
-  public Set<URI> getGeneratedFileUris() {
-    return Collections.unmodifiableSet(generatedFileUris);
+  public boolean isGenerated(InputFile inputFile) {
+    return generatedFileUris.contains(inputFile.uri().toString()); // Case-Insensitive check
```
andrei-epure-sonarsource, line 101: I would put the comment above the method - it's part of a public method

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/AbstractGlobalProtobufFileProcessorTest.java
```diff
@@ -66,125 +71,183 @@ public void prepare() {
...
   @Test
   public void process_encoding_new_prop() throws IOException {
-    project2.setProperty("sonar.foo.analyzer.projectOutPaths", mockEncodingProtoReport("UTF-8", "encodingutf8").toString());
+    project2.setProperty("sonar.foo.analyzer.projectOutPaths", mockEncoding("UTF-8", "encodingutf8"));
 
     underTest.build(context);
-    assertThat(underTest.getRoslynEncodingPerUri()).containsOnly(entry(toUri("encodingutf8"), StandardCharsets.UTF_8));
-    assertThat(underTest.getGeneratedFileUris()).isEmpty();
+    assertThat(underTest.getRoslynEncodingPerUri()).containsOnly(entry(toUriString("encodingutf8"), StandardCharsets.UTF_8));
+    assertThat(underTest.isGenerated(mockInputFile("encodingutf8"))).isFalse();
```
andrei-epure-sonarsource, line 148: maybe also test that `isGenerated` for UTF_8 is false as well? otherwise, the reader may wonder if it should be true

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/AbstractGlobalProtobufFileProcessorTest.java
```diff
@@ -66,125 +71,183 @@ public void prepare() {
...
+  }
+
+  @Test
+  public void is_not_case_sensitive() throws IOException {
+    project2.setProperty("sonar.foo.analyzer.projectOutPaths", mockGenerated("upper/lower/mixed"));
```
andrei-epure-sonarsource, line 153: I find it confusing that `upper/lower/mixed` are all lower case... shouldn't they be `UPPER/lower/MiXeD` to be consistent with the names?

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFileTest.java
```diff
@@ -51,40 +50,44 @@ public void should_treat_as_match_when_roslyn_entry_missing_for_file() throws IO
...
+
+    assertEncodingMatch(StandardCharsets.UTF_8, URI.create(fileUri.toString().toUpperCase()), sqCharset, false);
+    assertEncodingMatch(StandardCharsets.UTF_8, URI.create(fileUri.toString().toLowerCase()), sqCharset, false);
+    assertEncodingMatch(StandardCharsets.UTF_16, URI.create(fileUri.toString().toUpperCase()), sqCharset, true);
+    assertEncodingMatch(StandardCharsets.UTF_16, URI.create(fileUri.toString().toLowerCase()), sqCharset, true);
+  }
+
+  private void assertEncodingMatch(Charset roslynCharset, URI fileUri, Charset sqCharset, boolean result) throws IOException {
     AbstractGlobalProtobufFileProcessor processor = mock(AbstractGlobalProtobufFileProcessor.class);
+    TreeMap<String, Charset> encodingPerUri = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
andrei-epure-sonarsource, line 89: I am confused by the fact that `encoding_per_file_is_not_case_sensitive` uses, via `assertEncodingMatch`, a `TreeMap`

if you change the `TreeMap` with a `HashMap` here on line 89, will it the case sensitivity test fail?

because if it does, it seems the unit test setup makes it case insensitive rather than the system under test

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/GeneratedFileFilterTest.java
pavel-mikula-sonarsource: It's no longer responsibility of GeneratedFileFilter to take care about case sensitivity. And it has lost it's value - take a look at the current mock 
```
when(processor.isGenerated(generatedInputFile)).thenReturn(true);
```
It doesn't really test anything related to case sensitivity. So I removed it to deobfuscate this fact.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractGlobalProtobufFileProcessor.java
pavel-mikula-sonarsource: I considered that before and I've chosen it this way. This should be really rare error message, because the code path is almost unreachable (File.cs and file.cs with **different encoding** in the same project). The `TreeMap` has sublinear `O(log n)` asymptotic complexity so I'm fine with calling it twice in this context. And `entry.getValue()` does just `return this.value`.

I think it doesn't worth adding two more lines and one more level of nesting `if` to fix this.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/AbstractGlobalProtobufFileProcessorTest.java
pavel-mikula-sonarsource: I don't think we should test for misreading of arguments, that's not functional. The "UTF-8" literal is value of `encoding` argument.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFileTest.java
pavel-mikula-sonarsource: Yes, the test setup takes care of this. The case-insensivity of the TreeMap itself is already tested by `AbstractGlobalProtobufFileProcessorTest.is_not_case_sensitive`. Here we just need to mock the same behavior.

This test has currently value as a safety net in case that non-trivial `encodingMatch()` is reimplemented in some way that will break this. So I'd like to keep this test and don't delete it.

andrei-epure-sonarsource: I find it weird having a test which is testing a behavior enforced by the underlying mocks.

I'm ok with keeping the test, but please add a comment to mention this fact. Someone seeing this later on may be tempted to remove the test (or may be confused).

PR #3508, opened 2020-08-17
Generate RspecStrings.resx files only on metadata update

costin-zaharia-sonarsource:
these files are now updated only after the new metadata is downloaded instead of each build run


costin-zaharia-sonarsource:
> could you do a test update of the rspecs (on top of this branch) to see how easy to review the diff is?

Sure

costin-zaharia-sonarsource:
> I suggest, before merging, to create the rspecs for a rule that's not yet implemented (e.g. one of the Regex rules that the Java folks created) - at least to see the experience before and get a feeling / see something that doesn't work.

Done. See: https://github.com/SonarSource/sonar-dotnet/pull/3510/files. As far as I can tell these kind of changes are easy to review.

Issue #3506, opened 2020-08-14
Add more complex integration tests for the C# plugin


andrei-epure-sonarsource:
in #3509 :
- [x] Create a solution with 2 C# projects, each with its own test solution
- [x] Add integration test for importing measures and highlighting
- [x] Add integration test for importing issues

in #3512  :
- [x] Add integration test for importing OpenCover coverage

left to do:
- [x] Add integration test for importing VS Coverage

PR #3509, opened 2020-08-17
Add IT with 2 projects (test issues, measures)

andrei-epure-sonarsource:
First part of #3506

PR #3512, opened 2020-08-17
Add opencover integration test for MultipleProjects

andrei-epure-sonarsource:
Related to  #3506

With this, we'll have more comprehensive tests for coverage import.


andrei-epure-sonarsource:
@costin-zaharia-sonarsource currently, this cannot be merged as it depends on #3509 , but the implementation is done

its/projects/MultipleProjects/opencover.xml
```diff
@@ -0,0 +1,1721 @@
+<?xml version="1.0" encoding="utf-8"?>
```
costin-zaharia-sonarsource, line 1: Should we add a small comment to document the exact command used to generate the file (in case we need to update the code and generate this again)?

andrei-epure-sonarsource: yes, I am adding a README in the next PR with visual studio coverage

costin-zaharia-sonarsource: Indeed, having it in a separate file sounds better.

andrei-epure-sonarsource:
before merging this, rebase #3523 to master

Issue #3518, opened 2020-08-18
null-coalescing assignment is not considered when computing cyclomatic complexity


costin-zaharia-sonarsource:
In the current implementation null-coalescing assignment is not considered when computing cyclomatic complexity.

e.g.
```
class MyClass
{
    void MyMethod(int? t) { t ??= 0; }
}
```

has cyclomatic complexity 1 instead of 2. 

For comparison check null-coalescing (??) operator: https://github.com/SonarSource/sonar-dotnet/blob/734df7968037387faade9353c7aedd1a77c010a0/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/MetricsTest.cs#L415-L416

PR #3519, opened 2020-08-18
Add repro for #3518

costin-zaharia-sonarsource:


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/MetricsTest.cs
```diff
@@ -414,6 +414,8 @@ public void Complexity()
                 .Should().Be(2);
             Complexity(AnalyzerLanguage.CSharp, "class MyClass { void MyMethod() { var t = null ?? 0; } }")
                 .Should().Be(2);
+            Complexity(AnalyzerLanguage.CSharp, "class MyClass { void MyMethod(int? t) { t ??= 0; } }")
```
andrei-epure-sonarsource, line 417: do you want to add a test with cognitive complexity on this PR as well?

costin-zaharia-sonarsource: There is one already: https://github.com/SonarSource/sonar-dotnet/blob/b6a7003b67dbd0ae7dae927cb8950e533f923d06/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CognitiveComplexity.cs#L173-L178

Issue #3513, opened 2020-08-18
Switch expressions are not considered when computing cyclomatic complexity metric


costin-zaharia-sonarsource:
According to [docs](https://docs.sonarqube.org/latest/user-guide/metric-definitions/): 

> Whenever the control flow of a function splits, the complexity counter gets incremented by one.

For switch expressions this does not happen. The number of arms is not counted and they are evaluated to 1.

e.g.  

```
class MyClass
{
    void MyMethod(int p)
    {
        switch (p)
        {
            case 0: break;
            default: break;
        }
    }
}
```
has complexity 2, but

```
class MyClass
{
    void MyMethod(int p)
    {
        var x = p switch
        {
            1 => "one",
            _ => string.Empty
        };
    }
}
```
has complexity 1.

Impacted rules: [S1541](https://rules.sonarsource.com/csharp/RSPEC-1541).

PR #3521, opened 2020-08-18
Count switch expression arms when computing cyclomatic complexity

costin-zaharia-sonarsource:
Fixes: #3513


costin-zaharia-sonarsource:
This will have an impact on [S1541](https://rules.sonarsource.com/csharp/RSPEC-1541).

PR #3524, opened 2020-08-18
Fix build break

andrei-epure-sonarsource:
I merged #3509 after #3521  which changed the way complexity was computed.

### DIFF
--- a/its/src/test/java/com/sonar/it/csharp/MultipleProjectsTest.java
+++ b/its/src/test/java/com/sonar/it/csharp/MultipleProjectsTest.java
@@ -185,13 +185,13 @@ public class MultipleProjectsTest {
 
   @Test
   public void complexityAtProjectLevel() {
-    assertThat(getProjectMeasureAsInt("complexity")).isEqualTo(17);
+    assertThat(getProjectMeasureAsInt("complexity")).isEqualTo(20);
   }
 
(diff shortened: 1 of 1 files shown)


its/src/test/java/com/sonar/it/csharp/MultipleProjectsTest.java
```diff
@@ -200,7 +200,7 @@ public void complexityAtFileLevel() {
     assertThat(getFileMeasureAsInt(FIRST_PROJECT_SECOND_CLASS_FILE, "complexity")).isEqualTo(1);
     assertThat(getFileMeasureAsInt(FIRST_PROJECT_ASSEMBLY_INFO_FILE, "complexity")).isEqualTo(0);
 
-    assertThat(getFileMeasureAsInt(SECOND_PROJECT_FIRST_CLASS_FILE, "complexity")).isEqualTo(3);
+    assertThat(getFileMeasureAsInt(SECOND_PROJECT_FIRST_CLASS_FILE, "complexity")).isEqualTo(6);
```
costin-zaharia-sonarsource, line 203: We now count the number of switch expressions arms when computing cyclomatic complexity. In this case 3.

costin-zaharia-sonarsource (approved):
LGTM!

PR #3525, opened 2020-08-19
Count null-coalescing assignment when computing cyclomatic complexity

### LINKED ISSUE #3518: null-coalescing assignment is not considered when computing cyclomatic complexity

costin-zaharia-sonarsource:
Fixes #3518

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/MetricsTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/MetricsTest.cs
@@ -422,7 +422,9 @@ End Class").Should().BeEquivalentTo(1, 2, 3, 4, 5, 6, 7, 8);
             Complexity(AnalyzerLanguage.CSharp, "class MyClass { void MyMethod() { var t = null ?? 0; } }")
                 .Should().Be(2);
             Complexity(AnalyzerLanguage.CSharp, "class MyClass { void MyMethod(int? t) { t ??= 0; } }")
-                .Should().Be(1); // null-coalescing assignment is not considered when computing cyclomatic complexity. See: https://github.com/SonarSource/sonar-dotnet/issues/3518
+                .Should().Be(2);
+            Complexity(AnalyzerLanguage.CSharp, "class MyClass { void MyMethod(int? a, int? b, int? c) => a ??= b ??= c ??= 0; }")
+                .Should().Be(4);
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/MetricsTest.cs
```diff
@@ -422,7 +422,9 @@ public void Complexity()
             Complexity(AnalyzerLanguage.CSharp, "class MyClass { void MyMethod() { var t = null ?? 0; } }")
                 .Should().Be(2);
             Complexity(AnalyzerLanguage.CSharp, "class MyClass { void MyMethod(int? t) { t ??= 0; } }")
-                .Should().Be(1); // null-coalescing assignment is not considered when computing cyclomatic complexity. See: https://github.com/SonarSource/sonar-dotnet/issues/3518
+                .Should().Be(2);
```
andrei-epure-sonarsource, line 425: from the [discussion with Ann](https://discuss.sonarsource.com/t/cognitive-complexity-for-null-coalesce-and-null-coalesce-assignment-operators/5432/2), I understand that we shouldn't treat this differently 

Although chaining these assignments is still a bit messy. I'll wait for the test to be updated.

andrei-epure-sonarsource (changes requested):

andrei-epure-sonarsource (approved):
LGTM

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Common/MetricsTest.cs
costin-zaharia-sonarsource: From what I understood the discussion was related to cognitive complexity and this issue is related to cognitive complexity which is a different topic.

When computing cyclomatic complexity, `?.` is used like:
```
            var x = c?.Name?.ToString();
//                   ^ Secondary [10] {{+1}}
//                         ^ Secondary@-1 [10] {{+1}}
```

PR #3529, opened 2020-08-19
Increase cognitive complexity when ??= is used

### LINKED ISSUE #3520: null-coalescing assignment is not considered when computing cognitive complexity
In the current implementation null-coalescing assignment is not considered when computing cognitive complexity.

e.g.
```
void NullCoalescenceAssignment() // Compliant - FN: Null coalescence assignment is not considered
{
    bool? value = null;

    value ??= true;
}
```

**Additional notes:**
the behavior is consistent with null-coalescing operator:

```
void NullCoalescence(string first, string second) // Compliant - FN: null-coalescing operator is not considered
{
    var x = first ?? second;
}
```

costin-zaharia-sonarsource:
Fixes #3520

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CognitiveComplexity.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CognitiveComplexity.cs
@@ -759,5 +759,25 @@ namespace Tests.Diagnostics
                     }
                 };
         }
+
+        void NullCoalescence(string first, string second) // Compliant - FN: null-coalescing operator is not considered: https://github.com/SonarSource/sonar-dotnet/issues/3528
+        {
+            var x = first ?? second;
+        }
+
+        void NullCoalescenceAssignment(int? first, int? second, int? third, int? fourth) // Noncompliant {{Refactor this method to reduce its Cognitive Complexity from 8 to the 0 allowed.}}
+        {
+            var a = first ??= 0;
+//                        ^^^ Secondary {{+1}}
+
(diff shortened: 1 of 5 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CognitiveComplexity.cs
```diff
@@ -759,5 +759,25 @@ void BlockAction()
...
+        {
+            var x = first ?? second;
+        }
+
+        void NullCoalescenceAssignment(int? first, int? second, int? third, int? fourth) // Noncompliant {{Refactor this method to reduce its Cognitive Complexity from 8 to the 0 allowed.}}
+        {
+            var a = first ??= 0;
+//                        ^^^ Secondary {{+1}}
+
+            var b = first ??= second ??= third ??= fourth;
```
pavel-mikula-sonarsource, line 773: Can you add a test demonstrating broken chain? Something like `a && b || c && d`? In this case, probably `fist ??= second = third ??= fourth`.

pavel-mikula-sonarsource (approved):
LGTM. I just left one suggestion about adding a test case.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/CognitiveComplexity.cs
```diff
@@ -759,5 +759,25 @@ void BlockAction()
                     }
                 };
         }
+
```
pavel-mikula-sonarsource, line 762: Line 193 in this file will need your attention 😄

PR #3530, opened 2020-08-19
Document cognitive complexity for null conditional and null coalescing operators

costin-zaharia-sonarsource:


andrei-epure-sonarsource (approved):
LGTM

Issue #3476, opened 2020-08-10
Remove deprecated APIs: org.sonar.api.batch.sensor.SensorDescriptor.global()


andrei-epure-sonarsource:
Remove `org.sonar.api.batch.sensor.SensorDescriptor.global()` from CoverageReportImportSensor, DotNetSensor
UnitTestResultsImportSensor

PR #3531, opened 2020-08-19
Remove deprecated APIs: SensorDescriptor.global() from UnitTestResultsImportSensor

pavel-mikula-sonarsource:
Contributes #3476

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsAggregator.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsAggregator.java
@@ -21,8 +21,8 @@ package org.sonar.plugins.dotnet.tests;
 
 import java.io.File;
 import java.util.function.Predicate;
-import org.sonar.api.batch.ScannerSide;
 import org.sonar.api.config.Configuration;
+import org.sonar.api.scanner.ScannerSide;
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 
@@ -45,9 +45,9 @@ public class UnitTestResultsAggregator {
   }
 
   UnitTestResultsAggregator(UnitTestConfiguration unitTestConf, Configuration configuration,
-                            VisualStudioTestResultsFileParser visualStudioTestResultsFileParser,
-                            NUnitTestResultsFileParser nunitTestResultsFileParser,
-                            XUnitTestResultsFileParser xunitTestResultsFileParser) {
+    VisualStudioTestResultsFileParser visualStudioTestResultsFileParser,
+    NUnitTestResultsFileParser nunitTestResultsFileParser,
+    XUnitTestResultsFileParser xunitTestResultsFileParser) {
     this.unitTestConf = unitTestConf;
     this.configuration = configuration;
     this.visualStudioTestResultsFileParser = visualStudioTestResultsFileParser;
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsAggregator.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsAggregator.java
@@ -77,21 +77,22 @@ public class UnitTestResultsAggregator {
     return hasKeyPredicate.test(unitTestConf.xunitTestResultsFilePropertyKey());
   }
 
-  UnitTestResults aggregate(WildcardPatternFileProvider wildcardPatternFileProvider, UnitTestResults unitTestResults) {
+  UnitTestResults aggregate(WildcardPatternFileProvider wildcardPatternFileProvider) {
+    UnitTestResults results = new UnitTestResults();
+
     if (hasVisualStudioTestResultsFile(configuration::hasKey)) {
-      aggregate(wildcardPatternFileProvider, configuration.getStringArray(unitTestConf.visualStudioTestResultsFilePropertyKey()), visualStudioTestResultsFileParser,
-        unitTestResults);
+      aggregate(wildcardPatternFileProvider, configuration.getStringArray(unitTestConf.visualStudioTestResultsFilePropertyKey()), visualStudioTestResultsFileParser, results);
     }
 
     if (hasNUnitTestResultsFile(configuration::hasKey)) {
-      aggregate(wildcardPatternFileProvider, configuration.getStringArray(unitTestConf.nunitTestResultsFilePropertyKey()), nunitTestResultsFileParser, unitTestResults);
+      aggregate(wildcardPatternFileProvider, configuration.getStringArray(unitTestConf.nunitTestResultsFilePropertyKey()), nunitTestResultsFileParser, results);
     }
 
     if (hasXUnitTestResultsFile(configuration::hasKey)) {
-      aggregate(wildcardPatternFileProvider, configuration.getStringArray(unitTestConf.xunitTestResultsFilePropertyKey()), xunitTestResultsFileParser, unitTestResults);
+      aggregate(wildcardPatternFileProvider, configuration.getStringArray(unitTestConf.xunitTestResultsFilePropertyKey()), xunitTestResultsFileParser, results);
     }
 
-    return unitTestResults;
+    return results;
   }
 
   private static void aggregate(WildcardPatternFileProvider wildcardPatternFileProvider, String[] reportPaths, UnitTestResultsParser parser, UnitTestResults unitTestResults) {
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensor.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensor.java
@@ -64,23 +64,19 @@ public class UnitTestResultsImportSensor implements ProjectSensor {
   @Override
   public void execute(SensorContext context) {
     if (unitTestResultsAggregator.hasUnitTestResultsProperty()) {
-      analyze(context, new UnitTestResults()); // FIXME: Refactoring
+      try {
+        saveTestMetrics(context);
+      } catch (Exception e) {
+        LOG.warn("Could not import unit test report: '{}'", e.getMessage());
+        analysisWarnings.addUnique(String.format("Could not import unit test report for '%s'. Please check the logs for more details.", languageName));
+      }
     } else {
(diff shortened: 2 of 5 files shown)


sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensorTest.java
```diff
@@ -143,44 +102,89 @@ public void describe_execute_only_when_key_present() {
...
+  }
+
   @Test
   public void import_two_reports_for_same_project_should_not_throw() throws Exception {
-    UnitTestResultsAggregator unitTestResultsAggregator = mock(UnitTestResultsAggregator.class);
+    UnitTestResultsAggregator aggregator = mock(UnitTestResultsAggregator.class);
+    DotNetPluginMetadata cSharpMetadata = mockCSharpMetadata();
+    DotNetPluginMetadata vbNetMetadata = mock(DotNetPluginMetadata.class);
+    when(vbNetMetadata.languageKey()).thenReturn("vb");
+    when(vbNetMetadata.languageName()).thenReturn("VB.NET");
```
costin-zaharia-sonarsource, line 134: We can extract a method also for mocking vb metadata. It will make this test easier to read.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensorTest.java
```diff
@@ -53,75 +54,32 @@
...
+    results.add(42, 1, 2, 3, null);
+    when(aggregator.hasUnitTestResultsProperty()).thenReturn(true);
+    when(aggregator.aggregate(any())).thenReturn(results);
+    SensorContextTester context = SensorContextTester.create(temp.newFolder());
+    when(aggregator.aggregate(any())).thenReturn(results);
 
-    verify(unitTestResultsAggregator).aggregate(Mockito.any(WildcardPatternFileProvider.class), Mockito.eq(results));
+    new UnitTestResultsImportSensor(aggregator, metadata, analysisWarnings).execute(context);
 
     assertThat(context.measures("projectKey"))
```
costin-zaharia-sonarsource, line 79: It would be great to add an explicit assertion that we expect the value for execution time to be missing (since it's passed as `null` in `results.add(42, 1, 2, 3, null);`). 

I think this will make the test (`should_not_save_metrics_with_empty_results`) easier to understand.

costin-zaharia-sonarsource (changes requested):

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensor.java
```diff
@@ -20,75 +20,59 @@
...
   private final String languageKey;
   private final String languageName;
   private final AnalysisWarnings analysisWarnings;
 
-  public UnitTestResultsImportSensor(UnitTestResultsAggregator unitTestResultsAggregator, ProjectDefinition projectDef,
-                                     DotNetPluginMetadata pluginMetadata, AnalysisWarnings analysisWarnings) {
-    this(unitTestResultsAggregator, projectDef, pluginMetadata.languageKey(), pluginMetadata.languageName(), analysisWarnings);
-  }
-
-  public UnitTestResultsImportSensor(UnitTestResultsAggregator unitTestResultsAggregator, ProjectDefinition projectDef,
```
andrei-epure-sonarsource, line 52: I see you removed one of the two public constructors - I imagine it wasn't used at all?

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensor.java
```diff
@@ -20,75 +20,59 @@
...
  * This class is responsible to handle all the C# and VB.NET unit test results reports (parse and report back to SonarQube).
  */
-public class UnitTestResultsImportSensor implements Sensor {
+public class UnitTestResultsImportSensor implements ProjectSensor {
 
   private static final Logger LOG = Loggers.get(UnitTestResultsImportSensor.class);
 
   private final WildcardPatternFileProvider wildcardPatternFileProvider = new WildcardPatternFileProvider(new File("."), File.separator);
   private final UnitTestResultsAggregator unitTestResultsAggregator;
-  private final ProjectDefinition projectDef;
```
andrei-epure-sonarsource, line 42: for my education: what was this replaced with?

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsAggregator.java
```diff
@@ -77,21 +77,22 @@ private boolean hasXUnitTestResultsFile(Predicate<String> hasKeyPredicate) {
     return hasKeyPredicate.test(unitTestConf.xunitTestResultsFilePropertyKey());
   }
 
-  UnitTestResults aggregate(WildcardPatternFileProvider wildcardPatternFileProvider, UnitTestResults unitTestResults) {
+  UnitTestResults aggregate(WildcardPatternFileProvider wildcardPatternFileProvider) {
+    UnitTestResults results = new UnitTestResults();
```
andrei-epure-sonarsource, line 81: nice!

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensor.java
andrei-epure-sonarsource: I see it was only being used in
```
    if (projectDef.getParent() == null) {
      analyze(context, new UnitTestResults());
    }
```
but I don't know what that was for

andrei-epure-sonarsource: Aaa, it was because it was ran at all modules, and it should only do something at the top-level module that didn't have any parent.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/UnitTestResultsAggregatorTest.java
```diff
@@ -98,10 +99,9 @@ public void aggregate() {
     VisualStudioTestResultsFileParser visualStudioTestResultsFileParser = mock(VisualStudioTestResultsFileParser.class);
     NUnitTestResultsFileParser nunitTestResultsFileParser = mock(NUnitTestResultsFileParser.class);
     XUnitTestResultsFileParser xunitTestResultsFileParser = mock(XUnitTestResultsFileParser.class);
-    UnitTestResults results = mock(UnitTestResults.class);
     new UnitTestResultsAggregator(unitTestConf, settings.asConfig(), visualStudioTestResultsFileParser, nunitTestResultsFileParser, xunitTestResultsFileParser)
-      .aggregate(wildcardPatternFileProvider, results);
-    verify(visualStudioTestResultsFileParser).accept(new File("foo.trx"), results);
+      .aggregate(wildcardPatternFileProvider);
```
andrei-epure-sonarsource, line 103: I find this test really hard to read because it's doing Arrange-Act-Assert multiple times for VS Test , NUnit, XUnit and then some more.

For the clean as you code benefit, would it be worth splitting these in single tests that do only one set of Arrange-Act-Assert ?

andrei-epure-sonarsource (changes requested):

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsAggregator.java
```diff
@@ -45,9 +45,9 @@ public UnitTestResultsAggregator(UnitTestConfiguration unitTestConf, Configurati
   }
 
   UnitTestResultsAggregator(UnitTestConfiguration unitTestConf, Configuration configuration,
-                            VisualStudioTestResultsFileParser visualStudioTestResultsFileParser,
-                            NUnitTestResultsFileParser nunitTestResultsFileParser,
-                            XUnitTestResultsFileParser xunitTestResultsFileParser) {
+    VisualStudioTestResultsFileParser visualStudioTestResultsFileParser,
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: We agreed that for .NET once we have some tooling for it. I don't like it like this here as well. But Java code is subjected to [Developer Toolset for Sonar-* Projects / Java code style configuration](https://github.com/SonarSource/sonar-developer-toolset#code-style-configuration-for-intellij).

I'm not going to fight against this autoformatting. But I'll be very happy if you raise this point in LT and update `sonar-formatter.xml`.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensorTest.java
```diff
@@ -53,75 +54,32 @@
   @Test
   public void coverage() {
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: AFAICT this test can be removed, because it's replaced by `execute_warns_when_no_key_is_present` with better assertion.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensor.java
pavel-mikula-sonarsource: It was used only in tests - see commit `Remove test-only constructor to improve coverage`. And having separate constructor for tests and then a separate test to test the normal constructor is useless 🤯

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsImportSensor.java
pavel-mikula-sonarsource: Exactly. The sensor was looking for the last top-level module to run on. And that's now implemented by inheriting `ProjectSensor`. So this workaround is not needed anymore and that's nice.

andrei-epure-sonarsource (changes requested):
nice, one last suggestion

andrei-epure-sonarsource (approved):
LGTM

costin-zaharia-sonarsource (approved):
LGMT!

PR #3536, opened 2020-08-20
Remove redundant step from doc

costin-zaharia-sonarsource:
**SonarAnalyzer.RuleDescriptorGenerator** is executed for both **cs** and **vbnet** during build.


pavel-mikula-sonarsource (approved):
LGTM

PR #3537, opened 2020-08-20
Fix Java plugin Code Smells: Major

pavel-mikula-sonarsource:


costin-zaharia-sonarsource:
I cannot merge this due to a conflict. @pavel-mikula-sonarsource will you take a look?

PR #3538, opened 2020-08-20
Fix Java plugin Code Smells: Minor in code

pavel-mikula-sonarsource:


pavel-mikula-sonarsource:
I will ignore the coverage QG in this case, because there's no real way to cover the 4th condition case.

sonar-csharp-plugin/src/test/java/org/sonar/plugins/csharp/CSharpTest.java
```diff
@@ -53,4 +54,17 @@ public void shouldGetCustomFileSuffixes() {
...
   }
 
+  @Test
+  public void equals_and_hashCode_considers_configuration() {
+    MapSettings otherSettings = new MapSettings();
+    otherSettings.setProperty("key", "value");
+    CSharp otherCSharp = new CSharp(otherSettings.asConfig());
+    CSharp sameCSharp = new CSharp(settings.asConfig());
+
+    assertThat(csharp).isEqualTo(sameCSharp)
```
andrei-epure-sonarsource, line 64: can you add a test comparing it to a VBNet?

sonar-csharp-plugin/src/test/java/org/sonar/plugins/csharp/CSharpTest.java
```diff
@@ -53,4 +54,17 @@ public void shouldGetCustomFileSuffixes() {
     assertThat(csharp.getFileSuffixes()).containsOnly(".cs", ".csharp");
   }
 
+  @Test
+  public void equals_and_hashCode_considers_configuration() {
```
andrei-epure-sonarsource, line 58: could you separate the test for equals and the test for hashCode?

sonar-vbnet-plugin/src/test/java/org/sonar/plugins/vbnet/VbNetTest.java
```diff
@@ -53,4 +54,17 @@ public void shouldGetCustomFileSuffixes() {
     assertThat(vbnet.getFileSuffixes()).containsOnly(".vb", ".vbnet");
   }
 
+  @Test
+  public void equals_and_hashCode_considers_configuration() {
```
andrei-epure-sonarsource, line 58: separate tests and add test comparing with a CSharp?

sonar-vbnet-plugin/src/main/java/org/sonar/plugins/vbnet/VbNet.java
```diff
@@ -36,4 +37,13 @@ public VbNet(Configuration configuration) {
     return configuration.getStringArray(VbNetPlugin.FILE_SUFFIXES_KEY);
   }
 
+  @Override
+  public boolean equals(Object o) {
+    return super.equals(o) && configuration == ((VbNet) o).configuration;
```
andrei-epure-sonarsource, line 42: also, to test all conditions, I think you can make a sneaky `TestAbstractLanguage extends AbstractLanguage` that is invoked with `super(VbNetPlugin.LANGUAGE_KEY, VbNetPlugin.LANGUAGE_NAME);` but will fail when casting to a VBNet

at least like this you'll make it clear that it will blow up in case this happens

andrei-epure-sonarsource (changes requested):

sonar-csharp-plugin/src/test/java/org/sonar/plugins/csharp/CSharpTest.java
pavel-mikula-sonarsource: Only with fake VbNet class, because it's in separate library.

sonar-vbnet-plugin/src/main/java/org/sonar/plugins/vbnet/VbNet.java
pavel-mikula-sonarsource: In overall, it's responsibility of the base class unit test to test the base class behavior.

andrei-epure-sonarsource: Later edit: what I suggested would test the `((VbNet) o).configuration;` , not the base class behavior.

andrei-epure-sonarsource: it's still a public class and you don't know exactly how it's going to be used by the scanner, right?

PR #3542, opened 2020-08-20
Remove unused .nuspec files

costin-zaharia-sonarsource:


andrei-epure-sonarsource (approved):
it is really weird that we are having nuspec files for an internal tool.

LGTM

andrei-epure-sonarsource:
there are 2 ongoing builds on master, I'd wait for them to be green before merging

PR #3544, opened 2020-08-20
NuGet packaging: Place SonarAnalyzer.CFG.dll in a framework specific folder

### LINKED ISSUE #3543: NuGet packaging: SonarAnalyzer.CFG.dll should be in a framework specific folder.
### Description

While building the `SonarAnalyzer.CFG.CSharp` package, NuGet logs the following warning:

```
WARNING: NU5101: The assembly 'lib\SonarAnalyzer.CFG.dll' is placed directly under 'lib' folder. It is recommended that assemblies be placed inside a framework-specific folder. Move it into a framework-specific folder.
```

According to the [documentation](https://docs.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference#assemblies-in-the-lib-root-are-ignored-when-the-package-is-installed-after-the-migration) assemblies in the lib root are ignored when the package is installed after the migration to PackageReference.

costin-zaharia-sonarsource:
A complete list of TFM can be found here: https://docs.microsoft.com/en-us/dotnet/standard/frameworks

Fixes #3543


costin-zaharia-sonarsource:
After the change, the folder structure will look like this:
image

PR #3553, opened 2020-08-24
S3254: Fix RedundantArgumentTest references

### LINKED ISSUE #3430: S3254: Fix RedundantArgumentTest references
### Description

RedundantArgument code fix generates wrong code when targeting .Net Core 3.1. It removed arguments from expression trees and this results in compiler error `[CS0854] An expression tree may not contain a call or invocation that uses optional arguments`

### Repro steps

`Expression<Action> expTreeVar = () => FuncWithOptionals(null);` -> `Expression<Action> expTreeVar = () => FuncWithOptionals();`

where `FuncWithOptionals` is:

```
private static string FuncWithOptionals(string str = null, params string[] args)
{
    return str;
}
```

### Related information

* C#/VB.NET Plugins Version: 8.9

pavel-mikula-sonarsource:
Fixes #3430


costin-zaharia-sonarsource (approved):
LGTM!

PR #3555, opened 2020-08-24
Do not retrieve symbols for Guids when processing ObjectCreationExpressions

costin-zaharia-sonarsource:
This PR reduces the `UnusedPrivateMember` analysis time for https://github.com/ksitarek/sonarscannersample/blob/master/vtb.Warehouse.Data/Migrations/20181005212624_InitialCreate.cs file from ~11 minutes to ~1 sec.

Main problem seems to be caused by the retrieval of unnecessary symbols for Guid when the constructor is called: 
```
values: new object[,]
                {
                    { new Guid("6833ef8a-2454-46bf-816e-b3b1157a145e"), new Guid("90d06fdc-4112-44b4-ab82-e38c4a2baa98"), 0 },
                    { new Guid("243abb4a-2484-42c7-9f01-5b4424040588"), new Guid("fb77981e-119b-4bb9-ba9b-26ff22c646d5"), 2 },
                    { new Guid("eb64d993-f585-4f8f-9480-de07d2279dd1"), new Guid("fb77981e-119b-4bb9-ba9b-26ff22c646d5"), 3 },
```

Partial fix for: #2474


andrei-epure-sonarsource:
you need to update the .NET IT expected results (see [failed test](https://sonarsource.visualstudio.com/DotNetTeam%20Project/_build/results?buildId=21110&view=logs&jobId=c7ac5d00-ca19-592b-a6ee-71358b142e5c&j=c7ac5d00-ca19-592b-a6ee-71358b142e5c&t=5bc48f64-1755-5faf-f681-1e0b2d826529))

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
```diff
@@ -179,10 +179,24 @@ private SymbolAccess ParentAccessType(SyntaxNode node)
...
+            }
+
             base.VisitObjectCreationExpression(node);
         }
 
+        private static bool IsInInitializerExpression(SyntaxNode node)
+        {
+            var parentKind = node.Parent.Kind();
+
+            return parentKind == SyntaxKind.ArrayInitializerExpression
```
andrei-epure-sonarsource, line 194: You've added the perf test, and I would add some more UTs to test the functionality, that the unused private fields are detected and issues are raised for the syntaxes covered here, when the private field exists and it's not used.


Including edge cases like
```
public class Foo
{
	private string x;
	private string[] y;
	public Foo()
	{
		y = new string[] { x = new String("foo".ToCharArray()) };
		new Qix(x = new String("bar".ToCharArray()));
		new Qix(x : x = new String("X".ToCharArray()));
		new List<string> { x = new String("Y".ToCharArray()) };
	}
}

public class Qix
{
	private string y;
	public Qix(string x)
	{
		y = x;
	}
}

public class Matrix
{
    private double[,] storage = new double[3, 3];

    public double this[int row, int column]
    {
        // The embedded array will throw out of range exceptions as appropriate.
        get { return storage[row, column]; }
        set { storage[row, column] = value; }
    }
}
```
unless we already test them

and also with [object initializers](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/object-and-collection-initializers#object-initializers) w/ and w/o anonymous types

and also check what happens when creating an object in a LINQ query - as a common scenario

andrei-epure-sonarsource (changes requested):
Nice! I think you should add more tests for the covered syntaxes.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
```diff
@@ -41,11 +42,15 @@ internal class CSharpSymbolUsageCollector : CSharpSyntaxWalker
...
+        // (see entity framework migration files) but use this list carefully since it can lead to false positives
+        // if the names are redefined in the user code.
+        private static readonly ImmutableHashSet<string> ignoredConstructors = ImmutableHashSet.Create("Guid", "System.Guid");
+
         private readonly Func<SyntaxNode, SemanticModel> getSemanticModel;
         private readonly HashSet<string> knownSymbolNames;
 
-        public HashSet<ISymbol> UsedSymbols { get; } =
-            new HashSet<ISymbol>();
+        public HashSet<ISymbol> UsedSymbols { get; } = new HashSet<ISymbol>();
```
andrei-epure-sonarsource, line 53: can this be an ISet?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
```diff
@@ -41,11 +42,15 @@ internal class CSharpSymbolUsageCollector : CSharpSyntaxWalker
             SyntaxKind.PreDecrementExpression
         };
 
+        // The types names (from .Net) defined here are very common and can slow down the analysis significantly
+        // (see entity framework migration files) but use this list carefully since it can lead to false positives
+        // if the names are redefined in the user code.
+        private static readonly ImmutableHashSet<string> ignoredConstructors = ImmutableHashSet.Create("Guid", "System.Guid");
```
andrei-epure-sonarsource, line 48: can you add a test with the FP to document it?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
```diff
@@ -314,7 +323,7 @@ static ISymbol GetOriginalDefinition(ISymbol candidateSymbol)
 
         private void TryStorePropertyAccess(ExpressionSyntax node, List<ISymbol> identifierSymbols)
         {
-            var propertySymbols = identifierSymbols.OfType<IPropertySymbol>();
+            var propertySymbols = identifierSymbols.OfType<IPropertySymbol>().ToList();
```
andrei-epure-sonarsource, line 326: why is the ToList necessary? because of the Any() followed by the foreach?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
```diff
@@ -249,7 +258,7 @@ public override void VisitPropertyDeclaration(PropertyDeclarationSyntax node)
             base.VisitPropertyDeclaration(node);
         }
 
-        private IMethodSymbol GetImplicitlyCalledConstructor(IMethodSymbol constructor)
+        private static IMethodSymbol GetImplicitlyCalledConstructor(IMethodSymbol constructor)
```
andrei-epure-sonarsource, line 261: maybe group static private methods together?

andrei-epure-sonarsource (changes requested):

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
costin-zaharia-sonarsource: This is no longer valid since I've changed the approach.

andrei-epure-sonarsource: BTW, some of these tests would still be valuable to have , in case we don't cover these sort of syntaxes

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
costin-zaharia-sonarsource: exactly. the enumeration is evaluated twice

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
costin-zaharia-sonarsource: Since this is not directly related to this PR I would like to postpone this to keep this PR focused on the problem at hand. It is already big enough.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/UnusedPrivateMemberTest.cs
```diff
@@ -161,58 +158,90 @@ public class AssetPostprocessor { }
...
+            // Once the NuGet packages are downloaded, the time to execute the analyzer on the given file is
+            // about ~1 sec. It was reduced from ~11 min by skipping Guids when processing ObjectCreationExpression.
+            // The threshold is set here to 30 seconds to avoid flaky builds due to slow build agents or network connections.
+            verifyAnalyzer.ExecutionTime().Should().BeLessOrEqualTo(30.Seconds());
         }
 
         [TestMethod]
-        [TestCategory("Rule")]
-        public void UnusedPrivateMember_UsedInGeneratedFile()
+        public void UnusedPrivateMember_GuidFP()
```
costin-zaharia-sonarsource, line 218: I can't put this inside the `UnusedPrivateMember.cs` with the other test cases because the codefix will remove the constructor and the file will not compile.

andrei-epure-sonarsource: shouldn't this test better be inside `UnusedPrivateMemberTest.Types.cs` ?

andrei-epure-sonarsource (approved):
LGTM, just a final suggestion

PR #3556, opened 2020-08-24
Fix RSPEC update leftover

andrei-epure-sonarsource:
my bad - it seems I missed them (there were other changes which I ignored and due to the noise I missed these)


andrei-epure-sonarsource:
@costin-zaharia-sonarsource  this should say as draft until #3557 gets merged

costin-zaharia-sonarsource (approved):
LGTM!

PR #3560, opened 2020-08-25
Update release workflow

tomverin:
Fixes #

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,7 +36,7 @@ jobs:
           RELEASE_SSH_USER: ${{ secrets.RELEASE_SSH_USER }}
           RELEASE_SSH_KEY: ${{ secrets.RELEASE_SSH_KEY }}
         # Put your action repo here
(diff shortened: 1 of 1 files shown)


malena-ebert-sonarsource (approved):
LGFM

.github/workflows/release.yml
```diff
@@ -7,28 +7,36 @@ on:
...
-        id: lt_release
-        
+        id: lt_release        
         with:
+          publish_to_binaries: true
           distribute: true
           attach_artifacts_to_github_release: true
           run_rules_cov: true
         env:
-          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
```
andrei-epure-sonarsource, line 29: for my education @tomverin  - the GH_TOKEN is not used anymore? I see it's not present in the environment anymore

andrei-epure-sonarsource: actually, I see there's a `RELEASE_GITHUB_TOKEN` now, so nevermind 👍

PR #3561, opened 2020-08-25
Remove ignored performance test

costin-zaharia-sonarsource:
- it does not run and is no longer relevant since #2474 is now fixed


pavel-mikula-sonarsource (approved):
LGTM

PR #3562, opened 2020-08-25
SQ plugin: Fall back to default when Roslyn encoding is not present

### LINKED ISSUE #3316: SQ plugin: Fall back to default when Roslyn encoding is not present
### Description

- the C# analyzer generates file-metadata.pb file with data extracted from the Roslyn API [here](https://github.com/SonarSource/sonar-dotnet/blob/734df7968037387faade9353c7aedd1a77c010a0/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/Utilities/FileMetadataAnalyzerBase.cs#L48) - note that if the Roslyn API gives no encoding, an empty string is used

- the SQ plugin reads the data from the protobuf file [here](https://github.com/SonarSource/sonar-dotnet/blob/366b5b7589beeede0d21019afa645e9c65c6798d/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractGlobalProtobufFileProcessor.java#L75) and tries to use the encoding [here](https://github.com/SonarSource/sonar-dotnet/blob/366b5b7589beeede0d21019afa645e9c65c6798d/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFile.java#L49). When no encoding is found, it logs a confusing message like 
```
    WARN: File ‘FOO.cs’ does not have encoding information. Skip it.
```
 - actually the encoding is not found in our metadata, even if the file does have an encoding information on it. 

See [this community thread](https://community.sonarsource.com/t/from-sonarqube-7-7-to-8-2-no-issues-found-with-sonarscanner-msbuild/23325/2)

### What needs to be done

The SQ plugin should probably fallback on the configured default encoding - `sonar.sourceEncoding` or the default system locale/encoding (the scanner has access to that information as it prints it ).
And the warning that the SQ plugin logs should be that a fallback has been used (specifying the value).

### Related information

* C#/VB.NET Plugins Version 8.6

Also note what the user said on the community thread
> I foud a fix with a SonarQube 8.2 c# plugin downgrade:  sonar-csharp-plugin-8.6.0.16497.jar -> sonar-csharp-plugin-7.11.0.8083.jar .

pavel-mikula-sonarsource:
Fixes #3316

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFile.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFile.java
@@ -47,7 +47,7 @@ public class EncodingPerFile {
     Charset roslynEncoding = globalReportProcessor.getRoslynEncodingPerUri().get(uri);
     if (roslynEncoding == null) {
       // Roslyn saw the file, but it didn't recognized it's encoding. We fall back to the default encoding and accept the file.
-      LOG.warn("Roslyn encoding was not detected for '{}', using default instead.", uri);
+      LOG.warn("Roslyn can not detect encoding for '{}', using default instead.", uri);
       return true;
     }
 
--- a/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFile.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFile.java
@@ -21,8 +21,8 @@ package org.sonarsource.dotnet.shared.plugins;
 
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import org.sonar.api.scanner.ScannerSide;
 import org.sonar.api.batch.fs.InputFile;
+import org.sonar.api.scanner.ScannerSide;
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 
@@ -44,10 +44,11 @@ public class EncodingPerFile {
     }
 
     // Case-Insensitive check
-    Charset roslynEncoding = globalReportProcessor.getRoslynEncodingPerUri().get(uri); 
+    Charset roslynEncoding = globalReportProcessor.getRoslynEncodingPerUri().get(uri);
     if (roslynEncoding == null) {
-      LOG.warn("File '{}' does not have encoding information. Skip it.", uri);
-      return false;
+      // Roslyn saw the file, but it didn't recognized it's encoding. We fall back to the default encoding and accept the file.
+      LOG.warn("Roslyn encoding was not found for '{}', using default instead.", uri);
+      return true;
     }
 
     Charset sqEncoding = inputFile.charset();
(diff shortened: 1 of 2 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFile.java
```diff
@@ -44,10 +44,11 @@ boolean encodingMatch(InputFile inputFile) {
     }
 
     // Case-Insensitive check
-    Charset roslynEncoding = globalReportProcessor.getRoslynEncodingPerUri().get(uri); 
+    Charset roslynEncoding = globalReportProcessor.getRoslynEncodingPerUri().get(uri);
     if (roslynEncoding == null) {
-      LOG.warn("File '{}' does not have encoding information. Skip it.", uri);
-      return false;
+      // Roslyn saw the file, but it didn't recognized it's encoding. We fall back to the default encoding and accept the file.
+      LOG.warn("Roslyn encoding was not found for '{}', using default instead.", uri);
```
costin-zaharia-sonarsource, line 50: Wouldn't be more clear to write something like: "File encoding cannot be identified from Roslyn metadata for '{}', using default instead."

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFile.java
```diff
@@ -44,10 +44,11 @@ boolean encodingMatch(InputFile inputFile) {
...
+      return true;
```
costin-zaharia-sonarsource, line 51: Shouldn't this be an info message instead of warning? Since we populate the encoding and we don't provide any advice for the user, I'm not sure the user can do something about it.

costin-zaharia-sonarsource (changes requested):

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFile.java
pavel-mikula-sonarsource: `Roslyn metadata` seems ambiguous to me. It could refer to real Roslyn internal state or the `file-metadata.pb`.

I've tried to update the message to `Roslyn encoding was not detected for '{}', using default instead.`

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFile.java
pavel-mikula-sonarsource: I'd prefer a warning. It should be rare, as I wasn't able to reproduce it. And it could lead to misencoded files in the SQ/SC if the situation happens.

And since the default isn't UTF-8 for some weird java-world reasons, user can either explicitly use the file with BOM, configure `<CodePage>` in project file or set `sonar.sourceEncoding` to fix this.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/EncodingPerFile.java
costin-zaharia-sonarsource: > Roslyn metadata seems ambiguous to me. It could refer to real Roslyn internal state or the file-metadata.pb

As far as I can tell, the Roslyn state (SyntaxTree.Encoding) is stored in `file-metadata.pb` and processed by this method so both are true :).

What about "Roslyn can not detect encoding for `{}`, using default instead."?

costin-zaharia-sonarsource (approved):
LGTM!

PR #3563, opened 2020-08-26
S1144, S4487: Ignore entity framework migration files

### LINKED ISSUE #2474: Analysis of Entity Framework migration files never completes

costin-zaharia-sonarsource:
Fixes #2474

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
@@ -52,12 +52,13 @@ namespace SonarAnalyzer.Rules.CSharp
                 fadeOutCode: true);
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(ruleS1144, ruleS4487);
 
-        private static readonly ImmutableArray<KnownType> IgnoredTypes =
+        private static readonly ImmutableArray<KnownType> ignoredTypes =
             ImmutableArray.Create(
                 KnownType.UnityEditor_AssetModificationProcessor,
                 KnownType.UnityEditor_AssetPostprocessor,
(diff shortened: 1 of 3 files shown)


pavel-mikula-sonarsource (changes requested):
I'd like to add a unit test to see it in action 🙂

costin-zaharia-sonarsource:
> I'd like to add a unit test to see it in action 🙂

I don't understand what do you mean 😄

pavel-mikula-sonarsource:
We already have a performance test that's implemented by ignoring Guid. But this PR says that everything inside Migration class is ignored. And we don't have a test for that.

I'd like to have a class inheriting Migration with few private unused lines (like in TestCases/UnusedPrivateMember.cs) with comments that it's compliant due to ignored type.

pavel-mikula-sonarsource (approved):
LGTM

PR #3564, opened 2020-08-26
Use the latest version of .Net 5

costin-zaharia-sonarsource:


azure-pipelines.yml
```diff
@@ -150,7 +150,7 @@ stages:
       - task: UseDotNet@2
         inputs:
           packageType: sdk
-          version: 5.0.100-preview.7.20366.6
+          version: 5.0.100-preview.8.20417.9
```
pavel-mikula-sonarsource, line 153: Same update should be done in `azure-pipelines-external.yml` for consistency.

pavel-mikula-sonarsource, line 153: Should `its\sources\Net5\global.json` be updated as well? There will be an impact on the GCP machines as well.

pavel-mikula-sonarsource, line 153: What is the motivation to do this now? Regarding the sideeffects on peachee jobs that take more time to finish.

pavel-mikula-sonarsource (changes requested):

azure-pipelines.yml
costin-zaharia-sonarsource: > Should its\sources\Net5\global.json be updated as well?

It's interesting that it worked also with the old version since it was not installed.

> There will be an impact on the GCP machines as well.

Not sure what you mean. GCP machines are running a different build and they have different config. Sure, we should update also them but they are not related.

costin-zaharia-sonarsource: Preview 8 was released and it's now feature complete. This makes preview 7 obsolete. The main idea is to have early feedback.

pavel-mikula-sonarsource: Hmm, it shouldn't work. There must be some preview-aware logic in place.

pavel-mikula-sonarsource: Side effects: To fully move to preview 8, we need to reconfigure/rebuild peachee VM to install that SDK.

pavel-mikula-sonarsource (approved):
LGTM

azure-pipelines.yml
costin-zaharia-sonarsource: Indeed, this is just the fist step. I will update the image and the build script as well.

PR #3570, opened 2020-08-27
S1144, S4487: Read symbols only for known names

costin-zaharia-sonarsource:
By retrieving symbols only for known names, when analyzing efcore, we reduce the analysis time from ~854678ms to 
~210854ms

Partial fix for #2474

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
@@ -20,7 +20,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -42,11 +41,6 @@ namespace SonarAnalyzer.Helpers
             SyntaxKind.PreDecrementExpression
         };
 
-        // The types names (from .Net) defined here are very common and can slow down the analysis significantly
-        // (see entity framework migration files) but use this list carefully since it can lead to false positives
-        // if the names are redefined in the user code.
-        private static readonly ImmutableHashSet<string> ignoredConstructors = ImmutableHashSet.Create("Guid", "System.Guid");
-
         private readonly Func<SyntaxNode, SemanticModel> getSemanticModel;
         private readonly HashSet<string> knownSymbolNames;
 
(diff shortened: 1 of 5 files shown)


pavel-mikula-sonarsource (changes requested):
Intermediate batch

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
```diff
@@ -103,104 +93,44 @@ public override void VisitAttribute(AttributeSyntax node)
...
-        private void TryStoreFieldAccess(IdentifierNameSyntax node, List<ISymbol> symbols)
-        {
-            var access = ParentAccessType(node);
-            var fieldSymbolUsagesList = GetFieldSymbolUsagesList(symbols);
-            if (HasFlag(access, SymbolAccess.Read))
+            if (HasKnownIdentifier(node))
             {
-                fieldSymbolUsagesList.ForEach(usage => usage.Readings.Add(node));
-            }
+                var symbols = GetSymbols(node).ToList();
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Symbols are iterated more than once and this is the reason why a list or other data structure which is not evaluated when consumed is needed. Returning immutable array will do.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
```diff
@@ -258,17 +193,66 @@ public override void VisitPropertyDeclaration(PropertyDeclarationSyntax node)
...
         /// </summary>
         /// <returns>List of symbols</returns>
-        private IEnumerable<ISymbol> GetSymbols<TSyntaxNode>(TSyntaxNode node, Func<TSyntaxNode, bool> condition)
+        private IEnumerable<ISymbol> GetSymbols<TSyntaxNode>(TSyntaxNode node)
             where TSyntaxNode : SyntaxNode
         {
-            return condition(node)
-                ? GetCandidateSymbols(getSemanticModel(node).GetSymbolInfo(node))
-                : Enumerable.Empty<ISymbol>();
+            return GetCandidateSymbols(getSemanticModel(node).GetSymbolInfo(node));
```
pavel-mikula-sonarsource, line 255: Local function `GetCandidateSymbols` doesn't have to exist anymore after this update.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
```diff
@@ -347,12 +331,38 @@ private AccessorAccess EvaluatePropertyAccesses(ExpressionSyntax node)
                 : AccessorAccess.Get;
         }
 
-        private bool IsKnownIdentifier(SimpleNameSyntax nameSyntax) =>
-            knownSymbolNames.Contains(nameSyntax.Identifier.ValueText);
+        private bool HasKnownIdentifier(SimpleNameSyntax nameSyntax) =>
```
pavel-mikula-sonarsource, line 334: Can you remove this function and embed it's content on call sites? There are just two.

While I was reading the `VisitIdentifierName` the name `HasKnownIdentifier` made an impression that all subnodes are inspected somehow. Even when the function name is good.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
```diff
@@ -103,114 +94,51 @@ public override void VisitAttribute(AttributeSyntax node)
...
+                UsedSymbols.UnionWith(GetSymbols(node));
+            }
+
             base.VisitGenericName(node);
         }
 
         public override void VisitElementAccessExpression(ElementAccessExpressionSyntax node)
         {
-            var symbols = GetSymbols(node, x => true).ToList();
+            var symbols = GetSymbols(node);
```
pavel-mikula-sonarsource, line 130: Can this be pre-filtered with the name-check as well? (maybe in another PR)

pavel-mikula-sonarsource (approved):
LGTM

PR #3573, opened 2020-08-28
Symbolic Execution refactoring

pavel-mikula-sonarsource:


pavel-mikula-sonarsource:
Burgr doesn't like my branch name

Issue #3403, opened 2020-06-26
SymbolicExecutionRunner throws SymbolicExecutionException


costin-zaharia-sonarsource:
### Description

Source: https://community.sonarsource.com/t/symbolicexecutionrunner-threw-an-exception-of-type-symbolicexecutionexception/27056

Our project is bound to SonarQube and has SonarAnalyzer.CSharp installed and the SonarLint plugin. When I build I’m seeing the following warning:

Warning AD0001 Analyzer ‘SonarAnalyzer.Rules.SymbolicExecution.SymbolicExecutionRunner’ threw an exception of type ‘SonarAnalyzer.SymbolicExecution.SymbolicExecutionException’ with message
'Error processing method: xxxxxxxxxxxxxxxx ##
Method file: C:\Perforce\xxx\xxxx\xxx\xxxxxxxxxxxx\xxxxxxxx\xxxxxxxxxxxxxx.cs ##
Method line: 1283,8 ##

### Stacktrace

```
Inner exception: System.NotSupportedException: Neither one of BoolConstraint, ObjectConstraint,StringConstraint or DisposableConstraint. ##
at SonarAnalyzer.SymbolicExecution.SymbolicValue.TrySetObjectConstraint(ObjectConstraint constraint, SymbolicValueConstraints oldConstraints, ProgramState programState) ##
at SonarAnalyzer.SymbolicExecution.SymbolicValue.TrySetConstraint(SymbolicValueConstraint constraint, ProgramState programState) ##
at System.Linq.Enumerable.d__172.MoveNext() ## at System.Linq.Enumerable.<SelectManyIterator>d__172.MoveNext() ##
at SonarAnalyzer.SymbolicExecution.CSharpExplodedGraph.VisitBinaryBranch(BinaryBranchBlock binaryBranchBlock, ExplodedGraphNode node, SyntaxNode instruction) ##
at SonarAnalyzer.SymbolicExecution.AbstractExplodedGraph.Walk() ##
at SonarAnalyzer.Rules.SymbolicExecution.SymbolicExecutionRunner.Analyze(CSharpExplodedGraph explodedGraph, SyntaxNodeAnalysisContext context) ##
[...]

PR #3574, opened 2020-08-28
Improve SymbolicValue error message and seal DisposableConstraint

pavel-mikula-sonarsource:
This should help with future diagnostics of #3403 

All SE constraints except DisposableConstraint were sealed.


costin-zaharia-sonarsource (approved):
LGTM!

PR #3576, opened 2020-08-28
Symbolic Execution refactoring

pavel-mikula-sonarsource:


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/SymbolicValues/NullableSymbolicValue.cs
```diff
@@ -68,31 +66,18 @@ public NullableSymbolicValue(SymbolicValue wrappedValue)
...
-        public override string ToString()
-        {
-            if (base.identifier != null)
-            {
-                return $"NULLABLE_SV_{base.identifier}";
-            }
-
-            return WrappedValue?.ToString() ?? base.ToString();
-        }
+        public override string ToString() => $"NULLABLE_SV_{identifier}" ;
```
costin-zaharia-sonarsource, line 81: The previous version displayed the `WrappedValue` when the `identifier` was null. This is no longer needed?

costin-zaharia-sonarsource (commented):
Just one comment regarding the NullableSymbolicValue.ToString change.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/SymbolicValues/NullableSymbolicValue.cs
pavel-mikula-sonarsource: There's no way to make `identifier` null. The `identifier` is set in base class constructor (it's the counter). This class cannot break it unless new constructor with `: base(null)` is added. So it was useless anyway and could be removed.

costin-zaharia-sonarsource (approved):
LGTM!

PR #3578, opened 2020-08-31
S1144, S4487: Check known types for ElementAccessExpression

### LINKED ISSUE #2474: Analysis of Entity Framework migration files never completes

costin-zaharia-sonarsource:
Fixes #2474

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
@@ -127,9 +127,14 @@ namespace SonarAnalyzer.Helpers
 
         public override void VisitElementAccessExpression(ElementAccessExpressionSyntax node)
         {
-            var symbols = GetSymbols(node);
-            UsedSymbols.UnionWith(symbols);
-            TryStorePropertyAccess(node, symbols);
+            if (node.Expression.IsKind(SyntaxKind.ThisExpression)
+                || knownSymbolNames.Contains(node.Expression.GetIdentifier()?.Identifier.ValueText))
+            {
+                var symbols = GetSymbols(node);
+                UsedSymbols.UnionWith(symbols);
+                TryStorePropertyAccess(node, symbols);
+            }
+
             base.VisitElementAccessExpression(node);
         }
 
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnusedPrivateMember.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnusedPrivateMember.cs
@@ -83,7 +83,7 @@ namespace Tests.Diagnostics
 
         public int GetItem(int i) => used[i];
 
-        private Dictionary<int, int> GetDictionary() => used;
+        private Dictionary<int, int> GetDictionary() => null;
 
         public int GetDictionaryItem(int i) => GetDictionary()[i];
 
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnusedPrivateMember.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnusedPrivateMember.cs
@@ -75,10 +75,17 @@ namespace Tests.Diagnostics
         private delegate void Delegate2(); // Noncompliant {{Remove the unused private type 'Delegate2'.}}
         private event Delegate Event; //Noncompliant {{Remove the unused private event 'Event'.}}
         private event Delegate MyEvent; //Noncompliant {{Remove this unread private field 'MyEvent' or refactor the code to use its value.}}
+        private int[][] array = new int[0][];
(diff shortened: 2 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
```diff
@@ -127,9 +127,14 @@ public override void VisitGenericName(GenericNameSyntax node)
 
         public override void VisitElementAccessExpression(ElementAccessExpressionSyntax node)
         {
-            var symbols = GetSymbols(node);
-            UsedSymbols.UnionWith(symbols);
-            TryStorePropertyAccess(node, symbols);
+            if (node.Expression.IsKind(SyntaxKind.ThisExpression)
+                || knownSymbolNames.Contains(node.Expression.GetIdentifier()?.Identifier.ValueText))
```
pavel-mikula-sonarsource, line 131: _Just thinking loud:_ In case `GetIdentifier()` returns `null`: We will not move forward and we will not inspect the symbol. Unless there's `null` in `knownSymbolNames`.

That feels as inconsistent and little hidden from readability POV. Can you do an explicit condition on the null identifier?

Not inspecting the symbol should be fine, right? Stuff like `GetDictionary()[index]` or `array[index][index]` should be recursively visited and handled on the top level.

pavel-mikula-sonarsource, line 131: Can you add testcases for `GetDictionary()[index]` or `array[index][index]`

pavel-mikula-sonarsource (changes requested):

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
costin-zaharia-sonarsource: Indeed:
- for `array[index][index]` we get null identifier but there will be a second call with `array[index]` which will have `array` as identifier. The behavior is OK since we are not concerned with usage of the first part.
- for `GetDictionary()[index]` is basically the same case null identifier here and `VisitIdentifierName` will take care of `GetDictionary`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnusedPrivateMember.cs
```diff
@@ -75,6 +75,17 @@ private struct Struct { }// Noncompliant {{Remove the unused private type 'Struc
...
         private event Delegate MyEvent; //Noncompliant {{Remove this unread private field 'MyEvent' or refactor the code to use its value.}}
+        private int[][] array = new int[0][];
+        private Dictionary<int, int> used = new Dictionary<int, int>();
+        private Dictionary<int, int> unused = new Dictionary<int, int>(); // Noncompliant
+
+        public int GetValue(int x, int y) => array[x][y];
+
+        public int GetItem(int i) => used[i];
+
+        private Dictionary<int, int> GetDictionary() => used;
```
pavel-mikula-sonarsource, line 86: Can you return null here instead? Like this, we're not sure if `used` is used due to `used[i]` or this call. Having `used[i]` has more value as originally intended.

```suggestion
        private Dictionary<int, int> GetDictionary() => null;
```

pavel-mikula-sonarsource (changes requested):
The last commit obfuscated the first one a little

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnusedPrivateMember.cs
costin-zaharia-sonarsource: In this case `.Fixed.cs` and `.Fixed.Batch.cs` need to be updated too.

pavel-mikula-sonarsource (approved):
LGTM

PR #3579, opened 2020-08-31
Verify Unit Tests only once in local tests

### LINKED ISSUE #2855: SonarDiagnosticAnalyzer.Initialize is called 4 times for unit tests
### Description

Running unit test in Visual Studio fires every rule 4 times. This is caused by `SonarDiagnosticAnalyzer.Initialize` method being called 4 times, therefore rule registers itself 4 times for running.

### Repro steps

Just put breakpoint into Initialize method and debug a test.

It has been confirmed on different machines, MSBuild 15 and 16 for these rules:
S1226 ParameterAssignedTo
S4143 DoNotOverwriteCollectionElement

### Expected behavior

Initialize should be called only once.

### Production

SonarLint and Integration tests fires the rule only once.

pavel-mikula-sonarsource:
Fixes #2855


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -207,40 +203,39 @@ private static void VerifyBuildErrors(ImmutableArray<Diagnostic> diagnostics, Co
...
-            Location location, string message, string issueId)
-        {
-            VerifyIssue(expectedIssues, issueFilter, location, message, null, false, issueId);
-        }
+        private static void VerifySecondaryIssue(Compilation compilation, IList<IIssueLocation> expectedIssues, Func<IIssueLocation, bool> issueFilter,
+            Location location, string message, string issueId) =>
+            VerifyIssue(compilation, expectedIssues, issueFilter, location, message, null, false, issueId);
 
-        private static string VerifyIssue(IList<IIssueLocation> expectedIssues, Func<IIssueLocation, bool> issueFilter,
+        private static string VerifyIssue(Compilation compilation, IList<IIssueLocation> expectedIssues, Func<IIssueLocation, bool> issueFilter,
```
costin-zaharia-sonarsource, line 214: - This method needs only the language string and not the full compilation object so we should pass version instead.
- We can use the more generic type `ICollection<IIssueLocation>` instead of `IList<IIssueLocation>` for expectedIssues

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/ParseOptionsHelper.cs
```diff
@@ -27,118 +27,77 @@
...
     {
-        private static readonly IEnumerable<ParseOptions> defaultParseOptions;
+        public static ImmutableArray<ParseOptions> BeforeCSharp7 { get; }
+        public static ImmutableArray<ParseOptions> BeforeCSharp8 { get; }
+
+        public static ImmutableArray<ParseOptions> FromCSharp6 { get; }
+        public static ImmutableArray<ParseOptions> FromCSharp7 { get; }
+        public static ImmutableArray<ParseOptions> FromCSharp8 { get; }
+
+        public static ImmutableArray<ParseOptions> FromVisualBasic12 { get; }
```
costin-zaharia-sonarsource, line 51: It seems this is used only in this class and it can be made private.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/ParseOptionsHelper.cs
```diff
@@ -27,118 +27,77 @@
...
+            var cs7 = CreateOptions(CSharp7, CSharp7_1, CSharp7_2, CSharp7_3);
+            var vb15 = CreateOptions(VisualBasic15, VisualBasic15_3, VisualBasic15_5);
+
+            BeforeCSharp7 = CreateOptions(CSharp5).Concat(CreateOptions(CSharp6)).FilterByEnvironment();
+            BeforeCSharp8 = BeforeCSharp7.Concat(cs7).FilterByEnvironment();
 
-            FromCSharp8 = CSharp8;
-            FromCSharp7 = CSharp7.Concat(FromCSharp8);
-            FromCSharp6 = CSharp6.Concat(FromCSharp7);
+            FromCSharp8 = CreateOptions(CSharp8).FilterByEnvironment();
```
costin-zaharia-sonarsource, line 67: As far as I can tell, when the code is run on local machines the full list of options will be create and then we will filter it and only take the last. It would be great to create it right from the beginning.

One option would be to have CreateOptions do also the filtering and take only the last item if needed. Something similar with:
`BeforeCSharp8 = CreateOptions(CSharp5, CSharp6, CSharp7, CSharp7_1, CSharp7_2, CSharp7_3);`
In this way the versions used are clear (no need to check other methods) and we will avoid multiple iterations.

costin-zaharia-sonarsource (changes requested):

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/DiagnosticVerifierTest.cs
```diff
@@ -242,11 +242,11 @@ public class ExpectedIssuesNotRaised
     }",
                     new BinaryOperationWithIdenticalExpressions());
 
-            action.Should().Throw<AssertFailedException>()
-                  .WithMessage(@"Issue(s) expected but not raised on line(s):
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: There were several concatenation in this file and only this last one was not concatenated. I did this to unify the file and to use the `Environment.NewLine` that is used in other tests.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/ParseOptionsHelper.cs
pavel-mikula-sonarsource: The create&filter is done only locally and only in tests. I'd keep it this way for simplicity, it's not expensive in this context. 

I considered the full-listing before doing this, but it would lead to situation like this in the future. And that would be hard to maintain and easy to make a typo.
```
BeforeCSharp8 = CreateOptions(CSharp5, CSharp6, CSharp7, CSharp7_1, CSharp7_2, CSharp7_3);
BeforeCSharp9 = CreateOptions(CSharp5, CSharp6, CSharp7, CSharp7_1, CSharp7_2, CSharp7_3, CSharp8);
BeforeCSharp10 = CreateOptions(CSharp5, CSharp6, CSharp7, CSharp7_1, CSharp7_2, CSharp7_3, CSharp8, CSharp9);
BeforeCSharp11 = CreateOptions(CSharp5, CSharp6, CSharp7, CSharp7_1, CSharp7_2, CSharp7_3, CSharp8, CSharp9, CSharp10);
```
and even worse (non-tabular)
```
FromCSharp11 = CreateOptions(CSharp11)
FromCSharp10 = CreateOptions(CSharp10, CSharp11)
FromCSharp9 = CreateOptions(CSharp9, CSharp10, CSharp11)
FromCSharp8 = CreateOptions(CSharp8, CSharp9, CSharp10, CSharp11)
FromCSharp7 = CreateOptions(CSharp7, CSharp7_1, CSharp7_2, CSharp7_3, CSharp8, CSharp9, CSharp10, CSharp11)
FromCSharp6 = CreateOptions(CSharp6, CSharp7, CSharp7_1, CSharp7_2, CSharp7_3, CSharp8, CSharp9, CSharp10, CSharp11);
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/DiagnosticVerifierTest.cs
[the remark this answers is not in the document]
costin-zaharia-sonarsource: I think we should keep things simple where possible. We don't need to apply the same rule for all the cases.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -76,13 +76,13 @@ public static void Verify(Compilation compilation, DiagnosticAnalyzer[] diagnost
             }
         }
 
-        private static void CompareActualToExpected(Compilation compilation, IEnumerable<Diagnostic> diagnostics, List<IIssueLocation> expectedIssues, bool compareIdToMessage)
+        private static void CompareActualToExpected(Compilation compilation, IEnumerable<Diagnostic> diagnostics, ICollection<IIssueLocation> expectedIssues, bool compareIdToMessage)
```
costin-zaharia-sonarsource, line 79: This method needs only the language string and not the full compilation object so we should pass version instead.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/ParseOptionsHelper.cs
costin-zaharia-sonarsource: Ok. It's a matter of preference in the end.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Helpers/DiagnosticVerifierTest.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Regular code base is easier to maintain. And the new version has the `Environment.NewLine` functional improvement.

costin-zaharia-sonarsource (approved):
LGTM!

PR #3580, opened 2020-09-01
Fix S4275 FP: Consider ??= as assignment

### LINKED ISSUE #3577: Fix S4275 FP: Consider ??= as assignment
### Repro steps

this doesn't work

```chsarp
        private ValueType? _value;

        [Inject]
        public ValueType Value
        {
            get => _value ??
                   throw new InvalidOperationException($"The {nameof(Value)} hasn't been initialized yet.");

            set
            {
                _value ??= value; // S4275

                if (!ReferenceEquals(_value, value))
                {
                    throw new InvalidOperationException($"The {nameof(Value)} was already initialized.");
                }
            }
        }

```

**Why would I want to do that?**

It seems to be the easiest way to guarantee that the value is set exactly once. This is only needed, when properties are injected via DI. This particular example is from Blazor.

### Known workarounds

the moment I write inside that setter `_value = value` it's ok.

### Related information

* C#/VB.NET Plugins Version: SonarAnalyzer.CSharp@8.12.0.21095
* Visual Studio Version: N/A
* MSBuild / dotnet version: .NET Core SDK (3.1.401)
* Operating System: MacOS

pavel-mikula-sonarsource:
Fixes #3577


costin-zaharia-sonarsource (approved):
LGTM!

PR #3581, opened 2020-09-01
S4017: Don't count Nullable<T> shorthand

### LINKED ISSUE #3277: S4017: Don't count Nullable<T> shorthand when used as a generic type parameter

costin-zaharia-sonarsource:
Fixes #3277

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotNestTypesInArguments.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotNestTypesInArguments.cs
@@ -99,7 +99,11 @@ namespace SonarAnalyzer.Rules.CSharp
                     return;
                 }
 
-                if (!namedTypeSymbol.ConstructedFrom.IsAny(ignoredTypes))
+                if (namedTypeSymbol.ConstructedFrom.IsAny(ignoredTypes))
+                {
+                    base.VisitGenericName(node);
+                }
+                else
                 {
                     if (depth == maxDepth - 1)
                     {
@@ -111,10 +115,6 @@ namespace SonarAnalyzer.Rules.CSharp
                         base.VisitGenericName(node);
                         depth--;
                     }
-                }
-                else
-                {
-                    base.VisitGenericName(node);
                 }
             }
         }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotNestTypesInArguments.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotNestTypesInArguments.cs
@@ -38,25 +38,29 @@ namespace SonarAnalyzer.Rules.CSharp
         private const string DiagnosticId = "S4017";
         private const string MessageFormat = "Refactor this method to remove the nested type argument.";
 
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        private static readonly DiagnosticDescriptor rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
 
-        protected override void Initialize(SonarAnalysisContext context)
-        {
+        protected override void Initialize(SonarAnalysisContext context) =>
             context.RegisterSyntaxNodeActionInNonGenerated(
-            c =>
-            {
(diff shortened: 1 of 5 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotNestTypesInArguments.cs
```diff
@@ -38,25 +38,29 @@ public sealed class DoNotNestTypesInArguments : SonarDiagnosticAnalyzer
...
+                    }
+                },
+                SyntaxKind.MethodDeclaration,
+                SyntaxKindEx.LocalFunctionStatement);
+
+        private static bool MaxDepthReached(SyntaxNode parameterSyntax, SemanticModel model)
+        {
+            var walker = new GenericWalker(2, model);
+            walker.SafeVisit(parameterSyntax);
+            return walker.HasReachedMaxDepth;
```
pavel-mikula-sonarsource, line 63: I suggest to give it the same name for consistency. Just to avoid confusion if the different wording has different meaning.
```suggestion
            return walker.MaxDepthReached;
```

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotNestTypesInArguments.cs
```diff
@@ -67,37 +71,52 @@ node switch
...
             {
-                return typeArguments.Max(p => GetNestingDepth(p, nextDepth, false));
-            }
+                if (!(model.GetSymbolInfo(node).Symbol is INamedTypeSymbol namedTypeSymbol))
+                {
+                    return;
+                }
 
-            return depth;
+                if (!namedTypeSymbol.ConstructedFrom.IsAny(ignoredTypes))
```
pavel-mikula-sonarsource, line 102: I'd prefer positive condition here and swap if/else blocks. The `not is any` is hard :)

pavel-mikula-sonarsource (changes requested):
Nice! The expected issues are still failing.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotNestTypesInArguments.cs
```diff
@@ -67,37 +71,52 @@ node switch
...
+                }
 
-            return depth;
+                if (namedTypeSymbol.ConstructedFrom.IsAny(ignoredTypes))
+                {
+                    base.VisitGenericName(node);
+                }
+                else
+                {
+                    if (depth == maxDepth - 1)
```
pavel-mikula-sonarsource, line 108: This condition feels weird. Can we change the initialization constant or do the `- 1` in constructor to have
```C#
if (depth == maxDepth)
{
    MaxDepthReached = true
}
```
Because you reach something once you get there. Current state declares that you're almost there.

pavel-mikula-sonarsource (approved):
LGTM, I just left one more improvement suggestion

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DoNotNestTypesInArguments.cs
costin-zaharia-sonarsource: The depth starts at 0 and increases every time `base.VisitGenericName(node);` is called. In our cases it reaches the max limit since we have to call the method but we exit before doing the incrementation (that's why we have -1 in the condition.). In my opinion starting from -1 would hide this and make the code harder to understand.

pavel-mikula-sonarsource: I didn't want to change the initial value of `depth` but the value of `maxDepth`

pavel-mikula-sonarsource: Like receive `depthCount` in constructor and store `this.maxDepth = depthCount - 1;`

PR #3582, opened 2020-09-01
Fix S3240 FP: Nested conditional operators

### LINKED ISSUE #3468: S3240 false positive on nested (potential) condtional operators
### Description
Rule S3240 suggest when a conditional operator can be used, also when one of the if-branches already contains a conditional operator. When applied, you get a nested conditional operator, which is not okay according to S3358. 

### Repro steps
``` C#
public static class S3240
{
    public static int NoConditionalOperator(int value)
    {
        if (value % 2 == 0) // none-compliant. Will lead to nested conditionals.
        {
            return value == 2 ? 1 : 0;
        }
        else
        {
            return 17;
        }
    }
}
```

### Related information
* v8.10

costin-zaharia-sonarsource:
Fixes #3468

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
@@ -369,6 +369,6 @@ namespace SonarAnalyzer.Rules.CSharp
         }
 
         private static bool IsConditionalStructure(SyntaxNode syntaxNode) =>
-            syntaxNode.IsAnyKind(SyntaxKind.ConditionalExpression, SyntaxKind.CoalesceExpression, SyntaxKindEx.CoalesceAssignmentExpression);
+            syntaxNode.IsAnyKind(SyntaxKind.ConditionalExpression, SyntaxKindEx.SwitchExpression);
     }
 }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
@@ -41,7 +41,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private const string MessageFormat = "Use the '{0}' operator here.";
 
         private static readonly DiagnosticDescriptor rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-        private static readonly ISet<SyntaxKind> EqualsOrNotEquals = new HashSet<SyntaxKind>
+        private static readonly ISet<SyntaxKind> equalsOrNotEquals = new HashSet<SyntaxKind>
         {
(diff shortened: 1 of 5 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalSimplification.FromCSharp8.cs
```diff
@@ -351,4 +351,28 @@ public int NestedTernary(bool condition1, bool condition2, int a, int b, int c)
             return c;
         }
     }
+
+    public int? NestedNullCoalescing(bool condition, int? a, int b)
+    {
+        if (condition) // Compliant, changing this will lead to nested conditionals
```
pavel-mikula-sonarsource, line 357: Are you sure about this? Having
```
return condition
  ? b
  : a ?? b;
```
seems simple and understandable even with more complicated expressions. If I remember what Ann said about the complexity metrics than `??` is considered simple by it's purpose.

I'd prefer to ban switch expressions instead.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
```diff
@@ -370,13 +360,15 @@ private static bool ExpressionCanBeNull(ExpressionSyntax expression, SemanticMod
...
-            ret.Add(IsCoalesceAssignmentSupportedKey, IsCoalesceAssignmentSupported(c).ToString());
+            var ret = new Dictionary<string, string> {{IsCoalesceAssignmentSupportedKey, IsCoalesceAssignmentSupported(c).ToString()}};
             if (simplifiedOperator != null)
             {
                 ret.Add(SimplifiedOperatorKey, simplifiedOperator);
             }
             return ret.ToImmutableDictionary();
         }
+
+        private static bool IsConditionalStructure(SyntaxNode syntaxNode) =>
```
pavel-mikula-sonarsource, line 371: Is this enough? I'd prefer to change the logic to `ContainsConditionalStructure`. This will not work for `return (condition ? a : b)` or some more complex expression (concatenations) containing ternary.

I think we should inspect the nested nodes and stop the nested nodes inspection on invocation syntax. To allow `SomeInvocation(condition ? a : b)`. Those inner arguments should be simple by-design.

pavel-mikula-sonarsource (changes requested):

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
pavel-mikula-sonarsource: Like this `syntaxNode.DescendantNodesAndSelf(x => !(x is InvocationExpressionSyntax))`

costin-zaharia-sonarsource: As discussed I will change this to look for ternary op and switch expressions in all descendants (and self).

pavel-mikula-sonarsource (approved):
LGTM

PR #3584, opened 2020-09-02
Make sure preprocessor symbols are not overwritten when signing assemblies

costin-zaharia-sonarsource:

### DIFF
--- a/sonaranalyzer-dotnet/src/targets/CommonProperties.targets
+++ b/sonaranalyzer-dotnet/src/targets/CommonProperties.targets
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <PropertyGroup>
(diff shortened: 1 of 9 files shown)


sonaranalyzer-dotnet/src/targets/CommonProperties.targets
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
pavel-mikula-sonarsource, line 1: Can you put this content in `Directory.Build.props`? You can remove the explicit reference from the csproj files.

pavel-mikula-sonarsource (changes requested):
Pre-review

sonaranalyzer-dotnet/src/targets/CommonProperties.targets
costin-zaharia-sonarsource: Nice! This simplified the PR quite a bit.

pavel-mikula-sonarsource (approved):
LGTM, nice!

PR #3586, opened 2020-09-02
Fix S1481 FP: "var using" declarations should not raise issue

### LINKED ISSUE #3137: Fix S1481 FP: "var using" declarations should not raise issue

pavel-mikula-sonarsource:
Fixes #3137


costin-zaharia-sonarsource (approved):
LGTM!

PR #3588, opened 2020-09-03
Add slack notification on central sync error

tomverin:
Fixes #

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,6 +35,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
           RELEASE_SSH_USER: ${{ secrets.RELEASE_SSH_USER }}
           RELEASE_SSH_KEY: ${{ secrets.RELEASE_SSH_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -35,6 +35,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
           RELEASE_SSH_USER: ${{ secrets.RELEASE_SSH_USER }}
           RELEASE_SSH_KEY: ${{ secrets.RELEASE_SSH_KEY }}
+          SLACK_API_TOKEN: ${{secrets.SLACK_API_TOKEN }}
```
costin-zaharia-sonarsource, line 38: This is needed just to be forwarded to repox?

tomverin: no it's needed for the sync of the java part to maven central, and it's just for notifying builders when this part fails which is quite often.
We don't break the release for that, but we need to be aware of it.

costin-zaharia-sonarsource: Thanks!

malena-ebert-sonarsource (approved):
LGFM

Issue #3565, opened 2020-08-26
S2589 FP: with StringSegment and literal null


rjgotten:
### Description

S2589 reports a false positive when using the equality operator to compare a `StringSegment` struct with a literal `null` value, claiming that the condition always evaluates to `false`. Similarly, with a non-equality operator it claims the result is always `true`.

In actuality `StringSegment` has an implicit conversion operator for the `String` type defined and will convert `null` to a `StringSegment` _pointing_ to a `null` string and then compare both `StringSegment` instances. The type implements custom equality operators which deem a `StringSegment` that points to a `null` string equal only to other `StringSegment` instances that point to a `null` string - i.e. the result of the operation is **not** universally fixed.

This is a dangerous false positive to have, as developers unaware of this nuance of the `StringSegment` type may remove null checks thinking that "it's a struct anyway, so it can't be null; this has to be a mistake..." and in doing so will actively remove conditions that should be guarded against or break logic branches in code.

### Repro steps

```cs
public void DoWork(StringSegment segment)
{
  if (segment == null) throw new ArgumentException("May not point to a null string", nameof(segment));
  // ...
}
```

### Expected behavior

S2589 is not raised.

### Actual behavior

S2589 is raised.

### Known workarounds

Don't compare with literal `null` but use the more obtuse `HasValue` property on the `StringSegment` instance. E.g.

```cs
public void DoWork(StringSegment segment)
{
  if (!segment.HasValue) throw new ArgumentException("May not point to a null string", nameof(segment));
  // ...
}
```

### Related information

* C#/VB.NET Plugins Version: Whatever ships with SonarLint 4.25.0.20544
* Visual Studio Version: 2019 (v16.6.5)
* MSBuild / dotnet version: 3.1.302
* Operating System: Windows 10

Corniel:
Is it not the case, in general, that the usage of this expression should be discouraged, but that it should not suggest that the expression evaluates to false? `StringSegment` is not the only struct with such behaviour.

mary-georgiou-sonarsource:
Fixed by https://github.com/SonarSource/sonar-dotnet/pull/7750

PR #3589, opened 2020-09-03
S2583: Add repro for #3565

pavel-mikula-sonarsource:
Reproduces #3565


pavel-mikula-sonarsource:
QG passed but it was not reported to GitHub
https://sonarcloud.io/dashboard?id=sonaranalyzer-dotnet&pullRequest=3589

andrei-epure-sonarsource (approved):

PR #3591, opened 2020-09-03
Fix S1128 FP: System namespace is needed for tuple deconstruction

### LINKED ISSUE #3408: Fix S1128 FP: System namespace is needed for tuple deconstruction

pavel-mikula-sonarsource:
Fixes #3408

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnnecessaryUsings.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnnecessaryUsings.cs
@@ -36,17 +36,14 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class UnnecessaryUsings : SonarDiagnosticAnalyzer
     {
-
         internal const string DiagnosticId = "S1128";
         private const string MessageFormat = "Remove this unnecessary 'using'.";
 
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        private static readonly DiagnosticDescriptor rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
 
-        protected override void Initialize(SonarAnalysisContext context)
-        {
+        protected override void Initialize(SonarAnalysisContext context) =>
             context.RegisterSyntaxNodeActionInNonGenerated(
                 c =>
                 {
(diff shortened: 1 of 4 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnnecessaryUsings.cs
```diff
@@ -65,32 +63,26 @@ protected override void Initialize(SonarAnalysisContext context)
...
-                if (comment.HasStructure)
-                {
-                    visitor.SafeVisit(comment.GetStructure());
-                }
+                visitor.SafeVisit(comment.GetStructure());
             }
         }
 
-        private static void CheckUnnecessaryUsings(SyntaxNodeAnalysisContext context, IEnumerable<UsingDirectiveSyntax> usingDirectives, HashSet<INamespaceSymbol> necessaryNamespaces)
+        private static void CheckUnnecessaryUsings(SyntaxNodeAnalysisContext context, IEnumerable<UsingDirectiveSyntax> usingDirectives, IEnumerable<INamespaceSymbol> necessaryNamespaces)
```
costin-zaharia-sonarsource, line 80: `necessaryNamespaces` can be iterated (and evaluated) multiple times since it is inside a foreach loop. We should let this be a HashSet

costin-zaharia-sonarsource (changes requested):

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnnecessaryUsings.cs
```diff
@@ -65,32 +63,26 @@ protected override void Initialize(SonarAnalysisContext context)
...
                 SyntaxKind.CompilationUnit);
 
-        }
-
-        private static void VisitContent(CSharpRemovableUsingWalker visitor, SyntaxList<MemberDeclarationSyntax> members, IEnumerable<SyntaxTrivia> trivias)
+        private static void VisitContent(CSharpSyntaxWalker visitor, SyntaxList<MemberDeclarationSyntax> members, IEnumerable<SyntaxTrivia> trivias)
         {
-            var comments = trivias.Where(trivia => trivia.Kind() == SyntaxKind.SingleLineDocumentationCommentTrivia ||
-                trivia.Kind() == SyntaxKind.MultiLineDocumentationCommentTrivia);
+            var comments = trivias.Where(trivia => trivia.Kind() == SyntaxKind.SingleLineDocumentationCommentTrivia || trivia.Kind() == SyntaxKind.MultiLineDocumentationCommentTrivia);
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I know that you meant something little different, but this should do the trick as well in reusable way.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnnecessaryUsings.cs
pavel-mikula-sonarsource: `necessaryNamespaces` are a Symbols while `usedNamespace` are syntaxes. The `Any` is needed in this case.

We can improve it with some larger refactoring, but there's usually not more than 20 imports in a single class. I don't think the performance gain is worth it.

costin-zaharia-sonarsource: I did not asked to remove the `Any` call 😄

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/UnnecessaryUsings.cs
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Looks good

costin-zaharia-sonarsource (approved):
lgmt!

PR #3592, opened 2020-09-03
Fix S3216 FP: the rule is not relevant for .NET Core

### LINKED ISSUE #2588: Fix S3216 FP: the rule is not relevant for .NET Core
### Description

See [this community thread](https://community.sonarsource.com/t/exclude-s3216-from-net-core/13001)

If the app is a .NET Core app, this rule should not run.

### Related information

* SonarC# Version 7.16

pavel-mikula-sonarsource:
Fixes #2588


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TaskConfigureAwait.cs
```diff
@@ -33,39 +33,30 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class TaskConfigureAwait : SonarDiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S3216";
```
costin-zaharia-sonarsource, line 35: You can make the DiagnosticId private

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TaskConfigureAwait.cs
```diff
@@ -33,39 +33,30 @@ namespace SonarAnalyzer.Rules.CSharp
...
-                    var awaitExpression = (AwaitExpressionSyntax)c.Node;
-                    var expression = awaitExpression.Expression;
-                    if (expression == null)
-                    {
-                        return;
-                    }
-
-                    var type = c.SemanticModel.GetTypeInfo(expression).Type;
-                    if (type.DerivesFrom(KnownType.System_Threading_Tasks_Task))
+                    if ((c.Node as AwaitExpressionSyntax).Expression is { } expression
```
costin-zaharia-sonarsource, line 53: We should use direct cast here. It is always an `AwaitExpressionSyntax`. If it isn't we would prefer to have a cast exception and not a null reference one.

sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Helpers/CompilationExtensions.cs
```diff
@@ -29,5 +29,10 @@ public static class CompilationExtensions
             (IMethodSymbol)compilation.GetSpecialType(type)
                 .GetMembers(methodName)
                 .SingleOrDefault();
+
+        public static bool IsNetFrameworkTarget(this Compilation compilation) =>
+            // There's no direct way of checking current target framework yet (09/2020).
```
costin-zaharia-sonarsource, line 34: I think the message is a bit confusing. We can check the target framework, by using precompiler symbols, but we cannot check the runtime.

costin-zaharia-sonarsource (changes requested):

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -151,17 +151,12 @@ private static void DumpActualDiagnostics(string languageVersion, IEnumerable<Di
             bool verifyNoException = true,
             CancellationToken? cancellationToken = null)
         {
-            var compilationOptions = compilation.Language == LanguageNames.CSharp
```
pavel-mikula-sonarsource, line 154: Bad things:
* The input `compilation.Options` were completely ignored.
* All tests were actually build with default `OutputKind.ConsoleApplication` causing most of them containing build errors because of missing `Main` entry point.
* This `GetAllDiagnostics` did a workaround that hides those build errors for the next `VerifyBuildErrors` step

New solution:
I moved the orchestration (`DynamicallyLinkedLibrary` and `allowUnsafe`) directly to the `AddProject` so now it's default for everything we use.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TaskConfigureAwaitTest.cs
```diff
@@ -34,12 +37,29 @@ public class TaskConfigureAwaitTest
...
         [TestMethod]
         [TestCategory("Rule")]
         public void TaskConfigureAwait_NetCore() =>
             Verifier.VerifyAnalyzer(@"TestCases\TaskConfigureAwait.NetCore.cs", new TaskConfigureAwait());
+
 #endif
 
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void TaskConfigureAwait_ConsoleApp()
```
costin-zaharia-sonarsource, line 52: Why it's relevant for this analyzer to run as console app?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -59,6 +60,12 @@ private ProjectBuilder AddProject(AnalyzerLanguage language, string projectName,
 
             var project = Solution.AddProject(projectName, projectName, languageName);
 
+            var compilationOptions = project.CompilationOptions.WithOutputKind(OutputKind.DynamicallyLinkedLibrary);
+            if (languageName == LanguageNames.CSharp)
```
costin-zaharia-sonarsource, line 64: Shouldn't we handle also VB.Net?

pavel-mikula-sonarsource: _Joke:_ Because it's always safe to use VB.NET 😆 

There's not `unsafe` code in VB, the option doesn't exist.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TaskConfigureAwaitTest.cs
pavel-mikula-sonarsource: This rule is relevant only for `OutputKind.DynamicallyLinkedLibrary` and there's a condition in ` OutputKind.DynamicallyLinkedLibrary` on line 46 for this (sorry, I can't get the link for some reason). This test covers that condition by using console app. Any other `OutputKind` would do the trick as well.

costin-zaharia-sonarsource (approved):
LGTM!

AlbertoMonteiro:
I think that should be reverted, explanation: https://github.com/SonarSource/sonar-dotnet/issues/2588#issuecomment-691495026

PR #3594, opened 2020-09-03
Fix S2931 FP: Handle IAsyncDisposable as disposing interface

### LINKED ISSUE #3324: Fix S2931 FP: Handle IAsyncDisposable as disposing interface
### Description

When implementing "dispose" using `IAsyncDisposable` S2931 is complaining the class should implement `IDispose`.

### Repro steps

```cs
public class TestClass : IAsyncDisposable
{
    private CancellationTokenSource? cancellationTokenSource;

    public Task MethodAsync()
    {
        this.cancellationTokenSource = new CancellationTokenSource();
        return Task.Delay(1000, this.cancellationTokenSource.Token);
    }

    public void Stop()
    {
        this.cancellationTokenSource?.Cancel();
    }

    public ValueTask DisposeAsync()
    {
        this.cancellationTokenSource?.Dispose();
        return new ValueTask();
    }
}
```

### Expected behavior

S2931 should accept `IDisposable` and `IAsyncDisposable`.

### Actual behavior

S2931 only accepts `IDisposable`.

### Known workarounds

N/A

### Related information

* C# Plugins Version: 8.6.1.17183
* Visual Studio Version: 16.5.4
* MSBuild / dotnet version: dotnetcore 3.1

costin-zaharia-sonarsource:
Fixes #3324

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableMemberInNonDisposableClass.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableMemberInNonDisposableClass.cs
@@ -105,8 +105,8 @@ namespace SonarAnalyzer.Rules.CSharp
                          .OfType<IFieldSymbol>();
         }
 
-        private static bool IsOwnerSinceDeclaration(IFieldSymbol field) =>
-            field.DeclaringSyntaxReferences.SingleOrDefault()?.GetSyntax() is VariableDeclaratorSyntax varDeclarator
+        private static bool IsOwnerSinceDeclaration(ISymbol symbol) =>
+            symbol.DeclaringSyntaxReferences.SingleOrDefault()?.GetSyntax() is VariableDeclaratorSyntax varDeclarator
             && varDeclarator.Initializer?.Value is ObjectCreationExpressionSyntax;
     }
 }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableMemberInNonDisposableClass.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableMemberInNonDisposableClass.cs
@@ -69,7 +69,8 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static bool ShouldExclude(ITypeSymbol typeSymbol) =>
             !typeSymbol.IsClass()
-            || typeSymbol.Implements(KnownType.System_IDisposable);
+            || typeSymbol.Implements(KnownType.System_IDisposable)
+            || typeSymbol.Implements(KnownType.System_IAsyncDisposable);
 
         private static string GetMessage(INamespaceOrTypeSymbol namedType, SymbolAnalysisContext analysisContext)
         {
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableMemberInNonDisposableClass.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableMemberInNonDisposableClass.cs
@@ -53,7 +53,6 @@ namespace SonarAnalyzer.Rules.CSharp
                     }
 
                     var message = GetMessage(namedType, analysisContext);
-
                     if (string.IsNullOrEmpty(message))
                     {
                         return;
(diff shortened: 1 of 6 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableMemberInNonDisposableClass.cs
```diff
@@ -28,102 +27,86 @@
...
-            return false;
+
+            return method.GetBodyDescendantNodes()
+                         .OfType<AssignmentExpressionSyntax>()
+                         .Where(n => n.IsKind(SyntaxKind.SimpleAssignmentExpression) && n.Right is ObjectCreationExpressionSyntax)
+                         .Select(n => compilation.GetSemanticModel(method.SyntaxTree).GetSymbolInfo(n.Left).Symbol)
+                         .OfType<IFieldSymbol>();
         }
 
+        private static bool IsOwnerSinceDeclaration(ISymbol symbol) =>
```
andrei-epure-sonarsource, line 108: nitpick: IsOwnerSinceDeclaration should be above GetAssignmentsToFieldsIn because it's called first

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableMemberInNonDisposableClass.cs
```diff
@@ -28,102 +27,86 @@
...
-                    method.ExpressionBody().DescendantNodes() :
-                    method.Body.DescendantNodes();
-                return methodNodes.
-                    OfType<AssignmentExpressionSyntax>().
-                    Where(n => n.IsKind(SyntaxKind.SimpleAssignmentExpression) && n.Right is ObjectCreationExpressionSyntax).
-                    Select(n => semanticModel.GetSymbolInfo(n.Left).Symbol).
-                    OfType<IFieldSymbol>();
-            }
-            return empty;
+            var disposableFields = namedType.GetMembers()
```
andrei-epure-sonarsource, line 75: if `disposableFields` is empty, you can return early, right?

that would save some unnecessary `GetAssignmentsToFieldsIn` calls which use the semantic model

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableMemberInNonDisposableClass.cs
```diff
@@ -28,102 +27,86 @@
...
-                        OrderBy(s => s));
+                    var message = GetMessage(namedType, analysisContext);
+                    if (string.IsNullOrEmpty(message))
+                    {
+                        return;
+                    }
 
-                    if (!string.IsNullOrEmpty(message))
+                    var typeDeclarations = new CSharpRemovableDeclarationCollector(namedType, analysisContext.Compilation).TypeDeclarations;
+                    foreach (var classDeclaration in typeDeclarations)
```
andrei-epure-sonarsource, line 61: for my education: I am looking at this loop and cannot understand why there must be more type declarations. Do you know why? If yes, is there an existing test case for it? Where we raise in multiple locations?

(I know it's not related to this change but I am trying to find out if you can simplify this in this PR)

andrei-epure-sonarsource (changes requested):

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableMemberInNonDisposableClass.cs
costin-zaharia-sonarsource: I would avoid this kind of conventions. Code changes, often methods are called from multiple places and the order in which the methods are called should not define the order in which they are defined.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/DisposableMemberInNonDisposableClass.cs
costin-zaharia-sonarsource: > I am looking at this loop and cannot understand why there must be more type declarations. Do you know why?

Partial classes can have multiple type declarations since they can be defined more than once.

> If yes, is there an existing test case for it? Where we raise in multiple locations?

We don't have tests for this case. I think this should be handled in a separate PR.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DisposableMemberInNonDisposableClassTest.cs
```diff
@@ -30,11 +31,38 @@ public class DisposableMemberInNonDisposableClassTest
...
+        public void DisposableMemberInNonDisposableClass() =>
             Verifier.VerifyAnalyzer(@"TestCases\DisposableMemberInNonDisposableClass.cs",
-                new DisposableMemberInNonDisposableClass(),
-                ParseOptionsHelper.FromCSharp8);
+                                    new DisposableMemberInNonDisposableClass(),
+                                    ParseOptionsHelper.FromCSharp8);
+
+#if NETCOREAPP // IAsyncDisposable is available only on .Net Core
+        [TestMethod]
+        public void DisposableMemberInNonDisposableClass_IAsyncDisposable() =>
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Nice. Just found that we didn't search for fields which implement `IAsyncDisposable `

andrei-epure-sonarsource (approved):
LGTM

PR #3595, opened 2020-09-04
Fail fast for SonarScanner for MSBuild <= 3.0

### LINKED ISSUE #3593: Fail fast for SonarScanner for MSBuild <= 3.0
Support for SonarScanner for MSBuild 3.0 was removed in previous release 8.12

We should fail fast the analysis in case that the `roslyn.reportFilePaths` property is missing for all projects (SQ modules).

We cannot fail fast when `analyzer.projectOutPaths` is missing because we intentionally don't generate the property for test projects. And solution containing only test projects would fail to analyze.

pavel-mikula-sonarsource:
Fixes #3593

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
@@ -93,8 +93,8 @@ public class DotNetSensor implements ProjectSensor {
       protobufDataImporter.importResults(context, protobufPaths, toRealPath);
     }
 
-    List<RoslynReport> roslynDirs = reportPathCollector.roslynDirs();
-    if (roslynDirs.isEmpty()) {
+    List<RoslynReport> roslynReports = reportPathCollector.roslynReports();
+    if (roslynReports.isEmpty()) {
       throw new IllegalStateException("No Roslyn issues report were found.");
     }
     Map<String, List<RuleKey>> activeRoslynRulesByPartialRepoKey = activeRoslynRulesByPartialRepoKey(pluginMetadata, context.activeRules()
@@ -102,6 +102,6 @@ public class DotNetSensor implements ProjectSensor {
       .stream()
       .map(ActiveRule::ruleKey)
       .collect(toList()));
-    roslynDataImporter.importRoslynReports(roslynDirs, context, activeRoslynRulesByPartialRepoKey, toRealPath);
+    roslynDataImporter.importRoslynReports(roslynReports, context, activeRoslynRulesByPartialRepoKey, toRealPath);
   }
 }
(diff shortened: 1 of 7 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
```diff
@@ -94,13 +94,14 @@ private void executeInternal(SensorContext context) {
...
 
     List<RoslynReport> roslynDirs = reportPathCollector.roslynDirs();
-    if (!roslynDirs.isEmpty()) {
-      Map<String, List<RuleKey>> activeRoslynRulesByPartialRepoKey = activeRoslynRulesByPartialRepoKey(pluginMetadata, context.activeRules()
-        .findAll()
-        .stream()
-        .map(ActiveRule::ruleKey)
-        .collect(toList()));
-      roslynDataImporter.importRoslynReports(roslynDirs, context, activeRoslynRulesByPartialRepoKey, toRealPath);
+    if (roslynDirs.isEmpty()) {
```
andrei-epure-sonarsource: is it possible that a project without any issues won't end up with any roslynDirs? or are these created during the BEGIN step by the scanner?

andrei-epure-sonarsource: in `org/sonarsource/dotnet/shared/plugins/AbstractModuleConfiguration.java` I see

```

  /**
   * Returns path of the directory containing all protobuf files, if:
   * - Property with it's location is defined
   * - The directory exists
   * - The directory contains at least one protobuf
   */
  public List<Path> protobufReportPaths() {
```

andrei-epure-sonarsource: it can be easily tested with an IT that has no issues

andrei-epure-sonarsource (changes requested):
would it be worth to add an IT with an "clean code project" with no issues?

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
pavel-mikula-sonarsource: Generation of the file is configured by `SonarQube.Integration.targets`. It's not configured when `SonarQubeExclude` is present. So in case all projects will have `SonarQubeExclude` we'll fail the analysis. But that kinda makes sense.

Project with no issues will still generate file with empty `results`, I've verified that: 
```
{
  "$schema": "http://json.schemastore.org/sarif-1.0.0",
  "version": "1.0.0",
  "runs": [
    {
      "tool": {
        "name": "Microsoft (R) Visual Basic Compiler",
        "version": "3.5.0.0",
        "fileVersion": "3.5.0-beta4-20153-05 (20b9af91)",
        "semanticVersion": "3.5.0",
        "language": "en-US"
      },
      "results": [
      ]
    }
  ]
}
```

pavel-mikula-sonarsource: This PR is not related to `protobufReportPaths`, see issue description #3593

pavel-mikula-sonarsource:
`NoSonarTest` seems that it doesn't contain any issues and would fail in that case.

pavel-mikula-sonarsource:
Similar for `DoNotAnalyzeTestFilesTest`

andrei-epure-sonarsource (approved):
lgtm

PR #3597, opened 2020-09-07
Update rspec

### LINKED ISSUE #3559: Update RSPEC for 8.13
Pay attention to include the resx files as well.
Check if the rule updates imply a rule change which hasn't been done yet (e.g. #3450 and #3330).

costin-zaharia-sonarsource:
Fixes #3559

### DIFF
--- a/sonaranalyzer-dotnet/rspec/cs/S5042_c#.html
+++ b/sonaranalyzer-dotnet/rspec/cs/S5042_c#.html
@@ -1,53 +1,87 @@
-<p>Expanding archive files is security-sensitive. For example, expanding archive files has led in the past to the following vulnerabilities:</p>
-<ul>
-  <li> <a href="http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-1263">CVE-2018-1263</a> </li>
-  <li> <a href="http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-16131">CVE-2018-16131</a> </li>
-</ul>
-<p>Applications that expand archive files (zip, tar, jar, war, 7z, ...) should verify the path where the archive's files are expanded and not trust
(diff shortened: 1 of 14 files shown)


sonaranalyzer-dotnet/rspec/cs/S5042_c#.html
```diff
@@ -1,53 +1,87 @@
...
+FileStream zipToOpen = new FileStream(@"ZipBomb.zip", FileMode.Open);
+ZipArchive archive = new ZipArchive(zipToOpen, ZipArchiveMode.Read);
+foreach (ZipArchiveEntry entry in archive.Entries)
+{
+  entry.ExtractToFile("./output_onlyfortesting.txt", true); // Sensitive
+}
+</pre>
+<h2>Compliant Solution</h2>
+<pre>
+int THRESHOLD_ENTRIES = 10000;
```
andrei-epure-sonarsource, line 33: RE: https://jira.sonarsource.com/browse/RSPEC-5044
@eric-therond-sonarsource - should we stop raising the hotspot when seeing a compliant example? In other words, do the S5042 changes require an issue to be created?

andrei-epure-sonarsource: Eric is on vacation. @costin-zaharia-sonarsource , I guess we can move forward with this one, allow the change and eventually open an issue to update the rule if Eric says so when back from vacation.

andrei-epure-sonarsource (approved):

sonaranalyzer-dotnet/rspec/cs/S5042_c#.html
costin-zaharia-sonarsource: Just quickly checked the code and as far as I can tell we consider both `ExtractToFile(..)` and `Open()` unsafe. This means that we will raise issues also for the compliant example -> code changes are needed before updating the docs.

I will revert the changes related to this.

andrei-epure-sonarsource (approved):

PR #3601, opened 2020-09-08
Analyze projects with no sources

### LINKED ISSUE #3600: Fix the S4MSB property detection error that surfaced in security
See this[ cirrus ci failed job ](https://cirrus-ci.com/task/5701631268356096?command=qa_csharp_msbuild16#L2850)

```
INFO: ------------- Run sensors on project
INFO: Sensor C# [csharp]
...
ERROR: No Roslyn issue reports were found.
```

After this gets merged, we need to use again `DEV` version in the security ITs instead of `LATEST_RELEASE` and test all security ITs work as expected.

Only after then we can release.

andrei-epure-sonarsource:
Fixes #3600


its/src/test/java/com/sonar/it/csharp/NoSourcesTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+    ORCHESTRATOR.executeBuild(beginStep);
+
+    TestUtils.runMSBuild(ORCHESTRATOR, projectDir, "/t:Rebuild");
+
+    ORCHESTRATOR.executeBuild(TestUtils.createEndStep(projectDir));
+  }
+
+  @Test
+  public void filesAtProjectLevel() {
```
andrei-epure-sonarsource, line 41: I cherry-picked this on `master` and it fails with the `No Roslyn issue reports were found.`

its/src/test/java/com/sonar/it/csharp/NoSourcesTest.java
```diff
@@ -0,0 +1,50 @@
+package com.sonar.it.csharp;
```
pavel-mikula-sonarsource, line 1: Java ITs failed on missing licence header in this file

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractModuleConfiguration.java
```diff
@@ -58,10 +58,14 @@
 
   private final Configuration configuration;
   private final String languageKey;
+  // we store the project key for logging
```
pavel-mikula-sonarsource, line 61: I think we can remove this comment. It's clearly visible from usages.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractModuleConfiguration.java
```diff
@@ -81,51 +85,51 @@ static String getAnalyzerReportDir(String languageKey) {
...
+      return Arrays.stream(strPaths)
+        .map(Paths::get)
+        .collect(Collectors.toList());
+    } else {
+      LOG.warn( "For project '{}', no Roslyn issues reports have been found.", projectKey);
+      return Collections.emptyList();
+    }
+  }
+
+  private static boolean validateOutputDir(Path analyzerOutputDir, String projectKey) {
```
pavel-mikula-sonarsource, line 109: The passing of class state into static method seems painful and non-natural. Can you remove the `static` instead to simplify the call site to original form and the state management?

```suggestion
  private boolean validateOutputDir(Path analyzerOutputDir) {
```

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractModuleConfiguration.java
```diff
@@ -81,51 +85,51 @@ static String getAnalyzerReportDir(String languageKey) {
...
     String path = analyzerOutputDir.toString();
     try {
       if (!analyzerOutputDir.toFile().exists()) {
-        LOG.warn("Analyzer working directory does not exist: '{}'. {}", path, MSG_SUFFIX);
+        LOG.warn("For project '{}', analyzer working directory does not exist: '{}'. {}", projectKey, path, MSG_SUFFIX);
```
pavel-mikula-sonarsource, line 113: I like the idea of having unified prefix in all messages. To avoid fight with English language, I'd use this technical form to avoid the problem:
```suggestion
        LOG.warn("Project '{}': Analyzer working directory does not exist '{}'. {}", projectKey, path, MSG_SUFFIX);
```

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/AbstractModuleConfigurationTest.java
```diff
@@ -67,7 +67,7 @@ public void giveWarningsWhenGettingProtobufPathAndNoPropertyAvailable() {
 
     AbstractModuleConfiguration config = createAbstractModuleConfiguration(configuration);
     assertThat(config.protobufReportPaths()).isEmpty();
-    assertThat(logTester.logs(LoggerLevel.WARN)).containsOnly("Property missing: 'sonar.cs.analyzer.projectOutPaths'. No protobuf files will be loaded for this project.");
+    assertThat(logTester.logs(LoggerLevel.WARN)).containsOnly("Property missing: 'sonar.cs.analyzer.projectOutPaths'. No protobuf files will be loaded for project '<NONE>'.");
```
pavel-mikula-sonarsource, line 70: Can you change mocking of this test to use some project key to see that the value is really used? The `<NONE>` is tested by the test below.

its/src/test/java/com/sonar/it/csharp/NoSourcesTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import static com.sonar.it.csharp.Tests.ORCHESTRATOR;
+import static com.sonar.it.csharp.Tests.getMeasureAsInt;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class NoSourcesTest {
```
pavel-mikula-sonarsource, line 16: This should be registered in `its/src/test/java/com/sonar/it/csharp/Tests.java` AFAIK

its/src/test/java/com/sonar/it/csharp/NoSourcesTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+  private static final String PROJECT = "ProjectWithNoSources";
+  @ClassRule
+  public static final Orchestrator orchestrator = Tests.ORCHESTRATOR;
+
+  @BeforeClass
+  public static void init() throws Exception {
+    TestUtils.reset(orchestrator);
+
+    Path projectDir = Tests.projectDir(temp, "ProjectWithNoSources");
```
pavel-mikula-sonarsource, line 29: The constant `PROJECT` should be used here and below

its/src/test/java/com/sonar/it/csharp/NoSourcesTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.junit.rules.TemporaryFolder;
+
+import static com.sonar.it.csharp.Tests.ORCHESTRATOR;
+import static com.sonar.it.csharp.Tests.getMeasureAsInt;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class NoSourcesTest {
+
+  @ClassRule
+  public static TemporaryFolder temp = TestUtils.createTempFolder();
```
pavel-mikula-sonarsource, line 19: _CodeSmell:_
```suggestion
  public static final TemporaryFolder temp = TestUtils.createTempFolder();
```

its/src/test/java/com/sonar/it/csharp/NoSourcesTest.java
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  @Test
+  public void filesAtProjectLevel() {
+    assertThat(getProjectMeasureAsInt("violations")).isEqualTo(4);
+  }
+
+  /* Helper methods */
+
+  private Integer getProjectMeasureAsInt(String metricKey) {
```
pavel-mikula-sonarsource, line 47: This is copy/pasted all over the place 7 times. Can you refactor it to some `TestHelper` class?

its/src/test/java/com/sonar/it/csharp/NoSourcesTest.java
pavel-mikula-sonarsource, line 29: This initialization seems to be copy/pasted several times, can it be factored out to some `TestHelper` class?
To clarify: Only for simple cases with single `projectDir` argument.

its/src/test/java/com/sonar/it/csharp/NoSourcesTest.java
```diff
@@ -0,0 +1,50 @@
...
+import static com.sonar.it.csharp.Tests.ORCHESTRATOR;
+import static com.sonar.it.csharp.Tests.getMeasureAsInt;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class NoSourcesTest {
+
+  @ClassRule
+  public static TemporaryFolder temp = TestUtils.createTempFolder();
+
+  private static final String PROJECT = "ProjectWithNoSources";
```
pavel-mikula-sonarsource, line 21: I'd prefer to have those two `@ClassRule` together and this constant above/below them. Current order and empty lines seems irregular.

pavel-mikula-sonarsource (changes requested):

its/src/test/java/com/sonar/it/csharp/NoSourcesTest.java
andrei-epure-sonarsource: Indeed,Ionly ranit in theide

its/src/test/java/com/sonar/it/csharp/NoSourcesTest.java
andrei-epure-sonarsource: `getMeasureAsInt` is already a helper. The `getProjectMeasureAsInt` is copy/pasted to avoid having `getMeasureAsInt(PROJECT, metricKey);` all over the tests and simply call `getProjectMeasureAsInt("lines")`

however in such a simple test it's an overkill, so I will call directly `getMeasureAsInt`

its/src/test/java/com/sonar/it/csharp/NoSourcesTest.java
andrei-epure-sonarsource: well, the `temp` needs to be passed from here as it's `ClassRule` annotated; and the projectDir needs to be passed... 

it's already in the `TestHelper`. 

Do you mean initializing _and_ doing the analysis? I think the way it is now makes the test understandable, each step

pavel-mikula-sonarsource: It still makes sense to factor it to a separate method with 3 args. Because it can be reused several times all over the place.
The test itself is readable, but the big picture copy/paste/duplication is not a great pattern to maintain.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractModuleConfiguration.java
```diff
@@ -85,43 +84,43 @@ static String getAnalyzerReportDir(String languageKey) {
...
     return analyzerWorkDirPaths.stream().map(x -> x.resolve(getAnalyzerReportDir(languageKey)))
-      .filter(dir -> validateOutputDir(dir, projectKey))
+      .filter(this::validateOutputDir)
       .collect(Collectors.toList());
   }
 
   public List<Path> roslynReportPaths() {
     String[] strPaths = configuration.getStringArray(getRoslynJsonReportPathProperty(languageKey));
-    LOG.debug("For project '{}', the Roslyn JSON report path has '{}'", projectKey, String.join(",", strPaths));
+    LOG.debug("Project '{}': The Roslyn JSON report path has '{}'", projectKey, String.join(",", strPaths));
```
andrei-epure-sonarsource, line 97: I want to add more UTs for the comments, but it will be a small update

its/src/test/java/com/sonar/it/csharp/MetricsTest.java
```diff
@@ -292,18 +280,10 @@ public void executableLines() {
 
   /* Helper methods */
 
-  private Measure getProjectMeasure(String metricKey) {
```
andrei-epure-sonarsource, line 295: these weren't used

its/src/test/java/com/sonar/it/csharp/DoNotAnalyzeTestFilesTest.java
```diff
@@ -37,31 +36,30 @@
...
   @Before
   public void init() {
-    TestUtils.reset(orchestrator);
+    TestUtils.reset(ORCHESTRATOR);
   }
 
   @Test
   public void should_not_increment_test() throws Exception {
-    Path projectDir = Tests.projectDir(temp, "DoNotAnalyzeTestFilesTest");
+    Path projectDir = Tests.projectDir(temp, PROJECT);
```
pavel-mikula-sonarsource, line 48: Nice cleanup 👍

its/src/test/java/com/sonar/it/csharp/MetricsTest.java
```diff
@@ -59,26 +59,14 @@ private static RuleChain getRuleChain() {
         @Override
         protected void before() throws Throwable {
           TestUtils.reset(ORCHESTRATOR);
-
-          Path projectDir = Tests.projectDir(temp, "MetricsTest");
-
-          ScannerForMSBuild beginStep = TestUtils.createBeginStep("MetricsTest", projectDir)
-            .setProfile("no_rule")
-            // Without that, the MetricsTest project is considered as a Test project :)
```
pavel-mikula-sonarsource, line 67: Can you preserve this comment? I find it useful. Similar one is in `NoSonarTest.java`

its/src/test/java/com/sonar/it/csharp/SharedFilesTest.java
```diff
@@ -40,26 +37,14 @@
...
 
-  @ClassRule
-  public static final Orchestrator orchestrator = Tests.ORCHESTRATOR;
-
   @Before
-  public void init() {
-    TestUtils.reset(orchestrator);
+  public void init(){
+    TestUtils.reset(ORCHESTRATOR);
   }
```
pavel-mikula-sonarsource, line 44: _Nitpick:_ 44 and 57 has some whitespace

its/src/test/java/com/sonar/it/csharp/Tests.java
```diff
@@ -83,6 +87,20 @@ public static Path projectDir(TemporaryFolder temp, String projectName) throws I
     return tmpProjectDir;
   }
 
+  static BuildResult analyzeProject(TemporaryFolder temp, String projectName, @Nullable String profileKey, String... keyValues) throws IOException {
```
pavel-mikula-sonarsource, line 90: _Optionally:_ Can you add second method with almost the same content with extra subProjectName argument and `TestUtils.createBeginStep(projectName, projectDir, subProjectName)`? That should simplify almost all of the rest of the invocations.

its/src/test/java/com/sonar/it/vbnet/MetricsTest.java
```diff
@@ -55,26 +56,14 @@ private static RuleChain getRuleChain() {
         @Override
         protected void before() throws Throwable {
           TestUtils.reset(ORCHESTRATOR);
-
-          Path projectDir = Tests.projectDir(temp, "VbMetricsTest");
-
-          ScannerForMSBuild beginStep = TestUtils.createBeginStep("VbMetricsTest", projectDir)
-            .setProfile("vbnet_no_rule")
-            // Without that, the MetricsTest project is considered as a Test project :)
```
pavel-mikula-sonarsource, line 63: Can this comment be preserved? I find it useful.

its/src/test/java/com/sonar/it/vbnet/NoSonarTest.java
```diff
@@ -36,37 +36,21 @@
...
 
   @BeforeClass
   public static void init() throws Exception {
     TestUtils.reset(orchestrator);
-
-    Path projectDir = Tests.projectDir(temp, "VbNoSonarTest");
-
-    ScannerForMSBuild beginStep = TestUtils.createBeginStep("VbNoSonarTest", projectDir)
-      .setProfile("vbnet_class_name")
-      // Without that, the NoSonarTest project is considered as a Test project :)
```
pavel-mikula-sonarsource, line 51: Same here

pavel-mikula-sonarsource (approved):
LGTM! Nice work and cleanup

its/src/test/java/com/sonar/it/csharp/NoSourcesTest.java
pavel-mikula-sonarsource: Awesome!

andrei-epure-sonarsource:
moving this to draft as we have warnings on non-C# builds (e.g. Java build)
from analyzing sonar-java

```
[INFO] Sensor C# Properties [csharp]
[WARNING] Project 'org.sonarsource.java:it-java-semantic': Property missing: 'sonar.cs.analyzer.projectOutPaths'. No protobuf files will be loaded for this project.
[WARNING] Project 'org.sonarsource.java:it-java-semantic': No Roslyn issues reports have been found.
[INFO] Sensor C# Properties [csharp] (done) | time=28ms
```

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
```diff
@@ -91,6 +91,8 @@ private void executeInternal(SensorContext context) {
     List<Path> protobufPaths = reportPathCollector.protobufDirs();
     if (!protobufPaths.isEmpty()) {
       protobufDataImporter.importResults(context, protobufPaths, toRealPath);
+    } else {
+      LOG.warn("No protobuf reports found - no metrics and highlighting will be imported.");
```
andrei-epure-sonarsource, line 95: @pavel-mikula-sonarsource  given that we fail if no roslyn reports are found... shouldn't we also (or even the more) fail if no highlighting can be made?

andrei-epure-sonarsource:
Another scenario to manually test before the release of 8,13: what if a CS file is referenced for some reason in the sources of a maven build?

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
pavel-mikula-sonarsource: I don't think so, I've chosen only the Roslyn property to fail fast because of this comment:
https://github.com/SonarSource/sonar-dotnet/blob/ee3a8953dacbde39b0ff18c86d04cb95724cd8a5/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractModuleConfiguration.java#L82

Having "SQ Project" containing only .NET test project and maybe some other non-.NET stuff you'd like to analyze is (weird but) legal setup with Roslyn in place and should be analyzed. This might change once `Tests as 1st class citizen` is done and we modify the pb generation and that comment.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
```diff
@@ -91,6 +91,8 @@ private void executeInternal(SensorContext context) {
     List<Path> protobufPaths = reportPathCollector.protobufDirs();
     if (!protobufPaths.isEmpty()) {
```
pavel-mikula-sonarsource, line 92: Can you please swap this condition to have a positive test?
```suggestion
    if (protobufPaths.isEmpty()) {
```

pavel-mikula-sonarsource (approved):
Still LGTM with one minor suggestion

Issue #3611, opened 2020-09-16
Add support for VarPattern inside switch expressions


costin-zaharia-sonarsource:
### Description

`VarPattern` inside switch expressions is not supported.

### Repro steps

Build the CFG for the following snippet: `string a = taintedString switch {var x => null};`

A `NotSupportedException` is thrown for the var pattern.

### Related information

* C#/VB.NET Plugins Version: 8.13

mary-georgiou-sonarsource:
SE engine has been rewritten and varpattern is supported there.
We won't support the legacy engine.

PR #3612, opened 2020-09-16
Add repro for #3611

costin-zaharia-sonarsource:
Reproduces #3611


pavel-mikula-sonarsource (approved):
LGTM, let's merge after version bump to 8.14

PR #3619, opened 2020-09-17
Specify scanner compatibility in embedded docs

andrei-epure-sonarsource:
We've dropped support for Scanner for MSBuild 3.x in the 8.12 release.

Also, some users wonder why C# analysis doesn't work with the `scanner-cli`, so it doesn't hurt to explicitly mention to use the S4MSB.


MikeBirnstiehl (approved):
LGTM!

PR #3635, opened 2020-09-28
S2699: Detect custom method with specific attribute as assertion

### LINKED ISSUE #3634: S2699: Detect custom method with specific attribute as assertion
### Description

Rule 2699 should recognize domain specific assertion methods as assertion methods as well.

In my particular case, my Blazor testing library [bUnit](https://bunit.egilhansen.com/) has a semantic HTML comparer exposed as the as the assertion method [`MarkupMatches(...)`](https://bunit.egilhansen.com/docs/verification/verify-markup.html#the-markupmatches-method) which is not recognized as a assertion method by SonarLint.

A simple suggestion would be to add the word `MATCHES` to the assertion method list. But that is a quick-fix for this particular case, and will likely exclude other domain specific assertion methods in the future, so perhaps an attribute is a safer approach, for those that want to be compatible with SonarLint?

### Repro steps

Simple example from the documentation link above:

```
// arrange
using var ctx = new TestContext();

// act
var cut = ctx.RenderComponent<Heading>();

// assert
cut.MarkupMatches(@"<h3 id=""heading-1337"" required>
                      Heading text
                      <small class=""mark text-muted"">Secondary text</small>
                    </h3>");
```

### Expected behavior

`MarkupMatches` to be recognized as an assertion method

### Actual behavior

`MarkupMatches` is not recognized as an assertion method.

### Known workarounds

Create another extension method that uses one of the keywords that is recognized as assertion statements.

### Related information

Discussed here: https://community.sonarsource.com/t/how-to-mark-custom-methods-as-assertion-methods-in-c/31437

Not sure about these, maybe @domn1995 can help out, since he reported the problem originally: https://github.com/egil/bUnit/discussions/216

* C#/VB.NET Plugins Version
* Visual Studio Version
* MSBuild / dotnet version
* Operating System

pavel-mikula-sonarsource:
Fixes #3634

Approach similar to S2259 / C#: [RSPEC-3634](https://jira.sonarsource.com/browse/RSPEC-3634)

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
@@ -27,6 +27,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.ShimLayer.CSharp;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
@@ -58,7 +59,7 @@ namespace SonarAnalyzer.Rules.CSharp
                 {
                     var methodDeclaration = (MethodDeclarationSyntax)c.Node;
                     if (methodDeclaration.Identifier.IsMissing ||
-                        methodDeclaration.Body == null)
+                        (methodDeclaration.Body == null && methodDeclaration.ExpressionBody == null))
                     {
                         return;
                     }
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
@@ -27,7 +27,6 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
-using SonarAnalyzer.ShimLayer.CSharp;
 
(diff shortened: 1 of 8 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
```diff
@@ -36,6 +36,8 @@ public sealed class TestMethodShouldContainAssertion : SonarDiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S2699";
         private const string MessageFormat = "Add at least one assertion to this test case.";
+        private const string CustomAssertionAttributeName = "AssertionMethodAttribute";
+
```
costin-zaharia-sonarsource, line 40: pedantic: there is an extra empty line

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
```diff
@@ -36,6 +36,8 @@ public sealed class TestMethodShouldContainAssertion : SonarDiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S2699";
```
costin-zaharia-sonarsource, line 37: optional (clean as you code): 
- DiagnosticId can be private
- IsTestIgnored can be private

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/TestMethodShouldContainAssertionTest.cs
```diff
@@ -193,5 +193,14 @@ public void ShouldNotThrowTest()
...
                     .Concat(NuGetMetadataReference.FluentAssertions(fluentVersion))
                     .Concat(MetadataReferenceFacade.GetSystemThreadingTasks()));
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void TestMethodShouldContainAssertion_CustomAssertionMethod() =>
+            Verifier.VerifyAnalyzer(@"TestCases\TestMethodShouldContainAssertion.Custom.cs",
+                new TestMethodShouldContainAssertion(),
+                additionalReferences: NuGetMetadataReference.MSTestTestFramework(Constants.NuGetLatestVersion));
+
```
costin-zaharia-sonarsource, line 203: pedantic: empty lines can be removed

costin-zaharia-sonarsource (approved):

Issue #3639, opened 2020-10-02
Fix S3971 FP: Support IAsyncDisposable


costin-zaharia-sonarsource:
### Description

Rule [S3971](https://rules.sonarsource.com/csharp/RSPEC-3971) raises a false positive when `GC.SuppressFinalize` is called from [IAsyncDisposable.DisposeAsync](https://docs.microsoft.com/en-us/dotnet/api/system.iasyncdisposable.disposeasync?view=dotnet-plat-ext-3.1#System_IAsyncDisposable_DisposeAsync). This contradicts the best practices described in the API doc https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#the-disposeasync-method

Source: https://community.sonarsource.com/t/false-positive-do-not-call-suppressfinalize/31740

### Repro steps

```
public class Custom : IAsyncDisposable
{
    public async ValueTask DisposeAsync()
    {
        await DisposeAsyncCore();

        GC.SuppressFinalize(this); // Noncompliant - FP, see: https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#the-disposeasync-method
    }

    protected virtual ValueTask DisposeAsyncCore() => default;
}
```

### Expected behavior

S3971 should not raise an issue for this case.

### Actual behavior

False positive is reported.

### Known workarounds

Implement `IDisposable` instead.

### Related information

* C#/VB.NET Plugins Version: 8.13.1.21947

dnnr:
Is it possible that there's a regression? I'm on v2025.3.1, and I'm getting S3971 complaints for calling SuppressFinalize in DisposeAsync.

denis-troller:
> Is it possible that there's a regression? I'm on v2025.3.1, and I'm getting [S3971](https://sonarsource.github.io/rspec/#/rspec/S3971/csharp) complaints for calling SuppressFinalize in DisposeAsync.

Would you mind sharing a reproducer?

pavel-mikula-sonarsource:
Hi, please open a new issue with the reproducer. Feel free to mention is issue as a related one / possible regression.

dnnr:
Thanks, new issue opened. Just for the record: It doesn't look like a regression but it happens when using an intermediate interface.

PR #3640, opened 2020-10-02
Add repro for #3639

costin-zaharia-sonarsource:
Reproduces #3639

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCallGCSuppressFinalize.NetCore.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCallGCSuppressFinalize.NetCore.cs
@@ -1,0 +1,17 @@
+using System;
+using System.Threading.Tasks;
+
+namespace Tests.Diagnostics
+{
+    public class Custom : IAsyncDisposable
+    {
+        public async ValueTask DisposeAsync()
+        {
+            await DisposeAsyncCore();
+
+            GC.SuppressFinalize(this); // Noncompliant - FP, see: https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#the-disposeasync-method
+        }
+
+        protected virtual ValueTask DisposeAsyncCore() => default;
+    }
+}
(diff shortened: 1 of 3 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCallGCSuppressFinalize.NetCore.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Threading.Tasks;
+
+namespace Tests.Diagnostics
+{
+    public class Custom : IAsyncDisposable
+    {
+        public async ValueTask DisposeAsync()
```
pavel-mikula-sonarsource, line 8: Can you add two more cases with True Positives that will invoke `GC.SuppressFinalize(this)`?
* Method `Something` in this class
* Another class that does NOT implement `IAsyncDisposable` but has `public void DisposeAsync()` with the same name

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCallGCSuppressFinalize.NetCore.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Threading.Tasks;
+
+namespace Tests.Diagnostics
+{
+    public class Custom : IAsyncDisposable
```
pavel-mikula-sonarsource, line 6: We usually add a link to the original issue with reproducers for reference
```suggestion
	// https://github.com/SonarSource/sonar-dotnet/issues/3639
    public class Custom : IAsyncDisposable
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCallGCSuppressFinalize.NetCore.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        protected virtual ValueTask DisposeAsyncCore() => default;
+    }
+
+    public class Fake
+    {
+        public async ValueTask DisposeAsync()
+        {
+            await DisposeAsyncCore();
```
pavel-mikula-sonarsource, line 28: This line causes compilation error for the snippet. Can it be removed? I don't think it's needed for purpose of this test case. 
```suggestion
```

costin-zaharia-sonarsource: It is there to avoid warnings for the `async` keyword but I forgot to add `DisposeAsyncCore` method

pavel-mikula-sonarsource (approved):
LGTM

PR #3643, opened 2020-10-06
Replace "reset" API call with "bulk_delete"

### LINKED ISSUE #3638: Remove 'api/orchestrator/reset' usage from integration tests
From the Sonarqube team:

>You should rely on regular web services to clean up data as needed. For example, if you need to remove a project and all associated data, call `api/projects/delete?project=myproject`.

They will cleanup all usages of orchestrator APIs.

costin-zaharia-sonarsource:
- "reset" is deprecated and needs to be removed

Fixes #3638


its/src/test/java/com/sonar/it/shared/TestUtils.java
```diff
@@ -181,14 +187,22 @@ public static String replaceLtsVersion(String version) {
...
+
+    LOG.info("TEST SETUP: deleting projects analyzed before: " + currentDateTime);
+
+    orchestrator.getServer()
+      .newHttpCall("/api/projects/bulk_delete")
       .setAdminCredentials()
+      .setMethod(HttpMethod.POST)
+      .setParams("analyzedBefore", currentDateTime)
       .execute();
-    // api/orchestrator/reset will clear the license, so reinstall it
```
costin-zaharia-sonarsource, line 188: This is no longer needed since instead of resetting the orchestrator we just delete the analyzed projects.

its/src/test/java/com/sonar/it/shared/TestUtils.java
```diff
@@ -181,14 +187,22 @@ public static String replaceLtsVersion(String version) {
...
+      .withZone( ZoneId.of("UTC"))
+      .format(instant);
+
+    LOG.info("TEST SETUP: deleting projects analyzed before: " + currentDateTime);
+
+    orchestrator.getServer()
+      .newHttpCall("/api/projects/bulk_delete")
       .setAdminCredentials()
+      .setMethod(HttpMethod.POST)
+      .setParams("analyzedBefore", currentDateTime)
```
pavel-mikula-sonarsource, line 204: Java DateTime API is a daytime nightmare. Can you replace it with this and remove all the orchestration and imports?
```suggestion
      .setParams("qualifiers", "TRK")
```

its/src/test/java/com/sonar/it/shared/TestUtils.java
```diff
@@ -181,14 +187,22 @@ public static String replaceLtsVersion(String version) {
...
       .execute();
-    // api/orchestrator/reset will clear the license, so reinstall it
-    if (orchestrator.getDistribution().isActivateLicense()) {
-      orchestrator.activateLicense();
-    }
   }
 
   static WsClient newWsClient(Orchestrator orch) {
```
pavel-mikula-sonarsource, line 208: Empty line before `createTempFolder`: There're some spaces that can be cleaned as we code.

pavel-mikula-sonarsource (changes requested):
It seems that we can simplify it a lot.

its/src/test/java/com/sonar/it/shared/TestUtils.java
costin-zaharia-sonarsource: Currently the API documentation says: `At least one parameter is required among analyzedBefore, projects and q` and this is a bit misleading.

costin-zaharia-sonarsource: Are you sure about this? The API returns now 400. See: https://sonarsource.visualstudio.com/DotNetTeam%20Project/_build/results?buildId=22472&view=logs&j=5a0b2a7a-818d-5a4c-a639-e2568d087f6b&t=407cbd3b-4ba1-5b9c-259b-e19d650c77d2

pavel-mikula-sonarsource: No. There's not-so-well-named method `checkAtLeastOneParameterIsPresent` that validates the request on SQ side :(

its/src/test/java/com/sonar/it/shared/TestUtils.java
```diff
@@ -181,22 +185,30 @@ public static String replaceLtsVersion(String version) {
...
-      .setMethod(HttpMethod.POST)
+    // We add one day to ensure that today's entries are deleted.
+    Instant instant = Instant.now().plus(1, ChronoUnit.DAYS);
+
+    // The expected format is yyyy-MM-dd.
+    String currentDateTime = DateTimeFormatter.ISO_LOCAL_DATE
+      .withZone(ZoneId.of("UTC"))
+      .format(instant);
+
+    LOG.info("TEST SETUP: deleting projects analyzed before: " + currentDateTime);
```
pavel-mikula-sonarsource, line 196: Do we need to log this? I'd remove the `currentDateTime` from the log, because it's in UTC. It runs against local SQ instance so there's no harm done.
I would just log that the all projects are deleted from local SQ (as it was in one of the intermediate commits)

pavel-mikula-sonarsource (approved):
LGTM with one open question

its/src/test/java/com/sonar/it/shared/TestUtils.java
costin-zaharia-sonarsource: I think this helps with debugging since it shows the used format. The API is very restrictive in this regard and displaying the value can help in case of problems.

mpaladin:
FYI it would have been enough to change project keys to unique keys for each test rather than using `bulk_delete`.

PR #3647, opened 2020-10-07
S3966: Remove FP on streams special handling

### LINKED ISSUE #3644: Rule S3966: FP when using StreamWriter
### Description

`StreamWriter` is closing and not disposing the underling stream. See: https://github.com/microsoft/referencesource/blob/master/mscorlib/system/io/streamwriter.cs#L269

Rule description needs to be updated too since it's based on this FP.

Source: https://github.com/SonarSource/sonar-dotnet/issues/209#issuecomment-703929235

### Repro steps

```
using (Stream stream = new FileStream("file.txt", FileMode.OpenOrCreate))
{
    using (StreamWriter writer = new StreamWriter(stream))  // Noncompliant: 'stream' will be disposed twice
    {
        // Use the writer object...
    }
}
```

### Expected behavior

There should be no diagnostic raised for `stream` in this case since it is not disposed twice.

### Actual behavior

A FP is raised reporting that the `stream` was disposed twice.

### Known workarounds

None.

### Related information

* C#/VB.NET Plugins Version: https://github.com/SonarSource/sonar-dotnet/releases/tag/8.13.1.21947

costin-zaharia-sonarsource:
Fixes #3644


sonaranalyzer-dotnet/rspec/cs/S3966_c#.html
```diff
@@ -1,33 +1,20 @@
 <p>A proper implementation of <code>IDisposable.Dispose</code> should allow for it to be called multiple times on the same object, however this is not
 guaranteed and could result in an exception being thrown.</p>
-<p>It is best not to rely on this behaviour and therefore make sure an object is disposed only once on all execution paths. This is particularly true
+<p>It is best not to rely on this behavior and therefore make sure an object is disposed only once on all execution paths. This is particularly true
```
andrei-epure-sonarsource, line 3: is the nested using statements still valid?
> This is particularly true when dealing with nested using statements.


I would check with @nicolas-harraudeau-sonarsource on how it's best to word the new implementation

sonaranalyzer-dotnet/its/expected/Nancy/Nancy.Tests-{776D244D-BC4D-4BBB-A478-CAF2D04520E1}-S3966.json
```diff
@@ -1,17 +0,0 @@
-{
-"issues":  [
-{
-"id":  "S3966",
-"message":  "Refactor this code to make sure 'innerStream' is disposed only once.",
```
andrei-epure-sonarsource, line 5: was this an instance of the StreamWriter problem?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ObjectsShouldNotBeDisposedMoreThanOnce.cs
```diff
@@ -65,67 +64,20 @@ public void Disposed_Using_WithExpressions()
             }
         }
 
-        public void Disposed_Using3(Stream str)
```
andrei-epure-sonarsource, line 68: I suggest to leave these tests and mark them as compliant.

andrei-epure-sonarsource (changes requested):
Just to drop a few remarks.

The [StreamWriter constructor docs](https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=netframework-4.5#System_IO_StreamReader__ctor_System_IO_Stream_System_Text_Encoding_System_Boolean_System_Int32_System_Boolean_) says

> Unless you set the leaveOpen parameter to true, the StreamReader object calls Dispose() on the provided Stream object when StreamReader.Dispose is called.

In practice it calls `Close()` but close eventually calls `Dispose()`. In the remarks for the [Close() docs](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.close?view=netframework-4.5#remarks)

>This method calls Dispose, specifying true to release all resources. You do not have to specifically call the Close method. Instead, ensure that every Stream object is properly disposed. You can declare Stream objects within a using block (or Using block in Visual Basic) to ensure that the stream and all of its resources are disposed, or you can explicitly call the Dispose method.

Even in the StreamWriter itself, [close() calls Dispose(true)](https://github.com/microsoft/referencesource/blob/master/mscorlib/system/io/streamwriter.cs#L231).

So the reasoning that `Close()` isn't the same thing as `Dispose()` is wrong.

I guess it's still ok to remove this example because we were suggesting to replace it with some cumbersome , not nice try-catch replacement.

sonaranalyzer-dotnet/its/expected/Nancy/Nancy.Tests-{776D244D-BC4D-4BBB-A478-CAF2D04520E1}-S3966.json
costin-zaharia-sonarsource: Yes.

```
using (var innerStream = new MemoryStream())
using (var streamWriter = new StreamWriter(innerStream, Encoding.UTF8) { AutoFlush = true })
```

andrei-epure-sonarsource:
After these changes... will this rule only catch when a resource is manually disposed a second time via `.Dispose()` , after a `using`? In this case, the rule wouldn't be that bad because we'd detect a code smell (hard to read - why `Dispose` if already `using` has been used)?

costin-zaharia-sonarsource:
> I guess it's still ok to remove this example because we were suggesting to replace it with some cumbersome , not nice try-catch replacement.

Additionally, calling dispose twice in this case is not a problem since there is no exception thrown.

sonaranalyzer-dotnet/rspec/cs/S3966_c#.html
nicolas-harraudeau-sonarsource: Thanks for pinging me @andrei-epure-sonarsource.
I rewrote the rule description with the help of @costin-zaharia-sonarsource.

nicolas-harraudeau-sonarsource: @andrei-epure-sonarsource @costin-zaharia-sonarsource I'm just wondering if we shouldn't keep the rule in sonar-way now that we removed the problematic use case. disposing the object twice explicitly is really strange to me.

costin-zaharia-sonarsource:
> After these changes... will this rule only catch when a resource is manually disposed a second time via .Dispose() , after a using? In this case, the rule wouldn't be that bad because we'd detect a code smell (hard to read - why Dispose if already using has been used)?

Yes, what's left now is the check that the `Dispose` was explicitly called twice or after a using.

sonaranalyzer-dotnet/rspec/cs/S3966_c#.html
costin-zaharia-sonarsource: After discussion with @nicolas-harraudeau-sonarsource we decided to add the rule back to Sonar Way. With the latest changes most of the FPs will disappear and the found issues on peach reveal code which is error prone.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ObjectsShouldNotBeDisposedMoreThanOnce.cs
costin-zaharia-sonarsource: As discussed I've added back 2 test cases with streams

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ObjectsShouldNotBeDisposedMoreThanOnce.cs
```diff
@@ -330,70 +167,21 @@ public void Disposed_UsingDeclaration()
...
                 d.Dispose(); // Noncompliant {{Refactor this code to make sure 'd' is disposed only once.}}
             }
-
-            public void Implicit_Disposed_UsingDeclaration(Stream str)
-            {
-                using var s = new FileStream("path", FileMode.Open);
-                using var sr = new StreamReader(s); // Noncompliant {{Refactor this code to make sure 's' is disposed only once.}}
-            }
-
-            public void Disposed_Using_Valid()
```
andrei-epure-sonarsource, line 340: this is still useful, right?

andrei-epure-sonarsource (approved):
LGTM

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ObjectsShouldNotBeDisposedMoreThanOnce.cs
costin-zaharia-sonarsource: We have a similar case (not exactly this one) at line 29.

costin-zaharia-sonarsource: I will add back a simplified version of this one

PR #3648, opened 2020-10-07
Rename Java parent project

pavel-mikula-sonarsource:
For historical reasons, the parent project for both .NET analyzers was still called `sonar-csharp`. With some leftovers on other places. Same confusion is in SonarCloud and [SonarC# parent
](https://sonarcloud.io/dashboard?id=org.sonarsource.dotnet%3Asonar-csharp) project name and project key.
This PR should fix that.

I've stopped CI to prevent creating new duplicate project on SC. My plan is:
* [x] Review this PR
* [x] Freeze commits
* [ ] Wait for last CI to finish in the repo
* [ ] Manually rename SC project and change project key (.NET Champion needed)
* [ ] Run CI against new SC setup
* [ ] Merge
* [ ] Rebase all branches
* [ ] Unfreeze commits

### DIFF
--- a/sonar-csharp-plugin/pom.xml
+++ b/sonar-csharp-plugin/pom.xml
@@ -27,34 +27,6 @@
     </license>
   </licenses>
 
-  <developers>
-    <developer>
-      <id>aleve</id>
-      <name>Amaury Levé</name>
-      <organization>SonarSource</organization>
-    </developer>
-    <developer>
-      <id>vhristov</id>
-      <name>Valeri Hristov</name>
-      <organization>SonarSource</organization>
-    </developer>
-    <developer>
-      <id>jgyerik</id>
-      <name>Janos Gyerik</name>
-      <organization>SonarSource</organization>
-    </developer>
-    <developer>
-      <id>dmeneses</id>
-      <name>Duarte Meneses</name>
-      <organization>SonarSource</organization>
-    </developer>
-    <developer>
-      <id>jhenry</id>
-      <name>Julien Henry</name>
-      <organization>SonarSource</organization>
-    </developer>
-  </developers>
-
   <scm>
     <connection>scm:git:git@github.com:SonarSource/sonar-dotnet.git</connection>
     <developerConnection>scm:git:git@github.com:SonarSource/sonar-dotnet.git</developerConnection>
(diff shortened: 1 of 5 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -199,7 +199,7 @@ public static bool IsStringEmpty(this ExpressionSyntax expression, SemanticModel
                 ?? methodDeclarationBase?.ExpressionBody()?.DescendantNodes()
                 ?? Enumerable.Empty<SyntaxNode>();
 
-            // See issue: https://github.com/SonarSource/sonar-csharp/issues/416
```
andrei-epure-sonarsource, line 202: could you please separate the commit of changing the github URLs (for which the redirect works fine) from this PR which only changes the Maven pom files ?

andrei-epure-sonarsource (changes requested):
Please create a separate PR with the github URL changes (btw the redirect works fine).

This PR should only deal with the maven changes.

Also @tomverin - can the parent pom changes affect anything in the process of deploying to Maven central?

sonar-csharp-plugin/pom.xml
```diff
@@ -4,7 +4,7 @@
 
   <parent>
     <groupId>org.sonarsource.dotnet</groupId>
-    <artifactId>sonar-csharp</artifactId>
+    <artifactId>sonar-dotnet</artifactId>
```
tomverin, line 7: You've changed the artifactId which is good but not the folder which make it inconsistent

pavel-mikula-sonarsource: This kind of confusion is exactly the motivation for the fix 😃 There are two things:

`sonar-csharp` - old root, renamed to `sonar-dotnet`
`sonar-csharp-plugin` - the real plugin, submodule and directory that remained unchanged. Only it's `<parent>` was updated to the new one.

tomverin: Indeed didn't check that it was the parent tag modified here, so let's say I did not write this comment ;)

tomverin (approved):

andrei-epure-sonarsource (approved):
LGTM

PR #3649, opened 2020-10-08
Update reproducer links to sonar-dotnet projects

pavel-mikula-sonarsource:
Cleanup related to #3648 

GitHub project was renamed from `sonar-csharp` to `sonar-dotnet` in the past. Redirect works, but the links are confusing at first glanc.


andrei-epure-sonarsource (approved):
LGTM

PR #3650, opened 2020-10-08
Add integration test with the import of external issues

andrei-epure-sonarsource:
To test [Importing Third-Party Issues](https://docs.sonarqube.org/latest/analysis/external-issues/)

### DIFF
--- a/its/src/test/java/com/sonar/it/csharp/ExternalIssuesTest.java
+++ b/its/src/test/java/com/sonar/it/csharp/ExternalIssuesTest.java
@@ -1,0 +1,106 @@
+/*
+ * SonarSource :: C# :: ITs :: Plugin
+ * Copyright (C) 2011-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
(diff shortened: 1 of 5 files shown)


its/src/test/java/com/sonar/it/csharp/ExternalIssuesTest.java
```diff
@@ -0,0 +1,106 @@
...
+  @Test
+  public void external_issues_imported_by_default_as_code_smells() throws Exception {
+    Tests.analyzeProject(temp, PROJECT, null);
+
+    assertThat(getComponent(PROGRAM_COMPONENT_ID)).isNotNull();
+    List<Issues.Issue> allIssues = getIssues(PROGRAM_COMPONENT_ID);
+
+    assertThat(allIssues).hasSize(5);
+    assertThat(filter(allIssues, SONAR_RULES_PREFIX, RuleType.CODE_SMELL)).hasSize(1);
+    assertThat(filter(allIssues, ROSLYN_RULES_PREFIX, RuleType.CODE_SMELL)).hasSize(4);
```
costin-zaharia-sonarsource, line 63: Since there are not so many issues we could specify them instead of checking how many there are.

its/src/test/java/com/sonar/it/csharp/ExternalIssuesTest.java
```diff
@@ -0,0 +1,106 @@
...
+  @Test
+  public void external_issues_categories_mapping() throws Exception {
+    Tests.analyzeProject(temp, PROJECT, null,
+      "sonar.cs.roslyn.bugCategories", "StyleCop.CSharp.DocumentationRules,StyleCop.CSharp.MaintainabilityRules",
+      "sonar.cs.roslyn.vulnerabilityCategories", "StyleCop.CSharp.OrderingRules");
+
+    assertThat(getComponent(PROGRAM_COMPONENT_ID)).isNotNull();
+    List<Issues.Issue> allIssues = getIssues(PROGRAM_COMPONENT_ID);
+
+    assertThat(filter(allIssues, ROSLYN_RULES_PREFIX, RuleType.CODE_SMELL)).isEmpty();
```
costin-zaharia-sonarsource, line 87: Would be nice to add one more case to map one of the Roslyn categories to code smell to be able to cover also that (`sonar.cs.roslyn.codeSmellCategories`). 

Note: I like that we cover the cases with `no category mapped` and with `multiple categories mapped` and I would keep them.

costin-zaharia-sonarsource (changes requested):
Nice! I've added though 2 comments related to tests.

costin-zaharia-sonarsource (approved):
Nice!

PR #3651, opened 2020-10-09
Update sonarcloud badges after project rename

andrei-epure-sonarsource:
related to #3648
I manually checked the links are correct


pavel-mikula-sonarsource (approved):
LGTM. Good catch!

PR #3663, opened 2020-10-16
GitHub workflow: Request review

pavel-mikula-sonarsource:
### Content
Add GitHub workflow when review is requested:
- Find all linked issues `Fixes #xxxx` or process the PR itself
- Remove assignees
- Assign to the reviewer
- Move card to "Review in progress" column or
- Add card to "Review in progress" column if the PR/Issue is not on Kanban yet

### Concept

For now, it's based on https://github.com/actions/github-script for simplicity (only single all-in-one YML file). Once it's ready for distribution to LT/SonarSource, we can extract it as a reusable action into a separate repo.

### Live playground
- https://github.com/pavel-mikula-sonarsource/GitHubActionPlayground
- Status and log of an action is visible here: https://github.com/pavel-mikula-sonarsource/GitHubActionPlayground/actions
- It takes around 20 sec to queue the action.

### DIFF
--- a/.github/workflows/RequestReview.yml
+++ b/.github/workflows/RequestReview.yml
@@ -73,7 +73,6 @@ jobs:
             }
             //
             async function processIssue(issue) {
-
                 removeAssignees(issue);
                 addAssignee(issue, context.payload.requested_reviewer.login);
                 const card = await findCard(issue.url);
--- a/.github/workflows/RequestReview.yml
+++ b/.github/workflows/RequestReview.yml
@@ -1,0 +1,102 @@
+name: Request review
+
+on:
+  pull_request:
+    types: ["review_requested"]
+
+jobs:
+    assign_and_move_card:
+        name: Assign issue to reviewer and move Kanban card
+        runs-on: ubuntu-latest
+
+        steps:
+            # https://github.com/actions/github-script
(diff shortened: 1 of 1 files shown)


.github/workflows/RequestReview.yml
```diff
@@ -0,0 +1,102 @@
...
+                                let card = cards.data.find(x => x.content_url == content_url);
+                                if (card) {
+                                    return card;
+                                }
+                            }
+                            console.log("Card not found for: " + content_url);
+                            return null;
+                        }
+                        //
+                        async function processIssue(issue) {
```
costin-zaharia-sonarsource, line 52: Shouldn't we check first if the issue is open before making any changes?

costin-zaharia-sonarsource (changes requested):

.github/workflows/RequestReview.yml
pavel-mikula-sonarsource: If we do a mistake by `Fixes #<closed issue>` than I would still move it on Kanban to raise our attention to it. 
I've added the check for direct PR processing, because there's no reason to do it there.

costin-zaharia-sonarsource (approved):
LGTM! Nice!

Issue #3661, opened 2020-10-16
S3353: False positive with implicit conversion


antrv:
### Description

S3353 raises a false positive with implicit conversion from a constant value.

### Repro steps

```
public struct MyStruct<T>
{
    public static implicit operator MyStruct<T>(T value) { return default; }
}

public static class Example
{
    public static MyStruct<int> Get()
    {
        MyStruct<int> x = 1;
        return x;
    }
}
```

### Expected behavior

No messages for line `MyStruct<int> x = 1;`.

### Actual behavior

S3353 message for `x` in line `MyStruct<int> x = 1;`.

### Related information

* C#/VB.NET Plugins Version 8.10.0.19839
* Visual Studio Version 2019 Pro 16.7.6
* MSBuild / dotnet version 
* Operating System - Win 10 x64 pro

pavel-mikula-sonarsource:
Hi @antrv,

Thank you for reporting this case. I can confirm it as a False Positive. The root cause is the same as in #3271 so I'll move your example there. Problem is that we don't check if the target type can/cannot be `const`.

PR #3664, opened 2020-10-20
S3353: Add 2nd FP repro for #3271

pavel-mikula-sonarsource:
Reproduces #3271 as reported in #3661


andrei-epure-sonarsource (approved):
LGTM

Issue #3668, opened 2020-10-26
Add test cases for all rules with relevant C# 9 syntax


costin-zaharia-sonarsource:
Back in summer we've evaluated what rules need to be updated ("Implement"), or at least sanity-checked ("Add UT") in [C#9 Affected rules](https://docs.google.com/spreadsheets/d/1kSWhy6R1al2bJl9JwnzT2oMnhwZH7d1uwidl01SONX0/edit#gid=547594608).

We need to go one by one and add in the existing unit tests the C# 9 syntax we believe may affect these rules.

**Important**: in the [C#9 Affected rules](https://docs.google.com/spreadsheets/d/1kSWhy6R1al2bJl9JwnzT2oMnhwZH7d1uwidl01SONX0/edit#gid=547594608) table, please change "Add UT" to "Implement" when you realize that something doesn't work out of the box.
- highlight with yellow, to underline the need to implement has been confirmed when adding a UT
- if, after adding the UTs, there's no need for an implementation, highlight "Add UT" with green (to know we did it)

![image](https://user-images.githubusercontent.com/38876598/97457912-7d8b2280-193a-11eb-8c9f-5214d982a16b.png)

- [x] S3442	AbstractTypesShouldNotHaveConstructors.cs
- [x] S4432	AesManagedShouldBeWithSecureMode.cs
- [x] S3244	AnonymousDelegateEventUnsubscribe.cs
- [x] S2330	ArrayCovariance.cs
- [x] S2306	AsyncAwaitIdentifier.cs
- [x] S3168	AsyncVoidMethod.cs
- [x] S1200	AvoidExcessiveClassCoupling.cs
- [x] S110	AvoidExcessiveInheritance.cs
- [x] S4583	BeginInvokePairedWithEndInvoke.cs
- [x] S3343	CallerInformationParametersShouldBeLast.cs
- [x] S4462	CallToAsyncMethodShouldNotBeBlocking.cs
- [x] S3247	CastShouldNotBeDuplicated.cs
- [x] S4830	CertificateValidationCheck.cs
- [x] S3928	CheckArgumentException.cs
- [x] S1451	CheckFileLicense.cs
- [x] S100, S101	ClassAndMethodName.cs
- [x] S3453	ClassNotInstantiatable.cs
- [x] S1694	ClassShouldNotBeAbstract.cs
- [x] S3897	ClassWithEqualityShouldImplementIEquatable.cs
[...]

PR #3698, opened 2020-10-30
Add C# 9 test cases (8 rules)

costin-zaharia-sonarsource:
Related to #3668

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnusedPrivateMember.CSharp9.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnusedPrivateMember.CSharp9.cs
@@ -1,12 +1,10 @@
 ﻿using System;
 
-int a = 0; // Compliant - FN
-
+int a = 0;
 int b = 1;
 Console.WriteLine(b);
 
-Product c = new("name", 1); // Compliant - FN
-
+Product c = new("name", 1);
 Product d = new("name", 2);
 Console.WriteLine(d);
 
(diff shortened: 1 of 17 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnusedPrivateMember.CSharp9.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
+
+int a = 0; // Compliant - FN
```
andrei-epure-sonarsource, line 3: this is a local variable in the `main` method - why is this an FN?

same for line 8

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnusedReturnValue.CSharp9.cs
```diff
@@ -0,0 +1,27 @@
...
+SomeGenericMethod<object>();
+var o = SomeGenericMethod2<object>();
+
+T SomeGenericMethod2<T>() where T : class { return null; }
+T SomeGenericMethod<T>() where T : class { return null; } // Compliant
+
+new R().MyMethod();
+i = new R().MyMethod2();
+
+int NeverUsed(int neverUsedVal) // FN - the returned value is not used
```
andrei-epure-sonarsource, line 18: does this work for local functions in C# 7?

andrei-epure-sonarsource (approved):
LGTM, I've left 2 comments

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnusedPrivateMember.CSharp9.cs
costin-zaharia-sonarsource: My bad. It's not.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/UnusedReturnValue.CSharp9.cs
costin-zaharia-sonarsource: Yes, we have a few test cases with local functions and static local functions and the analysis works as expected. It seems we have issues only with local functions defined as top level statements.

PR #3700, opened 2020-10-30
Run all unit tests from the dev-build script

andrei-epure-sonarsource:
to be able to check everything passes before opening a PR


costin-zaharia-sonarsource (approved):
Nice!

PR #3703, opened 2020-11-02
Add C# 9 test cases (8 rules)

costin-zaharia-sonarsource:
Related to #3668

### DIFF
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ThreadStaticNonStaticField.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ThreadStaticNonStaticField.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ThreadStaticNonStaticField.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ThreadStaticNonStaticField.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
--- a/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TooManyGenericParameters.cs
+++ b/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TooManyGenericParameters.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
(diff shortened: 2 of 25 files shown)


sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TooManyGenericParameters.cs
```diff
@@ -90,23 +90,19 @@ protected override void Initialize(ParameterLoadingAnalysisContext context)
...
-
-                case SyntaxKind.StructDeclaration:
-                    return "struct";
-
-                case SyntaxKind.InterfaceDeclaration:
-                    return "interface";
+                SyntaxKind.ClassDeclaration => "class",
+                SyntaxKind.StructDeclaration => "struct",
+                SyntaxKind.InterfaceDeclaration => "interface",
+                _ => GetUnknownType(typeDeclaration)
```
costin-zaharia-sonarsource, line 99: The coverage is low due to this line. The default branch is not reachable since we explicitly subscribe only to class, struct and interface declaration.

I would not remove this though since one could subscribe also for records and forget to update this switch.

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/TooManyGenericParameters.cs
```diff
@@ -127,7 +123,7 @@ private static string GetEnclosingTypeName(MethodDeclarationSyntax methodDeclara
                         return ((InterfaceDeclarationSyntax)parent).Identifier.ValueText;
 
                     default:
-                        parent = methodDeclaration.Parent;
+                        parent = parent.Parent;
```
costin-zaharia-sonarsource, line 126: We had an infinite loop here in case of records :).

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ToStringNoNull.CSharp9.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
```
andrei-epure-sonarsource, line 1: what happens if inside a top-level-statement file, you add a local function

`string ToString() { return null; }`?

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/ThreadStaticNonStaticField.cs
```diff
@@ -1,4 +1,4 @@
-/*
+﻿/*
```
andrei-epure-sonarsource, line 1: nitpick: can revert this

andrei-epure-sonarsource (approved):
LGTM, two minor comments

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ToStringNoNull.CSharp9.cs
costin-zaharia-sonarsource: I've added a new test case: it will not raise an issue since it's not a ToString override.

andrei-epure-sonarsource:
how come you were able to merge with a red quality gate?

costin-zaharia-sonarsource:
> how come you were able to merge with a red quality gate?

Not sure. It was approved so I though that's why the button was accessible but now that you ask I have second thoughts 😄

PR #3704, opened 2020-11-02
Add C# 9 test cases

costin-zaharia-sonarsource:
Related to #3668

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StringOrIntegralTypesForIndexers.CSharp9.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StringOrIntegralTypesForIndexers.CSharp9.cs
@@ -1,0 +1,18 @@
+﻿record Foo
+{
+    public int this[Class index] { get { return 0; } }
+//                  ^^^^^ Noncompliant {{Use string or an integral type here, or refactor this indexer into a method.}}
+
+    public int this[Record index] { get { return 0; } }
+
+    public int this[int index] { get { return 0; } }
+
+    public int this[string index] { get { return 0; } }
+
+    public int this[nint index] { get { return 0; } }
+
+    public int this[nuint index] { get { return 0; } }
+}
+
+class Class { }
+record Record { }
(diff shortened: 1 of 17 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SuppressFinalizeUseless.CSharp9.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+GC.SuppressFinalize(new object()); // Compliant - class is static. Should we raise if `this` is not passed as parameter?
```
andrei-epure-sonarsource, line 3: > Should we raise if `this` is not passed as parameter?

this seems like a different rule? This rule is ""GC.SuppressFinalize" should not be invoked for types without destructors"

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StringOrIntegralTypesForIndexers.CSharp9.cs
```diff
@@ -0,0 +1,18 @@
+﻿record Foo
+{
+    public int this[Class index] { get { return 0; } }
+//                  ^^^^^ Noncompliant {{Use string or an integral type here, or refactor this indexer into a method.}}
+
+    public int this[Record index] { get { return 0; } }
```
andrei-epure-sonarsource, line 6: should add // FN ?

andrei-epure-sonarsource (approved):
LGTM

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/StringOrIntegralTypesForIndexers.CSharp9.cs
andrei-epure-sonarsource: to avoid another round, it's ok to just mark it in the spread sheet that we implement it

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/SuppressFinalizeUseless.CSharp9.cs
costin-zaharia-sonarsource: I've noticed that [CA1816](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1816#cause) raises an issue for this but I like your suggestion too.

PR #3710, opened 2020-11-04
Add C# 9 tests for S4004, S2971, S3240, S1862 and S2760

andrei-epure-sonarsource:
Related to #3668

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalSimplification.FromCSharp9.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalSimplification.FromCSharp9.cs
@@ -1,0 +1,73 @@
+﻿using System;
+
+Apple a = null, b = null;
+bool condition = false;
+
+a = a is not null ? (a) : b; // FN
+
+Apple x;
+if (a is not null) // Noncompliant {{Use the '?:' operator here.}}
+{
+    x = a;
+}
+else
+{
+    x = b;
+}
+
+if (a is null) // FN
+{
+    a = b;
+}
+
+a = (a is not null) ? a : Identity(new()); // FN ??=
+b = (a is not null) ? Identity(a) : Identity(new()); // FN ??
+a = a ?? new(); // Noncompliant {{Use the '??=' operator here.}}
+
+Fruit elem;
+if (condition) // FN, in C# 9 has target typed conditionals
+{
+    elem = new Apple();
+}
+else
+{
+    elem = new Orange();
+}
+
+Apple Identity(Apple o) => o;
+
+// we ignore lambdas because of type resolution for conditional expressions, see CS0173
+Action<int, int> LambdasAreIgnored(bool condition)
+{
+    Action<int, int> myAction;
+    if (condition)
+    {
+        myAction = static (int i, int j) => { };
+    }
+    else
+    {
+        myAction = static (_, _) => { Console.WriteLine(); };
+    }
+    return myAction;
+}
+
+int NestedExpressionWithSwitch(bool condition, int a, int b)
+{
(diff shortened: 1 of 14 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalSimplification.FromCSharp9.cs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+if (a is null) // FN
+{
+    a = b;
+}
+
+a = (a is not null) ? a : Identity(new()); // FN ??=
+b = (a is not null) ? Identity(a) : Identity(new()); // FN ??
+a = a ?? new(); // Noncompliant {{Use the '??=' operator here.}}
```
costin-zaharia-sonarsource, line 25: it's nice that we already support that :)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalSimplification.FromCSharp9.cs
```diff
@@ -0,0 +1,73 @@
...
+
+Fruit elem;
+if (condition) // FN, in C# 9 has target typed conditionals
```
costin-zaharia-sonarsource, line 28: pedantic: "in" is redundant

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalStructureSameCondition.CSharp9.cs
```diff
@@ -0,0 +1,51 @@
+﻿object o;
+int i;
+
+void SimpleTest()
+{
+    if (o is not null) // Secondary [line11]
```
costin-zaharia-sonarsource, line 6: I think it would be better to use something else for the labels than line number since it can get hard to maintain.

If we want to add one more case for which an "using" is needed we would need to update all the labels.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalStructureSameCondition.CSharp9.cs
```diff
@@ -0,0 +1,51 @@
...
+//      ^^^^^^^^^^^^^
+    {
+        // foo
+    }
+    else if (o is not null) // Noncompliant [line11] {{This branch duplicates the one on line 6.}}
+    {
+        var x = 1;
+    }
+}
```
costin-zaharia-sonarsource, line 15: Could we add also a testcase with:
```
if (o is not null) {}
else (o != null) {}
```
in order to document the behavior.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalsWithSameCondition.CSharp9.cs
```diff
@@ -0,0 +1,53 @@
...
+}
+if (c is null) // FN
+{
+    doTheThing(c);
+}
+if (c is not null) // Compliant
+{
+    doTheThing(c);
+}
+
```
costin-zaharia-sonarsource, line 14: Here also we can add a case with:
```
if (o is not null) { ... }
if (o != null) { ... }
```
to document the rule behavior.

costin-zaharia-sonarsource (approved):

andrei-epure-sonarsource:
I found that a rule wasn't working due to TopLevelStatements, please have another look, @costin-zaharia-sonarsource , as I've changed the tests a bit

costin-zaharia-sonarsource (approved):
LGTM!

PR #3712, opened 2020-11-04
Avoid duplication for workaround to support C#9 init

costin-zaharia-sonarsource:
Related to #3668

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
@@ -32,6 +32,7 @@ namespace SonarAnalyzer.UnitTest.TestFramework
     internal struct SolutionBuilder
     {
         private const string GeneratedAssemblyName = "project";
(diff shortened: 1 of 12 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -32,6 +32,7 @@ namespace SonarAnalyzer.UnitTest.TestFramework
     internal struct SolutionBuilder
     {
         private const string GeneratedAssemblyName = "project";
+        private const string InitSnippet = @"namespace System.Runtime.CompilerServices { public class IsExternalInit { } } // See https://github.com/dotnet/roslyn/issues/45510";
```
andrei-epure-sonarsource, line 35: maybe take out the comment and put it above?

andrei-epure-sonarsource (approved):
LGTM, minor comment left

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
costin-zaharia-sonarsource: I was thinking to have it here in case we get an error (3 years from now) for duplicate type (since it will appear on the same line with the error).

PR #3717, opened 2020-11-05
Compute coverage for both .Net 4.8 and .Net 5

costin-zaharia-sonarsource:
As far as I can tell opencover is considering only the first run

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -255,6 +255,7 @@ stages:
 
           # This step needs to be manually done on CI agents. See https://github.com/OpenCover/opencover/raw/master/main/OpenCover.Documentation/Usage.pdf
           regsvr32 $(NUGET_PACKAGES)\opencover\4.7.922\tools\x86\OpenCover.Profiler.dll
+          regsvr32 $(NUGET_PACKAGES)\opencover\4.7.922\tools\x64\OpenCover.Profiler.dll
 
           $supportedFrameworks = @('net48', 'net5.0')
           Foreach ($framework in $supportedFrameworks)
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -66,7 +66,7 @@ stages:
 
           Write-Host "Remove .Net 5 target for UnitTest project"
 
-          ((Get-Content -Path $testProjFileName -raw) -Replace "net48;netcoreapp3.1;net5.0","net48;netcoreapp3.1") | Set-Content -Path $testProjFileName
+          ((Get-Content -Path $testProjFileName -raw) -Replace "net48;net5.0","net48") | Set-Content -Path $testProjFileName
 
           Write-Host "Show test project after we modified the target frameworks to exclude net5"
           Get-Content -Path $testProjFileName
@@ -255,7 +255,7 @@ stages:
           regsvr32 $(NUGET_PACKAGES)\opencover\4.7.922\tools\x86\OpenCover.Profiler.dll
           regsvr32 $(NUGET_PACKAGES)\opencover\4.7.922\tools\x64\OpenCover.Profiler.dll
 
-          $supportedFrameworks = @('net48','netcoreapp3.1','net5.0')
+          $supportedFrameworks = @('net48', 'net5.0')
           Foreach ($framework in $supportedFrameworks)
           {
             & $(NUGET_PACKAGES)\opencover\4.7.922\tools\OpenCover.Console.exe -target:"dotnet.exe" -targetargs:"test $testProjFileName --no-build -c $(BuildConfiguration) -f $framework" -returntargetcode -filter:"+[SonarAnalyzer.CFG]* +[SonarAnalyzer]* +[SonarAnalyzer.CSharp]* +[SonarAnalyzer.Utilities]* +[SonarAnalyzer.VisualBasic]*" -output:tests\coverage.$framework.xml -oldStyle
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -251,8 +251,9 @@ stages:
           & dotnet build $testProjFileName -c $(BuildConfiguration)
           Test-ExitCode "ERROR: UnitTest project build FAILED."
 
-          Write-Host "Register OpenCover profilers. This step needs to be done manually on CI agents. See https://github.com/OpenCover/opencover/raw/master/main/OpenCover.Documentation/Usage.pdf"
-          regsvr32 $(NUGET_PACKAGES)\opencover\4.7.922\tools\x86\OpenCover.Profiler.dll
+          Write-Host "Register OpenCover profilers!"
+
+          # This step needs to be manually done on CI agents. See https://github.com/OpenCover/opencover/raw/master/main/OpenCover.Documentation/Usage.pdf
           regsvr32 $(NUGET_PACKAGES)\opencover\4.7.922\tools\x64\OpenCover.Profiler.dll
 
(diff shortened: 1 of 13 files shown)


costin-zaharia-sonarsource:
Test sample: https://sonarcloud.io/component_measures?id=sonaranalyzer-dotnet&metric=new_coverage&pullRequest=3717&selected=sonaranalyzer-dotnet%3Asonaranalyzer-dotnet%2Fsrc%2FSonarAnalyzer.CSharp%2FCoverageCheck.cs&view=list

azure-pipelines.yml
```diff
@@ -251,20 +251,16 @@ stages:
           & dotnet build $testProjFileName -c $(BuildConfiguration)
           Test-ExitCode "ERROR: UnitTest project build FAILED."
 
-          Write-Host "List Sonar*.dll files"
-          Get-ChildItem -Recurse .\tests\SonarAnalyzer.UnitTest\bin\${BuildConfiguration}\ -Filter Sonar*.dll
+          Write-Host "Register OpenCover profilers. This step needs to be done manually on CI agents. See https://github.com/OpenCover/opencover/raw/master/main/OpenCover.Documentation/Usage.pdf"
```
pavel-mikula-sonarsource, line 254: It seems to me that only the beginning is a message. And the rest is a comment that's actionable only in the YML file, right?
```suggestion
          Write-Host "Register OpenCover profilers."
          #This step needs to be done manually on CI agents. See https://github.com/OpenCover/opencover/raw/master/main/OpenCover.Documentation/Usage.pdf"
```

azure-pipelines.yml
```diff
@@ -251,20 +251,16 @@ stages:
...
+          regsvr32 $(NUGET_PACKAGES)\opencover\4.7.922\tools\x86\OpenCover.Profiler.dll
```
pavel-mikula-sonarsource, line 255: The primary way `-register:user` option doesn't work?
I'd prefer OpenCover to do the dirty jobs by itself. Otherwise, do we know if both 32 and 64 are needed and if/when we can remove 32 for example?

azure-pipelines.yml
```diff
@@ -251,20 +251,16 @@ stages:
...
-          & dotnet test $testProjFileName -c $(BuildConfiguration) --no-build --no-restore --nologo -f netcoreapp3.1
-          Test-ExitCode "ERROR: Unit tests for .NET Core 3.1 FAILED."
-
-          Write-Host "UTs .Net 5"
-          & dotnet test $testProjFileName -c $(BuildConfiguration) --no-build --no-restore --nologo -f net5.0
-          Test-ExitCode "ERROR: Unit tests for .NET 5 FAILED."
+          $supportedFrameworks = @('net48', 'net5.0')
+          Foreach ($framework in $supportedFrameworks)
+          {
+            & $(NUGET_PACKAGES)\opencover\4.7.922\tools\OpenCover.Console.exe -target:"dotnet.exe" -targetargs:"test $testProjFileName --no-build -c $(BuildConfiguration) -f $framework" -returntargetcode -filter:"+[SonarAnalyzer.CFG]* +[SonarAnalyzer]* +[SonarAnalyzer.CSharp]* +[SonarAnalyzer.Utilities]* +[SonarAnalyzer.VisualBasic]*" -output:tests\coverage.$framework.xml -oldStyle
```
pavel-mikula-sonarsource, line 261: Are the parenthesis around build configuration needed? There are no parenthesis around $framework.
```suggestion
            & $(NUGET_PACKAGES)\opencover\4.7.922\tools\OpenCover.Console.exe -target:"dotnet.exe" -targetargs:"test $testProjFileName --no-build -c $BuildConfiguration -f $framework" -returntargetcode -filter:"+[SonarAnalyzer.CFG]* +[SonarAnalyzer]* +[SonarAnalyzer.CSharp]* +[SonarAnalyzer.Utilities]* +[SonarAnalyzer.VisualBasic]*" -output:tests\coverage.$framework.xml -oldStyle
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/MetadataReferences/AspNetCoreMetadataReference.cs
```diff
@@ -18,7 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-#if NETCOREAPP
+#if NET5_0
```
pavel-mikula-sonarsource, line 21: In more files:
```suggestion
#if NET
```

azure-pipelines.yml
```diff
@@ -66,7 +66,7 @@ stages:
 
           Write-Host "Remove .Net 5 target for UnitTest project"
 
-          ((Get-Content -Path $testProjFileName -raw) -Replace "net48;netcoreapp3.1;net5.0","net48;netcoreapp3.1") | Set-Content -Path $testProjFileName
+          ((Get-Content -Path $testProjFileName -raw) -Replace "net48;net5.0","net48") | Set-Content -Path $testProjFileName
```
pavel-mikula-sonarsource, line 69: Do we still need this entire step `Exclude net5 from the main build`? It was introduced during .NET 5 preview times. Back then, net48 and netcoreapp3.1 were build. So can we just remove the step now?

pavel-mikula-sonarsource (changes requested):
I mostly left some questions to answer.

azure-pipelines.yml
costin-zaharia-sonarsource: Yes, `-register:user` does not work for .Net Core and .Net

costin-zaharia-sonarsource: The options are `x86` and `x64`. I will try to see if `x64` is enough.

azure-pipelines.yml
costin-zaharia-sonarsource: Yes, they are needed. 

- `$framework` is a [powershell variable](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_variables?view=powershell-7.1#:~:text=A%20variable%20is%20a%20unit,include%20spaces%20and%20special%20characters.)
- `$(BuildConfiguration)` is an [Azure DevOps predefined variable](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/visual-studio-build?view=azure-devops)

azure-pipelines.yml
costin-zaharia-sonarsource: No it cannot be removed at this moment. `VSBuild` Azure task currently fails when building projects targeting .Net 5.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/MetadataReferences/AspNetCoreMetadataReference.cs
costin-zaharia-sonarsource: I'm not sure about this. It is not in documentation and it seems that [they didn't took a decision yet](https://github.com/dotnet/docs/pull/21556#issuecomment-726853578). When can change this later when we will know more about it.

pavel-mikula-sonarsource (approved):
LGTM, we'll see how to registration check ends

azure-pipelines.yml
costin-zaharia-sonarsource: Unfortunately we need both. `x86` version is used for .Net framwework tests and `x64` is used for .Net 5

PR #3725, opened 2020-11-06
Add C# 9 test cases for S3869, S1854, S3346, S3903, S4524, S3172

andrei-epure-sonarsource:
Related to #3668

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DeclareTypesInNamespacesTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DeclareTypesInNamespacesTest.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
@@ -20,6 +20,8 @@
 
 extern alias csharp;
 extern alias vbnet;
+
+using Microsoft.CodeAnalysis;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarAnalyzer.UnitTest.TestFramework;
 using CSharp = csharp::SonarAnalyzer.Rules.CSharp;
@@ -32,11 +34,9 @@ namespace SonarAnalyzer.UnitTest.Rules
     {
         [TestMethod]
         [TestCategory("Rule")]
-        public void DeclareTypesInNamespaces_CS()
-        {
+        public void DeclareTypesInNamespaces_CS() =>
             Verifier.VerifyAnalyzer(@"TestCases\DeclareTypesInNamespaces.cs",
                 new CSharp.DeclareTypesInNamespaces());
-        }
 
         [TestMethod]
         [TestCategory("Rule")]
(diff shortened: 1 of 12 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DeclareTypesInNamespacesTest.cs
```diff
@@ -54,11 +54,17 @@ public void DeclareTypesInNamespaces_CS_Before8()
 
         [TestMethod]
         [TestCategory("Rule")]
-        public void DeclareTypesInNamespaces_VB()
-        {
+        public void DeclareTypesInNamespaces_CS_AfterCSharp9() =>
+            Verifier.VerifyAnalyzer(@"TestCases\DeclareTypesInNamespaces.AfterCSharp9.cs",
```
costin-zaharia-sonarsource, line 58: Why not follow the same convention with `CSharp9`?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.CSharp9.cs
```diff
@@ -0,0 +1,88 @@
...
+
+    nint two = 2; // Noncompliant
+    two = 3;
+    Foo(two);
+}
+
+void PatternMatch(object param)
+{
+    object a = param;
+    if (a is not null)
```
costin-zaharia-sonarsource, line 36: `Not` is not implemented on our side and I think we should add a case to show this. Maybe something similar with:
```
void PatternMatch2(int a, int b)
{
    if (b is not 5)
    {
        a = 1;
    }
    else if (b is 5)
    {
        a = 2; // Compliant - FN, the parameter value is overwritten on all possible paths
    }
}
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.CSharp9.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        a = null; // Compliant
+        Foo(a);
+    }
+
+    int i = 100;
+    if (i is not > 50 and < 200)
```
costin-zaharia-sonarsource, line 43: Same comment as above. We don't support the new patterns:
```
var c = 5;
    switch (c)
    {
        case < 5:
            c = 5; // Compliant, FN
            break;
        case >= 5:
            c = 56; // Compliant, FN
            break;
    }
}
```

costin-zaharia-sonarsource (approved):
LGTM! Please see my comments with missing test cases for DeadStores.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DeclareTypesInNamespacesTest.cs
andrei-epure-sonarsource: to have the same pattern with the other files
`DeclareTypesInNamespaces_CS_Before8`
`DeclareTypesInNamespaces_CS_After8`

same for the test file name

PR #3726, opened 2020-11-06
Add C# 9 test cases for S2292, S4049, S4275, S1450, S4000, S4214, S3251, S4428 and S927

costin-zaharia-sonarsource:

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PartialMethodNoImplementationTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PartialMethodNoImplementationTest.cs
@@ -35,6 +35,6 @@ namespace SonarAnalyzer.UnitTest.Rules
         [TestMethod]
         [TestCategory("Rule")]
         public void PartialMethodNoImplementation_CSharp9() =>
-            Verifier.VerifyAnalyzer(@"TestCases\PartialMethodNoImplementation.CSharp9.cs", new PartialMethodNoImplementation(), ParseOptionsHelper.FromCSharp9);
+            Verifier.VerifyAnalyzerFromCSharp9Library(@"TestCases\PartialMethodNoImplementation.CSharp9.cs", new PartialMethodNoImplementation());
     }
 }
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PartialMethodNoImplementationTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PartialMethodNoImplementationTest.cs
@@ -35,10 +35,6 @@ namespace SonarAnalyzer.UnitTest.Rules
         [TestMethod]
         [TestCategory("Rule")]
         public void PartialMethodNoImplementation_CSharp9() =>
-            Verifier.VerifyAnalyzer(@"TestCases\PartialMethodNoImplementation.CSharp9.cs",
-                                    new PartialMethodNoImplementation(),
-                                    ParseOptionsHelper.FromCSharp9,
-                                    // Compilation errors are ignored since they are intented.
-                                    checkMode: CompilationErrorBehavior.Ignore);
+            Verifier.VerifyAnalyzer(@"TestCases\PartialMethodNoImplementation.CSharp9.cs", new PartialMethodNoImplementation(), ParseOptionsHelper.FromCSharp9);
     }
 }
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PartialMethodNoImplementationTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PartialMethodNoImplementationTest.cs
@@ -30,10 +30,15 @@ namespace SonarAnalyzer.UnitTest.Rules
     {
         [TestMethod]
         [TestCategory("Rule")]
-        public void PartialMethodNoImplementation()
-        {
-            Verifier.VerifyAnalyzer(@"TestCases\PartialMethodNoImplementation.cs",
-                new PartialMethodNoImplementation());
-        }
+        public void PartialMethodNoImplementation() => Verifier.VerifyAnalyzer(@"TestCases\PartialMethodNoImplementation.cs", new PartialMethodNoImplementation());
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void PartialMethodNoImplementation_CSharp9() =>
+            Verifier.VerifyAnalyzer(@"TestCases\PartialMethodNoImplementation.CSharp9.cs",
+                                    new PartialMethodNoImplementation(),
+                                    ParseOptionsHelper.FromCSharp9,
+                                    // Compilation errors are ignored since they are intented.
+                                    checkMode: CompilationErrorBehavior.Ignore);
     }
 }
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PointersShouldBePrivateTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PointersShouldBePrivateTest.cs
@@ -30,10 +30,13 @@ namespace SonarAnalyzer.UnitTest.Rules
     {
         [TestMethod]
         [TestCategory("Rule")]
-        public void PointersShouldBePrivate()
-        {
-            Verifier.VerifyAnalyzer(@"TestCases\PointersShouldBePrivate.cs",
-                new PointersShouldBePrivate());
-        }
+        public void PointersShouldBePrivate() => Verifier.VerifyAnalyzer(@"TestCases\PointersShouldBePrivate.cs", new PointersShouldBePrivate());
+
+#if NET5_0 // Function pointers are supported only by .Net 5 runtime
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void PointersShouldBePrivate_CSharp9() =>
+            Verifier.VerifyAnalyzer(@"TestCases\PointersShouldBePrivate.CSharp9.cs", new PointersShouldBePrivate(), ParseOptionsHelper.FromCSharp9);
+#endif
     }
 }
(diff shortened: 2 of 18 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PropertyToAutoProperty.CSharp9.cs
```diff
@@ -0,0 +1,37 @@
...
+
+public class MyAttribute : Attribute { }
+
+public record Record
+{
+    private int field1;
+    private string field2;
+    private string field3;
+
+    [My()]
```
andrei-epure-sonarsource, line 11: what is the attribute testing?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PropertyToAutoProperty.CSharp9.cs
```diff
@@ -0,0 +1,37 @@
...
+        set { fieldWithAttribute = value; }
+    }
+
+    public string PropWithGetAndSet // Noncompliant
+    {
+        get { return field2; }
+        set { field2 = value; }
+    }
+
+    public string PropWithGetAndInit // Compliant - FN
```
andrei-epure-sonarsource, line 32: you should also add a
```
public string PropWithGetAndInit { get; init; } // Compliant
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PrivateFieldUsedAsLocalVariable.CSharp9.cs
```diff
@@ -0,0 +1,28 @@
+﻿record Record
+{
+    private int F0 = 0; // Compliant - unused
+    private int F1 = 0; // Compliant - FN
+    public int F2 = 0; // Compliant - Public
```
andrei-epure-sonarsource, line 5: also add an `internal`, `protected`, `protected internal` and `private protected`?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PointersShouldBePrivateTest.cs
```diff
@@ -30,10 +30,13 @@ public class PointersShouldBePrivateTest
...
         [TestMethod]
         [TestCategory("Rule")]
-        public void PointersShouldBePrivate()
-        {
-            Verifier.VerifyAnalyzer(@"TestCases\PointersShouldBePrivate.cs",
-                new PointersShouldBePrivate());
-        }
+        public void PointersShouldBePrivate() => Verifier.VerifyAnalyzer(@"TestCases\PointersShouldBePrivate.cs", new PointersShouldBePrivate());
+
+#if NET5_0 // Function pointers are supported only by .Net 5 runtime
```
andrei-epure-sonarsource, line 35: hmm, I used function pointers in my DeadStore test, w/o compilation directives... and there was no failure https://github.com/SonarSource/sonar-dotnet/pull/3725/files#diff-e80332c6a531ede00e7288259bbce92e353e19b067cc3111b0ccaff17671403fR87

the error would happen at runtime?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PInvokesShouldNotBeVisible.CSharp9.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Runtime.InteropServices;
+
+public record Record
+{
```
andrei-epure-sonarsource, line 4: can you also add a Noncompliant method on the record?

```
        [DllImport("kernel32.dll", CharSet = CharSet.Unicode)]
        public static extern bool RemoveDirectory(string name);  // Noncompliant
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PartialMethodNoImplementationTest.cs
```diff
@@ -30,10 +30,15 @@ public class PartialMethodNoImplementationTest
...
-        }
+        public void PartialMethodNoImplementation() => Verifier.VerifyAnalyzer(@"TestCases\PartialMethodNoImplementation.cs", new PartialMethodNoImplementation());
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void PartialMethodNoImplementation_CSharp9() =>
+            Verifier.VerifyAnalyzer(@"TestCases\PartialMethodNoImplementation.CSharp9.cs",
+                                    new PartialMethodNoImplementation(),
+                                    ParseOptionsHelper.FromCSharp9,
+                                    // Compilation errors are ignored since they are intented.
```
andrei-epure-sonarsource, line 41: nitpick s/intented/intended

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PartialMethodNoImplementation.CSharp9.cs
```diff
@@ -0,0 +1,20 @@
+﻿public partial record Record
+{
+    partial void Method(); //Noncompliant {{Supply an implementation for this partial method.}}
+
+    private partial void M2(); // Noncompliant - FP (there is a compiler error already)
```
andrei-epure-sonarsource, line 5: if you also mention `// Error [CSXXXX]` , will you still need `checkMode: CompilationErrorBehavior.Ignore` ?

you can use `// Noncompliant @-1` I think to have the comment on the second line, to have both `Error` and `Noncompliant`

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterNamesInPartialMethod.CSharp9.cs
```diff
@@ -0,0 +1,19 @@
+﻿public partial record Record
+{
+    partial void M1(int x, int y);
+
+    private partial void M2(int x, int y);
+    internal partial void M3(int x, int y);
+    protected partial void M4(int x, int y);
+    protected internal partial void M5(int x, int y);
+    public partial void M6(int x, int y);
```
andrei-epure-sonarsource, line 9: to test the [extended support](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/extending-partial-methods) for partial methods, you could also add a non-void method like
```
partial bool TryFoo(string input, out string output);
```

andrei-epure-sonarsource (changes requested):

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PropertyToAutoProperty.CSharp9.cs
costin-zaharia-sonarsource: This is a double check that we don't raise for fields which have attributes.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PrivateFieldUsedAsLocalVariable.CSharp9.cs
costin-zaharia-sonarsource: I've assumed they are covered in the test cases we had already but they are not (not all). I've added them.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PointersShouldBePrivateTest.cs
costin-zaharia-sonarsource: I've added the check since I've got this error during analysis:
```
[CS8889]: The target runtime doesn't support extensible or runtime-environment default calling conventions. on line 25
```
on line 25: ```public delegate* unmanaged<int, int> f5;```
I've tried to google it a bit but currently there is not a lot of info about it.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PartialMethodNoImplementation.CSharp9.cs
costin-zaharia-sonarsource: Nice!

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PointersShouldBePrivateTest.cs
andrei-epure-sonarsource: this snippet works in sharplab.io , see [here](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQBMQGoA+ABAzAAmwCZ8BhAWACgBvKq/Bwg7ARgDZCAWfAGQHsA5gAoAlPmr4AvvUYBXAHZQIAMzhdCAVhgBLPvIAKfbfPgJRMhrUqMb+FHAA2cARHgAqADzZOAPnwAHGAQWfABefAAyUgA6fgEAbnwAeiT8ADk9AGM+MH8HbQgTC1tA4LDImLjElPSsnLyCoutGaUpWoA==)
```
public class C
{

    public static void Log() { }
    unsafe void FunctionPointer()
    {
        delegate*<void> ptr1 = &C.Log; // Noncompliant
        ptr1 = &C.Log; // Noncompliant
    }
}
```

andrei-epure-sonarsource (approved):
LGTM

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/PointersShouldBePrivateTest.cs
costin-zaharia-sonarsource: yes, the problem appears when using `unmanaged` (reproducible also on sharplab.io). I think we prefer to keep the case though to document the behavior when these are used.

PR #3729, opened 2020-11-09
Add C# 9 tests for Dispos* rules

andrei-epure-sonarsource:
Related to #3668


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DisposableTypesNeedFinalizers.cs
```diff
@@ -37,21 +37,8 @@ public class Foo_Compliant : IDisposable // Error [CS0535]
             private IntPtr myResource;
             private bool disposed = false;
 
-            protected virtual void Dispose(bool disposing)
```
andrei-epure-sonarsource, line 40: I removed this because the rule currently works with an empty `Dispose()`, so no need to add complex logic in the test

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DisposableNotDisposed.CSharp9.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Net;
+
+var fs0 = new FileStream(@"c:\foo.txt", FileMode.Open); // Noncompliant
+FileStream fs1 = new(@"c:\foo.txt", FileMode.Open); // FN
+
+var fs2 = new FileStream(@"c:\foo.txt", FileMode.Open); // Compliant, passed to a method
+NoOperation(fs2);
+
+FileStream fs3; // Compliant - not instantiated
+using var fs5 = new FileStream(@"c:\foo.txt", FileMode.Open); // Compliant
```
pavel-mikula-sonarsource, line 12: Let's add
```
using FileStream fs6 = new(@"c:\foo.txt", FileMode.Open); // Compliant
```
for future reference.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DisposableNotDisposed.CSharp9.cs
```diff
@@ -0,0 +1,78 @@
...
+    public void Foo()
+    {
+        field_fs4.Dispose();
+
+        FileStream fs5; // Compliant - used properly
+        using (fs5 = new FileStream(@"c:\foo.txt", FileMode.Open))
+        {
+            // do nothing but dispose
+        }
+
```
pavel-mikula-sonarsource, line 71: Let's add
```
        using (fs5 = new(@"c:\foo.txt", FileMode.Open))
        {
            // do nothing but dispose
        }
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DisposableNotDisposed.CSharp9.cs
```diff
@@ -0,0 +1,78 @@
...
+        FileStream fs1 = new(@"c:\foo.txt", FileMode.Open); // FN
+        var fs2 = File.Open(@"c:\foo.txt", FileMode.Open); // Noncompliant - instantiated with factory method
+            // Noncompliant@-1
```
pavel-mikula-sonarsource, line 74: To be consistent with markings above, same for the WebClient
```suggestion
            // Noncompliant@-1 - duplicate
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DisposableReturnedFromUsing.CSharp9.cs
```diff
@@ -0,0 +1,56 @@
...
+    };
+    return result; // FN, we don't track aliasing
+}
+
+FileStream TargetTypedNew()
+{
+    using FileStream fs1 = new(@"c:\foo.txt", FileMode.Open); // Noncompliant
+    return fs1;
+}
+
```
pavel-mikula-sonarsource, line 45: Let's add the older syntax as well
```
FileStream TargetTypedNew2()
{
    using (FileStream fs1 = new(@"c:\foo.txt", FileMode.Open) // Noncompliant
    {
        return fs1;
    }
}
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DisposeFromDisposeTest.cs
```diff
@@ -32,18 +33,19 @@ public class DisposeFromDisposeTest
     {
         [TestMethod]
         [TestCategory("Rule")]
-        public void DisposeFromDispose()
-        {
-            Verifier.VerifyAnalyzer(@"TestCases\DisposeFromDispose.cs", new DisposeFromDispose(), ParseOptionsHelper.FromCSharp8);
-        }
+        public void DisposeFromDispose_BeforeCSharp8() =>
+            Verifier.VerifyAnalyzer(@"TestCases\DisposeFromDispose.BeforeCSharp8.cs", new DisposeFromDispose(),
+                ImmutableArray.Create(new CSharpParseOptions(LanguageVersion.CSharp7_2)));
```
pavel-mikula-sonarsource, line 38: This looks weird compared to method name. Two using statements can be removed afterwards.
```suggestion
                ParseOptionsHelper.BeforeCSharp8);
```
If there's a reason for 7_2-only, we should rename the method and the test case file.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DisposeNotImplementingDispose.CSharp9.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    public void Dispose()
+    {
+        // Dispose(10)
+    }
+}
+
+public partial record MyPartial
+{
+    public void Dispose(int i) // FN, partial classes are not processed, see https://github.com/dotnet/roslyn/issues/3748
```
pavel-mikula-sonarsource, line 33: I don't understand this case. What I see:
There's a `Dispose()` method implementing `IDisposable` for this class. So it's not a FN.
`Dispose(int i)` has similar name, but probably should not have an effect on the test (I haven't seen standalone overload like this in the old test).

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DisposeNotImplementingDispose.CSharp9.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+
+void Dispose() { } // top level local function, compliant
+
+public record R
+{
+    void Dispose() { } //Noncompliant {{Either implement 'IDisposable.Dispose', or totally rename this method to prevent confusion.}}
+}
+
```
pavel-mikula-sonarsource, line 9: Let's add partial class with partial Dispose method.

pavel-mikula-sonarsource (changes requested):
Some improvement suggestions.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DisposeNotImplementingDispose.CSharp9.cs
andrei-epure-sonarsource: this was actually a copy paste of the test in `TestCases\DisposeNotImplementingDispose.cs` ... indeed

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DisposeFromDisposeTest.cs
andrei-epure-sonarsource:
```
   Test method SonarAnalyzer.UnitTest.Rules.DisposeFromDisposeTest.DisposeFromDispose_BeforeCSharp8 threw exception:
SonarAnalyzer.UnitTest.TestFramework.UnexpectedDiagnosticException: CSharp5: Unexpected build error [CS8026]: Feature 'readonly structs' is not available in C# 5. Please use language version 7.2 or greater. on line 7
  Stack Trace:
```

I'll update the name

andrei-epure-sonarsource: Note: I don't see any reason in adding tests before C# 7.2, which already is 3 years old

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DisposableMemberInNonDisposableClass.cs
```diff
@@ -125,4 +125,22 @@ private sealed class DisposableStuff : IDisposable
             public void Dispose() { }
         }
     }
+
+    partial class PartialMethod : IDisposable
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: In the new commit: `PartialMethod1` looks equal to `PartialMethod2` to me from functional and RSPEC POV. But one is compliant and one is noncompliant. That doesn't match :) So the `PartialMethod2` should be marked as FP?

pavel-mikula-sonarsource (commented):
There's one last thing that confuses me.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DisposableMemberInNonDisposableClass.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: No, because it doesn't implement IDisposable.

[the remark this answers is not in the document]
pavel-mikula-sonarsource: That's what I've missed 💡

pavel-mikula-sonarsource (approved):
LGTM

PR #3739, opened 2020-11-11
Add C# 9 tests for the rest of D* rules

andrei-epure-sonarsource:
Related to #3668


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotExposeListT.cs
```diff
@@ -34,7 +34,7 @@ public class Bar<T>
 
         public void Foo()
         {
-            Action<List<int>> x = (List<int> x) => { };
+            Action<List<int>> x = (List<int> y) => { };
```
andrei-epure-sonarsource, line 37: fix needed for older C# versions

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotShiftByZeroOrIntSize.CSharp9.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+
+// Native Integers can be either 32b or 64b depending on the underlying system.
+// We should only raise when the shift is done with values >= 64.
```
pavel-mikula-sonarsource, line 4: I think we should not raise for big values, because we don't know what's the `nint` size of the run machine. Will be 128 one day. Could be less on some super exotic scenario 🤯

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotThrowFromDestructors.CSharp9.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        ~Program()
+        {
+            throw new Exception(); // Noncompliant {{Remove this 'throw' statement.}}
+//          ^^^^^^^^^^^^^^^^^^^^^^
+
+            void Inner()
+            {
+                throw new Exception(); // Noncompliant, C# 7+
```
pavel-mikula-sonarsource, line 21: _Nitpick:_ We don't need this comment since we're in CSharp9 file.
```suggestion
                throw new Exception(); // Noncompliant
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotThrowFromDestructors.CSharp9.cs
```diff
@@ -0,0 +1,47 @@
...
+                Foo();
+            }
+            catch (Exception)
+            {
+                throw; // Noncompliant, rethrowing has the same effect as throwing
+            }
+        }
+    }
+
+    class C
```
pavel-mikula-sonarsource, line 43:
```suggestion
    record C
```

pavel-mikula-sonarsource (changes requested):
Some small stuff

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DoNotExposeListTTest.cs
```diff
@@ -30,11 +31,20 @@ public class DoNotExposeListTTest
...
-        {
-            Verifier.VerifyAnalyzer(@"TestCases\DoNotExposeListT.cs",
-                new DoNotExposeListT());
-        }
+        public void DoNotExposeListT() =>
+            Verifier.VerifyAnalyzer(@"TestCases\DoNotExposeListT.cs", new DoNotExposeListT());
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void DoNotExposeListT_CSharp9_TopLevelStatements() =>
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: because the records will be nested types, and to avoid different behavior for nested types vs top-level types

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotShiftByZeroOrIntSize.CSharp9.cs
andrei-epure-sonarsource: So do you suggest to avoid raising on `nint`? I'm fine with that

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DoNotExposeListTTest.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: from https://jira.sonarsource.com/browse/RSPEC-3956

> This rule raises an issue every time a System.Collections.Generic.List<T> is exposed:
    As an externally visible member.
    As the return type of an externally visible method.
    As a parameter type of an an externally visible method.

I am not sure how the types nested under Main can be accessed.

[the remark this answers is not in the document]
pavel-mikula-sonarsource: The public class is not nested according to [sharplab](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQBMQGoA+ABATARgFgAobABgAJt8AWAbhJIDcIEKAPCgXgvweJLYAzFVwUAQqwA8AFQB8JAN4kKqigHp1FKAAsA9gFcANigoBLAHYXEouFHMWKMPQAcAtEbhM4R7TAjwYHAWMBQAxnoWUHqeFBAuLvYIEGZQcAD8KmrCFAAyqTCychQAsnAw+ij4RQAUMnEIAOYAlNzFFsZGdBpaAHKREWAuRmYQISQAviRAA===)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/DoNotShiftByZeroOrIntSize.CSharp9.cs
pavel-mikula-sonarsource: Yes. Let's don't raise on `nint` and `nuint` even for `1 << 512`. 

I did it the same way in CBDE overflow rule. It's even more visible on that example.

pavel-mikula-sonarsource (commented):
Let's make the update on the last open thread or finish the discussion there.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DoNotExposeListTTest.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: good catch!

pavel-mikula-sonarsource (approved):
LGTM

PR #3741, opened 2020-11-11
Run tests for all platforms in dev build

andrei-epure-sonarsource:
This should have been done in #3700 ... better late than never.

In #3700 I forced the same behavior as on AZP.

By running `.\scripts\build\dev-build.ps1 -build -test -release` before opening a PR we'll make sure we won't have surprises.

### DIFF
--- a/scripts/build/build-utils.ps1
+++ b/scripts/build/build-utils.ps1
@@ -147,6 +147,15 @@ function Invoke-UnitTests([string]$binPath, [bool]$failsIfNotTest) {
 
     & (Get-VsTestPath) $testFiles /Parallel /Enablecodecoverage /InIsolation /Logger:trx /TestAdapterPath:$testDirs
     Test-ExitCode "ERROR: Unit Tests execution FAILED."
+
+    $testProjFileName = "tests\SonarAnalyzer.UnitTest\SonarAnalyzer.UnitTest.csproj"
+    & dotnet build $testProjFileName
+
+    dotnet test  $testProjFileName --no-build --no-restore --nologo -f netcoreapp3.1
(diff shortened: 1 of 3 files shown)


scripts/build/build-utils.ps1
```diff
@@ -147,6 +147,15 @@ function Invoke-UnitTests([string]$binPath, [bool]$failsIfNotTest) {
 
     & (Get-VsTestPath) $testFiles /Parallel /Enablecodecoverage /InIsolation /Logger:trx /TestAdapterPath:$testDirs
     Test-ExitCode "ERROR: Unit Tests execution FAILED."
+
```
pavel-mikula-sonarsource, line 150: I might be useful to add log like `Write-Header "Running unit tests .NET Core 3.1"` to report the stage. Also update the message for the first run on line 131

scripts/build/build-utils.ps1
```diff
@@ -147,6 +147,15 @@ function Invoke-UnitTests([string]$binPath, [bool]$failsIfNotTest) {
...
+    $testProjFileName = "tests\SonarAnalyzer.UnitTest\SonarAnalyzer.UnitTest.csproj"
+    & dotnet build $testProjFileName
```
pavel-mikula-sonarsource, line 152: Line 138 expects `*.UnitTest.dll` to exist already. Why is it build here then?

pavel-mikula-sonarsource (changes requested):
Improvement suggestion and one question that confuses me.

scripts/build/build-utils.ps1
andrei-epure-sonarsource: I actually had to pass the buildConfiguration given to the script

pavel-mikula-sonarsource (approved):
LGTM

PR #3742, opened 2020-11-11
Add C# 9 tests for In* rules

pavel-mikula-sonarsource:
Related to #3668


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InsecureHashAlgorithm.NetCore.cs
```diff
@@ -0,0 +1,100 @@
+﻿using System.Security.Cryptography;
```
costin-zaharia-sonarsource, line 1: This duplicates much of the code tested by InsecureHashAlgorithm on .Net Framework and I think it does not bring much value since the test is more library oriented. 

If you really want to add tests for .Net core I think we should move the code which can be tested by all languages in a file and then have dedicated files per framework just with differences.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/InterfaceMethodsShouldBeCallableByChildTypes.CSharp9.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+
+namespace Tests.Diagnostics
```
costin-zaharia-sonarsource, line 3: pedantic: I think we can skip namespaces from most of the tests (if we remove all the boilerplate code the test cases will get a bit more visible)

costin-zaharia-sonarsource (changes requested):

costin-zaharia-sonarsource (approved):
LGTM!

PR #3744, opened 2020-11-12
Add C# 9 test cases for S1226,  S1206, S3447, S3466, S4050, S4069, S1123, S3937 and S3459

costin-zaharia-sonarsource:
Related to #3668

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterAssignedTo.CSharp9.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterAssignedTo.CSharp9.cs
@@ -1,0 +1,21 @@
+﻿using System;
+
+static void Foo(int x)
+{
+    Action<int, int> discard = (_, _) => _ = 10; // Compliant
+
+    Action<int, int> foo = (a, b) => a = 10; // Noncompliant
+}
+
+public record Record
+{
+    void Foo(int x)
+    {
+        x = 42; // Noncompliant
+    }
+
+    void Foo(nint x)
+    {
+        x += 42; // Compliant
+    }
+}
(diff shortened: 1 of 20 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterAssignedTo.CSharp9.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+static void Foo(int x)
+{
+    Action<int, int> discard = (_, _) => _ = 10; // Compliant
```
andrei-epure-sonarsource, line 5: also a test would be nice
```
    Action<int> underscoreName = (_) => _ = 10; // Noncompliant
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterAssignedTo.CSharp9.cs
```diff
@@ -0,0 +1,21 @@
...
+
+    Action<int, int> foo = (a, b) => a = 10; // Noncompliant
+}
+
+public record Record
```
andrei-epure-sonarsource, line 10: how does this rule behave with `set` and `init` for Properties, if `value` is assigned?
IMO it should raise

andrei-epure-sonarsource (approved):
LGTM, I suggest to add two more tests for ParameterAssignedTo before merging

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/ParameterAssignedTo.CSharp9.cs
costin-zaharia-sonarsource: Nice! I've added the test cases. 

It works as expected but I did not think at all that you might assign to `value`.

PR #3745, opened 2020-11-12
Add C# 9 tests fro S1309, S5659, S4433, S2184

pavel-mikula-sonarsource:
Related to #3668


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/LossOfFractionInDivision.CSharp9.cs
```diff
@@ -0,0 +1,17 @@
+﻿decimal dec = 3 / 2;    // Noncompliant
+dec = 3L / 2;           // Noncompliant
+Method(3 / 2);          // Noncompliant
+dec = (decimal)3 / 2;
+Method(3.0F / 2);
+
+decimal ndec = (nint)3 / 2; // FN
```
costin-zaharia-sonarsource, line 7: I think we should add cases when both operands are nint or nuint.
something like:
```
nint a = 2;
nint b = 4;
decimal x = b/a;
```

costin-zaharia-sonarsource (approved):
LGTM! Just a small comment to add a few more testcases.

PR #3746, opened 2020-11-12
Add C# 9 test for Member* and Method* rules

pavel-mikula-sonarsource:
Related to #3668


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MemberOverrideCallsBaseMember.CSharp9.cs
```diff
@@ -0,0 +1,16 @@
+﻿
```
costin-zaharia-sonarsource, line 1: pedantic: empty line

there are a few more in the other test case files

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MemberShouldBeStatic.CSharp9.cs
```diff
@@ -0,0 +1,23 @@
+﻿
+int TopLevelMethod() => 0;              // FN
+
+var localVariable = 42;
+int LocalFunction() => localVariable;   // Compliant
```
costin-zaharia-sonarsource, line 5: I think we can add one more case with static lambdas to document the behavior.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodsShouldNotHaveTooManyLines_CustomValues.CSharp9.cs
```diff
@@ -0,0 +1,33 @@
+﻿
```
costin-zaharia-sonarsource, line 1: could we add one more case with a local function declared at the top level?

costin-zaharia-sonarsource (approved):
LGTM! I would add though a few more test cases. Please see my comments.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MemberShouldBeStatic.CSharp9.cs
pavel-mikula-sonarsource: There's `New RSPEC: Use static lambdas when possible.` note in the spreadsheet for lambdas. And since lambda is not a member, I didn't added as it's not relevant for this rule.

costin-zaharia-sonarsource: In order to document the behavior we can still add a case to show that this rule will not trigger for those cases and that it's a decision we took not a case we missed.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/MethodsShouldNotHaveTooManyLines_CustomValues.CSharp9.cs
```diff
@@ -1,4 +1,17 @@
...
+i++;
+i++;
+i++;
+
+void LocalFunction() // FN
+{
+    i++;
+    i++;
+    i++;
+    i++;
```
costin-zaharia-sonarsource, line 12: Nice! We found now a few more FNs

pavel-mikula-sonarsource: It was a nasty nice catch :)

PR #3748, opened 2020-11-13
Add C# * tests for F* rules

pavel-mikula-sonarsource:
Related to #3668


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldBeReadonly.CSharp9.cs
```diff
@@ -0,0 +1,28 @@
+﻿record Person
+{
+    private int birthYear;     // Noncompliant {{Make 'birthYear' 'readonly'.}}
+    int birthMonth = 3;        // Noncompliant
+
+    int legSize1 = 3;
+    int legSize2 = 3;
+    bool usedInInit = false;    // FN, only written in property init
+
```
andrei-epure-sonarsource, line 9: what about static fields and `Module Initializers` from C# 9 (which is like a static constructor)?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldNotBePublic.CSharp9.cs
```diff
@@ -0,0 +1,30 @@
+﻿public record Sample
+{
+    public readonly double Pi3 = 3.14;  // Noncompliant {{Make 'Pi3' private.}}
```
andrei-epure-sonarsource, line 3: can you add a test with 
- static field
- const field

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldsShouldNotBePublic.CSharp9.cs
```diff
@@ -0,0 +1,27 @@
+﻿public record PublicClass
```
andrei-epure-sonarsource, line 1: FieldShouldBeReadonly and FieldsShouldBeReadonly is quite confusing

could we rename FieldsShouldNotBePublic to FieldsShouldBeEncapsulatedInProperties ? 

I left a message on https://jira.sonarsource.com/browse/RSPEC-1104 whether RSPEC-2357  should be deprecated in favor to RSPEC-1104

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FunctionComplexity.CSharp9.cs
```diff
@@ -0,0 +1,42 @@
+﻿//FN
```
andrei-epure-sonarsource, line 1: given we have boolean expressions in the main test, could you add some test cases with the new Pattern matching improvements ?

andrei-epure-sonarsource (changes requested):
I left some small suggestions

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/FieldShouldBeReadonlyTest.cs
```diff
@@ -30,23 +30,22 @@ public class FieldShouldBeReadonlyTest
...
-                new FieldShouldBeReadonly(),
-                ParseOptionsHelper.FromCSharp8);
-        }
+        public void FieldShouldBeReadonly() =>
+            Verifier.VerifyAnalyzer(@"TestCases\FieldShouldBeReadonly.cs", new FieldShouldBeReadonly(), ParseOptionsHelper.FromCSharp8);
+
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void FieldShouldBeReadonly_CSharp9() =>
+            Verifier.VerifyAnalyzerFromCSharp9Library(@"TestCases\FieldShouldBeReadonly.CSharp9.cs", new FieldShouldBeReadonly());
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: No. We haven't added codefix tests in this sprint. And there's nothing C# 9 specific in the field declaration itself where the codefix is.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShadowsParentField.CSharp9.cs
```diff
@@ -0,0 +1,11 @@
+﻿public record Fruit
+{
+    protected int ripe;
+    protected static int leafs;
+}
+
+public record Raspberry : Fruit
+{
+    private bool ripe;  // Noncompliant {{'ripe' is the name of a field in 'Fruit'.}}
+    protected static int leafs; // Compliant, static is ignored
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: The C# specific RSPEC has
```
protected string plantingSeason;
```
that looks equal to the existing case:
```
protected int ripe;
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldBeReadonly.CSharp9.cs
pavel-mikula-sonarsource: This rule is not relevant for static fields. Only instance fields are inspected.

https://github.com/SonarSource/sonar-dotnet/blob/204f23b90d8c3e70c1921f01df3877989f8ff5f5/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/FieldShouldBeReadonly.cs#L343

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldNotBePublic.CSharp9.cs
pavel-mikula-sonarsource: Rule's not relevant for static fields. 
https://github.com/SonarSource/sonar-dotnet/blob/204f23b90d8c3e70c1921f01df3877989f8ff5f5/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Rules/FieldShouldNotBePublicBase.cs#L37

Case with const field is in the `FieldShouldNotBePublic.cs`. For C# 9 context, I wanted to test only the minimal compliant/noncompliant scenarios.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/FieldShouldBeReadonlyTest.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: I added codefix tests, I saw one interesting problem in one of them

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShadowsParentField.CSharp9.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: I meant the fact that this rule specifically ignores `static` fields
It's not written anywhere in the RSPEC, normally we have an `Exceptions` sections
```
public class C {
    public static string field;
}

public class D : C
{
    public static string Field;
}
```

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldBeReadonly.CSharp9.cs
andrei-epure-sonarsource: fair point, the RSPEC mentions `readonly fields can only be assigned in a class constructor. `

I left a comment on the RSPEC anyway https://jira.sonarsource.com/browse/RSPEC-2933

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldNotBePublic.CSharp9.cs
andrei-epure-sonarsource: regarding the argument that the rule is just for static, if we remove that line, will a test fail?

if we remove lines from a rule and its tests don't fail, are we really testing the behavior of the rule?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldBeReadonly.CSharp9.cs
andrei-epure-sonarsource: although it would be nice to have a test to document this behavior

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShouldNotBePublic.CSharp9.cs
andrei-epure-sonarsource: nevermind, there are tests for static in the main test case

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/FieldShouldBeReadonlyTest.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: This is so small and simple that it doesn't seem to require a specific test in C# 9 context. The codefix doesn't leave `FieldDeclarationSyntax` node.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/FieldShadowsParentField.CSharp9.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I'm sorry, I've miss read your comment :(

andrei-epure-sonarsource (approved):
LGTM

sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/RspecStrings.resx
```diff
@@ -11234,7 +11234,7 @@
     <value>Critical Vulnerability</value>
   </data>
   <data name="S5547_Description" xml:space="preserve">
-    <value>Strong cipher algorithms are cryptographic systems resistant to cryptanalysis, they are not vulnerable to well-known attacks like brute force attacks for example.</value>
+    <value>Strong cipher algorithms are cryptographic systems resistant to cryptanalysis, they are not vulnerable to well-known attacks like brute force attacks for example. </value>
```
andrei-epure-sonarsource, line 11237: nitpick: unrelated change, but I'll merge anyway

pavel-mikula-sonarsource: ah, I just checked the HTML files and missed this one :( Where's the space coming from? I wanted to fix it in Jira before rspec update, but I didn't find it.

PR #3749, opened 2020-11-13
Add C#9 tests for S1186, S1116, S2291 , S2197, S4035, S4220

andrei-epure-sonarsource:
Related to #3668

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/EmptyMethodTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/EmptyMethodTest.cs
@@ -47,8 +47,7 @@ namespace SonarAnalyzer.UnitTest.Rules
         [TestCategory("Rule")]
         public void EmptyMethod_CSharp9()
         {
-            Verifier.VerifyAnalyzer(@"TestCases\EmptyMethod.CSharp9.cs",
-                new EmptyMethod(), ParseOptionsHelper.FromCSharp9, OutputKind.ConsoleApplication);
+            Verifier.VerifyAnalyzerFromCSharp9Console(@"TestCases\EmptyMethod.CSharp9.cs", new EmptyMethod());
         }
 #endif
 
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/EmptyMethodTest.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/EmptyMethodTest.cs
@@ -23,6 +23,7 @@ using Microsoft.VisualStudio.TestTools.UnitTesting;
 using csharp::SonarAnalyzer.Rules.CSharp;
 using SonarAnalyzer.UnitTest.MetadataReferences;
 using SonarAnalyzer.UnitTest.TestFramework;
+using Microsoft.CodeAnalysis;
 
 namespace SonarAnalyzer.UnitTest.Rules
 {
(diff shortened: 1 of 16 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/EmptyMethodTest.cs
```diff
@@ -41,6 +42,16 @@ public void EmptyMethod()
                 options: ParseOptionsHelper.FromCSharp8);
         }
 
+#if NET5_0
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void EmptyMethod_CSharp9()
+        {
+            Verifier.VerifyAnalyzer(@"TestCases\EmptyMethod.CSharp9.cs",
```
costin-zaharia-sonarsource, line 50: It was intended to use this instead of `VerifyAnalyzerFromCSharp9Console`?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/EmptyMethod.CSharp9.cs
```diff
@@ -0,0 +1,109 @@
...
+    public override void F2()
+    {
+    }
+}
+
+class WithProp
+{
+    public string Prop
+    {
+        init { }
```
costin-zaharia-sonarsource, line 62: Shouldn't this be a FN since it's not an auto-implemented property?

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/EventHandlerDelegateShouldHaveProperArguments.CSharp9.cs
```diff
@@ -0,0 +1,31 @@
...
+//                    ^^^^^^^^^^^^^^^^
+            SomeEvent?.Invoke(this, null); // Noncompliant {{Use 'EventArgs.Empty' instead of null as the event args of this event invocation.}}
+            SomeEvent?.Invoke(null, null); // Noncompliant
+                                           // Noncompliant@-1
+
+            SomeStaticEvent?.Invoke(null, e);
+            SomeStaticEvent?.Invoke(this, e); // Noncompliant {{Make the sender on this static event invocation null.}}
+            SomeStaticEvent?.Invoke(null, null); // Noncompliant {{Use 'EventArgs.Empty' instead of null as the event args of this event invocation.}}
+            SomeStaticEvent?.Invoke(this, null); // Noncompliant
+                                                // Noncompliant@-1
```
costin-zaharia-sonarsource, line 28: Should we add some cases with direct invocation like below?
```
SomeEvent(null, e);
SomeStaticEvent(null, null);
```

costin-zaharia-sonarsource (approved):

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/EmptyMethod.CSharp9.cs
andrei-epure-sonarsource: the current rule only looks at methods, so it's the same behavior as for `set`.
I've opened https://github.com/SonarSource/sonar-dotnet/issues/3753 to do it

costin-zaharia-sonarsource: Shouldn't we mark both as FNs then?

PR #3752, opened 2020-11-16
Add C# 9 unit tests for S4055

pavel-mikula-sonarsource:
Related to #3668

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/LiteralsShouldNotBePassedAsLocalizedParameters.CSharp9.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/LiteralsShouldNotBePassedAsLocalizedParameters.CSharp9.cs
@@ -1,0 +1,27 @@
+﻿using System.ComponentModel;
+
+TopLevelLocalFunction("literal", "literal"); //Noncompliant
+                                             //Noncompliant@-1
+
+void TopLevelLocalFunction([Localizable(true)] string param1, string message)
+{
+}
+
+record Record
+{
+    [Localizable(true)]
+    public string Property1 { get; set; }
+    public string Property2 { get; set; }
+
+    public void Method()
+    {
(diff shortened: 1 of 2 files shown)


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/LiteralsShouldNotBePassedAsLocalizedParameters.CSharp9.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.ComponentModel;
+
+TopLevelLocalFunction("literal", "literal"); //Noncompliant
+                                             //Noncompliant@-1
```
costin-zaharia-sonarsource, line 4: Isn't the second diagnostic a FP? As far as I can tell only the first parameter is decorated. For `Property2` there is no issue raised.

costin-zaharia-sonarsource (approved):
LGTM! I've added just a small comment.

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/LiteralsShouldNotBePassedAsLocalizedParameters.CSharp9.cs
pavel-mikula-sonarsource: The RSPEC states that the rule is trigger also for
>The parameter or property name contains "Text", "Message", or "Caption".

As the [original test](https://github.com/SonarSource/sonar-dotnet/blob/2850f216baf6a41e1cd7478918abc61a3eea73d2/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/LiteralsShouldNotBePassedAsLocalizedParameters.cs#L23) tests this explicit list, I wanted to demonstrate one attribute-based and one name-based scenario in C# 9 context.

So it should be a true positive.

Issue #3751, opened 2020-11-14
S2699: should be skipped on tests with TestCaseSourceAttribute and return values


alexandersorokin:
### Description

Rule ID: S2699

It does not respect `NUnit` `TestCaseSourceAttribute` for methods with return value.

### Repro steps
1. Reference NUnit 3;
2. Run following code:
```
[TestFixture]
internal class Equals_Should
{
  private static readonly TestCaseData[] Cases =
  {
    new TestCaseData("foo", "bar").Returns(false),
    new TestCaseData(100, 100).Returns(true),
  };

  [Test]
  [TestCaseSource(nameof(Cases))]
  public bool Test(object obj1, object obj2)
  {
    return obj1.Equals(obj2);
  }
}
```

### Expected behavior

No warnings as return value is checked by NUnit.

### Actual behavior

```
Warning	S2699	Add at least one assertion to this test case
```

### Known workarounds

No workarounds was found.

### Related information

* C# 8
* Visual Studio 2019 Pro. Version 16.7.6
* donetcore3.1
* Windows 10. Version 20H2.

alexandersorokin:
So, I think that the `S2699` rule should run on test methods with `TestCaseSourceAttribute` only if the test method returns `void` or `Task`.

costin-zaharia-sonarsource:
Hi @alexandersorokin. Thanks for your feedback. I can confirm the false positive.

Would you be interested to write a fix for it?

Thanks,
Costin

alexandersorokin:
Yes, I can fix it. But firstly we should discuss how it should be fixed.

I think that the right way to fix the bug is to split the rule `S2699` into two rules.

The first rule should ensure that a test method has at least one assertion or has a return value. So it is going to be the same as the currently implemented rule `S2699` apart from two things. Firstly, it should not make use of `TestCaseAttribute`. Secondly, it should not show the warning if a test method has a return value.
[...]

PR #3755, opened 2020-11-16
Rule S2699: Document false positive

costin-zaharia-sonarsource:
Related to #3751

### DIFF
--- a/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.NUnit.cs
+++ b/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.NUnit.cs
@@ -282,6 +282,21 @@
         public void TestCaseSource15() // Don't raise on skipped test methods
         {
         }
+
+        private static readonly TestCaseData[] Cases =
+        {
+          new TestCaseData("foo", "bar").Returns(false),
+          new TestCaseData(100, 100).Returns(true),
+        };
+
+        // See: https://github.com/SonarSource/sonar-dotnet/issues/3751
+        // Returned value is verified by the framework and no assertion is needed.
+        [Test]
+        [TestCaseSource(nameof(Cases))] 
+        public bool Test(object obj1, object obj2) // Noncompliant - FP
+        {
+            return obj1.Equals(obj2);
+        }
     }
 
     [TestFixture]


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.NUnit.cs
```diff
@@ -282,6 +282,21 @@ public void TestCaseSource14()
...
+        {
+          new TestCaseData("foo", "bar").Returns(false),
+          new TestCaseData(100, 100).Returns(true),
+        };
+
+        // See: https://github.com/SonarSource/sonar-dotnet/issues/3751
+        // Returned value is verified by the framework and no assertion is needed.
+        [Test]
+        [TestCaseSource(nameof(Cases))] 
+        public bool Test(object obj1, object obj2) // Noncompliant - FP
```
pavel-mikula-sonarsource, line 296: Is there a documentation for the `bool` return type? I've only found `void` examples in the [TestCaseSource docs](https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcasesource.html)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.NUnit.cs
```diff
@@ -282,6 +282,21 @@ public void TestCaseSource14()
...
+        {
```
pavel-mikula-sonarsource, line 297: Can you please add two more testcases with `void` return type? One TP without an assert. One compliant with assert. Just to be sure we don't ignore all methods with that attribute.

pavel-mikula-sonarsource (changes requested):

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.NUnit.cs
costin-zaharia-sonarsource: Not sure why it does not appear anymore in docs, I've used that in the past and the method is still there in  the latest NUnit version (3.12). 

You can find this mentioned in [2.6.4 docs](https://docs.nunit.org/2.6.4/testCaseSource.html)

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.NUnit.cs
costin-zaharia-sonarsource: As far as I can tell both these cases were already covered in the test cases above. See lines 210 and 217.

pavel-mikula-sonarsource (approved):
LGTM

PR #3772, opened 2020-11-17
Run C#9 test cases only on .Net 5

### LINKED ISSUE #3738: Run C#9 tests only `#if NET`
There's no use case where C# 9 is going to be able to run under .NET Framework, so no need to run tests there. Or in NETCOREAPP 3.x.

costin-zaharia-sonarsource:
Fixes #3738


sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/AvoidExcessiveClassCouplingTest.cs
```diff
@@ -33,10 +33,12 @@ public class AvoidExcessiveClassCouplingTest
         public void AvoidExcessiveClassCoupling() =>
             Verifier.VerifyAnalyzer(@"TestCases\AvoidExcessiveClassCoupling.cs", new AvoidExcessiveClassCoupling { Threshold = 1 });
 
+#if NET
```
andrei-epure-sonarsource, line 36: also
- AvoidExcessiveClassCoupling_InRecord_Enums_Not_Counted
- AvoidExcessiveClassCoupling_Base_Records_Interfaces_NotCounted
- AvoidExcessiveClassCoupling_Primitive_Types_Not_Counted

sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/Rules/DisposableMemberInNonDisposableClassTest.cs.orig
```diff
@@ -0,0 +1,100 @@
+﻿/*
```
andrei-epure-sonarsource, line 1: this seems that it's an artefact that needs to be removed?

andrei-epure-sonarsource (changes requested):
There are also tests which use `ParseOptionsHelper.FromCSharp9` and I see they haven't been ported to the new API, like:
- CbdeHandler_CS_FromCSharp9
- CookiesShouldBeHttpOnly_Net
- CookiesShouldBeSecure_Net
- XmlExternalEntityShouldNotBeParsed_XmlReader_CSharp9

And code fixes:
- CheckFileLicenseCodeFix_CSharp9_ShouldBeNoncompliant_CS
- ConditionalSimplification_FromCSharp9_CodeFix
- EmptyStatement_CodeFix_CSharp9
- UseShortCircuitingOperators_CSharp9_CodeFix
- VirtualEventField_CSharp9_CodeFix

costin-zaharia-sonarsource:
> There are also tests which use ParseOptionsHelper.FromCSharp9

Yes! I forgot to search for `ParseOptionsHelper.FromCSharp9` usages. 

For code fixes we don't have methods specific for C#9 and since there are just a few calls I prefer to let them this way.

andrei-epure-sonarsource (approved):
LGTM!

PR #3775, opened 2020-11-19
Update rules.xml generation to include status

### LINKED ISSUE #3287: Fix RuleDescriptorGenerator: the generated `rules.xml` does not contain the rule STATUS
For details see [this internal thread](https://discuss.sonarsource.com/t/deprecated-rules-do-not-have-synced-status-in-sonarcloud/4270/6).

We upload the rules metadata using a `rules.xml` file. A rule has the STATUS attribute which, by default, is READY, but it can also be DEPRECATED.

The rules definition are loaded in [AbstractRulesDefinition #define](https://github.com/SonarSource/sonar-dotnet/blob/8.4.0.15306/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractRulesDefinition.java#L60) using the XML file /org/sonar/plugins/csharp/rules.xml

This file gets generated during the build in e.g. `C:\Workspace\sonar-dotnet\sonaranalyzer-dotnet\src\SonarAnalyzer.RuleDescriptorGenerator\bin\Debug\net46\cs`

We need to improve the RuleDescriptorGenerator to include the rule status.

costin-zaharia-sonarsource:
Fixes #3287

### DIFF
--- a/sonar-csharp-plugin/src/test/java/org/sonar/plugins/csharp/CSharpSonarRulesDefinitionTest.java
+++ b/sonar-csharp-plugin/src/test/java/org/sonar/plugins/csharp/CSharpSonarRulesDefinitionTest.java
@@ -84,4 +84,16 @@ public class CSharpSonarRulesDefinitionTest {
     assertThat(rule.type()).isEqualTo(RuleType.VULNERABILITY);
     assertThat(rule.securityStandards()).containsExactlyInAnyOrder("cwe:326", "owaspTop10:a3", "owaspTop10:a6");
   }
+
+  @Test
+  public void test_all_rules_have_status_set(){
+    CSharpSonarRulesDefinition definition = new CSharpSonarRulesDefinition();
+    RulesDefinition.Context context = new RulesDefinition.Context();
+    definition.define(context);
+    RulesDefinition.Repository repository = context.repository("csharpsquid");
+
+    for (RulesDefinition.Rule rule:repository.rules()) {
+      assertThat(rule.status()).isNotNull();
+    }
+  }
 }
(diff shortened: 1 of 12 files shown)


andrei-epure-sonarsource (changes requested):
don't we have any tests on the java plugin on this, related to `AbstractRulesDefinition` ?

andrei-epure-sonarsource (approved):
LGTM!

sonar-csharp-plugin/src/test/java/org/sonar/plugins/csharp/CSharpSonarRulesDefinitionTest.java
```diff
@@ -84,4 +84,16 @@ public void test_security_standards_with_vulnerability() {
...
+
+  @Test
+  public void test_all_rules_have_status_set(){
+    CSharpSonarRulesDefinition definition = new CSharpSonarRulesDefinition();
+    RulesDefinition.Context context = new RulesDefinition.Context();
+    definition.define(context);
+    RulesDefinition.Repository repository = context.repository("csharpsquid");
+
+    for (RulesDefinition.Rule rule:repository.rules()) {
+      assertThat(rule.status()).isNotNull();
```
andrei-epure-sonarsource, line 96: another idea is to check that the status is one of the allowed values for SQ, to make sure it won't fail at startup

costin-zaharia-sonarsource: I've initially thought at this and found out that the same test will hard fail if the status from `rules.xml` is not one of the enum values.

PR #3779, opened 2020-11-20
Update rspec before release

costin-zaharia-sonarsource:


andrei-epure-sonarsource (approved):
LGTM (I looked out of curiosity)

pavel-mikula-sonarsource (commented):
S1313 has a description update that we can use. While skipping the "Exceptions" section if not implemented.

pavel-mikula-sonarsource (commented):
What's the situation with S2228 deprecation? Rule api generates note that's not a part of this PR

pavel-mikula-sonarsource (commented):
S3925 (that's not part of this PR) seems to have some formatting issue in JIRA and rule-api generates bad list. Can you fix the specification?

pavel-mikula-sonarsource (changes requested):
~~VB.NET / S1821 description should be updated. Rule is implemented, I've fixed missing examples in the RSPEC recently~~

I take this one back, it's there. I somehow overlooked it.

pavel-mikula-sonarsource (approved):
LGTM. I left 3 comments/questions above to clarify situation.

costin-zaharia-sonarsource:
> What's the situation with S2228 deprecation? Rule api generates note that's not a part of this PR

We clarified this offline, the S2228 is already updated on master by https://github.com/SonarSource/sonar-dotnet/pull/3775

PR #3793, opened 2020-11-30
Rename directory sonaranalyzer-dotnet to analyzers

pavel-mikula-sonarsource:
We were having a pain with `sonaranalyzer-dotnet` directory.
* It's hard to quickly distinguish between the .NET part and Java part of the project
* It's sorted differently in GitHub vs. Windows dir, making the navigation harder to dispatch to muscle memory

Solution: We'll rename it to `analyzers`


pavel-mikula-sonarsource:
I had to split the ITs into several steps, because they contain manually added files, that are ignored by `.gitignore` by default.
And it was not possible to manually add them all at once to preserve history correctly.

Commit `MOve its/sources/ember-mm step 3 - remove other .gitignored DLLs` removes DLLs that were originally added, but should not be part of the project according to it's own `ember-mm/.gitignore`

pavel-mikula-sonarsource:
Commit `Nancy: fix CRLF issue` contains change from CRLF to LF as an isolated commit. Git wasn't happy with rename and this change together and the history was hard to follow otherwise.

Corniel:
Good luck with this one. Seems like a hard thing to do. :)

andrei-epure-sonarsource (approved):
LGTM

(for the record, I manually checked the VSIX project still loads)

PR #3794, opened 2020-12-01
Fix CodeSmells with type Bug

pavel-mikula-sonarsource:


pavel-mikula-sonarsource:
I'll need another branch.

PR #3795, opened 2020-12-01
Fix CodeSmells with type Bug

pavel-mikula-sonarsource:


andrei-epure-sonarsource:
It's weird that I don't see these on the [Overall Code Period](https://sonarcloud.io/summary/overall?id=sonaranalyzer-dotnet)

image

andrei-epure-sonarsource (approved):
Great! LGTM

PR #3798, opened 2020-12-02
GitHub workflow: Don't run Request review for external PRs

pavel-mikula-sonarsource:
This fixes issue that appeared when the action was triggered for PR from external fork.

### DIFF
--- a/.github/workflows/RequestReview.yml
+++ b/.github/workflows/RequestReview.yml
@@ -8,6 +8,8 @@ jobs:
   assign_and_move_card:
     name: Assign issue to reviewer and move Kanban card
     runs-on: ubuntu-latest
+    # PRs from forks don't have required token authorization
+    if: github.event.pull_request.head.repo.full_name == github.repository
 
     steps:
       # https://github.com/actions/github-script
--- a/.github/workflows/RequestReview.yml
+++ b/.github/workflows/RequestReview.yml
@@ -73,23 +73,26 @@ jobs:
             }
             //
             async function processIssue(issue) {
-                removeAssignees(issue);
-                addAssignee(issue, context.payload.requested_reviewer.login);
-                const card = await findCard(issue.url);
-                if (card) {
-                    console.log("Moving card");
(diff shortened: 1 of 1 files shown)


.github/workflows/RequestReview.yml
```diff
@@ -73,23 +75,26 @@ jobs:
...
-                    console.log("Moving card");
-                    github.projects.moveCard({ card_id: card.id, position: "bottom", column_id: REVIEW_IN_PROGRESS_COLUMN });
-                } else if (issue.pull_request) {
-                    console.log("Creating PR card");
-                    github.projects.createCard({ column_id: REVIEW_IN_PROGRESS_COLUMN, content_id: context.payload.pull_request.id, content_type: "PullRequest" });
-                } else {
-                    console.log("Creating Issue card");
-                    github.projects.createCard({ column_id: REVIEW_IN_PROGRESS_COLUMN, content_id: issue.id, content_type: "Issue" });
+                if(issue.state == "open") {
+                  removeAssignees(issue);
```
andrei-epure-sonarsource, line 79: if there is a 2nd reviewer added, I believe we shouldn't reset+add, but add another face on the kanban, to reflect the people that should have a look

andrei-epure-sonarsource (changes requested):

.github/workflows/RequestReview.yml
pavel-mikula-sonarsource: Let's stick with baby steps here as well. Review state is very hard to read from API. I can't get list of reviewers with "yellow dot in UI".

pavel-mikula-sonarsource: I added a note about this idea to my Kanban backlog.

andrei-epure-sonarsource: sure, it's an edge case, if it's not easy to fix, we leave it for later

andrei-epure-sonarsource (approved):
LGTM

PR #3801, opened 2020-12-03
Set max line length to 200

costin-zaharia-sonarsource:


.editorconfig
```diff
@@ -27,6 +27,7 @@ indent_size = 4
 insert_final_newline = true
 trim_trailing_whitespace = true
 end_of_line = lf
+max_line_length = 200
```
pavel-mikula-sonarsource, line 30: In the [docs](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#max_line_length), it says
>Forces hard line wrapping ...

We don't really want that for some cases like `KnowTypes.cs` (even when it doesn't work like that, in some IDEs). And there's [some controversy](https://github.com/editorconfig/editorconfig/issues/387) around this rule. Is there some way to use rulers setting instead? I haven't found much except "editor.rulers" setting for VS Code, but that doesn't seem to be relevant for `.editorconfig`

Does it really worth adding?

pavel-mikula-sonarsource (changes requested):
Topic to discuss

.editorconfig
costin-zaharia-sonarsource: > Is there some way to use rulers setting instead?

As far as I can tell, we cannot use rulers as they are not implemented yet: https://github.com/editorconfig/editorconfig/issues/89

> Does it really worth adding?

I think yes, it's great to have a visual guideline for the max length. I suggest to try this for a while and if it doesn't fit well to our needs we can remove it anytime.

pavel-mikula-sonarsource (approved):
LGTM

.editorconfig
andrei-epure-sonarsource: did you test what happens when you open KnownTypes.cs? can it be disabled for specific files?

costin-zaharia-sonarsource: Behavior:
- on Rider, the IDE will display a nice guideline and will not do any changes by itself if the lines are not changed or added. When you finish to update a line (insert a `;` for e.g.), if the line does not respect the limit it will be split. This can be solved easily by undo which will undo jus the automated change. Of course the behavior can be very different depending on the IDE preferences (e.g. if you have clean up on save for e.g.)
- on Visual Studio: this setting seems to be ignored by the IDE for now. It looks that I will be the only one affected by the change 😄.

pavel-mikula-sonarsource: I did test it in VS while reviewing. Nothing harmful happened.

PR #3804, opened 2020-12-04
CFG generation throws StackOverlowException when syntax tree is too deep

### LINKED ISSUE #3630: CFG generation throws StackOverlowException when syntax tree is too deep
### Description

CFG generation can throw `StackOverlofException` that will break the build of analyzed project.

Project `Lucene.NET` has [a funny file](https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Analysis.Common/Analysis/CharFilter/HTMLStripCharFilter.cs) with constants initialized to 1800 string concatenations. When we generate CFG for such syntax node with a very deep structure, it fails on `StackOverflowException`. 

Same problem occurs for all `BinaryExpressionSyntax` (&&, ||, ...). Maximum depth varies on .NET Framework vs .NET Core build.
### Reproducer

Issue is reproducible with simple test:
```C#
[TestMethod]
public void CauseStackOverflow()
{
    const int count = 2000;
    const string dna = "ACGT";
    var concatenations = Enumerable.Repeat($@"""{dna}""", count).JoinStr(" +\n");
    var cfg = Build($"var x = {concatenations};");
    VerifyMinimalCfg(cfg);
    VerifyAllInstructions(cfg.EntryBlock);
}
```

Note: `StackOverflowException` cannot be caught. Failure of the test is visible in Test output window.
### Possible solutions

1. Use Roslyn CFG
2. Validate and limit nesting depth of supported CFG
3. Process BinaryExpressionSyntax in non-recursive way
4. Do both 2. and 3.

### After it's fixed

Once this is fixed, we need to:
* Create issue on SONARSEC side to update `SonarAnalyzer.CFG.dll` version to the latest released
* Remove `SupportedNodeCountLimit` from Sonar-Securify C# FE `ConstructorUcfgBuilder.ProcessInitializerValue`

pavel-mikula-sonarsource:
Fixes #3630

### DIFF
--- a/analyzers/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/analyzers/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -31,6 +31,8 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
 {
     public sealed class CSharpControlFlowGraphBuilder : AbstractControlFlowGraphBuilder
     {
+        private const int SupportedExpressionNodeCountLimit = 500;
+
         private readonly Stack<Block> breakTarget = new Stack<Block>();
         private readonly Stack<Block> continueTargets = new Stack<Block>();
         private readonly Stack<Dictionary<object, List<JumpBlock>>> switchGotoJumpBlocks = new Stack<Dictionary<object, List<JumpBlock>>>();
@@ -217,6 +219,10 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
             if (expression == null)
             {
                 return currentBlock;
+            }
+            if (!IsTooComplex(expression))
+            {
+                throw new NotSupportedException("Too complex expression");
             }
 
             switch (expression.Kind())
@@ -526,6 +532,12 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
             }
 
             throw new NotSupportedException($"{patternSyntaxWrapper.SyntaxNode.Kind()}");
+        }
+
+        private static bool IsTooComplex(SyntaxNode node)
+        {
+            var count = 0;  // Limit descending for performance reasons
+            return node.DescendantNodes(x => ++count < SupportedExpressionNodeCountLimit).Count() < SupportedExpressionNodeCountLimit;
         }
 
         #endregion Top level Build*
--- a/analyzers/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/analyzers/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -220,7 +220,7 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
             {
                 return currentBlock;
             }
-            if (!IsTooComplex(expression))
+            if (IsTooComplex(expression))
             {
                 throw new NotSupportedException("Too complex expression");
             }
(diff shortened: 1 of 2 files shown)


analyzers/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -528,6 +534,12 @@ private Block BuildPatternExpression(PatternSyntaxWrapper patternSyntaxWrapper,
             throw new NotSupportedException($"{patternSyntaxWrapper.SyntaxNode.Kind()}");
         }
 
+        private static bool IsTooComplex(SyntaxNode node)
```
andrei-epure-sonarsource, line 537: given you only use it in `BuildExpression`, can't this be a static local function?

(this class already has lots of methods - and you group this private static method inside the region `Top level Build` - you may as well keep it as a local function)

analyzers/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -177,6 +173,52 @@ public class Bar
             VerifyAllInstructions(exit);
         }
 
+        [TestMethod]
+        [TestCategory("CFG")]
+        public void Cfg_ExtremelyNestedExpression_NotSupported_FromExpression()
```
andrei-epure-sonarsource, line 178: both `Cfg_ExtremelyNestedExpression_NotSupported_FromExpression` and `Cfg_ExtremelyNestedExpression_NotSupported_FromBodyMethod` are from body methods...

You could rename 
```suggestion
        public void Cfg_ExtremelyNestedExpression_NotSupported_FromAssignment()
```

analyzers/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -177,6 +173,52 @@ public class Bar
...
+            var method = CompileWithMethodBody(string.Format(TestInput, $"var x = {ExtremelyNestedExpression()};"), "Bar", out var semanticModel);
+            Action a = () => CSharpControlFlowGraph.Create(method.Body, semanticModel);
+
+            a.Should().Throw<NotSupportedException>().WithMessage("Too complex expression");
+            CSharpControlFlowGraph.TryGet(method.Body, semanticModel, out _).Should().BeFalse();
+        }
+
+        [TestMethod]
+        [TestCategory("CFG")]
+        public void Cfg_ExtremelyNestedExpression_NotSupported_FromBodyMethod()
```
andrei-epure-sonarsource, line 189:
```suggestion
        public void Cfg_ExtremelyNestedExpression_NotSupported_FromBodyMethodReturn()
```

andrei-epure-sonarsource (changes requested):
Do you think it's worth adding an integration test as well? (because we don't test what's the end result when throwing the exception)

analyzers/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
pavel-mikula-sonarsource: Method `BuildExpression` itself has almost 300 LOC. I don't want to add local function there.

analyzers/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
pavel-mikula-sonarsource: Good catch! I actually wanted to check `FromExpression`. My intention is to cover all branches here:
https://github.com/SonarSource/sonar-dotnet/blob/84c8cde333395a1f13e27f95832b9e1937c8721c/analyzers/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs#L79-L90

analyzers/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
pavel-mikula-sonarsource: `FromArrowMethod` and `FromBodyMethod` makes a symmetic pair in the list.

pavel-mikula-sonarsource:
I don't think adding integration test is needed. The UTs are testing the `TryGet` method to return `null`.

andrei-epure-sonarsource (approved):
LGTM

PR #3806, opened 2020-12-04
GitHub workflow: Start progress

pavel-mikula-sonarsource:
Assign to sender when card is moved from `To do` to `In progress` column on the Best Kanban

### DIFF
--- a/.github/workflows/StartProgress.yml
+++ b/.github/workflows/StartProgress.yml
@@ -35,5 +35,3 @@ jobs:
                 addAssignee(issue, context.payload.sender.login);
             }
             console.log(`Done`);
-
-
--- a/.github/workflows/StartProgress.yml
+++ b/.github/workflows/StartProgress.yml
@@ -1,0 +1,39 @@
+name: Start Progress
+
+on:
+  project_card:
+    types: ["moved"]
+
+jobs:
+  assign_to_self:
+    runs-on: ubuntu-latest
+    if: |
+        github.event.changes.column_id.from == 4971951
+        && github.event.project_card.column_id == 4971952
+        && github.event.project_card.content_url != null
+    steps:
+      # https://github.com/actions/github-script
+      - uses: actions/github-script@v3
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            //
+            async function addAssignee(issue, login) {
+                console.log("Assigning to: " + login);
+                await github.issues.addAssignees({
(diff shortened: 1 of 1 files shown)


.github/workflows/StartProgress.yml
```diff
@@ -0,0 +1,39 @@
...
+
+on:
+  project_card:
+    types: ["moved"]
+
+jobs:
+  assign_to_self:
+    runs-on: ubuntu-latest
+    if: |
+        github.event.changes.column_id.from == 4971951
```
andrei-epure-sonarsource, line 11:
```suggestion
        github.event.changes.column_id.from == 4971951 # To Do
```

.github/workflows/StartProgress.yml
```diff
@@ -0,0 +1,39 @@
...
+        && github.event.project_card.column_id == 4971952
```
andrei-epure-sonarsource, line 12:
```suggestion
        && github.event.project_card.column_id == 4971952 # In Progress
```

.github/workflows/StartProgress.yml
```diff
@@ -0,0 +1,39 @@
...
+            //
+            const url = context.payload.project_card.content_url;
+            console.log("Processing " + url);
+            var issue = (await github.request(url)).data;
+            if(!issue.assignee){
+                addAssignee(issue, context.payload.sender.login);
+            }
+            console.log(`Done`);
+
+
```
andrei-epure-sonarsource, line 39:
```suggestion

```

andrei-epure-sonarsource (approved):
LGTM (I left some suggestions before merging)

.github/workflows/StartProgress.yml
pavel-mikula-sonarsource: It's not possible to add comments inside `if`. It doesn't work as `//` nor as `#`. See this run:

https://github.com/pavel-mikula-sonarsource/GitHubActionPlayground/actions/runs/400781201

.github/workflows/StartProgress.yml
pavel-mikula-sonarsource: Same here

.github/workflows/StartProgress.yml
pavel-mikula-sonarsource: For reference, run with `//` is one of the old/early versions: https://github.com/pavel-mikula-sonarsource/GitHubActionPlayground/actions/runs/293418742

PR #3807, opened 2020-12-04
Fix S1172 FP: Issue is raised when parameter is used in tuple creation

### LINKED ISSUE #3255: Fix S1172 FP: Issue is raised when parameter is used in tuple creation
As reported on this [community thread](https://community.sonarsource.com/t/c-s1172-parameters-not-recognized-in-value-tuples/22011/6)

### Description

```C#
public string UsedInTuple(string value) // Noncompliant FP
{
    var x = (value, 7);
    return x.value;
}
```

### Related information

* SonarC# 8.5

pavel-mikula-sonarsource:
Fixes #3255

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
@@ -1529,7 +1529,7 @@ public class Sample
         {
             var cfg = Build("var a = b ?? c;");
             VerifyCfg(cfg, 4);
-            var branchBlock = (BinaryBranchBlock) cfg.EntryBlock;
+            var branchBlock = (BinaryBranchBlock)cfg.EntryBlock;
             var blocks = cfg.Blocks.ToList();
             var bNullBlock = blocks[1];
             var assignmentBlock = blocks[2];
@@ -1553,7 +1553,7 @@ public class Sample
             var cfg = Build("a = a ?? c;");
             VerifyCfg(cfg, 4);
 
-            var branchBlock = (BinaryBranchBlock) cfg.EntryBlock;
+            var branchBlock = (BinaryBranchBlock)cfg.EntryBlock;
             var blocks = cfg.Blocks.ToList();
             var bNullBlock = blocks[1];
             var afterOp = blocks[2];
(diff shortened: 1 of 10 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -5219,6 +5219,40 @@ public void Cfg_New_TargetTyped()
 
         #endregion
 
+        #region "Tuples"
+
+        [TestMethod]
+        [TestCategory("CFG")]
+        public void Cfg_Tuple_Create()
```
andrei-epure-sonarsource, line 5226: it would also be nice to have a test with expressions in the tuple creation, e.g.
`var (a,b) = (Foo(), 1 + Bar());`

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
```diff
@@ -258,4 +258,36 @@ public void Init<T>(ref T field)
...
+        public void DoSomething(bool arg)
+        {
+            var trueValue = true;
+            var tuple = (arg, trueValue);
+
+            if (tuple.arg)
+            {
+            }
+
+            if (tuple.trueValue)    // FN
```
andrei-epure-sonarsource, line 273: nitpick - is there a reason to have so many spaces before the comment here and at line 288?

andrei-epure-sonarsource (changes requested):
Nice work! I suggest to add a test with more complex logic in tuple creation

andrei-epure-sonarsource:
(for some reason the GH action didn't trigger, moving the parent issue manually)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
pavel-mikula-sonarsource: VS tab stops are 4 spaces wide and the `)` ends on one, so next tab produces 4 spaces in this case.

I moved the line 288, as there were two tabs.

andrei-epure-sonarsource (approved):
LGTM

PR #3808, opened 2020-12-04
Update SonarLint.xml

pavel-mikula-sonarsource:
Update QP and SonarLint.xml according to current QP. 

Current version of SL uses slightly different format and path. I also removed it from `Directory.Build.props` to avoid the problem with the message about duplicate SonarLint.xml files.


andrei-epure-sonarsource (approved):
LGTM!

PR #3815, opened 2020-12-07
S1144 FP: tuple construction should be considered for member usage

### LINKED ISSUE #2333: S1144 FP: tuple construction should be considered for member usage
The following code flags [S1144](https://rules.sonarsource.com/csharp/RSPEC-1144) on the Name setter but when I remove it compilation fails with CS0200.

```
namespace S1144_False_Positive_3
{
    class Class
    {
        void M()
        {
            OtherClass oc = new OtherClass();
            (oc.Name, oc.Type) = ("A", "B");
        }

        class OtherClass
        {
            public string Name { get; set; }
            public string Type { get; set; }
        }
    }
}
```

Is this a false positive?

### Related information

* SonarC# Version : Latest
* Visual Studio Version: 15.9.8

pavel-mikula-sonarsource:
Fixes #2333

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
@@ -287,7 +287,7 @@ namespace SonarAnalyzer.Helpers
                     return AccessorAccess.Both;
                 }
             }
-            else if (topmostSyntax.Parent is ArgumentSyntax argument && IsAssignmentToTuple(argument))
+            else if (topmostSyntax.Parent is ArgumentSyntax argument && argument.IsAssignmentToTuple())
             {
                 return AccessorAccess.Set;
             }
@@ -302,14 +302,6 @@ namespace SonarAnalyzer.Helpers
                 return topmostSyntax.Parent.IsAnyKind(IncrementKinds) ? AccessorAccess.Both : AccessorAccess.Get;
             }
         }
-
-        //FIXME: Rebase and extract into an extension method
-        // (arg, b) = something
-        private static bool IsAssignmentToTuple(ArgumentSyntax argument) =>
-            argument.Parent is { } tupleExpression
-            && ShimLayer.CSharp.TupleExpressionSyntaxWrapper.IsInstance(tupleExpression)
-            && tupleExpression.Parent is AssignmentExpressionSyntax assignment
-            && assignment.Left == tupleExpression;
 
         private bool IsKnownIdentifier(SyntaxToken identifier) =>
             knownSymbolNames.Contains(identifier.ValueText);
--- a/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
@@ -287,7 +287,7 @@ namespace SonarAnalyzer.Helpers
                     return AccessorAccess.Both;
                 }
             }
-            else if (topmostSyntax.Parent is ArgumentSyntax argument && argument.IsAssignmentToTuple())
+            else if (topmostSyntax.Parent is ArgumentSyntax argument && argument.IsInTupleAssignmentTarget())
             {
                 return AccessorAccess.Set;
             }
--- a/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarAnalyzer for .NET
  * Copyright (C) 2015-2020 SonarSource SA
  * mailto: contact AT sonarsource DOT com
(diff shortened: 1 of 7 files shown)


analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -411,11 +411,17 @@ public static bool IsComment(this SyntaxTrivia trivia)
...
+                ? argumentSyntax.Select(x => x.Expression).ToImmutableArray()
+                : ImmutableArray<ExpressionSyntax>.Empty;
         }
+
+        // (arg, b) = something
+        public static bool IsAssignmentToTuple(this ArgumentSyntax argument) =>
+            argument.Parent is { } tupleExpression
+            && TupleExpressionSyntaxWrapper.IsInstance(tupleExpression)
+            && tupleExpression.Parent is AssignmentExpressionSyntax assignment
+            && assignment.Left == tupleExpression;
```
andrei-epure-sonarsource, line 424: I find it easier to follow if you write
```
&& tupleExpression == assignment.Left;
```
because we read left to rigth (and at this point, we're focused on the tuple)

analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -411,11 +411,17 @@ public static bool IsComment(this SyntaxTrivia trivia)
...
-                return argumentSyntax.Select(x => x.Expression).ToImmutableArray();
-            }
-            return ImmutableArray<ExpressionSyntax>.Empty;
+            return methodParameterLookup.TryGetSyntax(parameterName, out var argumentSyntax)
+                ? argumentSyntax.Select(x => x.Expression).ToImmutableArray()
+                : ImmutableArray<ExpressionSyntax>.Empty;
         }
+
+        // (arg, b) = something
+        public static bool IsAssignmentToTuple(this ArgumentSyntax argument) =>
```
andrei-epure-sonarsource, line 420: I think the name `IsAssignmentToTuple` can be improved (did so in #3810 as well, but didn't have a better idea).

Because when looking at `(a, b) = (c, d)` we have arguments (and tuples) on both sides of the assignment, so `argument.IsAssignmentToTuple` is hard to follow. My first thought is to read as "argument is assigned to tuple" (like in `(a,b) = argument`) which is not the case. 

I suggest 
- `IsTupleArgumentOnAssignmentLeftSide` (not satisfied with it either...)
- `IsInTupleArgumentOnLeftSideOfAssignment` which would align with another name `IsInNameOfArgument()`

analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
```diff
@@ -298,34 +273,34 @@ private void StorePropertyAccess(IPropertySymbol propertySymbol, AccessorAccess
...
                 else
                 {
                     // Prop += value --> get/set
                     return AccessorAccess.Both;
                 }
             }
-
-            // nameof(Prop) --> get/set
-            if (node.IsInNameOfArgument(GetSemanticModel(node)))
+            else if (topmostSyntax.Parent is ArgumentSyntax argument && argument.IsAssignmentToTuple())
```
andrei-epure-sonarsource, line 290: this is for `(a,b) = someValue`, which is indeed Set.
How about `someValue = (a,b)` which should be a Get - could you add a test for it?

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SyntaxHelperTest.cs
```diff
@@ -90,6 +91,47 @@ public void NameIs_VB()
...
+public class Sample
+{
+    public void Method()
+    {
+        int a, b;
+        (a, b) = (42, 42);
+    }
+}";
+            var argument = Parse_CS(code).OfType<CSharp.ArgumentSyntax>().First();
+            argument.IsAssignmentToTuple().Should().BeTrue();
```
andrei-epure-sonarsource, line 107: can you also add a test with 
```
int a = 1, b = 2;
var tuple = (a, b);
```
and verify `IsAssignmentToTuple()` returns false?

andrei-epure-sonarsource (changes requested):

andrei-epure-sonarsource:
Also, please have a look at the code smells

analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
pavel-mikula-sonarsource: Last 2 lines
```
            && tupleExpression.Parent is AssignmentExpressionSyntax assignment
            && assignment.Left == tupleExpression;
```
are inspecting the `assignment`. So to me it's like "it's assignment and the assignment matches the constraint". It also follow the logic of going deeper in the structure.

analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
pavel-mikula-sonarsource: `IsInTupleAssignmentTarget` as discussed offline

andrei-epure-sonarsource (approved):
LGTM
Please have a look at the code smells before merging.

PR #3817, opened 2020-12-07
Add support for Deterministic Source Paths

### LINKED ISSUE #3362: Deterministic Source Paths Breaks Coverage
### Description

If deterministic source paths are enabled during the builds, the scanner will fail to find the code coverage for the code.  (The coverage reports will be indexed, but it will say 0 files matched).  This happens by default with the Microsoft.SourceLink package.

### Repro steps

Create a dotnet core project with 1 class and 1 unit test project, add the Microsoft.SourceLink package, and then run the sonar scanner against it.  It will fail to produce code coverage.

### Known workarounds

Turn deterministic source paths off.

### Related information
* see https://github.com/dotnet/sourcelink/issues/615
* see https://github.com/dotnet/roslyn/issues/43476
* C#/VB.NET Plugins Version - 8.3.1
* Visual Studio Version - N/A
* MSBuild / dotnet version - Core 3.1
* If running through the Scanner for MSBuild, its version - dotnet-scanner 4.8
* Operating System - Win 10.

andrei-epure-sonarsource:
Fixes  #3362

### DIFF
--- a/its/projects/CoverageWithDeterministicSourcePaths/VisualStudio.coveragexml
+++ b/its/projects/CoverageWithDeterministicSourcePaths/VisualStudio.coveragexml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <results>
   <modules>
-    <module name="coveragewithdeterministicsourcepaths.tests.dll" path="coveragewithdeterministicsourcepaths.tests.dll" id="D56BC170BF52284F8137CB445BC6115201000000" block_coverage="100.00" line_coverage="100.00" blocks_covered="3" blocks_not_covered="0" lines_covered="1" lines_partially_covered="0" lines_not_covered="0">
+    <module name="coveragewithdeterministicsourcepaths.tests.dll" path="coveragewithdeterministicsourcepaths.tests.dll" id="323A4A5218B417449EC6A3D6F87D1B2A01000000" block_coverage="100.00" line_coverage="100.00" blocks_covered="3" blocks_not_covered="0" lines_covered="1" lines_partially_covered="0" lines_not_covered="0">
       <functions>
         <function id="8284" token="0x6000003" name="TestMethod1()" namespace="Repro3362.Tests" type_name="UnitTest1" block_coverage="100.00" line_coverage="100.00" blocks_covered="3" blocks_not_covered="0" lines_covered="1" lines_partially_covered="0" lines_not_covered="0">
           <ranges>
@@ -14,7 +14,7 @@
         </source_file>
       </source_files>
     </module>
-    <module name="coveragewithdeterministicsourcepaths.dll" path="coveragewithdeterministicsourcepaths.dll" id="DCBBF5335692704B877E036875DA445301000000" block_coverage="50.00" line_coverage="50.00" blocks_covered="1" blocks_not_covered="1" lines_covered="3" lines_partially_covered="0" lines_not_covered="3">
+    <module name="coveragewithdeterministicsourcepaths.dll" path="coveragewithdeterministicsourcepaths.dll" id="56A4C88902D07047AB06B82DBC63818001000000" block_coverage="50.00" line_coverage="50.00" blocks_covered="1" blocks_not_covered="1" lines_covered="3" lines_partially_covered="0" lines_not_covered="3">
       <functions>
         <function id="8272" token="0x6000001" name="Covered()" namespace="Repro3362" type_name="Foo" block_coverage="100.00" line_coverage="100.00" blocks_covered="1" blocks_not_covered="0" lines_covered="3" lines_partially_covered="0" lines_not_covered="0">
           <ranges>
@@ -32,7 +32,7 @@
         </function>
       </functions>
       <source_files>
-        <source_file id="0" path="/_/its/projects/CoverageWithDeterministicSourcePaths/CoverageWithDeterministicSourcePaths/Foo.cs" checksum_type="unknown">
+        <source_file id="0" path="/_/CoverageWithDeterministicSourcePaths/CoverageWithDeterministicSourcePaths/Foo.cs" checksum_type="unknown">
         </source_file>
       </source_files>
     </module>
--- a/its/projects/CoverageWithDeterministicSourcePaths/VisualStudio.coveragexml
+++ b/its/projects/CoverageWithDeterministicSourcePaths/VisualStudio.coveragexml
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<results>
+  <modules>
+    <module name="coveragewithdeterministicsourcepaths.tests.dll" path="coveragewithdeterministicsourcepaths.tests.dll" id="D56BC170BF52284F8137CB445BC6115201000000" block_coverage="100.00" line_coverage="100.00" blocks_covered="3" blocks_not_covered="0" lines_covered="1" lines_partially_covered="0" lines_not_covered="0">
+      <functions>
+        <function id="8284" token="0x6000003" name="TestMethod1()" namespace="Repro3362.Tests" type_name="UnitTest1" block_coverage="100.00" line_coverage="100.00" blocks_covered="3" blocks_not_covered="0" lines_covered="1" lines_partially_covered="0" lines_not_covered="0">
+          <ranges>
+            <range source_id="0" covered="yes" start_line="9" start_column="38" end_line="9" end_column="57" />
+          </ranges>
+        </function>
+      </functions>
+      <source_files>
+        <source_file id="0" path="C:\Workspace\sonar-dotnet\its\projects\CoverageWithDeterministicSourcePaths\CoverageWithDeterministicSourcePaths.Tests\FooTests.cs" checksum_type="unknown">
+        </source_file>
+      </source_files>
+    </module>
+    <module name="coveragewithdeterministicsourcepaths.dll" path="coveragewithdeterministicsourcepaths.dll" id="DCBBF5335692704B877E036875DA445301000000" block_coverage="50.00" line_coverage="50.00" blocks_covered="1" blocks_not_covered="1" lines_covered="3" lines_partially_covered="0" lines_not_covered="3">
+      <functions>
+        <function id="8272" token="0x6000001" name="Covered()" namespace="Repro3362" type_name="Foo" block_coverage="100.00" line_coverage="100.00" blocks_covered="1" blocks_not_covered="0" lines_covered="3" lines_partially_covered="0" lines_not_covered="0">
+          <ranges>
+            <range source_id="0" covered="yes" start_line="6" start_column="9" end_line="6" end_column="10" />
+            <range source_id="0" covered="yes" start_line="7" start_column="13" end_line="7" end_column="27" />
+            <range source_id="0" covered="yes" start_line="8" start_column="9" end_line="8" end_column="10" />
+          </ranges>
+        </function>
+        <function id="8288" token="0x6000002" name="NotCovered()" namespace="Repro3362" type_name="Foo" block_coverage="0.00" line_coverage="0.00" blocks_covered="0" blocks_not_covered="1" lines_covered="0" lines_partially_covered="0" lines_not_covered="3">
+          <ranges>
+            <range source_id="0" covered="no" start_line="11" start_column="9" end_line="11" end_column="10" />
+            <range source_id="0" covered="no" start_line="12" start_column="13" end_line="12" end_column="27" />
+            <range source_id="0" covered="no" start_line="13" start_column="9" end_line="13" end_column="10" />
+          </ranges>
+        </function>
+      </functions>
+      <source_files>
+        <source_file id="0" path="/_/its/projects/CoverageWithDeterministicSourcePaths/CoverageWithDeterministicSourcePaths/Foo.cs" checksum_type="unknown">
+        </source_file>
+      </source_files>
+    </module>
+  </modules>
+</results>
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/NCover3ReportParser.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/NCover3ReportParser.java
@@ -35,11 +35,11 @@ public class NCover3ReportParser implements CoverageParser {
 
   private static final String EXCLUDED_ID = "0";
   private static final Logger LOG = Loggers.get(NCover3ReportParser.class);
-  private final FileService coverageFileValidator;
+  private final FileService fileService;
   private final AnalysisWarnings analysisWarnings;
 
-  NCover3ReportParser(FileService validator, AnalysisWarnings analysisWarnings) {
-    this.coverageFileValidator = validator;
+  NCover3ReportParser(FileService fileService, AnalysisWarnings analysisWarnings) {
+    this.fileService = fileService;
     this.analysisWarnings = analysisWarnings;
   }
 
@@ -119,7 +119,7 @@ public class NCover3ReportParser implements CoverageParser {
 
       if (documents.containsKey(doc) && !isExcludedLine(line)) {
         String path = documents.get(doc);
-        if (coverageFileValidator.isSupportedAbsolute(path)) {
+        if (fileService.isSupportedAbsolute(path)) {
 
           LOG.trace("Found coverage for line '{}', vc '{}' when analyzing the doc '{}' with the path '{}'.",
             line, vc, doc, path);
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/NCover3ReportParser.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/NCover3ReportParser.java
@@ -35,10 +35,10 @@ public class NCover3ReportParser implements CoverageParser {
 
   private static final String EXCLUDED_ID = "0";
   private static final Logger LOG = Loggers.get(NCover3ReportParser.class);
-  private final CoverageFileValidator coverageFileValidator;
+  private final FileService coverageFileValidator;
   private final AnalysisWarnings analysisWarnings;
 
-  NCover3ReportParser(CoverageFileValidator validator, AnalysisWarnings analysisWarnings) {
+  NCover3ReportParser(FileService validator, AnalysisWarnings analysisWarnings) {
     this.coverageFileValidator = validator;
     this.analysisWarnings = analysisWarnings;
   }
@@ -119,7 +119,7 @@ public class NCover3ReportParser implements CoverageParser {
 
       if (documents.containsKey(doc) && !isExcludedLine(line)) {
         String path = documents.get(doc);
-        if (coverageFileValidator.isSupported(path)) {
+        if (coverageFileValidator.isSupportedAbsolute(path)) {
 
           LOG.trace("Found coverage for line '{}', vc '{}' when analyzing the doc '{}' with the path '{}'.",
             line, vc, doc, path);
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/NCover3ReportParser.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/NCover3ReportParser.java
@@ -23,7 +23,6 @@ import java.io.File;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.function.Predicate;
 import org.sonar.api.notifications.AnalysisWarnings;
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
@@ -36,11 +35,11 @@ public class NCover3ReportParser implements CoverageParser {
 
   private static final String EXCLUDED_ID = "0";
   private static final Logger LOG = Loggers.get(NCover3ReportParser.class);
-  private final Predicate<String> isSupported;
+  private final CoverageFileValidator coverageFileValidator;
   private final AnalysisWarnings analysisWarnings;
 
-  NCover3ReportParser(Predicate<String> isSupported, AnalysisWarnings analysisWarnings) {
-    this.isSupported = isSupported;
+  NCover3ReportParser(CoverageFileValidator validator, AnalysisWarnings analysisWarnings) {
+    this.coverageFileValidator = validator;
     this.analysisWarnings = analysisWarnings;
   }
 
@@ -120,7 +119,7 @@ public class NCover3ReportParser implements CoverageParser {
 
       if (documents.containsKey(doc) && !isExcludedLine(line)) {
         String path = documents.get(doc);
-        if (isSupported.test(path)) {
+        if (coverageFileValidator.isSupported(path)) {
 
           LOG.trace("Found coverage for line '{}', vc '{}' when analyzing the doc '{}' with the path '{}'.",
             line, vc, doc, path);
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
@@ -29,13 +29,13 @@ import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 
 @ScannerSide
-public class FileSystemCoverageFileValidator implements FileService {
-  private static final Logger LOG = Loggers.get(FileSystemCoverageFileValidator.class);
+public class ScannerFileService implements FileService {
+  private static final Logger LOG = Loggers.get(ScannerFileService.class);
   private static final Pattern DETERMINISTIC_SOURCE_PATH_PREFIX = Pattern.compile("^(([a-zA-Z]:)?[\\\\][_][\\\\])|([\\/][_][\\/])");
   private FileSystem fileSystem;
   private String languageKey;
 
-  public FileSystemCoverageFileValidator(String languageKey, FileSystem fileSystem) {
+  public ScannerFileService(String languageKey, FileSystem fileSystem) {
     this.languageKey = languageKey;
     this.fileSystem = fileSystem;
   }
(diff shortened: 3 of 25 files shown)


andrei-epure-sonarsource:
@pavel-mikula-sonarsource , if you want, you can have an initial look. Currently, this implementation is not working (at least in the ITs - which is an artificial use case, as the coverage file is checked in and the code runs under a temporary folder - see the details in 6969a9a commit message).

Left things:
- have a working integration test
- do manual testing on Windows and Linux
- add an integration test with `_`, `_1` and `_2` placeholders inside the deterministic source paths (which won't work, as this PR is an initial baby step)
- add a setting to enable this feature - there are a few things that could go wrong, like CI systems where `_` may be the starting folder (hard to predict) - I would rather have this feature require an explicit opt-in , get feedback , and avoid live site incidents

andrei-epure-sonarsource:
@pavel-mikula-sonarsource I should still test this manually on Linux , and maybe with a more complex project than the one in the integration tests

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/ScannerFileServiceTest.java
```diff
@@ -0,0 +1,232 @@
...
+    when(fs.hasFiles(any())).thenReturn(result);
+    when(fs.predicates()).thenReturn(mock(FilePredicates.class));
+    return fs;
+  }
+
+  private InputFile mockInput(String path) {
+    return new TestInputFileBuilder("mod", path).setLanguage("cs").build();
+  }
+
+}
```
andrei-epure-sonarsource, line 232: add new line

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParserTest.java
```diff
@@ -38,7 +43,14 @@
...
   @Rule
   public ExpectedException thrown = ExpectedException.none();
-  private Predicate<String> alwaysTrue = s -> true;
+  private FileService alwaysTrue;
+
+  @Before
+  public void prepare() {
+    alwaysTrue = mock(FileService.class);
+    when(alwaysTrue.isSupportedAbsolute(anyString())).thenReturn(true);
+
```
andrei-epure-sonarsource, line 52: rm

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
```diff
@@ -0,0 +1,100 @@
...
+
+  public boolean isSupportedAbsolute(String absolutePath) {
+    FilePredicates fp = fileSystem.predicates();
+    return fileSystem.hasFiles(
+      fp.and(
+        fp.hasAbsolutePath(absolutePath),
+        fp.hasLanguage(languageKey)));
+  }
+
+  public Optional<InputFile> getFilesByRelativePath(String filePath) {
```
andrei-epure-sonarsource, line 52:
```suggestion
  public Optional<InputFile> getFileByRelativePath(String filePath) {
```

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
```diff
@@ -0,0 +1,100 @@
...
+import org.sonar.api.batch.fs.FileSystem;
+import org.sonar.api.batch.fs.InputFile;
+import org.sonar.api.scanner.ScannerSide;
+import org.sonar.api.utils.log.Logger;
+import org.sonar.api.utils.log.Loggers;
+
+@ScannerSide
+public class ScannerFileService implements FileService {
+  private static final Logger LOG = Loggers.get(ScannerFileService.class);
+  private static final Pattern DETERMINISTIC_SOURCE_PATH_PREFIX = Pattern.compile("^(([a-zA-Z]:)?[\\\\][_][\\\\]|([/][_][/]))");
```
andrei-epure-sonarsource, line 35: add comment to explain regex

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/ScannerFileServiceTest.java
```diff
@@ -0,0 +1,232 @@
...
+import org.sonar.api.batch.fs.internal.TestInputFileBuilder;
+import org.sonar.api.utils.log.LogTester;
+import org.sonar.api.utils.log.LoggerLevel;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class ScannerFileServiceTest {
```
andrei-epure-sonarsource, line 41: also add test with `_1`, `_2` in the deterministic path

andrei-epure-sonarsource:
I opened  #3832 to improve coverage statistics in the logging.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/ScannerFileServiceTest.java
```diff
@@ -0,0 +1,232 @@
...
+    FileSystem fs = mock(FileSystem.class);
+    FilePredicates filePredicates = mock(FilePredicates.class);
+
+    ArgumentCaptor<String> argumentCaptor = ArgumentCaptor.forClass(String.class);
+    when(filePredicates.hasAbsolutePath(argumentCaptor.capture())).thenReturn(mock(FilePredicate.class));
+    when(fs.predicates()).thenReturn(filePredicates);
+    when(fs.hasFiles(any())).thenReturn(true);
+
+    // act
+    ScannerFileService sut = new ScannerFileService("key", fs);
```
pavel-mikula-sonarsource, line 57: Not sure what `sut` stands for. Renaming artifact?
```suggestion
    ScannerFileService sfs = new ScannerFileService("key", fs);
```

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/ScannerFileServiceTest.java
```diff
@@ -0,0 +1,232 @@
...
+    assertThat(argumentCaptor.getValue()).isEqualTo(andMock);
+  }
+
+  @Test
+  public void getFilesByRelativePath_when_no_indexed_files_returns_empty() {
+    FileSystem fs = createFileSystemForInputFiles(Collections.emptyList());
+
+    // act
+    ScannerFileService sut = new ScannerFileService("key", fs);
+    Optional<InputFile> result = sut.getFilesByRelativePath("C:\\_\\some\\path\\file.cs");
```
pavel-mikula-sonarsource, line 119: Have you seen the `C:\_\` path on some generated coverage file? 
Do we know what happens when you build from `\\server\path\` ?

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/ScannerFileServiceTest.java
```diff
@@ -0,0 +1,232 @@
...
+
+  @Test
+  public void getFilesByRelativePath_when_multiple_indexed_files_match_returns_empty() {
+    FileSystem fs = createFileSystemForInputFiles(Arrays.asList(mockInput("root1/some/path/file.cs"), mockInput("root2/some/path/file.cs")));
+
+    ScannerFileService sut = new ScannerFileService("key", fs);
+    Optional<InputFile> result = sut.getFilesByRelativePath("/_/some/path/file.cs");
+
+    assertThat(result).isEmpty();
+    assertThat(logTester.logs(LoggerLevel.TRACE)).containsExactly(
```
pavel-mikula-sonarsource, line 149: This pair would be easier to read if the first one is placed as one long line.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/ScannerFileServiceTest.java
```diff
@@ -0,0 +1,232 @@
...
+  public void getFilesByRelativePath_when_single_indexed_files_match_returns_file() {
+    InputFile expectedResult = mockInput("root/some/path/file.cs");
+    FileSystem fs = createFileSystemForInputFiles(Arrays.asList(
+      mockInput("one"),
+      mockInput("two"),
+      expectedResult,
+      mockInput("four")));
+
+    ScannerFileService sut = new ScannerFileService("key", fs);
+    Optional<InputFile> result = sut.getFilesByRelativePath("/_/some/path/file.cs");
```
pavel-mikula-sonarsource, line 164: To demonstrate that it's not just the `root` single-level prefix, but any relative path.
```suggestion
    Optional<InputFile> result = sut.getFilesByRelativePath("/_/path/file.cs");
```

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/ScannerFileServiceTest.java
```diff
@@ -0,0 +1,232 @@
...
+    assertThat(result).contains(expectedResult);
+    assertThat(logTester.logs(LoggerLevel.TRACE)).hasSize(2);
+    assertThat(logTester.logs(LoggerLevel.TRACE).get(0))
+      .isEqualTo("It seems Deterministic Source Paths are used, replacing 'C:\\_\\some\\path\\file.cs' with 'some\\path\\file.cs'.");
+    assertThat(logTester.logs(LoggerLevel.TRACE).get(1))
+      .startsWith("Found indexed file ")
+      .endsWith("root/some/path/file.cs' for coverage entry 'some/path/file.cs'.");
+
+    result = sut.getFilesByRelativePath("D:\\_\\some\\path\\file.cs");
+    assertThat(result).contains(expectedResult);
```
pavel-mikula-sonarsource, line 190: `hasValue` calls `contains` directly, but this name seems more readable. The `getFileS` suggests that it's a collection that should contain a member. That's not what the test is doing (even when it goes away after rename)
```suggestion
    assertThat(result).hasValue(expectedResult);
```
Applies below as well.

pavel-mikula-sonarsource (changes requested):
I left some questions and improvement suggestions.

>add an integration test with _, _1 and _2 placeholders 

I'd add just UTs. In one of the comments, I suggested implementing those, as it's easy.

>add a setting to enable this feature

Can you demonstrate a case in UTs with what could go wrong? Having CI with `_` root will just do the relative-path magic and should end in the same place as the absolute path. Shouldn't it?

I'm in for keeping it as out of the box feature. Coverage has already some pain to configure.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/NCover3ReportParser.java
```diff
@@ -36,11 +35,11 @@
 
   private static final String EXCLUDED_ID = "0";
   private static final Logger LOG = Loggers.get(NCover3ReportParser.class);
-  private final Predicate<String> isSupported;
+  private final FileService coverageFileValidator;
   private final AnalysisWarnings analysisWarnings;
 
-  NCover3ReportParser(Predicate<String> isSupported, AnalysisWarnings analysisWarnings) {
-    this.isSupported = isSupported;
+  NCover3ReportParser(FileService validator, AnalysisWarnings analysisWarnings) {
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: indeed, it should be `fileService` rather than validator; the rename didn't do the job for variables and params

its/projects/CoverageWithDeterministicSourcePaths/VisualStudio.coveragexml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: It's a good question. However, the same question applies for the file that gets generated without Deterministic Source Paths.

It seems VS accepts only the binary file (e.g. `Andrei_ANDREI-WINDOWS_2020-12-11.15_57_21.coverage`) which gets generated during `dotnet test`.

We import the XML which needs another processing step `CodeCoverage.exe analyze`.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/ScannerFileServiceTest.java
andrei-epure-sonarsource: system under test - Martin Fowler abbreviation (see [Mocks Aren't Stubs](https://martinfowler.com/articles/mocksArentStubs.html)). Costin is using it as well, and given it's industry standard-ish (because Fowler is a guru), I think it's fine to have it in the codebase.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParserTest.java
```diff
@@ -194,18 +206,61 @@ public void valid_with_complex_test_case() throws Exception {
...
+        Assertions.entry(32, 0),
+        Assertions.entry(33, 0),
+        Assertions.entry(34, 0));
+
+    assertThat(logTester.logs(LoggerLevel.INFO).get(0)).startsWith("Parsing the Visual Studio coverage XML report ");
+    assertThat(logTester.logs(LoggerLevel.DEBUG).get(0)).startsWith("The current user dir is ");
+    assertThat(logTester.logs(LoggerLevel.DEBUG).get(1))
+      .startsWith("Found indexed file '/test/file/Calc.cs' for coverage entry '")
+      .endsWith("\\CalcMultiplyTest\\MultiplyTest.cs'.");
   }
```
andrei-epure-sonarsource, line 264: missing test with `deterministic_source_paths.coveragexml`

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/ScannerFileServiceTest.java
andrei-epure-sonarsource: No, I've only seen something like `/_/x/y/Foo.cs`, including on Windows systems

but when the input `path` is `/_/its/projects/CoverageWithDeterministicSourcePaths/CoverageWithDeterministicSourcePaths/Foo.cs`, [new File(path).getCanonicalPath()](https://github.com/SonarSource/sonar-dotnet/blob/6ef1c842c09232c065b477f79910effb8554523f/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParser.java#L111) will return `C:\_\its\projects\CoverageWithDeterministicSourcePaths\CoverageWithDeterministicSourcePaths\Foo.cs` 

So the drive letter is more for the behavior of the Java library call that we use.

andrei-epure-sonarsource: I've mentioned this behavior in [this comment](https://github.com/SonarSource/sonar-dotnet/issues/3362#issuecomment-738183321) as well

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  private static String getNormalizedRelativePath(String filePath) {
+    String relativePath = replaceDeterministicSourcePath(filePath);
+    return relativePath.replace('\\', '/');
+  }
+
+  private static String replaceDeterministicSourcePath(String filePath) {
+
+    Matcher matcher = DETERMINISTIC_SOURCE_PATH_PREFIX.matcher(filePath);
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: It's already a class static field... `private static final Pattern DETERMINISTIC_SOURCE_PATH_PREFIX = Pattern.compile("^(([a-zA-Z]:)?[\\\\][_][\\\\]|([/][_][/]))")`

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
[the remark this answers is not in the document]
andrei-epure-sonarsource: I've answered about `D:\_\f.cs` patterns in [this comment](https://github.com/SonarSource/sonar-dotnet/pull/3817#discussion_r540934572)

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/ScannerFileServiceTest.java
andrei-epure-sonarsource: I now see the logic is twisted:
- the deterministic source paths are unix style `/_/x/y/Foo.cs`
- then `getCanonicalPath` makes them Windows style on Windows systems
- then we transform it again to unix style to match the scanner normalized paths

From this point of view, it's better to work with the original path from the beginning instead of the canonicalPath, when the canonicalPath is not found as an absolute path by the scanner.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
```diff
@@ -0,0 +1,100 @@
...
+    } else if (count > 1) {
+      LOG.debug("Found {} indexed files for relative path '{}'. Will skip this coverage entry.", count, normalizedRelativePath);
+      return Optional.empty();
+    } else {
+      LOG.trace("Found indexed file '{}' for coverage entry '{}'.", foundFile.uri().getPath(), normalizedRelativePath);
+      return Optional.of(foundFile);
+    }
+  }
+
+  private static String getNormalizedRelativePath(String filePath) {
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: probably we need to log both paths for maximum debug info

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/VisualStudioCoverageXmlReportParserTest.java
```diff
@@ -38,7 +43,14 @@
 
   @Rule
   public ExpectedException thrown = ExpectedException.none();
-  private Predicate<String> alwaysTrue = s -> true;
+  private FileService alwaysTrue;
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: because in the case of VS coverage, when the `isAbsolutePath` is false, it will call `getFileByRelativePath`.
For the others, the expectation is not to call this.
I clarified these assumptions in the tests.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
```diff
@@ -0,0 +1,100 @@
...
+      String path = file.uri().getPath();
+      if (path.endsWith(normalizedRelativePath)) {
+        count++;
+        foundFile = file;
+      }
+    }
+    if (count == 0) {
+      LOG.trace("Did not find any indexed file for '{}'.", normalizedRelativePath);
+      return Optional.empty();
+    } else if (count > 1) {
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: good point, I will refactor to have the happy path first (and error handling after)

andrei-epure-sonarsource:
@pavel-mikula-sonarsource the new commits , addressing your comments, start with "rename method and fields; modify UT" 3add2dd

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
```diff
@@ -0,0 +1,89 @@
...
+
+  private static String getNormalizedRelativePath(String filePath) {
+    String relativePath = replaceDeterministicSourcePath(filePath);
+    return relativePath.replace('\\', '/');
+  }
+
+  private static String replaceDeterministicSourcePath(String filePath) {
+    Matcher matcher = DETERMINISTIC_SOURCE_PATH_PREFIX.matcher(filePath);
+    if (matcher.find())
+    {
```
pavel-mikula-sonarsource, line 82: Let's fix this code smell (Java code tainted by C# convention :) )
```suggestion
    if (matcher.find()) {
```

pavel-mikula-sonarsource (changes requested):
GitHub won't let me add note to existing comment. Please take a look at the note that I will write in a minute :)

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Change of plan: I'll rewrite this in a new comment to be up to date.

pavel-mikula-sonarsource (changes requested):
Rewritten comment

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/ScannerFileServiceTest.java
```diff
@@ -157,10 +154,10 @@ public void getFileByRelativePath_when_single_indexed_files_match_returns_file()
     Optional<InputFile> result = sut.getFileByRelativePath("/_/path/file.cs");
 
     assertThat(result).hasValue(expectedResult);
-    assertThat(logTester.logs(LoggerLevel.TRACE)).hasSize(2);
-    assertThat(logTester.logs(LoggerLevel.TRACE).get(1))
+    assertThat(logTester.logs(LoggerLevel.TRACE)).hasSize(1);
+    assertThat(logTester.logs(LoggerLevel.TRACE).get(0))
       .startsWith("Found indexed file ")
-      .endsWith("root/some/path/file.cs' for '/_/path/file.cs' (normalized to 'path/file.cs').");
+      .endsWith("sonar-dotnet/sonar-dotnet-shared-library/mod/root/some/path/file.cs' for '/_/path/file.cs' (normalized to 'path/file.cs').");
```
andrei-epure-sonarsource, line 160: this is relying on the behavior of `TestInputFileBuilder`, however I wanted to make sure the full path returned by the filesystem will be logged (and not some trimmed version)

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
```diff
@@ -51,39 +50,29 @@ public boolean isSupportedAbsolute(String absolutePath) {
...
+        if (path.endsWith(relativePath)) {
+          count++;
+          foundFile = file;
+        }
+      }
+      if (count == 1) {
+        LOG.trace("Found indexed file '{}' for '{}' (normalized to '{}').", foundFile.uri().getPath(), filePath, relativePath);
+        return Optional.of(foundFile);
+      } else {
+        LOG.debug("Found {} indexed files for '{}' (normalized to '{}'). Will skip this coverage entry. Verify sonar.sources in .sonarqube\\out\\sonar-project.properties.",
```
andrei-epure-sonarsource, line 70: `.sonarqube\\out\\sonar-project.properties` should use the local system path separator - I'll make a separate PR for that, as I want to improve the general logging

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
```diff
@@ -0,0 +1,78 @@
...
+      Iterable<InputFile> files = fileSystem.inputFiles(fileSystem.predicates().hasLanguage(languageKey));
+      int count = 0;
+      InputFile foundFile = null;
+      for (InputFile file : files) {
+        String path = file.uri().getPath();
+        if (path.endsWith(relativePath)) {
+          count++;
+          foundFile = file;
+        }
+      }
```
pavel-mikula-sonarsource, line 65: This new layout is great, as it's easy to follow. Now it's way more visible what's going.

_Open question:_ The filtering in the `for` is actually a manual way of implementing lambda. `predicate` actually in Java terms. So can we replace this section with something predicate based?

There exists `predicates().hasRelativePath()`. Does that one solve our issue (I'm not sure if it does the same)? If not, isn't it simpler to implement a predicate? Even in-lined one with the `new FilePredicate(){accept(...)}` java syntax for inline abstract class implementation. As it should be 3 LOC.

We would not need the `count` variable.
We would return to the structure I asked you to remove previously :D 
```
    FilePredicates fp = fileSystem.predicates();
    Iterable<InputFile> files = fileSystem.inputFiles(
      fp.and(
        fp.hasRelativePath(path),
        fp.hasLanguage(languageKey)
      ));
```

pavel-mikula-sonarsource (changes requested):
One last thing I didn't spot previously.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
andrei-epure-sonarsource: as previously mentioned during a standup on Friday, this doesn't work. This is why I ended up doing this logic.

andrei-epure-sonarsource: I read your suggestion better now, getting it back in progress

pavel-mikula-sonarsource: Copy from Slack, as I just found out:
The relative path predicate does this:
```
return path.equals(f.relativePath()); 
```
so it will not help us :disappointed:
As seen here: https://github.com/SonarSource/sonar-enterprise/blob/021961b86da874e4b733a566ce252[…]nar/api/batch/fs/internal/predicates/RelativePathPredicate.java
PathPatternPredicate could do the trick, but it doesn't worth the trouble with crafting the dynamic regex
So if you see easy way how to implement own predicate, let's do it. Otherwise, let's merge it as is.

pavel-mikula-sonarsource: PathPatternPredicate = predicates().matchesPathPattern()

pavel-mikula-sonarsource (approved):
LGTM. As discussed, let's refactor the predicate in another PR.

PR #3818, opened 2020-12-08
Fix S1172 FP: Parameter captured in a referenced local function

### LINKED ISSUE #3134: Fix S1172 FP: Parameter captured in a referenced local function
### Description
A S1172 warning is shown for a parameter captured by a local function.
This occurs only when the local function is returned from the method.

### Repro steps

```
private static Predicate<DateTime> GetFilter(DateTime dateTime)  
{  
        bool Filter(DateTime time) => time.Year == dateTime.Year;  
        return Filter;  
}  
```

### Expected behavior
No S1172 warning is shown for this case.

### Actual behavior
A S1172 warning is shown even though the parameter is required.

### Known workarounds
Suppress the warning, assign the parameter to a discard, or disabling the SonarLint extension.

### Related information

* SonarLint Version - 4.17.0.14702
* SonarC# Version - 4.17.0.14702
* Visual Studio Version -  2017 Enterprise 15.9.19
* MSBuild / dotnet version - (dotnet) 4.0.30319
* Operating System - Windows 10 Enterprise 1909

pavel-mikula-sonarsource:
Fixes #3134

Review per commit is suggested to follow LVA refactoring


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MethodParameterUnused.cs
```diff
@@ -426,10 +426,56 @@ object TupleArgument((string adress, bool state)? e)
...
+            Enumerable.Empty<object>().Where(IsTrue);
+
+            bool IsTrue(object x) => condition;
+        }
+
+        private void LocalFunctionReferencedBody(bool condition)
+        {
+            Enumerable.Empty<object>().Where(IsTrue);
+
+            bool IsTrue(object x) => condition;
```
andrei-epure-sonarsource, line 446: should this be 
```suggestion
            bool IsTrue(object x) { return condition; };
```

analyzers/src/SonarAnalyzer.CSharp/LiveVariableAnalysis/CSharp/CSharpLiveVariableAnalysis.cs
```diff
@@ -35,61 +34,62 @@ public sealed class CSharpLiveVariableAnalysis : AbstractLiveVariableAnalysis
...
-            assignedInBlock = new HashSet<ISymbol>(); // Kill (The set of variables that are assigned a value.)
-            usedInBlock = new HashSet<ISymbol>(); // Gen (The set of variables that are used before any assignment.)
-
-            var assignmentLhs = new HashSet<SyntaxNode>();
+            var ret = new State();
+            ProcessBlockInternal(block, ret);
+            return ret;
+        }
 
+        private void ProcessBlockInternal(Block block, State state)
```
andrei-epure-sonarsource, line 70: wouldn't it be better for this to return `State` as well (to avoid modifying the parameter)?

You could still leave the other methods as they are, as an initial baby step

andrei-epure-sonarsource (approved):
LGTM, I left only small comments

Nice refactoring!

analyzers/src/SonarAnalyzer.CSharp/LiveVariableAnalysis/CSharp/CSharpLiveVariableAnalysis.cs
pavel-mikula-sonarsource: It's part of recursive loop filling the values, so it's not easy to do just return without some heavy coping of values.

PR #3822, opened 2020-12-09
Fix S3956 FP: Don't raise on properties marked with XmlElement attribute

### LINKED ISSUE #3728: Fix S3956 FP: Don't raise on properties marked with XmlElement attribute
As reported on the [internal discuss](https://discuss.sonarsource.com/t/s3956-raises-issues-in-xmlserializer-context/6281) thread.

XmlSerializer (used mainly in Web Services) doesn't serialize `IList`/`IList<T>` types contrary to `DataContractSerializer` from WCF. MS recommends [this workaround](https://social.msdn.microsoft.com/Forums/en-US/aeff4155-ba84-488c-a400-90834f6c2c74/xmlserializer-and-ilistlttgt?forum=asmxandxml) to the problem.

We should not raise S3956 on properties marked with `XmlElementAttribute` to avoid FPs:
```
[XmlElement(Name="Orders")]
public List<Order> _DoNotUse_Orders
```

pavel-mikula-sonarsource:
Fixes #3728

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotExposeListT.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotExposeListT.cs
@@ -102,7 +102,6 @@ namespace SonarAnalyzer.Rules.CSharp
 
                     if (fieldSymbol != null
                         && fieldSymbol.IsPubliclyAccessible()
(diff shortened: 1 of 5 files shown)


andrei-epure-sonarsource (approved):
LGTM

PR #3825, opened 2020-12-10
S3168: Broaden detection of event handlers

### LINKED ISSUE #1181: Rule S3168: Broaden detection of event handlers
### Description

There are several frameworks which do not follow the standard signature for event handlers of what are conceptually events.

Here are 2 examples off the top of my head:
1. The Reactive Extensions' IObservable.Subscribe extension method that takes a delegate
2. Prism's EventAggregator and the region navigation callbacks (OnNavigatedTo, etc)

I believe that the approach of the current implementation of S3168 is far too restrictive and leads very easily to disabling this rule altogether.

May I suggest considering methods whose name starts with 'On' as event handlers, and therefore acceptable for async void?
(Of course it would have to be a bit less naïve, like matching with something along the lines of `On[^a-z]\w*` to avoid accepting `OnboardClient()`) 
That would force the user to reconsider using async void if such a prefix were not suitable for the method.

### Repro steps

1. Create a method `OnPersonRegistered` that handles an event raised by Prism's EventAggregator.
2. Make `OnPersonRegistered` await an asynchronous method (`Task.Yield()`, for instance) so that it requires a signature with async void.
3. Run analysis with rule S3168 activated

### Expected behavior

The method `OnPersonRegistered` being the handler of what is conceptually and by name (**On**Something) an event, the async void in its signature is unavoidable and should be acceptable.
Therefore, rule S3168 should not be considered violated.

### Actual behavior

The analyzer raises a violation for rule S3168.

### Related information

* SonarC# Version 6.8.1.4648
* Visual Studio Version 15.5.4

pavel-mikula-sonarsource:
Fixes #1181

### DIFF
--- a/analyzers/rspec/cs/S3168_c#.html
+++ b/analyzers/rspec/cs/S3168_c#.html
@@ -49,5 +49,5 @@ class HttpPrinter
 <h2>Exceptions</h2>
 <p>Event handlers, i.e. methods with two arguments, first one matching type <code>object</code> or name <code>sender</code> and the second being or
 inheriting from <code>EventArgs</code>, are ignored.</p>
-<p>Event handlers named as <code>OnSomething</code> are also ignored.</p>
+<p>Methods named as <code>OnSomething</code> are also ignored.</p>
 
--- a/analyzers/rspec/cs/S3168_c#.html
+++ b/analyzers/rspec/cs/S3168_c#.html
@@ -47,6 +47,7 @@ class HttpPrinter
 }
 </pre>
 <h2>Exceptions</h2>
-<p>Event handlers, i.e. methods with two arguments, first one matching <code>object sender</code> and the second being or inheriting from
-<code>EventArgs</code>, are ignored.</p>
+<p>Event handlers, i.e. methods with two arguments, first one matching type <code>object</code> or name <code>sender</code> and the second being or
+inheriting from <code>EventArgs</code>, are ignored.</p>
+<p>Event handlers named as <code>OnSomething</code> are also ignored.</p>
 
--- a/analyzers/src/SonarAnalyzer.Common/Helpers/KnownMethods.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Helpers/KnownMethods.cs
@@ -51,9 +51,9 @@ namespace SonarAnalyzer.Helpers
 
         public static bool IsObjectEquals(this IMethodSymbol methodSymbol) =>
             methodSymbol != null
-                && (methodSymbol.IsOverride || methodSymbol.IsInType(KnownType.System_Object))
                 && methodSymbol.MethodKind == MethodKind.Ordinary
                 && methodSymbol.Name == nameof(object.Equals)
+                && (methodSymbol.IsOverride || methodSymbol.IsInType(KnownType.System_Object))
                 && methodSymbol.Parameters.Length == 1
                 && methodSymbol.Parameters[0].Type.Is(KnownType.System_Object)
                 && methodSymbol.ReturnType.Is(KnownType.System_Boolean);
@@ -66,10 +66,10 @@ namespace SonarAnalyzer.Helpers
                 && methodSymbol.MethodKind == MethodKind.Ordinary
                 && methodSymbol.Name == nameof(object.Equals)
                 && methodSymbol.IsInType(KnownType.System_Object)
-                && HasCorrectArguments()
+                && HasCorrectParameters()
                 && methodSymbol.ReturnType.Is(KnownType.System_Boolean);
 
-            bool HasCorrectArguments() =>
+            bool HasCorrectParameters() =>
                 methodSymbol.Parameters.Length == 2
                 && methodSymbol.Parameters[0].Type.Is(KnownType.System_Object)
                 && methodSymbol.Parameters[1].Type.Is(KnownType.System_Object);
@@ -77,17 +77,17 @@ namespace SonarAnalyzer.Helpers
 
         public static bool IsObjectGetHashCode(this IMethodSymbol methodSymbol) =>
             methodSymbol != null
-            && (methodSymbol.IsOverride || methodSymbol.IsInType(KnownType.System_Object))
             && methodSymbol.MethodKind == MethodKind.Ordinary
             && methodSymbol.Name == nameof(object.GetHashCode)
+            && (methodSymbol.IsOverride || methodSymbol.IsInType(KnownType.System_Object))
             && methodSymbol.Parameters.Length == 0
             && methodSymbol.ReturnType.Is(KnownType.System_Int32);
 
         public static bool IsObjectToString(this IMethodSymbol methodSymbol) =>
             methodSymbol != null
-            && (methodSymbol.IsOverride || methodSymbol.IsInType(KnownType.System_Object))
             && methodSymbol.MethodKind == MethodKind.Ordinary
             && methodSymbol.Name == nameof(object.ToString)
+            && (methodSymbol.IsOverride || methodSymbol.IsInType(KnownType.System_Object))
             && methodSymbol.Parameters.Length == 0
             && methodSymbol.ReturnType.Is(KnownType.System_String);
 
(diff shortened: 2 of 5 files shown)


pavel-mikula-sonarsource:
Let's merge this one even with failed QG. It's coming from cleanup of old code.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/AsyncVoidMethod.cs
```diff
@@ -48,7 +48,7 @@ private async void EventHandlerCases_MyEvent(object sender, bool e)
             // Property changed implementation goes here
         }
 
-        private async void ArbreDesClefs_ItemInvoked(TreeView sender, TreeViewItemInvokedEventArgs args) { }    // Noncompliant FP
+        private async void ArbreDesClefs_ItemInvoked(TreeView sender, TreeViewItemInvokedEventArgs args) { }    // Compliant, doesn't inherit from EventArgs, but looks like it
```
andrei-epure-sonarsource, line 60: what if ` private async void ArbreDesClefs_ItemInvoked(TreeView somethingElse, TreeViewItemInvokedEventArgs arguments)` ? is using `sender` and `args` a pattern enforced by MS guidelines?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/AsyncVoidMethod.cs
```diff
@@ -33,6 +34,27 @@ private async void EventHandlerCases_MyEvent(object sender, bool e)
...
+        {
+            // Property changed implementation goes here
+        }
+
+        private async void ArbreDesClefs_ItemInvoked(TreeView sender, TreeViewItemInvokedEventArgs args) { }    // Noncompliant FP
+
+
+        // Substitute for reference to Xamarin.Forms, Windows.UI.Xaml.Controls
+        public class BindableObject { }
+        public class TreeView { }
```
andrei-epure-sonarsource, line 65: maybe add also the namespace `Windows.UI.Xaml.Controls` and test with more than TreeView or `TreeViewItemInvokedEventArgs` , and rely also if there are items in that namespace?

although it may be better to keep it simple with `EventArgs` suffix

analyzers/src/SonarAnalyzer.Common/Helpers/KnownMethods.cs
```diff
@@ -31,194 +31,162 @@ public static bool IsMainMethod(this IMethodSymbol methodSymbol)
...
-                methodSymbol.Parameters.Length == 0 &&
-                methodSymbol.ReturnType.Is(KnownType.System_String);
-        }
+            return methodSymbol != null
+                && !methodSymbol.IsOverride
+                && methodSymbol.IsStatic
+                && methodSymbol.MethodKind == MethodKind.Ordinary
+                && methodSymbol.Name == nameof(object.Equals)
+                && methodSymbol.IsInType(KnownType.System_Object)
+                && HasCorrectArguments()
```
andrei-epure-sonarsource, line 69:
```suggestion
                && HasCorrectParameters()
```

analyzers/src/SonarAnalyzer.Common/Helpers/KnownMethods.cs
```diff
@@ -31,194 +31,162 @@ public static bool IsMainMethod(this IMethodSymbol methodSymbol)
...
-                methodSymbol.Parameters[1].Type.Is(KnownType.System_Runtime_Serialization_StreamingContext) &&
-                methodSymbol.ReturnsVoid &&
-                (methodSymbol.Name == "GetObjectData" ||
-                methodSymbol.Name == explicitName);
+            return methodSymbol != null
+                && methodSymbol.Parameters.Length == 2
+                && methodSymbol.Parameters[0].Type.Is(KnownType.System_Runtime_Serialization_SerializationInfo)
+                && methodSymbol.Parameters[1].Type.Is(KnownType.System_Runtime_Serialization_StreamingContext)
+                && methodSymbol.ReturnsVoid
+                && (methodSymbol.Name == "GetObjectData" || methodSymbol.Name == explicitName);
```
andrei-epure-sonarsource, line 120: I think you can move name checks before the Parameters Type checks, right? 

and maybe the same for all name checks which are done after `Type.Is()`  (e.g. lines 109,178,182) - string comparisons should be cheaper than the type checks which eventually [do multiple comparisons](https://github.com/SonarSource/sonar-dotnet/blob/6ef1c842c09232c065b477f79910effb8554523f/analyzers/src/SonarAnalyzer.Common/Helpers/TypeHelper.cs#L60)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/AsyncVoidMethod.cs
```diff
@@ -35,15 +35,24 @@ private async void EventHandlerCases_MyEvent(object sender, bool e)
...
+        static async void On() { }      // Noncompliant
+        static async void Onboard() { } // Noncompliant
+        static async void ToX() { }     // Noncompliant
+        static async void OnX() { }
+
+        static async void OnŘád() { }
+        static async void Onřád() { }   // Noncompliant
+
+        static async void OnΘ() { }     // Compliant, Uppercase Theta
+        static async void Onθ() { }     // Noncompliant, Lowercase Theta
```
andrei-epure-sonarsource, line 52: also 

- `static async void On3People` 
- `static async void ONCAPS` 
- `static async void On_Underscore`
- `static async void onEvent`

analyzers/rspec/cs/S3168_c#.html
```diff
@@ -47,6 +47,7 @@ <h2>Compliant Solution</h2>
 }
 </pre>
 <h2>Exceptions</h2>
-<p>Event handlers, i.e. methods with two arguments, first one matching <code>object sender</code> and the second being or inheriting from
-<code>EventArgs</code>, are ignored.</p>
+<p>Event handlers, i.e. methods with two arguments, first one matching type <code>object</code> or name <code>sender</code> and the second being or
+inheriting from <code>EventArgs</code>, are ignored.</p>
+<p>Event handlers named as <code>OnSomething</code> are also ignored.</p>
```
andrei-epure-sonarsource, line 52: I would say "Methods named as `OnSomething` ...", right? We don't know for sure they're event handlers?

andrei-epure-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/AsyncVoidMethod.cs
pavel-mikula-sonarsource: Can you please clarify this question in more detail? I didn't understand it.
I'll update the comment a little, as the actual fix was about the sender args.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/AsyncVoidMethod.cs
pavel-mikula-sonarsource: There [A LOT](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls?view=winrt-19041) of types ending with `EventArgs`, so for Xamarin, it should be fine with the suffix. 

I wanted to stay Xamarin-reference-free to have an universal solution. With any class as a sender and any `WhateverNameEndingWithEventArgs` argument (as supported by the old implementation)

analyzers/src/SonarAnalyzer.Common/Helpers/KnownMethods.cs
```diff
@@ -66,58 +66,58 @@ public static bool IsStaticObjectEquals(this IMethodSymbol methodSymbol)
...
             && methodSymbol.Name == nameof(object.ToString)
+            && (methodSymbol.IsOverride || methodSymbol.IsInType(KnownType.System_Object))
             && methodSymbol.Parameters.Length == 0
             && methodSymbol.ReturnType.Is(KnownType.System_String);
 
         public static bool IsIDisposableDispose(this IMethodSymbol methodSymbol)
         {
             const string explicitName = "System.IDisposable.Dispose";
             return methodSymbol != null
+                && (methodSymbol.Name == nameof(IDisposable.Dispose) || methodSymbol.Name == explicitName)
```
andrei-epure-sonarsource, line 98: I'm not sure this change makes sense - checking a bool or int property should be less expensive than checking the name

(I guess you can make this change in a different PR if you want)

andrei-epure-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.Common/Helpers/KnownMethods.cs
andrei-epure-sonarsource: (discussed offline - the changes were made to keep a common checking pattern  - first name, then params)

Issue #3724, opened 2020-11-06
Rule S3329: Cipher Block Chaining IV's should be random and unique


pierre-loup-tristant-sonarsource:
https://github.com/SonarSource/security-expected-issues/tree/master/dotnet/NetCore3MVC/Controllers/RSPEC_3329

pavel-mikula-sonarsource:
This should be SE rule

PR #3830, opened 2020-12-11
Add handling for updating ByteArray and CryptographyIV constraints

costin-zaharia-sonarsource:
Related to #3724


pavel-mikula-sonarsource (approved):
LGTM. Can you add some fragment crafted from failed peach example to some rule UT to act as integration test?

csaba-sagi-sonarsource (approved):
LGTM

PR #3833, opened 2020-12-11
S2583 and S2589 FP: Pattern matching generates wrong bool constraint

### LINKED ISSUE #3110: S2583 FP: Pattern matching generates wrong bool constraint

### LINKED ISSUE #3123: S2589 FP with 'is' pattern
### Description

[S2589](https://rules.sonarsource.com/csharp/RSPEC-2589) 

When casting an object to `bool` with `is`, we can easily check if the value is true by chaining the two expressions, like below.

Value can be either true or false, so there should be no S2589 occurrence.

Probably when encountering the `is` expression we are not assigning the correct contraints.

### Repro steps

```
        public bool Foo(object x) => x is bool value && value; // Noncompliant FP S2589
                                                        ^^^^^

        
        public bool Foo(object x) => x is bool value && value == true; // Noncompliant FP S2589

```

### Known workarounds

None

### Related information

* SonarC# Version 8.3

pavel-mikula-sonarsource:
Fixes #3110
Fixes #3123

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
@@ -649,7 +649,7 @@ namespace SonarAnalyzer.SymbolicExecution
             {
                 // Push value for the discard, it will be popped when visiting the block for the
                 // corresponding case statement.
-                newProgramState = newProgramState.PushValue(SymbolicValue.Create());
+                newProgramState = newProgramState.PushValue(new SymbolicValue());
             }
             else if (SingleVariableDesignationSyntaxWrapper.IsInstance(variableDesignation))
             {
@@ -660,12 +660,12 @@ namespace SonarAnalyzer.SymbolicExecution
                 var variableSymbol = SemanticModel.GetDeclaredSymbol(singleVariableDesignation);
                 var newSymbolicValue = SymbolicValue.Create();
                 newProgramState = SetNewSymbolicValueIfTracked(variableSymbol, newSymbolicValue, newProgramState);
+                // When the pattern is "x is Type t" we know that "t != null", hence (SV != null)
+                newProgramState = newProgramState.SetConstraint(newSymbolicValue, ObjectConstraint.NotNull);
 
                 if (singleVariable)
                 {
-                    // When the pattern is "x is Type t" we know that "t != null", hence (SV != null)
-                    newProgramState = newProgramState.SetConstraint(newSymbolicValue, ObjectConstraint.NotNull);
-                    newProgramState = newProgramState.PushValue(newSymbolicValue);
+                    newProgramState = newProgramState.PushValue(new SymbolicValue());
                 }
             }
             else if (ParenthesizedVariableDesignationSyntaxWrapper.IsInstance(variableDesignation))
--- a/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
@@ -635,15 +635,15 @@ namespace SonarAnalyzer.SymbolicExecution
             //   - VarKeyword                                 - PredefinedType
             //   - SingleVariableDesignation                  - SingleVariableDesignation
             //      - IdentifierToken                             - IdentifierToken
-            VisitVariableDesignation(varPatternSyntax.Designation, programState, singleVariable: true);
+            VisitVariableDesignation(varPatternSyntax.Designation, programState);
 
         private ProgramState VisitDeclarationPattern(DeclarationPatternSyntaxWrapper declarationPattern, ProgramState newProgramState) =>
             // "x is string s" is equivalent to "s = x" and "s" gets NotNull constraint
             // "x is (string s, int i)" is recursive pattern with PositionalPatternClause and doesn't reach this path
-            VisitVariableDesignation(declarationPattern.Designation, newProgramState, singleVariable: true);
+            VisitVariableDesignation(declarationPattern.Designation, newProgramState);
 
-        /// <param name="singleVariable">True for top-level designated variables in "x is string s". NotNull constraint is added. False for variables nested in "var (string s, int i)".</param>
-        private ProgramState VisitVariableDesignation(VariableDesignationSyntaxWrapper variableDesignation, ProgramState programState, bool singleVariable)
+        /// <param name="isInParanthesizedVariableDesignation">True for variables nested in ParenthesizedVariableDesignationSyntax "var (string s, int i)".</param>
+        private ProgramState VisitVariableDesignation(VariableDesignationSyntaxWrapper variableDesignation, ProgramState programState, bool isInParanthesizedVariableDesignation = false)
         {
             var newProgramState = programState;
             if (DiscardDesignationSyntaxWrapper.IsInstance(variableDesignation))
@@ -662,7 +662,7 @@ namespace SonarAnalyzer.SymbolicExecution
                 var newSymbolicValue = SymbolicValue.Create();
                 newProgramState = SetNewSymbolicValueIfTracked(variableSymbol, newSymbolicValue, newProgramState);
 
-                if (singleVariable)
+                if (!isInParanthesizedVariableDesignation)
                 {
                     // When the pattern is "x is Type t" we know that "t != null", hence (SV != null)
                     newProgramState = newProgramState.SetConstraint(newSymbolicValue, ObjectConstraint.NotNull);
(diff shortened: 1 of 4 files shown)


analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
```diff
@@ -649,7 +649,7 @@ private ProgramState VisitVariableDesignation(VariableDesignationSyntaxWrapper v
             {
                 // Push value for the discard, it will be popped when visiting the block for the
                 // corresponding case statement.
-                newProgramState = newProgramState.PushValue(SymbolicValue.Create());
+                newProgramState = newProgramState.PushValue(new SymbolicValue());
```
pavel-mikula-sonarsource, line 652: This has no functional effect. It's just to use same API as below (because we don't need the logic inside `.Create()` here).

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
```diff
@@ -660,12 +660,12 @@ private ProgramState VisitVariableDesignation(VariableDesignationSyntaxWrapper v
                 var variableSymbol = SemanticModel.GetDeclaredSymbol(singleVariableDesignation);
                 var newSymbolicValue = SymbolicValue.Create();
                 newProgramState = SetNewSymbolicValueIfTracked(variableSymbol, newSymbolicValue, newProgramState);
+                // When the pattern is "x is Type t" we know that "t != null", hence (SV != null)
+                newProgramState = newProgramState.SetConstraint(newSymbolicValue, ObjectConstraint.NotNull);
```
pavel-mikula-sonarsource, line 664: This is just general improvement - it's safe to add this constraint every time we hit this path. Not only when `singleVariable` is true.

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
```diff
@@ -660,12 +660,12 @@ private ProgramState VisitVariableDesignation(VariableDesignationSyntaxWrapper v
...
 
                 if (singleVariable)
                 {
-                    // When the pattern is "x is Type t" we know that "t != null", hence (SV != null)
-                    newProgramState = newProgramState.SetConstraint(newSymbolicValue, ObjectConstraint.NotNull);
-                    newProgramState = newProgramState.PushValue(newSymbolicValue);
+                    newProgramState = newProgramState.PushValue(new SymbolicValue());
```
pavel-mikula-sonarsource, line 668: This is the actual fix. We need to push the result of `value is Something` witch is a bool. At this stage, it has no SE-logical relation to the designated variable `value is Something newVariable` from SymbolicValue point of view. 

SV for `newVariable` is a side effect of this designation, but doesn't effect the actual result of `value is Something` condition. So it should not be pushed (same as the discard pattern above)

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
```diff
@@ -660,12 +660,12 @@ private ProgramState VisitVariableDesignation(VariableDesignationSyntaxWrapper v
                 var variableSymbol = SemanticModel.GetDeclaredSymbol(singleVariableDesignation);
                 var newSymbolicValue = SymbolicValue.Create();
                 newProgramState = SetNewSymbolicValueIfTracked(variableSymbol, newSymbolicValue, newProgramState);
+                // When the pattern is "x is Type t" we know that "t != null", hence (SV != null)
+                newProgramState = newProgramState.SetConstraint(newSymbolicValue, ObjectConstraint.NotNull);
 
                 if (singleVariable)
```
andrei-epure-sonarsource, line 666: this is really confusing:
- I see `if (SingleVariableDesignationSyntaxWrapper.IsInstance(variableDesignation))`
AND
- `if (singleVariable)`

Uhm... could you rename `singleVariable` to something better? From what I understand, this makes the difference between `s1` when inside `x is string s1` vs `x is (string s1, string s2)` - however when it reaches line 668, both are `SingleVariableDesignationSyntaxWrapper`

rather than `isSingleVariable` I suggest `isInsideParanthesizedVariableDesignation`, which would reverse the current boolean checks & values - WDYT?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp7.cs
```diff
@@ -252,17 +252,24 @@ public int DoSomething(object value)
             if (!(value is bool valueBool))
```
andrei-epure-sonarsource, line 252: could you please also add a test with 
```
    if (x is (string s, int i))
    {
        if (s == null)  // Noncompliant
        {
        }
    }
```
inside `SonarAnalyzer.UnitTest\TestCases\ConditionEvaluatesToConstant.CSharp8.cs`

andrei-epure-sonarsource: and also
```
        if (x is (((string s1, (string  s2, String s3)), string s), int i))
        {
            if (s3 == null) // Noncompliant
            {
            }
        }
```

And maybe a test with generated code with inner tuples to see if it throws at some point? Or maybe that's a good test for the nested expression stress test?

andrei-epure-sonarsource (changes requested):
Nice! I've left a few comments for "clean as you code"

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
```diff
@@ -649,7 +649,7 @@ private ProgramState VisitVariableDesignation(VariableDesignationSyntaxWrapper v
             {
                 // Push value for the discard, it will be popped when visiting the block for the
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: It's little more complicated (see 1st review commit).
I also reversed the refactoring about setting `NotNull` constraint. I covered the path with test, so coverage is happy now (it was not covered before), but the var pattern doesn't inspect the body anyway, so the effect of the constraint is not visible. I'll see if I can fix it easily.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp7.cs
pavel-mikula-sonarsource: It's little more complicated. `x is(string s, int i)` is C# 7 and it's not `ParanthesizedVariableDesignation`. It's recursive pattern.

I covered this in first review commit. I'll add the recursive hell :)

pavel-mikula-sonarsource:
I suggest to do the 2nd review per commit to see the evolution.

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Update after last commit: The reverted refactoring is now covered with test that would fail.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp7.cs
andrei-epure-sonarsource: thanks!

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/ParseOptionsHelper.cs
```diff
@@ -97,7 +97,7 @@ language switch
         private static ImmutableArray<ParseOptions> FilterByEnvironment(this IEnumerable<ParseOptions> options) =>
             TestContextHelper.IsAzureDevOpsContext
                 ? options.ToImmutableArray()
-                : ImmutableArray.Create(options.Last()); // Use only the latest version for local test run and debug
+                : ImmutableArray.Create(options.First()); // Use only the oldest version for local test run and debug
```
andrei-epure-sonarsource, line 100: nice!

andrei-epure-sonarsource (approved):
LGTM!
Impressive , great! 🚀

PR #3834, opened 2020-12-11
Rule S3329: IV is safe by default

costin-zaharia-sonarsource:
Related to #3724

This updates the current implementation to consider the IV safe by default.

According to the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.symmetricalgorithm.createencryptor?view=net-5.0#System_Security_Cryptography_SymmetricAlgorithm_CreateEncryptor):
> If the current Key property is null, the GenerateKey method is called to create a new random Key. If the current IV property is null, the GenerateIV method is called to create a new random IV.

### DIFF
--- a/analyzers/rspec/cs/S3329_c#.html
+++ b/analyzers/rspec/cs/S3329_c#.html
@@ -15,24 +15,22 @@ twice will yield two different ciphertexts.</p>
 public void Encrypt(byte[] key, byte[] data, MemoryStream target)
 {
     byte[] initializationVector = new byte[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16 };
-    using (AesCryptoServiceProvider aes = new AesCryptoServiceProvider())
-    {
-        ICryptoTransform encryptor = aes.CreateEncryptor(key, initializationVector); // Noncompliant, hardcoded value is used
-        using (CryptoStream cryptoStream = new CryptoStream(target, encryptor, CryptoStreamMode.Write))
-        {
-            cryptoStream.Write(data);
-        }
-    }
+
+    using var aes = new AesCryptoServiceProvider();
+    var encryptor = aes.CreateEncryptor(key, initializationVector); // Noncompliant, hardcoded value is used
+
+    using var cryptoStream = new CryptoStream(target, encryptor, CryptoStreamMode.Write);
+    cryptoStream.Write(data);
 }
 </pre>
 <h2>Compliant Solution</h2>
 <pre>
 public byte[] Encrypt(byte[] key, byte[] data, MemoryStream target)
 {
-    using AesCryptoServiceProvider aes = new AesCryptoServiceProvider();
-    var encryptor = aes.CreateEncryptor(key, aes.IV);
+    using var aes = new AesCryptoServiceProvider();
+    var encryptor = aes.CreateEncryptor(key, aes.IV); // aes.IV is automatically generated to random secure value
 
-    using CryptoStream cryptoStream = new CryptoStream(target, encryptor, CryptoStreamMode.Write);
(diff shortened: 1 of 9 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/InitializationVectorShouldBeRandom.cs
```diff
@@ -183,9 +165,6 @@ argument.Expression switch
             private static bool HasNotInitializedIVConstraint(SymbolicValue value, ProgramState programState) =>
```
csaba-sagi-sonarsource, line 165: If we want to comply with SA1204StaticElementsMustAppearBeforeInstanceElements than the static elements should go before all instance elements.

pavel-mikula-sonarsource: We don't. We're currently dogfooding SA rules and I've raised this one to be [discussed/removed](https://sonarsource.slack.com/archives/C012KBFFYD6/p1607018101012200).

analyzers/rspec/cs/S3329_c#.html
```diff
@@ -29,16 +29,13 @@ <h2>Compliant Solution</h2>
...
-    {
-        aes.GenerateIV();
-        ICryptoTransform encryptor = aes.CreateEncryptor(key, aes.IV);
-        using (CryptoStream cryptoStream = new CryptoStream(target, encryptor, CryptoStreamMode.Write))
-        {
-            cryptoStream.Write(data);
-        }
-        return aes.IV;
-    }
+    using AesCryptoServiceProvider aes = new AesCryptoServiceProvider();
```
pavel-mikula-sonarsource, line 32: Just to keep it shorter
```suggestion
    using var aes = new AesCryptoServiceProvider();
```

analyzers/rspec/cs/S3329_c#.html
```diff
@@ -29,16 +29,13 @@ <h2>Compliant Solution</h2>
...
+    var encryptor = aes.CreateEncryptor(key, aes.IV);
```
pavel-mikula-sonarsource, line 33: Should we explain what's going on and why?
```suggestion
    var encryptor = aes.CreateEncryptor(key, aes.IV); // aes.IV is automatically generated to random secure value
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InitializationVectorShouldBeRandom.cs
```diff
@@ -270,4 +257,25 @@ public class CustomAes : Aes
...
+
+        public SymmetricalEncryptorWrapper()
+        {
+            algorithm = Aes.Create();
+        }
+
+
+        public void GenerateIV()
+        {
+            algorithm.GenerateIV();
```
pavel-mikula-sonarsource, line 273: Same scenario (SV not found) will happen for argument, right? `GenerateIV(SymmetricAlgorithm arg) => arg.GenerateIV();`

pavel-mikula-sonarsource (approved):
LGTM. I left some minor comments.

csaba-sagi-sonarsource (approved):
LGTM

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InitializationVectorShouldBeRandom.cs
costin-zaharia-sonarsource: No. As far as I can tell, for arguments we will get a symbol since they are defined while processing parameter list. The problem appears for class fields.

PR #3837, opened 2020-12-14
Fix S2325 FP: Do not raise for methods inside generic class

### LINKED ISSUE #3204: Fix S2325 FP: Do not raise for methods inside generic class
### Description

S2325 is now raised for methods inside generic type.

### Repro steps

``` C#
public class S2325<TFirst, TSecond>
{
    public int BuildSomething()
    {
        return 42;
    }
}
```
`BuildSomething` can be invoked as `instance.BuildSomething`
Changing method into static forces callers to invoke it as `S2325<SomeTypeA, SomeTypeB>.BuildSomething` that makes the code difficult to read.

It's also difficult to use the method from another base class that is not aware of final types yet.

See [CA1000](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1000?view=vs-2019)
See [this case](https://sonarcloud.io/project/issues?id=sonaranalyzer-dotnet&issues=AXCtNZOzAFhjSEJCuYhA&open=AXCtNZOzAFhjSEJCuYhA)

### Related information

* SonarC# Version 8.4

pavel-mikula-sonarsource:
Fixes #3204

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/MemberShouldBeStatic.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/MemberShouldBeStatic.cs
@@ -133,7 +133,9 @@ namespace SonarAnalyzer.Rules.CSharp
 
             bool IsExcludedByEnclosingType() =>
                 methodOrPropertySymbol.ContainingType.IsInterface()
+                // Any generic type in nesting chain with member accessible from outside (through the whole nesting chain) is excluded.
                 || (methodOrPropertySymbol.ContainingType.IsGenericType && IsAccessibleOutsideTheType(methodOrPropertySymbol.GetEffectiveAccessibility()))
+                // Any nested private generic type with member accessible from outside that type (not the whole nesting chain) is also excluded.
                 || (methodOrPropertySymbol.ContainingType.TypeArguments.Any() && IsAccessibleOutsideTheType(methodOrPropertySymbol.DeclaredAccessibility));
 
             bool IsAccessibleOutsideTheType(Accessibility accessibility) =>
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/MemberShouldBeStatic.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/MemberShouldBeStatic.cs
@@ -104,7 +104,8 @@ namespace SonarAnalyzer.Rules.CSharp
             if (methodOrPropertySymbol == null
                 || IsStaticVirtualAbstractOrOverride()
                 || MethodNameWhitelist.Contains(methodOrPropertySymbol.Name)
-                || IsOverrideOrInterface()
+                || IsOverrideInterfaceOrNew()
+                || IsInExcludedType()
                 || methodOrPropertySymbol.GetAttributes().Any(IsIgnoredAttribute)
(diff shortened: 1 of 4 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MemberShouldBeStatic.cs
```diff
@@ -216,10 +216,26 @@ public class Foo
...
+        private int PrivateMethod() => 42;      // Noncompliant, private member is easy to invoke => should be private
+        protected int ProtectedMethod() => 42;  // Noncompliant
+
+        internal int InternalMethod() => 42;                // Compliant
+        protected internal int ProtectedInternal() => 42;   // Compliant as internal invocation would be hard
+
+        public class NestedInGenericType
+        {
+            public int BuildSomething() => 42;  // Compliant
+        }
```
andrei-epure-sonarsource, line 238:
```suggestion
        public class PublicNestedInGenericType
        {
            public int BuildSomething() => 42;  // Compliant
        }
        private class PrivateNestedInGenericType
        {
            public int BuildSomething() => 42;  // Noncompliant
        }
        public static class PublicStaticNestedInGenericType
        {
            public static int BuildSomething() => 42;  // Compliant
        }
        private static class PrivateStaticNestedInGenericType
        {
            public static int BuildSomething() => 42;  // Compliant
        }
```

andrei-epure-sonarsource (approved):
LGTM, please add some additional tests if you find them useful

andrei-epure-sonarsource:
@pavel-mikula-sonarsource  Also... for the coverage - could you please fill it in?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MemberShouldBeStatic.cs
pavel-mikula-sonarsource: The private nested is a good catch!

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberShouldBeStatic.cs
```diff
@@ -133,16 +133,11 @@ protected override void Initialize(SonarAnalysisContext context)
...
-
-            bool IsAccessibleOutsideTheType() =>
-                methodOrPropertySymbol.GetEffectiveAccessibility() switch
-                {
-                    Accessibility.Public => true,
-                    Accessibility.Internal => true,
-                    Accessibility.ProtectedOrInternal => true,
-                    _ => false
-                };
+                || (methodOrPropertySymbol.ContainingType.IsGenericType && IsAccessibleOutsideTheType(methodOrPropertySymbol.GetEffectiveAccessibility()))
```
andrei-epure-sonarsource, line 136: I don't get this: when is `methodOrPropertySymbol.ContainingType.IsGenericType` false and `methodOrPropertySymbol.ContainingType.TypeArguments.Any()` true?

and why not use `methodOrPropertySymbol.GetEffectiveAccessibility()` in both cases? maybe add a comment to explain this, as it's not straight forward

andrei-epure-sonarsource (approved):
could you please a comment to clarify `IsExcludedByEnclosingType`?

LGTM, otherwise.

PR #3841, opened 2020-12-15
Refactor ScannerFileService to use PathSuffixPredicate

andrei-epure-sonarsource:

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/FileService.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/FileService.java
@@ -28,6 +28,8 @@ public interface FileService {
   /**
    * Note that the absolute path returned by the Scanner may be different from the absolute path returned by the Operating System
    * @see org.sonar.api.batch.fs.InputFile#uri()
+   *
+   * @param deterministicBuildPath - the path in the code coverage report when builds are done with the `-deterministic` option
    */
-  Optional<String> getAbsolutePath(String relativePath);
+  Optional<String> getAbsolutePath(String deterministicBuildPath);
 }
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/FileService.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/FileService.java
@@ -20,11 +20,14 @@
 package org.sonar.plugins.dotnet.tests;
 
 import java.util.Optional;
-import org.sonar.api.batch.fs.InputFile;
 
 public interface FileService {
 
   boolean isSupportedAbsolute(String absolutePath);
 
-  Optional<InputFile> getFileByRelativePath(String relativePath);
+  /**
+   * Note that the absolute path returned by the Scanner may be different from the absolute path returned by the Operating System
+   * @see org.sonar.api.batch.fs.InputFile#uri()
+   */
+  Optional<String> getAbsolutePath(String relativePath);
 }
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/RelativePathPredicate.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/RelativePathPredicate.java
@@ -1,0 +1,42 @@
+/*
+ * SonarSource :: .NET :: Shared library
+ * Copyright (C) 2014-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.dotnet.tests;
+
+import org.sonar.api.batch.fs.FilePredicate;
+import org.sonar.api.batch.fs.InputFile;
+import org.sonar.api.internal.google.common.annotations.VisibleForTesting;
+
+class RelativePathPredicate implements FilePredicate {
+  private final String relativePath;
+
+  RelativePathPredicate(String relativePath) {
+    this.relativePath = relativePath;
+  }
+
+  @Override
+  public boolean apply(InputFile inputFile) {
+    return inputFile.uri().getPath().endsWith(relativePath);
+  }
+
+  @VisibleForTesting
+  String getRelativePath() {
+    return relativePath;
+  }
+}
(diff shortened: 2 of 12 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
```diff
@@ -53,22 +55,19 @@ public boolean isSupportedAbsolute(String absolutePath) {
...
-        if (path.endsWith(relativePath)) {
-          count++;
-          foundFile = file;
-        }
-      }
-      if (count == 1) {
+      List<InputFile> foundFiles = new ArrayList<>();
+      FilePredicates fp = fileSystem.predicates();
+
+      fileSystem.inputFiles(fp.and(fp.hasLanguage(languageKey), new RelativePathPredicate(relativePath))).forEach(foundFiles::add);
```
pavel-mikula-sonarsource, line 61: Omg to those inside-out Java APIs 🤦‍♂️ I found just this alternative to make it less reversed and readable.
```
ImmutableList<InputFile> files = ImmutableList.copyOf(fileSystem.inputFiles(fp.and(fp.hasLanguage(languageKey), new RelativePathPredicate(relativePath))));
```

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/ScannerFileService.java
```diff
@@ -53,22 +55,19 @@ public boolean isSupportedAbsolute(String absolutePath) {
...
+
+      int foundFilesCount = foundFiles.size();
```
pavel-mikula-sonarsource, line 63: It's used once for logic and second time for logging. I'd not use the variable to spare the line & state. The `.size()` is cheap and easy to understand.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/RelativePathPredicateTest.java
```diff
@@ -0,0 +1,76 @@
...
+    return Arrays.asList(new Object[][] {
+        {"", "", true},
+        {"some/path/file.cs", "/_/some/path/file.cs", true},
+        {"some/path/file.cs", "/home/test/long/path/some/path/file.cs", true},
+        {"some/path/file.cs", "some/path/file.cs", true},
+        {"some/path/file.cs", "/path/file.cs", false},
+        {"SOME/path/file.cs", "some/path/file.cs", false},
+        {"foo", "bar", false},
+        {"foo", "", false},
+        {"\\some\\path\\file.cs", "/some/path/file.cs", false}
```
pavel-mikula-sonarsource, line 59: Let's add this for completeness
```
 {"\\some\\path\\file.cs", "\\path\\file.cs", true}
```

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/RelativePathPredicateTest.java
```diff
@@ -0,0 +1,76 @@
...
+    this.relativePath = relativePath;
+    this.testInput = testInput;
+    this.expectedResult = expectedResult;
+  }
+
+  @Parameterized.Parameters
+  public static Collection input() {
+    return Arrays.asList(new Object[][] {
+        {"", "", true},
+        {"some/path/file.cs", "/_/some/path/file.cs", true},
```
pavel-mikula-sonarsource, line 52: Let's add
```
        {"some/PATH/file.cs", "/_/some/path/file.cs", true},
        {"some/path/file.cs", "/_/some/PATH/file.cs", true},
```

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/ScannerFileServiceTest.java
```diff
@@ -85,54 +90,75 @@ public void isSupportedAbsolute_returns_fileSystem_result_when_false() {
...
+    // assert
     assertThat(result).isEmpty();
     assertThat(logTester.logs(LoggerLevel.DEBUG)).containsExactly("Found 0 indexed files for '/_/some/path/file.cs' (normalized to 'some/path/file.cs'). Will skip this coverage entry. Verify sonar.sources in .sonarqube\\out\\sonar-project.properties.");
   }
 
   @Test
-  public void getFileByRelativePath_when_multiple_indexed_files_match_returns_empty() {
-    FileSystem fs = createFileSystemForInputFiles(Arrays.asList(mockInput("root1/some/path/file.cs"), mockInput("root2/some/path/file.cs")));
+  public void getFileByRelativePath_when_multiple_indexed_files_match_returns_empty_and_logs() {
+    FileSystem fs = createFileSystemForInputFiles(Arrays.asList(mockInput("foo"), mockInput("bar")));
```
pavel-mikula-sonarsource, line 161: The test is named `when_multiple_indexed_files_match`, but the link that the FS always returns all files is little hidden in this test. And it's not visible from `createFileSystemForInputFiles` name nor the `foo`/`bar` values. So it confuses me, because it loos like it's empty because none of those files matches.

Can we rename the method to something like `createFileSystemReturningAllFiles`?

pavel-mikula-sonarsource (changes requested):
Some minor polishing

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/RelativePathPredicateTest.java
andrei-epure-sonarsource: I had them initially, but the Java [URI class](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html) implements the URI RFC specs which only use slashes - the constructor will throw when a backslash is in the string.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/RelativePathPredicate.java
```diff
@@ -0,0 +1,42 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.dotnet.tests;
+
+import org.sonar.api.batch.fs.FilePredicate;
+import org.sonar.api.batch.fs.InputFile;
+import org.sonar.api.internal.google.common.annotations.VisibleForTesting;
+
+class RelativePathPredicate implements FilePredicate {
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Following this rename, we should also rename `relativePath` constructor argument, `relativePath` field and `getRelativePath()` method.

In context of scanner, the term `relative path` refers to path from the root. Here, we're working with `pathSuffix`.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/FileService.java
```diff
@@ -20,11 +20,14 @@
...
 public interface FileService {
 
   boolean isSupportedAbsolute(String absolutePath);
 
-  Optional<InputFile> getFileByRelativePath(String relativePath);
+  /**
+   * Note that the absolute path returned by the Scanner may be different from the absolute path returned by the Operating System
+   * @see org.sonar.api.batch.fs.InputFile#uri()
+   */
+  Optional<String> getAbsolutePath(String relativePath);
```
pavel-mikula-sonarsource, line 32: Similar to other comment, I think we should rename this argument as well. As it's not the usual relative path in scanner context.

Here and in all implementations.
```suggestion
  Optional<String> getAbsolutePath(String pathSuffix);
```

pavel-mikula-sonarsource (changes requested):
Last round - just more renaming.

pavel-mikula-sonarsource (approved):
LGTM

PR #3844, opened 2020-12-16
Make sure that all hotspots are run on .Net ITs

costin-zaharia-sonarsource:
Hotspots run for ITs only if they are added in the `SonarLint.xml` files from https://github.com/SonarSource/sonar-dotnet/tree/master/analyzers/its/config. 

Currently the configuration is a bit dated and this PR will address the following:
- removes the SonarLint.xml from https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/its/SonarLint.xml since it is not used
- removes the SonarLint.xml from https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/its/sources/Ember-MM/SonarLint.xml since it is overwritten during build by the one from `its/config/Ember-MM`
- removes the SonarLint.xml from https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/its/sources/akka.net/SonarLint.xml since it is overwritten during build by the one from `its/config/akka.net`
- adds the missing Hotspot rule ids
- updates the expected issues files


csaba-sagi-sonarsource (approved):
LGTM

PR #3845, opened 2020-12-16
Fix S3927 FP and FN: Internal, protected and static methods

### LINKED ISSUE #3734: Fix S3927 FP and FN: Internal, protected and static methods
FP: Issue should not raise on `internal` and `protected` members.

FN: Issue should raise on `static` methods. RSPEC and message needs to be updated.

pavel-mikula-sonarsource:
Fixes #3734

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Rules/ImplementSerializationMethodsCorrectlyBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/ImplementSerializationMethodsCorrectlyBase.cs
@@ -32,6 +32,7 @@ namespace SonarAnalyzer.Rules
         protected const string MessageFormat = "Make this method {0}.";
         private const string problemParameterText = "have a single parameter of type 'StreamingContext'";
         private const string problemGenericParameterText = "have no type parameters";
+        private const string problemPublicText = "non-public";
 
         private static readonly ImmutableArray<KnownType> serializationAttributes =
             ImmutableArray.Create(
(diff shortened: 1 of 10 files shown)


analyzers/src/SonarAnalyzer.Common/Rules/ImplementSerializationMethodsCorrectlyBase.cs
```diff
@@ -62,39 +71,26 @@ protected override void Initialize(SonarAnalysisContext context)
...
                 SymbolKind.Method);
-        }
-
-        protected abstract GeneratedCodeRecognizer GeneratedCodeRecognizer { get; }
-        protected abstract string MethodReturnTypeShouldBeVoidMessage { get; }
 
-        protected abstract Location GetIdentifierLocation(IMethodSymbol methodSymbol);
-
-        private IEnumerable<string> FindIssues(IMethodSymbol methodSymbol)
+        private List<string> FindIssues(IMethodSymbol methodSymbol)
```
andrei-epure-sonarsource, line 80: the return type can stay `IEnumerable<string>` (it's only being enumerated upon, right?)

being private and a short class, I think we can live with it as it is, as well

andrei-epure-sonarsource (approved):
LGTM! (minor comment left)

analyzers/src/SonarAnalyzer.Common/Rules/ImplementSerializationMethodsCorrectlyBase.cs
pavel-mikula-sonarsource: I had to rewrite the `.Count` check. And that inspired me to improve the condition, simplify the flow and remove 6 unneeded rows :) So it's a valuable update.

PR #3848, opened 2020-12-16
Fix S3971 FP: Support IAsyncDisposable

### LINKED ISSUE #3639: Fix S3971 FP: Support IAsyncDisposable

pavel-mikula-sonarsource:
Fixes #3639

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotCallGCSuppressFinalize.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotCallGCSuppressFinalize.cs
@@ -36,17 +36,17 @@ namespace SonarAnalyzer.Rules.CSharp
         private const string DiagnosticId = "S3971";
         private const string MessageFormat = "Do not call 'GC.SuppressFinalize'.";
 
-        private static readonly DiagnosticDescriptor rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
-        private static readonly IEnumerable<MemberDescriptor> checkedMethods = new List<MemberDescriptor>
+        private static readonly IEnumerable<MemberDescriptor> CheckedMethodList = new List<MemberDescriptor>
         {
             new MemberDescriptor(KnownType.System_GC, "SuppressFinalize")
         };
-        internal override IEnumerable<MemberDescriptor> CheckedMethods => checkedMethods;
 
-        protected override bool ShouldReportOnMethodCall(InvocationExpressionSyntax invocation,
-            SemanticModel semanticModel, MemberDescriptor memberDescriptor)
+        internal override IEnumerable<MemberDescriptor> CheckedMethods => CheckedMethodList;
+
+        protected override bool ShouldReportOnMethodCall(InvocationExpressionSyntax invocation, SemanticModel semanticModel, MemberDescriptor memberDescriptor)
         {
             var methodDeclaration = invocation.FirstAncestorOrSelf<MethodDeclarationSyntax>();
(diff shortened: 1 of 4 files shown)


andrei-epure-sonarsource (changes requested):
nice work with adding more tests!

my comment is a clean-as-you-code suggestion for performance

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotCallGCSuppressFinalize.cs
```diff
@@ -56,12 +56,8 @@ public sealed class DoNotCallGCSuppressFinalize : DoNotCallMethodsCSharpBase
             }
 
             var methodSymbol = semanticModel.GetDeclaredSymbol(methodDeclaration);
-            if (!methodSymbol.IsIDisposableDispose())
-            {
-                return true;
-            }
-
-            return false;
+            return !methodSymbol.IsIDisposableDispose()
```
andrei-epure-sonarsource, line 59: inside `SonarAnalyzer.Common\Rules\DoNotCallMethodsBase.cs`, `ShouldReportOnMethodCall` is called at the end of the method, after using the semantic model multiple times

could you add an AST-based name check on the `methodDeclaration` (`Dispose` or `DisposeAsync`) in the override of `IsInValidContext` which is called at the very beginning? like this, you'd save a lot of semantic model calls from the starts

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotCallGCSuppressFinalize.cs
andrei-epure-sonarsource: actually, my comment is wrong

however, we can still improve the implementation of `DoNotCallMethodsBase` by moving some name checks earlier, and like this we'd improve 3 different rules easily.

andrei-epure-sonarsource (approved):
LGTM!

please check my comments - I think we can make the perf improvement in a separate PR (will you do it or should I?)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCallGCSuppressFinalize.NetCore.cs
```diff
@@ -15,13 +15,61 @@ public async ValueTask DisposeAsync()
...
+        }
+
+        protected virtual ValueTask DisposeAsyncCore() => default;
+    }
+
+    public class FakeA
+    {
+        public async ValueTask DisposeAsync()
+        {
+            await DisposeAsyncCore();
```
andrei-epure-sonarsource, line 45: do we have a rule for `await DisposeAsyncCore();` ? to make sure `DisposeAsync` is properly implemented?

andrei-epure-sonarsource: I opened https://github.com/SonarSource/sonar-dotnet/issues/3853

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotCallGCSuppressFinalize.cs
andrei-epure-sonarsource: I opened https://github.com/SonarSource/sonar-dotnet/issues/3854

Issue #2334, opened 2019-03-12
Feature request: S1155 should have a code fix


Corniel:
### Description
It would be nice if rule S1155 ("Any()" should be used to test for emptiness) has a code fix too.

#### For Fix
```csharp
public bool Zero()
{
    return this.SomeCollection.Count(SomeCodition) == 0;
}
public bool MoreThanZero()
{
    return this.SomeCollection.Count(SomeCodition) > 0;
}
public bool NotZero()
{
    return this.SomeCollection.Count(SomeCodition) != 0;
}
```

#### After Fix

```csharp
public bool Zero()
{
    return !this.SomeCollection.Any(SomeCodition);
}
public bool MoreThanZero()
{
    return this.SomeCollection.Any(SomeCodition);
}
public bool NotZero()
{
    return this.SomeCollection.Any(SomeCodition);
}
```

PR #3849, opened 2020-12-16
Fix S1155 FN: Count() != 0

Corniel:
Will working on #2334 , I noticed that both `enumerable.Count() != 0` and `enumerable.Count(condition) ?? 0|1` where not covered by tests. The second already worked, but the first one didn't, so hence a simple fix.


andrei-epure-sonarsource (approved):
LGTM, nice catch!

PR #3855, opened 2020-12-17
Rule S5332: Using clear-text protocols is security-sensitive

### LINKED ISSUE #3721: Rule S5332: Using clear-text protocols is security-sensitive
Code examples:
https://github.com/SonarSource/security-expected-issues/tree/master/dotnet/NetCore3MVC/Controllers/RSPEC_5332

costin-zaharia-sonarsource:
Fixes #3721


analyzers/its/config/Ember-MM/SonarLint.xml
```diff
@@ -31,6 +31,7 @@
     <Rule><Key>S4829</Key></Rule>
     <Rule><Key>S4834</Key></Rule>
     <Rule><Key>S5042</Key></Rule>
+	<Rule><Key>S5332</Key></Rule>
```
csaba-sagi-sonarsource, line 34: pedantic: the sonarlint.xml files use spaces instead of tabs, so it would be nice to change the tabs to spaces in them.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/ClearTextProtocolsAreSensitiveTest.cs
```diff
@@ -0,0 +1,56 @@
+/*
```
csaba-sagi-sonarsource, line 1: Maybe a   Verifier.VerifyNoIssueReported test case can be added to verify that the hotspot configuration checking is done correctly.

costin-zaharia-sonarsource: Good catch! I forgot about this.

csaba-sagi-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.CSharp/Helpers/InterpolatedStringExpressionSyntaxExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace SonarAnalyzer.Helpers
+{
+    public static class InterpolatedStringExpressionSyntaxExtensions
+    {
+        public static string GetContentsText(this InterpolatedStringExpressionSyntax interpolatedStringExpression) =>
```
pavel-mikula-sonarsource, line 27: Nice! Can you reuse this directly in here?

https://github.com/SonarSource/sonar-dotnet/blob/fec881e2659764c9ad157be105e8af1c5b571a28/analyzers/src/SonarAnalyzer.CSharp/Rules/SqlKeywordsDelimitedBySpace.cs#L186

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,167 @@
...
+        private const string TelnetPatternForIdentifier = @"(Telnet)(?![a-z])";
+        private const string LocalHost = "localhost";
+        private const string LocalHostIp = "127.0.0.1";
+        private const string EnableSslName = "EnableSsl";
+
+        private static readonly DiagnosticDescriptor DefaultRule =
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager).WithNotConfigurable();
+
+        private static readonly DiagnosticDescriptor EnableSslRule =
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, EnableSslMessage, RspecStrings.ResourceManager).WithNotConfigurable();
```
pavel-mikula-sonarsource, line 54: Single line would be more readable. And actually allows to remove the empty line between them.

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        private readonly Dictionary<string, string> recommendedProtocols = new Dictionary<string, string>
+        {
+            {"telnet", "ssh"},
+            {"ftp", "sftp, scp or ftps"},
+            {"http", "https"},
+            {"clear-text SMTP", "SMTP over SSL/TLS or SMTP with STARTTLS" }
+        };
+
+        private readonly ImmutableArray<string> validServerValues = ImmutableArray.Create(LocalHost, LocalHostIp);
```
pavel-mikula-sonarsource, line 64: The localhost detection is very scattered around. Adding another exception would be hard, it has to be done multiple times.
Let's extract this, to have just one place to maintain.
```
	ValidPattern = "localhost|127.0.0.1";
```
Reuse this in `HttpPattern`/`FtpPattern` creation. And add 
```
ValidServerPattern = "^" + ValidPattern + "$"
```
to replace `validServerValues`.

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        private readonly CSharpObjectInitializationTracker objectInitializationTracker =
+            new CSharpObjectInitializationTracker(constantValue => constantValue is bool value && value,
+                                                  ImmutableArray.Create(KnownType.System_Net_Mail_SmtpClient, KnownType.System_Net_FtpWebRequest),
+                                                  propertyName => EnableSslName == propertyName);
```
pavel-mikula-sonarsource, line 69: This looks like reversed Java convention copy/paste issue.
```suggestion
                                                  propertyName => propertyName == EnableSslName);
```

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(DefaultRule, EnableSslRule);
+
+        public ClearTextProtocolsAreSensitive() : this(AnalyzerConfiguration.Hotspot) { }
+
+        public ClearTextProtocolsAreSensitive(IAnalyzerConfiguration analyzerConfiguration) : base(analyzerConfiguration) { }
+
+        protected override void Initialize(SonarAnalysisContext context) =>
+            context.RegisterCompilationStartAction(ccc =>
```
pavel-mikula-sonarsource, line 78: Seems like a copy/paste issue. We have some nested `c`, `cc`, `ccc` cases, but it should not be needed here.
```suggestion
            context.RegisterCompilationStartAction(c =>
```

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,167 @@
...
+    public sealed class ClearTextProtocolsAreSensitive : HotspotDiagnosticAnalyzer
+    {
+        private const string DiagnosticId = "S5332";
+        private const string MessageFormat = "Using {0} protocol is insecure. Use {1} instead.";
+        private const string EnableSslMessage = "EnableSsl should be set to true.";
+        private const string HttpPattern = @"^http:\/\/(?!localhost|127.0.0.1).+";
+        private const string FtpPattern = @"^ftp:\/\/.*@(?!localhost|127.0.0.1)";
+        private const string Telnet = "telnet";
+        private const string TelnetPattern = @"^telnet:\/\/.*@(?!localhost|127.0.0.1)";
+        private const string TelnetPatternForIdentifier = @"(Telnet)(?![a-z])";
```
pavel-mikula-sonarsource, line 45: Parenthesis should not be needed
```suggestion
        private const string TelnetPatternForIdentifier = @"Telnet(?![a-z])";
```

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,167 @@
...
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public sealed class ClearTextProtocolsAreSensitive : HotspotDiagnosticAnalyzer
+    {
+        private const string DiagnosticId = "S5332";
+        private const string MessageFormat = "Using {0} protocol is insecure. Use {1} instead.";
+        private const string EnableSslMessage = "EnableSsl should be set to true.";
+        private const string HttpPattern = @"^http:\/\/(?!localhost|127.0.0.1).+";
+        private const string FtpPattern = @"^ftp:\/\/.*@(?!localhost|127.0.0.1)";
+        private const string Telnet = "telnet";
```
pavel-mikula-sonarsource, line 43: It took me some time to understand that this is the key for the next dictionary.
```suggestion
        private const string TelnetKey = "telnet";
```

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,167 @@
...
+            {
+                c.ReportDiagnosticWhenActive(Diagnostic.Create(DefaultRule, c.Node.GetLocation(), unsafeProtocol, recommendedProtocols[unsafeProtocol]));
+            }
+        }
+
+        private bool IsServerSafe(ObjectCreationExpressionSyntax objectCreation) =>
+            objectCreation.ArgumentList?.Arguments.Count > 0
+            && validServerValues.Contains(GetText(objectCreation.ArgumentList.Arguments[0].Expression));
+
+        private static bool TryGetUnsafeProtocol(string text, out string unsafeProtocol)
```
pavel-mikula-sonarsource, line 139: Having this, we could remove the `out` parameter and simplify the flow and remove all indirection. Each `if/elseif/else` would directly return value.
```suggestion
        private static string GetUnsafeProtocol(string text)
```

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,167 @@
...
+                && !enableSslValue)
+            {
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(EnableSslRule, assignment.GetLocation()));
+            }
+        }
+
+        private void VisitStringExpressions(SyntaxNodeAnalysisContext c)
+        {
+            var text = GetText(c.Node);
+            if (TryGetUnsafeProtocol(text, out var unsafeProtocol))
```
pavel-mikula-sonarsource, line 129: Following the suggestion below for `TryGetUnsafeProtocol`, this could be updated afterwards
```suggestion
            if (GetUnsafeProtocol(GetText(c.Node)) is {} unsafeProtocol)
```

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        private bool IsServerSafe(ObjectCreationExpressionSyntax objectCreation) =>
+            objectCreation.ArgumentList?.Arguments.Count > 0
+            && validServerValues.Contains(GetText(objectCreation.ArgumentList.Arguments[0].Expression));
+
+        private static bool TryGetUnsafeProtocol(string text, out string unsafeProtocol)
+        {
+            unsafeProtocol = null;
+
+            if (Regex.IsMatch(text, HttpPattern))
```
andrei-epure-sonarsource, line 143: given these will be executed a lot, while not make `HttpPattern`, `FtpPattern` compiled regexes rather than strings, to speed the checks?  (see [Interpreted vs. compiled regular expressions](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices#interpreted-vs-compiled-regular-expressions) - compiled regexes will slow the startup, however if many checks are done, the benefit will show )

pavel-mikula-sonarsource (changes requested):
Minor improvements and few more tests

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,167 @@
...
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public sealed class ClearTextProtocolsAreSensitive : HotspotDiagnosticAnalyzer
+    {
+        private const string DiagnosticId = "S5332";
+        private const string MessageFormat = "Using {0} protocol is insecure. Use {1} instead.";
+        private const string EnableSslMessage = "EnableSsl should be set to true.";
+        private const string HttpPattern = @"^http:\/\/(?!localhost|127.0.0.1).+";
```
andrei-epure-sonarsource, line 41: I would suggest to organize the fields differently : 
- keep the Regexes together
- keep the messages and message components together 
(and maybe separate the two groups by an empty line)

(and I suggest making them `new Regex(pattern, RegexOptions.Singleline | RegexOptions.Compiled);`, see [this comment](https://github.com/SonarSource/sonar-dotnet/pull/3855/files#r545770235)

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,167 @@
...
+                context.RegisterSyntaxNodeActionInNonGenerated(VisitStringExpressions, SyntaxKind.StringLiteralExpression, SyntaxKind.InterpolatedStringExpression);
+                context.RegisterSyntaxNodeActionInNonGenerated(VisitObjectCreation, SyntaxKind.ObjectCreationExpression);
+                context.RegisterSyntaxNodeActionInNonGenerated(VisitInvocationExpression, SyntaxKind.InvocationExpression);
+                context.RegisterSyntaxNodeActionInNonGenerated(VisitAssignments, SyntaxKind.SimpleAssignmentExpression);
+            });
+
+        private void VisitObjectCreation(SyntaxNodeAnalysisContext context)
+        {
+            var objectCreation = (ObjectCreationExpressionSyntax)context.Node;
+            if (!IsServerSafe(objectCreation) && objectInitializationTracker.ShouldBeReported(objectCreation, context.SemanticModel))
```
andrei-epure-sonarsource, line 94: should this be using `if / else` ?
-  if not, could you please add a test case where both messages are being raised?
-  if it should be if/else, could you do the pattern match before the semanticModel check?

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
andrei-epure-sonarsource: using compiled regexes applies to all places in this class

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
andrei-epure-sonarsource, line 129: rather than the `TryGet` pattern, you could simply return `unsafeProtocol` from the method and do a `if (GetUnsafeProtocol(text) is {} unsafeProtocol)` here (to avoid dealing with a boolean and a string when there's no need to)

andrei-epure-sonarsource (changes requested):
I've left some comments, my main suggestion is to use compiled regexes for better performance

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -38,20 +38,19 @@ public sealed class ClearTextProtocolsAreSensitive : HotspotDiagnosticAnalyzer
...
+        private const string ValidServerPattern = "localhost|127.0.0.1|::1";
 
-        private static readonly DiagnosticDescriptor DefaultRule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager).WithNotConfigurable();
+        private static readonly DiagnosticDescriptor DefaultRule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager).WithNotConfigurable();
+        private static readonly DiagnosticDescriptor EnableSslRule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, EnableSslMessage, RspecStrings.ResourceManager).WithNotConfigurable();
 
-        private static readonly DiagnosticDescriptor EnableSslRule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, EnableSslMessage, RspecStrings.ResourceManager).WithNotConfigurable();
+        private readonly Regex httpRegex = new Regex(@$"^http:\/\/(?!{ValidServerPattern}).+", RegexOptions.Singleline | RegexOptions.Compiled);
```
andrei-epure-sonarsource, line 49: nitpick: `new Regex(pattern, RegexOptions.Singleline | RegexOptions.Compiled);` can be a `CompiledRegex(string pattern)` private static method

andrei-epure-sonarsource (approved):
LGTM

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,196 @@
...
+
+            var f = $"ftp://anonymous@example.com"; // Noncompliant
+            var g = $"ftp://{part}@example.com"; // Noncompliant
+            var h = "ssh://anonymous@example.com";
+            var i = "sftp://anonymous@example.com";
+            var ftpProtocolScheme = "ftp://"; // It's compliant when standalone
+            var j = "ftp://" + user + "@example.com"; // Compliant - FN (protocol is compliant when standallone, check previous case)
+            var k = "ftp://anonymous@" + domain;// Noncompliant
+            var l = $"ftp://anonymous@{domain}"; // Noncompliant
+            var m = $"{ftp}://anonymous@example.com"; // Compliant - FN
```
pavel-mikula-sonarsource, line 46: Why FN? It's true negative for me, as the ftp variable might contain anything
```suggestion
            var m = $"{ftp}://anonymous@example.com"; // Compliant
```

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,165 @@
...
+        private const string ValidServerPattern = "localhost|127.0.0.1|::1";
+
+        private static readonly DiagnosticDescriptor DefaultRule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager).WithNotConfigurable();
+        private static readonly DiagnosticDescriptor EnableSslRule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, EnableSslMessage, RspecStrings.ResourceManager).WithNotConfigurable();
+
+        private readonly Regex httpRegex = CreateCompiledRegex(@$"^http:\/\/(?!{ValidServerPattern}).+");
+        private readonly Regex ftpRegex = CreateCompiledRegex(@$"^ftp:\/\/.*@(?!{ValidServerPattern})");
+        private readonly Regex telnetRegex = CreateCompiledRegex(@$"^telnet:\/\/.*@(?!{ValidServerPattern})");
+        private readonly Regex telnetRegexForIdentifier = CreateCompiledRegex(@"Telnet(?![a-z])");
+        private readonly Regex validServerRegex = CreateCompiledRegex(ValidServerPattern);
```
pavel-mikula-sonarsource, line 53: We need to fix the begin and end, as mentioned in the original suggestion. Let's add this test case:
```
using var localhosting = new SmtpClient("localhosting", 25);  // Noncompliant
```
```suggestion
        private readonly Regex validServerRegex = CreateCompiledRegex($"^{ValidServerPattern}$");
```

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,165 @@
...
+        private const string EnableSslMessage = "EnableSsl should be set to true.";
+
+        private const string TelnetKey = "telnet";
+        private const string EnableSslName = "EnableSsl";
+        private const string ValidServerPattern = "localhost|127.0.0.1|::1";
+
+        private static readonly DiagnosticDescriptor DefaultRule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager).WithNotConfigurable();
+        private static readonly DiagnosticDescriptor EnableSslRule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, EnableSslMessage, RspecStrings.ResourceManager).WithNotConfigurable();
+
+        private readonly Regex httpRegex = CreateCompiledRegex(@$"^http:\/\/(?!{ValidServerPattern}).+");
```
pavel-mikula-sonarsource, line 49: This `+` should not is not needed
```suggestion
        private readonly Regex httpRegex = CreateCompiledRegex(@$"^http:\/\/(?!{ValidServerPattern}).");
```

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,165 @@
...
+
+        private static string GetText(SyntaxNode node) =>
+            node switch
+            {
+                InterpolatedStringExpressionSyntax interpolatedStringExpression => interpolatedStringExpression.GetContentsText(),
+                LiteralExpressionSyntax literalExpression => literalExpression.Token.ValueText,
+                _ => string.Empty
+            };
+
+        private static Regex CreateCompiledRegex(string pattern) =>
```
pavel-mikula-sonarsource, line 162: Just to keep it short
```suggestion
        private static Regex CreateRegex(string pattern) =>
```
or
```suggestion
        private static Regex CompileRegex(string pattern) =>
```

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,165 @@
...
+            new Regex(pattern, RegexOptions.Singleline | RegexOptions.Compiled);
```
pavel-mikula-sonarsource, line 163: Can you either remove the SingleLine or add a test case demonstrating multiline string where the `.` matches the new line character?
```suggestion
            new Regex(pattern, RegexOptions.Compiled);
```

pavel-mikula-sonarsource, line 163: I think we'll need to add `IgnoreCase` to support stuff like this for each protocol
```
Localhost
LOCALHOST
Http
HTTP
```

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,165 @@
...
+            else if (ftpRegex.IsMatch(text))
+            {
+                return "ftp";
+            }
+            else if (telnetRegex.IsMatch(text))
+            {
+                return "telnet";
+            }
+
+            return null;
```
pavel-mikula-sonarsource, line 151: According to convention, to keep the flows explicitly trivial. VB.NET has better syntax for this :trollface: 😄 
```suggestion
        else
        {
            return null;
        }
```

pavel-mikula-sonarsource (changes requested):
More fun with regexes

analyzers/src/SonarAnalyzer.CSharp/Rules/ClearTextProtocolsAreSensitive.cs
costin-zaharia-sonarsource: Then we will match also strings like `TelNet`

pavel-mikula-sonarsource: Please add more tests - for each protocol, as mentioned above.
```
HTTP://example.com
http://localhost
FTP://anonymous@example.com
ftp://anonymous@LOCALHOST
TELNET://anonymous@example.com
telnet://anonymous@localhost
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ClearTextProtocolsAreSensitive.cs
```diff
@@ -0,0 +1,203 @@
...
+using System.Net;
+using System.Net.Mail;
+
+namespace Tests.Diagnostics
+{
+    class ClearTextProtocols
+    {
+        private const string a = "http://example.com"; // Noncompliant {{Using http protocol is insecure. Use https instead.}}
+        private const string b = "https://example.com";
+        private const string c = "http://localhost";
```
pavel-mikula-sonarsource, line 13: Let's add last one to be sure:
```
        private const string c = "http://localhost/path/query?query=xxx";
        private const string c = "http://example.com/localhost/?query=xxx";
        private const string c = "http://example.com/path/?query=localhost";
```

pavel-mikula-sonarsource (approved):
LGTM, please add more testcases from those last two opened conversations

PR #3856, opened 2020-12-17
Fix S4423 FPs: If statements, unary negations

### LINKED ISSUE #3720: Rule S4423: Weak SSL and TLS protocols should not be used
Code examples:
https://github.com/SonarSource/security-expected-issues/tree/master/dotnet/NetCore3MVC/Controllers/RSPEC_4423

csaba-sagi-sonarsource:
Fixes #3720

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/WeakSslTlsProtocols.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/WeakSslTlsProtocols.cs
@@ -34,11 +34,18 @@ namespace SonarAnalyzer.Rules.CSharp
     {
         protected override GeneratedCodeRecognizer GeneratedCodeRecognizer { get; } = CSharpGeneratedCodeRecognizer.Instance;
 
-        protected override SyntaxKind SyntaxKind { get; } = SyntaxKind.IdentifierName;
+        protected override SyntaxKind SyntaxKindOfInterest { get; } = SyntaxKind.IdentifierName;
 
         protected override DiagnosticDescriptor Rule { get; } = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
         protected override string GetIdentifierText(IdentifierNameSyntax identifierNameSyntax) =>
             identifierNameSyntax.Identifier.Text;
+
+        protected override bool IsNodeOfInterest(SyntaxNode node)
+        {
+            var topNode = node.Parent.GetSelfOrTopParenthesizedExpression();
+            return !topNode.Parent.IsKind(SyntaxKind.BitwiseNotExpression)
+                   && !topNode.Parent.Parent.IsKind(SyntaxKind.IfStatement);
+        }
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/WeakSslTlsProtocols.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/WeakSslTlsProtocols.cs
@@ -56,7 +56,6 @@ namespace SonarAnalyzer.Rules.CSharp
             }
 
             var current = topNode;
-            var parent = current.Parent;
             while (current.Parent != null && !current.Parent.IsKind(SyntaxKind.IfStatement))
             {
                 current = current.Parent;
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/WeakSslTlsProtocols.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/WeakSslTlsProtocols.cs
@@ -43,9 +43,32 @@ namespace SonarAnalyzer.Rules.CSharp
 
         protected override bool IsNodeOfInterest(SyntaxNode node)
         {
+            if (!(node.Parent is MemberAccessExpressionSyntax))
+            {
+                return false;
+            }
+
             var topNode = node.Parent.GetSelfOrTopParenthesizedExpression();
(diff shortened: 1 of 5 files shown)


analyzers/src/SonarAnalyzer.Common/Rules/WeakSslTlsProtocolsBase.cs
```diff
@@ -47,24 +48,27 @@ public abstract class WeakSslTlsProtocolsBase<TSyntaxKind, TIdentifierNameSyntax
 
         protected abstract GeneratedCodeRecognizer GeneratedCodeRecognizer { get; }
 
-        protected abstract TSyntaxKind SyntaxKind { get; }
+        protected abstract TSyntaxKind SyntaxKindOfInterest { get; }
```
costin-zaharia-sonarsource, line 51: I would go for the simple name `SyntaxKind` since `OfInterest` does not add a lot of details.

analyzers/src/SonarAnalyzer.Common/Rules/WeakSslTlsProtocolsBase.cs
```diff
@@ -47,24 +48,27 @@ public abstract class WeakSslTlsProtocolsBase<TSyntaxKind, TIdentifierNameSyntax
...
 
         protected abstract DiagnosticDescriptor Rule { get; }
 
         protected abstract string GetIdentifierText(TIdentifierNameSyntax identifierNameSyntax);
 
+        protected abstract bool IsNodeOfInterest(SyntaxNode node);
```
costin-zaharia-sonarsource, line 57: `IsInsideIfCondition`?

analyzers/src/SonarAnalyzer.VisualBasic/Rules/WeakSslTlsProtocols.cs
```diff
@@ -34,11 +34,34 @@ public sealed class WeakSslTlsProtocols : WeakSslTlsProtocolsBase<SyntaxKind, Id
...
+            var parent = current.Parent;
+            while (parent != null && !parent.IsKind(SyntaxKind.IfStatement))
+            {
+                current = parent;
+                parent = parent.Parent;
+            }
+
+            if (parent != null && parent.IsKind(SyntaxKind.IfStatement))
+            {
+                return (parent as IfStatementSyntax).Condition != current;
```
costin-zaharia-sonarsource, line 61: I would use a direct cast here since we check the syntax kind previously and if the cast fails we will get a NRE anyway.

analyzers/src/SonarAnalyzer.VisualBasic/Rules/WeakSslTlsProtocols.cs
```diff
@@ -34,11 +34,34 @@ public sealed class WeakSslTlsProtocols : WeakSslTlsProtocolsBase<SyntaxKind, Id
...
+        protected override bool IsNodeOfInterest(SyntaxNode node)
+        {
+            if (!(node.Parent is MemberAccessExpressionSyntax))
+            {
+                return false;
+            }
+
+            var current = node;
+            var parent = current.Parent;
+            while (parent != null && !parent.IsKind(SyntaxKind.IfStatement))
```
costin-zaharia-sonarsource, line 53: I've realized now that the condition ca be in different constructs too (e.g. Ternary operator). I suggest to document the FP and let it be for the moment. We could handle it next time we touch this rule.

analyzers/src/SonarAnalyzer.VisualBasic/Rules/WeakSslTlsProtocols.cs
```diff
@@ -34,11 +34,34 @@ public sealed class WeakSslTlsProtocols : WeakSslTlsProtocolsBase<SyntaxKind, Id
...
             identifierNameSyntax.Identifier.Text;
+
+        protected override bool IsNodeOfInterest(SyntaxNode node)
+        {
+            if (!(node.Parent is MemberAccessExpressionSyntax))
+            {
+                return false;
+            }
+
+            var current = node;
```
costin-zaharia-sonarsource, line 51: I think we can replace `parent` with `current.Parent` in order to have just a local variable and skip some assignments.

costin-zaharia-sonarsource (changes requested):
Looks good. I've added just a few minor comments.

analyzers/src/SonarAnalyzer.VisualBasic/Rules/WeakSslTlsProtocols.cs
csaba-sagi-sonarsource: You mean that  I should remove the if and just do a cast?

costin-zaharia-sonarsource: No, I would replace the `as` with a direct cast.
```suggestion
                return ((IfStatementSyntax)parent).Condition != current;
```

analyzers/src/SonarAnalyzer.CSharp/Rules/WeakSslTlsProtocols.cs
```diff
@@ -57,15 +57,14 @@ protected override bool IsNodeOfInterest(SyntaxNode node)
 
             var current = topNode;
             var parent = current.Parent;
```
costin-zaharia-sonarsource, line 59: I think we can remove the parent.

costin-zaharia-sonarsource (approved):
LGTM!

PR #3859, opened 2020-12-18
Add missing BOMs and update rspec.ps1 to use it

pavel-mikula-sonarsource:


scripts/rspec/rspec.ps1
```diff
@@ -127,7 +127,7 @@ function UpdateTestEntry($rule) {
     $ruleTypeTestCase = "${sonaranalyzerPath}\\tests\\SonarAnalyzer.UnitTest\\PackagingTests\\$fileToEdit.cs"
     $ruleId = $ruleKey.Substring(1)
     (Get-Content "${ruleTypeTestCase}") -replace "//\[`"$ruleId`"\]", "[`"$ruleId`"] = `"$ruleType`"" |
-        Set-Content "${ruleTypeTestCase}"
+        Set-Content "${ruleTypeTestCase}" -Encoding UTF8
```
andrei-epure-sonarsource, line 130: should this be 
```suggestion
        Set-Content "${ruleTypeTestCase}" -Encoding utf8BOM
```
>utf8BOM: Encodes in UTF-8 format with Byte Order Mark (BOM)

andrei-epure-sonarsource (changes requested):

scripts/rspec/rspec.ps1
pavel-mikula-sonarsource: Yes, it's better to be on the safe explicit side. Using UTF8 also works, but seems glitchy:
File UTF8 & no encoding parameter => UTF8 is saved, BOM is lost if present
File UTF8 & encoding UTF8 parameter => BOM is preserved

andrei-epure-sonarsource (approved):
LGTM!

Issue #2478, opened 2019-07-08
S1144 FP: `Deconstruct` method in nested type should be considered


antoinebj:
### Description

A false positive exists for rule S1144 when using a deconstructible type.
This is similar to #2126 but the example given in reply failed to reproduce it because the deconstructible type was not nested.

### Repro steps

Run analysis on this code:

```csharp
internal sealed class S1144
{
    public void SomeMethod()
    {
        var (a, b) = new SomeDeconstructibleType();
    }

    private sealed class SomeDeconstructibleType
    {
        public void Deconstruct(out object a, out object b)
        {
            a = b = null;
        }
    }
}
```

### Expected behavior

No violation of S1144 should be detected.

### Actual behavior

A violation of S1144 is detected.

### Known workarounds

Suppression of the warning.

### Related information

* SonarC# Version 7.15.0.8572
* Visual Studio Version 16.1.4

andrei-epure-sonarsource:
thanks @antoinebj  I confirm it's a FP

stefanloerwald:
This is still an active issue now. Is there any plan to fix it?

andrei-epure-sonarsource:
It's in our backlog, @stefanloerwald , so we will address it at some point in the future. I don't have an ETA for that, though.

PR #3862, opened 2020-12-18
Improve S1144 for protected internal members

pavel-mikula-sonarsource:
Related to #2478

Spin off from https://github.com/SonarSource/sonar-dotnet/pull/3843#discussion_r545271559

Having only public facing member, the rule didn't trigger.
Having only one Deconstruct, the code with access checking didn't run.

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Helpers/AccessibilityExtensions.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Helpers/AccessibilityExtensions.cs
@@ -1,0 +1,30 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2020 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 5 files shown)


andrei-epure-sonarsource (approved):
looks great to me 👍

analyzers/src/SonarAnalyzer.Common/Helpers/AccessibilityExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Microsoft.CodeAnalysis;
+
+namespace SonarAnalyzer.Helpers
+{
+    public static class AccessibilityExtensions
```
andrei-epure-sonarsource, line 25: can you please add a comment here about `ProtectedOrInternal` and `ProtectedAndInternal` because I find it a brain teaser

PR #3863, opened 2020-12-18
Restore VSIX nuget packages

andrei-epure-sonarsource:
When running 
- `git clean -dfx`
- `.\scripts\build\dev-build.ps1 -buildJava -build -test -release` 

The build fails for the VSIX project (at least on my machine). This should fix it.


costin-zaharia-sonarsource (approved):
LGTM!

Issue #2528, opened 2019-08-01
handling of null conditional and null coalescing operators can lead to a false positive for S2259


ChaosCrafter:
When an IF statement uses the null conditional rule and (or?) the null coalescing operator then the else path is not marked as cannot be null, resulting in false S2259 messages.

Eg.
if ((x?.value??0)==0)
{ // this happens if x is null, x.value is null or x.value == 0
}
else
{ // this part can't be called if x is null or x.value is null
  text = $"{x.value} items";
}

In this example, the line 
  text = $"{x.value} items";
gets a message S2259 'x' is null on at least one execution path.
Unless I'm misunderstanding something, that is not true.

andrei-epure-sonarsource:
This is most likely a duplicate of #2361 which got fixed during this sprint, it needs to be reviewed and confirmed it got fixed.

christophe-zurn-sonarsource:
@andrei-epure-sonarsource I had a look and this is not actually a duplicate of #2361. The issue here comes from the fact that we do not support constraints on integers in the symbolic execution engine. 
Basically, when `(x?.value??0)==0` is false, this indeed implies that `x` is not null, but since we do not track integers, we cannot infer the non-nullability of `x` from the `equals` comparing integers.

A simpler reproducer for this FP:
```
string Foo(object x)
        {
            if ((x == null ? 0 : 1) == 0)
            {
                return "x is null";
            }
            else
            {
                return $"{x.value} items";
            }
        }
```

andrei-epure-sonarsource:
Another repro [on community.](https://community.sonarsource.com/t/c-null-coalescing-x-is-null-on-at-least-one-execution-path/35221)

And this can also manifest as a False Negative in S2583/S2589, e.g.

```
    public class Repro_2528
    {
        public int Count { get; }
        public int Foo()
        {
            Repro_2528 x = null;
            if (x == null) { } // Noncompliant
            if ((x?.Count ?? 0) == 0) // FN
            {
                return -1;
            }

            return 1;
        }
    }
```

pavel-mikula-sonarsource:
Another repro, from #4537 
```
        private void Sample()
        {
            string someString = null;

            if (!someString?.Contains("a") ?? true)
                Console.WriteLine("It's null or doesn't contain 'a'");
            else
                Console.WriteLine(someString.Length); //S2259 reported here
        }
```

andrei-epure-sonarsource:
[...]

PR #3869, opened 2020-12-22
Add FP and FN reproducers for issue 2528

andrei-epure-sonarsource:
Adding test cases for #2528


andrei-epure-sonarsource:
@csaba-sagi-sonarsource we like to document behavior even if we don't fix the issue right away

csaba-sagi-sonarsource (approved):
LGTM

PR #3871, opened 2020-12-23
Rule S1313 should not raise when loopback, broadcast addresses are used (#3450)

### LINKED ISSUE #3450: Rule S1313 should not raise when loopback, broadcast addresses are used
The subtask C#/VB RSPEC-1313 inherits now the complete description from [the main RSPEC](https://jira.sonarsource.com/browse/RSPEC-1313) and in particular this text block about exceptions:
```
Loopback addresses 127.0.0.0/8 in CIDR notation (from 127.0.0.0 to 127.255.255.255)
Broadcast address 255.255.255.255
Non routable address 0.0.0.0
Strings of the form 2.5.<number>.<number> as they often match Object Identifiers (OID).
```

So, the rule should be updated to not raise when loopback, broadcast, non routable addresses are used and when OIDs are used.

csaba-sagi-sonarsource:
Fixes #3450

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/HardcodedIpAddress.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/HardcodedIpAddress.cs
@@ -53,7 +53,7 @@ namespace SonarAnalyzer.Rules.CSharp
             literalExpression.Ancestors().AnyOfKind(SyntaxKind.Attribute);
 
         protected override string GetAssignedVariableName(LiteralExpressionSyntax stringLiteral) =>
-            stringLiteral.FirstAncestorOrSelf<SyntaxNode>(IsVariableIdentifier)?.ToString().ToUpperInvariant();
+            stringLiteral.FirstAncestorOrSelf<SyntaxNode>(IsVariableIdentifier)?.ToString();
 
         private static bool IsVariableIdentifier(SyntaxNode syntaxNode) =>
             syntaxNode is StatementSyntax
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/HardcodedIpAddress.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/HardcodedIpAddress.cs
@@ -32,11 +32,6 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class HardcodedIpAddress : HardcodedIpAddressBase<SyntaxKind, LiteralExpressionSyntax>
     {
-        private static bool IsVariableIdentifier(SyntaxNode syntaxNode) =>
-            syntaxNode is StatementSyntax
-            || syntaxNode is VariableDeclaratorSyntax
-            || syntaxNode is ParameterSyntax;
-
         protected override GeneratedCodeRecognizer GeneratedCodeRecognizer { get; } = CSharpGeneratedCodeRecognizer.Instance;
 
         protected override SyntaxKind SyntaxKind { get; } = SyntaxKind.StringLiteralExpression;
@@ -61,5 +56,10 @@ namespace SonarAnalyzer.Rules.CSharp
 
         protected override string GetAssignedVariableName(LiteralExpressionSyntax stringLiteral) =>
             stringLiteral.FirstAncestorOrSelf<SyntaxNode>(IsVariableIdentifier)?.ToString().ToUpperInvariant();
+
+        private static bool IsVariableIdentifier(SyntaxNode syntaxNode) =>
+            syntaxNode is StatementSyntax
+            || syntaxNode is VariableDeclaratorSyntax
+            || syntaxNode is ParameterSyntax;
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/HardcodedIpAddress.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/HardcodedIpAddress.cs
@@ -18,23 +18,30 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System.Collections.Immutable;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.Helpers.CSharp;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     [Rule(DiagnosticId)]
-    public sealed class HardcodedIpAddress : HardcodedIpAddressBase<LiteralExpressionSyntax>
+    public sealed class HardcodedIpAddress : HardcodedIpAddressBase<SyntaxKind, LiteralExpressionSyntax>
     {
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager)
-                .WithNotConfigurable();
+        private static bool IsVariableIdentifier(SyntaxNode syntaxNode) =>
+            syntaxNode is StatementSyntax
+            || syntaxNode is VariableDeclaratorSyntax
+            || syntaxNode is ParameterSyntax;
+
+        protected override GeneratedCodeRecognizer GeneratedCodeRecognizer { get; } = CSharpGeneratedCodeRecognizer.Instance;
+
+        protected override SyntaxKind SyntaxKind { get; } = SyntaxKind.StringLiteralExpression;
+
+        protected override DiagnosticDescriptor Rule { get; } = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager).WithNotConfigurable();
 
         public HardcodedIpAddress()
             : base(AnalyzerConfiguration.Hotspot)
@@ -46,25 +53,6 @@ namespace SonarAnalyzer.Rules.CSharp
         {
         }
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } =
-            ImmutableArray.Create(rule);
-
-        protected override void Initialize(SonarAnalysisContext context)
-        {
-            context.RegisterCompilationStartAction(
-                ccc =>
-                {
-                    if (!IsEnabled(ccc.Options))
-                    {
-                        return;
-                    }
-
-                    context.RegisterSyntaxNodeActionInNonGenerated(
-                        GetAnalysisAction(rule),
-                        SyntaxKind.StringLiteralExpression);
-                });
-        }
-
         protected override string GetValueText(LiteralExpressionSyntax literalExpression) =>
             literalExpression.Token.ValueText;
 
@@ -73,10 +61,5 @@ namespace SonarAnalyzer.Rules.CSharp
 
         protected override string GetAssignedVariableName(LiteralExpressionSyntax stringLiteral) =>
             stringLiteral.FirstAncestorOrSelf<SyntaxNode>(IsVariableIdentifier)?.ToString().ToUpperInvariant();
-
-        private static bool IsVariableIdentifier(SyntaxNode syntaxNode) =>
-            syntaxNode is StatementSyntax ||
-            syntaxNode is VariableDeclaratorSyntax ||
-            syntaxNode is ParameterSyntax;
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/HardcodedIpAddress.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/HardcodedIpAddress.cs
@@ -36,15 +36,13 @@ namespace SonarAnalyzer.Rules.CSharp
 
         protected override SyntaxKind SyntaxKind { get; } = SyntaxKind.StringLiteralExpression;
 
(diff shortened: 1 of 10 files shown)


analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -20,74 +20,130 @@
...
+        }
 
-                if (HasAttributes(stringLiteral))
-                {
-                    return;
-                }
+        private void CheckForHardcodedIpAddresses(SyntaxNodeAnalysisContext context)
+        {
+            var stringLiteral = (TLiteralExpression)context.Node;
+            var variableName = GetAssignedVariableName(stringLiteral);
```
pavel-mikula-sonarsource, line 109: Fixing old technical debt: 
* `ignoredVariableNames` contains upper case values
* `Contains` below is case sensitive
* There's no sign that `GetAssignedVariableName` should or does return upper case
* Inheriting classes do `ToUpperInvariant`
* These facts are hard to follow

Let's 
* Add `StringComparer.InvariantCultureIgnoreCase` to that `HashSet` constructor argument to have case insensitive matching
* Remove `ToUpperInvariant` from C# and VB

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -20,74 +20,130 @@
...
 
-                c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, stringLiteral.GetLocation(), literalValue));
-            };
+            var literalValue = GetValueText(stringLiteral);
+
+            if (!IPAddress.TryParse(literalValue, out var address))
+            {
+                return;
+            }
+
```
pavel-mikula-sonarsource, line 128: Manual checking for loopback is tricky. Did you tried `IPAddress.IsLoopback(address)` instead? Are there any issues with that?

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -20,74 +20,130 @@
...
+        private static bool DoesItLookLikeObjectIdentifier(string s) =>
+            s.StartsWith("2.5.");
+
         protected const string DiagnosticId = "S1313";
+
         protected const string MessageFormat = "Make sure using this hardcoded IP address '{0}' is safe here.";
 
-        private static readonly ISet<string> IgnoredVariableNames =
+        private const int NumOfIPv4AddressParts = 4;
```
pavel-mikula-sonarsource, line 49: Just to simplify the name. It's integer, so it's obvious that it's a count
```suggestion
        private const int IPv4AddressParts = 4;
```

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -20,74 +20,130 @@
...
+            context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, stringLiteral.GetLocation(), literalValue));
+        }
+
+        private bool IsLoopbackAddress(string ip) =>
+            ip.StartsWith("127.") || ipv6LoopbackPattern.IsMatch(ip);
 
-        private static bool IsLocalHost(string s) =>
-            s == "::" || s == "127.0.0.1";
+        private bool IsNonRoutableAddress(string ip) =>
+             ip.Equals("0.0.0.0") || ipv6NonRoutablePattern.IsMatch(ip);
```
pavel-mikula-sonarsource, line 147:
```suggestion
             ip == "0.0.0.0" || ipv6NonRoutablePattern.IsMatch(ip);
```

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -20,74 +20,130 @@
...
+        private const int NumOfIPv4AddressParts = 4;
+
+        private readonly ISet<string> ignoredVariableNames =
             new HashSet<string>
             {
                 "VERSION",
                 "ASSEMBLY",
             };
 
+        private readonly Regex ipv6LoopbackPattern = new Regex("^(?:0*:)*?:?0*1$", RegexOptions.Compiled);
```
pavel-mikula-sonarsource, line 58: `^(?:0*:)*?:?0*1$` Original
`^(0*:)*?:?0*1$` Why using noncapturing groups when we don't capture anything? 
`^(0*:)*:?0*1$` As we fix `begin`, `end` and have defined content, this should behave the same

But it still matches many inputs that are not IPv6 address. Like `:::::1`, `::000001`, `0::0::1` and so on. Let's get rid of this if possible with the other comment about `IPAddress.IsLoopback`

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -20,74 +20,130 @@
...
+
+        private readonly Regex ipv6NonRoutablePattern = new Regex("^(?:0*:)*?:?0*$", RegexOptions.Compiled);
```
pavel-mikula-sonarsource, line 60: This regex is hard to build right. Can you use `address.GetAddressBytes().All(x => x==0)` instead? It will cover the IPv4 case as well. Or any other address family.

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -20,74 +20,130 @@
...
+            {
+                return;
+            }
+
+            if (IsLoopbackAddress(literalValue) || IsNonRoutableAddress(literalValue) || IsBroadcast(literalValue) || DoesItLookLikeObjectIdentifier(literalValue))
+            {
+                return;
+            }
+
+            if (address.AddressFamily == AddressFamily.InterNetwork
```
pavel-mikula-sonarsource, line 134: Could be single line, as it's short and boring

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -20,74 +20,130 @@
...
+                && literalValue.Split('.').Length != NumOfIPv4AddressParts)
```
pavel-mikula-sonarsource, line 135: This should be cheaper, since we won't generate the intermediate string array
```suggestion
                && literalValue.Count(x => x == '.') != NumOfIPv4AddressParts - 1)
```

analyzers/src/SonarAnalyzer.CSharp/Rules/HardcodedIpAddress.cs
```diff
@@ -18,23 +18,30 @@
...
 {
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     [Rule(DiagnosticId)]
-    public sealed class HardcodedIpAddress : HardcodedIpAddressBase<LiteralExpressionSyntax>
+    public sealed class HardcodedIpAddress : HardcodedIpAddressBase<SyntaxKind, LiteralExpressionSyntax>
     {
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager)
-                .WithNotConfigurable();
+        private static bool IsVariableIdentifier(SyntaxNode syntaxNode) =>
```
pavel-mikula-sonarsource, line 35: This should be left at the end. We'll deactivate the SA1204

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -20,74 +20,130 @@
...
+            {
+                return;
+            }
+
+            context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, stringLiteral.GetLocation(), literalValue));
+        }
+
+        private bool IsLoopbackAddress(string ip) =>
```
pavel-mikula-sonarsource, line 143: This and the below could be static, as it depends on static state only (while changing the regexes to static, as they do not depend on the instance)

analyzers/src/SonarAnalyzer.VisualBasic/Rules/HardcodedIpAddress.cs
```diff
@@ -31,10 +30,18 @@ namespace SonarAnalyzer.Rules.VisualBasic
 {
     [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
     [Rule(DiagnosticId)]
-    public sealed class HardcodedIpAddress : HardcodedIpAddressBase<LiteralExpressionSyntax>
+    public sealed class HardcodedIpAddress : HardcodedIpAddressBase<SyntaxKind, LiteralExpressionSyntax>
     {
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        private static bool IsVariableIdentifier(SyntaxNode syntaxNode) =>
```
pavel-mikula-sonarsource, line 35: Same here, SA1204 should be ignored in this case

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/HardcodedIpAddress.cs
```diff
@@ -29,26 +29,49 @@ public HardcodedIpAddress()
...
+            string broadcastAddress = "255.255.255.255"; //Compliant
+            string loopbackAddress1 = "127.0.0.1"; //Compliant
+            string loopbackAddress2 = "127.2.3.4"; //Compliant
+            string nonRoutableAddress = "0.0.0.0"; //Compliant
+            string notAnIp1 = "0.0.0.1234"; //Compliant
+            string country_oid = "2.5.6.2"; //Compliant 
+            string subschema_oid = "2.5.20.1";
+            string not_considered_as_an_oid = "2.59.6.2"; // Noncompliant
+//                                            ^^^^^^^^^^
+
```
pavel-mikula-sonarsource, line 63: Let's add few cases with invalid IPv6 strings (too many digits, more than one `::`).

pavel-mikula-sonarsource (changes requested):
First batch :)

pavel-mikula-sonarsource (changes requested):
GitHub UI tricked you and some comments were left out :)

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -30,64 +30,86 @@
...
-        private static bool IsLocalHost(string s) =>
-            s == "::" || s == "127.0.0.1";
+            if (HasAttributes(stringLiteral))
+            {
+                return;
+            }
+
+            var literalValue = GetValueText(stringLiteral);
+
+            IPAddress address;
```
pavel-mikula-sonarsource, line 92: The original `out var address` was better. We're using `out var` a lot. This costs us 3 lines of code

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -30,64 +30,86 @@
...
+
+            if (!IPAddress.TryParse(literalValue, out address))
```
pavel-mikula-sonarsource, line 94: This and the following two `if` could be merged to single sequence. The final `|| (... && ...)` is acceptable.

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -30,64 +30,86 @@
...
+            {
+                return;
+            }
+
+            if (IPAddress.IsLoopback(address)
+                || address.GetAddressBytes().All(x => x == 0)
```
pavel-mikula-sonarsource, line 100: This line deserves a short explanation now, as it's not that clear without the original variable names
```suggestion
                || address.GetAddressBytes().All(x => x == 0) // Nonroutable 0.0.0.0 or 0::0
```

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -30,64 +30,86 @@
...
+                || literalValue == IPv4Broadcast
+                || literalValue.StartsWith("2.5."))
```
pavel-mikula-sonarsource, line 102: This line deserves a short explanation now, as it's not that clear without the original variable names
```suggestion
                || literalValue.StartsWith("2.5."))  // Looks like OID
```

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -30,64 +30,86 @@
...
+        private readonly string[] ignoredVariableNames =
             {
                 "VERSION",
                 "ASSEMBLY",
             };
 
+        private readonly DiagnosticDescriptor rule;
+
+        protected abstract GeneratedCodeRecognizer GeneratedCodeRecognizer { get; }
+
```
pavel-mikula-sonarsource, line 51: I would remove this empty line, as it's a block of abstract properties (easier to read)

analyzers/src/SonarAnalyzer.Common/Rules/HardcodedIpAddressBase.cs
```diff
@@ -30,64 +30,86 @@
...
+        protected abstract TSyntaxKind SyntaxKind { get; }
+
         protected abstract string GetAssignedVariableName(TLiteralExpression stringLiteral);
+
```
pavel-mikula-sonarsource, line 55: I would remove this and the following empty line, as it's a block of abstract methods (easier to read)

pavel-mikula-sonarsource (approved):
LGTM, please take a look at those final nitpicks before merging

PR #3877, opened 2020-12-29
Fix S4830 FN: Match DangerousAcceptAnyServerCertificateValidator

### LINKED ISSUE #3824: Fix S4830 FN: Match DangerousAcceptAnyServerCertificateValidator
### Description

[`HttpClientHandler.DangerousAcceptAnyServerCertificateValidator`](https://docs.microsoft.com/dotnet/api/system.net.http.httpclienthandler.dangerousacceptanyservercertificatevalidator) was added in dotnet/corefx#19908 (.NET Core 2.0) to have the same behavior as a delegate returning true, including on backends that do not support callbacks.  S4830 does not trigger when this property is used instead of a delegate, even though they have the same effect.

### Repro steps

```csharp
using System.Net.Http;

namespace NoServerCertValidation
{
    public static class NoServerCertValidationTest
    {
	public static void WithHandlerDelegate()
	{
	    new HttpClientHandler().ServerCertificateCustomValidationCallback =
		(sender, certificate, chain, sslErrors) => true;
	}

	public static void WithHandlerProp()
	{
	    new HttpClientHandler().ServerCertificateCustomValidationCallback =
		HttpClientHandler.DangerousAcceptAnyServerCertificateValidator;
	}
    }
}
```

### Expected behavior

S4830 on both line 9 and 15.

### Actual behavior

S4830 on line 9 only.

### Related information

Using [SonarAnalyzer.CSharp](https://www.nuget.org/packages/SonarAnalyzer.CSharp/) version 8.15.0.24505 building with .NET 5.0:

<details>
<summary><code>dotnet --info</code></summary>

    .NET SDK (reflecting any global.json):
     Version:   5.0.101
     Commit:    d05174dc5a
    
    Runtime Environment:
     OS Name:     debian
     OS Version:  
     OS Platform: Linux
     RID:         debian-x64
     Base Path:   /usr/share/dotnet/sdk/5.0.101/
    
    Host (useful for support):
      Version: 5.0.1
      Commit:  b02e13abab
    
    .NET SDKs installed:
      5.0.101 [/usr/share/dotnet/sdk]
    
    .NET runtimes installed:
      Microsoft.AspNetCore.App 5.0.1 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
      Microsoft.NETCore.App 5.0.1 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
    
    To install additional .NET runtimes or SDKs:
      https://aka.ms/dotnet-download

</details>

pavel-mikula-sonarsource:
Fixes #3824

### DIFF
--- a/analyzers/src/SonarAnalyzer.VisualBasic/Rules/CertificateValidationCheck.cs
+++ b/analyzers/src/SonarAnalyzer.VisualBasic/Rules/CertificateValidationCheck.cs
@@ -48,6 +48,14 @@ namespace SonarAnalyzer.Rules.VisualBasic
     {
         public CertificateValidationCheck() : base(RspecStrings.ResourceManager) { }
 
+        internal override AbstractMethodParameterLookup<ArgumentSyntax> CreateParameterLookup(SyntaxNode argumentListNode, IMethodSymbol method) =>
+            argumentListNode switch
+            {
+                InvocationExpressionSyntax invocation => new VisualBasicMethodParameterLookup(invocation.ArgumentList, method),
+                ObjectCreationExpressionSyntax ctor => new VisualBasicMethodParameterLookup(ctor.ArgumentList, method),
+                _ => throw new ArgumentException("Unexpected type.", nameof(argumentListNode)) // This should be throw only by bad usage of this method, not by input dependency
+            };
+
         protected override void Initialize(SonarAnalysisContext context)
         {
             // C# += equivalent:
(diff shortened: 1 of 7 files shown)


pavel-mikula-sonarsource:
We'll merge this with failed QG. Coverage is failing due to old defensive code

analyzers/src/SonarAnalyzer.VisualBasic/Rules/CertificateValidationCheck.cs
```diff
@@ -34,74 +34,66 @@ namespace SonarAnalyzer.Rules.VisualBasic
...
+            };
+
         protected override void Initialize(SonarAnalysisContext context)
         {
-            //C# += equivalent:
-            //AddHandler situation does not exist in VB.NET. Delegate is pointer to one function only and can not have multiple handlers. It's not an event.
-            //Only assignment and object creation are valid cases for VB.NET
+            // C# += equivalent:
+            // AddHandler situation does not exist in VB.NET. Delegate is pointer to one function only and can not have multiple handlers. It's not an event.
+            // Only assignment and object creation are valid cases for VB.NET
```
csaba-sagi-sonarsource, line 63: Is this comment really needed?
Actually wondering if any of these  handling of comments are needed in both c# and vbnet as the code itself is pretty self descriptive.

csaba-sagi-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.VisualBasic/Rules/CertificateValidationCheck.cs
pavel-mikula-sonarsource: These comments were there from original version. I added them back then because I was the only one knowing this out of the box. I find it still useful for non-VB experts :)

pavel-mikula-sonarsource: ..it explains why is the C# registration different from VB

PR #3878, opened 2020-12-29
Fix S4784 FN: Process const regex

### LINKED ISSUE #3298: Fix S4784 FN: Process const regex
### Description

Security hotspot **S4784 : Using regular expressions is security-sensitive** is not detected when regex value is a const.

### Repro steps

```csharp
    public static class ReDoSAttack
    {
        private const string UnsafeRegex = @"^([(?>\.\-)*|\w]+)@\w+(?>(([\.-]?\w+)(?!$)))*(\.\w{2,3})+$";

        public static bool IsValid(string input)
        {
            var regex = new Regex(UnsafeRegex);
            return regex.IsMatch(input);
        }
    }
```

### Expected behavior

A new security hotspot is detected.

### Actual behavior

Security hotspot is not detected.

### Known workarounds

Hard code regex value into the constructor.

### Related information

* Visual Studio 2019 (16.5.4)
* dotnet version 3.1
* SonarScanner 4.9 for MSBuild 4.8
* Microsoft Windows Server 2019 Standard 10.0.17763

Could I provide anythings else to help with my issue ?
Thanks

pavel-mikula-sonarsource:
Fixes #3298

### DIFF
--- a/analyzers/its/expected/Ember-MM/scraper.EmberCore-{EF6A550E-DD76-4F4D-8250-8598140F828B}-S4784.json
+++ b/analyzers/its/expected/Ember-MM/scraper.EmberCore-{EF6A550E-DD76-4F4D-8250-8598140F828B}-S4784.json
@@ -32,10 +32,166 @@
 "location":  {
 "uri":  "sources\Ember-MM\Addons\scraper.EmberCore\Scraper\clsScrapeIMDB.vb",
 "region":  {
+"startLine":  222,
+"startColumn":  56,
+"endLine":  222,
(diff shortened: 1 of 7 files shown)


csaba-sagi-sonarsource:
The UsingRegularExpressionsTest class is violating rule SA1209UsingAliasDirectivesMustBePlacedAfterOtherUsingDirectives, so maybe it would be a good idea to clean up that file too if we are working on that rule

csaba-sagi-sonarsource (approved):
LGTM

PR #3882, opened 2020-12-30
Rule S2077 concatenated value tracking (#3292)

### LINKED ISSUE #3292: Fix S2077 FN track concatenated values
With https://github.com/SonarSource/sonar-dotnet/pull/3106 variables with string interpolation values "are tracked" :metal:

```
string sensitiveQuery = string.Format("INSERT INTO Users (name) VALUES (\"{0}\")", param); //  SQL query with string interpolation
string oString3 = sensitiveQuery; // assignment
SqlCommand command = new SqlCommand(oString3); // Noncompliant
command.CommandText = oString3; // Noncompliant
```

However variables with string concatenation values and used in sql queries do not trigger the rule: 
```
string oString2 = "SELECT * FROM Table1 WHERE col1 = '" + param + "'"; //  SQL query with string concatenation
SqlCommand oCmd2 = new SqlCommand(oString2, myConnection); // FN
oCmd2.CommandText = oString2; // FN
```

Related community thread:
https://community.sonarsource.com/t/owasp-rules-not-applied-on-c-and-vb-net-code/22919

Eric

csaba-sagi-sonarsource:
Fixes #3292

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
@@ -68,9 +68,7 @@ namespace SonarAnalyzer.Rules.CSharp
         protected override bool IsConcat(ExpressionSyntax argument, SemanticModel semanticModel) =>
             IsStringMethodInvocation("Concat", argument, semanticModel)
             || IsConcatenation(argument, semanticModel)
-            || (argument is IdentifierNameSyntax identifierNameSyntax
-                && semanticModel.GetDeclaringSyntaxNode(identifierNameSyntax) is VariableDeclaratorSyntax variableDeclaratorSyntax
-                && IsConcat(variableDeclaratorSyntax.Initializer?.Value, semanticModel));
+            || IsVariableDeclarationWithConcat(argument, semanticModel);
 
         protected override bool IsInterpolated(ExpressionSyntax expression, SemanticModel semanticModel) =>
             expression switch
@@ -92,6 +90,11 @@ namespace SonarAnalyzer.Rules.CSharp
             expression.IsKind(SyntaxKind.AddExpression)
             && expression is BinaryExpressionSyntax concatenation
             && !IsConcatenationOfConstants(concatenation, semanticModel);
+
+        private bool IsVariableDeclarationWithConcat(ExpressionSyntax argument, SemanticModel semanticModel) =>
+            (argument is IdentifierNameSyntax identifierNameSyntax
+             && semanticModel.GetDeclaringSyntaxNode(identifierNameSyntax) is VariableDeclaratorSyntax variableDeclaratorSyntax
+             && IsConcat(variableDeclaratorSyntax.Initializer?.Value, semanticModel));
 
         private bool IsInterpolated(IdentifierNameSyntax identifier, SemanticModel semanticModel) =>
             semanticModel.GetDeclaringSyntaxNode(identifier) is VariableDeclaratorSyntax variableDeclaratorSyntax
@@ -117,15 +120,16 @@ namespace SonarAnalyzer.Rules.CSharp
                 var nestedBinary = nestedLeft as BinaryExpressionSyntax;
                 while (nestedBinary != null)
                 {
-                    if (!nestedBinary.Right.IsConstant(semanticModel)
-                        || (!nestedBinary.IsKind(SyntaxKind.AddExpression)
-                            && !nestedBinary.IsConstant(semanticModel)))
+                    if (nestedBinary.Right.IsConstant(semanticModel)
+                        && (nestedBinary.IsKind(SyntaxKind.AddExpression) || nestedBinary.IsConstant(semanticModel)))
+                    {
+                        nestedLeft = nestedBinary.Left;
+                        nestedBinary = nestedLeft as BinaryExpressionSyntax;
+                    }
+                    else
                     {
                         return false;
                     }
-
-                    nestedLeft = nestedBinary.Left;
-                    nestedBinary = nestedLeft as BinaryExpressionSyntax;
                 }
                 return nestedLeft.IsConstant(semanticModel);
             }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
@@ -66,11 +66,11 @@ namespace SonarAnalyzer.Rules.CSharp
                 : null;
 
         protected override bool IsTracked(ExpressionSyntax argument, SemanticModel semanticModel) =>
-            argument != null &&
-                (IsConcatenation(argument, semanticModel)
-                 || argument.IsKind(SyntaxKind.InterpolatedStringExpression)
-                 || (argument is InvocationExpressionSyntax invocation && IsInvocationOfInterest(invocation, semanticModel))
-                 || IsTrackedVariableDeclaration(argument, semanticModel));
+            argument != null
+            && (IsConcatenation(argument, semanticModel)
+                || argument.IsKind(SyntaxKind.InterpolatedStringExpression)
+                || (argument is InvocationExpressionSyntax invocation && IsInvocationOfInterest(invocation, semanticModel))
+                || IsTrackedVariableDeclaration(argument, semanticModel));
 
         private static bool IsInvocationOfInterest(InvocationExpressionSyntax invocation, SemanticModel semanticModel) =>
             (invocation.IsMethodInvocation(KnownType.System_String, "Format", semanticModel) || invocation.IsMethodInvocation(KnownType.System_String, "Concat", semanticModel))
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
@@ -19,7 +19,6 @@
  */
 
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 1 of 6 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ExecutingSqlQueries_Net46.vb
```diff
@@ -109,7 +96,7 @@ Namespace Tests.Diagnostics
             command.CommandText = "SELECT * FROM mytable WHERE mycol=" & param ' Noncompliant
             command.CommandText = $"SELECT * FROM mytable WHERE mycol={param}" ' Noncompliant
             Dim adapter = New OdbcDataAdapter(String.Format("INSERT INTO Users (name) VALUES (""{0}"")", param), "") ' Noncompliant
-            Dim adapter1 = New odbcdataadapter(sTRing.foRmAt("INSERT INTO Users (name) VALUES (""{0}"")", param), "") ' Noncompliant
+            Dim adapter1 = New odbcdataadapter(String.Format("INSERT INTO Users (name) VALUES (""{0}"")", param), "") ' Noncompliant
```
pavel-mikula-sonarsource, line 99: This is little hard to maintain, but the CraZY caSINg should be preserved to demonstrace case insensivity for VB.NET

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ExecutingSqlQueries_Net46.vb
```diff
@@ -134,7 +121,7 @@ Namespace Tests.Diagnostics
         Public Sub NonCompliant_OracleCommands(ByVal connection As OracleConnection, ByVal transaction As OracleTransaction, ByVal query As String, ByVal param As String)
             Dim command = New OracleCommand("SELECT * FROM mytable WHERE mycol=" & param) ' Noncompliant
             command.CommandText = String.Format("INSERT INTO Users (name) VALUES (""{0}"")", param) ' Noncompliant
-            command.CommandText = string.forMAT("INSERT INTO Users (name) VALUES (""{0}"")", param) ' Noncompliant
+            command.CommandText = String.Format("INSERT INTO Users (name) VALUES (""{0}"")", param) ' Noncompliant
```
pavel-mikula-sonarsource, line 124: Same here

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ExecutingSqlQueries_Net46.vb
```diff
@@ -164,5 +151,36 @@ Namespace Tests.Diagnostics
             command.CommandText = String.Format("INSERT INTO Users (name) VALUES (""{0}"")", param) ' Noncompliant
         End Sub
 
+        Public Sub Bar(ByVal connection As SqlConnection, ByVal param As String)
```
pavel-mikula-sonarsource, line 154: To follow the style of previous methods. (I personally don't like `Foo` & `Bar` unless it has absolutely no meaning for the test, like an outer class name when the test is related to a method)
```suggestion
        Public Sub ConcatAndStringFormat(ByVal connection As SqlConnection, ByVal param As String)
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ExecutingSqlQueries_Net46.cs
```diff
@@ -193,6 +193,14 @@ public void Bar(SqlConnection connection, string param)
             string stillSensitive = sensitiveQuery;
             command.CommandText = stillSensitive; // Noncompliant
 
+            string sensitiveConcatQuery = "SELECT * FROM Table1 WHERE col1 = '" + param + "'";
+            command = new SqlCommand(sensitiveConcatQuery); // Noncompliant
+
```
pavel-mikula-sonarsource, line 198: Does this behaves the same? Let's add a test cases.
```
command = new SqlCommand("SELECT * FROM Table1 WHERE col1 = '" + param + "'"); // Noncompliant
command.CommandText = "SELECT * FROM Table1 WHERE col1 = '" + param + "'"; // Noncompliant

```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/ExecutingSqlQueriesBase.cs
```diff
@@ -30,7 +31,8 @@ public abstract class ExecutingSqlQueriesBase<TSyntaxKind, TExpressionSyntax> :
         protected const string DiagnosticId = "S2077";
         protected const string MessageFormat = "Make sure that formatting this SQL query is safe here.";
 
-        private readonly KnownType[] constructors = {
+        private readonly KnownType[] constructors =
+            {
```
pavel-mikula-sonarsource, line 35: _Formatting:_ Having block and it's content aligned on the same line doesn't seem right

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/ExecutingSqlQueriesBase.cs
```diff
@@ -30,7 +31,8 @@ public abstract class ExecutingSqlQueriesBase<TSyntaxKind, TExpressionSyntax> :
         protected const string DiagnosticId = "S2077";
         protected const string MessageFormat = "Make sure that formatting this SQL query is safe here.";
```
pavel-mikula-sonarsource, line 32: This can be private now

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
```diff
@@ -101,29 +88,38 @@ expression switch
...
             {
                 var nestedLeft = binaryExpression.Left;
                 var nestedBinary = nestedLeft as BinaryExpressionSyntax;
                 while (nestedBinary != null)
                 {
-                    if (!nestedBinary.Right.IsConstant(semanticModel) ||
-                        (!nestedBinary.IsKind(SyntaxKind.AddExpression) && !nestedBinary.IsConstant(semanticModel)))
+                    if (!nestedBinary.Right.IsConstant(semanticModel)
+                        || (!nestedBinary.IsKind(SyntaxKind.AddExpression)
+                            && !nestedBinary.IsConstant(semanticModel)))
```
pavel-mikula-sonarsource, line 122: This one is seriously hard to read.
v1: Let's keep the `&&` on single line to see the main `||` logic clearly.
```suggestion
                    if (!nestedBinary.Right.IsConstant(semanticModel)
                        || (!nestedBinary.IsKind(SyntaxKind.AddExpression) && !nestedBinary.IsConstant(semanticModel)))
```
v2: Can we invert the condition? That should make it readable. Something like
```
if(right is constant && (nested is constant || nested is add))
  do the nesting
else
  return false
```

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/ExecutingSqlQueries.cs
```diff
@@ -125,8 +121,9 @@ private static bool IsConcatenationOfConstants(BinaryExpressionSyntax binaryExpr
                 var nestedBinary = nestedLeft as BinaryExpressionSyntax;
                 while (nestedBinary != null)
                 {
-                    if (!nestedBinary.Right.IsConstant(semanticModel) ||
-                        (!IsConcatenationOperator(nestedBinary) && !nestedBinary.IsConstant(semanticModel)))
+                    if (!nestedBinary.Right.IsConstant(semanticModel)
```
pavel-mikula-sonarsource, line 124: Same as C#

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
```diff
@@ -74,12 +66,11 @@ public ExecutingSqlQueries()
...
         protected override bool IsConcat(ExpressionSyntax argument, SemanticModel semanticModel) =>
-            IsStringMethodInvocation("Concat", argument, semanticModel) ||
-            (
-                argument.IsKind(SyntaxKind.AddExpression) &&
-                argument is BinaryExpressionSyntax concatenation &&
-                !IsConcatenationOfConstants(concatenation, semanticModel)
-            );
+            IsStringMethodInvocation("Concat", argument, semanticModel)
+            || IsConcatenation(argument, semanticModel)
+            || (argument is IdentifierNameSyntax identifierNameSyntax
```
pavel-mikula-sonarsource, line 71: While you've extracted 3 `&&` conditions to `IsConcatenation`, this new 3 `&&` condition would also make sense to be extracted to a method to keep the code structure regular.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
```diff
@@ -74,12 +66,11 @@ public ExecutingSqlQueries()
...
+                && semanticModel.GetDeclaringSyntaxNode(identifierNameSyntax) is VariableDeclaratorSyntax variableDeclaratorSyntax
```
pavel-mikula-sonarsource, line 72: This condition doesn't seem right. We're inside `IsConcat(ExpressionSyntax)` context. And the condition checks if the parent is variable declarator with initializer and recursion. That feels hacky.

I'd expect some `AddExpression` syntax to be checked here.

pavel-mikula-sonarsource (changes requested):
I think we'll need to reimplement the condition.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
pavel-mikula-sonarsource: Waaay better :) Now we can see what it does

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
pavel-mikula-sonarsource: This looks the same way as it did before. Am I missing something?

pavel-mikula-sonarsource: Let's refactor it. It's a nightmare to maintain.

The Base `IsConcat`, `IsFormat` and `IsInterpolated` should be merged to single `IsTracked`, while removing `IsStringMethodInvocation` indirection. That will also remove 6 of those private helpers.

The ObjectCreation and PropertyAccess can be then embedded directly into the tracker, as it's not reused (just rename the `context =>` to `c =>` for example)

On the C#/VB side, the variable declarator step interpolation can be decoupled and simplified as well.

pavel-mikula-sonarsource (approved):
Awesome! I just left two formatting comments

PR #3886, opened 2020-12-30
Improve S2068: Detect constant concatenation

### LINKED ISSUE #3802: Improve S2068: Detect constant concatenation
```
            using (var conn = OpenConn("Server = " + server + "; Database = Test; User = SA; Password =" + "Secret123")) { } // FN
            var mySecret = "123";
            using (var conn = OpenConn("Server = " + server + "; Database = Test; User = SA; Password =" + mySecret)) { } // FN
```

We are currently able to detect when a `private const string secret = "constantValue";` is passed (see where `secret` is used in [the test](https://github.com/SonarSource/sonar-dotnet/blob/50e3d6240212086a5ec9cfd7887bfbf265f79860/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHardcodeCredentials_DefaultValues.cs#L13)), but not with variables initialized with a constant.

```
            new NetworkCredential("username", secret); // Noncompliant {{Please review this hard-coded password.}}
```

This should probably be done in 2 steps:
- first fix the concatenation of hardcoded strings and with `const` fields or variables
- fix concatenation with variables which are not `const`

pavel-mikula-sonarsource:
Fixes #3802


csaba-sagi-sonarsource (approved):
LGTM

PR #3890, opened 2021-01-04
Improve S2068: Detect string.Format with constant values

### LINKED ISSUE #3884: Improve S2068: Detect string.Format with constant values
Spinoff from #3802

Add support for string.Format strings with constant values:
```
var a = String.Format("Server = localhost; Database = Test; User = SA; Password = {0}", secretConst); // FN
```

pavel-mikula-sonarsource:
Fixes #3884

This is a draft, rebase will be needed. Only commits from 69f85c2 needs review.

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -72,7 +73,7 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private class VariableDeclarationBannedWordsFinder : CredentialWordsFinderBase<VariableDeclaratorSyntax>
         {
-            public VariableDeclarationBannedWordsFinder(DoNotHardcodeCredentialsBase<SyntaxKind> analyzer) : base(analyzer) { }
+            public VariableDeclarationBannedWordsFinder(DoNotHardcodeCredentials analyzer) : base(analyzer) { }
 
             protected override string GetAssignedValue(VariableDeclaratorSyntax syntaxNode, SemanticModel semanticModel) =>
                 syntaxNode.Initializer?.Value.GetStringValue();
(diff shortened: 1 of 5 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
```diff
@@ -146,5 +160,68 @@ private static bool ShouldHandle(SyntaxNode method, SyntaxNode current)
...
+                catch (FormatException)
+                {
+                    return null;
+                }
+            }
+
+            protected override string GetVariableName(InvocationExpressionSyntax syntaxNode) => null;
+
+            protected override bool ShouldHandle(InvocationExpressionSyntax syntaxNode, SemanticModel semanticModel) =>
+                syntaxNode.IsMethodInvocation(KnownType.System_String, "Format", semanticModel)
```
csaba-sagi-sonarsource, line 222: Maybe it would be useful to create a helper extension method IsMethodInvocationWithParams. For me seems like something that could be useful in other places too, and that way we would not do semanticModel.GetSymbolInfo(syntaxNode).Symbol is IMethodSymbol symbol twice.

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/DoNotHardcodeCredentials.cs
```diff
@@ -152,5 +164,70 @@ private static bool ShouldHandle(SyntaxNode method, SyntaxNode current)
...
+                catch (FormatException)
+                {
+                    return null;
+                }
+            }
+
+            protected override string GetVariableName(InvocationExpressionSyntax syntaxNode) => null;
+
+            protected override bool ShouldHandle(InvocationExpressionSyntax syntaxNode, SemanticModel semanticModel) =>
+                syntaxNode.IsMethodInvocation(KnownType.System_String, "Format", semanticModel)
```
csaba-sagi-sonarsource, line 228: If we opt to create the extension, this can be replaced too.

csaba-sagi-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DoNotHardcodeCredentials.cs
pavel-mikula-sonarsource: I checked the other places and no one is checking the args (yet). So maybe later. It's closed closely to each other, so Roslyn internal symbol cache should deal with the performance penalty in this case.

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/DoNotHardcodeCredentials.cs
pavel-mikula-sonarsource: _Note:_ Second extension would needed to be created for VB, as they don't share argument types.

PR #3901, opened 2021-01-05
Rule S2077: support of more database libraries (#3293)

### LINKED ISSUE #3293: Rule S2077: support of more database libraries
Some database librairies like Mysql.Data are not supported:
```
MySqlConnection myConnection = new MySqlConnection(strConn);
string sensitiveString2 = "SELECT * FROM Table1 WHERE col1 = '" + param + "'";
MySql.Data.MySqlClient.MySqlHelper.ExecuteNonQuery(myConnection, sensitiveString2); // FN
```
Ideally S2077 should follow rule S3649 (all the sinks defined in S3649 should be supported in S2077)

Related discussions:
- https://github.com/SonarSource/sonar-dotnet/issues/2800#issuecomment-584036701
- https://jira.sonarsource.com/browse/RSPEC-4873?focusedCommentId=232648&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-232648

Eric

csaba-sagi-sonarsource:
Fixes #3293

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/ExecutingSqlQueriesBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/ExecutingSqlQueriesBase.cs
@@ -47,6 +47,9 @@ namespace SonarAnalyzer.Rules
                 KnownType.MySql_Data_MySqlClient_MySqlCommand,
                 KnownType.MySql_Data_MySqlClient_MySqlDataAdapter,
                 KnownType.MySql_Data_MySqlClient_MySqlScript,
+                KnownType.Microsoft_Data_Sqlite_SqliteCommand,
+                KnownType.System_Data_Sqlite_SqliteCommand,
+                KnownType.System_Data_Sqlite_SQLiteDataAdapter,
             };
 
         private readonly KnownType[] constructorsWithSecondParam =
@@ -59,6 +62,11 @@ namespace SonarAnalyzer.Rules
                 new MemberDescriptor(KnownType.Microsoft_EntityFrameworkCore_RelationalDatabaseFacadeExtensions, "ExecuteSqlCommandAsync"),
                 new MemberDescriptor(KnownType.Microsoft_EntityFrameworkCore_RelationalDatabaseFacadeExtensions, "ExecuteSqlCommand"),
                 new MemberDescriptor(KnownType.Microsoft_EntityFrameworkCore_RelationalQueryableExtensions, "FromSql"),
+            };
+
+        private readonly MemberDescriptor[] invocationsWithFirstParam =
+            {
+                new MemberDescriptor(KnownType.System_Data_Sqlite_SqliteCommand, "Execute"),
             };
 
         private readonly MemberDescriptor[] invocationsWithSecondParam =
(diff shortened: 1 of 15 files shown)


analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/ExecutingSqlQueriesBase.cs
```diff
@@ -41,13 +43,63 @@ public abstract class ExecutingSqlQueriesBase<TSyntaxKind, TExpressionSyntax> :
...
+            };
+
+        private readonly MemberDescriptor[] properties =
+            {
+                new MemberDescriptor(KnownType.System_Data_Odbc_OdbcCommand, "CommandText"),
+                new MemberDescriptor(KnownType.System_Data_OracleClient_OracleCommand, "CommandText"),
+                new MemberDescriptor(KnownType.System_Data_SqlClient_SqlCommand, "CommandText"),
+                new MemberDescriptor(KnownType.System_Data_SqlServerCe_SqlCeCommand, "CommandText"),
+                new MemberDescriptor(KnownType.MySql_Data_MySqlClient_MySqlCommand, "CommandText"),
+                new MemberDescriptor(KnownType.Microsoft_Data_Sqlite_SqliteCommand, "CommandText"),
```
pavel-mikula-sonarsource, line 95: I think there's and issue with `System.Data.Sqllite` and that is, that there exists
`System.Data.Sqlite` for .NET Fx and .NET Core 2.1
`System.Data.SQLite` for .NET Core 3.1 and .NET 5
It's documented in sonar-security ITs and I'm not sure if current configuration supports both (probably not yet).

Can you take a look and if it's confirmed, then add tests for netcore with `SQLite`?

pavel-mikula-sonarsource (changes requested):
I left few minor comments and one about adding more for SQLite

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/ExecutingSqlQueriesBase.cs
csaba-sagi-sonarsource: I could not find the nuget package version with System.Data.Sqlite.

pavel-mikula-sonarsource: That was probably related to the `Microsoft.Data.Sqlite` => `System.Data.SQLite` move

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ExecutingSqlQueries_Net46.cs
```diff
@@ -106,7 +110,7 @@ public void OdbcCommands(OdbcConnection connection, OdbcTransaction transaction,
             OdbcDataAdapter adapter;
             adapter = new OdbcDataAdapter(); // Compliant
             adapter = new OdbcDataAdapter(command); // Compliant
-            adapter = new OdbcDataAdapter(query, ""); // Compliant
+            adapter = new OdbcDataAdapter(query, $" non constant {query}"); // Compliant
```
pavel-mikula-sonarsource, line 113: This is little confusing, as it looks that the type has two query parameters and it's not clear why we don't raise on the second one.
```suggestion
            adapter = new OdbcDataAdapter(query, $"concatenated connection string {query}"); // Compliant
```
or with new parameter in the method
```suggestion
            adapter = new OdbcDataAdapter(query, $"connection string {server}"); // Compliant
```

pavel-mikula-sonarsource: Applies to all

pavel-mikula-sonarsource (approved):
LGTM, please take a look at the last comment before merging

PR #3915, opened 2021-01-07
Happy New Year 2021!

### LINKED ISSUE #3896: Update license headers 2021
Use Maven parent 55 and update headers for java & C#

pavel-mikula-sonarsource:
Fixes #3896


analyzers/tests/SonarAnalyzer.UnitTest/Helpers/BlockIdMapTest.cs
```diff
@@ -1,23 +1,22 @@
-﻿extern alias csharp;
```
andrei-epure-sonarsource, line 1: I don't know why this was being used in the first place, but it's great to see it gone 🎉

analyzers/src/SonarAnalyzer.CFG/License.txt
```diff
@@ -1,7 +1,7 @@
 ﻿                   GNU LESSER GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
 
- Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
```
andrei-epure-sonarsource, line 4: I think this should be reverted - it's just the generic GNU Lesser GPL license

analyzers/src/SonarAnalyzer.Vsix/License.txt
```diff
@@ -1,7 +1,7 @@
 ﻿                   GNU LESSER GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
 
- Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
```
andrei-epure-sonarsource, line 4: revert

andrei-epure-sonarsource (changes requested):
please revert the GNU license changes

andrei-epure-sonarsource:
(and you can also clean the unused usings :) )

pavel-mikula-sonarsource:
The unused usings are old technical debt that I don't want to address in this PR. It just popped up because of reactivation of the rule.

andrei-epure-sonarsource (approved):
LGTM

PR #3922, opened 2021-01-07
update guava version

### LINKED ISSUE #3916: Update Guava test dependency
We are using the Guava library in `<scope>test</scope>` in our Java projects.

We should update it to

```
<dependency>
  <groupId>com.google.guava</groupId>
  <artifactId>guava</artifactId>
  <version>[24.1.1,)</version>
</dependency>
```

as the current version has a vulnerability. As we're only using it in tests, it's not urgent.

andrei-epure-sonarsource:
Fixes #3916

### DIFF
--- a/sonar-csharp-plugin/pom.xml
+++ b/sonar-csharp-plugin/pom.xml
@@ -100,7 +100,7 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>[24.1.1,)</version>
+      <version>30.1-jre</version>
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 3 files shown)


sonar-csharp-plugin/pom.xml
```diff
@@ -100,7 +100,7 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>19.0</version>
+      <version>[24.1.1,)</version>
```
pavel-mikula-sonarsource, line 103: Why not the [latest](https://github.com/google/guava/releases)? As also used in https://github.com/SonarSource/sonar-java/blob/26910434c68fd9ece1c0e6f898112c30833f2dad/pom.xml#L153-L157
```suggestion
      <version>30.1-jre</version>
```

pavel-mikula-sonarsource (changes requested):

pavel-mikula-sonarsource (approved):
LGTM

Issue #3910, opened 2021-01-06
S2589 FP and FN: Conditional access and pattern matching


jjolidon:
### Description

S2589 triggers a false positive for a null check after a first test with pattern matching.

### Repro steps

```cs
private void Test(Exception ex)
{
    if (ex?.Data["FOO"] is string myString)
    {
        // intentionally empty for the reproducer
    }
    else
    {
        if (ex != null) // This test triggers S2589
        {
            // intentionally empty for the reproducer
        }
    }
}
```

### Expected behavior

The above code should not trigger S2589

### Actual behavior

The above code triggers S2589

### Known workarounds

No known workaround

### Related information

* SonarAnalyzer.CSharp version: 8.13.0.25740
* Visual Studio version: 16.8.3
* MSBuild / dotnet version: 16.8.2.56705 / 5.0.101
* SonarScanner for .NET version (if used)
* Operating System: Win 10

pavel-mikula-sonarsource:
Hi @jjolidon ,

Thank you for reporting this case. I can confirm it as False Positive.

andrei-epure-sonarsource:
other conditional access FPs: #3416 ,  #2528, #2704

mary-georgiou-sonarsource:
Fixed by https://github.com/SonarSource/sonar-dotnet/pull/7750

PR #3924, opened 2021-01-08
S2589: Add FP and FN repro for #3910

pavel-mikula-sonarsource:
Reproduces #3910

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp7.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp7.cs
@@ -348,16 +348,6 @@ namespace Tests.Diagnostics
             }
         }
 
-        public void ElementAccess2(Exception ex)
-        {
-            if (ex?.Data["Key"] is string value)
-            {
-            }
-            else if (ex.Data != null) // FN
-            {
-            }
-        }
-
         public void MemberBinding1(Exception ex)
         {
             if (ex?.Message is string value)
@@ -372,6 +362,9 @@ namespace Tests.Diagnostics
         {
             if (ex?.Message is string value)
             {
+            }
+            else if (ex == null)    // Noncompliant FP
+            {                       // Secondary
             }
             else if (ex.Message != null) // FN
             {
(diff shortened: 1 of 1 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp7.cs
```diff
@@ -334,4 +334,48 @@ public int Foo()
...
+            {
+            }
+        }
+
+        public void MemberBinding2(Exception ex)
+        {
+            if (ex?.Message is string value)
+            {
+            }
+            else if (ex.Message != null) // FN
```
andrei-epure-sonarsource, line 376: on the `else` branch there can be 3 cases:
1. `ex` is null
2. `ex` is not null, `ex.Message` is  null
3. `ex` is not null, `ex.Message` is  not null and not string

Given that `Message` may be either null or not null, a null check makes sense

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp7.cs
```diff
@@ -334,4 +334,48 @@ public int Foo()
...
+            {
+            }
+        }
+
+        public void ElementAccess2(Exception ex)
+        {
+            if (ex?.Data["Key"] is string value)
+            {
+            }
+            else if (ex.Data != null) // FN
```
andrei-epure-sonarsource, line 356: in this case, this is relying on the fact that if ex.Data is null, it will throw an exception at line 353, right? 

could you please add a comment with this, because I have to use pen and paper to figure it out?

andrei-epure-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp7.cs
pavel-mikula-sonarsource: `Message` is of type string, therefore `not string` is not a concern.
If you enter `ex?.Message is string value` branch, `Message` is definitely not null. 
In else branch, `ex` might be null, resulting in null reference exception. But `ex.Message` is definitely null then, therefore FN.

I'll add the `ex == null` check.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp7.cs
pavel-mikula-sonarsource: Good point, this is actually a bad test case.

andrei-epure-sonarsource (approved):
LGTM

PR #3926, opened 2021-01-08
Rule S2077: add secondary locations when tracking values (#3294)

### LINKED ISSUE #3294: Rule S2077: add secondary locations when tracking values
When tracking values (https://github.com/SonarSource/sonar-dotnet/issues/3292) the developer can be lost identifying where the SQL query come from:

```
string sensitiveQuery = string.Format("INSERT INTO Users (name) VALUES (\"{0}\")", param); //  initialization: secondary location
//...
//...
//...
//...
//...
//...
string oString3 = sensitiveQuery; // assignment: secondary location
//...
//...
//...
//...
//...
//...
SqlCommand command = new SqlCommand(oString3); // Noncompliant: main location
command.CommandText = oString3; // Noncompliant: main location
```

Secondary locations should be added when initialization/assignments involve the highlighted SQL query variable when querying database (main location).

Eric

csaba-sagi-sonarsource:
Fixes #3294

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
@@ -31,10 +31,8 @@ namespace SonarAnalyzer.Rules.CSharp
 {
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     [Rule(DiagnosticId)]
-    public sealed class ExecutingSqlQueries : ExecutingSqlQueriesBase<SyntaxKind, ExpressionSyntax>
+    public sealed class ExecutingSqlQueries : ExecutingSqlQueriesBase<SyntaxKind, ExpressionSyntax, IdentifierNameSyntax>
     {
-        private static readonly AssignmentFinder AssignmentFinder = new CSharpAssignmentFinder();
-
         public ExecutingSqlQueries() : this(AnalyzerConfiguration.Hotspot) { }
 
         internal /*for testing*/ ExecutingSqlQueries(IAnalyzerConfiguration analyzerConfiguration) : base(RspecStrings.ResourceManager)
@@ -42,7 +40,11 @@ namespace SonarAnalyzer.Rules.CSharp
             InvocationTracker = new CSharpInvocationTracker(analyzerConfiguration, Rule);
             PropertyAccessTracker = new CSharpPropertyAccessTracker(analyzerConfiguration, Rule);
             ObjectCreationTracker = new CSharpObjectCreationTracker(analyzerConfiguration, Rule);
+            AssignmentFinder = new CSharpAssignmentFinder();
         }
+
+        protected override string GetIdentifierName(IdentifierNameSyntax identifierNameSyntax) =>
+            identifierNameSyntax.Identifier.ValueText;
 
         protected override ExpressionSyntax GetInvocationExpression(SyntaxNode expression) =>
             expression is InvocationExpressionSyntax invocation
@@ -67,11 +69,27 @@ namespace SonarAnalyzer.Rules.CSharp
         protected override bool IsTracked(ExpressionSyntax expression, BaseContext context) =>
             IsSensitiveExpression(expression, context.SemanticModel) || IsTrackedVariableDeclaration(expression, context);
 
-        private static bool IsSensitiveExpression(ExpressionSyntax expression, SemanticModel semanticModel) =>
+        protected override bool IsSensitiveExpression(ExpressionSyntax expression, SemanticModel semanticModel) =>
             expression != null
             && (IsConcatenation(expression, semanticModel)
             || expression.IsKind(SyntaxKind.InterpolatedStringExpression)
             || (expression is InvocationExpressionSyntax invocation && IsInvocationOfInterest(invocation, semanticModel)));
+
+        protected override Location SecondaryLocationForExpression(ExpressionSyntax node, string identifierName)
+        {
+            if (node == null)
+            {
+                return Location.None;
+            }
+
+            if (node.Parent is EqualsValueClauseSyntax equalsValueClause
+                && equalsValueClause.Parent is VariableDeclaratorSyntax declarationSyntax)
+            {
+                return declarationSyntax.Identifier.GetLocation();
+            }
+
+            return node.Parent is AssignmentExpressionSyntax assignment ? assignment.Left.GetLocation() : Location.None;
+        }
 
         private static bool IsInvocationOfInterest(InvocationExpressionSyntax invocation, SemanticModel semanticModel) =>
             (invocation.IsMethodInvocation(KnownType.System_String, "Format", semanticModel) || invocation.IsMethodInvocation(KnownType.System_String, "Concat", semanticModel))
@@ -81,57 +99,6 @@ namespace SonarAnalyzer.Rules.CSharp
             expression.IsKind(SyntaxKind.AddExpression)
             && expression is BinaryExpressionSyntax concatenation
             && !IsConcatenationOfConstants(concatenation, semanticModel);
-
-        private static bool IsTrackedVariableDeclaration(ExpressionSyntax expression, BaseContext context)
-        {
-            if (expression is IdentifierNameSyntax)
-            {
-                var node = expression;
-                Location location;
-                do
-                {
-                    var identifierName = (node as IdentifierNameSyntax).Identifier.ValueText;
-                    node = AssignmentFinder.FindLinearPrecedingAssignmentExpression(identifierName, node) as ExpressionSyntax;
-                    location = GetSecondaryLocationForExpression(node);
-
-                    if (IsSensitiveExpression(node, context.SemanticModel))
-                    {
-                        if (location != Location.None)
-                        {
-                            context.AddSecondaryLocation(new SecondaryLocation(location, string.Format(AssignmentWithFormattingMessage, identifierName)));
-                        }
-
-                        return true;
-                    }
-                    else
-                    {
-                        if (location != Location.None)
-                        {
-                            context.AddSecondaryLocation(new SecondaryLocation(location, string.Format(AssignmentMessage, identifierName)));
-                        }
-                    }
-                }
-                while (node is IdentifierNameSyntax);
-            }
-
-            return false;
-        }
-
-        private static Location GetSecondaryLocationForExpression(ExpressionSyntax node)
-        {
-            if (node == null)
-            {
-                return Location.None;
-            }
-
-            if (node.Parent is EqualsValueClauseSyntax equalsValueClauseSyntax
-                && equalsValueClauseSyntax.Parent is VariableDeclaratorSyntax declarationSyntax)
-            {
-                return declarationSyntax.Identifier.GetLocation();
-            }
-
-            return node.Parent is AssignmentExpressionSyntax assignment ? assignment.Left.GetLocation() : Location.None;
-        }
 
         private static bool AllConstants(IEnumerable<ArgumentSyntax> arguments, SemanticModel semanticModel) =>
             arguments.All(a => a.Expression.HasConstantValue(semanticModel));
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
@@ -50,17 +50,17 @@ namespace SonarAnalyzer.Rules.CSharp
                 : null;
 
         protected override ExpressionSyntax GetArgumentAtIndex(InvocationContext context, int index) =>
-            context.Invocation is InvocationExpressionSyntax invocation
+            context.Node is InvocationExpressionSyntax invocation
                 ? invocation.ArgumentList.Get(index)
                 : null;
 
         protected override ExpressionSyntax GetArgumentAtIndex(ObjectCreationContext context, int index) =>
-            context.Expression is ObjectCreationExpressionSyntax objectCreation
+            context.Node is ObjectCreationExpressionSyntax objectCreation
                 ? objectCreation.ArgumentList.Get(index)
                 : null;
 
         protected override ExpressionSyntax GetSetValue(PropertyAccessContext context) =>
-            context.Expression is MemberAccessExpressionSyntax setter && setter.IsLeftSideOfAssignment()
+            context.Node is MemberAccessExpressionSyntax setter && setter.IsLeftSideOfAssignment()
                 ? ((AssignmentExpressionSyntax)setter.GetSelfOrTopParenthesizedExpression().Parent).Right.RemoveParentheses()
                 : null;
 
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
@@ -94,14 +94,21 @@ namespace SonarAnalyzer.Rules.CSharp
                     node = AssignmentFinder.FindLinearPrecedingAssignmentExpression(identifierName, node) as ExpressionSyntax;
                     location = GetSecondaryLocationForExpression(node);
 
-                    if (location != Location.None)
-                    {
-                        context.AddSecondaryLocation(location);
-                    }
-
                     if (IsSensitiveExpression(node, context.SemanticModel))
                     {
+                        if (location != Location.None)
+                        {
+                            context.AddSecondaryLocation(new SecondaryLocation(location, string.Format(AssignmentWithFormattingMessage, identifierName)));
+                        }
+
                         return true;
+                    }
+                    else
+                    {
+                        if (location != Location.None)
+                        {
+                            context.AddSecondaryLocation(new SecondaryLocation(location, string.Format(AssignmentMessage, identifierName)));
+                        }
                     }
                 }
                 while (node is IdentifierNameSyntax);
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
@@ -66,7 +66,7 @@ namespace SonarAnalyzer.Rules.CSharp
                 ? ((AssignmentExpressionSyntax)setter.GetSelfOrTopParenthesizedExpression().Parent).Right.RemoveParentheses()
                 : null;
 
-        protected override bool IsTracked(ExpressionSyntax expression, BaseContext context) =>
+        protected override bool IsTracked(ExpressionSyntax expression, SyntaxBaseContext context) =>
             IsSensitiveExpression(expression, context.SemanticModel) || IsTrackedVariableDeclaration(expression, context);
 
         protected override bool IsSensitiveExpression(ExpressionSyntax expression, SemanticModel semanticModel) =>
@@ -75,8 +75,10 @@ namespace SonarAnalyzer.Rules.CSharp
             || expression.IsKind(SyntaxKind.InterpolatedStringExpression)
             || (expression is InvocationExpressionSyntax invocation && IsInvocationOfInterest(invocation, semanticModel)));
 
-        protected override Location SecondaryLocationForExpression(ExpressionSyntax node, string identifierName)
+        protected override Location SecondaryLocationForExpression(ExpressionSyntax node, string identifierNameToFind, out string identifierNameFound)
         {
+            identifierNameFound = identifierNameToFind;
+
             if (node == null)
             {
                 return Location.None;
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
@@ -33,10 +33,9 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class ExecutingSqlQueries : ExecutingSqlQueriesBase<SyntaxKind, ExpressionSyntax>
     {
-        public ExecutingSqlQueries()
-            : this(AnalyzerConfiguration.Hotspot)
-        {
-        }
+        private static readonly AssignmentFinder AssignmentFinder = new CSharpAssignmentFinder();
+
+        public ExecutingSqlQueries() : this(AnalyzerConfiguration.Hotspot) { }
 
         internal /*for testing*/ ExecutingSqlQueries(IAnalyzerConfiguration analyzerConfiguration) : base(RspecStrings.ResourceManager)
         {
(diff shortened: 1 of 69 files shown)


analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/SyntaxTrackerBase.cs
```diff
@@ -24,29 +24,42 @@
...
 
-        protected bool IsEnabled(AnalyzerOptions options)
-        {
-            analyzerConfiguration.Initialize(options);
+            void TrackAndReportIfNecessary(SyntaxNodeAnalysisContext c)
+            {
+                var baseContext = IsTracked(c.Node, c.SemanticModel, conditions, out var location);
 
-            return analyzerConfiguration.IsEnabled(Rule.Id);
+                if (baseContext != null)
```
pavel-mikula-sonarsource, line 57: After addressing other comments, this could look something like this
```suggestion
                if (CreateContext(c.Node, c.SemanticModel, conditions) is {} context
                    && context.PrimaryLocation != null)
```

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/ElementAccessContext.cs
```diff
@@ -24,18 +24,16 @@
 namespace SonarAnalyzer.Helpers
 {
     /// <summary>
-    /// Syntax and semantic information about a single element access
+    /// Syntax and semantic information about a single element access.
     /// </summary>
-    public class ElementAccessContext
+    public class ElementAccessContext : BaseContext
     {
         public SyntaxNode Expression { get; }
```
pavel-mikula-sonarsource, line 31: This line repeats a lot. Let's add `SyntaxBaseContext`

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/SyntaxTrackerBase.cs
```diff
@@ -24,29 +24,42 @@
...
+                          TrackedSyntaxKinds);
+                  }
+              });
 
-        protected bool IsEnabled(AnalyzerOptions options)
-        {
-            analyzerConfiguration.Initialize(options);
+            void TrackAndReportIfNecessary(SyntaxNodeAnalysisContext c)
+            {
+                var baseContext = IsTracked(c.Node, c.SemanticModel, conditions, out var location);
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Later update: I've added a comment about adding `SyntaxBaseContext`. Having that, and considering fact that lines `location = express.GetLocation();` repeats itself a lot, you can add overridable readonly property `PrimaryLocation` to the `SyntaxBaseContext` class.

Overridable, because `BaseTypeTracker` does some more magic with it and `BaseTypeContext` will need to override that.

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/ObjectCreationTracker.cs
```diff
@@ -21,82 +21,56 @@
...
-                    && context.InvokedConstructorSymbol.Value.IsConstructor()
-                    && context.InvokedConstructorSymbol.Value.ContainingType.Implements(baseType);
+                       && context.InvokedConstructorSymbol.Value.IsConstructor()
+                       && context.InvokedConstructorSymbol.Value.ContainingType.Implements(baseType);
+
+        protected override BaseContext IsTracked(SyntaxNode expression, SemanticModel semanticModel, ObjectCreationCondition[] conditions, out Location location)
+        {
+            var context = new ObjectCreationContext(expression, semanticModel);
+            location = expression.GetLocation();
+            return conditions.All(c => c(context)) ? context : null;
```
pavel-mikula-sonarsource, line 73: This repeats a lot.
The location should be fixed by another comments. 
Can you do a little bit of generic magic between context, condition and tracker to be able to move the last line in `SyntaxTrackerBase`?
That would allow this method to be just `CreateContext` and have very clean design.

pavel-mikula-sonarsource (changes requested):
First batch for the refactoring (I didn't went through all files yet).

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/SyntaxTrackerBase.cs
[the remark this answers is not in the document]
csaba-sagi-sonarsource: There is already a variable with name context in this scope.

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/BaseTypeTracker.cs
```diff
@@ -114,5 +64,28 @@ internal BaseClassCondition MatchSubclassesOf(params KnownType[] types)
...
+            if (baseTypeList == null || !baseTypeList.Any())
+            {
+                return null;
+            }
+
+            var context = new BaseTypeContext(expression, baseTypeList, semanticModel);
+
+            // We can't pass the issueLocation to the lambda directly so we need a temporary variable
+            Location locationToReport = null;
+            if (conditions.All(c => c(context, out locationToReport)))
```
pavel-mikula-sonarsource, line 82: This is really ugly old design. The location magic should happen inside that specific condition, directly to it's context.

That makes all the condition delegates the same. So there should be just a single generic condition delegate in the base type. And the `conditions.All` part could be moved to base class, as mentioned in another comment.

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/PropertyAccessTracker.cs
```diff
@@ -20,91 +20,50 @@
...
+            var propertyName = GetPropertyName(expression);
+            if (propertyName == null)
+            {
+                return null;
+            }
 
-        #endregion
+            var context = new PropertyAccessContext(expression, propertyName, semanticModel);
+            location = expression.GetLocation();
+            return conditions.All(c => c(context)) ? context : null;
```
pavel-mikula-sonarsource, line 66: After adressing other comments, this could be something like
```
return GetPropertyName(expression) is {} propertyName
    ? CreateContext(...)
    : null;

File: analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/PropertyAccessTracker.cs [hunk 7]
```diff
@@ -20,91 +20,50 @@
...
-            }
-        }
+        private bool CaseInsensitiveComparison { get; }
 
-        #region Syntax-level standard conditions
+        protected PropertyAccessTracker(IAnalyzerConfiguration analyzerConfiguration, DiagnosticDescriptor rule, bool caseInsensitiveComparison = false) : base(analyzerConfiguration, rule) =>
+           CaseInsensitiveComparison = caseInsensitiveComparison;
 
         public PropertyAccessCondition MatchProperty(params MemberDescriptor[] properties) =>
             context =>
```
pavel-mikula-sonarsource, line 44: The body `context => ....` could be single line

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/InvocationTracker.cs
```diff
@@ -41,69 +40,47 @@ public abstract class InvocationTracker<TSyntaxKind> : SyntaxTrackerBase<TSyntax
...
+                       && context.MethodSymbol.Value.ReturnType.DerivesFrom(returnType);
 
         internal InvocationCondition ArgumentIsBoolConstant(string parameterName, bool expectedValue) =>
             context => ConstArgumentForParameter(context, parameterName) is bool boolValue
-                    && boolValue == expectedValue;
+                       && boolValue == expectedValue;
+
+        protected override BaseContext IsTracked(SyntaxNode expression, SemanticModel semanticModel, InvocationCondition[] conditions, out Location location)
+        {
+            var methodName = GetMethodName(expression);
```
pavel-mikula-sonarsource, line 74: This method could be simplified like the other comment with `GetPropertyName`

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/FieldAccessTracker.cs
```diff
@@ -20,88 +20,49 @@
...
             {
-                if (IsTrackedField(c.Node, c.SemanticModel))
-                {
-                    c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, c.Node.GetLocation()));
-                }
+                return null;
             }
 
-            bool IsTrackedField(SyntaxNode expression, SemanticModel semanticModel)
+            var fieldName = GetFieldName(expression);
```
pavel-mikula-sonarsource, line 57: This method could be simplified like the other comment with `GetPropertyName`

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/BaseTypeTracker.cs
```diff
@@ -114,5 +64,28 @@ internal BaseClassCondition MatchSubclassesOf(params KnownType[] types)
                 return false;
             };
         }
+
+        protected override BaseContext IsTracked(SyntaxNode expression, SemanticModel semanticModel, BaseClassCondition[] conditions, out Location location)
+        {
+            location = Location.None;
```
pavel-mikula-sonarsource, line 70: This method could be simplified like the other comment with `GetPropertyName` after addressing all other comments. Something like
```
return GetBaseTypeNodes(expression) is {} baseTypeList
    && baseTypeList.Any()
    ? CreateContext(...)
    : null;
```
Well, in form of arrow method actually.

pavel-mikula-sonarsource (changes requested):
Batch 2/3 to finish the refactoring

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
```diff
@@ -81,10 +82,56 @@ public ExecutingSqlQueries()
...
-             && semanticModel.GetDeclaringSyntaxNode(identifierNameSyntax) is VariableDeclaratorSyntax variableDeclaratorSyntax
-             && IsTracked(variableDeclaratorSyntax.Initializer?.Value, semanticModel));
+        private static bool IsTrackedVariableDeclaration(ExpressionSyntax expression, BaseContext context)
+        {
+            if (expression is IdentifierNameSyntax)
+            {
+                var node = expression;
+                Location location;
+                do
+                {
```
pavel-mikula-sonarsource, line 92: This should make your life easier with number of conditions and you won't need the `(node as IdentifierNameSyntax)` later
```suggestion
            var node = expression;
            Location location;
            while (node is IdentifierNameSyntax is identifierName)
            {
```

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
```diff
@@ -81,10 +82,56 @@ public ExecutingSqlQueries()
...
+                        }
+                    }
+                }
+                while (node is IdentifierNameSyntax);
+            }
+
+            return false;
+        }
+
+        private static Location GetSecondaryLocationForExpression(ExpressionSyntax node)
```
pavel-mikula-sonarsource, line 120: I don't like `Get` prefixes for many reasons (Costin disagrees), I'd use it only when there's no other choice and the method really only gets something, that directly exists.
```suggestion
        private static Location SecondaryLocationForExpression(ExpressionSyntax node)
```

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
```diff
@@ -81,10 +82,56 @@ public ExecutingSqlQueries()
             && expression is BinaryExpressionSyntax concatenation
             && !IsConcatenationOfConstants(concatenation, semanticModel);
 
-        private bool IsTrackedVariableDeclaration(ExpressionSyntax argument, SemanticModel semanticModel) =>
-            (argument is IdentifierNameSyntax identifierNameSyntax
-             && semanticModel.GetDeclaringSyntaxNode(identifierNameSyntax) is VariableDeclaratorSyntax variableDeclaratorSyntax
-             && IsTracked(variableDeclaratorSyntax.Initializer?.Value, semanticModel));
+        private static bool IsTrackedVariableDeclaration(ExpressionSyntax expression, BaseContext context)
```
pavel-mikula-sonarsource, line 85: This seems that it can be moved to base class (while adding `TIdentifierNameSyntax` and moving the `AssignmentFinder` to a protected property of a base class initialized in a constructor, just like the trackers.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
```diff
@@ -81,10 +82,56 @@ public ExecutingSqlQueries()
...
+        }
+
+        private static Location GetSecondaryLocationForExpression(ExpressionSyntax node)
+        {
+            if (node == null)
+            {
+                return Location.None;
+            }
+
+            if (node.Parent is EqualsValueClauseSyntax equalsValueClauseSyntax
```
pavel-mikula-sonarsource, line 127: For simplicity, we don't need the `Syntax` suffix unless it collides with it's symbol (like `methodSyntax` vs `methodSymbol`)
```suggestion
            if (node.Parent is EqualsValueClauseSyntax equalsValueClause
```

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
```diff
@@ -81,10 +82,56 @@ public ExecutingSqlQueries()
...
+                        if (location != Location.None)
+                        {
+                            context.AddSecondaryLocation(new SecondaryLocation(location, string.Format(AssignmentWithFormattingMessage, identifierName)));
+                        }
+
+                        return true;
+                    }
+                    else
+                    {
+                        if (location != Location.None)
```
pavel-mikula-sonarsource, line 108: Would it make sense to move this condition inside `AddSecondaryLocation` instead of the `Debug.Assert`? To remove the responsibility from each user and keep the class responsible for it's own state? You wouldn't need the `location` variable then.

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/ExecutingSqlQueries.cs
```diff
@@ -82,10 +85,54 @@ public ExecutingSqlQueries()
...
+        {
+            if (node == null)
+            {
+                return Location.None;
+            }
+
+            if (node.Parent is EqualsValueSyntax equalsValueSyntax
+                && equalsValueSyntax.Parent is VariableDeclaratorSyntax declarationSyntax)
+            {
+                var identifier = declarationSyntax.Names.FirstOrDefault(name => name.Identifier.ValueText == identifierName);
```
pavel-mikula-sonarsource, line 130: VB is case insensitive, we'll need to add an UT for that as well.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/UsingCommandLineArgumentsBase.cs
```diff
@@ -18,23 +18,29 @@
...
+        private const string MessageFormat = "Make sure that command line arguments are used safely here.";
 
-        protected MethodDeclarationTracker<TSyntaxKind> MethodDeclarationTracker { get; set; }
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
 
-        protected override void Initialize(SonarAnalysisContext context)
-        {
+        protected MethodDeclarationTracker MethodDeclarationTracker { get; set; }
+
+        protected DiagnosticDescriptor Rule { get; }
```
pavel-mikula-sonarsource, line 36: These 3 friends could be without empty lines

pavel-mikula-sonarsource (approved):
Last batch

pavel-mikula-sonarsource (changes requested):
Should not be approved yet

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ExecutingSqlQueries_Net46.cs
```diff
@@ -278,35 +278,38 @@ public void NonCompliant_SystemDataSqlite(SQLiteConnection connection, SQLiteTra
...
 
-            string stillSensitive = sensitiveQuery;
-            command.CommandText = stillSensitive;                                                       // Noncompliant
+            string stillSensitive = sensitiveQuery;                                                     // Secondary [3] {{SQL query is assigned to stillSensitive.}}
+            command.CommandText = stillSensitive;                                                       // Noncompliant [3]
 
-            string sensitiveConcatQuery = "SELECT * FROM Table1 WHERE col1 = '" + param + "'";
-            command = new SqlCommand(sensitiveConcatQuery);                                             // Noncompliant
+            string sensitiveConcatQuery = "SELECT * FROM Table1 WHERE col1 = '" + param + "'";          // Secondary [6,7,8] {{SQL Query is dynamically formatted and assigned to sensitiveConcatQuery.}}
+            //     ^^^^^^^^^^^^^^^^^^^^
```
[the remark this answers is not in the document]
csaba-sagi-sonarsource: Nope, I removed it

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ExecutingSqlQueries_Net46.vb
```diff
@@ -244,34 +244,37 @@ Namespace Tests.Diagnostics
...
+            Dim sensitiveQuery As String = String.Format("INSERT INTO Users (name) VALUES (""{0}"")", param)    ' Secondary [1,2,3,4] {{SQL Query is dynamically formatted and assigned to sensitiveQuery.}}
+            '   ^^^^^^^^^^^^^^
+            Dim command = New SqlCommand(sensitiveQuery)                                                        ' Noncompliant [1]
+            command.CommandText = sensitiveQuery                                                                ' Noncompliant [2]
 
-            Dim stillSensitive As String = sensitiveQuery
-            command.CommandText = stillSensitive                                                                ' Noncompliant
+            Dim stillSensitive As String = sensitiveQuery                                                       ' Secondary [3] {{SQL query is assigned to STILLsENSITIVE.}}
+            '   ^^^^^^^^^^^^^^
+            command.CommandText = STILLsENSITIVE                                                                ' Noncompliant [3] ^13#19
```
pavel-mikula-sonarsource, line 254: This secondary message is confusing here. The query is assigned to `stillSensitive` and then read from or propagated through `STILLsENSITIVE`.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ExecutingSqlQueries.cs
pavel-mikula-sonarsource: The key here was to remove the redundant `if (expression is IdentifierNameSyntax)` check

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/SyntaxTrackerBase.cs
```diff
@@ -18,35 +18,57 @@
...
+                          GeneratedCodeRecognizer,
+                          TrackAndReportIfNecessary,
+                          TrackedSyntaxKinds);
+                  }
+              });
 
-            return analyzerConfiguration.IsEnabled(Rule.Id);
+            void TrackAndReportIfNecessary(SyntaxNodeAnalysisContext c)
+            {
+                if (CreateContext(c.Node, c.SemanticModel) is { } trackingContext
```
pavel-mikula-sonarsource, line 59: What about changing the signature to this? All the way through all context constructors as well all the way down to `SyntaxContextBase`. 
```suggestion
                if (CreateContext(c) is { } trackingContext
```

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/BaseTypeTracker.cs
```diff
@@ -22,97 +22,48 @@
...
         /// </summary>
         protected abstract IEnumerable<SyntaxNode> GetBaseTypeNodes(SyntaxNode contextNode);
 
+        protected BaseTypeTracker(IAnalyzerConfiguration analyzerConfiguration, DiagnosticDescriptor rule) : base(analyzerConfiguration, rule) { }
+
         internal BaseClassCondition MatchSubclassesOf(params KnownType[] types)
         {
             var immutableTypes = types.ToImmutableArray();
-            return (BaseTypeContext context, out Location issueLocation) =>
+            return (context) =>
```
pavel-mikula-sonarsource, line 47:
```suggestion
            return context =>
```

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/BaseTypeTracker.cs
```diff
@@ -22,97 +22,48 @@
...
                     if (context.SemanticModel.GetTypeInfo(baseTypeNode).Type.DerivesOrImplementsAny(immutableTypes))
                     {
-                        issueLocation = baseTypeNode.GetLocation();
+                        context.PrimaryLocation = baseTypeNode.GetLocation();
                         return true; // assume there won't be more than one matching node
                     }
                 }
 
-                issueLocation = null;
+                context.PrimaryLocation = null;
```
pavel-mikula-sonarsource, line 58: This line is not needed anymore. It was there to make compiler happy before.

pavel-mikula-sonarsource (changes requested):
Final polishing, looking great

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/SyntaxTrackerBase.cs
```diff
@@ -18,35 +18,57 @@
...
 
+using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 
 namespace SonarAnalyzer.Helpers
 {
-    public abstract class SyntaxTrackerBase<TSyntaxKind>
+    public delegate bool TrackingCondition<in TContext>(TContext trackingContext);
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: That cleared the air a lot! :)

analyzers/src/SonarAnalyzer.Common/SyntaxTrackers/SyntaxTrackerBase.cs
pavel-mikula-sonarsource: I meant to update `public SyntaxBaseContext(SyntaxNode node, SemanticModel semanticModel)` constructor to single argument as well to deduplicate the propagation of the pair Node&SemanticModel.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/ExecutingSqlQueriesBase.cs
```diff
@@ -129,18 +138,46 @@ protected override void Initialize(SonarAnalysisContext context)
...
+        {
+            var node = expression;
+            while (node is TIdentifierNameSyntax identifierNameSyntax)
+            {
+                var identifierName = GetIdentifierName(identifierNameSyntax);
+                node = AssignmentFinder.FindLinearPrecedingAssignmentExpression(identifierName, node) as TExpressionSyntax;
+
+                if (IsSensitiveExpression(node, context.SemanticModel))
+                {
+                    context.AddSecondaryLocation(
```
pavel-mikula-sonarsource, line 158: This is long and repetitive. Can you craft the `SecondaryLocation` instance inside `AddSecondaryLocation`? Something like `AddSecondaryLocation(Location location, String message, params String[] formatArgs)`

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/ExecutingSqlQueriesBase.cs
```diff
@@ -129,18 +138,46 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                        new SecondaryLocation(
+                                SecondaryLocationForExpression(node, identifierName, out var foundName),
```
pavel-mikula-sonarsource, line 160: This is long piece of code computed in all cases. Can you precompute it before the `if`? `var location = ...`

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/ExecutingSqlQueries.cs
```diff
@@ -44,34 +45,64 @@ public ExecutingSqlQueries()
...
+            if (node == null)
+            {
+                return Location.None;
+            }
+
+            if (node.Parent is EqualsValueSyntax equalsValue
+                && equalsValue.Parent is VariableDeclaratorSyntax declarationSyntax)
+            {
+                var identifier = declarationSyntax.Names.FirstOrDefault(name => name.Identifier.ValueText.Equals(identifierNameToFind, StringComparison.OrdinalIgnoreCase));
+                identifierNameFound = identifier != null ? identifier.Identifier.ValueText : string.Empty;
```
pavel-mikula-sonarsource, line 94:
```suggestion
                identifierNameFound = identifier == null ? string.Empty : identifier.Identifier.ValueText;
```

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/ExecutingSqlQueries.cs
```diff
@@ -44,34 +45,64 @@ public ExecutingSqlQueries()
...
+                return identifier != null ? identifier.GetLocation() : Location.None;
```
pavel-mikula-sonarsource, line 95: Line 84 assigns default value and that is re-assigned here for 2nd time. And the `if` condition is here twice. Let's rewrite it as
```
if(identifier==null)
{
    return Location.None;
}
else
{  
    identifierFound = identifier.Identifier.ValueText;
    return identifier.GetLocation();
}
```

pavel-mikula-sonarsource (changes requested):
Last round

pavel-mikula-sonarsource (approved):
LGTM

Issue #3185, opened 2020-03-06
Rule S5753: Disabling ASP.NET "Request Validation" feature is security-sensitive


agigleux:
Implements https://jira.sonarsource.com/browse/RSPEC-5753

andrei-epure-sonarsource:
Also see https://jira.sonarsource.com/browse/RSPEC-4564 which is already implemented

agigleux:
Would be great to tackle https://github.com/SonarSource/sonar-dotnet/issues/3211 during the same sprint.

andrei-epure-sonarsource:
- [ ] start from the RSPEC-4564 implementation and change it to be according to the RSPEC-5753 - only Controller level. Send a PR for this baby step.
- [ ] Add IT with a `web.config` Additional File. Read the `web.config` as Additional File (as we do with the SonarLint.xml file), parse it and verify the 2 attributes mentioned by RSPEC-5753. Update the analyzers docs to mention that `web.config` files must be manually added as AdditionalFiles in order to be scanned.
- [ ] open an issue for the S4MSB to add `web.config` files as AdditionalFiles w/o user action (and provide an opt-out setting/flag). Also write in the issue to update the analyzer docs - instead of manually adding `web.config`, mention to use the correct S4MSB version.

andrei-epure-sonarsource:
Update: after a discussion with Pavel we decided to search for the web.config ourselves, rather than relying on additional config from the user.

andrei-epure-sonarsource:
Update: we're going to ship just controller-level detection with this issue.

#3951 will bring `web.config` detection.

PR #3927, opened 2021-01-08
Implement first part of S5753

andrei-epure-sonarsource:
First part of #3185


analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DisablingRequestValidation.cs
```diff
@@ -0,0 +1,79 @@
...
+                    var attributes = c.Symbol.GetAttributes();
+                    if (attributes.IsEmpty)
+                    {
+                        return;
+                    }
+                    var attributeWithFalseParameter = attributes.FirstOrDefault(a =>
+                        a.ConstructorArguments.Length == 1
+                        && a.ConstructorArguments[0].Kind == TypedConstantKind.Primitive
+                        && a.ConstructorArguments[0].Value is bool b
+                        && b == false);
```
andrei-epure-sonarsource, line 70: here I prefer `b == false` instead of `!b` because it's the actual `false` literal that we are looking for

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DisablingRequestValidation.cs
```diff
@@ -0,0 +1,79 @@
...
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public sealed class DisablingRequestValidation : SonarDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S5753";
+        private const string MessageFormat = "Make sure disabling ASP.NET Request Validation feature is safe here.";
+        private static readonly DiagnosticDescriptor Rule =
```
pavel-mikula-sonarsource, line 36: _Nitpick:_ Should be single line to match current coding style. We need to improve our templates for RSPEC script.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DisablingRequestValidation.cs
```diff
@@ -0,0 +1,79 @@
...
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager).WithNotConfigurable();
+
+        private readonly IAnalyzerConfiguration analyzerConfiguration;
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public DisablingRequestValidation() : this(AnalyzerConfiguration.Hotspot)
+        {
+        }
```
pavel-mikula-sonarsource, line 45: In this case..
```suggestion
        public DisablingRequestValidation() : this(AnalyzerConfiguration.Hotspot) { }
```

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DisablingRequestValidation.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        public DisablingRequestValidation(IAnalyzerConfiguration analyzerConfiguration)
```
pavel-mikula-sonarsource, line 47: _Nitpick:_ Could be arrow method

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DisablingRequestValidation.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            this.analyzerConfiguration = analyzerConfiguration;
+        }
+
+        protected override void Initialize(SonarAnalysisContext context) =>
```
pavel-mikula-sonarsource, line 52: No VB.NET alternative? It's high value, low hanging and the upcomming `web.config` part is language independent.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DisablingRequestValidation.cs
```diff
@@ -0,0 +1,79 @@
...
+                    if (attributes.IsEmpty)
+                    {
+                        return;
+                    }
+                    var attributeWithFalseParameter = attributes.FirstOrDefault(a =>
+                        a.ConstructorArguments.Length == 1
+                        && a.ConstructorArguments[0].Kind == TypedConstantKind.Primitive
+                        && a.ConstructorArguments[0].Value is bool b
+                        && b == false);
+                    if (attributeWithFalseParameter != null && attributeWithFalseParameter.AttributeClass.Is(KnownType.System_Web_Mvc_ValidateInputAttribute))
```
pavel-mikula-sonarsource, line 71: This doesn't seem right. We need to check the AttributeClass inside the Linq above, actually it should be checked before inspecting the arguments for semantic reasons. Having
```
[RandomAttribute(false)]
[ValidateInput(false)]
```
would not work.

analyzers/rspec/cs/S5753_c#.html
```diff
@@ -0,0 +1,74 @@
...
+<pre>
+[ValidateInput(false)]
+public ActionResult Welcome(string name)
+{
+  ...
+}
+</pre>
+<p>At application level, configured in the Web.config file:</p>
+<pre>
+&lt;configuration&gt;
```
pavel-mikula-sonarsource, line 29: Just raising awareness, that this `web.config` example is present here. Little ahead of it's time.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/DisablingRequestValidationTest.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace SonarAnalyzer.UnitTest.Rules
+{
+    [TestClass]
+    public class DisablingRequestValidationTest
+    {
+        [DataTestMethod]
+        [DataRow("3.0.20105.1")]
+        [DataRow(Constants.NuGetLatestVersion)]
+        [TestCategory("Rule")]
+        public void DisablingRequestValidation(string aspNetMvcVersion) =>
```
pavel-mikula-sonarsource, line 37: For hotspot, there should be also test `DisablingRequestValidation_Disabled` similar to this one: https://github.com/SonarSource/sonar-dotnet/blob/617541f8991ec9ed5a48d680aca645c8ca280019/analyzers/tests/SonarAnalyzer.UnitTest/Rules/UsingRegularExpressionsTest.cs#L40-L45

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DisablingRequestValidation.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        [ValidateInput(false)] // Noncompliant
+        public void VoidFoo() { }
+
+        [ValidateInput(false)] // Noncompliant
+        private ActionResult ArrowFoo(string i) => null;
+    }
+
+    public class CompliantController : Controller
```
pavel-mikula-sonarsource, line 41: Let's add `[ValidateInput(true)]` also here for completeness

pavel-mikula-sonarsource (changes requested):

analyzers/rspec/cs/S5753_c#.html
andrei-epure-sonarsource: indeed. as this won't close the issue, I find this tolerable. If we don't solve the other bit, we'll update the RSPEC and let AppSec know.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DisablingRequestValidation.cs
andrei-epure-sonarsource: good catch

analyzers/rspec/cs/S5753_c#.html
andrei-epure-sonarsource: I rebased it on a feature branch

analyzers/rspec/vbnet/S5753_vb.net.html
```diff
@@ -0,0 +1,74 @@
...
+</ul>
+<p>There is a risk if you answered yes to any of those questions.</p>
+<h2>Recommended Secure Coding Practices</h2>
+<ul>
+  <li> Activate the Request Validation feature for all HTTP requests </li>
+</ul>
+<h2>Sensitive Code Example</h2>
+<p>At Controller level:</p>
+<pre>
+[ValidateInput(false)]
```
pavel-mikula-sonarsource, line 21: I've added VB.NET subtask to Jira. Please regenerate this (and check if my subtask works as expected :) )

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DisablingRequestValidation.cs
```diff
@@ -0,0 +1,34 @@
...
+using SonarAnalyzer.Common;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public sealed class DisablingRequestValidation : DisablingRequestValidationBase
+    {
+        public DisablingRequestValidation() : this(AnalyzerConfiguration.Hotspot) { }
+        public DisablingRequestValidation(IAnalyzerConfiguration analyzerConfiguration) : base(RspecStrings.ResourceManager, analyzerConfiguration) { }
```
pavel-mikula-sonarsource, line 32: For consistency (add empty line between them). 
```suggestion
        public DisablingRequestValidation() : this(AnalyzerConfiguration.Hotspot) { }
        
        public DisablingRequestValidation(IAnalyzerConfiguration analyzerConfiguration) : base(RspecStrings.ResourceManager, analyzerConfiguration) { }
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DisablingRequestValidation.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        [HttpPost]
+        [ValidateInput(false)] // Noncompliant
+        [Obsolete]
+        public ActionResult FooWithMoreAttributes(string input)
+        {
+            return Foo(input);
+        }
+
+
```
pavel-mikula-sonarsource, line 35: _Nitpick:_ One of the review commits added empty line here

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DisablingRequestValidation.vb
```diff
@@ -0,0 +1,109 @@
...
+
+        <ValidateInput(False)> Public Function Foo(ByVal input As String) As ActionResult ' Noncompliant {{Make sure disabling ASP.NET Request Validation feature is safe here.}}
+'        ^^^^^^^^^^^^^^^^^^^^
+            Return Nothing
+        End Function
+
+        <CLSCompliant(False)>
+        <ValidateInput(False)> ' Noncompliant
+        Public Function WithTwoFalse(ByVal input As String) As ActionResult
+            Return Foo(input)
```
pavel-mikula-sonarsource, line 19: _Just a note (no action required):_ Function in VB doesn't have to return a value. So the test cases could be just
```
<ValidateInput(False)>
Public Function Something(....
End Function
```

analyzers/tests/SonarAnalyzer.UnitTest/Rules/DisablingRequestValidationTest.cs
```diff
@@ -0,0 +1,69 @@
...
+        [DataRow(Constants.NuGetLatestVersion)]
+        [TestCategory("Rule")]
+        public void DisablingRequestValidation_CS(string aspNetMvcVersion) =>
+            Verifier.VerifyAnalyzer(@"TestCases\DisablingRequestValidation.cs",
+                new CS.DisablingRequestValidation(AnalyzerConfiguration.AlwaysEnabled),
+                additionalReferences: NuGetMetadataReference.MicrosoftAspNetMvc(aspNetMvcVersion));
+
+        [DataTestMethod]
+        [DataRow("3.0.20105.1")]
+        [DataRow(Constants.NuGetLatestVersion)]
```
pavel-mikula-sonarsource, line 44: I don't think we need the `DataRow` schema for purpose of this test. It won't change with different version so removing it could make it simple and save us few test runs (CSharp7, 7.1, 7.2, 7.3, 8, 9 for one of those nugets).

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DisablingRequestValidationBase.cs
```diff
@@ -0,0 +1,74 @@
...
+    public abstract class DisablingRequestValidationBase : SonarDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S5753";
+        private const string MessageFormat = "Make sure disabling ASP.NET Request Validation feature is safe here.";
+
+        private readonly IAnalyzerConfiguration analyzerConfiguration;
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        protected DiagnosticDescriptor Rule { get; }
```
pavel-mikula-sonarsource, line 38: Could be private readonly field (together with `analyzerConfiguration` above, no empty line between them), as it's not used in inheriting classes.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DisablingRequestValidationBase.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using SonarAnalyzer.Common;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules
+{
+    public abstract class DisablingRequestValidationBase : SonarDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S5753";
```
pavel-mikula-sonarsource, line 31: Could be protected.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DisablingRequestValidationBase.cs
```diff
@@ -0,0 +1,74 @@
...
+                    if (attributes.IsEmpty)
+                    {
+                        return;
+                    }
+                    var attributeWithFalseParameter = attributes.FirstOrDefault(a =>
+                        a.ConstructorArguments.Length == 1
+                        && a.ConstructorArguments[0].Kind == TypedConstantKind.Primitive
+                        && a.ConstructorArguments[0].Value is bool b
+                        && b == false
+                        && a.AttributeClass.Is(KnownType.System_Web_Mvc_ValidateInputAttribute));
```
pavel-mikula-sonarsource, line 65: I'd check the `AttributeClass` first, to follow the semantics. We're looking for `ValidateInput attribute with one argument that has value False`.

We're already working with symbol information, so there's no performance penalty.

pavel-mikula-sonarsource (changes requested):
Final polishing

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DisablingRequestValidationBase.cs
andrei-epure-sonarsource: it's a micro-optimization, but [IsMatch](https://github.com/SonarSource/sonar-dotnet/blob/f400007f77b756f6262070ef14a7476353359bfc/analyzers/src/SonarAnalyzer.Common/Helpers/TypeHelper.cs#L60) which gets called by `Is` does more checks on the type, which we don't need to do for most attributes.

I'd rather keep it as is, I don't think it's difficult to read.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/DisablingRequestValidationTest.cs
pavel-mikula-sonarsource: I've originally meant this for the `_Disabled` tests. If you're aware of any behavior differences between 3.0.20105.1 and latest, we might keep it for the functional test.

pavel-mikula-sonarsource (approved):
LGTM. We should be good to merge. I left one discussion opened to consider it on your side.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/DisablingRequestValidationTest.cs
andrei-epure-sonarsource: I realized now that 3.0.20105.1 is from 2012. I will just use 5.x which is the latest MVC version for .NET Framework, since 2014, and we shouldn't aim to support anything else (as .NET Framework is in maintenance mode).

Issue #3931, opened 2021-01-11
Fix S3264 FP: Event invoked through its invocation list acquired by GetInvocationList()


csaba-sagi-sonarsource:
### Description

Rule S3264 raises a false positive when the event is invoked through its invocation list acquired by GetInvocationList().

Source: https://community.sonarsource.com/t/remove-unused-event/35730

### Repro steps

```
        public delegate void SomeDelegate(int a);

        public class ReproForIssue
        {
            public event SomeDelegate ActuallyUsedEvent; // Noncompliant FP

            public void SomeMethod(SomeDelegate usedEvent)
            {
                var invocations = usedEvent.GetInvocationList();

                foreach (var invocation in invocations)
                {
                    invocation.DynamicInvoke(1623);
                }
            }
        }
```

### Expected behavior

S3264  should not raise an issue for this case.

### Actual behavior

False positive is reported.

### Known workarounds

None

### Related information

* C#/VB.NET Plugins version: 8.16

pavel-mikula-sonarsource:
Internal ticket [NET-1929](https://sonarsource.atlassian.net/browse/NET-1929)

[NET-1929]: https://sonarsource.atlassian.net/browse/NET-1929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3932, opened 2021-01-11
Rule S3264: add repro for #3931

csaba-sagi-sonarsource:
Reproduces #3931

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/UninvokedEventDeclaration.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/UninvokedEventDeclaration.cs
@@ -38,27 +38,23 @@ namespace SonarAnalyzer.Rules.CSharp
         internal const string DiagnosticId = "S3264";
         private const string MessageFormat = "Remove the unused event '{0}' or invoke it.";
 
-        private static readonly Accessibility maxAccessibility = Accessibility.Public;
+        private static readonly Accessibility MaxAccessibility = Accessibility.Public;
 
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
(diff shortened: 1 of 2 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UninvokedEventDeclaration.cs
```diff
@@ -204,5 +204,23 @@ public void InvokeAll()
                 outer.MyAction6?.Invoke(this, EventArgs.Empty);
             }
         }
+
+        // https://github.com/SonarSource/sonar-dotnet/issues/3931
+        public delegate void SomeDelegate(int a);
+
+        public class ReproForIssue3931
+        {
+            public event SomeDelegate ActuallyUsedEvent; // Noncompliant FP
```
costin-zaharia-sonarsource, line 213: I'm not sure I get it. Where is `ActuallyUsedEvent` used?

costin-zaharia-sonarsource (changes requested):
Just a small question related to the usage of `ActuallyUsedEvent`

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UninvokedEventDeclaration.cs
csaba-sagi-sonarsource: I messed it up, I`ll fix it. I tried to combine it with the example from the community and messed it up. Fix is on its way. Sorry...

costin-zaharia-sonarsource: No worries, it looks good now. 

I would add though an additional case also with a call of `.GetInvocationList();` on another event for which we don't do an iteration. We will have to decide if a call to `.GetInvocationList()` will "sanitize" the event or if we want to actually check for iterations and invoke calls.

costin-zaharia-sonarsource (approved):
LGTM!

analyzers/src/SonarAnalyzer.CSharp/Rules/UninvokedEventDeclaration.cs
```diff
@@ -38,62 +38,55 @@ public sealed class UninvokedEventDeclaration : SonarDiagnosticAnalyzer
...
-                .GetRemovableFieldLikeDeclarations(eventSyntax, maxAccessibility)
+                .GetRemovableFieldLikeDeclarations(EventSyntax, MaxAccessibility)
                 .ToList();
 
             if (!removableEventFields.Any())
             {
                 return;
             }
 
-            var invokedSymbols = GetInvokedEventSymbols(removableDeclarationCollector);
```
andrei-epure-sonarsource, line 77: we'd need to come up with a reproducer, but I suspect this change lead to the #4498 perf regression

PR #3933, opened 2021-01-11
S4036 Searching OS commands in PATH is security-sensitive

### LINKED ISSUE #3437: S4036 Searching OS commands in PATH is security-sensitive
Implements https://jira.sonarsource.com/browse/RSPEC-4036 (C# specific [RSPEC](https://jira.sonarsource.com/browse/RSPEC-5894) here).

Should raise when the cmd/file argument (first arg) of these functions:
_System.Diagnostics.ProcessStartInfo.ProcessStartInfo(string)
System.Diagnostics.ProcessStartInfo.ProcessStartInfo(string, string)
System.Diagnostics.Process.Start(string)
System.Diagnostics.Process.Start(string, string)
System.Diagnostics.Process.Start(string, string, System.Security.SecureString, string)
System.Diagnostics.Process.Start(string, string, string, System.Security.SecureString, string)
System.Diagnostics.ProcessStartInfo.FileName.set_

is not an absolute path or relative path (ie: the PATH environment variable is used to search the binary).

A check could be to look for the first chars of the string, if different from:
- ./
- .\
- ../
- ..\
- \
- /
- C:\

it should raise, example:
```
Process p = new Process();
p.StartInfo.FileName = "/usr/bin/binary"; // Compliant

Process p = new Process();
 p.StartInfo.FileName = "binary"; // Sensitive
```

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/dotnet/NetCore3MVC/Controllers/RSPEC_4036

pavel-mikula-sonarsource:
Fixes #3437

This is baby step version of the rule, but good enough to close the issue. 

Object initialization is not perfect, because `CSharpObjectInitializationTracker` doesn't inherit `SyntaxTrackerBase` and doesn't have VB alternative. I'll probably do another PR to reimplement this after #3926 gets merged.


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CommandPath.cs
```diff
@@ -1,8 +1,130 @@
-using System;
+﻿using System.Collections.Generic;
+using System.Diagnostics;
+using System.Security;
```
andrei-epure-sonarsource, line 3: could you also create your own "ProcessStartInfo" class in a different namespace and use that?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CommandPath.vb
```diff
@@ -56,6 +56,11 @@ Public Class Program
 
         psi = New ProcessStartInfo("bad.exe")   ' Noncompliant, later assignment is not tracked
         psi.FileName = "C:\file.exe"
+
+        psi = New ProcessStartInfo() With {.FileName = "C:\file.exe"}
+        psi = New ProcessStartInfo() With {.FileName = "bad.exe"}   ' FN
```
andrei-epure-sonarsource, line 61: why is this FN, and the next is Compliant? it's difficult to follow because you are using in both cases the same variable, which gets overwritten and not exploited

also, what if between line 62 and 63 there's a bunch of logic using the bad value - why should line 62 be compliant?

andrei-epure-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CommandPath.cs
```diff
@@ -0,0 +1,128 @@
...
+
+public class Program
+{
+    private string compliantField = @"C:\file.exe";
+    private string noncompliantField = "file.exe";
+
+    private Process field = Process.Start("file.exe");                      // Noncompliant
+    public Process PropertyRW { get; set; } = Process.Start("file.exe");    // Noncompliant
```
andrei-epure-sonarsource, line 11: maybe a property that doesn't have anything assigned would fix the partial coverage problem?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CommandPath.cs
```diff
@@ -1,8 +1,130 @@
...
+        Process.Start(@"C:\dir\file.exe");
+        Process.Start(@"D:\file.exe");
+        Process.Start(@"z:\file.exe");
+        Process.Start(@"Z:\file.exe");
+        Process.Start(@"\\server\file.exe");
+        Process.Start(@"\\server\dir\file.exe");
+        Process.Start(@"\\server\c$\file.exe");
+        Process.Start(@"\\10.0.0.1\dir\file.exe");
+
+        Process.Start("file");          // FN
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Good catch. I'll implement it, as it's easy. Also applies to custom protocols like `skype:echo123?call`

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/MetadataReferenceFacade.cs
```diff
@@ -87,6 +87,17 @@ internal static class MetadataReferenceFacade
             };
 #endif
 
+        internal static References GetSystemDiagnosticsProcess() =>
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Because `System.DiagnosticsProcess` doesn't work without it. Line
```
private Process field = Process.Start("file.exe");
```
fails with

>The type 'Component' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.ComponentModel.Primitives, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a

because type `Process` inherits `Component`

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/MetadataReferenceFacade.cs
```diff
@@ -87,6 +87,17 @@ internal static class MetadataReferenceFacade
...
+#if NETFRAMEWORK
+            Enumerable.Empty<MetadataReference>();
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: The .NET Framework tests works just fine. It only doesn't need any additional reference.

analyzers/src/SonarAnalyzer.CSharp/SyntaxTrackers/CSharpPropertyAccessTracker.cs
```diff
@@ -40,20 +40,19 @@ public class CSharpPropertyAccessTracker : PropertyAccessTracker<SyntaxKind>
 
         public CSharpPropertyAccessTracker(IAnalyzerConfiguration analyzerConfiguration, DiagnosticDescriptor rule) : base(analyzerConfiguration, rule) { }
 
+        public override object AssignedValue(PropertyAccessContext context) =>
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: It backtracks constant values, so it could be renamed. But I'd keep it as `AssignedValue` because it could be value tracked from `x.Property = variable` back to it's constant source. Where from API user point of view, it provides whatever value we're able to track.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CommandPath.vb
pavel-mikula-sonarsource: Are you looking at the latest version or on intermediate commit? I don't see `FN` in the UTs. I prepared some UTs for the object initialization tracker (that's somehow aware of following reassignment), but I didn't implement it so final version describes actual state.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CommandPath.cs
pavel-mikula-sonarsource: Not sure, let's try

pavel-mikula-sonarsource: Did not help :(

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/MetadataReferenceFacade.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: could you add a comment specifying this, please? it will help future maintainers

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CommandPath.vb
andrei-epure-sonarsource: it's on the "all commits" changes.
```
psi = New ProcessStartInfo() With {.FileName = "C:\file.exe"}
psi = New ProcessStartInfo() With {.FileName = "bad.exe"}   ' FN
psi = New ProcessStartInfo() With {.FileName = "bad.exe"}   ' Compliant, reassigned later
psi.FileName = "C:\file.exe"
```

And I am saying that you can have

```
psi = New ProcessStartInfo() With {.FileName = "bad.exe"}   ' Compliant, reassigned later

Process.Start(psi); // BOOM!

psi.FileName = "C:\file.exe"
```

andrei-epure-sonarsource: my suggestion is that both are FNs

andrei-epure-sonarsource (changes requested):
Minor 2 comments left

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CommandPath.vb
pavel-mikula-sonarsource: Agrh, I missed that it's on VB file 🤦‍♂️

andrei-epure-sonarsource (approved):
LGTM :shipit:

PR #3935, opened 2021-01-12
Cleanup additional license files

pavel-mikula-sonarsource:
License should be placed only inside file headers.


analyzers/CBDE/LICENSE.txt
```diff
@@ -1,4 +1,4 @@
-Copyright (C) 2012-2020 SonarSource SA
+Copyright (C) 2012-2021 SonarSource SA
```
pavel-mikula-sonarsource, line 1: Leftover from happy new year.

I'm not really sure what to do with this file.

analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.CSharp.nuspec
```diff
@@ -6,7 +6,7 @@
     <title>SonarAnalyzer for C#</title>
     <authors>SonarSource</authors>
     <owners>SonarSource</owners>
-    <license type="file">License.txt</license>
+    <license type="expression">LGPL-3.0-only</license>
```
pavel-mikula-sonarsource, line 9: I manually checked nuget.org UI and this translates into link https://licenses.nuget.org/LGPL-3.0-only

docs/code-of-conduct.md
```diff
@@ -1,71 +1,71 @@
...
-faith may face temporary or permanent repercussions as determined by other
-members of the project's leadership.
-
-## Attribution
-
-This Code of Conduct is adapted from the [Contributor Covenant][homepage], version 1.4,
-available at https://www.contributor-covenant.org/version/1/4/code-of-conduct.html
-
-[homepage]: https://www.contributor-covenant.org
+## Our Pledge
```
pavel-mikula-sonarsource, line 1: Moved file to keep project root clean

README.md
```diff
@@ -50,7 +50,7 @@ languages in [SonarQube](http://www.sonarqube.org/), [SonarCloud](https://sonarc
 ## How to contribute
 
 There are many ways you can contribute to the `sonar-dotnet` project. 
-When contributing, please respect our [Code of Conduct](./CODE_OF_CONDUCT.md).
+When contributing, please respect our [Code of Conduct](./docs/code-of-conduct.md).
```
pavel-mikula-sonarsource, line 53: This is just cleanup of the root directory (moved file).

And the end of this file, there's `## License` section. That's probably fine (?).

analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.CSharp.nuspec
andrei-epure-sonarsource: nice!

README.md
andrei-epure-sonarsource: I'd keep the license in the root, as it is

LICENSE.txt
```diff
@@ -1,165 +0,0 @@
-                   GNU LESSER GENERAL PUBLIC LICENSE
```
andrei-epure-sonarsource, line 1: please revert this change, we still need the license in the root

NOTICE.txt
```diff
@@ -1,6 +0,0 @@
-SonarQube C# Plugin
```
andrei-epure-sonarsource, line 1: given we have this notice in all our repos, let's revert this until we clarify if it's needed.
I will ask around

andrei-epure-sonarsource (changes requested):
Let's keep LICENSE and NOTICE in the root folder to be aligned with all our other open-source repositories.

README.md
pavel-mikula-sonarsource: Does it imply that you want to remove the `## License` section from `readme.md`? As `license.txt` file is discussed below.

andrei-epure-sonarsource: I don't mind either way, TBH, we can leave it as is

andrei-epure-sonarsource (approved):
LGTM

PR #3936, opened 2021-01-12
Deprecated S2070 and update S4790

### LINKED ISSUE #3360: Rule S4790: its content should be replaced by S2070
For hashing algorithms, several rules exist, in particular these two:
- A security-hotspot: https://jira.sonarsource.com/browse/RSPEC-4790
- That supersedes this vulnerability: https://jira.sonarsource.com/browse/RSPEC-2070

It's not possible to maintain two rules on exactly the same subject, for us and the end users, so:
- S2070 will be deprecated
- the content/implementation of S2070 is more relevant than S4790 because S4790 raises everywhere a hash function is used (even when secure hash function, like SHA-256 is used) and S2070 raises only when a weak hash function is used (like MD5)
- so the content/implementation of S2070 should "be moved" to S4790 (the key of the rule should be updated in Sonar-dotnet), because the type of issue (hotspot) is more relevant

costin-zaharia-sonarsource:
Fixes #3360

### DIFF
--- a/analyzers/its/expected/Nancy/Nancy-{34576216-0DCA-4B0F-A0DC-9075E75A676F}-S4790.json
+++ b/analyzers/its/expected/Nancy/Nancy-{34576216-0DCA-4B0F-A0DC-9075E75A676F}-S4790.json
@@ -4,45 +4,6 @@
 "id":  "S4790",
 "message":  "Make sure this weak hash algorithm is not used in a sensitive context here.",
 "location":  {
-"uri":  "sources\Nancy\src\Nancy\Cryptography\DefaultHmacProvider.cs",
-"region":  {
-"startLine":  14,
-"startColumn":  43,
-"endLine":  14,
-"endColumn":  59
-}
-}
-},
-{
-"id":  "S4790",
-"message":  "Make sure this weak hash algorithm is not used in a sensitive context here.",
-"location":  {
-"uri":  "sources\Nancy\src\Nancy\Cryptography\DefaultHmacProvider.cs",
-"region":  {
-"startLine":  60,
-"startColumn":  33,
-"endLine":  60,
-"endColumn":  57
-}
-}
-},
-{
-"id":  "S4790",
-"message":  "Make sure this weak hash algorithm is not used in a sensitive context here.",
-"location":  {
-"uri":  "sources\Nancy\src\Nancy\Diagnostics\DiagnosticsSession.cs",
-"region":  {
-"startLine":  28,
-"startColumn":  29,
-"endLine":  28,
-"endColumn":  48
-}
-}
-},
-{
-"id":  "S4790",
-"message":  "Make sure this weak hash algorithm is not used in a sensitive context here.",
-"location":  {
 "uri":  "sources\Nancy\src\Nancy\Responses\EmbeddedFileResponse.cs",
 "region":  {
 "startLine":  66,
--- a/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
@@ -44,13 +44,17 @@ namespace SonarAnalyzer.Rules
         private readonly string[] unsafeAlgorithms =
         {
             "DSA",
+            "System.Security.Cryptography.DSA",
             "HMACMD5",
+            "System.Security.Cryptography.HMACMD5",
             "HMACRIPEMD160",
+            "System.Security.Cryptography.HMACRIPEMD160",
             "HMACSHA1",
+            "System.Security.Cryptography.HMACSHA1",
             "MD5",
             "System.Security.Cryptography.MD5",
             "RIPEMD160",
-            "RIPEMD160Managed",
+            "System.Security.Cryptography.RIPEMD160",
             "SHA1",
             "System.Security.Cryptography.SHA1"
         };
--- a/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
@@ -41,7 +41,7 @@ namespace SonarAnalyzer.Rules
             KnownType.System_Security_Cryptography_SHA1
         };
 
-        private readonly string[] unsafeAlgorithms = { "SHA1", "MD5", "DSA", "HMACMD5", "HMACRIPEMD160", "RIPEMD160", "RIPEMD160Managed" };
+        private readonly string[] unsafeAlgorithms = { "SHA1", "System.Security.Cryptography.SHA1", "MD5", "System.Security.Cryptography.MD5", "DSA", "HMACMD5", "HMACRIPEMD160", "HMACSHA1", "RIPEMD160", "RIPEMD160Managed" };
 
         protected ObjectCreationTracker<TSyntaxKind> ObjectCreationTracker { get; set; }
 
--- a/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
@@ -41,7 +41,19 @@ namespace SonarAnalyzer.Rules
             KnownType.System_Security_Cryptography_SHA1
         };
 
-        private readonly string[] unsafeAlgorithms = { "SHA1", "System.Security.Cryptography.SHA1", "MD5", "System.Security.Cryptography.MD5", "DSA", "HMACMD5", "HMACRIPEMD160", "HMACSHA1", "RIPEMD160", "RIPEMD160Managed" };
+        private readonly string[] unsafeAlgorithms =
+        {
+            "DSA",
+            "HMACMD5",
+            "HMACRIPEMD160",
+            "HMACSHA1",
+            "MD5",
+            "System.Security.Cryptography.MD5",
+            "RIPEMD160",
+            "RIPEMD160Managed",
+            "SHA1",
+            "System.Security.Cryptography.SHA1"
+        };
 
         protected ObjectCreationTracker<TSyntaxKind> ObjectCreationTracker { get; set; }
 
--- a/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
@@ -28,50 +28,48 @@ namespace SonarAnalyzer.Rules
         protected const string DiagnosticId = "S4790";
         protected const string MessageFormat = "Make sure this weak hash algorithm is not used in a sensitive context here.";
 
-        private const string Create = "Create";
+        private const string CreateMethodName = "Create";
 
         private readonly KnownType[] algorithmTypes =
         {
-            KnownType.System_Security_Cryptography_SHA1,
-            KnownType.System_Security_Cryptography_MD5,
             KnownType.System_Security_Cryptography_DSA,
-            KnownType.System_Security_Cryptography_RIPEMD160,
-            KnownType.System_Security_Cryptography_HMACSHA1,
             KnownType.System_Security_Cryptography_HMACMD5,
-            KnownType.System_Security_Cryptography_HMACRIPEMD160
+            KnownType.System_Security_Cryptography_HMACRIPEMD160,
+            KnownType.System_Security_Cryptography_HMACSHA1,
+            KnownType.System_Security_Cryptography_MD5,
+            KnownType.System_Security_Cryptography_RIPEMD160,
+            KnownType.System_Security_Cryptography_SHA1
         };
 
         private readonly string[] unsafeAlgorithms = { "SHA1", "MD5", "DSA", "HMACMD5", "HMACRIPEMD160", "RIPEMD160", "RIPEMD160Managed" };
 
         protected ObjectCreationTracker<TSyntaxKind> ObjectCreationTracker { get; set; }
 
-        protected InvocationTracker<TSyntaxKind> ParameterlessFactoryInvocationTracker { get; set; }
-
-        protected InvocationTracker<TSyntaxKind> ParameterizedFactoryInvocationTracker { get; set; }
+        protected InvocationTracker<TSyntaxKind> InvocationTracker { get; set; }
 
         protected override void Initialize(SonarAnalysisContext context)
         {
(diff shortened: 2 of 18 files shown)


analyzers/its/expected/Nancy/Nancy-{34576216-0DCA-4B0F-A0DC-9075E75A676F}-S4790.json
```diff
@@ -4,45 +4,6 @@
 "id":  "S4790",
 "message":  "Make sure this weak hash algorithm is not used in a sensitive context here.",
 "location":  {
-"uri":  "sources\Nancy\src\Nancy\Cryptography\DefaultHmacProvider.cs",
```
costin-zaharia-sonarsource, line 7: These are the lines which are now compliant:
```
private readonly int hmacLength = new HMACSHA256().HashSize / 8;

var hmacGenerator = new HMACSHA256(this.key);

var algorithm = new SHA256Managed();
```

they are all related to SHA 256 which is considered secure.

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/CreatingHashAlgorithms.cs
```diff
@@ -46,8 +46,8 @@ public CreatingHashAlgorithms()
         internal /*for testing*/ CreatingHashAlgorithms(IAnalyzerConfiguration analyzerConfiguration)
         {
             ObjectCreationTracker = new VisualBasicObjectCreationTracker(analyzerConfiguration, rule);
-            InvocationTracker = new VisualBasicInvocationTracker(analyzerConfiguration, rule);
-            BaseTypeTracker = new VisualBasicBaseTypeTracker(analyzerConfiguration, rule);
+            ParameterlessFactoryInvocationTracker = new VisualBasicInvocationTracker(analyzerConfiguration, rule);
+            ParameterizedFactoryInvocationTracker = new VisualBasicInvocationTracker(analyzerConfiguration, rule);
```
pavel-mikula-sonarsource, line 50: This is suspicious. Tracker instance doesn't persist any state about tracked methods and can be reused. One tracker of given type should be just fine. We can call `Track` multiple times with different conditions.
```suggestion
            InvocationTracker = new VisualBasicInvocationTracker(analyzerConfiguration, rule);
```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
```diff
@@ -28,23 +28,50 @@ public abstract class CreatingHashAlgorithmsBase<TSyntaxKind> : SonarDiagnosticA
...
         protected override void Initialize(SonarAnalysisContext context)
         {
-            ObjectCreationTracker.Track(context,
-                ObjectCreationTracker.WhenDerivesFrom(KnownType.System_Security_Cryptography_HashAlgorithm));
+            ObjectCreationTracker.Track(context, ObjectCreationTracker.WhenDerivesOrImplementsAny(algorithmTypes));
 
-            InvocationTracker.Track(context,
-                InvocationTracker.MethodNameIs("Create"),
-                InvocationTracker.MethodReturnTypeIs(KnownType.System_Security_Cryptography_HashAlgorithm));
+            ParameterlessFactoryInvocationTracker.Track(context,
```
pavel-mikula-sonarsource, line 56: Can we use something like
```
ParameterlessFactoryInvocationTracker.MatchMethod(algorithmTypes.Select(x => new MemberDescriptor(x, CreateMethodName)).ToArray()),
```
to avoid the repetition? It will create few more instances, but no harm is done as the Create method of those would be insecure as well.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
```diff
@@ -28,23 +28,50 @@ public abstract class CreatingHashAlgorithmsBase<TSyntaxKind> : SonarDiagnosticA
...
+                                                                                                          new MemberDescriptor(KnownType.System_Security_Cryptography_HMAC, Create),
+                                                                                                          new MemberDescriptor(KnownType.System_Security_Cryptography_MD5, Create),
+                                                                                                          new MemberDescriptor(KnownType.System_Security_Cryptography_RIPEMD160, Create),
+                                                                                                          new MemberDescriptor(KnownType.System_Security_Cryptography_SHA1, Create)),
+                                                        ParameterlessFactoryInvocationTracker.MethodHasParameters(0));
 
-            BaseTypeTracker.Track(context,
-                BaseTypeTracker.MatchSubclassesOf(KnownType.System_Security_Cryptography_HashAlgorithm));
+            ParameterizedFactoryInvocationTracker.Track(context,
```
pavel-mikula-sonarsource, line 64: Same as before, can we use lambda to avoid repetition? With one additional `"CreateFromName"` concat.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CreatingHashAlgorithms.cs
```diff
@@ -1,77 +1,104 @@
...
+    public class InsecureHashAlgorithm
     {
-        // RSPEC 4790: https://jira.sonarsource.com/browse/RSPEC-4790
-        public void ComputeHash()
+        public void Hash(byte[] temp)
         {
-            // Review all instantiations of classes that inherit from HashAlgorithm, for example:
-            HashAlgorithm hashAlgo = HashAlgorithm.Create();
-//                                   ^^^^^^^^^^^^^^^^^^^^^^    {{Make sure this weak hash algorithm is not used in a sensitive context here.}}
-            HashAlgorithm hashAlgo2 = HashAlgorithm.Create("SHA1");
```
pavel-mikula-sonarsource, line 13: I can't find new version of this test. Can you add a line for each `unsafeAlgorithms` value?

pavel-mikula-sonarsource, line 13: With the new tracking we have, you can add one test to document variable tracking:
```
var name = "SHA512";
name = "SHA1";
HashAlgorithm.Create(name); // Noncompliant
name = "SHA512";
HashAlgorithm.Create(name); // Compliant

File: analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CreatingHashAlgorithms.cs [hunk 6]
```diff
@@ -1,77 +1,104 @@
...
+namespace Tests.Diagnostics
+{
+    public class InsecureHashAlgorithm
     {
-        // RSPEC 4790: https://jira.sonarsource.com/browse/RSPEC-4790
-        public void ComputeHash()
+        public void Hash(byte[] temp)
         {
-            // Review all instantiations of classes that inherit from HashAlgorithm, for example:
-            HashAlgorithm hashAlgo = HashAlgorithm.Create();
```
pavel-mikula-sonarsource, line 11: This overload became safe now? That's suspicious. Should we document that it's compliant now in a test case?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CreatingHashAlgorithms.cs
```diff
@@ -1,77 +1,104 @@
...
+            }
+            using (var sha256 = HashAlgorithm.Create("SHA256Managed"))
+            {
+            }
         }
 
-        public void AdditionalTests(SHA1CryptoServiceProvider sha1)
+        public void Md5Calls()
         {
-            var myHash = new MyHashAlgorithm();     // Noncompliant
```
pavel-mikula-sonarsource, line 27: This is gone now, because it was meanwhile handled by S2257?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CreatingHashAlgorithms.cs
```diff
@@ -1,77 +1,104 @@
...
-        public static MyHashAlgorithm CreateHash() => null;
-        public static MyHashAlgorithm DoCreate() => null;
-
-        public override void Initialize() { /* no-op */ }
-        protected override void HashCore(byte[] array, int ibStart, int cbSize) { /* no-op */ }
-        protected override byte[] HashFinal()
+        public void DsaCalls()
         {
-            throw new System.NotImplementedException();
+            using (var dsa = new DSACryptoServiceProvider()) // Noncompliant
```
pavel-mikula-sonarsource, line 61: I don't see a test case for `RIPEMD160` and `HMACRIPEMD160`

pavel-mikula-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
costin-zaharia-sonarsource: I prefer not to do this change since not all of them have a create method.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CreatingHashAlgorithms.cs
costin-zaharia-sonarsource: S2070 does not check for this and I've checked already with @eric-therond-sonarsource and we should mirror it's implementation but sure I can add a test case to document the behavior.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
costin-zaharia-sonarsource: I prefer not to do this change since not all of them have a create method.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CreatingHashAlgorithms.cs
costin-zaharia-sonarsource: No, the reason is that we want to mirror the functionality from S2070 which does not check for custom inheritance of `HashAlgorithm`

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CreatingHashAlgorithms.cs
costin-zaharia-sonarsource: We do have a similar case for MD5 and it's handled the same for all. I can change it a bit to check also a valid value but I would not add test cases for each algorithm since the code which is handling them is the same and they will grow a lot and will be hard to maintain.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
```diff
@@ -28,23 +28,48 @@ public abstract class CreatingHashAlgorithmsBase<TSyntaxKind> : SonarDiagnosticA
...
+            KnownType.System_Security_Cryptography_HMACMD5,
+            KnownType.System_Security_Cryptography_HMACRIPEMD160,
+            KnownType.System_Security_Cryptography_HMACSHA1,
+            KnownType.System_Security_Cryptography_MD5,
+            KnownType.System_Security_Cryptography_RIPEMD160,
+            KnownType.System_Security_Cryptography_SHA1
+        };
 
-        protected InvocationTracker<TSyntaxKind> InvocationTracker { get; set;  }
+        private readonly string[] unsafeAlgorithms = { "SHA1", "MD5", "DSA", "HMACMD5", "HMACRIPEMD160", "RIPEMD160", "RIPEMD160Managed" };
```
pavel-mikula-sonarsource, line 44: "HMACSHA1" is missing in this list and UTs.

pavel-mikula-sonarsource, line 44: Value `System.Security.Cryptography.SHA1` is also valid algorithm name according to [docs](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.hashalgorithm.create?view=netframework-4.8#System_Security_Cryptography_HashAlgorithm_Create_System_String_). All forms would be scattered around. We should support them or document FNs.

pavel-mikula-sonarsource (changes requested):
Last cross check. And VB UT is failing on compilation issue.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
costin-zaharia-sonarsource: Yes, it wasn't in the initial implementation of S2070. I will add it.

pavel-mikula-sonarsource: The same concept applies to all:
https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.hmac.create?view=netframework-4.8

I haven't found a list in docs for `RIPEMD160` and `RIPEMD160Managed` as the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.ripemd160.create?view=netframework-4.8#System_Security_Cryptography_RIPEMD160_Create_System_String_) are missing the list of valid values. But quick test shows, that:
`"System.Security.Cryptography.RIPEMD160"` creates instance of `RIPEMD160Managed` type, so we should support it.
`"RIPEMD160Managed"` is not a valid value for the argument, it returns `null`. So we should remove it.

pavel-mikula-sonarsource: I'm sorry I missed that the full list of valid values is scattered around :/

pavel-mikula-sonarsource (changes requested):
I found more values for the `unsafeAlgorithms`

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CreatingHashAlgorithmsBase.cs
costin-zaharia-sonarsource: > I haven't found a list in docs for RIPEMD160 and RIPEMD160Managed as the docs are missing the list of valid values. But quick test shows, that:
"System.Security.Cryptography.RIPEMD160" creates instance of RIPEMD160Managed type, so we should support it.
"RIPEMD160Managed" is not a valid value for the argument, it returns null. So we should remove it.

They were present from the old implementation. I wonder now if it was a mistake or they were really needed.

costin-zaharia-sonarsource: I've removed **RIPEMD160Managed** and added test cases with namespaces for the rest of the values

pavel-mikula-sonarsource (approved):
LGTM

PR #3941, opened 2021-01-14
Refactor MetadataReferenceFacade

pavel-mikula-sonarsource:
Remove `Get` prefix and `()` suffix to align with source members.


costin-zaharia-sonarsource (approved):
LGTM!

PR #3942, opened 2021-01-14
Implement web.config analysis

andrei-epure-sonarsource:
Related to #3185


andrei-epure-sonarsource:
@costin-zaharia-sonarsource  - I still need to add the .NET-level integration tests in the `regression-test.ps1` script, and make that work. However, this is ready for an initial round of code review.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DisablingRequestValidationBase.cs
```diff
@@ -69,5 +80,79 @@ protected DisablingRequestValidationBase(System.Resources.ResourceManager rspecR
...
                 SymbolKind.Method);
+
+            context.RegisterCompilationAction(c =>
+            {
+                analyzerConfiguration.Initialize(c.Options);
+                if (!analyzerConfiguration.IsEnabled(DiagnosticId))
+                {
+                    return;
+                }
+                foreach (var fullPath in Directory.GetFiles(rootPath, "web.config", SearchOption.AllDirectories).Select(p => Path.GetFullPath(p)))
```
costin-zaharia-sonarsource, line 91: I would extract `Directory.GetFiles(rootPath, "web.config", SearchOption.AllDirectories).Select(p => Path.GetFullPath(p))` to a method.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DisablingRequestValidationBase.cs
```diff
@@ -69,5 +80,79 @@ protected DisablingRequestValidationBase(System.Resources.ResourceManager rspecR
...
+                }
+            });
+        }
+
+        private void ReportValidateRequest(XDocument doc, string webConfigPath, CompilationAnalysisContext c)
+        {
+            foreach (var pages in doc.XPathSelectElements("configuration/system.web/pages"))
+            {
+                if (pages.Attribute("validateRequest") is { } validateRequest
+                    && validateRequest.Value == "false"
```
costin-zaharia-sonarsource, line 108: Not sure about it but `False` can also be correctly parsed, right? Shouldn't we do a case insensitive check or actually parse the value?

I ask about this since, as far as I can tell, the framework uses behind [NameValueCollection](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.namevaluecollection?view=netframework-4.8) when parsing the file, which uses a case insensitive comparer.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DisablingRequestValidationBase.cs
```diff
@@ -69,5 +80,79 @@ protected DisablingRequestValidationBase(System.Resources.ResourceManager rspecR
...
+            var lineInfo = (IXmlLineInfo)element;
+            if (lineInfo.HasLineInfo())
+            {
+                // the IXmlLineInfo.LineNumber is 1-based, whereas Roslyn is zero-based
+                var lineNumber = lineInfo.LineNumber - 1;
+                // IXmlLineInfo.LinePosition is 1-based
+                var start = lineInfo.LinePosition + element.ToString().IndexOf(attribute) - 2;
+                var end = start + attribute.Length;
+                var linePos = new LinePositionSpan(new LinePosition(lineNumber, start), new LinePosition(lineNumber, end));
+                return Location.Create(path, new TextSpan(lineNumber, end - start), linePos);
```
costin-zaharia-sonarsource, line 141: Isn't `end - start` the `attribute.Length`?

analyzers/tests/SonarAnalyzer.UnitTest/Rules/DisablingRequestValidationTest.cs
```diff
@@ -31,33 +37,95 @@ namespace SonarAnalyzer.UnitTest.Rules
...
+            var expectedIssuesInRoot = new IssueLocationCollector().GetExpectedIssueLocations(GetLines(rootWebConfig)).ToList();
+            DiagnosticVerifier.CompareActualToExpected(compilation.LanguageVersionString(), actualIssuesInRoot, expectedIssuesInRoot, false);
+
+            // verify subfolder issues
+            foreach (var subFolder in subFolders)
+            {
+                var subFolderWebConfig = Path.Combine(rootDirectory, subFolder, WebConfig);
+                var actualSubFolderIssues = allDiagnostics.Where(d => d.Location.GetLineSpan().Path.EndsWith(subFolderWebConfig));
+                var expectedSubFolderIssues = new IssueLocationCollector().GetExpectedIssueLocations(GetLines(subFolderWebConfig)).ToList();
+                DiagnosticVerifier.CompareActualToExpected(compilation.LanguageVersionString(), actualSubFolderIssues, expectedSubFolderIssues, false);
```
costin-zaharia-sonarsource, line 93: At the first glance this looks duplicated. Can we extract a method?

costin-zaharia-sonarsource (changes requested):

analyzers/its/expected/WebConfig/Framework48-{6AC7E627-A21A-459C-8831-C45B6058E24D}-S1451.json
```diff
@@ -0,0 +1,69 @@
+{
+"issues":  [
+{
+"id":  "S1451",
+"message":  "Add or update the header of this file.",
+"location":  {
+"uri":  "sources\WebConfig\App_Start\RouteConfig.cs",
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: We don't need the files, but I want to keep a working Web project (to be able to test the functionality of multiple WebConfig overriden config, for example).

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DisablingRequestValidationBase.cs
andrei-epure-sonarsource: good point (even if XML complains only "false" respects the schema boolean type)

analyzers/tests/SonarAnalyzer.UnitTest/Rules/DisablingRequestValidationTest.cs
```diff
@@ -31,33 +37,95 @@ namespace SonarAnalyzer.UnitTest.Rules
     public class DisablingRequestValidationTest
     {
         private const string AspNetMvcVersion = "5.2.7";
+        private const string WebConfig = "Web.config";
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: yes, and we'll need to support the extra `xdt:Transform` syntaxes... as we're quite advanced in the sprint, I will postpone doing this in my next iteration

costin-zaharia-sonarsource (approved):
LGTM!

PR #3946, opened 2021-01-15
Fix NRE in MethodParameterLookup

pavel-mikula-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpMethodParameterLookup.cs
```diff
@@ -31,12 +31,12 @@ internal class CSharpMethodParameterLookup : AbstractMethodParameterLookup<Argum
         }
 
         public CSharpMethodParameterLookup(ArgumentListSyntax argumentList, SemanticModel semanticModel)
```
csaba-sagi-sonarsource, line 33: Maybe these constructors could be single line.

analyzers/src/SonarAnalyzer.VisualBasic/Helpers/VisualBasicMethodParameterLookup.cs
```diff
@@ -26,12 +26,12 @@ namespace SonarAnalyzer.Helpers
     internal class VisualBasicMethodParameterLookup : AbstractMethodParameterLookup<ArgumentSyntax>
     {
         public VisualBasicMethodParameterLookup(ArgumentListSyntax argumentList, IMethodSymbol methodSymbol)
```
csaba-sagi-sonarsource, line 28: Same here, if you want you can move the constructors to single line.

csaba-sagi-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpMethodParameterLookup.cs
pavel-mikula-sonarsource: The 2nd one has some logic that should be on a separate line. So I'll keep the 1st one the same for consitency.

PR #3954, opened 2021-01-18
Deprecate S4564 in favor of S5753

### LINKED ISSUE #3160: False positives in RSPEC-4564, input validation checks.
### Description

RSPEC-4564 is reporting false positives for HttpPost functions without ValidateInput..

### Repro steps

I am getting a lot of false positives for this rule, claiming I need ValidateInput(true) to be set, which I know I don't.

First issue is that this rule is triggering on posts that do not accept strings as parameters. For example, the rule is triggering on this function:
[HttpPost]
public JsonResult Delete(int id)

Second issue is that since ASP.NET 4.0, HTML input is validated by default unless explicitly disabled in web.config or by using ValidateInput(false). We are doing neither. When testing on an HttpPost function that does accept a string, we proved that the input is being validated fine without needing ValidateInput(true), but Sonar is not recognizing this.

### Expected behavior

* I expect this rule to not trigger on HttpPost functions with no string parameters.
* I expect this rule to not trigger on ASP.NET 4.0+ projects unless input validation has been explicitly disabled.

### Actual behavior

* This rule is triggering on HttpPost functions that only have integer or boolean parameters.
* This rule is triggering on an ASP.NET 4.0 project, and input validation has not been explicitly disabled.

### Known workarounds

* None, save for manually mitigating each function as a false positive after we proved malicious input is not possible.

### Related information

* SonarC#: 8.4
* Visual Studio Version: 16.4.5
* Operating System: Windows Server 2016

### LINKED ISSUE #3211: Deprecate S4564 in favor of S5753
https://jira.sonarsource.com/browse/RSPEC-4564 (Vulnerability) is deprecated in favor of https://jira.sonarsource.com/browse/RSPEC-5753 (Security Hotspot).

costin-zaharia-sonarsource:
Fixes #3211
Fixes #3160 (as we deprecated it, we won't fix the FP)

### DIFF
--- a/analyzers/rspec/cs/S4564_c#.html
+++ b/analyzers/rspec/cs/S4564_c#.html
@@ -43,4 +43,6 @@ public class FooBarController : Controller
   <li> <a href="https://www.sans.org/top25-software-errors/#cat1">SANS Top 25</a> - Insecure Interaction Between Components </li>
   <li> <a href="https://www.owasp.org/index.php/ASP.NET_Request_Validation">OWASP ASP.NET Request Validation</a> </li>
 </ul>
+<h2>Deprecated</h2>
+<p>This rule is deprecated; use {rule:csharpsquid:S5753} instead.</p>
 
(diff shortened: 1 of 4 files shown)


andrei-epure-sonarsource (approved):
LGTM

Issue #3955, opened 2021-01-18
FIX S2583: False positive with ref parameter


csaba-sagi-sonarsource:
### Description

Rule S2583 raises a false positive with ref parameters.

Source: https://community.sonarsource.com/t/sonar-is-not-analyzing-the-reference-objects/35834

### Repro steps

```
public void TestExample(ref bool stop)
{
    while (!stop)
    {
        while (true)
        {
            if (stop)
            {
                break;
            }
        }
    }
}
```

### Expected behavior

S2583 should not raise an issue for this case.

### Actual behavior

False positive is reported.

### Known workarounds

None.

### Related information

* C#/VB.NET Plugins version: 8.16

andrei-epure-sonarsource:
This seems to be a duplicate of #3353 - or am I missing something?

csaba-sagi-sonarsource:
This issue is a duplicate of [Fix S2589 FP: ref parameters in multithreading](https://github.com/SonarSource/sonar-dotnet/issues/3353)

PR #3956, opened 2021-01-18
Rule S2583: Add repro for #3955

csaba-sagi-sonarsource:
relates to #3955

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
@@ -267,38 +267,6 @@ namespace Tests.Diagnostics
         }
     }
 
-    public class PassingArgumentsByReference
-    {
-        private static object _gate = new object();
-
-        public void InitWithLock<T>(ref T field)
-            where T : new()
-        {
-            if (field == null)
-            {
-                lock (_gate)
-                {
-                    if (field == null) // Noncompliant, FP - in multithreading context it makes sense to check for null twice
-                    {
-                        field = new T();
-                    }
-                }
-            }
-        }
-
-        public void Init<T>(ref T field)
-            where T : new()
-        {
-            if (field == null)
-            {
-                if (field == null) // Noncompliant, we already checked for null
-                {
-                    field = new T();
-                }
-            }
-        }
-    }
-
     public class Tuples
     {
         public void DoSomething(bool arg)
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.cs
@@ -2720,3 +2720,24 @@ namespace Repro_LocalFunction
         }
     }
 }
+
+// https://github.com/SonarSource/sonar-dotnet/issues/3955 for S2583
+namespace Repro_RefParam
+{
+    public class Repro
+    {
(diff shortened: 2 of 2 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.cs
```diff
@@ -2720,3 +2720,24 @@ void LocalFunction()
         }
     }
 }
+
+// https://github.com/SonarSource/sonar-dotnet/issues/3955 for S2583
+namespace Repro_RefParam
```
andrei-epure-sonarsource, line 2725: I think this test cases is similar to the one in the `ConditionEvaluatesToConstant.CSharp8.cs ` file - see https://github.com/SonarSource/sonar-dotnet/pull/3358/files#diff-1b8068370be2d846e5ffe4e1592e213810d5b148f949f60deca575648ac66e6cR230

Although I like this more, as it's more readable.

andrei-epure-sonarsource (changes requested):
I think this is a duplicate, but I may be missing something

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.cs
andrei-epure-sonarsource: or maybe the syntax is slightly different there?

andrei-epure-sonarsource: my suggestion is to:
- remove the other example from the CSharp8 test case, as it's not necessarily C# 8 syntax
-  as I said, I like you example more because it's simpler. I'd delete the other one, and update the link in your sample to point to the original issue.
- in the original issue thread, also add this snippet so that we can see the simple syntax directly from the issue , without going to the test case

andrei-epure-sonarsource: and mention here that the problem is in a multi-threaded environment, where the `stop` can be modified in a different thread while this thread is running

andrei-epure-sonarsource (approved):
LGTM

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
```diff
@@ -267,38 +267,6 @@ public void EmptyCollectionCheck2()
...
-                {
-                    if (field == null) // Noncompliant, FP - in multithreading context it makes sense to check for null twice
-                    {
-                        field = new T();
-                    }
-                }
-            }
-        }
-
-        public void Init<T>(ref T field)
```
andrei-epure-sonarsource, line 289: sorry for changing my mind here, but I realize that this example documents one valid Noncompliant and one FP...

- the FP is where there's a `lock` and it seems we talk about a multi-threaded env
- the TP is where there's no `lock`, so no specific suggestion the code should be used in multi-threaded env

maybe we should just drop the PR?

andrei-epure-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
csaba-sagi-sonarsource: I moved the old FP example as we agreed on Slack

andrei-epure-sonarsource (approved):
LGTM!

PR #3978, opened 2021-01-21
Rule S2387, S4025: FieldShadowsParentField

### LINKED ISSUE #3393: Fix S2387 FP: Do not raise issue when 'new' modifier is present
### Description

https://rules.sonarsource.com/csharp/RSPEC-2387
In the example they have shown standard class inheritance but does it not consider the C# 'new' keyword for hiding parent properties. I understand it could still cause confusion as the bug says but I feel it should be allowed in case we explicitly follow C#'s own pragmatics 

### Repro steps

```
class A
{
 protected readonly BaseClass prop;
}
class B : A
{
 private new readonly ChildClass prop; // bug error line
}

```
`// the BaseClass is another class which is inherited by ChildClass
// prop is private readonly property. 
`
### Expected behavior
I would still like to maintain same property name.

I am expecting that adding **new** keyword tells the class to hide the underlying parent field and get rid of this error

### Actual behavior

Error still persists even with or without **new** keyword

### Related information

- C#
-  7.0
- Visual Studio Version - 2017 Enterprise
- Win 10
- Sonar Qube Developer edition 8.1
- Sonar Lint add on latest

### LINKED ISSUE #3969: Rule: S4025 Child class fields should not differ from parent class fields only by capitalization
Implement https://jira.sonarsource.com/browse/RSPEC-4025

### LINKED ISSUE #3971: Rule: S2387 Child class fields should not shadow parent class fields
Implement https://jira.sonarsource.com/browse/RSPEC-2387

pavel-mikula-sonarsource:
Fixes #3393
Fixes #3969
Fixes #3971


analyzers/src/SonarAnalyzer.Common/Rules/FieldShadowsParentFieldBase.cs
```diff
@@ -0,0 +1,77 @@
...
+            s2387 = DiagnosticDescriptorBuilder.GetDescriptor(S2387DiagnosticId, S2387MessageFormat, rspecResources);
+            s4025 = DiagnosticDescriptorBuilder.GetDescriptor(S4025DiagnosticId, S4025MessageFormat, rspecResources);
+        }
+
+        protected IEnumerable<Diagnostic> CheckFields(SemanticModel semanticModel, TVariableDeclaratorSyntax variableDeclarator)
+        {
+            if (semanticModel.GetDeclaredSymbol(variableDeclarator) is IFieldSymbol fieldSymbol)
+            {
+                var fieldName = fieldSymbol.Name;
+                var fieldNameLower = fieldSymbol.Name.ToUpperInvariant();
```
csaba-sagi-sonarsource, line 56: Why is it called fieldNameLower  if it is the result of ToUpperInvariant?

analyzers/src/SonarAnalyzer.VisualBasic/Rules/FieldShadowsParentField.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace SonarAnalyzer.Rules.VisualBasic
+{
+    [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
+    [Rule(S2387DiagnosticId)]
+    [Rule(S4025DiagnosticId)]
+    public sealed class FieldShadowsParentField : FieldShadowsParentFieldBase<ModifiedIdentifierSyntax>
+    {
+        public FieldShadowsParentField() : base(RspecStrings.ResourceManager) { }
+
+        protected override void Initialize(SonarAnalysisContext context) =>
```
csaba-sagi-sonarsource, line 38: I have a feeling that Initialize could have gone to the base class too. Only the SyntaxKind is different plus the foreach line.

csaba-sagi-sonarsource (changes requested):
Overall LGTM, just a few minor changes.

analyzers/src/SonarAnalyzer.VisualBasic/Rules/FieldShadowsParentField.cs
pavel-mikula-sonarsource: I gave it a try, but there's to much syntax pain to do it (`TSyntaxKind`, `TFieldDeclarationSyntax`, `GeneratedCodeRecognizer`, `SyntaxKind` and `FieldVariables`)

analyzers/src/SonarAnalyzer.Common/Rules/FieldShadowsParentFieldBase.cs
pavel-mikula-sonarsource: Good catch, I missed this glitch when copy/pasting the implementation.

csaba-sagi-sonarsource (approved):
LGTM

PR #3981, opened 2021-01-21
Remove useless Helpers sub-namespaces

pavel-mikula-sonarsource:
Remove useless suffixes from
* `SonarAnalyzer.Helpers.CSharp`
* `SonarAnalyzer.Helpers.VisualBasic`
to
* `SonarAnalyzer.Helpers`
because all nested classes are prefixes with the language name anyway


csaba-sagi-sonarsource (approved):
LGTM

andrei-epure-sonarsource (approved):

PR #3982, opened 2021-01-21
Rule: S927 Parameter names should match base declaration

### LINKED ISSUE #3968: Rule: S927 Parameter names should match base declaration
Implement https://jira.sonarsource.com/browse/RSPEC-927

pavel-mikula-sonarsource:
Fixes #3968

Rebase will be needed after #3981 is merged. First commit `Remove useless namespaces` doesn't need a review for the same reason.


csaba-sagi-sonarsource (changes requested):
I would really like to rename the ParameterNamesInPartialMethod classes as Partial methods are not applicable for VB at all and  that is not the only thing the rule does so it is a bit confusing.
Maybe something like ParameterNameBaseDeclarationCheck.

csaba-sagi-sonarsource (approved):
LGTM

PR #3986, opened 2021-01-22
Rule: S3998 Threads should not lock on objects with weak identity

### LINKED ISSUE #3963: Rule: S3998 Threads should not lock on objects with weak identity
Implement https://jira.sonarsource.com/browse/RSPEC-3998

pavel-mikula-sonarsource:
Fixes #3963


csaba-sagi-sonarsource (approved):
LGTM.  PS- it is soo nice to see when almost the whole logic is in the base class.

PR #3997, opened 2021-01-27
Rule: S3884 "CoSetProxyBlanket" and "CoInitializeSecurity" should not be used

### LINKED ISSUE #3959: Rule: S3884 "CoSetProxyBlanket" and "CoInitializeSecurity" should not be used
Implement https://jira.sonarsource.com/browse/RSPEC-3884

csaba-sagi-sonarsource:
Fixes #3959


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SecurityPInvokeMethodShouldNotBeCalled.vb
```diff
@@ -0,0 +1,61 @@
...
+        RequireFullSIC = &H200
+        AutoImpersonate = &H400
+        NoCustomMarshal = &H2000
+        DisableAAA = &H1000
+    End Enum
+
+    <DllImport("ole32.dll")>
+    Public Shared Function CoInitializeSecurity(pVoid As IntPtr, cAuthSvc As Integer, asAuthSvc As IntPtr, pReserved1 As IntPtr, level As RpcAuthnLevel, impers As RpcImpLevel, pAuthList As IntPtr, dwCapabilities As EoAuthnCap, pReserved3 As IntPtr) As Integer
+    End Function
+
```
pavel-mikula-sonarsource, line 51: VB.NET has also alternative (and preferred in VB world) syntax to do this that we should also cover:
```
    Declare Function CoInitializeSecurity Lib "ole32.dll" (pVoid As IntPtr, cAuthSvc As Integer, asAuthSvc As IntPtr, pReserved1 As IntPtr, level As RpcAuthnLevel, impers As RpcImpLevel, pAuthList As IntPtr, dwCapabilities As EoAuthnCap, pReserved3 As IntPtr) As Integer
```
and other variant of the same so we need to check for the identifier in Alias
```
    Declare Function CustomName Lib "ole32.dll" Alias "CoInitializeSecurity" (pVoid As IntPtr, cAuthSvc As Integer, asAuthSvc As IntPtr, pReserved1 As IntPtr, level As RpcAuthnLevel, impers As RpcImpLevel, pAuthList As IntPtr, dwCapabilities As EoAuthnCap, pReserved3 As IntPtr) As Integer
```
The alias has a priority, so let's document in UT that we don't have FP on this:
```
    Declare Function CoInitializeSecurity Lib "ole32.dll" Alias "SomethingElse" (pVoid As IntPtr, cAuthSvc As Integer, asAuthSvc As IntPtr, pReserved1 As IntPtr, level As RpcAuthnLevel, impers As RpcImpLevel, pAuthList As IntPtr, dwCapabilities As EoAuthnCap, pReserved3 As IntPtr) As Integer

File: analyzers/tests/SonarAnalyzer.UnitTest/Rules/SecurityPInvokeMethodShouldNotBeCalledTest.cs [hunk 2]
```diff
@@ -29,8 +30,13 @@ public class SecurityPInvokeMethodShouldNotBeCalledTest
     {
         [TestMethod]
         [TestCategory("Rule")]
-        public void SecurityPInvokeMethodShouldNotBeCalled() =>
+        public void SecurityPInvokeMethodShouldNotBeCalled_CS() =>
             Verifier.VerifyAnalyzer(@"TestCases\SecurityPInvokeMethodShouldNotBeCalled.cs",
                 new CS.SecurityPInvokeMethodShouldNotBeCalled());
```
pavel-mikula-sonarsource, line 35: Let's make the Verifier a single line, just as VB

analyzers/src/SonarAnalyzer.Common/Rules/SecurityPInvokeMethodShouldNotBeCalledBase.cs
```diff
@@ -0,0 +1,100 @@
...
+        protected const string InteropDllName = "ole32.dll";
+
+        protected static readonly ISet<string> InvalidMethods = new HashSet<string>
+        {
+            "CoSetProxyBlanket",
+            "CoInitializeSecurity"
+        };
+
+        protected abstract TSyntaxKind SyntaxKind { get; }
+        protected abstract ILanguageFacade LanguageFacade { get; }
```
pavel-mikula-sonarsource, line 46: I renamed this to `Language` for simplicity in all other rules in my WIP.
```suggestion
        protected abstract ILanguageFacade Language { get; }
```

analyzers/src/SonarAnalyzer.Common/Rules/SecurityPInvokeMethodShouldNotBeCalledBase.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        protected abstract SyntaxNode Expression(TInvocationExpressionSyntax invocationExpression);
```
pavel-mikula-sonarsource, line 48: _Note:_ Once S4583 is merged, this will not be needed anymore. `Language.Syntax.NodeExpression(invocation)` will do the trick.

analyzers/src/SonarAnalyzer.Common/Rules/SecurityPInvokeMethodShouldNotBeCalledBase.cs
```diff
@@ -0,0 +1,100 @@
...
+        protected abstract SyntaxToken Identifier(TIdentifierNameSyntax identifierName);
```
pavel-mikula-sonarsource, line 49: _Note:_ Once S4583 is merged, it would be nice to introduce `Language.Syntax.NodeIdentifier` similar to the `NodeExpression` and use it instead. It repeats a lot.

analyzers/src/SonarAnalyzer.Common/Rules/SecurityPInvokeMethodShouldNotBeCalledBase.cs
```diff
@@ -0,0 +1,100 @@
...
+    public abstract class SecurityPInvokeMethodShouldNotBeCalledBase<TSyntaxKind, TInvocationExpressionSyntax, TIdentifierNameSyntax> : SonarDiagnosticAnalyzer
+        where TSyntaxKind : struct
+        where TInvocationExpressionSyntax : SyntaxNode
+        where TIdentifierNameSyntax : SyntaxNode
+    {
+        protected const string DiagnosticId = "S3884";
+        protected const string MessageFormat = "Refactor the code to remove this use of '{0}'.";
+        protected const string InteropDllName = "ole32.dll";
+
+        protected static readonly ISet<string> InvalidMethods = new HashSet<string>
```
pavel-mikula-sonarsource, line 39: There's a codesmell about static member in generic class. This field doesn't need to be static so you can fix it.

analyzers/src/SonarAnalyzer.Common/Rules/SecurityPInvokeMethodShouldNotBeCalledBase.cs
```diff
@@ -0,0 +1,100 @@
...
+            }
+
+            var dllImportAttribute = methodCallSymbol.Symbol.GetAttributes(KnownType.System_Runtime_InteropServices_DllImportAttribute).FirstOrDefault();
+
+            if (dllImportAttribute == null)
+            {
+                return;
+            }
+
+            if (dllImportAttribute.ConstructorArguments.Any(x => x.Value.Equals(InteropDllName)))
```
pavel-mikula-sonarsource, line 94: It looks to me that this entire section can be changed to 7 `&&` positive conditions.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SecurityPInvokeMethodShouldNotBeCalled.vb
pavel-mikula-sonarsource: I'm not sure how much work it would be to implement the `Alias` correctly and how much can you get from semantic model. 

Let's say `CoInitializeSecurity Lib "ole32.dll" Alias "SomethingElse"` is acceptable FP (let's just document it if we can't handle it). Why would you do it anyway?

I'm worried about `CustomName Lib "ole32.dll" Alias "CoInitializeSecurity"` as current implementation filters the `CustomName` out early as this makes sense for C#.

pavel-mikula-sonarsource (changes requested):
Situation is not so easy for VB in this case

analyzers/src/SonarAnalyzer.Common/Rules/SecurityPInvokeMethodShouldNotBeCalledBase.cs
```diff
@@ -0,0 +1,83 @@
...
+        protected virtual string GetMethodName(ISymbol symbol, SyntaxToken syntaxToken) =>
+            syntaxToken.ValueText;
+
+        private void CheckForIssue(SyntaxNodeAnalysisContext analysisContext)
+        {
+            if (analysisContext.Node is TInvocationExpressionSyntax invocation
+                && Expression(invocation) is TIdentifierNameSyntax directMethodCall
+                && analysisContext.SemanticModel.GetSymbolInfo(directMethodCall) is SymbolInfo methodCallSymbol
+                && methodCallSymbol.Symbol != null
+                && InvalidMethods.Contains(GetMethodName(methodCallSymbol.Symbol, Identifier(directMethodCall)))
```
pavel-mikula-sonarsource, line 74: This name check was originally before symbol retrieving for C#. That's a major performance optimization, because it's rare to have methods named like this. Currently, we'd get symbol for each invocation for C# and that's expensive.

I think it's worth the trouble to have this optimization in place for C# (and re-validate the name for 2nd time after symbol) and to not have it for VB, as there's no other way around it (only the current symbol based check will kick in)

pavel-mikula-sonarsource (changes requested):
No easy task indeed

analyzers/src/SonarAnalyzer.CSharp/Rules/SecurityPInvokeMethodShouldNotBeCalled.cs
```diff
@@ -32,64 +29,24 @@ namespace SonarAnalyzer.Rules.CSharp
...
+        public SecurityPInvokeMethodShouldNotBeCalled() : base(RspecStrings.ResourceManager) { }
 
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        protected override SyntaxNode Expression(InvocationExpressionSyntax invocationExpression) =>
+            invocationExpression.Expression;
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        protected override SyntaxToken Identifier(SyntaxNode syntaxNode) =>
+            ((IdentifierNameSyntax)syntaxNode).Identifier;
```
pavel-mikula-sonarsource, line 43: TBH, I didn't see this type cast coming and your original `TIdentifierNameSyntax` was a clearer solution. But with the `SyntaxFacade` on it's way, I guess we can leave it as it is.

analyzers/src/SonarAnalyzer.VisualBasic/Rules/SecurityPInvokeMethodShouldNotBeCalled.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        protected override string GetMethodName(ISymbol symbol) =>
+            symbol.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax() is DeclareStatementSyntax declaration
+            && declaration.AliasName != null
+                ? declaration.AliasName.GetStringValue()
+                : symbol.Name;
+
+        protected override IMethodSymbol MethodSymbolForInvalidInvocation(SyntaxNode syntaxNode, SemanticModel semanticModel) =>
+            semanticModel.GetSymbolInfo(syntaxNode).Symbol is IMethodSymbol methodSymbol
+            && InvalidMethods.Any(x => x.Equals(GetMethodName(methodSymbol), StringComparison.OrdinalIgnoreCase))
```
pavel-mikula-sonarsource, line 60: `GetMethodName` is expensive and it's currently inside loop. Let's extract it to `&& GetMethodName(..) is {} methodName`.

Like that, it would be better to swap the operands to `methodName.Equals(x, ...)`

pavel-mikula-sonarsource (approved):
LGTM, Nice! I just left one performance comment to fix before merging.

pavel-mikula-sonarsource:
The coverage issue is purely defensive coding.

analyzers/src/SonarAnalyzer.Common/Rules/SecurityPInvokeMethodShouldNotBeCalledBase.cs
```diff
@@ -0,0 +1,87 @@
...
+            "CoInitializeSecurity"
+        };
+
+        protected SecurityPInvokeMethodShouldNotBeCalledBase(System.Resources.ResourceManager rspecResources) =>
+            Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, rspecResources);
+
+        protected override void Initialize(SonarAnalysisContext context) =>
+            context.RegisterSyntaxNodeActionInNonGenerated(Language.GeneratedCodeRecognizer, CheckForIssue, SyntaxKind);
+
+        protected virtual bool IsImportFromInteropDll(ISymbol symbol) =>
```
andrei-epure-sonarsource, line 63: For the coverage of `SecurityPInvokeMethodShouldNotBeCalledBase.cs` , [lines 63-66](https://sonarcloud.io/code?id=sonaranalyzer-dotnet&pullRequest=3997&selected=sonaranalyzer-dotnet%3Aanalyzers%2Fsrc%2FSonarAnalyzer.Common%2FRules%2FSecurityPInvokeMethodShouldNotBeCalledBase.cs&line=63), there is room for coverage improvement.

`DllImportAttribute` has [several fields](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.dllimportattribute.bestfitmapping?view=net-5.0) which can be passed to the Attribute creation.

You could add some snippets that compile with different combinations of arguments, like:

```
    <DllImport("oLe32" , BestFitMapping:=false, CallingConvention:=CallingConvention.FastCall)>
    Public Shared Function CoSetProxyBlanket22() As Integer
    End Function
```

And some snippets with compile errors (that can happen in the IDE, as the analysis is done during build on WIP code)

```
        [DllImport(BestFitMapping = false)] // Error [CS7036]
        public static extern int CoInitializeSecurity1();

        [DllImport(BestFitMapping =false, EntryPoint="ole32.dll", ExactSpelling = true)] // Error [CS7036]
        public static extern int CoInitializeSecurity2();
```

andrei-epure-sonarsource (changes requested):
I think there's a minor coverage improvement to be made

andrei-epure-sonarsource (approved):
LGTM!

Issue #3970, opened 2021-01-18
Rule: S2437 Silly bit operations should not be performed


pavel-mikula-sonarsource:
Implement https://jira.sonarsource.com/browse/RSPEC-2437

PR #4009, opened 2021-02-01
Fix S2437 FP: Tracked Enum variable initialized to zero

pavel-mikula-sonarsource:
Fix FP found on Peach, related to #3970


csaba-sagi-sonarsource (approved):
LGTM

PR #4011, opened 2021-02-01
Rule S1125: Add code example from community

costin-zaharia-sonarsource:
See: https://community.sonarsource.com/t/s1905-false-positives-generated-within-c-ternary-operator-statement/36717


csaba-sagi-sonarsource (approved):
LGTM

PR #4019, opened 2021-02-02
Remove S3949 from SonarWay

### LINKED ISSUE #4008: Remove S3949 rule from SonarWay (until CFG gets better)
Currently, [S3949](https://rules.sonarsource.com/csharp/RSPEC-3949) has the CBDE symbolic execution engine inside which works on the MLIR format.

The problem with it is that, due to some CFG structual problems we have, it will ignore the methods which contain unknown syntaxes, and the list of unsupported syntaxes is quite bit. Thus, it will ignore a big part of the analyzed code (have small benefit).

In the same time:
- we've seen on the community and the internal forum, the logs produces by the CBDE engine and also the fact that it's running in an external process brings confusion both for users and for us.
- the rule sometimes performs badly in terms of speed (with low added value, as mentioned above)

We need to remove this rule from SonarWay for C# , until the current implementation will be better and the gain will be bigger than the pain.

andrei-epure-sonarsource:
Fixes #4008


costin-zaharia-sonarsource (approved):
LGTM!

Issue #4015, opened 2021-02-02
Fix S3353 FP: Variable is used within an Expression


lg2de:
### Description

S3353 is reported for variable if it is used within an `Expression`.
Following the suggestion to make the variable `const` will change the expression.

If the expression is analyzed in frameworks like "NHibernate" the changed expression cause totally different behavior.

I think the rule should not apply for variables using within any `Expression`.

### Repro steps

```cs
namespace SonarDotnetS3353
{
    public class Class1
    {
        public void Method()
        {
            string localVariable = null;

            this.ExpressionMethod(() => localVariable);
        }

        private void ExpressionMethod(Expression<Func<string>> expression)
        {
            if (expression.Body is MemberExpression)
            {
                return;
            }

            throw new InvalidOperationException(
                $"The expression body is a {expression.Body.GetType().Name}, but FieldExpression is expected.");
        }
    }
}
```

When changing `localVariable` to `const string` the analyzing method throws an exception.

### Expected behavior

Rule S3353 should not apply.

### Actual behavior

`SonarDotnetS3353\Class1.cs(10,20,10,41): warning S3353: Add the 'const' modifier to 'localVariable'.`

### Known workarounds

N/A

### Related information

* C#/VB.NET Plugins version: 8.17.0.26580
* Visual Studio version: 2019 16.8.3
* dotnetcore version: 3.1
* Operating System: Windows

pavel-mikula-sonarsource:
Hi @lg2de ,

Thank you for reporting this interesting case. I confirm it as a False Positive.

lg2de:
Maybe the next days I'll try to solve this with a PR.

lg2de:
I failed to fix it. The variations of `Expression` I faced are too complex (for me).

There seams already to be a check for variable usage within `Expression`, but it is limited to assignment.
I'm not sure whether every usage within an expression should stop raising S3533 or not.

pavel-mikula-sonarsource:
Hi @lg2de, thank you for giving it a try.

PR #4022, opened 2021-02-03
Add FP repro for #4015

pavel-mikula-sonarsource:
Reproduces #4015


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UnchangedLocalVariablesShouldBeConst.cs
```diff
@@ -347,3 +348,23 @@ public NormalClass NormalClassSetToNull()
 
     }
 }
+
+// https://github.com/SonarSource/sonar-dotnet/issues/4015
```
andrei-epure-sonarsource, line 352: interesting! should the issue be only about `Expression` or , in general, classes from `System.Linq.Expressions`?

andrei-epure-sonarsource (approved):
LGTM

PR #4023, opened 2021-02-03
Fix NRE S4583 and S5547: Invocation expression is null for VB

csaba-sagi-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Rules/BeginInvokePairedWithEndInvokeBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/BeginInvokePairedWithEndInvokeBase.cs
@@ -53,7 +53,8 @@ namespace SonarAnalyzer.Rules
             context.RegisterSyntaxNodeActionInNonGenerated(Language.GeneratedCodeRecognizer, c =>
             {
                 var invocation = (TInvocationExpressionSyntax)c.Node;
-                if (Language.Syntax.NodeExpression(invocation).ToStringContains(BeginInvoke)
+                if (Language.Syntax.NodeExpression(invocation) is { } expression
+                    && expression.ToStringContains(BeginInvoke)
                     && c.SemanticModel.GetSymbolInfo(invocation).Symbol is IMethodSymbol methodSymbol
                     && methodSymbol.Name == BeginInvoke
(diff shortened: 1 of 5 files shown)


analyzers/src/SonarAnalyzer.Common/Rules/BeginInvokePairedWithEndInvokeBase.cs
```diff
@@ -53,7 +53,8 @@ public abstract class BeginInvokePairedWithEndInvokeBase<TSyntaxKind, TInvocatio
             context.RegisterSyntaxNodeActionInNonGenerated(Language.GeneratedCodeRecognizer, c =>
             {
                 var invocation = (TInvocationExpressionSyntax)c.Node;
-                if (Language.Syntax.NodeExpression(invocation).ToStringContains(BeginInvoke)
+                if (Language.Syntax.NodeExpression(invocation) is { } expression
```
pavel-mikula-sonarsource, line 56: Can you add a testcase to rule specific file as well?

analyzers/src/SonarAnalyzer.Common/Rules/DoNotCallInsecureSecurityAlgorithmBase.cs
```diff
@@ -61,12 +61,9 @@ private void CheckInvocation(SyntaxNodeAnalysisContext context)
         {
             var invocation = (TInvocationExpressionSyntax)context.Node;
 
-            if (!(context.SemanticModel.GetSymbolInfo(InvocationExpression(invocation)).Symbol is IMethodSymbol methodSymbol))
-            {
-                return;
-            }
-
-            if (methodSymbol.ReturnType.DerivesFromAny(AlgorithmTypes) || IsInsecureBaseAlgorithmCreationFactoryCall(methodSymbol, invocation))
+            if (InvocationExpression(invocation) is { } expression
```
pavel-mikula-sonarsource, line 64: Can you add a testcase to rule specific file as well?

pavel-mikula-sonarsource (changes requested):
Two more testcases please

pavel-mikula-sonarsource (approved):
LGTM

PR #4027, opened 2021-02-05
Add OpenCover support for deterministic build

### LINKED ISSUE #3976: Add coverage support for Deterministic Source Paths (OpenCover format)
This is a follow-up of the work done in #3817.

To reproduce the problem, clone https://github.com/costin-zaharia-sonarsource/Repro3362 and run the commands:
```
dotnet sonarscanner begin /k:repro3362 /d:sonar.verbose=true /d:sonar.cs.opencover.reportsPaths="opencover.xml"

dotnet build --no-incremental -p:DeterministicSourcePaths=true

OpenCover.Console.exe -output:"opencover.xml" -register:administrator -target:"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe" -targetargs:".\Repro3362.Tests\bin\Debug\netcoreapp3.1\Repro3362.Tests.dll"

dotnet sonarscanner end
```

Actual coverage in SQ: 0%.
Expected: Should list some coverage. The OpenCover console says:
>Visited Classes 2 of 3 (66.67), Visited Methods 2 of 4 (50), Visited Points 4 of 9 (44.44), Visited Branches 2 of 4 (50)

In the OpenCover.xml we see
```
      <Files>
        <File uid="5" fullPath="/_/Repro3362/Foo.cs" />
      </Files>
```
so we should be able to support it by updating `OpenCoverReportParser.java` with the same logic as in `VisualStudioCoverageXmlReportParser.java`.

_After the implementation, we could ask @TimHess to confirm it works with their setup (see [comment](https://github.com/SonarSource/sonar-dotnet/issues/3362#issuecomment-763875778))._

andrei-epure-sonarsource:
Fixes #3976

### DIFF
--- a/its/src/test/java/com/sonar/it/csharp/CoverageTest.java
+++ b/its/src/test/java/com/sonar/it/csharp/CoverageTest.java
@@ -113,7 +113,7 @@ public class CoverageTest {
 
     assertThat(linesToCover.getValue()).isEqualTo("6");
     assertThat(uncoveredLines.getValue()).isEqualTo("3");
-    assertCoverageMetrics("CoverageWithDeterministicSourcePaths", 6, 3, 9, 4);
+    assertCoverageMetrics("CoverageWithDeterministicSourcePaths", 6, 3, 2, 1);
   }
 
   @Test
--- a/its/src/test/java/com/sonar/it/csharp/CoverageTest.java
+++ b/its/src/test/java/com/sonar/it/csharp/CoverageTest.java
@@ -27,7 +27,7 @@ import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
-import org.sonarqube.ws.Measures;
+import org.sonarqube.ws.Measures.Measure;
 
 import static com.sonar.it.csharp.Tests.ORCHESTRATOR;
 import static com.sonar.it.csharp.Tests.getMeasure;
@@ -52,8 +52,8 @@ public class CoverageTest {
       .doesNotContain("Sensor C# Tests Coverage Report Import")
       .doesNotContain("Coverage Report Statistics:");
 
-    Measures.Measure linesToCover = getMeasure("CoverageTest", "lines_to_cover");
-    Measures.Measure uncoveredLines = getMeasure("CoverageTest", "uncovered_lines");
+    Measure linesToCover = getMeasure("CoverageTest", "lines_to_cover");
+    Measure uncoveredLines = getMeasure("CoverageTest", "uncovered_lines");
 
     assertThat(linesToCover.getValue()).isEqualTo("2");
     assertThat(uncoveredLines.getValue()).isEqualTo("2");
@@ -109,8 +109,8 @@ public class CoverageTest {
       "Sensor C# Tests Coverage Report Import",
       "Coverage Report Statistics: 1 files, 1 main files, 1 main files with coverage, 0 test files, 0 project excluded files, 0 other language files.");
 
-    Measures.Measure linesToCover = getMeasure("CoverageWithDeterministicSourcePaths", "lines_to_cover");
-    Measures.Measure uncoveredLines = getMeasure("CoverageWithDeterministicSourcePaths", "uncovered_lines");
+    Measure linesToCover = getMeasure("CoverageWithDeterministicSourcePaths", "lines_to_cover");
+    Measure uncoveredLines = getMeasure("CoverageWithDeterministicSourcePaths", "uncovered_lines");
 
     assertThat(linesToCover.getValue()).isEqualTo("6");
     assertThat(uncoveredLines.getValue()).isEqualTo("3");
@@ -162,8 +162,8 @@ public class CoverageTest {
       "Sensor C# Tests Coverage Report Import",
       "Coverage Report Statistics: 1 files, 1 main files, 1 main files with coverage, 0 test files, 0 project excluded files, 0 other language files.");
 
-    Measures.Measure linesToCover = getMeasure("CoverageWithDeterministicSourcePaths", "lines_to_cover");
-    Measures.Measure uncoveredLines = getMeasure("CoverageWithDeterministicSourcePaths", "uncovered_lines");
+    Measure linesToCover = getMeasure("CoverageWithDeterministicSourcePaths", "lines_to_cover");
+    Measure uncoveredLines = getMeasure("CoverageWithDeterministicSourcePaths", "uncovered_lines");
 
     assertThat(linesToCover.getValue()).isEqualTo("6");
     assertThat(uncoveredLines.getValue()).isEqualTo("3");
--- a/its/src/test/java/com/sonar/it/csharp/CoverageTest.java
+++ b/its/src/test/java/com/sonar/it/csharp/CoverageTest.java
@@ -101,6 +101,22 @@ public class CoverageTest {
   }
 
   @Test
+  public void open_cover_with_deterministic_source_paths() throws Exception {
+    BuildResult buildResult = Tests.analyzeProject(temp, "CoverageWithDeterministicSourcePaths", "no_rule", "sonar.cs.opencover.reportsPaths", "opencover.xml");
+
+    assertThat(buildResult.getLogs()).contains(
+      "Sensor C# Tests Coverage Report Import",
+      "Coverage Report Statistics: 1 files, 1 main files, 1 main files with coverage, 0 test files, 0 project excluded files, 0 other language files.");
+
+    org.sonarqube.ws.Measures.Measure linesToCover = getMeasure("CoverageWithDeterministicSourcePaths", "lines_to_cover");
+    org.sonarqube.ws.Measures.Measure uncoveredLines = getMeasure("CoverageWithDeterministicSourcePaths", "uncovered_lines");
+
+    assertThat(linesToCover.getValue()).isEqualTo("6");
+    assertThat(uncoveredLines.getValue()).isEqualTo("3");
+    assertCoverageMetrics("CoverageWithDeterministicSourcePaths", 6, 3, 9, 4);
+  }
+
+  @Test
   public void dotcover() throws Exception {
     BuildResult buildResult = analyzeCoverageTestProject("sonar.cs.dotcover.reportsPaths", "reports/dotcover.html");
 
--- a/its/src/test/java/com/sonar/it/csharp/CoverageTest.java
+++ b/its/src/test/java/com/sonar/it/csharp/CoverageTest.java
@@ -44,21 +44,6 @@ public class CoverageTest {
   }
 
   @Test
-  public void should_import_coverage_with_deterministic_source_paths() throws Exception {
-    BuildResult buildResult = Tests.analyzeProject(temp, "CoverageWithDeterministicSourcePaths", "no_rule", "sonar.cs.vscoveragexml.reportsPaths", "VisualStudio.coveragexml");
-
-    assertThat(buildResult.getLogs()).contains(
-      "Sensor C# Tests Coverage Report Import",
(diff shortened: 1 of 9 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -108,17 +120,17 @@ private void handleSequencePointTag(XmlParserHelper xmlParserHelper) {
...
 
       if (files.containsKey(fileId)) {
-        String filePath = files.get(fileId);
+        Optional<String> optionalPath = files.get(fileId);
 
-        if (fileService.isSupportedAbsolute(filePath)) {
-          LOG.trace("OpenCover parser: add hits for fileId '{}', filePath '{}', line '{}', visitCount '{}'",
-            fileId, filePath, line, visitCount);
+        if (optionalPath.isPresent()) {
+          String absoluteFilePath = optionalPath.get();
```
costin-zaharia-sonarsource, line 126: I think the name can be misleading since, at  this point, we don't know for sure if the path is absolute or relative. We can name it just `filePath` or `path`.

costin-zaharia-sonarsource: I thought that a functional approach with `ifPresentOrElse` would work better but it seems that we need to set the language version to 9 in order to be able to do that.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -108,17 +120,17 @@ private void handleSequencePointTag(XmlParserHelper xmlParserHelper) {
...
+          coverage.addHits(absoluteFilePath, line, visitCount);
 
-          coverage.addHits(filePath, line, visitCount);
+          LOG.trace("OpenCover parser: add hits for fileId '{}', filePath '{}', line '{}', visitCount '{}'",
+            fileId, absoluteFilePath, line, visitCount);
         } else {
-          LOG.debug("Skipping the fileId '{}', line '{}', vc '{}' because file '{}'" +
-              " is not indexed or does not have the supported language.",
-            fileId, line, visitCount, filePath);
+          LOG.debug("Skipping the fileId '{}', line '{}', visitCount '{}' because file is not indexed or does not have the supported language.",
```
costin-zaharia-sonarsource, line 132: Can we read the full path from xml in order to be able to print it? 

I'm thinking that it can be helpful when analyzing the logs to see directly the file name. Otherwise we will have to open the report and search for fileId.

costin-zaharia-sonarsource (commented):
Just a couple of questions. Otherwise I think it looks good.

sonar-dotnet-shared-library/src/test/resources/opencover/deterministic_source_paths.xml
```diff
@@ -0,0 +1,958 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: yes, and it's the same for VS coverage. Deterministic builds are just for production code.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
andrei-epure-sonarsource: good point, I should update to Java 11 which is the current LTS

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
andrei-epure-sonarsource: This is a conscious change I did. We read file paths in a different function (`handleFileTag`). Previously, we were storing the path from the coverage entry (actually, the "canonicalPath") in the `files` dictionary (map). Now, I store `Optional<String>` instead of the path.

Why did I do it? To avoid doing IO operations _(get the canonical path, to the scanner, to `fileService`)_ for each coverage and branch coverage entry.

It's a trade off here. I'm fine also with changing the approach and do the following instead:
- storing the original path from the coverage entry in the `files` map _(we need this for the deterministic build path resolution)_
- for each (branch) coverage entry: get the canonical path, and then make the `fileService` call if necessary

For the scanner search, I imagine it keeps all indexed files in memory. However, whenever it would find a not-indexed file, it will go to the filesystem search.

WDYT? It's a tradeoff between performance (and for large codebases, I imagine it could be a hit) and logs usefulness.

costin-zaharia-sonarsource: From what I've noticed so far is that from time to time we do have either questions or issues related to coverage which are sometimes hard to investigate. I think that both performance and logging details are important and I think that having them both would be great 😄 

The solution you suggested with a class which contains both the original path and the resolved one by the file service sounds very good.

its/src/test/java/com/sonar/it/csharp/CoverageTest.java
```diff
@@ -66,8 +52,8 @@ public void should_not_import_coverage_without_report() throws Exception {
       .doesNotContain("Sensor C# Tests Coverage Report Import")
       .doesNotContain("Coverage Report Statistics:");
 
-    org.sonarqube.ws.Measures.Measure linesToCover = getMeasure("CoverageTest", "lines_to_cover");
-    org.sonarqube.ws.Measures.Measure uncoveredLines = getMeasure("CoverageTest", "uncovered_lines");
+    Measures.Measure linesToCover = getMeasure("CoverageTest", "lines_to_cover");
+    Measures.Measure uncoveredLines = getMeasure("CoverageTest", "uncovered_lines");
```
costin-zaharia-sonarsource, line 56: pedantic: you can use `import org.sonarqube.ws.Measures.Measure;` instead of `import org.sonarqube.ws.Measures;` to simplify this even more.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -145,17 +164,54 @@ private void handleBranchPointTag(XmlParserHelper xmlParserHelper) {
...
   }
+
+  /**
+   * Represents a file mentioned in the OpenCover report list of files, which may or may not be supported by the FileService.
+   * @see FileService
+   */
+  private class CoveredFile
+  {
+    // the file ID from the OpenCover report
+    private String uid;
```
costin-zaharia-sonarsource, line 190: We should make the fields final since they are not meant to be modified.

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParserTest.java
```diff
@@ -118,6 +119,119 @@ public void valid() throws Exception {
 
     List<String> debugLogs = logTester.logs(LoggerLevel.DEBUG);
     assertThat(debugLogs.get(0)).startsWith("The current user dir is '");
+
+    // FIXME the test case may be wrong
```
costin-zaharia-sonarsource, line 123: Should we add an issue or card for this to be sure we don't forget to address it?

costin-zaharia-sonarsource (changes requested):
I've left some comments and I've also noticed that there are 2 unhandled code smells.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -46,8 +48,8 @@ public void accept(File file, Coverage coverage) {
 
     private final File file;
     private final Coverage coverage;
-    // the key is the file ID, the value is the file path
-    private final Map<String, String> files = new HashMap<>();
+    // the key is the file ID, the value is the CoveredFile
+    private final Map<String, CoveredFile> files = new HashMap<>();
     // the key is the file path
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: nice catch!

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -145,17 +164,54 @@ private void handleBranchPointTag(XmlParserHelper xmlParserHelper) {
...
+   * @see FileService
+   */
+  private class CoveredFile
+  {
+    // the file ID from the OpenCover report
+    private String uid;
+    // the path from the coverage file
+    private String originalPath;
+    // the path which is supported by the FileService
+    private String supportedPath;
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: I thought of this name as well. Indeed, "supported" is a bit too ambiguous.
I also thought about "scannerPath" or "indexedPath", as it needs to be supported by the scanner. 

I'll use "indexedPath" because that's actually what we need for coverage to work: the path needs to be known to the scanner.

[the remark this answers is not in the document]
costin-zaharia-sonarsource: Sounds good!

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -86,14 +88,31 @@ private void handleFileRef(XmlParserHelper xmlParserHelper) {
...
+        // maybe it's a deterministic build file path
+        Optional<String> optionalPath = fileService.getAbsolutePath(pathInCoverageFile);
+        if (optionalPath.isPresent()) {
+          coveredFile = new CoveredFile(uid, pathInCoverageFile, optionalPath.get());
+          LOG.debug("Found indexed file '{}' for coverage entry '{}'.", coveredFile.supportedPath, coveredFile.originalPath);
+        } else {
+          coveredFile = new CoveredFile(uid, pathInCoverageFile, null);
+        }
+      }
+      files.put(uid, coveredFile);
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: I was logging at debug level when we were resolving the deterministic path because that's only what I considered as needed to be logged.

If the path is not found at all, it's being logged in different places (for SequencePoint / BranchPoint), and now we have some redundancy - see the new test cases.

However, I'll keep this redundancy and see if it will bother us in practice.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/OpenCoverReportParser.java
```diff
@@ -212,7 +212,7 @@ boolean isIndexed() {
     @Override
     public String toString() {
       return indexedPath == null
-        ? String.format("(ID '%s', path '%s')", uid, originalPath)
+        ? String.format("(ID '%s', path '%s', NO INDEXED PATH)", uid, originalPath)
```
andrei-epure-sonarsource, line 215: I prefer to make it obvious

costin-zaharia-sonarsource (approved):
LGTM! 🚀

PR #4028, opened 2021-02-05
Deprecate S4261

### LINKED ISSUE #3928: Fix S4261 RSPEC: Rule is based off of errant examples and not the text rule.
Both the compliant and noncompliant examples are missing the async modifier for the function. That is what determines if the method should be suffixed with Async or not.

current compliant example
``` csharp
public class Foo
  {
    public Task ReadAsync(byte [] buffer, int offset, int count, CancellationToken cancellationToken)
  }
```
what it should be
``` csharp
public class Foo
  {
    public async Task ReadAsync(byte [] buffer, int offset, int count, CancellationToken cancellationToken) {}
  }
```
… though even this is strange as the method name is Read but this doesnt return the result of reading anything.

What looks like is that this rule has referenced the example code that is incorrect, and not the [section](https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap#naming-parameters-and-return-types) that states the guidance.

> Asynchronous methods in TAP include the `Async` suffix after the operation name for methods that return awaitable types, such as [Task](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task), [Task<TResult>](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task-1), [ValueTask](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask), and [ValueTask<TResult>](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1). For example, an asynchronous `Get` operation that returns a `Task<String>` can be named `GetAsync`

Also posted at https://community.sonarsource.com/t/s4261-has-incorrect-examples/36324/3

costin-zaharia-sonarsource:
Fixes #3928

More details here: https://community.sonarsource.com/t/s4261-has-incorrect-examples/36324/7

### DIFF
--- a/analyzers/rspec/cs/S4261_c#.html
+++ b/analyzers/rspec/cs/S4261_c#.html
@@ -37,4 +37,6 @@ namespace myLibrary
   <li> <a href="https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap">Task-based
   Asynchronous Pattern</a> </li>
 </ul>
+<h2>Deprecated</h2>
+<p>This rule is deprecated, and will eventually be removed.</p>
 
--- a/analyzers/rspec/cs/S4261_c#.json
+++ b/analyzers/rspec/cs/S4261_c#.json
@@ -1,7 +1,7 @@
 {
   "title": "Methods should be named according to their synchronicities",
   "type": "CODE_SMELL",
-  "status": "ready",
+  "status": "deprecated",
   "remediation": {
     "func": "Constant\/Issue",
     "constantCost": "5min"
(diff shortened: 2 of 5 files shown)


analyzers/rspec/cs/S4261_c#.html
```diff
@@ -37,4 +37,6 @@ <h2>See</h2>
   <li> <a href="https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap">Task-based
   Asynchronous Pattern</a> </li>
 </ul>
+<h2>Deprecated</h2>
```
pavel-mikula-sonarsource, line 40: We should also put something like `[Obsolete("This rule is deprecated in favor of S4790")]` to the class like here https://github.com/SonarSource/sonar-dotnet/blob/1043838c7665a38fd2cb409506f82c5dab4e5a1b/analyzers/src/SonarAnalyzer.CSharp/Rules/InsecureHashAlgorithm.cs#L33-L37

pavel-mikula-sonarsource (changes requested):
Shouldn't there be "Fixes #3928" in the description? As it's a solution for that problem.

pavel-mikula-sonarsource (approved):
LGTM

PR #4030, opened 2021-02-05
Tracker refactoring

pavel-mikula-sonarsource:
Key ideas:
- remove duplicated language specific tracker initialization from rules
- remove tracker storage in specific rules
- remove state from trackers, as they don't need it and there's lot of duplicated code to propagate it
- add TrackerFacade that provides all trackers


analyzers/src/SonarAnalyzer.Common/DiagnosticAnalyzer/TrackerDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        protected abstract ILanguageFacade<TSyntaxKind> Language { get; }
+        protected abstract void Initialize(TrackerInput input);
+
+        public sealed override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        protected TrackerHotspotDiagnosticAnalyzer(IAnalyzerConfiguration configuration, string diagnosticId, string messageFormat, ResourceManager rspecResources) : base(configuration)
+        {
+            Rule = DiagnosticDescriptorBuilder.GetDescriptor(diagnosticId, messageFormat, rspecResources);
+            if (configuration == AnalyzerConfiguration.Hotspot)
```
csaba-sagi-sonarsource, line 41: I am not sure it this check is needed,  because that way non hotspot rules could also inherit this class and would work, but  if that is the case than why is it called TrackerHotspotDiagnosticAnalyzer?

analyzers/src/SonarAnalyzer.Common/Rules/BypassingAccessibilityBase.cs
```diff
@@ -18,24 +18,26 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
 
 namespace SonarAnalyzer.Rules
 {
-    public abstract class BypassingAccessibilityBase<TSyntaxKind> : SonarDiagnosticAnalyzer
+    public abstract class BypassingAccessibilityBase<TSyntaxKind> : TrackerHotspotDiagnosticAnalyzer<TSyntaxKind>
```
csaba-sagi-sonarsource, line 26: This is not a Hotspot, it is a code smell.

analyzers/src/SonarAnalyzer.Common/DiagnosticAnalyzer/TrackerDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+using System.Collections.Immutable;
+using System.Resources;
+using Microsoft.CodeAnalysis;
+using SonarAnalyzer.Common;
+
+namespace SonarAnalyzer.Helpers
+{
+    public abstract class TrackerHotspotDiagnosticAnalyzer<TSyntaxKind> : HotspotDiagnosticAnalyzer
```
csaba-sagi-sonarsource, line 28: The file name and the class name are not the same

csaba-sagi-sonarsource (changes requested):
TrackerHotspotDiagnosticAnalyzer  is used for non hotspot rules also. If that was the intention than the naming is a bit misleading, otherwise it should not be used for not hotspot rules.

csaba-sagi-sonarsource:
Also seems to me that by time the Language facade will be part of almost every analyzer, so maybe we could consider it as a candidate to move into the SonarDiagnosticAnalyzer class?

analyzers/src/SonarAnalyzer.Common/Rules/BypassingAccessibilityBase.cs
pavel-mikula-sonarsource: But it uses a lot of functionality from the base class so it's handy to use it as well. As in the other comment, I plan to rename those base classes in the future. This will be probably something like `TrackerDiagnosticAnalyzer` and the `HotspotDiagnosticAnalyzer` will be something like `ConfigurableDiagnosticAnalyzer`

pavel-mikula-sonarsource:
I plan to do renaming in another PR. And introduce new base classes like the one with `LanguageFacade` - It will not be used in all rules, but it will be handy in many rules common for C# and VB.NET. Another idea for new base class is a base to deal with the repetitive `Rule` and `SupportedDiagnostic` issue scaffolding. The parameter loading base also needs a strict look.

csaba-sagi-sonarsource (approved):
LGTM

PR #4031, opened 2021-02-05
Moving hotspot rules to dedicated folders

csaba-sagi-sonarsource:


Corniel:
Just out of curiosity, what do you want to achieve with this? If a rule is a _hotspot_ or not feels quite arbitrary.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/NonStandardCryptographicAlgorithmsShouldNotBeUsedTest.cs
```diff
@@ -33,28 +33,28 @@ public class NonStandardCryptographicAlgorithmsShouldNotBeUsedTest
         [TestMethod]
         [TestCategory("Rule")]
         public void NonStandardCryptographicAlgorithmsShouldNotBeUsed_CSharp() =>
-            Verifier.VerifyAnalyzer(@"TestCases\NonStandardCryptographicAlgorithmsShouldNotBeUsed.cs",
+            Verifier.VerifyAnalyzer(@"TestCases\Hotspots\NonStandardCryptographicAlgorithmsShouldNotBeUsed.cs",
```
pavel-mikula-sonarsource, line 36: While looking at the files in the directory, this rule has too long name. Let's rename this type and all related stuff to `UsingNonstandardCryptography`

pavel-mikula-sonarsource (approved):
LGTM. Let's rename one of the rules while we're on move/rename topic.

Master if frozen now, so let's don't merge it before version bumping.

csaba-sagi-sonarsource:
> Just out of curiosity, what do you want to achieve with this? If a rule is a _hotspot_ or not feels quite arbitrary.
Our plan is to organize a bit better the solution, and move things into subfolders and that way increase the navigability.

PR #3791, opened 2020-11-30
S5944: Use explicit return statements

Corniel:
Fixes #2159

Looking back to some old work, I noticed I never pushed #2159.
As everything is already [documented](https://jira.sonarsource.com/browse/RSPEC-5944), it should be a walk in the park.

Retry of #3789

PR #4032, opened 2021-02-06
Rule S5944: Use explicit return statements

### LINKED ISSUE #2159: Rule S5944: VB.NET functions should use the return statement
### Description
In Visual Basic, it allowed by to assign the return value to a variable (that does not have to be declared) with the same name of the function, instead of using a return statement. That is a bad practice that should be discouraged.

```vb
Public Class UseReturnStatement

	Public Function FunctionName() As Integer
		FunctionName = 42 ' Not compliant
	End Function

	Public Function ReturnStatement() As Integer
		Return 42 ' Compliant
	End Function

	Public Function ReturnWithVariable() As Integer
		Dim value As Integer = 42
		Return value  ' compliant
	End Function


	Public Function ReturnWithFunctionName() As Integer
		ReturnWithFunctionName= 42
		Return ReturnWithFunctionName  ' Not compliant
	End Function

End Class
```

Corniel:
Fixes #2159

Due to issues with #3791


pavel-mikula-sonarsource (approved):
LGTM! Thank you @Corniel for your contribution. 🎉 

Full review history is in #3791 that had troubles with Burgr step in CI.

Issue #3735, opened 2020-11-10
Rule S3981 : should also check `<0`


andrei-epure-sonarsource:
The [RSPEC for S3981](https://jira.sonarsource.com/browse/RSPEC-3981) says

>So testing that a size or length is greater than or equal to zero doesn't make sense, since the result is always true. Similarly testing that it is less than zero will always return false. 

Currently, the rule only checks for `>=0`, not also for `<0`.

Corniel:
Consider pattern matching similar to what have been introduced in #3857:
```
count() <= 0
count() < 0
count() <= negative
count() < negative
// reverse too
```

PR #3857, opened 2020-12-17
Adding fix provider for S1155

Corniel:
Fixes #2334. All replacements seem okay, the formatting (and the test) are a bit off. Help is wanted for that part.

PR #4033, opened 2021-02-06
Introduction of CountComparisonSyntax

Corniel:
There are two rules involved for checking `Count`, `Length`, and `Count()`, on collections. Those are implemented for both languages and with a Fix for one on the way (#3857) the code duplication involved to do so, felt like a burden to me.

So with this helper syntax (struct), I tried to centralize the different types of Count comparisons we might be interested in:
* equivalent to Any()
* equivalent to !Any()
* Without meaning (`< 0`, `>= 0`, `!= -1`, `== -1`)

If a `BinaryExpressionSyntax` is a relavent count expression is not taken into account yet (up to the rule in play). Reason for that is that S1155 is not triggered by IList, ICollection, Array etc,  and S3981 is. It is beyond the scope of this proposal to decide on that here I think.

Just another remark: most of this code is already in #3857, but internal, C# only, and only taking Any() and !Any() into account.

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
@@ -31,6 +31,18 @@ namespace SonarAnalyzer.Helpers.Facade
     {
         public override SyntaxKind Kind(SyntaxNode node) => node.Kind();
 
+        public override ComparisonKind ComparisonKind(SyntaxNode node) =>
+           Kind(node) switch
+           {
+               SyntaxKind.EqualsExpression => Helpers.ComparisonKind.Equals,
+               SyntaxKind.NotEqualsExpression => Helpers.ComparisonKind.NotEquals,
+               SyntaxKind.LessThanExpression => Helpers.ComparisonKind.LessThan,
+               SyntaxKind.LessThanOrEqualExpression => Helpers.ComparisonKind.LessThanOrEqual,
+               SyntaxKind.GreaterThanExpression => Helpers.ComparisonKind.GreaterThan,
+               SyntaxKind.GreaterThanOrEqualExpression => Helpers.ComparisonKind.GreaterThanOrEqual,
+               _ => Helpers.ComparisonKind.None,
+           };
+
         public override bool IsKind(SyntaxNode node, SyntaxKind kind) => node.IsKind(kind);
 
         public override bool IsAnyKind(SyntaxNode node, ISet<SyntaxKind> syntaxKinds) => node.IsAnyKind(syntaxKinds);
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessChecking.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessChecking.cs
@@ -36,115 +36,70 @@ namespace SonarAnalyzer.Rules.CSharp
         internal const string DiagnosticId = "S1155";
         private const string MessageFormat = "Use '.Any()' to test whether this 'IEnumerable<{0}>' is empty or not.";
 
-        private static readonly CSharpExpressionNumericConverter ExpressionNumericConverter = new CSharpExpressionNumericConverter();
-
         private static readonly DiagnosticDescriptor rule =
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
+        private ILanguageFacade<SyntaxKind> Language => CSharpFacade.Instance;
+
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
 
-        protected override void Initialize(SonarAnalysisContext context)
-        {
-            context.RegisterSyntaxNodeActionInNonGenerated(
-                c =>
-                {
-                    var binary = (BinaryExpressionSyntax) c.Node;
-                    CheckCountZero(binary.Right, binary.Left, c);
-                    CheckCountOne(binary.Left, binary.Right, c);
-                },
-                SyntaxKind.GreaterThanExpression);
+        protected override void Initialize(SonarAnalysisContext context) =>
             context.RegisterSyntaxNodeActionInNonGenerated(
                 c =>
                 {
                     var binary = (BinaryExpressionSyntax)c.Node;
-                    CheckCountZero(binary.Left, binary.Right, c);
-                    CheckCountOne(binary.Right, binary.Left, c);
-                },
-                SyntaxKind.LessThanExpression);
-            context.RegisterSyntaxNodeActionInNonGenerated(
-                c =>
-                {
-                    var binary = (BinaryExpressionSyntax)c.Node;
-                    CheckCountOne(binary.Right, binary.Left, c);
-                    CheckCountZero(binary.Left, binary.Right, c);
-                },
-                SyntaxKind.GreaterThanOrEqualExpression);
-            context.RegisterSyntaxNodeActionInNonGenerated(
-                c =>
-                {
-                    var binary = (BinaryExpressionSyntax)c.Node;
-                    CheckCountOne(binary.Left, binary.Right, c);
-                    CheckCountZero(binary.Right, binary.Left, c);
-                },
-                SyntaxKind.LessThanOrEqualExpression);
-            context.RegisterSyntaxNodeActionInNonGenerated(
-                c =>
-                {
-                    var binary = (BinaryExpressionSyntax)c.Node;
-                    CheckCountZero(binary.Left, binary.Right, c);
-                    CheckCountZero(binary.Right, binary.Left, c);
-                },
-                SyntaxKind.EqualsExpression);
-            context.RegisterSyntaxNodeActionInNonGenerated(
-                c =>
-                {
-                    var binary = (BinaryExpressionSyntax)c.Node;
-                    CheckCountZero(binary.Left, binary.Right, c);
-                    CheckCountZero(binary.Right, binary.Left, c);
-                },
-                SyntaxKind.NotEqualsExpression);
-        }
+                    var left = Language.ExpressionNumericConverter.TryGetConstantIntValue(binary.Left, out var l_out) ? (int?)l_out : null;
+                    var right = Language.ExpressionNumericConverter.TryGetConstantIntValue(binary.Right, out var r_out) ? (int?)r_out : null;
 
-        private static void CheckCountZero(ExpressionSyntax zero, ExpressionSyntax count, SyntaxNodeAnalysisContext context)
-        {
-            if (ExpressionNumericConverter.TryGetConstantIntValue(zero, out var value) &&
-                value == 0 &&
-                TryGetCountCall(count, context.SemanticModel, out var reportLocation, out var typeArgument))
-            {
-                context.ReportDiagnosticWhenActive(Diagnostic.Create(rule, reportLocation, typeArgument));
-            }
-        }
-        private static void CheckCountOne(ExpressionSyntax one, ExpressionSyntax count, SyntaxNodeAnalysisContext context)
-        {
-            if (ExpressionNumericConverter.TryGetConstantIntValue(one, out var value) &&
-                value == 1 &&
-                TryGetCountCall(count, context.SemanticModel, out var reportLocation, out var typeArgument))
-            {
-                context.ReportDiagnosticWhenActive(Diagnostic.Create(rule, reportLocation, typeArgument));
-            }
-        }
+                    if ((left ?? right) is int constant)
+                    {
+                        var comparision = left is null
+                            ? Language.Syntax.ComparisonKind(binary)
+                            : Language.Syntax.ComparisonKind(binary).Mirror();
+                        var expression = left is null ? binary.Left : binary.Right;
+
+                        if (comparision.Count(constant).AnyOrNotAny()
+                            && TryGetCountCall(expression, c.SemanticModel, out var location, out var typeArgument))
+                        {
+                            c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, location, typeArgument));
+                        }
+                    }
+                },
+                SyntaxKind.GreaterThanExpression,
+                SyntaxKind.GreaterThanOrEqualExpression,
+                SyntaxKind.LessThanExpression,
+                SyntaxKind.LessThanOrEqualExpression,
+                SyntaxKind.EqualsExpression,
+                SyntaxKind.NotEqualsExpression);
 
         private static bool TryGetCountCall(ExpressionSyntax expression, SemanticModel semanticModel, out Location countLocation, out string typeArgument)
         {
             countLocation = null;
             typeArgument = null;
             var invocation = expression as InvocationExpressionSyntax;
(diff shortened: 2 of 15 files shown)


Corniel:
Fixed issue #3735 too.

analyzers/src/SonarAnalyzer.CSharp/Helpers/CountComparisonSyntaxHelper.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace SonarAnalyzer.Helpers
```
pavel-mikula-sonarsource, line 6: We're moving extensions to `Extensions` directory and namespace to clean up the code base.

analyzers/src/SonarAnalyzer.CSharp/Helpers/CountComparisonSyntaxHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    public static class CountComparisonSyntaxHelper
```
pavel-mikula-sonarsource, line 8: To follow previous comment, this class and file should be
```suggestion
    public static class BinaryExpressionSyntaxExtensions
```

analyzers/src/SonarAnalyzer.CSharp/Helpers/CountComparisonSyntaxHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                left = l_out;
+            }
+            if (Converter.TryGetConstantIntValue(binary.Right, semanticModel, out var r_out))
+            {
+                right = r_out;
+            };
+            return new CountComparisonSyntax(left, binary.Operator(), right);
+        }
+        private static CountComparisonSyntax.Kind Operator(this CSharpSyntaxNode syntax) =>
```
pavel-mikula-sonarsource, line 26: _Nitpick:_ Missing EOL
```suggestion
        }
        
        private static CountComparisonSyntax.Kind Operator(this CSharpSyntaxNode syntax) =>
```

analyzers/src/SonarAnalyzer.Common/Helpers/CountComparisonSyntax.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            Equals,
+            NotEquals,
+            LessThan,
+            LessThanOrEqual,
+            GreaterThan,
+            GreaterThanOrEqual,
+        }
+
+        public CountComparisonSyntax(int? left, Kind @operator, int? right)
```
pavel-mikula-sonarsource, line 18: Please try to sort all of these members according to our https://github.com/SonarSource/sonar-dotnet/blob/master/docs/coding-style.md

and add empty lines between methods. Having block of properties without empty lines between them is fine.

analyzers/src/SonarAnalyzer.Common/Helpers/CountComparisonSyntax.cs
```diff
@@ -0,0 +1,94 @@
...
+            && Operator == other.Operator;
+        public override int GetHashCode() =>
+            Left.GetHashCode()
+            ^ (Right.GetHashCode() << 14)
+            ^ (Operator.GetHashCode() << 28);
+        public override string ToString() => $"{Left} {Operator} {Right}";
+
+        private static readonly CountComparisonSyntax[] Empties = new[]
+        {
+            new CountComparisonSyntax(default, Kind.Equals, 0),
```
pavel-mikula-sonarsource, line 74: Please use `null` instead of `default` for all nullable.
```suggestion
            new CountComparisonSyntax(null, Kind.Equals, 0),
```

analyzers/src/SonarAnalyzer.Common/Helpers/CountComparisonSyntax.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+        public int? Left { get; }
+        public int? Right { get; }
+        public Kind Operator { get; }
+
+        public bool ExpressionIsLeft() => !Left.HasValue;
+        public bool IsValid() => Left.HasValue ^ Right.HasValue;
+        public bool IsEmpty() => Empties.Contains(Normalized());
+        public bool IsAny() => Anys.Contains(Normalized());
+        public bool IsEmptyOrAny() => IsEmpty() || IsAny();
```
pavel-mikula-sonarsource, line 32: These are not trivial so they should be formatted like this (according to our latest coding style update)
```
public bool ExpressionIsLeft() => 
	!Left.HasValue;

```

analyzers/src/SonarAnalyzer.Common/Helpers/CountComparisonSyntax.cs
```diff
@@ -0,0 +1,94 @@
...
+        public override bool Equals(object obj) => obj is CountComparisonSyntax other && Equals(other);
+        public bool Equals(CountComparisonSyntax other) =>
+            Left == other.Left
+            && Right == other.Right
+            && Operator == other.Operator;
+        public override int GetHashCode() =>
+            Left.GetHashCode()
+            ^ (Right.GetHashCode() << 14)
+            ^ (Operator.GetHashCode() << 28);
+        public override string ToString() => $"{Left} {Operator} {Right}";
```
pavel-mikula-sonarsource, line 70: Let's make this a class and remove this overhead. I don't see a scenario where we would profit from special behavior for structs in this case.
```suggestion
```

analyzers/src/SonarAnalyzer.CSharp/Helpers/CountComparisonSyntaxHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+        private static readonly IExpressionNumericConverter Converter = new CSharpExpressionNumericConverter();
+
+        public static CountComparisonSyntax CountComparison(this BinaryExpressionSyntax binary, SemanticModel semanticModel)
+        {
+            int? left = default;
+            int? right = default;
+            if (Converter.TryGetConstantIntValue(binary.Left, semanticModel, out var l_out))
+            {
+                left = l_out;
+            }
```
pavel-mikula-sonarsource, line 19: It took me a while to actually understand this. Let's make it more direct and explicit
```suggestion
            int? left = Converter.TryGetConstantIntValue(binary.Left, semanticModel, out var leftValue) ? leftValue : null;
```

analyzers/src/SonarAnalyzer.CSharp/Helpers/CountComparisonSyntaxHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+            if (Converter.TryGetConstantIntValue(binary.Right, semanticModel, out var r_out))
+            {
+                right = r_out;
+            };
+            return new CountComparisonSyntax(left, binary.Operator(), right);
+        }
+        private static CountComparisonSyntax.Kind Operator(this CSharpSyntaxNode syntax) =>
+            syntax.Kind() switch
+            {
+                SyntaxKind.EqualsExpression => CountComparisonSyntax.Kind.Equals,
```
pavel-mikula-sonarsource, line 29: I'm thinking about how to make this shorter. Either by using static import or moving the `Kind` out of `CountCOmparisonSyntax` to be a `ComparisonKind`.

pavel-mikula-sonarsource (changes requested):
Hi @Corniel,

I didn't have enough time to review all yet, I just left bunch of beauty comments to make it easier for me to deal with the logic next time :)

analyzers/src/SonarAnalyzer.Common/Helpers/CountComparisonSyntax.cs
Corniel: This is key to its behaviour. With a class the behaviour will not work.

analyzers/src/SonarAnalyzer.CSharp/Helpers/CountComparisonSyntaxHelper.cs
Corniel: I was not sure either.  `ComparisonKind` might be worth trying.

analyzers/src/SonarAnalyzer.CSharp/Extensions/BinaryExpressionSyntaxExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                left = l_out;
+            }
+            if (Converter.TryGetConstantIntValue(binary.Right, semanticModel, out var r_out))
+            {
+                right = r_out;
+            };
+            return new CountComparisonSyntax(left, binary.Operator(), right);
+        }
+        private static CountComparisonSyntax.ComparisonKind Operator(this CSharpSyntaxNode syntax) =>
```
pavel-mikula-sonarsource, line 27: Adding EOL
```suggestion
        }
        
        private static CountComparisonSyntax.ComparisonKind Operator(this CSharpSyntaxNode syntax) =>
```

analyzers/src/SonarAnalyzer.VisualBasic/Extensions/BinaryExpressionSyntaxExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        private static readonly IExpressionNumericConverter Converter = new VisualBasicExpressionNumericConverter();
+
+        public static CountComparisonSyntax CountComparision(this BinaryExpressionSyntax binary, SemanticModel semanticModel)
+        {
+            int? left = Converter.TryGetConstantIntValue(binary.Left, semanticModel, out var l_out) ? l_out : null;
+            int? right = Converter.TryGetConstantIntValue(binary.Right, semanticModel, out var r_out) ? r_out : null;
+            return new CountComparisonSyntax(left, binary.Operator(), right);
+        }
+        private static CountComparisonSyntax.ComparisonKind Operator(this VisualBasicSyntaxNode syntax) =>
```
pavel-mikula-sonarsource, line 19: Add EOL
```suggestion
        }

        private static CountComparisonSyntax.ComparisonKind Operator(this VisualBasicSyntaxNode syntax) =>
```

analyzers/src/SonarAnalyzer.CSharp/Helpers/CountComparisonSyntaxHelper.cs
pavel-mikula-sonarsource: Looks better. The return type of the methods needs to be fixed, it doesn't compile now.

analyzers/src/SonarAnalyzer.Common/Helpers/CountComparisonSyntax.cs
```diff
@@ -0,0 +1,118 @@
...
+using System;
+using System.Linq;
+
+namespace SonarAnalyzer.Helpers
+{
+    public readonly struct CountComparisonSyntax : IEquatable<CountComparisonSyntax>
+    {
+        public int? Left { get; }
+        public int? Right { get; }
+        public ComparisonKind Operator { get; }
```
pavel-mikula-sonarsource, line 30: Normalization is key component of this class. And also a key pain. Let's change the inner state to normalized one. To store
* IsValid
* ComparisonKind Operator
* int Value

`int?` should not be needed any more, as we don't care when `IsValid = false`, right?

That should make things simple.

Corniel: @pavel-mikula-sonarsource I like the suggestion storing the result of the analysis. However, that is - I think - not enough.

If we take a look to the rule `DoNotCheckZeroSizeCollection` we can - and I think, should - extend this rule to all statements that always evaluate to to `true` or `false`. So `collection.Count < -1`, and `array.Length > -10` should also trigger this one. In that case, you need the inter value and the comparison operator to give a describtive message.

We might want to consider to do the normalization directly, and only store the constant to compare with, although it might turn out that for an additional fix provider it worth knowing, which part is where. I think that is a hard call.

The idea was to get the normalization part in the class/struct only, to localize the pain. When that has been achieved (and I think we did that now), we can focus on on reducing the pain too in this struct. What do you think?

pavel-mikula-sonarsource: We can store the normalized state to make this a class and remove all the code that makes it complicated. And we can also store the original state to be able to generate proper error messages.
That would be something like

// Normalized state (private)
* IsValid
* Operator (the normalized one)
* Value

// Original state (public)
* ExpressionLeft (or OriginalLeft)
* ExpressionRight (or OriginalRight)
* ExpressionOperator (or OriginalOperator)

My key idea is to achieve this:
```
public bool IsEmpty() =>
            IsMatch(ComparisonKind.Equals, 0)
            || IsMatch(ComparisonKind.LessThan, 1)
            || IsMatch(ComparisonKind.LessThanOrEqual, 0);
```
because it simple and straight forward. We don't need any additional equal-logic and persistent instances:
```
        private static readonly CountComparisonSyntax[] Empties = new[]
        {
            new CountComparisonSyntax(null, ComparisonKind.Equals, 0),
            new CountComparisonSyntax(null, ComparisonKind.LessThan, 1),
            new CountComparisonSyntax(null, ComparisonKind.LessThanOrEqual, 0),
        };
```

Corniel: I still think that you want to communicate something like this: "array.length < -1 always evaluates as false, and therfor without meaning.

And for the fix for the `Any()` and `!Any()` is it beneficial to use `CountComparisonSyntax`, and by doing so, not have to evaluate twice which part of the expression is the constant.

For me, the added value here, is to make the code that uses this, the easiest to read/reason about. And if that makes this a bit harder to read, I think that's the compromise.

pavel-mikula-sonarsource: I'm sorry but I'm lost with the latest commit. Where did the `CountComparisonSyntax` type go? It was a good type for holding the complicated logic on one isolated place.

Corniel: Well, the outcome of the evalutation is now a simple enum, al the complex logic is here: https://github.com/SonarSource/sonar-dotnet/pull/4033/files#diff-517470f94b2d684730153b4a3f4011df2483f680ed478bfc323cd3aa07ea1d99

analyzers/src/SonarAnalyzer.RuleDescriptorGenerator/SonarAnalyzer.RuleDescriptorGenerator.csproj
```diff
@@ -37,7 +37,7 @@
     checking the file extension.
    -->
   <Target Name="PostBuild" AfterTargets="PostBuildEvent" Condition="$(TargetFileName.EndsWith(&quot;exe&quot;))">
-    <Exec Command="cd $(OutDir)&#xD;&#xA;$(TargetFileName) cs&#xD;&#xA;$(TargetFileName) vbnet" />
+    <!--<Exec Command="cd $(OutDir)&#xD;&#xA;$(TargetFileName) cs&#xD;&#xA;$(TargetFileName) vbnet" />-->
```
Corniel, line 40: Reminder: remove this from the pull request.

Corniel:
New attempt in #4648

PR #4035, opened 2021-02-08
Fix S5944 FP: Attribute and named argument

pavel-mikula-sonarsource:
Related to #2159 

Attribute name issue was found on Peach.

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
@@ -1,10 +1,12 @@
-﻿Public Class ImplicitReturnStatementsAreNoncompliant
+﻿Imports System
+
+Public Class ImplicitReturnStatementsAreNoncompliant
     Implements IInterface
-    
+
     Public Function AssignedReturnValueOnly() As Integer
         AssignedReturnValueOnly = 42 ' Noncompliant ^9#23
     End Function
-    
+
     Public Function AssignementStatements() As Integer
         AssignementStatements = 101 ' Noncompliant {{Use a 'Return' statement; assigning returned values to function names is obsolete.}}
         AssignementStatements += 42 ' Noncompliant {{Use a 'Return' statement; assigning returned values to function names is obsolete.}}
@@ -17,7 +19,7 @@
         AssignementStatements <<= 2 ' Noncompliant {{Use a 'Return' statement; assigning returned values to function names is obsolete.}}
         AssignementStatements >>= 1 ' Noncompliant {{Use a 'Return' statement; assigning returned values to function names is obsolete.}}
     End Function
-    
+
     Public Function CaseInsensitive() As Integer
         CASEINSENSITIVE = 42  ' Noncompliant
     End Function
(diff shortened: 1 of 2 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
```diff
@@ -17,37 +19,44 @@
...
     End Function
-    
-     Public Function ImplementedInterfaceMethod() As Integer Implements IInterface.ImplementedInterfaceMethod
+
+    Public Function ImplementedInterfaceMethod() As Integer Implements IInterface.ImplementedInterfaceMethod
         ImplementedInterfaceMethod = 42 ' Noncompliant
     End Function
 
+    Private Function ArgumentName() As Integer
+        WithExplicitArgumentName(ArgumentName:=ArgumentName)    ' Noncompliant
```
andrei-epure-sonarsource, line 54: why is this Noncompliant? isn't this a named argument assignment rather than assigning a value to the function name?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
```diff
@@ -87,33 +96,50 @@ Public Class DoesNotApplyOn
...
     End Function
-    
+
+    <CustomAttribute>   ' Compliant
+    Public Function CustomAttribute() As Integer
+    End Function
+
+    Private Function ArgumentName() As Integer
+        WithExplicitArgumentName(ArgumentName:=42)  ' Compliant
+        WithExplicitArgumentName(ArgumentName:=42)
```
andrei-epure-sonarsource, line 116: line 54 and this are very hard to read because the type of the argument is string... and here you are passing an integer...

why is this compliant?

andrei-epure-sonarsource:
Also, as a side note, the syntax in lines 64-72 is messed up - it's not worth making another commit now just for this as this is a fix for the release, but please do a commit after to fix it (otherwise it will create noise in a future PR).

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
pavel-mikula-sonarsource: First `ArgumentName` is really argument name of the `WithExplicitArgumentName(ArgumentName As Integer)` method.
`Name:=Value` is equivalent to `name : value` explicit parameter naming when you need to pass them out of order.
Second `ArgumentName` is the noncompliant stuff. It's the hidden variable declared by enclosing function name.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
pavel-mikula-sonarsource: I fixed the `String` issue, that wasn't intended.

It's compliant, because it's parameter name like `param : value` C# syntax. It doesn't use the hidden variable.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
pavel-mikula-sonarsource: I made this more explicit by renaming the parameter.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
```diff
@@ -17,37 +19,44 @@
...
     End Function
-    
-     Public Function ImplementedInterfaceMethod() As Integer Implements IInterface.ImplementedInterfaceMethod
+
+    Public Function ImplementedInterfaceMethod() As Integer Implements IInterface.ImplementedInterfaceMethod
         ImplementedInterfaceMethod = 42 ' Noncompliant
     End Function
 
+    Private Function ArgumentName() As Integer
+        WithExplicitArgumentName(Something:=ArgumentName)    ' Noncompliant
```
andrei-epure-sonarsource, line 54: I don't understand the purpose of this test.
- no value is assigned to the function name
- the fact that it's an implicit return value should work with just a function call, right?

andrei-epure-sonarsource: like
```
Private Function X() As Integer
  FooThatReturnsInteger()
End Function
```

pavel-mikula-sonarsource: Missing `Return` statement is not important here. Important is the reading of `ArgumentName` implicit variable.

pavel-mikula-sonarsource: There's a difference between
```
WithExplicitArgumentName(Something:=ArgumentName)   ' Reading implicit variable that holds the return value
WithExplicitArgumentName(Something:=ArgumentName()) ' Recursive invocation
```

andrei-epure-sonarsource (approved):
LGTM

PR #4042, opened 2021-02-10
Update commonTemplates to v1.0.10

pavel-mikula-sonarsource:
Use new version of common templates:
https://sonarsource.visualstudio.com/DotNetTeam%20Project/_git/pipelines-yaml-templates/pullrequest/14

As this step renames steps, we need to:
* remove `Sonar.Net (Promote artifacts on repox and notify Burgr promoteBurgr)` required checks
* merge this
* add `Sonar.Net (Artifacts: Notify Burgr)` required checks

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,7 +21,7 @@ resources:
     - repository: commonTemplates
       type: git
       name: pipelines-yaml-templates
-      ref:  refs/tags/v1.0.6
+      ref:  refs/tags/v1.0.10
 
 stages:
 - template: stage-with-burgr-notifications.yml@commonTemplates


azure-pipelines.yml
```diff
@@ -21,7 +21,7 @@ resources:
     - repository: commonTemplates
       type: git
       name: pipelines-yaml-templates
-      ref:  refs/tags/v1.0.6
+      ref:  refs/tags/v1.0.10
```
andrei-epure-sonarsource, line 24: I don't see this tag here yet : https://sonarsource.visualstudio.com/DotNetTeam%20Project/_git/pipelines-yaml-templates/tags

andrei-epure-sonarsource (changes requested):

azure-pipelines.yml
pavel-mikula-sonarsource: Added now. It didn't propagated from the PR :/

andrei-epure-sonarsource (approved):
LGTM

PR #4043, opened 2021-02-10
Clean hotpot rules

csaba-sagi-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/Rules/BypassingAccessibilityTest.cs
```diff
@@ -33,30 +33,18 @@ public class BypassingAccessibilityTest
         [TestCategory("Rule")]
         [TestCategory("Hotspot")]
```
costin-zaharia-sonarsource, line 34: I think we can remove the attribute since this is not a hotspot anymore.

costin-zaharia-sonarsource (approved):
LGTM! Just a small comment regarding wrong Hotspot category in `BypassingAccessibility` tests

PR #4046, opened 2021-02-11
Make contact email consistent with other licence headers

MaartenGDev:
When checking the following commit: https://github.com/SonarSource/sonar-dotnet/commit/1af2e34548c0873aa437a11d02f9d51373ed7177,  that bumps the year to 2021 I spotted that the mailto was different in the `CSharpDiagnosticAnalyzerContextHelper.cs` file.

This commit makes it consistent with the files in the following directory: https://github.com/SonarSource/sonar-dotnet/tree/master/analyzers/src/SonarAnalyzer.CSharp/Rules


pavel-mikula-sonarsource:
Hi @MaartenGDev, thank you for noticing this. 

It actually revealed root cause of some other problems. First is that rule S1451 should trigger to detect this. Second is that the file itself is not present in SonarCloud for some reason.

I'll close this PR for now as it would hide the evidence and there might be more files affected by the same issue.

PR #4050, opened 2021-02-11
Rule S2612: [C#] Setting loose file permissions is security-sensitive

### LINKED ISSUE #3979: Rule S2612: Setting loose file permissions is security-sensitive
https://jira.sonarsource.com/browse/MMF-2159
https://jira.sonarsource.com/browse/RSPEC-2612
https://github.com/SonarSource/security-expected-issues/pull/239

costin-zaharia-sonarsource:
Fixes #3979


analyzers/its/config/NetCore31WithConfigurableRules/SonarLint.xml
```diff
@@ -22,6 +23,9 @@
     <Rule><Key>S4834</Key></Rule>
     <Rule><Key>S5042</Key></Rule>
     <Rule><Key>S5332</Key></Rule>
+    <Rule><Key>S5693</Key></Rule>
```
csaba-sagi-sonarsource, line 26: These three are redundant, they are already enlisted in the file.

analyzers/rspec/cs/Sonar_way_recommended_profile.json
```diff
@@ -0,0 +1,6 @@
+{
```
csaba-sagi-sonarsource, line 1: If I am correct than this file is not needed.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/LooseFilePermissions.cs
```diff
@@ -0,0 +1,152 @@
...
+            "DefaultPermissions",
+            "OtherExecute",
+            "OtherWrite",
+            "OtherRead",
+            "OtherReadWriteExecute"
+        };
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager).WithNotConfigurable();
```
csaba-sagi-sonarsource, line 54: Styling: Order of members inside the class is not correct.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/LooseFilePermissions.cs
```diff
@@ -0,0 +1,152 @@
...
+
+        public LooseFilePermissions() : this(AnalyzerConfiguration.Hotspot) { }
+
+        internal LooseFilePermissions(IAnalyzerConfiguration configuration) : base(configuration)
```
csaba-sagi-sonarsource, line 58: Styling: This is my preference so no need to tackle if you do not agree. The opening and closing parenthesis can be moved to line 58 as nothing happens in the constructor.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/LooseFilePermissions.cs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        private static bool IsFileAccessPermissions(SimpleNameSyntax identifierNameSyntax, SemanticModel semanticModel) =>
+            WeakFileAccessPermissions.Contains(identifierNameSyntax.Identifier.Text)
+            && identifierNameSyntax.IsKnownType(KnownType.Mono_Unix_FileAccessPermissions, semanticModel);
+
+        private static void VisitInvocations(SyntaxNodeAnalysisContext context)
+        {
+            var invocation = (InvocationExpressionSyntax)context.Node;
+            if (!invocation.IsMemberAccessOnKnownType("SetAccessRule", KnownType.System_Security_AccessControl_FileSystemSecurity, context.SemanticModel) &&
```
csaba-sagi-sonarsource, line 91: Styling: && should be in the next line.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/LooseFilePermissions.cs
```diff
@@ -0,0 +1,152 @@
...
+                !invocation.IsMemberAccessOnKnownType("AddAccessRule", KnownType.System_Security_AccessControl_FileSystemSecurity, context.SemanticModel))
+            {
+                return;
+            }
+
+            if (GetObjectCreation(invocation, context.SemanticModel) is { })
```
csaba-sagi-sonarsource, line 97: Maybe the the two ifs could be merged if( (SetAccessRule || AddAccessRule) && GetObjectCreation)

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/LooseFilePermissions.cs
```diff
@@ -0,0 +1,152 @@
...
+                     .FirstOrDefault(objectCreation => IsFileSystemAccessRuleForEveryoneWithAllow(objectCreation, semanticModel));
+        }
+
+        private static bool IsFileSystemAccessRuleForEveryoneWithAllow(ObjectCreationExpressionSyntax objectCreation, SemanticModel semanticModel) =>
+            objectCreation.IsKnownType(KnownType.System_Security_AccessControl_FileSystemAccessRule, semanticModel)
+            && objectCreation.ArgumentList is { } argumentList
+            && IsEveryone(argumentList.Arguments.First().Expression, semanticModel)
+            && semanticModel.GetConstantValue(argumentList.Arguments.Last().Expression) is {HasValue: true, Value: 0};
+
+        private static bool IsEveryone(SyntaxNode syntaxNode, SemanticModel semanticModel)
```
csaba-sagi-sonarsource, line 130: Styling: If could be converted to conditional expression and expression bodied method

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/WeakSslTlsProtocols.cs
```diff
@@ -122,8 +122,7 @@ public void SslProtocolsNonComplaint()
 
             var protocol = SslProtocols.None;
 
-            // FP - the following cases should be valid
-            bool isSafe = protocol == SslProtocols.Tls ? true : false; // Noncompliant
+            bool isSafe = protocol == SslProtocols.Tls ? true : false;
```
csaba-sagi-sonarsource, line 125: Maybe this could be moved to SslProtocolsCompliant method. And as we now do not raise for whilestatements. We should add a case for that too.

csaba-sagi-sonarsource (changes requested):
Overall LGTM, just a few minor changes needed.

analyzers/rspec/cs/Sonar_way_recommended_profile.json
costin-zaharia-sonarsource: The reason why this file was generated is that the rule is in `Sonar way recommended` profile. I wrote though to @pierre-loup-tristant-sonarsource to confirm that this is intentional.

csaba-sagi-sonarsource: There are a bunch of rules that are part of that profile but this was the first time the file was added. We either do not use that file and do not have to create it, or a lot of rules are missing from them. If the second is the case than we should create a task to add all the other rules that are part of that profile too.

costin-zaharia-sonarsource: I've discussed it in the end with @saberduck and it seems that this was a mistake. The change was needed only for JS. He fixed the RSPEC and I will regenerate the files.

csaba-sagi-sonarsource (approved):
LGTM

PR #4051, opened 2021-02-12
Extend NugetMetadataFactory to support specifying a runtime

costin-zaharia-sonarsource:
- this is needed in order to support dependencies like [Mono.Posix](https://www.nuget.org/packages/Mono.Posix.NETStandard)


csaba-sagi-sonarsource (approved):
LGTM

PR #4057, opened 2021-02-12
Don't count lambdas in too complex CFG expression count

### LINKED ISSUE #4056: Don't count lambdas in too complex CFG expression count
As seen in sonar-security ITs, content of lambdas are counted inside `CSharpControlFlowGraphBuilder.IsTooComplex` do determine expression complexity.

As the layout page content in Razor XSS is rendered inside async lambda, the main CFG for the layout page is not build, because it considers the lambda argument contains too complex expression.

As a fun fact, the Sonar-Security IT for NetCore2 fails, because it contains bunch of `BeginContext` and `EndContext` that were removed from .NET Core 3 Razor build system. So it doesn't fail on NetCore3 by pure coincidence.

pavel-mikula-sonarsource:
Fixes #4056

### DIFF
--- a/analyzers/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
+++ b/analyzers/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
@@ -500,7 +500,11 @@ namespace SonarAnalyzer.ControlFlowGraph.CSharp
         private static bool IsTooComplex(SyntaxNode node)
         {
             var count = 0;  // Limit descending for performance reasons
-            return node.DescendantNodes(x => ++count < SupportedExpressionNodeCountLimit).Count() >= SupportedExpressionNodeCountLimit;
+            // The evaluation doesn't descend into the content of lambda expression. We need to tolerate these for Razor Layout pages.
+            return node.DescendantNodes(x => ++count < SupportedExpressionNodeCountLimit && !IsLambda(x)).Count() >= SupportedExpressionNodeCountLimit;
+
+            static bool IsLambda(SyntaxNode node) =>
+                 node is LambdaExpressionSyntax;
         }
 
         #endregion Top level Build*
--- a/analyzers/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
@@ -183,7 +183,7 @@ namespace NS
             Action a = () => CSharpControlFlowGraph.Create(equalsValueSyntax.Value, semanticModel);
 
             a.Should().Throw<NotSupportedException>().WithMessage("Too complex expression");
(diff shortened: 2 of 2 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -183,7 +183,7 @@ public void Cfg_ExtremelyNestedExpression_NotSupported_FromExpression()
             Action a = () => CSharpControlFlowGraph.Create(equalsValueSyntax.Value, semanticModel);
 
             a.Should().Throw<NotSupportedException>().WithMessage("Too complex expression");
-            CSharpControlFlowGraph.TryGet(method.Body, semanticModel, out _).Should().BeFalse();
+            CSharpControlFlowGraph.TryGet(equalsValueSyntax.Value, semanticModel, out _).Should().BeFalse();
```
pavel-mikula-sonarsource, line 186: Fixing old bug, `equalsValueSyntax.Value` was intended here, just as in the `Create(..)` above.

analyzers/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
```diff
@@ -500,7 +500,11 @@ private Block BuildExpression(ExpressionSyntax expression, Block currentBlock)
         private static bool IsTooComplex(SyntaxNode node)
         {
             var count = 0;  // Limit descending for performance reasons
-            return node.DescendantNodes(x => ++count < SupportedExpressionNodeCountLimit).Count() >= SupportedExpressionNodeCountLimit;
+            // The evaluation doesn't descend into the content of lambda expression. We need to tolerate these for Razor Layout pages.
```
andrei-epure-sonarsource, line 503: >The evaluation doesn't descend into the content of lambda expression

Where this logic happens? Is it in this builder as well, or a different class? I'm just thinking ahead if this may change in the future, the evaluation may descend into lambas, and not to forget about this IsTooComplex.

I suggest to add there a comment that `IsTooComplex` should be revisited if we decide to descent into lambda expressions in the future.

andrei-epure-sonarsource (approved):
LGTM, minor suggestion to add a comment (if it makes sense)

analyzers/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs
andrei-epure-sonarsource: Or is the comment just an explanation for the logic below?

pavel-mikula-sonarsource: It's related to the fact, that we don't support lambdas and we don't evaluate their inner behavior. 
This `IsTooComplex` problem should go away by (re)design to Roslyn CFG so we don't need to put too much attention to it.

PR #4064, opened 2021-02-15
Log warning when there are only tests in the SQ/SC project

### LINKED ISSUE #4052: Warn users when no MAIN projects is found
Inside `DotNetSensor`:
Raise a warning there are 0 `MAIN` files and at least one `TEST` files.

andrei-epure-sonarsource:
Fixes #4052

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
@@ -72,20 +72,18 @@ public class DotNetSensor implements ProjectSensor {
   }
 
   private boolean shouldExecuteOnProject(FileSystem fs) {
-    boolean hasMainFiles = hasFilesOfType(fs, Type.MAIN);
-    boolean hasTestFiles = hasFilesOfType(fs, Type.TEST);
-
-    if (!hasMainFiles) {
-      if (hasTestFiles) {
-        String noResultsMessage = "Your SonarQube/SonarCloud project will not have results for " + pluginMetadata.languageKey() + " files.";
-        LOG.warn("This solution contains only TEST files. Because there are no MAIN files, this sensor will be skipped. " + noResultsMessage);
-      } else {
-        LOG.debug("No files to analyze. Skip Sensor.");
-      }
-      return false;
+    if (hasFilesOfType(fs, Type.MAIN)) {
+      return true;
+    } else if (hasFilesOfType(fs, Type.TEST)) {
+      LOG.warn("This sensor will be skipped, because the current solution contains only TEST files and no MAIN files. " +
+          "Your SonarQube/SonarCloud project will not have results for {} files. " +
+          "You can read more about the detection of test projects here: https://github.com/SonarSource/sonar-scanner-msbuild/wiki/Analysis-of-product-projects-vs.-test-projects",
+        pluginMetadata.languageKey());
+    } else {
+      // it's not a .NET project
+      LOG.debug("No files to analyze. Skip Sensor.");
     }
-
-    return true;
+    return false;
   }
 
   private boolean hasFilesOfType(FileSystem fs, Type fileType) {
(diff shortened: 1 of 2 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
```diff
@@ -73,16 +72,24 @@ public void execute(SensorContext context) {
...
-      LOG.debug("No files to analyze. Skip Sensor.");
+    boolean hasMainFiles = hasFilesOfType(fs, Type.MAIN);
+    boolean hasTestFiles = hasFilesOfType(fs, Type.TEST);
+
+    if (!hasMainFiles) {
+      if (hasTestFiles) {
+        String noResultsMessage = "Your SonarQube/SonarCloud project will not have results for " + pluginMetadata.languageKey() + " files.";
+        LOG.warn("This solution contains only TEST files. Because there are no MAIN files, this sensor will be skipped. " + noResultsMessage);
+      } else {
+        LOG.debug("No files to analyze. Skip Sensor.");
```
andrei-epure-sonarsource, line 83: given that this is DEBUG, I think I can add `Your SonarQube/SonarCloud project will not have results for LANG files.` here too, right?

Verbose logs should make things obvious.

pavel-mikula-sonarsource: I think it's good as it is. The "No files to analyze" is clear enough. Given it's a DEBUG.

I would not scare or mislead users using S4MSB for SQL and other projects.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
```diff
@@ -73,16 +72,24 @@ public void execute(SensorContext context) {
...
+      }
```
pavel-mikula-sonarsource, line 84: I'd do two changes:
- Make conditions positive.
- I think that 'hasFilesOfType' is expensive under the hood. So call it only when needed.
```suggestion
    if (hasFilesOfType(fs, Type.MAIN)) {
      return true;
    } else if (hasFilesOfType(fs, Type.TEST)) {
        String noResultsMessage = "Your SonarQube/SonarCloud project will not have results for " + pluginMetadata.languageKey() + " files.";
        LOG.warn("This solution contains only TEST files. Because there are no MAIN files, this sensor will be skipped. " + noResultsMessage);
    } else {
      LOG.debug("No files to analyze. Skip Sensor.");
    }
    return false;
```

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/DotNetSensorTest.java
```diff
@@ -164,10 +155,64 @@ public void whenReportsArePresentThereAreNoWarnings() {
...
+
+  @Test
+  public void whenThereAreOnlyTestFiles_logWarning() {
+    addTestFileToFs();
+
+    sensor.execute(tester);
+
+    assertThat(logTester.logs(LoggerLevel.DEBUG)).isEmpty();
+    assertThat(logTester.logs(LoggerLevel.WARN))
+      .containsExactly("This solution contains only TEST files. Because there are no MAIN files, this sensor will be skipped. Your SonarQube/SonarCloud project will not have results for LANG_KEY files.");
```
pavel-mikula-sonarsource, line 178: _To consider:_ I'd reorder the given information. First the main action (skip), followed by a reasoning (main/test). And the SQ/SC consequence as is.
```suggestion
      .containsExactly("This sensor will be skipped, because current solution contains only TEST files and no MAIN files. Your SonarQube/SonarCloud project will not have results for LANG_KEY files.");
```

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/DotNetSensorTest.java
```diff
@@ -164,10 +155,64 @@ public void whenReportsArePresentThereAreNoWarnings() {
...
+
+  private void addTestFileToFs() {
+    DefaultInputFile testFile = new TestInputFileBuilder("mod", "SomeTest.language")
+      .setLanguage(LANG_KEY)
+      .setType(Type.TEST)
+      .build();
+    addFileToFs(testFile);
+  }
+
+  private void addFileToFs(InputFile inputFile) {
```
pavel-mikula-sonarsource, line 206: By adding type and name arguments here, we can remove the duplicated logic from the two methods above. It takes 5 steps to build the file.

pavel-mikula-sonarsource (changes requested):
Minor improvements suggested

pavel-mikula-sonarsource (approved):
LGTM

PR #4065, opened 2021-02-15
Rule S2612 [VB.Net]: Setting loose file permissions is security-sensitive

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.VisualBasic/Rules/LooseFilePermissions.cs
```diff
@@ -0,0 +1,125 @@
...
+                var diagnostic = invocationLocation.GetLineSpan().StartLinePosition.Line == secondaryLocation.GetLineSpan().StartLinePosition.Line
+                    ? Diagnostic.Create(Rule, invocationLocation)
+                    : Diagnostic.Create(Rule, invocationLocation, additionalLocations: new[] {secondaryLocation});
+
+                context.ReportDiagnosticWhenActive(diagnostic);
+            }
+        }
+
+        private static bool IsFileAccessPermissions(SimpleNameSyntax identifierNameSyntax, SemanticModel semanticModel) =>
+            LooseFilePermissionsConfig.WeakFileAccessPermissions.Contains(identifierNameSyntax.Identifier.Text)
```
csaba-sagi-sonarsource, line 73: Maybe identifierNameSyntax.Identifier could go into the SyntaxFacade class and if that is possible than the whole method could go to the base class.

analyzers/src/SonarAnalyzer.VisualBasic/Rules/LooseFilePermissions.cs
```diff
@@ -0,0 +1,125 @@
...
+{
+    [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
+    [Rule(DiagnosticId)]
+    public sealed class LooseFilePermissions : LooseFilePermissionsBase<SyntaxKind>
+    {
+        protected override ILanguageFacade<SyntaxKind> Language => VisualBasicFacade.Instance;
+
+        public LooseFilePermissions() : this(AnalyzerConfiguration.Hotspot) { }
+
+        internal LooseFilePermissions(IAnalyzerConfiguration configuration) : base(configuration, RspecStrings.ResourceManager)
```
csaba-sagi-sonarsource, line 42: styling: can be single line with { }

csaba-sagi-sonarsource (approved):
LGTM, if you want you can move that method to the base class.

analyzers/src/SonarAnalyzer.VisualBasic/Rules/LooseFilePermissions.cs
costin-zaharia-sonarsource: I'm not sure about this. In order to move `identifierNameSyntax.Identifier` there, we would need to use the base class `SyntaxNode` and do a cast in derived types. With this we will lose the type safety check which is now done at the compile time.

costin-zaharia-sonarsource: Thanks for the tip with `Language.Syntax.NodeIdentifier(syntaxNode)`. Looks better now 👍

PR #4066, opened 2021-02-15
Rule: S5443 Using publicly writable directories is security-sensitive

### LINKED ISSUE #3999: Rule: S5443 Using publicly writable directories is security-sensitive
https://jira.sonarsource.com/browse/MMF-2159
https://jira.sonarsource.com/browse/RSPEC-5443
https://github.com/SonarSource/security-expected-issues/pull/256

csaba-sagi-sonarsource:
Fixes #3999


analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,132 @@
...
+        private const string MessageFormat = "Make sure publicly writable directories are used safely here.";
+        private const string TwoPartPatternWithVariableSlash = @"^[\/\\]{0}[\/\\]{1}([\/\\]|$)";
+        private const string TwoPartPatternWithForwardSlash = @"^\/{0}\/{1}(\/|$)";
+        private const string ThreePartPatternWithForwardSlash = @"^\/{0}\/{1}\/{2}(\/|$)";
+        private const string TmpRegex = @"^([a-zA-z]:)?[\\\/]?[\/\\]tmp([\/\\]|$)";
+        private const string TempRegex = @"^([a-zA-z]:)?[\\\/]?[\/\\]temp([\/\\]|$)";
+        private const string WindowsTempRegex = @"^([a-zA-z]:)?[\\\/]?[\\\/]windows[\\\/]temp([\\\/]|$)";
+        private const string UserProfileRegex = @"%USERPROFILE%\\AppData\\Local\\Temp([\\\/]|$)";
+        private const string EnvironmentVariableTemplate = @"%{0}%([\\\/]|$)";
+        private static readonly string VarTmpRegex = string.Format(TwoPartPatternWithVariableSlash, "var", "tmp");
```
costin-zaharia-sonarsource, line 43: Since we get all these warnings for static fields on generic types we can extract a class responsible for finding the matches and use it from here.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,132 @@
...
+                RegexOptions.Compiled | RegexOptions.IgnoreCase | RegexOptions.CultureInvariant);
+
+        protected abstract ILanguageFacade<TSyntaxKind> Language { get; }
+
+        private protected abstract bool IsGetTempPathAssignment(TInvocationExpression invocationExpression, KnownType type, string methodName, SemanticModel semanticModel);
+        private protected abstract bool IsInsecureEnvironmentVariableRetrieval(TInvocationExpression invocation, KnownType type, string methodName, SemanticModel semanticModel);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        protected DiagnosticDescriptor Rule { get; }
```
costin-zaharia-sonarsource, line 76: can be private

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        protected static readonly string[] InsecureEnvironmentVariables =
+            new[]
+                {
+                    "tmp",
+                    "temp",
+                    "tmpdir"
+                };
```
costin-zaharia-sonarsource, line 84: Can be simplified a bit: `protected static readonly string[] InsecureEnvironmentVariables = {"tmp", "temp", "tmpdir"};`

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        protected PubliclyWritableDirectoriesBase(IAnalyzerConfiguration configuration, System.Resources.ResourceManager rspecResources) : base(configuration) =>
```
costin-zaharia-sonarsource, line 86: we can add a using `using System.Resources;` to simplify this

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/PubliclyWritableDirectories.cs
```diff
@@ -0,0 +1,52 @@
...
+    public sealed class PubliclyWritableDirectories : PubliclyWritableDirectoriesBase<SyntaxKind, InvocationExpressionSyntax>
+    {
+        protected override ILanguageFacade<SyntaxKind> Language => VisualBasicFacade.Instance;
+
+        public PubliclyWritableDirectories() : this(AnalyzerConfiguration.Hotspot) { }
+
+        internal PubliclyWritableDirectories(IAnalyzerConfiguration configuration) : base(configuration, RspecStrings.ResourceManager) { }
+
+        private protected override bool IsGetTempPathAssignment(InvocationExpressionSyntax invocationExpression, KnownType type, string methodName, SemanticModel semanticModel) =>
+            VisualBasicSyntaxHelper.IsMethodInvocation(invocationExpression, type, methodName, semanticModel)
```
costin-zaharia-sonarsource, line 43:
```suggestion
            invocationExpression.IsMethodInvocation(type, methodName, semanticModel)
```

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/PubliclyWritableDirectories.cs
```diff
@@ -0,0 +1,52 @@
...
+            && (invocationExpression.Parent is EqualsValueSyntax
+                || invocationExpression.Parent is AssignmentStatementSyntax);
+
+        private protected override bool IsInsecureEnvironmentVariableRetrieval(InvocationExpressionSyntax invocation, KnownType type, string methodName, SemanticModel semanticModel) =>
+            VisualBasicSyntaxHelper.IsMethodInvocation(invocation, type, methodName, semanticModel)
```
costin-zaharia-sonarsource, line 48:
```suggestion
            invocation.IsMethodInvocation(type, methodName, semanticModel)
```

analyzers/src/SonarAnalyzer.VisualBasic/Extensions/InterpolatedStringExpressionSyntaxExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using Microsoft.CodeAnalysis.VisualBasic.Syntax;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Extensions
+{
+    public static class InterpolatedStringExpressionSyntaxExtensions
+    {
+        public static string GetContentsText(this InterpolatedStringExpressionSyntax interpolatedStringExpression) =>
+            interpolatedStringExpression.Contents.JoinStr("", content => content.ToString());
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I'd prefer `null`, as we don't really need any separator.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
pavel-mikula-sonarsource: Or just remove `static` keyword, as we don't need static nature of the regex.

analyzers/rspec/cs/S5443_c#.html
```diff
@@ -0,0 +1,71 @@
...
+<p> * <code>/run/lock</code></p>
+<p> * <code>/var/run/lock</code></p>
+<p> * <code>/Library/Caches</code></p>
+<p> * <code>/Users/Shared</code></p>
+<p> * <code>/private/tmp</code></p>
+<p> * <code>/private/var/tmp</code></p>
+<p> * <code>\Windows\Temp</code></p>
+<p> * <code>\Temp</code></p>
+<p> * <code>\TMP</code></p>
+<p> * <code>%USERPROFILE%\AppData\Local\Temp</code></p>
```
pavel-mikula-sonarsource, line 25: This doesn't seem right, I'd expect `<ul>` `<li>` here.

analyzers/rspec/cs/S5443_c#.html
```diff
@@ -0,0 +1,71 @@
+<p>Operating systems have global directories where any user has write access. Those folders are mostly used as temporary storage areas like
+<code>/tmp</code> in Linux based systems. An application manipulating files from these folders is exposed to race conditions on filenames: a malicious
+user can try to create a file with a predictable name before the application does. A successful attack can result in other files being accessed,
+modified, corrupted or deleted. This risk is even higher if the application runs with elevated permissions.</p>
+<p>In the past, it has led to the following vulnerabilities:</p>
+<p> * <a href="https://nvd.nist.gov/vuln/detail/CVE-2012-2451">CVE-2012-2451</a></p>
+<p> * <a href="https://nvd.nist.gov/vuln/detail/CVE-2015-1838">CVE-2015-1838</a></p>
```
pavel-mikula-sonarsource, line 7: Spacing issue in Jira, there should be `<ul>`/`<li>` around here

analyzers/rspec/cs/S5443_c#.html
```diff
@@ -0,0 +1,71 @@
...
+</ul>
+<p>There is a risk if you answered yes to any of those questions.</p>
+<h2>Recommended Secure Coding Practices</h2>
+<p>Out of the box, .NET is missing secure-by-design APIs to create temporary files. To overcome this, one of the following options can be used:</p>
+<p> * Use a dedicated sub-folder with tightly controlled permissions</p>
+<p> * Created temporary files in a publicly writable folder and make sure:</p>
+<p> <strong></strong> Generated filename is unpredictable</p>
+<p> <strong></strong> File is readable and writable only by the creating user ID</p>
+<p> <strong></strong> File descriptor is not inherited by child processes</p>
+<p> <strong></strong> File is destroyed as soon as it is closed</p>
```
pavel-mikula-sonarsource, line 39: Spacing issue in Jira, there should be `<ul>`/`<li>` around here

pavel-mikula-sonarsource, line 39: The `<strong></strong>` doesn't look intentional on these lines (looks like a Jira issue)

analyzers/rspec/cs/S5443_c#.html
```diff
@@ -0,0 +1,71 @@
...
+<h2>Sensitive Code Example</h2>
+<pre>
+using (var writer = new StreamWriter("/tmp/f")) // Sensitive
+{
+// ...
+}
```
pavel-mikula-sonarsource, line 45: _Optional:_ New syntax to reduce parenthesis hell in the demonstration?
```suggestion
using var writer = new StreamWriter("/tmp/f"); // Sensitive
```

analyzers/rspec/vbnet/S5443_vb.net.html
```diff
@@ -0,0 +1,70 @@
+<p>Operating systems have global directories where any user has write access. Those folders are mostly used as temporary storage areas like
```
pavel-mikula-sonarsource, line 1: _Note for myself:_ Check formatting after Jira update.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/PubliclyWritableDirectories.cs
```diff
@@ -0,0 +1,100 @@
...
+            using (var tmpDir = new StreamReader(@"/tmp/f")) { };  // Noncompliant
+            using (var tmpDir = new StreamReader("D:\\Windows\\Temp\\f")) { }; // Noncompliant
+            using (var tmpDir = new StreamReader("\\\\Windows\\Temp\\f")) { }; // Noncompliant
+            using (var tmpDir = new StreamReader("\\Windows\\Temp\\f")) { }; // Noncompliant
+            using (var tmpDir = new StreamReader(@"C:\Windows\Temp\f")) { }; // Noncompliant
+            using (var tmpDir = new StreamReader(@"D:\Windows\Temp\f")) { }; // Noncompliant
+            using (var tmpDir = new StreamReader(@"\\Windows\Temp\f")) { }; // Noncompliant
+            using (var tmpDir = new StreamReader(@"\Windows\Temp\f")) { }; // Noncompliant
+            tmp = "/tmp/f"; // Noncompliant
+            tmp = "/tEmP/f"; // Noncompliant
```
pavel-mikula-sonarsource, line 80: Why is this specific case noncompliant? Linux file system is case sensitive so `/tEmP/` is not `/temp/`

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/PubliclyWritableDirectories.cs
```diff
@@ -0,0 +1,100 @@
+﻿using System;
+using System.IO;
+
+namespace Tests.Diagnostics
+{
+    public class Program
+    {
+        public void Compliant()
+        {
+            var randomPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName()); // Compliant
```
pavel-mikula-sonarsource, line 10: Let's also document `Path.GetTempFileName()`

pavel-mikula-sonarsource: Well, is `Path.GetTempFileName()` compliant or sensitive? It directly creates the file for you, so it doesn't have the autodelete component in it.
Unless you open it like that afterwards.

analyzers/src/SonarAnalyzer.Common/Facade/SyntaxFacade.cs
```diff
@@ -34,6 +34,7 @@ public abstract class SyntaxFacade<TSyntaxKind>
         public abstract SyntaxToken? InvocationIdentifier(SyntaxNode invocation);
         public abstract SyntaxNode NodeExpression(SyntaxNode node);
         public abstract SyntaxToken? NodeIdentifier(SyntaxNode node);
+        public abstract string GetStringTextValue(SyntaxNode node);
```
pavel-mikula-sonarsource, line 37: As it expects any Node as input (contrary to InvocatonIdentifier that expect Invocation as an input)
```suggestion
        public abstract string NodeStringTextValue(SyntaxNode node);
```

analyzers/src/SonarAnalyzer.VisualBasic/Extensions/InterpolatedStringExpressionSyntaxExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using Microsoft.CodeAnalysis.VisualBasic.Syntax;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Extensions
+{
+    public static class InterpolatedStringExpressionSyntaxExtensions
+    {
+        public static string GetContentsText(this InterpolatedStringExpressionSyntax interpolatedStringExpression) =>
+            interpolatedStringExpression.Contents.JoinStr(string.Empty, content => content.ToString());
```
pavel-mikula-sonarsource, line 29: As we don't need any separator there
```suggestion
            interpolatedStringExpression.Contents.JoinStr(null, content => content.ToString());
```

pavel-mikula-sonarsource (changes requested):
First round, I'll check the implementation itself later.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/PubliclyWritableDirectories.cs
csaba-sagi-sonarsource: From what I understood, all the cases after the //variates comment are not linux only so they should have a case insensitive check.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/PubliclyWritableDirectories.cs
csaba-sagi-sonarsource: I would say it is compliant as this rule targets directories, and [S5445](https://jira.sonarsource.com/browse/RSPEC-5445) targets insecure temporary files.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,133 @@
...
+                        && (IsGetTempPathAssignment(invocation, KnownType.System_IO_Path, "GetTempPath", c.SemanticModel)
+                            || IsInsecureEnvironmentVariableRetrieval(invocation, KnownType.System_Environment, "GetEnvironmentVariable", c.SemanticModel)))
+                    {
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, node.GetLocation()));
+                    }
+                },
+                Language.SyntaxKind.InvocationExpression);
+        }
+
+        private static class PubliclyWritableDirectoriesRegexMatcher
```
pavel-mikula-sonarsource, line 91: As we're inside `PubliclyWritableDirectoriesBase` scope
```suggestion
        private static class RegexMatcher
```

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PubliclyWritableDirectories.cs
```diff
@@ -0,0 +1,54 @@
...
+        public PubliclyWritableDirectories() : this(AnalyzerConfiguration.Hotspot) { }
+
+        internal PubliclyWritableDirectories(IAnalyzerConfiguration configuration) : base(configuration, RspecStrings.ResourceManager) { }
+
+        private protected override bool IsGetTempPathAssignment(InvocationExpressionSyntax invocationExpression, KnownType type, string methodName, SemanticModel semanticModel) =>
+            invocationExpression.IsMethodInvocation(type, methodName, semanticModel)
+            && (invocationExpression.Parent is EqualsValueClauseSyntax
+                || invocationExpression.Parent is AssignmentExpressionSyntax
+                || invocationExpression.Parent is ArrowExpressionClauseSyntax
+                || invocationExpression.Parent is ReturnStatementSyntax);
```
pavel-mikula-sonarsource, line 47: This can be replaced with `invocationExpression.Parent.IsAnyKind(...)` if I'm not missing fact that some of those classes are basetype for many syntax kinds.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PubliclyWritableDirectories.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        private protected override bool IsInsecureEnvironmentVariableRetrieval(InvocationExpressionSyntax invocation, KnownType type, string methodName, SemanticModel semanticModel) =>
+            invocation.IsMethodInvocation(type, methodName, semanticModel)
+            && invocation.ArgumentList?.Arguments.FirstOrDefault() is { } firstArgument
+            && InsecureEnvironmentVariables.Any(x => x.Equals(firstArgument.Expression?.GetStringValue(), StringComparison.OrdinalIgnoreCase));
```
pavel-mikula-sonarsource, line 52: It's a hotspot and most of the hotspots uses `.FindStringConstant` to also detect value propagation. Should we use that instead here as well?

We can handle the interpolated string separately.

analyzers/src/SonarAnalyzer.CSharp/Extensions/InterpolatedStringExpressionSyntaxExtensions.cs
```diff
@@ -26,6 +26,6 @@ namespace SonarAnalyzer.Extensions
     public static class InterpolatedStringExpressionSyntaxExtensions
     {
         public static string GetContentsText(this InterpolatedStringExpressionSyntax interpolatedStringExpression) =>
-            interpolatedStringExpression.Contents.JoinStr("", content => content.ToString());
+            interpolatedStringExpression.Contents.JoinStr(null, content => content.ToString());
```
pavel-mikula-sonarsource, line 29: I'm not very happy with this being a common extension as the empty separator is questionable. For example in hardcoded credentials rule, we needed to replace it with something nonempty. 

To avoid this problem, can we remove the interpolated string support from this rule and create an issue to deal with it in universal way for `FindConstant` extension for all other rules?

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/PubliclyWritableDirectories.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public PubliclyWritableDirectories() : this(AnalyzerConfiguration.Hotspot) { }
+
+        internal PubliclyWritableDirectories(IAnalyzerConfiguration configuration) : base(configuration, RspecStrings.ResourceManager) { }
+
+        private protected override bool IsGetTempPathAssignment(InvocationExpressionSyntax invocationExpression, KnownType type, string methodName, SemanticModel semanticModel) =>
+            invocationExpression.IsMethodInvocation(type, methodName, semanticModel)
+            && (invocationExpression.Parent is EqualsValueSyntax
+                || invocationExpression.Parent is AssignmentStatementSyntax
+                || invocationExpression.Parent is ReturnStatementSyntax);
```
pavel-mikula-sonarsource, line 46: Same as C# vs. IsAnyKind

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/PubliclyWritableDirectories.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        private protected override bool IsInsecureEnvironmentVariableRetrieval(InvocationExpressionSyntax invocation, KnownType type, string methodName, SemanticModel semanticModel) =>
+            invocation.IsMethodInvocation(type, methodName, semanticModel)
+            && invocation.ArgumentList?.Arguments.FirstOrDefault() is { } firstArgument
+            && InsecureEnvironmentVariables.Any(x => x.Equals(firstArgument.GetExpression()?.GetStringValue(), StringComparison.OrdinalIgnoreCase));
```
pavel-mikula-sonarsource, line 51: Same as C# vs. FindConstant

pavel-mikula-sonarsource (changes requested):
Second round. I'll inspect the regexes in 3rd round.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,133 @@
...
+            private static readonly string LibraryCachesRegex = string.Format(TwoPartPatternWithForwardSlash, "library", "caches");
+            private static readonly string UsersSharedRegex = string.Format(TwoPartPatternWithForwardSlash, "users", "shared");
+            private static readonly string PrivateTmpRegex = string.Format(TwoPartPatternWithForwardSlash, "private", "tmp");
+            private static readonly string PrivateVarTmpRegex = string.Format(ThreePartPatternWithForwardSlash, "private", "var", "tmp");
+            private static readonly string TempEnvVariable = string.Format(EnvironmentVariableTemplate, "temp");
+            private static readonly string TmpEnvVariable = string.Format(EnvironmentVariableTemplate, "tmp");
+            private static readonly string TmpDirEnvVariable = string.Format(EnvironmentVariableTemplate, "tmpdir");
+
+            private static readonly Regex LinuxPubliclyWritabelDirs =
+                new Regex($"{DevMqueueRegex}|{RunLockRegex}|{VarRunLockRegex}",
```
pavel-mikula-sonarsource, line 116: `|` needs to be in parenthesis to avoid unexpected behavior
```suggestion
                new Regex($"({DevMqueueRegex}|{RunLockRegex}|{VarRunLockRegex})",
```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,133 @@
...
+                    RegexOptions.Compiled);
+
+            private static readonly Regex WindowsAndMacPubliclyWritabelDirs =
+                new Regex($"{TmpRegex}|{VarTmpRegex}|{UsrTmpRegex}|{DevShmRegex}|{LibraryCachesRegex}|{UsersSharedRegex}|{PrivateTmpRegex}|{PrivateVarTmpRegex}|{TempRegex}|{WindowsTempRegex}",
```
pavel-mikula-sonarsource, line 120:
```suggestion
                new Regex($"({TmpRegex}|{VarTmpRegex}|{UsrTmpRegex}|{DevShmRegex}|{LibraryCachesRegex}|{UsersSharedRegex}|{PrivateTmpRegex}|{PrivateVarTmpRegex}|{TempRegex}|{WindowsTempRegex})",
```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,133 @@
...
+                    RegexOptions.Compiled | RegexOptions.IgnoreCase | RegexOptions.CultureInvariant);
+
+            private static readonly Regex EnvironmentVariables =
+                new Regex($"{UserProfileRegex}|{TempEnvVariable}|{TmpEnvVariable}|{TmpDirEnvVariable}",
```
pavel-mikula-sonarsource, line 124:
```suggestion
                new Regex($"({UserProfileRegex}|{TempEnvVariable}|{TmpEnvVariable}|{TmpDirEnvVariable})",
```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        private static class PubliclyWritableDirectoriesRegexMatcher
+        {
+            private const string TwoPartPatternWithVariableSlash = @"^[\/\\]{0}[\/\\]{1}([\/\\]|$)";
+            private const string TwoPartPatternWithForwardSlash = @"^\/{0}\/{1}(\/|$)";
+            private const string ThreePartPatternWithForwardSlash = @"^\/{0}\/{1}\/{2}(\/|$)";
+            private const string TmpRegex = @"^([a-zA-z]:)?[\\\/]?[\/\\]tmp([\/\\]|$)";
+            private const string TempRegex = @"^([a-zA-z]:)?[\\\/]?[\/\\]temp([\/\\]|$)";
+            private const string WindowsTempRegex = @"^([a-zA-z]:)?[\\\/]?[\\\/]windows[\\\/]temp([\\\/]|$)";
+            private const string UserProfileRegex = @"%USERPROFILE%\\AppData\\Local\\Temp([\\\/]|$)";
```
pavel-mikula-sonarsource, line 99: These are not regex objects, but patterns
```suggestion
            private const string UserProfilePattern = @"%USERPROFILE%\\AppData\\Local\\Temp([\\\/]|$)";
```
or
```suggestion
            private const string UserProfile = @"%USERPROFILE%\\AppData\\Local\\Temp([\\\/]|$)";
```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,133 @@
...
+                },
+                Language.SyntaxKind.InvocationExpression);
+        }
+
+        private static class PubliclyWritableDirectoriesRegexMatcher
+        {
+            private const string TwoPartPatternWithVariableSlash = @"^[\/\\]{0}[\/\\]{1}([\/\\]|$)";
+            private const string TwoPartPatternWithForwardSlash = @"^\/{0}\/{1}(\/|$)";
+            private const string ThreePartPatternWithForwardSlash = @"^\/{0}\/{1}\/{2}(\/|$)";
+            private const string TmpRegex = @"^([a-zA-z]:)?[\\\/]?[\/\\]tmp([\/\\]|$)";
```
pavel-mikula-sonarsource, line 96: It would allow way too much stuff there
```suggestion
            private const string TmpRegex = @"^([a-zA-Z]:)?[\\\/]?[\/\\]tmp([\/\\]|$)";
```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,133 @@
...
+            private const string TempRegex = @"^([a-zA-z]:)?[\\\/]?[\/\\]temp([\/\\]|$)";
```
pavel-mikula-sonarsource, line 97: Same
```suggestion
            private const string TempRegex = @"^([a-zA-Z]:)?[\\\/]?[\/\\]temp([\/\\]|$)";
```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,133 @@
...
+            private const string WindowsTempRegex = @"^([a-zA-z]:)?[\\\/]?[\\\/]windows[\\\/]temp([\\\/]|$)";
```
pavel-mikula-sonarsource, line 98:
```suggestion
            private const string WindowsTempRegex = @"^([a-zA-Z]:)?[\\\/]?[\\\/]windows[\\\/]temp([\\\/]|$)";
```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
pavel-mikula-sonarsource: Actually, it's used in `RegexOptions.IgnoreCase`, so there's no point in listing both
```suggestion
            private const string TmpRegex = @"^([a-z]:)?[\\\/]?[\/\\]tmp([\/\\]|$)";
```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,133 @@
...
+                    {
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, node.GetLocation()));
+                    }
+                },
+                Language.SyntaxKind.InvocationExpression);
+        }
+
+        private static class PubliclyWritableDirectoriesRegexMatcher
+        {
+            private const string TwoPartPatternWithVariableSlash = @"^[\/\\]{0}[\/\\]{1}([\/\\]|$)";
```
pavel-mikula-sonarsource, line 93: There's sometimes `[\/\\]` and sometimes `[\\\/]`. Please unify them to `[\\\/]` to make it easier to read.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
pavel-mikula-sonarsource, line 97: These two can be merged. It will have better performance:
```suggestion
            private const string TempRegex = @"^([a-zA-z]:)?[\\\/]?[\/\\]te?mp([\/\\]|$)";
```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
pavel-mikula-sonarsource, line 98: There three can be actually merged, for better performance.
```suggestion
            private const string TmpRegex = @"^([a-z]:)?[\\\/]?[\/\\](windows[\\\/])?te?mp([\/\\]|$)";
```
It will also match stuff like `/windows/tmp`

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
pavel-mikula-sonarsource, line 97: If I read this correctly, it matches stuff like `C:\\temp\` (invalid path) and `\\temp`. As `temp` would be a server name, I don't understand why are two backslashes matched here.
I think we need a prefix like this `^([a-zA-z]:[\\\/]?|[\\\/][\\\/][^\\\/]+[\\\/])?tmp([\\\/]|$)` instead.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
pavel-mikula-sonarsource, line 99: Current version is inconsistent with approach to Windows paths (compared to other regexes and end of this one)
```suggestion
            private const string UserProfileRegex = @"%USERPROFILE%[\\\/]AppData[\\\/]Local[\\\/]Temp([\\\/]|$)";
```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,133 @@
...
+            private const string TwoPartPatternWithForwardSlash = @"^\/{0}\/{1}(\/|$)";
+            private const string ThreePartPatternWithForwardSlash = @"^\/{0}\/{1}\/{2}(\/|$)";
+            private const string TmpRegex = @"^([a-zA-z]:)?[\\\/]?[\/\\]tmp([\/\\]|$)";
+            private const string TempRegex = @"^([a-zA-z]:)?[\\\/]?[\/\\]temp([\/\\]|$)";
+            private const string WindowsTempRegex = @"^([a-zA-z]:)?[\\\/]?[\\\/]windows[\\\/]temp([\\\/]|$)";
+            private const string UserProfileRegex = @"%USERPROFILE%\\AppData\\Local\\Temp([\\\/]|$)";
+            private const string EnvironmentVariableTemplate = @"%{0}%([\\\/]|$)";
```
pavel-mikula-sonarsource, line 100: Let's put this template together with the first 3 templates. I'd also remove `Pattern` from their names and use `Template` like `VariableSlashTwoPartTemplate`

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,133 @@
...
+            private static readonly string RunLockRegex = string.Format(TwoPartPatternWithForwardSlash, "run", "lock");
+            private static readonly string VarRunLockRegex = string.Format(ThreePartPatternWithForwardSlash, "var", "run", "lock");
+            private static readonly string LibraryCachesRegex = string.Format(TwoPartPatternWithForwardSlash, "library", "caches");
+            private static readonly string UsersSharedRegex = string.Format(TwoPartPatternWithForwardSlash, "users", "shared");
+            private static readonly string PrivateTmpRegex = string.Format(TwoPartPatternWithForwardSlash, "private", "tmp");
+            private static readonly string PrivateVarTmpRegex = string.Format(ThreePartPatternWithForwardSlash, "private", "var", "tmp");
+            private static readonly string TempEnvVariable = string.Format(EnvironmentVariableTemplate, "temp");
+            private static readonly string TmpEnvVariable = string.Format(EnvironmentVariableTemplate, "tmp");
+            private static readonly string TmpDirEnvVariable = string.Format(EnvironmentVariableTemplate, "tmpdir");
```
pavel-mikula-sonarsource, line 113: This indirection is too complicated. Let's remove these three and their template and use just this instead `%(temp|tmp|tmpdir)%([\\\/]|$)`.
That would move `UserProfileRegex` as a separate regex that can be crafted directly.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
pavel-mikula-sonarsource, line 100: Why are these two without `^`? I think they should have it.
```suggestion
            private const string UserProfileRegex = @"^%USERPROFILE%\\AppData\\Local\\Temp([\\\/]|$)";
            private const string EnvironmentVariableTemplate = @"^%{0}%([\\\/]|$)";
```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
pavel-mikula-sonarsource: And  then, it can have a separate regex to simplify it's creation.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            private static readonly Regex LinuxPubliclyWritabelDirs =
+                new Regex($"{DevMqueueRegex}|{RunLockRegex}|{VarRunLockRegex}",
+                    RegexOptions.Compiled);
+
+            private static readonly Regex WindowsAndMacPubliclyWritabelDirs =
+                new Regex($"{TmpRegex}|{VarTmpRegex}|{UsrTmpRegex}|{DevShmRegex}|{LibraryCachesRegex}|{UsersSharedRegex}|{PrivateTmpRegex}|{PrivateVarTmpRegex}|{TempRegex}|{WindowsTempRegex}",
+                    RegexOptions.Compiled | RegexOptions.IgnoreCase | RegexOptions.CultureInvariant);
```
pavel-mikula-sonarsource, line 121: I've asked to isolate few new regexes separately. Then we can use a constant to avoid this repetition 4 or 5 times (it's fine to have it twice).

pavel-mikula-sonarsource (changes requested):
Regex rumble

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PubliclyWritableDirectories.cs
csaba-sagi-sonarsource: I would say we do not need that as whether propagation is used or not, we will raise at the location wherever the string literal is used?

pavel-mikula-sonarsource: I don't understand :) Raising at the usage location is fine in all cases. Difference is if we're able to detect this
```
var name="%temp%";
Environment.GetEnvironmentVariable(name);
```

csaba-sagi-sonarsource: We register for StringLiteralExpression and InterpolatedStringExpression so we do not care about value propagation.

analyzers/src/SonarAnalyzer.CSharp/Extensions/InterpolatedStringExpressionSyntaxExtensions.cs
csaba-sagi-sonarsource: As we discussed, we will leave it as is as it is already used in multiple rules.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,133 @@
...
+            private const string TwoPartPatternWithVariableSlash = @"^[\/\\]{0}[\/\\]{1}([\/\\]|$)";
+            private const string TwoPartPatternWithForwardSlash = @"^\/{0}\/{1}(\/|$)";
+            private const string ThreePartPatternWithForwardSlash = @"^\/{0}\/{1}\/{2}(\/|$)";
+            private const string TmpRegex = @"^([a-zA-z]:)?[\\\/]?[\/\\]tmp([\/\\]|$)";
+            private const string TempRegex = @"^([a-zA-z]:)?[\\\/]?[\/\\]temp([\/\\]|$)";
+            private const string WindowsTempRegex = @"^([a-zA-z]:)?[\\\/]?[\\\/]windows[\\\/]temp([\\\/]|$)";
+            private const string UserProfileRegex = @"%USERPROFILE%\\AppData\\Local\\Temp([\\\/]|$)";
+            private const string EnvironmentVariableTemplate = @"%{0}%([\\\/]|$)";
+            private static readonly string VarTmpRegex = string.Format(TwoPartPatternWithVariableSlash, "var", "tmp");
+            private static readonly string UsrTmpRegex = string.Format(TwoPartPatternWithVariableSlash, "usr", "tmp");
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Did you create an issue for the FN?

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+        protected string[] InsecureEnvironmentVariables { get; } = { "tmp", "temp", "tmpdir" };
+
+        protected PubliclyWritableDirectoriesBase(IAnalyzerConfiguration configuration, ResourceManager rspecResources) : base(configuration)
+        {
+            rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, rspecResources).WithNotConfigurable();
+            linuxPubliclyWritabelDirs = new Regex($"({devMqueue}|{runLock}|{varRunLock})", RegexOptions.Compiled);
+            windowsAndMacPubliclyWritabelDirs = new Regex($"({varTmp}|{usrTmp}|{devShm}|{libraryCaches}|{usersShared}|{privateTmp}|{privateVarTmp}|{Temp})", WindowsAndUnixOptions);
```
pavel-mikula-sonarsource, line 70: Let's put leading `^` here, we'll improve the performance a little (probably not measurable), we can remove repetition of `^` above and we'll be sure that everyone is matched from start (contrary the case where %variables% weren't fixed at the start)

pavel-mikula-sonarsource, line 70: Let's move the ending  `([\/]|$)` and `([\/\\]|$)` to main regexes here to simplify the situation.

pavel-mikula-sonarsource, line 70: Let's split this one to `windowsDirs` and `macDirs` to keep the logic separated per domain. It will allow remove `Temp` and initialize it directly in it's standalone regex.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,110 @@
...
+        private readonly string devShm = string.Format(VariableSlashTwoPartTemplate, "dev", "shm");
+        private readonly string devMqueue = string.Format(ForwardSlashTwoPartTemplate, "dev", "mqueue");
+        private readonly string runLock = string.Format(ForwardSlashTwoPartTemplate, "run", "lock");
+        private readonly string varRunLock = string.Format(ForwardSlashThreePartTemplate, "var", "run", "lock");
+        private readonly string libraryCaches = string.Format(ForwardSlashTwoPartTemplate, "library", "caches");
+        private readonly string usersShared = string.Format(ForwardSlashTwoPartTemplate, "users", "shared");
+        private readonly string privateTmp = string.Format(ForwardSlashTwoPartTemplate, "private", "tmp");
+        private readonly string privateVarTmp = string.Format(ForwardSlashThreePartTemplate, "private", "var", "tmp");
+        private readonly DiagnosticDescriptor rule;
+        private readonly Regex linuxPubliclyWritabelDirs;
```
pavel-mikula-sonarsource, line 53: To keep things simple and aligned with my other comment. We're inside `PubliclyWritableDirectoriesBase` scope.
```suggestion
        private readonly Regex linuxDirs;
```

pavel-mikula-sonarsource (changes requested):
Final polishing

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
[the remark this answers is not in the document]
csaba-sagi-sonarsource: Not yet but I will before the end of the day. a reproducer for the UTs is already in place

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,112 @@
...
+{
+    public abstract class PubliclyWritableDirectoriesBase<TSyntaxKind, TInvocationExpression> : HotspotDiagnosticAnalyzer
+        where TSyntaxKind : struct
+        where TInvocationExpression : SyntaxNode
+    {
+        protected const string DiagnosticId = "S5443";
+        private const string MessageFormat = "Make sure publicly writable directories are used safely here.";
+        private const RegexOptions WindowsAndUnixOptions = RegexOptions.Compiled | RegexOptions.IgnoreCase | RegexOptions.CultureInvariant;
+        private const string ForwardSlashTwoPartTemplate = @"\/{0}\/{1}";
+        private const string ForwardSlashThreePartTemplate = @"\/{0}\/{1}\/{2}";
```
pavel-mikula-sonarsource, line 38: Now it's clear that we can remove these and use `list.JoinStr("|", x => Regex.Escape(x))` instead.

So we can introduce two new array fields with listed directorlies like `/var/tmp` (one for linux, one for mac) and construc regexes from them in a short and simple way

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/PubliclyWritableDirectoriesBase.cs
```diff
@@ -0,0 +1,112 @@
...
+            var runLock = string.Format(ForwardSlashTwoPartTemplate, "run", "lock");
+            var varRunLock = string.Format(ForwardSlashThreePartTemplate, "var", "run", "lock");
+            var libraryCaches = string.Format(ForwardSlashTwoPartTemplate, "library", "caches");
+            var usersShared = string.Format(ForwardSlashTwoPartTemplate, "users", "shared");
+            var privateTmp = string.Format(ForwardSlashTwoPartTemplate, "private", "tmp");
+            var privateVarTmp = string.Format(ForwardSlashThreePartTemplate, "private", "var", "tmp");
+            rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, rspecResources).WithNotConfigurable();
+            linuxDirectories = new Regex($@"^({devMqueue}|{runLock}|{varRunLock})(\/|$)", RegexOptions.Compiled);
+            macDirectories = new Regex($@"^({libraryCaches}|{usersShared}|{privateTmp}|{privateVarTmp}|{varTmp}|{usrTmp}|{devShm})(\/|$)", WindowsAndUnixOptions);
+            windowsDirectories = new Regex(@"^([a-z]:[\\\/]?|[\\\/][\\\/][^\\\/]+[\\\/]|[\\\/])(windows[\\\/])?te?mp([\\\/]|$)", WindowsAndUnixOptions);
```
pavel-mikula-sonarsource, line 70: This field can be initialized inline

pavel-mikula-sonarsource (changes requested):
Final round

pavel-mikula-sonarsource (approved):
LGTM!

PR #4067, opened 2021-02-15
Rule S5693: Allowing requests with excessive content length FN fix and secondary locations

csaba-sagi-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/RequestsWithExcessiveLength.vb
```diff
@@ -13,13 +13,13 @@ Namespace Tests.TestCases
         End Function
 
         <HttpPost>
-        <RequestSizeLimit(2_000_001)> ' Noncompliant ^10#27 {{Make sure the content length limit is safe here.}}
+        <RequestSizeLimit(8_000_001)> ' Noncompliant ^10#27 {{Make sure the content length limit is safe here.}}
```
andrei-epure-sonarsource, line 16: what does `^10#27` mean?

also, the comment in line 10 confuses me ... `^10#25`?

(same question for analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/RequestsWithExcessiveLength_CustomValues.vb)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/RequestsWithExcessiveLength.vb
```diff
@@ -78,6 +78,19 @@ Namespace Tests.TestCases
...
         End Function
 
+        <HttpPost>
+        <RequestSizeLimit(Integer.MaxValue)> ' Noncompliant
+        Public Function RequestSizeLimitWithoutNumericLiteral() As ActionResult
+            Return Nothing
+        End Function
+
+        <HttpPost>
+        <RequestFormLimits(MultipartBodyLengthLimit:=1000000000)> ' Noncompliant [1]
```
andrei-epure-sonarsource, line 88: what happens when
- `RequestFormLimits MultipartBodyLengthLimit` is above limit and `RequestSizeLimit` is below limit - will the secondary location appear as well?
- viceversa: `RequestSizeLimit` is above limit and ``RequestFormLimits MultipartBodyLengthLimit`` is above limit?

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -59,28 +58,76 @@ protected RequestsWithExcessiveLengthBase(System.Resources.ResourceManager rspec
...
+                            var attribute = (TAttributeSyntax)cc.Node;
+
+                            if (attribute.IsKnownType(KnownType.Microsoft_AspNetCore_Mvc_DisableRequestSizeLimitAttribute, cc.SemanticModel))
+                            {
+                                cc.ReportDiagnosticWhenActive(Diagnostic.Create(rule, attribute.GetLocation()));
+                            }
+
+                            if ((IsInvalidRequestSizeLimit(attribute, cc.SemanticModel)
+                                 || IsInvalidRequestFormLimits(attribute, cc.SemanticModel))
+                                && GetMethodOrClassDeclaration(attribute, cc.SemanticModel) is { } delcaration)
```
andrei-epure-sonarsource, line 83: typo 
```suggestion
                                && GetMethodOrClassDeclaration(attribute, cc.SemanticModel) is { } declaration)
```

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -59,28 +58,76 @@ protected RequestsWithExcessiveLengthBase(System.Resources.ResourceManager rspec
...
 
         private bool IsEnabled(AnalyzerOptions options)
         {
             analyzerConfiguration.Initialize(options);
             return SupportedDiagnostics.Any(d => analyzerConfiguration.IsEnabled(d.Id));
         }
+
+        private struct Attributes
+        {
+            public SyntaxNode InvalidRequestForm { get; set; }
```
andrei-epure-sonarsource, line 128: do the setters need to be public? can't you just make a constructor for the initialization,  and then have getter-only properties?

from what I see, you don't overwrite the values, so this should be immutable

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/RequestsWithExcessiveLength.cs
```diff
@@ -14,15 +14,15 @@ public ActionResult PostRequestWithNoLimit()
         }
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: also, add tests with attributes on local functions (C# 9 tests) - to document the functionality

I'm not sure it's valid behavior, but try to have a method with an attribute and inside it a local function with an attribute

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -59,28 +58,76 @@ protected RequestsWithExcessiveLengthBase(System.Resources.ResourceManager rspec
...
+
+                            if (attribute.IsKnownType(KnownType.Microsoft_AspNetCore_Mvc_DisableRequestSizeLimitAttribute, cc.SemanticModel))
+                            {
+                                cc.ReportDiagnosticWhenActive(Diagnostic.Create(rule, attribute.GetLocation()));
+                            }
+
+                            if ((IsInvalidRequestSizeLimit(attribute, cc.SemanticModel)
+                                 || IsInvalidRequestFormLimits(attribute, cc.SemanticModel))
+                                && GetMethodOrClassDeclaration(attribute, cc.SemanticModel) is { } delcaration)
+                            {
```
andrei-epure-sonarsource, line 84: this makes the overall method a bit complex, please move to a separate method, maybe named "CollectAttributesOverTheLimit"

the the pair method in EndAction could be extracted in "ReportOnCollectedAttributes"

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -59,28 +58,76 @@ protected RequestsWithExcessiveLengthBase(System.Resources.ResourceManager rspec
...
+                                var isRequestFormLimits = attribute.IsKnownType(KnownType.Microsoft_AspNetCore_Mvc_RequestFormLimitsAttribute, cc.SemanticModel);
```
andrei-epure-sonarsource, line 85: you are doing this query twice (once in the `if`, then here) and this is quite an expensive call (whenever you use the SemanticModel, think : "expensive performance cost").

Could you, when doing the `Is<Attribute>` call , extrac the `TAttributeSyntax` node , and then based on what's null or not null , create the `Attribute`?

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -59,28 +58,76 @@ protected RequestsWithExcessiveLengthBase(System.Resources.ResourceManager rspec
...
+
+                                attributesOverTheLimit.AddOrUpdate(
+                                    delcaration,
+                                    new Attributes
+                                    {
+                                        InvalidRequestForm = isRequestFormLimits ? attribute : null,
```
andrei-epure-sonarsource, line 91: It seems that `InvalidRequestForm` is exclusive with `InvalidRequestSize` - which means that one of the fields will always be null.  That makes the code hard to follow.

What I see you want to do is:
- collect all attributes of the sought types for a method
- iterate over them

Can't you just keep a collection of `TAttributeSyntax` per Declaration , make sure you don't add duplicates, and then use that? And like this, you get rid of the unnecessary `Attribute` struct.

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
andrei-epure-sonarsource: (also, this was hinted in the SonarCloud cognitive complexity issue)

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -55,28 +58,76 @@ protected RequestsWithExcessiveLengthBase(System.Resources.ResourceManager rspec
...
-                },
-                Language.SyntaxKind.Attribute);
+                    var attributesOverTheLimit = new ConcurrentDictionary<SyntaxNode, Attributes>();
+
+                    c.RegisterSyntaxNodeActionInNonGenerated(Language.GeneratedCodeRecognizer,
+                        cc =>
+                        {
+                            if (!IsEnabled(c.Options))
+                            {
+                                return;
```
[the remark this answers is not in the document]
csaba-sagi-sonarsource: SecurityHotspotTest.SecurityHotspotRules_AreDisabledByDefault. This is the test that does the verification for both CS and VB hotspots. IT was not moved there in #4033. Those tests already existed so we actually had 2 sets of tests verifying the same functionality so I just removed the tests from the Rule test cases.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/RequestsWithExcessiveLength.vb
csaba-sagi-sonarsource: ^10#27 means the issue begins on column 10 and has a length of 27

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
andrei-epure-sonarsource: Or you could make the `Attribute` a bit smarter, and let it have the same fields, but let it deal with them inside, in the constructor.

And the getter should be "MainAttribute" and "SecondaryAttribute" which can be optional.

But the logic would be contained in the `Attribute` class.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/RequestsWithExcessiveLength.vb
andrei-epure-sonarsource: could you please use the alternative syntax `//        ^^^^^^^^^^^^` because this is a visual way of showing what's underlined. Otherwise, we cannot really verify in github, and it's painful to verify in the IDE, as well.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/RequestsWithExcessiveLength.cs
```diff
@@ -41,14 +41,19 @@ public sealed class RequestsWithExcessiveLength : RequestsWithExcessiveLengthBas
         protected override bool IsInvalidRequestFormLimits(AttributeSyntax attribute, SemanticModel semanticModel) =>
             attribute.IsKnownType(KnownType.Microsoft_AspNetCore_Mvc_RequestFormLimitsAttribute, semanticModel)
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: Actually, it may not be obvious on how to move the logic in `IsKnownType` , so we should leave that for a separate PR/issue.

Here you can just check the name with the expected attribute name. An better than using contains you can just compare the two attribute names (with or without the Attribute suffix). And given you'd do this across multiple places, extract a method for it.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/RequestsWithExcessiveLength.vb
csaba-sagi-sonarsource: For CS it is with the `// ^^^^^^^` syntax but for VB it is not possible. It is nor permitted to insert a comment between the attribute and the method.

andrei-epure-sonarsource: maybe you can use `'         ^^^^^@-1` for the VB case? see [ex1](https://github.com/SonarSource/sonar-dotnet/blob/d9451117ff541bf5aa2e83998ba7b945e8f78b5a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/BooleanLiteralUnnecessary.vb#L46) , [ex2](https://github.com/SonarSource/sonar-dotnet/blob/d9451117ff541bf5aa2e83998ba7b945e8f78b5a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CognitiveComplexity.vb#L359) , [ex3](https://github.com/SonarSource/sonar-dotnet/blob/d9451117ff541bf5aa2e83998ba7b945e8f78b5a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CognitiveComplexity.vb#L482)

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -59,28 +62,92 @@ protected RequestsWithExcessiveLengthBase(System.Resources.ResourceManager rspec
...
 
         private bool IsEnabled(AnalyzerOptions options)
         {
             analyzerConfiguration.Initialize(options);
             return SupportedDiagnostics.Any(d => analyzerConfiguration.IsEnabled(d.Id));
         }
+
+        private struct Attributes
+        {
+            private readonly SyntaxNode invalidRequestForm;
```
andrei-epure-sonarsource, line 131: I suggest to name them "requestForm" and "requestSize"

this struct doesn't ensure they are invalid or not - it can hold valid attributes as well

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -59,28 +62,92 @@ protected RequestsWithExcessiveLengthBase(System.Resources.ResourceManager rspec
...
+            public SyntaxNode SecondaryAttribute =>
+                invalidRequestForm != null && invalidRequestSize != null ? invalidRequestSize : null;
+
+            public Attributes(SyntaxNode invalidRequestForm, SyntaxNode invalidRequestSize)
+            {
+                this.invalidRequestForm = invalidRequestForm;
+                this.invalidRequestSize = invalidRequestSize;
+            }
+
+            public Attributes(SyntaxNode invalidRequestForm, SyntaxNode invalidRequestSize, Attributes oldAttributes)
```
andrei-epure-sonarsource, line 146: could you add a test with multiple instances of attributes?

```

    [RequestSizeLimit(8_000_001)]
    [RequestSizeLimit(8_000_002)]
    [RequestSizeLimit(10)]
    [RequestSizeLimit(8_000_003)]
    public class RequestSizeLimitAboveController : Controller
    {

    }
```

andrei-epure-sonarsource (changes requested):
Looks much better now 👍 

I've left another round of suggestions

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/RequestsWithExcessiveLength.cs
```diff
@@ -38,17 +39,28 @@ public sealed class RequestsWithExcessiveLength : RequestsWithExcessiveLengthBas
...
-            && Language.ExpressionNumericConverter.TryGetConstantIntValue(firstArgument.Expression, out var intValue)
-            && intValue > StandardSizeLimit;
+            && semanticModel.GetConstantValue(firstArgument.Expression) is { HasValue: true } constantValue
+            && constantValue.Value is int intValue
+            && intValue > FileUploadSizeLimit
+            && attribute.IsKnownType(KnownType.Microsoft_AspNetCore_Mvc_RequestSizeLimitAttribute, semanticModel)
+                ? attribute
+                : null;
+
+        protected override SyntaxNode GetMethodOrClassDeclaration(AttributeSyntax attribute, SemanticModel semanticModel) =>
```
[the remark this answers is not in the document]
csaba-sagi-sonarsource: I do get it, There are test cases that verify that the issue is found if the attribute is on the class itself.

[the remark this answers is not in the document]
csaba-sagi-sonarsource: How about MethodLcalFunctionOrClassDeclaration ?

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
csaba-sagi-sonarsource: The build will fail as multiple attributes are not valid, and as this is a Hotspot and it does not run in Sonarlint, Should we cover these cases? The implementation with the struct was based on the assumption not to cover such cases.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/RequestsWithExcessiveLength.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: sorry, I looked at `MemberDeclarationSyntax` and I saw `MethodDeclarationSyntax` , nevermind, you can ignore the comment

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
andrei-epure-sonarsource: Ok. Maybe mention it in the code where the `Attribute` struct is used that there is this assumption. Maybe in the future this rule will be migrated and it will help devs to understand the initial assumptions.

andrei-epure-sonarsource (approved):
LGTM. I suggest that you can specify in a comment the assumption that there will be only one attribute per method and that this rule is not expected to run in SonarLint (where code may not be compileable)

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -136,7 +136,7 @@ private bool IsEnabled(AnalyzerOptions options)
             return SupportedDiagnostics.Any(d => analyzerConfiguration.IsEnabled(d.Id));
         }
 
-        private struct Attributes
+        private struct Attributes // This struct is used as the same attributes can not be applied multiple times to the same declaration.
```
andrei-epure-sonarsource, line 139: - nitpick: should be on the line above, we avoid inline comments (especially when it's long)

PR #4072, opened 2021-02-16
Don't inspect assembly name in test detection

### LINKED ISSUE #4062: Don't inspect assembly name in test detection
`Main` vs. `Test` project detection is based on several factors like assembly name and referenced projects. Assembly name was inspected to detect test projects based on presence of "Test" and "Tests" words. 

This has major drawbacks:
- Users that want to try the product will likely use "Test" as part of the testing project name.
- Behavior is inconsistent between products and tools (SonarLint, Scanner for MSBuild, Analyzer, Security engine).

We need to remove the part inspecting the assembly name.

pavel-mikula-sonarsource:
Fixes #4062

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectTypeHelperTests.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectTypeHelperTests.cs
@@ -19,8 +19,10 @@
  */
 
 using System.Collections.Generic;
+using System.Threading;
 using FluentAssertions;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarAnalyzer.Helpers;
 using SonarAnalyzer.UnitTest.MetadataReferences;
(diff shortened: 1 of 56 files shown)


its/src/test/java/com/sonar/it/shared/TestUtils.java
```diff
@@ -177,14 +177,20 @@ public static Integer getMeasureAsInt(Orchestrator orch, String componentKey, St
...
   }
 
   public static void reset(Orchestrator orchestrator) {
+    // Only for local debugging:
+    if (orchestrator.getServer() == null) {
+      // Tests.ORCHESTRATOR is a jUnit rule that is automagically started in it's beforeAll() action for all tests.
+      // Running individual test doesn't execute the @ClassRule annotation on ORCHESTRATOR in Tests class.
+      orchestrator.start();
+    }
+
```
pavel-mikula-sonarsource, line 193: As discussed, this will start orchestrator if needed for a local test run of an individual test.

its/src/test/java/com/sonar/it/csharp/NoSourcesTest.java
```diff
@@ -44,7 +44,7 @@ public static void init() throws Exception {
   }
 
   @Test
-  public void filesAtProjectLevel() {
+  public void raiseOnSharedFiles() {
```
pavel-mikula-sonarsource, line 47: Just a cleanup to deconfuse it. Name was copy pasted from `MetricsTest.java`

its/src/test/java/com/sonar/it/csharp/NoSonarTest.java
```diff
@@ -40,19 +40,16 @@
...
     TestUtils.reset(ORCHESTRATOR);
-
-    // Without setting the testProjectPattern, the NoSonarTest project is considered as a Test project :)
-    Tests.analyzeProject(temp, PROJECT, "class_name", "sonar.msbuild.testProjectPattern", "noTests");
+    Tests.analyzeProject(temp, PROJECT, "class_name");
   }
 
   @Test
-  public void filesAtProjectLevel() {
+  public void excludeNoSonarComment() {
```
pavel-mikula-sonarsource, line 47: Just a cleanup to deconfuse it. Name was copy pasted from `MetricsTest.java`

its/src/test/java/com/sonar/it/csharp/NoSonarTest.java
```diff
@@ -40,19 +40,16 @@
   @BeforeClass
   public static void init() throws Exception {
     TestUtils.reset(ORCHESTRATOR);
-
-    // Without setting the testProjectPattern, the NoSonarTest project is considered as a Test project :)
-    Tests.analyzeProject(temp, PROJECT, "class_name", "sonar.msbuild.testProjectPattern", "noTests");
+    Tests.analyzeProject(temp, PROJECT, "class_name");
```
pavel-mikula-sonarsource, line 43: Replaced deprecated `sonar.msbuild.testProjectPattern` with `<SonarQubeTestProject>false</SonarQubeTestProject>` to simplify the situation (new scanner is not released yet).

its/projects/NoSonarTest/Class1.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace CSLib.foo
+﻿[assembly: System.Reflection.AssemblyVersion("1.0.0")]
```
pavel-mikula-sonarsource, line 1: This one is funny in a bitter way: `sonar.msbuild.testProjectPattern` marked project as a MAIN for S4MSB, but analyzer considered it as a test anyway. So only S101 was triggered (partially lucky), because of it's MAIN+TEST scope.

Changing analyzer implementation cleaned up the air and S3904 started to raise. I added this line to fix it and keep the assertions clean & focused.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectTypeHelperTests.cs
```diff
@@ -18,24 +18,40 @@
...
+using Microsoft.CodeAnalysis;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.UnitTest.MetadataReferences;
+using SonarAnalyzer.UnitTest.TestFramework;
 
 namespace SonarAnalyzer.UnitTest.Helpers
 {
     [TestClass]
     public class ProjectTypeHelperTests
```
andrei-epure-sonarsource, line 32: can you also add a small test for `IsTest(this SyntaxNodeAnalysisContext context)` ?

andrei-epure-sonarsource (changes requested):
Nice! I've left a small suggestion to add another UT

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectTypeHelperTests.cs
pavel-mikula-sonarsource: As the context has a longer chain, I swapped existing tests to context and added small one for compilation (it produces less scaffolding for more value like this)

andrei-epure-sonarsource (approved):
LGTM

Issue #3719, opened 2020-11-05
Fix S1854 FP: Variable used in local function as delegate


pavel-mikula-sonarsource:
Separate issue for case reported in https://github.com/SonarSource/sonar-dotnet/issues/3169#issuecomment-671863449 by @Rubbiroid.

```C#
public void StoreEvents(IList<IEvent> events)
{
    var debouncedEventTypes = new HashSet<Type>();
    var debouncedEvents = events.Where(NotYetDebounced);

    foreach (var debouncedEvent in debouncedEvents)
    {
        StoreEvent(debouncedEvent);
    }

     bool NotYetDebounced(IEvent @event)
     {
         return !(@event is IWithDebounce) || debouncedEventTypes.Add(@event.GetType());
     }
}
```

sonar says that debouncedEventTypes is not used: Remove this useless assignment to local variable S1854

We have Sonar Version 8.3.1 (build 34397) and C# plugin 8.6.1 (build 17183)

pavel-mikula-sonarsource:
This is a special case of #3126 when the local method is used as a delegate

Issue #3126, opened 2020-02-17
Fix S1854 FP: When a variable is used inside local function


volgunin:
### Description

When a variable is used inside local function, S1854 is raised.

### Repro steps

        public string Test()
        {
            string buffer = new string('a', 1);
            return Local();

            string Local()
            {
                return buffer;
            }
        }

### Expected behavior

No warning.

### Actual behavior

warning S1854: Remove this useless assignment to local variable 'buffer'.

### Related information

* SonarAnalyzer.CSharp.8.2.0.14119
* VisualStudio 2019 16.4.4

andrei-epure-sonarsource:
We see this in our own project as well ([see sonarcloud](https://sonarcloud.io/project/issues?id=sonaranalyzer-dotnet&open=AXAbMGyqo9K0T9AzRkIS&resolved=false&types=CODE_SMELL))
putting a screenshot for the reference, as sonarcloud link will become onbsolete once we fix this

![image](https://user-images.githubusercontent.com/38876598/75554501-b8028f80-5a3a-11ea-857c-0442d500bae4.png)

paulhickman-a365:
I have encountered a similar case, but rather than calling the local function that uses the variable, I'm passing it as a delegate to another function to be called:

Sonar Analzyer Version: 8.4.0.15306
```
        public void Foo(Func<int> a)
        {
            a();
        }

        public void Bar()
        {
            var x = 4;
            Foo(local);

            int local()
            {
                return x;
            }
        }
````

pavel-mikula-sonarsource:
Another example on [community forum](https://community.sonarsource.com/t/sonarlint-for-visual-studio-useless-assignment-warning-for-value-captured-in-local-function/32658)

costin-zaharia-sonarsource:
And another one: https://community.sonarsource.com/t/s1854-false-positive-s1854-unused-assignments-should-be-removed

PR #4074, opened 2021-02-16
S1854: Add FP repro for #3719

lg2de:
I tried to the fix the bug #3719.
But I was only able to create the repro. If you agree we can make it public available.

Unfortunately I was not able to fix the bug because the implementation of "DeadStore" is very complex and the documentation (comments) is "limited"...
Maybe someone else is able to fix the problem with my repro.

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Runtime.Versioning;
 
 namespace Tests.Diagnostics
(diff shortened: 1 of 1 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DeadStores.cs
```diff
@@ -921,4 +922,41 @@ public int[] Ranges(int[] values)
...
+            {
+                return usedBool;
+            }
+
+            bool BoolInitializer(bool value)
+            {
+                return value;
+            }
+        }
+
```
pavel-mikula-sonarsource, line 944: This case is already documented on line 883 from #3126
```suggestion
```

pavel-mikula-sonarsource (approved):
LGTM, thank you for taking time to document this @lg2de. It will help us later to implement it correctly.

PR #4075, opened 2021-02-18
Log MAIN and TEST projects summary as INFO level

### LINKED ISSUE #4054: Report Main and Test projects in DotNetSensor
Gather test vs main project statistics in a module level sensor.
Report INFO with number of main and test projects analyzed.

andrei-epure-sonarsource:
Fixes #4054

### DIFF
--- a/its/src/test/java/com/sonar/it/csharp/MultipleProjectsTest.java
+++ b/its/src/test/java/com/sonar/it/csharp/MultipleProjectsTest.java
@@ -81,8 +81,8 @@ public class MultipleProjectsTest {
           ScannerForMSBuild beginStep = TestUtils.createBeginStep(PROJECT, projectDir);
 
           CommandExecutor.create().execute(Command.create("nuget")
-            .addArguments("restore")
-            .setDirectory(projectDir.toFile()),
+              .addArguments("restore")
+              .setDirectory(projectDir.toFile()),
             10 * 60 * 1000);
 
           ORCHESTRATOR.executeBuild(beginStep);
@@ -96,7 +96,7 @@ public class MultipleProjectsTest {
 
   @Test
   public void projectTypesInfoIsLogged() {
-    assertThat(buildResult.getLogs()).contains("Found 4 MSBuild projects. 2 MAIN project(s). 2 TEST project(s).");
+    assertThat(buildResult.getLogs()).contains("Found 4 MSBuild projects. 2 MAIN projects. 2 TEST projects.");
   }
 
   @Test
--- a/its/src/test/java/com/sonar/it/csharp/MultipleProjectsTest.java
+++ b/its/src/test/java/com/sonar/it/csharp/MultipleProjectsTest.java
@@ -20,6 +20,7 @@
 package com.sonar.it.csharp;
 
 import com.sonar.it.shared.TestUtils;
+import com.sonar.orchestrator.build.BuildResult;
 import com.sonar.orchestrator.build.ScannerForMSBuild;
 import com.sonar.orchestrator.util.Command;
 import com.sonar.orchestrator.util.CommandExecutor;
@@ -59,6 +60,9 @@ public class MultipleProjectsTest {
   @ClassRule
   public static RuleChain chain = getRuleChain();
 
+  // populated in getRuleChain()
+  private static BuildResult buildResult;
+
   private static RuleChain getRuleChain() {
     assertThat(SystemUtils.IS_OS_WINDOWS).withFailMessage("OS should be Windows.").isTrue();
 
@@ -85,9 +89,14 @@ public class MultipleProjectsTest {
 
           TestUtils.runMSBuild(ORCHESTRATOR, projectDir, "/t:Rebuild");
 
-          ORCHESTRATOR.executeBuild(TestUtils.createEndStep(projectDir));
+          buildResult = ORCHESTRATOR.executeBuild(TestUtils.createEndStep(projectDir));
         }
       });
+  }
+
+  @Test
+  public void projectTypesInfoIsLogged() {
+    assertThat(buildResult.getLogs()).contains("Found 4 MSBuild projects. 2 MAIN project(s). 2 TEST project(s).");
   }
 
   @Test
--- a/sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/CSharpPlugin.java
+++ b/sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/CSharpPlugin.java
@@ -25,7 +25,9 @@ import org.sonarsource.dotnet.shared.plugins.CodeCoverageProvider;
 import org.sonarsource.dotnet.shared.plugins.DotNetPluginMetadata;
 import org.sonarsource.dotnet.shared.plugins.DotNetSensor;
 import org.sonarsource.dotnet.shared.plugins.EncodingPerFile;
+import org.sonarsource.dotnet.shared.plugins.FileTypeSensor;
 import org.sonarsource.dotnet.shared.plugins.GeneratedFileFilter;
+import org.sonarsource.dotnet.shared.plugins.ProjectTypeCollector;
 import org.sonarsource.dotnet.shared.plugins.PropertiesSensor;
 import org.sonarsource.dotnet.shared.plugins.ProtobufDataImporter;
 import org.sonarsource.dotnet.shared.plugins.ReportPathCollector;
@@ -55,6 +57,7 @@ public class CSharpPlugin implements Plugin {
     context.addExtensions(
       METADATA,
       CSharp.class,
+      ProjectTypeCollector.class,
       ReportPathCollector.class,
       CSharpSonarRulesDefinition.class,
       DotNetSensor.class,
@@ -64,6 +67,7 @@ public class CSharpPlugin implements Plugin {
       WrongEncodingFileFilter.class,
       EncodingPerFile.class,
       GeneratedFileFilter.class,
+      FileTypeSensor.class,
       PropertiesSensor.class,
       SonarLintProfileExporter.class,
       ProtobufDataImporter.class,
--- a/sonar-dotnet-shared-library/pom.xml
+++ b/sonar-dotnet-shared-library/pom.xml
@@ -63,9 +63,9 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-engine</artifactId>
-      <version>5.7.1</version>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.13.1</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/sonar-dotnet-shared-library/pom.xml
+++ b/sonar-dotnet-shared-library/pom.xml
@@ -63,9 +63,9 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <version>4.13.1</version>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <version>5.7.1</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
@@ -75,9 +75,9 @@ public class DotNetSensor implements ProjectSensor {
   }
 
   private boolean shouldExecuteOnProject(FileSystem fs) {
-    if (hasFilesOfType(fs, Type.MAIN)) {
+    if (SensorContextUtils.hasFilesOfType(fs, Type.MAIN, pluginMetadata.languageKey())) {
       return true;
-    } else if (hasFilesOfType(fs, Type.TEST)) {
+    } else if (SensorContextUtils.hasFilesOfType(fs, Type.TEST, pluginMetadata.languageKey())) {
       LOG.warn("This sensor will be skipped, because the current solution contains only TEST files and no MAIN files. " +
           "Your SonarQube/SonarCloud project will not have results for {} files. " +
           "You can read more about the detection of test projects here: https://github.com/SonarSource/sonar-scanner-msbuild/wiki/Analysis-of-product-projects-vs.-test-projects",
@@ -87,10 +87,6 @@ public class DotNetSensor implements ProjectSensor {
       LOG.debug("No files to analyze. Skip Sensor.");
     }
     return false;
-  }
-
-  private boolean hasFilesOfType(FileSystem fs, Type fileType) {
-    return fs.inputFiles(fs.predicates().and(fs.predicates().hasType(fileType), fs.predicates().hasLanguage(pluginMetadata.languageKey()))).iterator().hasNext();
   }
 
   private void executeInternal(SensorContext context) {
--- a/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
@@ -49,11 +49,13 @@ public class DotNetSensor implements ProjectSensor {
   private final RoslynDataImporter roslynDataImporter;
   private final DotNetPluginMetadata pluginMetadata;
   private final ReportPathCollector reportPathCollector;
+  private final ProjectTypeCollector projectTypeCollector;
 
-  public DotNetSensor(DotNetPluginMetadata pluginMetadata, ReportPathCollector reportPathCollector,
-    ProtobufDataImporter protobufDataImporter, RoslynDataImporter roslynDataImporter) {
+  public DotNetSensor(DotNetPluginMetadata pluginMetadata, ReportPathCollector reportPathCollector,ProjectTypeCollector projectTypeCollector,
+                      ProtobufDataImporter protobufDataImporter, RoslynDataImporter roslynDataImporter) {
     this.pluginMetadata = pluginMetadata;
     this.reportPathCollector = reportPathCollector;
+    this.projectTypeCollector = projectTypeCollector;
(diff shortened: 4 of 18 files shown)


andrei-epure-sonarsource:
I will sync with @costin-zaharia-sonarsource on how to split adding more ITs, but that can be done in a separate PR

sonar-dotnet-shared-library/pom.xml
```diff
@@ -63,9 +63,9 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <version>4.13.1</version>
+      <groupId>org.junit.jupiter</groupId>
```
andrei-epure-sonarsource, line 66: I initially updated it for parameterized tests (junit 4 offers a pretty bad experience), but then realized it's better to not do parameterized tests for my case (too complex)

however, it's not hurting to be on the latest

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensor.java
```diff
@@ -0,0 +1,95 @@
...
+    Optional<String> projectName = configuration.get("sonar.projectName");
+    if (projectName.isPresent()) {
+      LOG.debug("Adding file type information (has MAIN '{}', has TEST '{}') for project '{}' (base dir '{}'). For debug info, see ProjectInfo.xml in '{}'.",
+        hasMainFiles, hasTestFiles, projectName.get(), getValueOrEmpty(configuration.get("sonar.projectBaseDir")), getValueOrEmpty(getProjectOutPath(configuration)));
+      projectTypeCollector.addProjectInfo(hasMainFiles, hasTestFiles);
+    }
+  }
+
+  // this is duplicated from future DotNetSensor
+  private boolean hasFilesOfType(FileSystem fs, Type fileType) {
```
andrei-epure-sonarsource, line 78: ok, I missed this after rebase. I will refactor to avoid duplication after I get the review comments, together

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+public class FileTypeSensorTest {
+  public static final String REPO_KEY = "REPO_KEY";
+  public static final String LANG_KEY = "LANG_KEY";
+  public static final String LANG_NAME = "LANG_NAME";
+
+  @Rule
+  public LogTester logTester = new LogTester();
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
```
andrei-epure-sonarsource, line 54: not used, remove

sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/CSharpPlugin.java
```diff
@@ -55,6 +57,7 @@ public void define(Context context) {
     context.addExtensions(
       METADATA,
       CSharp.class,
+      ProjectTypeCollector.class,
```
pavel-mikula-sonarsource, line 60: _Out of curiosity:_ Does order of these types have any impact on their dependencies? Can we sort them?

its/src/test/java/com/sonar/it/csharp/MultipleProjectsTest.java
```diff
@@ -85,11 +89,16 @@ protected void before() throws Throwable {
...
 
-          ORCHESTRATOR.executeBuild(TestUtils.createEndStep(projectDir));
+          buildResult = ORCHESTRATOR.executeBuild(TestUtils.createEndStep(projectDir));
         }
       });
   }
 
+  @Test
+  public void projectTypesInfoIsLogged() {
+    assertThat(buildResult.getLogs()).contains("Found 4 MSBuild projects. 2 MAIN project(s). 2 TEST project(s).");
```
pavel-mikula-sonarsource, line 99: We have `pluralize(..)` duplicated in `ProtobufDataImporter` and `RoslynDataImporter`. It would be useful here as well so it would be worth extracting it to a reusable method.

sonar-dotnet-shared-library/pom.xml
andrei-epure-sonarsource: 0% coverage may be due to this change, I will revert the `pom.xml` changes

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensor.java
```diff
@@ -0,0 +1,95 @@
...
+
+  private Optional<String> getProjectOutPath(Configuration configuration) {
+    if (pluginMetadata.languageKey().equalsIgnoreCase("cs")) {
+      return configuration.get("sonar.cs.analyzer.projectOutPath");
+    } else if (pluginMetadata.languageKey().equalsIgnoreCase("vbnet")) {
+      return configuration.get("sonar.vbnet.analyzer.projectOutPath");
+    } else {
+      return Optional.empty();
+    }
+  }
```
pavel-mikula-sonarsource, line 90: There must be a better, generic way of doing this. In worst case scenario,
`return configuration.get("sonar."+pluginMetadata.languageKey().toLowerCase()+".analyzer.projectOutPath");`

Shared base should not be aware of explicit list of it's users.

pavel-mikula-sonarsource: While looking at this, it seems save to assume we can use language key directly and don't care about case sensitivity issue.

 https://github.com/SonarSource/sonar-dotnet/blob/fb6f4ae1deda8d2e9fe0c8b8699744c596cea024/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/CodeCoverageProvider.java#L45-L62

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+
+  @Test
+  public void should_describe() {
+
+    SensorDescriptor desc = mock(SensorDescriptor.class);
+    sensor.describe(desc);
+
+    verify(desc).name("Verify what types of files (MAIN, TEST) are in LANG_KEY projects.");
```
pavel-mikula-sonarsource, line 85: ..and also verify that language filter is empty?

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+    projectTypeCollectorMock = mock(ProjectTypeCollector.class);
+
+    pluginMetadataMock = mock(DotNetPluginMetadata.class);
+    when(pluginMetadataMock.languageKey()).thenReturn(LANG_KEY);
+
+    sensor = new FileTypeSensor(projectTypeCollectorMock, pluginMetadataMock);
+  }
+
+
```
pavel-mikula-sonarsource, line 78: _Nitpick:_ We're not covered by stylecop here :) 
```suggestion

```

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
```diff
@@ -0,0 +1,188 @@
...
+    when(settingsMock.getString("sonar.projectName")).thenReturn(null);
+
+    sensor.execute(tester);
+
+    assertThat(logTester.logs()).isEmpty();
+    verifyZeroInteractions(projectTypeCollectorMock);
+  }
+
+  @Test
+  public void whenLanguageKeyIsPresent_logsCsOutputFile() {
```
pavel-mikula-sonarsource, line 99: Cs was CSharp originally?
```suggestion
  public void whenLanguageKeyIsPresent_logsOutputFile() {
```

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
```diff
@@ -0,0 +1,188 @@
...
+    verify(projectTypeCollectorMock, times(1)).addProjectInfo( false, true);
+  }
+
+  @Test
+  public void whenProjectNamePresent_andHasOnlyMainFiles_shouldAddCorrectInfo() {
+    when(settingsMock.getString("sonar.projectName")).thenReturn("FOO_PROJ");
+    addFileToFileSystem("foo.language", Type.MAIN);
+
+    sensor.execute(tester);
+
```
pavel-mikula-sonarsource, line 146: Let's add log assert with `has MAIN 'true', has TEST 'false'` to show that it can happen and it should be there.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  @Test
+  public void whenProjectNamePresent_andHasBothMainAndTestFiles_shouldAddCorrectInfo() {
+    when(settingsMock.getString("sonar.projectName")).thenReturn("FOO_PROJ");
+    addFileToFileSystem("foo.language", Type.MAIN);
+    addFileToFileSystem("bar.language", Type.TEST);
+
+    sensor.execute(tester);
+
```
pavel-mikula-sonarsource, line 157: Let's add log assert with `has MAIN 'true', has TEST 'true'` to show that it can happen and it should be there.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
pavel-mikula-sonarsource: This one is actually just to keep the test structure regular. The next one below is more interesting.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
```diff
@@ -49,11 +49,13 @@
   private final RoslynDataImporter roslynDataImporter;
   private final DotNetPluginMetadata pluginMetadata;
   private final ReportPathCollector reportPathCollector;
+  private final ProjectTypeCollector projectTypeCollector;
 
-  public DotNetSensor(DotNetPluginMetadata pluginMetadata, ReportPathCollector reportPathCollector,
-    ProtobufDataImporter protobufDataImporter, RoslynDataImporter roslynDataImporter) {
+  public DotNetSensor(DotNetPluginMetadata pluginMetadata, ReportPathCollector reportPathCollector,ProjectTypeCollector projectTypeCollector,
```
pavel-mikula-sonarsource, line 54: _Pinkie nitpick:_ Missing space. I have a macro for Ctrl+S to format & save.
```suggestion
  public DotNetSensor(DotNetPluginMetadata pluginMetadata, ReportPathCollector reportPathCollector, ProjectTypeCollector projectTypeCollector,
```

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/ProjectTypeCollectorTest.java
```diff
@@ -0,0 +1,143 @@
...
+    assertThat(underTest.getSummary()).isEmpty();
+  }
+
+  @Test
+  public void withNoFiles_twoModules_logOne() {
+    ProjectTypeCollector underTest = new ProjectTypeCollector();
+    underTest.addProjectInfo(false, false);
+    underTest.addProjectInfo(false, false);
+
+    assertThat(underTest.getSummary()).hasValue("Found 1 MSBuild projects. 1 with no MAIN or TEST files.");
```
pavel-mikula-sonarsource, line 48: ?
```suggestion
    assertThat(underTest.getSummary()).hasValue("Found 1 MSBuild projects. 1 with no MAIN nor TEST files.");
```

pavel-mikula-sonarsource, line 48: Why `1` when two `false, false` projects were added?

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/ProjectTypeCollector.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+      } else {
+        // no main files
+        if (project.hasTestFiles) {
+          onlyTest++;
+        } else {
+          noFiles++;
+        }
+      }
+    }
```
pavel-mikula-sonarsource, line 62: Isn't this too complicated? I'd say we need `int none = -1, main, test, mixed;` and very simple logic in `addProjectInfo` to increment the correct one.

The `-1` will compensate the top-level module (with appropriate explanation comment).

pavel-mikula-sonarsource (changes requested):
Some improvement suggestions.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
andrei-epure-sonarsource: yes, initially I had 2 tests, one for `cs` and one for `vbnet` - refactoring leftover

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/ProjectTypeCollectorTest.java
andrei-epure-sonarsource: the first is considered global module

sonar-csharp-plugin/src/main/java/org/sonar/plugins/csharp/CSharpPlugin.java
andrei-epure-sonarsource: should be a different PR
I'll do it after

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
pavel-mikula-sonarsource: This one seems missed. 
I'd like to add check to verify that the description language filter is empty.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/ProjectTypeCollectorTest.java
pavel-mikula-sonarsource: Can you add comment as in the other tests please? It's confusing
>// this will be skipped, considered as top-level module

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/ProjectTypeCollectorTest.java
```diff
@@ -0,0 +1,206 @@
...
+  public void mixedProjects_test_none() {
+    ProjectTypeCollector underTest = new ProjectTypeCollector();
+
+    // this will be skipped - considered as top-level module
+    addProjectWithNoFiles(underTest);
+
+    addTestProject(underTest);
+    addTestProject(underTest);
+
+    addProjectWithNoFiles(underTest);
```
pavel-mikula-sonarsource, line 157: Other tests don't have empty line between adding. 
```suggestion
    addTestProject(underTest);
    addTestProject(underTest);
    addProjectWithNoFiles(underTest);
```

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/ProjectTypeCollectorTest.java
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  private void addMainProject(ProjectTypeCollector projectTypeCollector) {
+    projectTypeCollector.addProjectInfo(true, false);
+  }
+
+  private void addMixedProject(ProjectTypeCollector projectTypeCollector) {
+    projectTypeCollector.addProjectInfo(true, true);
+  }
+}
```
pavel-mikula-sonarsource, line 206: Missing EOF EOL

pavel-mikula-sonarsource (approved):
LGTM. Please take a look at those 2 new EOL nitpicks and 2 opened conversations above.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
andrei-epure-sonarsource: sorry, I didn't find a way to do it yesterday, I'll check again today

pavel-mikula-sonarsource: Should be easy like this, right? https://github.com/SonarSource/sonar-dotnet/blob/1bf06b4b6c5aa030a31c90c27e4b2c00a95e9de7/sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/DotNetSensorTest.java#L93-L99

Will you update this in the next PR with sorting?

andrei-epure-sonarsource: I already updated using the mock, I'm not sure there's any benefit in using your suggested approach.

The way I did it ensures that it only calls what we expect it to call - setting the name.

PR #4077, opened 2021-02-19
Remove old ToDo comments

costin-zaharia-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
@@ -169,8 +169,6 @@ namespace SonarAnalyzer.Rules
 
         protected abstract string FileName { get; }
 
-        // ToDo: Remove this hardcoded hack as soon as the following ticket is fixed:
-        // https://jira.sonarsource.com/browse/MMF-672
         private static bool ShouldGenerateMetrics(SyntaxTree tree) =>
             FileExtensionWhitelist.Contains(Path.GetExtension(tree.FilePath));
     }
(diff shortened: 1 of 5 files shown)


analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -169,8 +169,6 @@ protected sealed override void Initialize(SonarAnalysisContext context)
 
         protected abstract string FileName { get; }
 
-        // ToDo: Remove this hardcoded hack as soon as the following ticket is fixed:
```
pavel-mikula-sonarsource, line 172: Do we have a follow up issue for this one? `".cs", ".csx", ".vb"` extension list is now hardcoded and the properties should be available from SQ/SC (not easy, but doable and desirable in the future).

The current hack was introduced in #1701

pavel-mikula-sonarsource (approved):
LGTM, there just one question

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
costin-zaharia-sonarsource: It looks to me that this is here to stay since the [related MMF](https://jira.sonarsource.com/browse/MMF-672) was closed as won't do.

pavel-mikula-sonarsource: The mentioned MMF is not strictly related to what this does. I think we should publish the metrics to whatever extensions are configured on SQ/SC.

PR #4079, opened 2021-02-19
Add UTs for Hotspot and Deprecated rules

### LINKED ISSUE #4034: Add UTs for Hotspot and Deprecated rules
## Context

Hotspot rules should have `WithNotConfigurable` rule configuration.

Non-hotspot rules should not. They should use `AnalysisConfiguration.AlwaysEnabled` if the underlaying base type needs `IAnalyzerConfiguration`.

## Problem

Some non-hotspot rules do have `WithNotConfigurable` in place. As observed on S3011 BypassingAccessibility.

## ToDo

Write a test that will do a cross check with rule type mapping and fail when it doesn't match.

Write a test that will do a cross check for deprecated rules - they should not be part of SonarWay (activated by default)

pavel-mikula-sonarsource:
Fixes #4034

S3011 was already fixed in previous refactoring PR

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
@@ -148,6 +148,15 @@ namespace SonarAnalyzer.UnitTest.Common
             SupportedDiagnostics(AnalyzerLanguage.VisualBasic).Should().OnlyContain(diagnostic => diagnostic.CustomTags.Contains(LanguageNames.VisualBasic));
 
         [TestMethod]
+        public void DeprecatedRules_AreNotInSonarWay()
+        {
+            foreach (var diagnostic in new RuleFinder().AllAnalyzerTypes.SelectMany(SupportedDiagnostics).Where(IsDeprecated))
+            {
+                IsSonarWay(diagnostic).Should().BeFalse($"{diagnostic.Id} is deprecated and should be removed from SonarWay.");
+            }
+        }
+
+        [TestMethod]
         public void AllRules_SonarWayTagPresenceMatchesIsEnabledByDefault()
         {
             var parameterized = new RuleFinder().AllAnalyzerTypes
@@ -159,9 +168,13 @@ namespace SonarAnalyzer.UnitTest.Common
             {
                 if (IsSecurityHotspot(diagnostic))
                 {
-                    // Security hotspots are enabled by default, but they will report issues only
-                    // when their ID is contained in SonarLint.xml
+                    // Security hotspots are enabled by default, but they will report issues only when their ID is contained in SonarLint.xml
+                    // Rule activation is done in DiagnosticDescriptorBuilder.WithNotConfigurable() to prevent rule supression and deactivation.
                     diagnostic.IsEnabledByDefault.Should().BeTrue($"{diagnostic.Id} should be enabled by default");
+                    if (IsDeprecated(diagnostic))
+                    {
+                        System.Diagnostics.Debugger.Break();
+                    }
                 }
                 else if (IsDeprecated(diagnostic))
                 {
@@ -174,7 +187,7 @@ namespace SonarAnalyzer.UnitTest.Common
                     // See https://github.com/SonarSource/sonar-dotnet/issues/1274
                     diagnostic.IsEnabledByDefault.Should().BeFalse($"{diagnostic.Id} has parameters and should be disabled by default");
                 }
-                else if (diagnostic.CustomTags.Contains(DiagnosticDescriptorBuilder.SonarWayTag))
+                else if (IsSonarWay(diagnostic))
                 {
                     diagnostic.IsEnabledByDefault.Should().BeTrue($"{diagnostic.Id} is in SonarWay");
                 }
@@ -220,6 +233,9 @@ namespace SonarAnalyzer.UnitTest.Common
 
         private static IEnumerable<Type> GetCodeFixProviderTypes(IEnumerable<Assembly> assemblies) =>
             assemblies.SelectMany(assembly => assembly.GetTypes()).Where(t => t.IsSubclassOf(typeof(SonarCodeFixProvider)));
+
+        private static bool IsSonarWay(DiagnosticDescriptor diagnostic) =>
+            diagnostic.CustomTags.Contains(DiagnosticDescriptorBuilder.SonarWayTag);
 
         private static bool IsDeprecated(DiagnosticDescriptor diagnostic)
         {
--- a/analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
@@ -38,20 +38,12 @@ namespace SonarAnalyzer.UnitTest.Common
     [TestClass]
     public class RuleTest
     {
-        private static IEnumerable<Type> GetCodeFixProviderTypes(IEnumerable<Assembly> assemblies)
-        {
-            return assemblies
-                .SelectMany(assembly => assembly.GetTypes())
-                .Where(t => t.IsSubclassOf(typeof(SonarCodeFixProvider)));
-        }
-
         [TestMethod]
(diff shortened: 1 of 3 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
```diff
@@ -18,11 +18,14 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+extern alias csharp;
+extern alias vbnet;
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
```
costin-zaharia-sonarsource, line 25: `using System.Collections.Immutable;` is no longer need

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
```diff
@@ -31,26 +34,19 @@
...
-        }
-
         [TestMethod]
         public void DiagnosticAnalyzerHasRuleAttribute()
         {
             foreach (var analyzer in new RuleFinder().AllAnalyzerTypes)
             {
                 var ruleDescriptors = analyzer.GetCustomAttributes<RuleAttribute>();
-
                 ruleDescriptors.Should().NotBeEmpty("RuleAttribute is missing from DiagnosticAnalyzer '{0}'", analyzer.Name);
```
costin-zaharia-sonarsource, line 50: Can be simplified even more:
```
analyzer.GetCustomAttributes<RuleAttribute>().Should().NotBeEmpty("RuleAttribute is missing from DiagnosticAnalyzer '{0}'", analyzer.Name);
```

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
```diff
@@ -122,129 +107,147 @@ public void SonarDiagnosticAnalyzer_IsUsedInAllRules()
...
                 {
                     // Security hotspots are enabled by default, but they will report issues only
                     // when their ID is contained in SonarLint.xml
-                    analyzer.IsEnabledByDefault.Should().BeTrue($"{analyzer.Id} should be enabled by default");
+                    diagnostic.IsEnabledByDefault.Should().BeTrue($"{diagnostic.Id} should be enabled by default");
+                }
+                else if (IsDeprecated(diagnostic))
+                {
+                    // Deprecated rules should be removed from SonarWay
+                    diagnostic.IsEnabledByDefault.Should().BeFalse($"{diagnostic.Id} is deprecated and should be disabled by default (removed from SonarWay)");
```
costin-zaharia-sonarsource, line 172: how do we handle deprecated hotspots? it looks to me that we will not reach this check for them

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
```diff
@@ -122,129 +107,147 @@ public void SonarDiagnosticAnalyzer_IsUsedInAllRules()
...
-                else if (parameterizedAnalyzers.Contains(analyzer))
+                else if (parameterized.Contains(diagnostic))
                 {
                     // Even if a a parametrized rule is in Sonar way profile, it is still disabled by default.
                     // See https://github.com/SonarSource/sonar-dotnet/issues/1274
-                    analyzer.IsEnabledByDefault.Should().BeFalse($"{analyzer.Id} has parameters and should be disabled by default");
+                    diagnostic.IsEnabledByDefault.Should().BeFalse($"{diagnostic.Id} has parameters and should be disabled by default");
                 }
-                else if (isInSonarWay)
+                else if (diagnostic.CustomTags.Contains(DiagnosticDescriptorBuilder.SonarWayTag))
```
costin-zaharia-sonarsource, line 180: Shouldn't this check run for all rules?

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
pavel-mikula-sonarsource: This check goes second, because deprecated hotspots are also active by default. That's not a problem as it's SonarLint.xml responsibility to decide if they should run or not (and they will not run in SL nor NuGet)

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
pavel-mikula-sonarsource: Rules that are not in SonarWay are not activated by default in VS. When using SL-only or NuGet, user needs to explicitly activate them.

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
costin-zaharia-sonarsource: Isn't the case with a deprecated hotspot in `Sonar Way` something we want to avoid?

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
costin-zaharia-sonarsource: 🤦 only now I've seen the comment: `Even if a a parametrized rule is in Sonar way profile, it is still disabled by default.`

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
pavel-mikula-sonarsource: That's a good point, but I'm it needs some more testing and investigation => will do later.

costin-zaharia-sonarsource: I don't understand why. Isn't this PR supposed to do this? 

> Write a test that will do a cross check for deprecated rules - they should not be part of SonarWay (activated by default)

pavel-mikula-sonarsource: I've investigate the problem, added UT for this scenario and few comments. There are different states
`SonarWay` - is in SonarWay or not according to Jira => has SonarWay custom tag
`IsActivatedByDefault` - resx key, is true for SonarWay rules
`IsEnabledByDefault` - diagnostic state, is true for nonparametrized SonarWay rules OR all hotspots (deprecated or not)

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
```diff
@@ -159,9 +168,13 @@ public void AllRules_SonarWayTagPresenceMatchesIsEnabledByDefault()
...
                 if (IsSecurityHotspot(diagnostic))
                 {
-                    // Security hotspots are enabled by default, but they will report issues only
-                    // when their ID is contained in SonarLint.xml
+                    // Security hotspots are enabled by default, but they will report issues only when their ID is contained in SonarLint.xml
+                    // Rule activation is done in DiagnosticDescriptorBuilder.WithNotConfigurable() to prevent rule supression and deactivation.
                     diagnostic.IsEnabledByDefault.Should().BeTrue($"{diagnostic.Id} should be enabled by default");
+                    if (IsDeprecated(diagnostic))
+                    {
+                        System.Diagnostics.Debugger.Break();
```
costin-zaharia-sonarsource, line 176: Is this still needed? it looks to me useful only for debugging

costin-zaharia-sonarsource (changes requested):
Looks good but I think we should remove the `Debugger.Break()` call

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
pavel-mikula-sonarsource: Not at all 🤦‍♂️

pavel-mikula-sonarsource: "Don't use debugger features in production" could be a new rule :)

costin-zaharia-sonarsource (approved):
LGTM!

PR #4080, opened 2021-02-19
Refactor NuGetMetadataFactory

pavel-mikula-sonarsource:
Deobfuscate complicated logic.
Procedural => OOP.
Simplify.
Cleanup

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
@@ -93,7 +93,7 @@ namespace SonarAnalyzer.UnitTest.MetadataReferences
             package.EnsurePackageIsInstalled();
 
             var allowedNuGetLibDirectoriesByPreference = allowedTargetFrameworks.Select((folder, priority) => new { folder, priority });
-            var packageDirectory = package.GetNuGetPackageDirectory();
+            var packageDirectory = package.PackageDirectory();
             LogMessage($"Download package directory: {packageDirectory}");
             if (!Directory.Exists(packageDirectory))
             {
--- a/analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
@@ -36,7 +36,7 @@ namespace SonarAnalyzer.UnitTest.MetadataReferences
 
         private static readonly PackageManager PackageManager = new PackageManager(CreatePackageRepository(), PackagesFolderRelativePath);
 
-        private static readonly string[] AllowedNuGetLibDirectoriesInOrderOfPreference = new string[]
+        private static readonly string[] SortedAllowedDirectories = new string[]
             {
                 "net",
                 "netstandard2.1",
@@ -52,14 +52,14 @@ namespace SonarAnalyzer.UnitTest.MetadataReferences
                 "net40",
                 "net20",
                 "portable-net45",
-                "lib",
+                "lib", // This has to be last, some packages have DLLs directly in "lib" directory
             };
 
         public static IEnumerable<MetadataReference> Create(string packageId, string packageVersion, string runtime, string targetFramework) =>
             Create(new Package(packageId, packageVersion, runtime), new[] { targetFramework });
 
         public static IEnumerable<MetadataReference> Create(string packageId, string packageVersion, string runtime = null) =>
-            Create(new Package(packageId, packageVersion, runtime), AllowedNuGetLibDirectoriesInOrderOfPreference);
+            Create(new Package(packageId, packageVersion, runtime), SortedAllowedDirectories);
 
         public static IEnumerable<MetadataReference> CreateNETStandard21()
         {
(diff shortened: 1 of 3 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -83,47 +83,26 @@ public static IEnumerable<MetadataReference> CreateNETStandard21()
                .ToImmutableArray();
         }
 
-        private static IEnumerable<MetadataReference> Create(Package package, string[] allowedTargetFrameworks)
+        private static IEnumerable<MetadataReference> Create(Package package, string[] allowedDirectories)
```
andrei-epure-sonarsource, line 86: please add a comment documenting the assumption that `allowedDirectiories` should be sorted - I believe `dllsPerDirectory[allowedDirectories.First(x => dllsPerDirectory.ContainsKey(x))];` relies on that

and could also name the parameter "sortedAllowedTargetFrameworkDirectories", explaining in the method comment that the directories are going to be verified one after another in the order they are given in

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -83,47 +83,26 @@ public static IEnumerable<MetadataReference> CreateNETStandard21()
...
-        private static void DumpSelectedGroup(Package package, IGrouping<string, FileInfo> fileGroup)
-        {
-            Console.WriteLine();
-            Console.WriteLine($"Package: {package.Id}");
-            Console.WriteLine($"Version: {package.Version}, chosen targetFramework: {fileGroup.Key}");
-            foreach (var file in fileGroup)
+            var dllsPerDirectory = Directory.GetFiles(packageDirectory, "*.dll", SearchOption.AllDirectories)
+                .GroupBy(x => new FileInfo(x).Directory.Name)
+                .ToDictionary(x => x.Key.Split('+').First(), x => x.AsEnumerable());
+            var dlls = dllsPerDirectory[allowedDirectories.First(x => dllsPerDirectory.ContainsKey(x))];
```
andrei-epure-sonarsource, line 100: I suggest to use `dir` rather than `x` in the lambda because when reading `dllsPerDirectory.ContainsKey(x)` the brain can wonder if `x` is a `dll` or a `directory`

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -83,47 +83,26 @@ public static IEnumerable<MetadataReference> CreateNETStandard21()
...
+            foreach (var file in dlls)
```
andrei-epure-sonarsource, line 101: is it `fileName` or `filePath`?

andrei-epure-sonarsource (changes requested):
Looks great, very nice refactoring.
I only left minor comments for readability.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -36,7 +36,7 @@ public static partial class NuGetMetadataFactory
 
         private static readonly PackageManager PackageManager = new PackageManager(CreatePackageRepository(), PackagesFolderRelativePath);
 
-        private static readonly string[] AllowedNuGetLibDirectoriesInOrderOfPreference = new string[]
+        private static readonly string[] SortedAllowedDirectories = new string[]
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I wanted to make this as short as possible. The long name was difficult to work with. Also check the latest commit [`f99d0e2` (#4080)](https://github.com/SonarSource/sonar-dotnet/pull/4080/commits/f99d0e2c700cce1d2406026ac9cbd8e11fbbae1a) - I've added a clear, meaningful and actionable message to let us work with it easily. Were missing such a message a lot (it always took me some investigation when I hit the "sequence contains no elements" issue)

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
pavel-mikula-sonarsource: Given the DLL situation in the last commit, it's not really a target framework in all cases. `allowedDirectory` is more precise name.

I'll add the sorted comment.

pavel-mikula-sonarsource: Thinking back on the to topic, the `target framework` component was actually a source of confusion in the deobfuscation phase.

pavel-mikula-sonarsource: We can consider renaming `Create(.., targetFramework)` argument to something else, like `dllDirectory` and update the comment accordingly.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
pavel-mikula-sonarsource: I prefer `x` as a convention from mathematical lambda calculus. It's usually λx for single argument. Like this, you don't need to ask yourself if `dir` is a local variable, argument or field. It's whatever the lambda is run for.

(The `x` situation doesn't apply for nested lambda or with multiple arguments)

andrei-epure-sonarsource: maybe we should document this in our coding style

andrei-epure-sonarsource (approved):
LGTM

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -97,7 +98,10 @@ private static IEnumerable<MetadataReference> Create(Package package, string[] a
             var dllsPerDirectory = Directory.GetFiles(packageDirectory, "*.dll", SearchOption.AllDirectories)
                 .GroupBy(x => new FileInfo(x).Directory.Name)
                 .ToDictionary(x => x.Key.Split('+').First(), x => x.AsEnumerable());
-            var dlls = dllsPerDirectory[allowedDirectories.First(x => dllsPerDirectory.ContainsKey(x))];
+            var directory = allowedDirectories.FirstOrDefault(x => dllsPerDirectory.ContainsKey(x))
```
andrei-epure-sonarsource, line 103: this is much better!

PR #4089, opened 2021-02-22
Fix project count logic - filter out global module

### LINKED ISSUE #4054: Report Main and Test projects in DotNetSensor

andrei-epure-sonarsource:
Fixes  #4054 

TL;DR;:
- we now count the MSBuild projects and show statistics - how many are MAIN, how many are TEST etc. Each MSBuild project has an equivalent scanner module.
- the global module is an artificial scanner module which has no files and no MSBuild equivalent.
- the global module has `projectName` and `projectKey` properties like all other module; to filter it out, this PR uses `projectOutPath` which is a property just for "real project" modules.

The IT,  `MultipleProjectsTest#projectTypesInfoIsLogged`, hasn't changed.

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensor.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensor.java
@@ -32,7 +32,7 @@ import org.sonar.api.utils.log.Loggers;
 
 /**
  * This class is a non-global sensor used to count the type of files in the .NET projects (i.e. Scanner modules).
- *
+ * <p>
  * Why is this needed?
  * - the Scanner for MSBuild categorizes projects as MAIN or TEST (see https://github.com/SonarSource/sonar-scanner-msbuild/wiki/Analysis-of-product-projects-vs.-test-projects)
  * - in SQ / SC, users can specify which files should be considered as MAIN (sources) or TEST (test sources) (see https://docs.sonarqube.org/latest/project-administration/narrowing-the-focus/)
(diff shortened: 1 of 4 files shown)


its/src/test/java/com/sonar/it/csharp/MetricsIncludeHeaderCommentTest.java
```diff
@@ -43,6 +43,10 @@
  * This is copy pasted from MetricsTest. It does not re-test everything, it just makes sure
  * that when 'ignoreHeaderComments' is set to False, it counts the header comments as well
  * and it does not modify the LOC metrics.
+ *
+ * <p>
```
andrei-epure-sonarsource, line 47: this is added automatically by IntelliJ when cleaning up, it's probably a javadoc convention

its/src/test/java/com/sonar/it/csharp/MetricsIncludeHeaderCommentTest.java
```diff
@@ -53,40 +57,10 @@
   private static final String FILE = "MetricsTest:foo/Class1.cs";
   private static final int NUMBER_OF_HEADER_COMMENT_LINES = 2;
 
+  // This is where the analysis is done.
   @ClassRule
   public static RuleChain chain = getRuleChain();
 
-  private static RuleChain getRuleChain() {
```
andrei-epure-sonarsource, line 59: I did this change in every file, just moved the private static method at the end

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensor.java
```diff
@@ -32,7 +32,7 @@
 
 /**
  * This class is a non-global sensor used to count the type of files in the .NET projects (i.e. Scanner modules).
- *
+ * <p>
```
andrei-epure-sonarsource, line 35: this is added automatically by IntelliJ, I think it's a javadoc standard

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensor.java
```diff
@@ -66,21 +66,24 @@ public void execute(SensorContext context) {
...
-    Optional<String> projectName = configuration.get("sonar.projectName");
-    if (projectName.isPresent()) {
-      // Note: the top-level module is artificially created by the scanner and will appear here with false-false.
-      LOG.debug("Adding file type information (has MAIN '{}', has TEST '{}') for project '{}' (base dir '{}'). For debug info, see ProjectInfo.xml in '{}'.",
-        hasMainFiles, hasTestFiles, projectName.get(), getValueOrEmpty(configuration.get("sonar.projectBaseDir")), getProjectOutPath(configuration));
+    Optional<String> projectOutPath = getProjectOutPath(configuration);
+    // We filter based on `projectOutPath` to avoid adding the top-level module, which has no files at all (is an artificial module with no MSBuild project equivalent).
+    // The top-level module has the `sonar.projectKey` and `sonar.projectName` properties, but does not have the `projectOutPath` property.
+    if (projectOutPath.isPresent()) {
+      LOG.debug("Adding file type information (has MAIN '{}', has TEST '{}') for project '{}' (project key '{}', base dir '{}'). For debug info, see ProjectInfo.xml in '{}'.",
```
andrei-epure-sonarsource, line 73: I added also the `projectKey` to the DEBUG logs, it may help us with support

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/ProjectTypeCollectorTest.java
```diff
@@ -35,48 +35,22 @@ public void withNoProjects() {
...
-    // If that doesn't happen, it returns empty string.
     assertThat(underTest.getSummary()).isEmpty();
-
-    addMainProject(underTest);
-    addMainProject(underTest);
-    assertThat(underTest.getSummary()).hasValue("Found 2 MSBuild projects. 3 MAIN projects.");
   }
 
   @Test
   public void withNoFiles_twoModules_logOne() {
```
costin-zaharia-sonarsource, line 42: The name looks to be out of sync. Only one project is added.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/ProjectTypeCollectorTest.java
```diff
@@ -35,48 +35,22 @@ public void withNoProjects() {
   public void withNoFiles_skip() {
     ProjectTypeCollector underTest = new ProjectTypeCollector();
```
costin-zaharia-sonarsource, line 36: I would use [sut](https://en.wikipedia.org/wiki/System_under_test) instead of `underTest`. It's short and the convention is well known.

See also the mention from Martin Fowler [here](https://martinfowler.com/articles/mocksArentStubs.html):
> I'll use System Under Test, or rather the abbreviation SUT.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/ProjectTypeCollectorTest.java
```diff
@@ -35,48 +35,22 @@ public void withNoProjects() {
...
 
-    // this will be skipped, considered as top-level module
-    addProjectWithNoFiles(underTest);
```
costin-zaharia-sonarsource, line 39: After this change both `withNoProjects' and `withNoFiles_skip` check the same behavior.

costin-zaharia-sonarsource (changes requested):
Looks good. Just a few minor comments regarding the tests.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/ProjectTypeCollectorTest.java
andrei-epure-sonarsource: yes, I had a discussion with Pavel some time ago on this, and since I'm using more explicit names.
Let's discuss it in the "how we work" and align with everyone.

costin-zaharia-sonarsource (approved):
LGTM!

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensor.java
costin-zaharia-sonarsource: It's for paragraphs separation. See: https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html

PR #4091, opened 2021-02-23
Display warnings in SQ/SC UI when only test projects are analyzed

andrei-epure-sonarsource:
Fixes #4063

Also some other minor improvements and fixing another found issue.

PR #4092, opened 2021-02-23
add IT for querying AnalysisWarnings from SQ

andrei-epure-sonarsource:
add IT and address leftover UT comment from #4091 review

### DIFF
--- a/its/src/test/java/com/sonar/it/shared/TestUtils.java
+++ b/its/src/test/java/com/sonar/it/shared/TestUtils.java
@@ -21,6 +21,7 @@ package com.sonar.it.shared;
 
 import com.sonar.orchestrator.Orchestrator;
 import com.sonar.orchestrator.build.Build;
+import com.sonar.orchestrator.build.BuildResult;
 import com.sonar.orchestrator.build.ScannerForMSBuild;
 import com.sonar.orchestrator.http.HttpMethod;
 import com.sonar.orchestrator.locator.FileLocation;
@@ -39,12 +40,14 @@ import java.time.format.DateTimeFormatter;
 import java.time.temporal.ChronoUnit;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.junit.rules.TemporaryFolder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.sonarqube.ws.Ce;
 import org.sonarqube.ws.Components;
 import org.sonarqube.ws.Issues;
 import org.sonarqube.ws.Measures;
@@ -52,6 +55,7 @@ import org.sonarqube.ws.Measures.Measure;
 import org.sonarqube.ws.client.HttpConnector;
 import org.sonarqube.ws.client.WsClient;
 import org.sonarqube.ws.client.WsClientFactories;
+import org.sonarqube.ws.client.ce.TaskRequest;
 import org.sonarqube.ws.client.components.ShowRequest;
 import org.sonarqube.ws.client.measures.ComponentRequest;
 
@@ -63,6 +67,14 @@ public class TestUtils {
 
   final private static Logger LOG = LoggerFactory.getLogger(TestUtils.class);
   private static final String MSBUILD_PATH = "MSBUILD_PATH";
+
+  public static Ce.Task getAnalysisWarningsTask(Orchestrator orchestrator, BuildResult buildResult) {
+    String taskId = extractCeTaskId(buildResult);
+    return newWsClient(orchestrator)
+      .ce()
+      .task(new TaskRequest().setId(taskId).setAdditionalFields(Collections.singletonList("warnings")))
+      .getTask();
+  }
 
   public static Location getPluginLocation(String pluginName) {
     Location pluginLocation;
@@ -245,4 +257,17 @@ public class TestUtils {
     }
   }
 
+  private static String extractCeTaskId(BuildResult buildResult) {
+    List<String> taskIds = extractCeTaskIds(buildResult);
+    if (taskIds.size() != 1) {
+      throw new IllegalStateException("More than one task id retrieved from logs.");
+    }
+    return taskIds.iterator().next();
+  }
+
+  private static List<String> extractCeTaskIds(BuildResult buildResult) {
+    return buildResult.getLogsLines(s -> s.contains("More about the report processing at")).stream()
+      .map(s -> s.substring(s.length() - 20))
+      .collect(Collectors.toList());
+  }
 }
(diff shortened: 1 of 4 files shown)


sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
```diff
@@ -104,8 +111,6 @@ public void whenProjectOutPathsNotPresent_shouldNotAddProjectInfo_shouldNotLog()
   @Test
   public void whenProjectOutPaths_returnsNull_shouldNotAddProjectInfo_shouldNotLog() {
     when(settingsMock.getStringArray("sonar.LANG_KEY.analyzer.projectOutPaths")).thenReturn(null);
-    addFileToFileSystem("foo.language", Type.MAIN);
```
andrei-epure-sonarsource, line 107: I removed these lines because files are not needed to log - if no files are present, it will log FALSE,FALSE as long as the `projectOutPaths` property is set

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
```diff
@@ -86,14 +86,21 @@ public void whenProjectOutPathsNotPresent_shouldNotAddProjectInfo_shouldNotLog()
...
-    // note that below it's the singular - Path
-    when(settingsMock.getString("sonar.cs.analyzer.projectOutPath")).thenReturn("CS OutPath");
-    when(settingsMock.getString("sonar.vbnet.analyzer.projectOutPath")).thenReturn("VB OutPath");
-    // should query the `getStringArray` method, not the `getString` method like below
-    when(settingsMock.getString("sonar.cs.analyzer.projectOutPaths")).thenReturn("CS OutPaths");
-    when(settingsMock.getString("sonar.vbnet.analyzer.projectOutPaths")).thenReturn("VB OutPaths");
-    addFileToFileSystem("foo.language", Type.MAIN);
-    addFileToFileSystem("bar.language", Type.TEST);
+    // note that below it's the singular in both calls - "projectOutputPath"
+    when(settingsMock.getString("sonar.LANG_KEY.analyzer.projectOutPath")).thenReturn("foo\\.sonarqube\\out\\0");
```
pavel-mikula-sonarsource, line 90: What's this `getString` mock for, when the next test claims that it should not be called. Can this line be removed?

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
```diff
@@ -86,14 +86,21 @@ public void whenProjectOutPathsNotPresent_shouldNotAddProjectInfo_shouldNotLog()
...
+
+    assertThat(logTester.logs()).isEmpty();
+    verifyZeroInteractions(projectTypeCollectorMock);
+  }
+
+  @Test
+  public void projectOutPaths_shouldNotBeRetrievedWithGetString() {
+    // the `projectOutPaths` property is multi-value and should be queried via getStringArray()
+    // this test ensures that only `getStringArray` is used, and not `getString`
+    when(settingsMock.getString("sonar.LANG_KEY.analyzer.projectOutPaths")).thenReturn("foo\\.sonarqube\\out\\0");
```
pavel-mikula-sonarsource, line 103: If I understand the mechanics of this test, then 
* getStringArray is empty, resulting in empty log
* getString is not empty, if somebody reads that, he'll get nonempty log
Am I missing something? That looks very indirect and unclear.

We should be more straight forward with this test. By not mocking the value and measuring it's consequences, but rather by failing if somebody queries that value instead.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
```diff
@@ -86,14 +86,21 @@ public void whenProjectOutPathsNotPresent_shouldNotAddProjectInfo_shouldNotLog()
...
-    when(settingsMock.getString("sonar.cs.analyzer.projectOutPath")).thenReturn("CS OutPath");
-    when(settingsMock.getString("sonar.vbnet.analyzer.projectOutPath")).thenReturn("VB OutPath");
-    // should query the `getStringArray` method, not the `getString` method like below
-    when(settingsMock.getString("sonar.cs.analyzer.projectOutPaths")).thenReturn("CS OutPaths");
-    when(settingsMock.getString("sonar.vbnet.analyzer.projectOutPaths")).thenReturn("VB OutPaths");
-    addFileToFileSystem("foo.language", Type.MAIN);
-    addFileToFileSystem("bar.language", Type.TEST);
+    // note that below it's the singular in both calls - "projectOutputPath"
+    when(settingsMock.getString("sonar.LANG_KEY.analyzer.projectOutPath")).thenReturn("foo\\.sonarqube\\out\\0");
+    when(settingsMock.getStringArray("sonar.LANG_KEY.analyzer.projectOutPath")).thenReturn(arrayOf("foo\\.sonarqube\\out\\0"));
```
pavel-mikula-sonarsource, line 91: What's the purpose of this singular form? To avoid typos in property name? Then we should mock to explicitly fail if somebody queries this property, shouldn't we?

If that's not possible (in a reasonable easy way), we should add more comments that this singular property should not be read and if the test reads and uses it, it would lead to a logged message and failed test - this mechanics is hard to see (if I even got it right :) )

pavel-mikula-sonarsource (changes requested):
The warning IT looks good, I have some doubts about the FileTypeSensorTest (not sure if I completely got it => probably confusing test)

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
pavel-mikula-sonarsource: I probably got it now. It was hard to see.

its/src/test/java/com/sonar/it/shared/TestUtils.java
```diff
@@ -39,19 +40,22 @@
...
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.junit.rules.TemporaryFolder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.sonarqube.ws.Ce;
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: > Compute Engine Server in charge of processing code analysis reports and saving them in the SonarQube Database
([docs](https://docs.sonarqube.org/latest/architecture/architecture-integration/))

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
andrei-epure-sonarsource: very good point, I've made the tests way too complicated

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/FileTypeSensorTest.java
andrei-epure-sonarsource: I've merged this and the above in one

pavel-mikula-sonarsource (approved):
LGTM! The `never()` is pretty clear about it's intention now :)

PR #4094, opened 2021-02-24
GitHub workflow: Milestone issue

pavel-mikula-sonarsource:
To work on a sprint, we need to add issues to ToDo column and set Milesones on all of them.

New intended process will be just adding milestones, as we have to do it anyway and choose the correct one. This action will automatically add the card to ToDo column for us.

Sprint issues without milestone needs to be added manually, as before.

### DIFF
--- a/.github/workflows/MilestoneIssue.yml
+++ b/.github/workflows/MilestoneIssue.yml
@@ -1,0 +1,22 @@
+name: Milesone Issue
+
+on:
+  issues:
+    types: ["milestoned"]
+
+jobs:
+  create_card:
+    name: Create card
+    runs-on: ubuntu-latest
+
+    steps:
+      # https://github.com/actions/github-script
+      - uses: actions/github-script@v3
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const TODO_COLUMN = 4971951;
+            console.log("Creating Issue card");
+            // Action will fail if a card already exists for this issue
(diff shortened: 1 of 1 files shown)


.github/workflows/MilestoneIssue.yml
```diff
@@ -0,0 +1,22 @@
...
+
+    steps:
+      # https://github.com/actions/github-script
+      - uses: actions/github-script@v3
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const TODO_COLUMN = 4971951;
+            console.log("Creating Issue card");
+            // Action will fail if a card already exists for this issue
```
pavel-mikula-sonarsource, line 20: It's too complicated to find a card with current API and there's no harm done => ROI would be bad now.

From online discussions, new API in alpha testing phase should provide better way of working with projects in the future.

andrei-epure-sonarsource (commented):
Thinking out loud:
- On this action, I would wait until we discuss in "how we work" if we want to keep one kanban or not.
- Our current kanban is quite "polluted" with out-of-sprint work (community PRs, minor improvements) and is hard to use it for getting an overview of the sprint, so I suggest to take into consideration using one kanban per sprint (and close it after).
- However, I now realize that we won't be able to use automation, so it would be better to keep this kanban which is automated, for sprint work, and a differnt kanban for out-of-sprint work

andrei-epure-sonarsource (approved):

PR #4099, opened 2021-02-26
Update rule specs; S2486 becomes a code smell

andrei-epure-sonarsource:
Most updates are probably a side-effect of the sprint on RSPECv2, where they cleaned a lot of formatting problems.


costin-zaharia-sonarsource (approved):
LGTM!

PR #4103, opened 2021-03-01
OutOfMemoryError possible using wildcard pattern with alternative path separators

### LINKED ISSUE #3852: OutOfMemoryError possible using wildcard pattern with alternative path separators
### Description

If you provide a wildcard pattern on Windows with forward slashes, the analyzer may search too high in the directory structure, eventually causing an OutOfMemoryError. It should support paths containing any acceptable path separator, not just the default.

### Repro steps

On Windows, provide an analyzer property such as:

> sonar.cs.vstest.reportsPaths=$(Agent.TempDirectory)/**/*.trx

Note that my example is on an Azure Pipelines self-hosted agent, so $(Agent.TempDirectory) is similar to `C:\agent_work9`.  

During analysis, you see the following DEBUG output:

```
18:40:23.959 INFO: Sensor C# Unit Test Results Import [csharp]
18:40:23.959 DEBUG: Pattern matcher extracted prefix/absolute path 'C:\agent_work9' from the given pattern 'C:\agent_work9\_temp/**/*.trx'.
18:40:23.959 DEBUG: Gathering files for wildcardPattern '_temp/**/*.trx'.
18:52:34.401 INFO: ------------------------------------------------------------------------
18:52:34.401 INFO: EXECUTION FAILURE
18:52:34.401 INFO: ------------------------------------------------------------------------
18:52:34.401 INFO: Total time: 12:18.950s
18:52:34.448 INFO: Final Memory: 9M/895M
18:52:34.448 INFO: ------------------------------------------------------------------------
##[error]18:52:34.448 ERROR: Error during SonarScanner execution
##[debug]Processed: ##vso[task.logissue type=error;]18:52:34.448 ERROR: Error during SonarScanner execution
18:52:34.448 ERROR: Error during SonarScanner execution
##[error]java.lang.OutOfMemoryError: Java heap space
	at java.io.WinNTFileSystem.resolve(Unknown Source)
	at java.io.File.<init>(Unknown Source)
	at java.io.File.listFiles(Unknown Source)
	at org.sonar.plugins.dotnet.tests.WildcardPatternFileProvider.listFiles(WildcardPatternFileProvider.java:140)
```

Note that the file search starts at `C:\agent_work9` and not `C:\agent_work9\_temp`, which causes it to search through every build ever conducted by this agent, including ancillary tools and packages, etc.

The root cause for this is that the [WildcardPatternFileProvider](https://github.com/SonarSource/sonar-dotnet/blob/8.15.0.24505/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/WildcardPatternFileProvider.java) class constructor accepts a `directorySeparator` parameter, which is only being provided as the default path separator (File.separator). Perhaps it should take [the same approach as .NET](https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats#canonicalize-separators) to accept both forward and back slashes.

### Known workarounds

Switch to the default Windows path separators:

> sonar.cs.vstest.reportsPaths=$(Agent.TempDirectory)/**/*.trx

And it will search at the correct level:
```
22:10:40.179 DEBUG: Pattern matcher extracted prefix/absolute path 'C:\agent_work1\_temp' from the given pattern 'C:\agent_work1\_temp\**\*.trx'.
```

### Related information

* SonarScanner for AzurePipelines 4.17.0
* Azure Pipelines Windows self-hosted agent
* C#
* netcoreapp3.1

csaba-sagi-sonarsource:
Fixes #3852

### DIFF
--- a/its/projects/MultipleProjects/WildcardPatternTest/opencover.xml
+++ b/its/projects/MultipleProjects/WildcardPatternTest/opencover.xml
@@ -1,0 +1,1520 @@
+<?xml version="1.0" encoding="utf-8"?>
+<CoverageSession xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Version="4.7.922.0">
+  <Summary numSequencePoints="61" visitedSequencePoints="47" numBranchPoints="34" visitedBranchPoints="22" sequenceCoverage="77.05" branchCoverage="64.71" maxCyclomaticComplexity="5" minCyclomaticComplexity="1" maxCrapScore="8.12" minCrapScore="1" visitedClasses="5" numClasses="7" visitedMethods="15" numMethods="22" />
+  <Modules>
+    <Module skippedDueTo="Filter" hash="19-A7-6E-09-56-92-7B-53-8D-DF-A8-1D-B0-92-BB-59-0C-61-4F-3C">
+      <ModulePath>C:\Windows\Microsoft.Net\assembly\GAC_32\mscorlib\v4.0_4.0.0.0__b77a5c561934e089\mscorlib.dll</ModulePath>
+      <ModuleTime>2020-01-09T19:46:54Z</ModuleTime>
+      <ModuleName>mscorlib</ModuleName>
+      <Classes />
+    </Module>
+    <Module skippedDueTo="MissingPdb" hash="91-C0-5F-E7-53-2D-F3-8B-30-38-BB-D8-3E-7F-F7-5B-7B-62-A1-36">
+      <ModulePath>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe</ModulePath>
+      <ModuleTime>2020-07-08T09:31:49.3780144Z</ModuleTime>
+      <ModuleName>vstest.console</ModuleName>
+      <Classes />
+    </Module>
+    <Module skippedDueTo="MissingPdb" hash="AD-4B-E4-78-B3-8D-E2-72-41-D6-0B-CE-66-65-9D-B9-A9-0B-8B-91">
+      <ModulePath>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\Extensions\TestPlatform\vstest.console.exe</ModulePath>
+      <ModuleTime>2020-07-08T09:31:50.868765Z</ModuleTime>
+      <ModuleName>vstest.console</ModuleName>
+      <Classes />
+    </Module>
+    <Module skippedDueTo="MissingPdb" hash="73-FC-22-E9-84-9B-8B-39-22-53-38-2D-5E-C8-41-5C-6B-22-F6-16">
+      <ModulePath>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\Extensions\TestPlatform\Microsoft.TestPlatform.CoreUtilities.dll</ModulePath>
+      <ModuleTime>2020-07-08T09:31:50.4228309Z</ModuleTime>
+      <ModuleName>Microsoft.TestPlatform.CoreUtilities</ModuleName>
+      <Classes />
+    </Module>
+    <Module skippedDueTo="Filter" hash="4A-9D-81-85-DC-CF-74-ED-91-AC-6F-7B-37-3D-3E-0B-36-5C-56-66">
+      <ModulePath>C:\Windows\Microsoft.Net\assembly\GAC_MSIL\System\v4.0_4.0.0.0__b77a5c561934e089\System.dll</ModulePath>
+      <ModuleTime>2019-07-24T02:58:28Z</ModuleTime>
+      <ModuleName>System</ModuleName>
+      <Classes />
+    </Module>
+    <Module skippedDueTo="MissingPdb" hash="EE-31-16-69-AE-0E-44-82-2F-77-53-E9-80-6A-BB-13-D4-31-24-DC">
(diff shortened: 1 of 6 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/WildcardPatternFileProvider.java
```diff
@@ -43,16 +43,14 @@
...
-  public WildcardPatternFileProvider(File baseDir, String directorySeparator) {
+  public WildcardPatternFileProvider(File baseDir) {
     this.baseDir = baseDir;
-    this.directorySeparator = directorySeparator;
   }
 
   Set<File> listFiles(String pattern) {
 
-    List<String> elements = Arrays.asList(pattern.split(Pattern.quote(directorySeparator)));
+    List<String> elements = Arrays.asList(pattern.split("["+ Pattern.quote(File.separator) +"/]"));
```
andrei-epure-sonarsource, line 53: can the `"["+ Pattern.quote(File.separator) +"/]"` be in a static final field? to avoid doing string concatenations at each call?

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/WildcardPatternFileProviderTest.java
```diff
@@ -236,6 +236,17 @@ public void should_fail_with_parent_folder_access_after_wildcard() {
     listFiles(path("*", "..", "foo.txt"), tmp.getRoot());
   }
 
+  @Test
+  public void given_pattern_with_windows_folder_separator_listFiles_detects_wrong_pattern(){
```
andrei-epure-sonarsource, line 240: the pattern is with mixed folder separators, right? Because `tmpRoot` and `FileSeparator` will be Windows, and `_temp/**/*.trx` are UNIX?

and rather than "detects", it should be "supports", because from the test I cannot see the detection itself (there's no message "Pattern detected")

its/src/test/java/com/sonar/it/csharp/CoverageTest.java
```diff
@@ -101,6 +101,18 @@ public void open_cover_on_MultipleProjects() throws Exception {
     assertCoverageMetrics("MultipleProjects:SecondProject/FirstClass.cs", 11, 3, 10, 4);
   }
 
+  @Test
+  public void open_cover_on_MultipleProjects_with_wildcardPattern() throws Exception {
```
andrei-epure-sonarsource, line 105: maybe improve and mention `with_UnixWildcardPattern` ?

its/src/test/java/com/sonar/it/csharp/CoverageTest.java
```diff
@@ -101,6 +101,18 @@ public void open_cover_on_MultipleProjects() throws Exception {
...
+    BuildResult buildResult = analyzeMultipleProjectsTestProject("sonar.cs.opencover.reportsPaths", "/*/opencover.xml");
+
+    // The original opencover is moved to a subfolder and parts of it are removed and that is how we know the correct one is matched.
```
andrei-epure-sonarsource, line 108: nice!

its/projects/MultipleProjects/WildcardPatternTest/opencover.xml
```diff
@@ -0,0 +1,1520 @@
+<?xml version="1.0" encoding="utf-8"?>
```
andrei-epure-sonarsource, line 1: maybe mention in a comment how you modified this file, and which is the source? 

You'll need to put the comment after the `<?xml` tag for it to be valid XML, as far as I remember

see [this example](https://github.com/SonarSource/sonar-dotnet/blob/master/its/projects/CoverageWithDeterministicSourcePaths/opencover.xml#L2)

andrei-epure-sonarsource (changes requested):
nice, I have some suggestions

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/WildcardPatternFileProviderTest.java
```diff
@@ -237,7 +237,7 @@ public void should_fail_with_parent_folder_access_after_wildcard() {
   }
 
   @Test
-  public void given_pattern_with_windows_folder_separator_listFiles_detects_wrong_pattern(){
+  public void given_pattern_with_mixed_folder_separator_and_UnixWildcard_listFiles_supports_pattern(){
```
andrei-epure-sonarsource, line 240: I think you copy pasted the UnixWildcard idea from the comment I made on `CoverageTest.java`...

if you put "mixed_folder_separator", you don't need the `UnixWildcard`, right?

andrei-epure-sonarsource (approved):
LGTM, please note you haven't addressed my initial comment on `CoverageTest.java` , and I left another one on

its/src/test/java/com/sonar/it/csharp/CoverageTest.java
andrei-epure-sonarsource: this hasn't been addressed

andrei-epure-sonarsource:
Also, please check the code smell as it's related to the performance of the regex: https://sonarcloud.io/project/issues?id=org.sonarsource.dotnet%3Asonar-dotnet&open=AXf4EY9bL2-FIDEPQJ15&pullRequest=4103&resolved=false&types=CODE_SMELL

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/WildcardPatternFileProviderTest.java
csaba-sagi-sonarsource: I accidentally merged two of your comments :)

andrei-epure-sonarsource:
UTs are failing due to the problem mentioned in #4115

PR #4105, opened 2021-03-02
[DO NOT MERGE]File format proposal

pavel-mikula-sonarsource:
This should not be merged!

Let's discuss file format to communicate from S4MSB -> Analyzer.

Do we need more fields?

### DIFF
--- a/SonarProjectConfig.xml
+++ b/SonarProjectConfig.xml
@@ -1,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<SonarProjectConfig xmlns="http://www.sonarsource.com/msbuild/analyzer/2021/1">
+  <AnalysisConfigPath>c:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\conf\SonarQubeAnalysisConfig.xml</AnalysisConfigPath>
+  <ProjectPath>C:\Projects\1h with Hacker\AwesomeBankWeb\AwesomeBankWeb.CSharp\AwesomeBankWeb.CSharp.csproj</ProjectPath>
+  <OutputDir>C:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\out\0</OutputDir>
+  <TestProject>false</TestProject>
+  <TargetFramework>netcoreapp3.1</TargetFramework>
+</SonarProjectConfig>
--- a/SonarProjectConfig.xml
+++ b/SonarProjectConfig.xml
@@ -4,6 +4,6 @@
   <ProjectPath>C:\Projects\1h with Hacker\AwesomeBankWeb\AwesomeBankWeb.CSharp\AwesomeBankWeb.CSharp.csproj</ProjectPath>
   <FilesToAnalyzePath>c:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\conf\0\FilesToAnalyze.txt</FilesToAnalyzePath>
   <OutputDir>C:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\out\0</OutputDir>
-  <TestProject>false</TestProject>
+  <ProjectType>Product</ProjectType>
   <TargetFramework>netcoreapp3.1</TargetFramework>
 </SonarProjectConfig>
--- a/SonarProjectConfig.xml
+++ b/SonarProjectConfig.xml
@@ -2,6 +2,7 @@
 <SonarProjectConfig xmlns="http://www.sonarsource.com/msbuild/analyzer/2021/1">
   <AnalysisConfigPath>c:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\conf\SonarQubeAnalysisConfig.xml</AnalysisConfigPath>
   <ProjectPath>C:\Projects\1h with Hacker\AwesomeBankWeb\AwesomeBankWeb.CSharp\AwesomeBankWeb.CSharp.csproj</ProjectPath>
+  <FilesToAnalyzePath>c:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\conf\0\FilesToAnalyze.txt</FilesToAnalyzePath>
   <OutputDir>C:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\out\0</OutputDir>
   <TestProject>false</TestProject>
   <TargetFramework>netcoreapp3.1</TargetFramework>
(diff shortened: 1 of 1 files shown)


SonarProjectConfig.xml
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
pavel-mikula-sonarsource, line 1: Filename itself is also to be reviewed.
`Sonar` - because additional files are passed to all analyzers
`Project` - because it's project-level context
`Config` - because it configures behavior of the analyzer on input side

SonarProjectConfig.xml
```diff
@@ -0,0 +1,8 @@
...
+<SonarProjectConfig xmlns="http://www.sonarsource.com/msbuild/analyzer/2021/1">
```
pavel-mikula-sonarsource, line 2: `SonarProjectConfig` - same as file name to be unique, even when it has dedicated xmlns

SonarProjectConfig.xml
```diff
@@ -0,0 +1,8 @@
...
+  <AnalysisConfigPath>c:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\conf\SonarQubeAnalysisConfig.xml</AnalysisConfigPath>
+  <ProjectPath>C:\Projects\1h with Hacker\AwesomeBankWeb\AwesomeBankWeb.CSharp\AwesomeBankWeb.CSharp.csproj</ProjectPath>
+  <OutputDir>C:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\out\0</OutputDir>
+  <TestProject>false</TestProject>
+  <TargetFramework>netcoreapp3.1</TargetFramework>
```
pavel-mikula-sonarsource, line 7: Multitarget projects will generate separate builds with separate conf/0 and out/0 directories.

SonarProjectConfig.xml
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<SonarProjectConfig xmlns="http://www.sonarsource.com/msbuild/analyzer/2021/1">
+  <AnalysisConfigPath>c:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\conf\SonarQubeAnalysisConfig.xml</AnalysisConfigPath>
+  <ProjectPath>C:\Projects\1h with Hacker\AwesomeBankWeb\AwesomeBankWeb.CSharp\AwesomeBankWeb.CSharp.csproj</ProjectPath>
+  <OutputDir>C:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\out\0</OutputDir>
+  <TestProject>false</TestProject>
```
pavel-mikula-sonarsource, line 6: This reflects `SonarQubeTestProject` property, that will be renamed to `SonarTestProject`. 

I'm avoiding `Sonar` prefix as everything in this `SonarProjectConfig` is `Sonar`.

SonarProjectConfig.xml
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<SonarProjectConfig xmlns="http://www.sonarsource.com/msbuild/analyzer/2021/1">
+  <AnalysisConfigPath>c:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\conf\SonarQubeAnalysisConfig.xml</AnalysisConfigPath>
+  <ProjectPath>C:\Projects\1h with Hacker\AwesomeBankWeb\AwesomeBankWeb.CSharp\AwesomeBankWeb.CSharp.csproj</ProjectPath>
+  <OutputDir>C:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\out\0</OutputDir>
```
pavel-mikula-sonarsource, line 5: Old name: `ProjectOutFolderPath.txt`
Original from `SonarQubeAnalysisConfig.xml`: `SonarOutputDir`

I'm avoiding `Sonar` prefix as everything in this `SonarProjectConfig` is `Sonar`.
I'm avoiding `Project` prefix as everything in this `SonarProjectConfig` is `Project`-scoped with notable exception of `ProjectPath` as it's not a prefix but a descriptor.

csaba-sagi-sonarsource (approved):
LGTM

SonarProjectConfig.xml
andrei-epure-sonarsource: where will this be stored under `C:\Projects\1h with Hacker\AwesomeBankWeb\.sonarqube\out\0`? if yes, it makes sense.

however, there will also be `ProjectInfo.xml` which is a "SonarProjectConfig" like file , too

could you name it `SonarAnalyzersProjectConfig`, to make it explicit that only analyzers will use it?

SonarProjectConfig.xml
andrei-epure-sonarsource: can you use the same terminology as used in the Analyzer?
Like `WorkDirectoryBasePath` ? (see in [UtilityAnalyzerBase](https://github.com/SonarSource/sonar-dotnet/blob/96ad8b26b3c9bce808b747bbac9d3be07dd6940e/analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs#L71))

Or AnalyzerWorkDir in the Java SQ Plugin code - [AbstractPropertyDefinitions](https://github.com/SonarSource/sonar-dotnet/blob/3135cc639f2024e5e6f8a836e2afbee51cdc204e/sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/AbstractPropertyDefinitions.java#L168)

andrei-epure-sonarsource (changes requested):

SonarProjectConfig.xml
andrei-epure-sonarsource: or even just AnalyzersProjectConfig, because everything under the `.sonarqube` folder is "Sonar"

pavel-mikula-sonarsource: No, this will be under `.sonarqube\conf\0` folder.
All additional files are passed to all analyzers => we need "Sonar" in the filename.

SonarProjectConfig.xml
pavel-mikula-sonarsource: I'll be happy to challenge and rename `WorkDirectoryBasePath` on the way, as it's not a good name 😄 

What's actually work directory? The current directory that the Scanner was run from? Current directory that the build was run from? Or is it `project` specific `output` `dir` from `config` AKA `(Sonar)ProjectConfig.OutputDir` 🙂

andrei-epure-sonarsource: I'm fine as long as we agree on a common vocabulary to use

Output dir is also `<PROJECT>\SastTest2\bin\Debug\netcoreapp2.2`, where the `Sast2.dll.RoslynCA.json` is output, but that's used as `sonar.cs.roslyn.reportFilePath`

Given we use `sonar.cs.analyzer.projectOutPath` - can it be `ProjectOutPath` and use this everywhere - Scanner, Config files, analyzer and SQ plugin?

pavel-mikula-sonarsource: `reportFilePath` has specific and quite accurate meaning.

`projectOutPath` is close enough to discuss. I'm generally avoiding `Project` prefix. It's aligned with other `*Path` fields in the XML. So `OutPath` it is?

andrei-epure-sonarsource: The problem is that `out` is the name of the folder. `out\0`  is actually the `ProjectOutPath`.

I'm fine with `OutPath` as well, as the config is per-project.

andrei-epure-sonarsource (approved):
LGTM

PR #4107, opened 2021-03-02
Update message and example for S3431

csaba-sagi-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/ExpectedExceptionAttributeShouldNotBeUsed.cs
```diff
@@ -33,40 +33,25 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class ExpectedExceptionAttributeShouldNotBeUsed : SonarDiagnosticAnalyzer
     {
-        internal const string DiagnosticId = "S3431";
-        private const string MessageFormat = "Replace the 'ExpectedException' attribute with a try/catch block.";
+        private const string DiagnosticId = "S3431";
+        private const string MessageFormat = "Replace the 'ExpectedException' attribute with a throw assertion or a try/catch block.";
```
andrei-epure-sonarsource, line 37: could you please add a test with the new compliant case , to make sure we don't raise on it and also to document the behavior in the tests?

andrei-epure-sonarsource (changes requested):

andrei-epure-sonarsource (approved):
LGTM!

PR #4108, opened 2021-03-02
S3216 RSPEC update

csaba-sagi-sonarsource:
As it can be seen in the following [blog post](https://devblogs.microsoft.com/dotnet/configureawait-faq/), ConfigureAwait(false) is not necessary in most cases in .Net Core as  there is no SynchronizationContext in most cases.
The rule was already raising only for .Net Framework but that was not mentioned in the rule description.


pavel-mikula-sonarsource (approved):
LGTM. 

Rspec update follows implementation of #2588

PR #4109, opened 2021-03-03
Use contains instead of containsExactly

costin-zaharia-sonarsource:
- additional warnings can be logged on the build machine

### DIFF
--- a/its/src/test/java/com/sonar/it/csharp/DoNotAnalyzeTestFilesTest.java
+++ b/its/src/test/java/com/sonar/it/csharp/DoNotAnalyzeTestFilesTest.java
@@ -54,7 +54,7 @@ public class DoNotAnalyzeTestFilesTest {
     assertThat(getMeasureAsInt(CSHARP_ONLY_TEST_PROJECT, "ncloc")).isNull();
 
     assertThat(buildResult.getLogsLines(l -> l.contains("WARN")))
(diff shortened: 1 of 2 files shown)


its/src/test/java/com/sonar/it/csharp/DoNotAnalyzeTestFilesTest.java
```diff
@@ -54,7 +54,7 @@ public void with_csharp_only_test_should_not_populate_metrics() throws Exception
     assertThat(getMeasureAsInt(CSHARP_ONLY_TEST_PROJECT, "ncloc")).isNull();
 
     assertThat(buildResult.getLogsLines(l -> l.contains("WARN")))
-      .containsExactly("WARN: This C# sensor will be skipped, because the current solution contains only TEST files and no MAIN files. " +
+      .contains("WARN: This C# sensor will be skipped, because the current solution contains only TEST files and no MAIN files. " +
```
costin-zaharia-sonarsource, line 57: Additional warnings, outside of our control, can be logged by the scanner. 

e.g. if `CodeCoverage.exe` is missing from the build machine (we don't have visual studio enterprise installed on the self hosted build machines).

andrei-epure-sonarsource (approved):
LGTM, this is needed for #4088 not to fail

PR #4115, opened 2021-03-03
Clean workspace for java steps

costin-zaharia-sonarsource:
if the workspace is not clean, the plugin is not generated correctly and the size is bigger than the threshold.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -251,6 +251,8 @@ stages:
 
     - job: runIntegrationTestsJob
       displayName: '.NET ITs'
+      workspace:
+        clean: all
       steps:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -251,6 +251,8 @@ stages:
 
     - job: runIntegrationTestsJob
       displayName: '.NET ITs'
+      workspace:
+        clean: all
```
costin-zaharia-sonarsource, line 255: For .Net ITs we also need to clean up in order to refresh the expected and actual folder.

andrei-epure-sonarsource (approved):
LGTM

Issue #4096, opened 2021-02-25
Fix S1172 FP: Parameter used in catch block


RobertSimonsSMA:
### Description

A S1172 warning is shown for a parameter used in a catch block if there is an unconditional return inside a using block inside the try clause

### Repro steps

In this code, there is a warning shown the the parameter errorCode is unused:
```
        private int UseAndReturn(int errorCode) {
            try {
                using (var textWriter = new StreamWriter("TestOutput.txt")) {
                    textWriter.Write("There is an errorcode");

                    return 0;
                }
            } catch (Exception) {
                return errorCode;
            }
        }

```

In contrast, in this code - that does functionally the same - the warning is not raised:
```
        private int TryAndReturn(int errorCode) {
            try {
                var textWriter = new StreamWriter("TestOutput.txt");
                try {
                    textWriter.Write("There is a code");

                    return 0;
                } finally {
                    textWriter.Dispose();
                }
            } catch (Exception) {
                return errorCode;
            }
        }

```

### Expected behavior

Both code snippets should not raise the unused parameter warning.

### Actual behavior

In the first snippet the use of the parameter in the catch block is not detected and a warning for an unused parameter is shown.

### Known workarounds

Unwrap all using blocks into try-finally blocks

### Related information
* Visual Studio 16.8.6
* SonarScanner for .NET version 8.18.0.27296
* Operating System Windows 10 Version 1909

RobertSimonsSMA:
So, will this be fixed in the foreseeable future?

Will it be weeks, months or years?

Sincerely
  Robert

pavel-mikula-sonarsource:
Hi @RobertSimonsSMA,

This should be fixed this year as part of [MMF-2394](https://jira.sonarsource.com/browse/MMF-2394).

PR #4117, opened 2021-03-03
S1172 FP repro with try catch block

csaba-sagi-sonarsource:
Adds reproducer for #4096

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MethodParameterUnused.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MethodParameterUnused.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MethodParameterUnused.cs
```diff
@@ -481,4 +482,25 @@ bool IsTrue(object x)
             bool Unused() => condition;
         }
     }
+
+    // https://github.com/SonarSource/sonar-dotnet/issues/4096
+    public class Repro_4096
+    {
+        private int TryCatchWithUsing(int errorCode) // Noncompliant
```
pavel-mikula-sonarsource, line 489:
```suggestion
        private int TryCatchWithUsing(int errorCode) // Noncompliant FP
```

pavel-mikula-sonarsource (approved):
LGTM, there's just one comment improvement before merging

Issue #4111, opened 2021-03-03
S5753 subtask -  update rule to get web.config from FilesToAnalyze.txt


andrei-epure-sonarsource:
Depends on #4110 and #4116.

Take the FilesToAnalyzePath from the SonarProjectConfig, read the `web.config` paths and use that in `DisablingRequestValidationBase.cs` (see #3938).

This issue is a sub-task for #3951 and should not be part of the milestone and should be merged to PR#3938 (on the feature branch).

Note: this means that the rule won't work in SonarLint (it's a security hotspot anyway) AND won't work anymore in the .NET ITs (where we force security hotspots to run, but now it won't work anymore).

PR #4120, opened 2021-03-04
Add ProjectConfigReader

### LINKED ISSUE #4110: S5753 subtask - create a configuration class for SonarProjectConfig.xml
See #4105.

Before we use it, we need a class to read it and provide getters for the properties inside. This class should be properly tested and resilient.

This issue is a sub-task for #3951 and should not be part of the milestone and should be merged to PR#3938 (on the feature branch)

andrei-epure-sonarsource:
Fixes #4110


andrei-epure-sonarsource:
@csaba-sagi-sonarsource @pavel-mikula-sonarsource - the ITs are failing because it's based on the feature branch.
Before merging, I will need to add a few more tests for coverage, but otherwise it's ready to review.

csaba-sagi-sonarsource (approved):
LGTM

analyzers/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
```diff
@@ -100,6 +103,9 @@
     <Content Include="CBDEMocks\CBDEWaitAndSucceeds.exe">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="ResourceTests\SonarProjectConfig\**\*">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
```
pavel-mikula-sonarsource, line 108: Can we use something like `ResourceTests\**\*` to be done with the topic for once and all and to remove the next 50 repetitive lines?

The pattern and purpose is pretty similar to test cases above
```
    <None Include="TestCases\**\*">
    </None>
```

analyzers/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
```diff
@@ -64,6 +64,9 @@
     <None Remove="CBDEMocks\CBDESucceedsWithIncorrectResults.exe" />
     <None Remove="CBDEMocks\CBDEWaitAndSucceeds.exe" />
     <None Remove="ResourceTests\RuleWithBooleanParameter\SonarLint.xml" />
+    <None Remove="ResourceTests\SonarProjectConfig\Invalid_DifferentNamespace\SonarProjectConfig.xml" />
+    <None Remove="ResourceTests\SonarProjectConfig\Invalid_MixedSeparators\SonarProjectConfig.xml" />
+    <None Remove="ResourceTests\SonarProjectConfig\Invalid_DifferentClassName\SonarProjectConfig.xml" />
```
pavel-mikula-sonarsource, line 69: I don't understand this piece:
Why are only 3 out of 13 files here? Is it intended? Does it have to be here? What's the reason for having 3 `None` and others as `Content`?

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        [DataTestMethod]
+        [DataRow("Invalid_MissingAnalysisConfigPath")]
+        [DataRow("Invalid_MissingOutPath")]
+        [DataRow("Invalid_MissingProjectPath")]
+        [DataRow("Invalid_MissingProjectType")]
+        [DataRow("Invalid_MissingTargetFramework")]
+
+        public void WhenProjectConfigHasMissingValues_FilesToAnalyzePath_ReturnsCorrectValue(string folder)
```
pavel-mikula-sonarsource, line 97: _Nitpick:_ Extra EOL
```suggestion
        [DataRow("Invalid_MissingTargetFramework")]
        public void WhenProjectConfigHasMissingValues_FilesToAnalyzePath_ReturnsCorrectValue(string folder)
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
```diff
@@ -0,0 +1,126 @@
...
+            sut.ProjectType.Should().Be(ProjectType.Product);
+            sut.TargetFramework.Should().Be("netcoreapp3.1");
+        }
+
+        [DataTestMethod]
+        [DataRow("Invalid_MissingAnalysisConfigPath")]
+        [DataRow("Invalid_MissingOutPath")]
+        [DataRow("Invalid_MissingProjectPath")]
+        [DataRow("Invalid_MissingProjectType")]
+        [DataRow("Invalid_MissingTargetFramework")]
```
pavel-mikula-sonarsource, line 95: Let's rename these cases from "Invalid_MissingXxxx" to "Missing_Xxx".
The XML schema and structure is valid. It has just unexpected values for whatever reason on S4MSB side, or weird project type or structure, or some new kind of project type that's not yet supported.
Invalid are only these cases:
```
Invalid_DifferentClassName
Invalid_DifferentNamespace
Invalid_Xml
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            var sut = new ProjectConfigReader(options);
+
+            sut.FilesToAnalyzePath.Should().Be(@"c:\foo\bar\.sonarqube\conf\0\FilesToAnalyze.txt");
+        }
+
+        [DataTestMethod]
+        [DataRow("Invalid_DifferentClassName")]
+        [DataRow("Invalid_DifferentNamespace")]
+        [DataRow("Invalid_WrongProjectTypeValue")]
```
pavel-mikula-sonarsource, line 109: Unexpected (not wrong) ProjectType is an enum on S4MSB side. Enums can have new values over time and we should not fail in this case if somebody uses new S4MSB with old Analyzer.
Let's add dedicated test for this with `WhenProjectConfigHasUnexpectedProjectType_FallBackToProduction`

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
```diff
@@ -0,0 +1,126 @@
...
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.UnitTest.Helpers
+{
+    [TestClass]
+
+    public class ProjectConfigReaderTests
+    {
+        [TestMethod]
+        public void WhenProjectConfigHasWindowsPathsAndProductType_ReturnsCorrectValue()
```
pavel-mikula-sonarsource, line 32: As we're in `ProjectConfigReaderTests`, we can omit `ProjectConfig` part of the test names to make them easier to read. Something like
`WindowsPathsAndProductType_ReturnsCorrectValue`
`WithWindowsPathsAndProductType_ReturnsCorrectValue`

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
```diff
@@ -0,0 +1,126 @@
...
+
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.UnitTest.Helpers
+{
+    [TestClass]
+
+    public class ProjectConfigReaderTests
```
pavel-mikula-sonarsource, line 29: _Nitpick:_ Extra EOL
```suggestion
    [TestClass]
    public class ProjectConfigReaderTests
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
pavel-mikula-sonarsource: We can also decide to remove something from the XML file in the future

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
```diff
@@ -0,0 +1,126 @@
...
+            var sut = new ProjectConfigReader(options);
+
+            sut.AnalysisConfigPath.Should().BeNull();
+            sut.ProjectPath.Should().BeNull();
+            sut.FilesToAnalyzePath.Should().BeNull();
+            sut.OutPath.Should().BeNull();
+            sut.ProjectType.Should().BeNull();
+            sut.TargetFramework.Should().BeNull();
+        }
+    }
```
pavel-mikula-sonarsource, line 125: Let's add _blind_ test similar to https://github.com/SonarSource/sonar-scanner-msbuild/pull/962 / `WriteProjectConfigFileTests.cs` / `Execute_AllPropertiesAreSet` ensuring that we have a test case for each new property we add in the future.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
pavel-mikula-sonarsource: We'll need to rename the first tests in this file to remove the "AndProductType" part.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
pavel-mikula-sonarsource, line 32: These first 3 tests has mixed responsibilities. They test path & everything else. As the windows vs. unix vs. mixed are special cases, I suggest to keep their content short and focused on that scenario. And have one test similar to the first one with name like "LoadsExpectedValues"

pavel-mikula-sonarsource (changes requested):
Round 1 for UTs, I'll review implementation after weekly

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                return null;
+            }
+            try
+            {
+                var ser = new XmlSerializer(typeof(ProjectConfig));
+                using var fs = File.Open(sonarProjectConfig.Path, FileMode.Open, FileAccess.Read, FileShare.Read);
+                return (ProjectConfig)ser.Deserialize(fs);
+            }
+            catch (Exception)
```
pavel-mikula-sonarsource, line 65: I think we should fail hard with AD0001 warning in this case. 
We're only in batch mode and something seriously unexpected happened here. 

Something was configured (like the exclusions, or a test project) and we'll silently behave very differently in this case. That would be a source of a poor user experience and big support pain

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        public ProjectConfigReader(AnalyzerOptions options)
+        {
+            projectConfig = new Lazy<ProjectConfig>(() => ReadContent(options));
+        }
+
+        private static ProjectConfig ReadContent(AnalyzerOptions options)
+        {
+            var sonarProjectConfig = options.AdditionalFiles.FirstOrDefault(IsSonarProjectConfig);
+            if (sonarProjectConfig == null || !File.Exists(sonarProjectConfig.Path))
```
pavel-mikula-sonarsource, line 55: Just thinking out loud about `File.Exists`: "SonarProjectConfig.xml" was added as additional file but doesn't exist. That's a serious violation of configuration of the project.

I think we should fail hard in that specific case as well, for the same reason as the other comment in "catch" line.

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
```diff
@@ -0,0 +1,76 @@
...
+            catch (Exception)
+            {
+                // cannot log exception
+                return null;
+            }
+        }
+
+        private static bool IsSonarProjectConfig(AdditionalText additionalText) =>
+            additionalText.Path != null
+            && new FileInfo(additionalText.Path).Name.Equals("SonarProjectConfig.xml", StringComparison.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: As just verified: `GetFileName` doesn't build `FileInfo` internally:
https://referencesource.microsoft.com/#mscorlib/system/io/path.cs,95facc58d06cadd0

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
```diff
@@ -0,0 +1,76 @@
...
+            try
+            {
+                var ser = new XmlSerializer(typeof(ProjectConfig));
+                using var fs = File.Open(sonarProjectConfig.Path, FileMode.Open, FileAccess.Read, FileShare.Read);
+                return (ProjectConfig)ser.Deserialize(fs);
+            }
+            catch (Exception)
+            {
+                // cannot log exception
+                return null;
```
pavel-mikula-sonarsource, line 68: Let's consider returning `new ProjectConfig()` with null values instead. It will remove bunch of `?` above and simplify the class inner state handling.

pavel-mikula-sonarsource (changes requested):
Part 2 - implementation

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectType.cs
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+namespace SonarAnalyzer.Helpers
+{
+    /// <summary>
+    /// Possible types of project
+    /// </summary>
+    public enum ProjectType
+    {
+        Product = 0,
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: I just copy-pasted it from the Scanner.
It's bad practice to have the default value assigned to a meaningful value. `0` should be `None`. But that's how the Scanner has it so won't change.

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
andrei-epure-sonarsource: I agree, however it's the same fail-soft approach we have when parsing the `SonarLint.xml` file. In both cases, the behavior will be unexpected...

pavel-mikula-sonarsource: `SonarLint.xml` is little different IMO, as users are expected to work with that file, SL in connected mode provides the file to the project and users are expected to mess things up sometimes. And it can just naturally go out of sync.

Scope of `SonarProjectConfig` is narrow and users are not supposed to mess with it.

pavel-mikula-sonarsource: Yes, we might rethink how to improve `SonarLint.xml`-troubleshooting user experience later. For example:
`Sxxxx: SonarLint.xml should exist and be valid` :D As we will have the path to the `csproj`, it could be the first rule 😄

We might actually raise some issue in the SonarLint.xml file itself 🐶 🥫

analyzers/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
pavel-mikula-sonarsource: Awesome!

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            sut.AnalysisConfigPath.Should().BeNull();
+            sut.ProjectPath.Should().BeNull();
+            sut.FilesToAnalyzePath.Should().BeNull();
+            sut.OutPath.Should().BeNull();
+            sut.ProjectType.Should().Be(ProjectType.Product);
+            sut.TargetFramework.Should().BeNull();
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(InvalidOperationException), "File SonarProjectConfig.xml has been added as an AdditionalFile but does not exist.")]
```
pavel-mikula-sonarsource, line 151: Isn't this violation of RSPEC-3431? The `Should().BeNull()` is a smoke and it's not clearly visible where it should fail.
This should do the trick:
```
sut.Invoking(x => x.AnalysisConfigPath).Should().Throw<InvalidOperationException>().WithMessage("File SonarProjectConfig.xml has been added as an AdditionalFile but does not exist.");
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
```diff
@@ -0,0 +1,160 @@
...
+        public void WhenFileIsMissing_ThrowException()
+        {
+            var options = TestHelper.CreateOptions("ResourceTests\\SonarProjectConfig\\FOO\\SonarProjectConfig.xml");
```
pavel-mikula-sonarsource, line 154: We can use shorter and more explicit path. For some scenarios, "FOO" as a name might actually exist.
```suggestion
            var options = TestHelper.CreateOptions("ThisPathDoesNotExist\\SonarProjectConfig.xml");
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            sut.OutPath.Should().Be(@"C:\foo\bar\.sonarqube\out\0");
+            sut.ProjectType.Should().Be(ProjectType.Product);
+            sut.TargetFramework.Should().Be("netcoreapp3.1");
+        }
+
+        [DataTestMethod]
+        [DataRow("Path_MixedSeparators", @"c:\foo\bar\.sonarqube\conf/0/FilesToAnalyze.txt")]
+        [DataRow("Path_Unix", @"/home/user/.sonarqube/conf/0/FilesToAnalyze.txt")]
+        [DataRow("Path_Windows", @"c:\foo\bar\.sonarqube\conf\0\FilesToAnalyze.txt")]
+        public void WithVariousPathFormats_ReturnsAsIsValue(string project, string expectedFilesToAnalyzePath)
```
pavel-mikula-sonarsource, line 65: Maybe like this?
```suggestion
        public void WithVariousPathFormats_ReturnsValueAsIs(string project, string expectedFilesToAnalyzePath)
```

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
```diff
@@ -0,0 +1,80 @@
+﻿/*
```
pavel-mikula-sonarsource, line 1: Is merge to the feature branch intentional on this one? I'd be in for merging it to master instead. The whole sprint would have to remain on a feature branch like this.

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        public string AnalysisConfigPath => projectConfig.Value.AnalysisConfigPath;
+        public string FilesToAnalyzePath => projectConfig.Value.FilesToAnalyzePath;
+        public string OutPath => projectConfig.Value.OutPath;
+        public string ProjectPath => projectConfig.Value.ProjectPath;
+        public string TargetFramework => projectConfig.Value.TargetFramework;
+        public ProjectType ProjectType =>
+            Enum.TryParse<ProjectType>(projectConfig.Value.ProjectType, out var result)
+            ? result
+            : ProjectType.Product;
```
pavel-mikula-sonarsource, line 48: We know that every single rule will  read this variable and therefore parse the enum. Let's add  `private ProjectType? projectType;` field and parse it only once here with little bit more logic.

pavel-mikula-sonarsource (changes requested):
Final polishing

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
pavel-mikula-sonarsource: ..after removing the older commits from this branch

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
andrei-epure-sonarsource: I thought about this but with a test that has 1 line I thought it shouldn't be a problem. Anyway, it's cleaner your way

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
andrei-epure-sonarsource: Yes, because I would keep it on the master branch until we get it working.

Either way, this isn't consumed by anything itself... it has no added value

pavel-mikula-sonarsource (approved):
LGTM

andrei-epure-sonarsource:
The ITs fail on
```
[ERROR]   WebConfigTest.should_raise_hotspot_on_web_config:48 expected:<[2]> but was:<[0]>
[ERROR]   WebConfigTest.should_raise_hotspot_on_web_config:48 expected:<[2]> but was:<[0]>
```
because #4111 is not done yet.

PR #4121, opened 2021-03-05
Use the NuGet Cache in the unit tests to speed up the build

costin-zaharia-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -86,7 +85,7 @@ private void Install()
                 var versionArgument = Version == Constants.NuGetLatestVersion ? string.Empty : $"-Version {Version}";
                 var configFile = ValidatedNuGetConfigPath();
                 // Explicitly specify the NuGet config to use to avoid being impacted by the NuGet config on the machine running the tests
-                var args = $"install {Id} {versionArgument} -OutputDirectory {Path.GetFullPath(PackagesFolderRelativePath)} -NonInteractive -ForceEnglishOutput -ConfigFile {configFile}";
+                var args = $"install {Id} {versionArgument} -OutputDirectory \"{Environment.GetEnvironmentVariable("NUGET_PACKAGES")}\" -NonInteractive -ForceEnglishOutput -ConfigFile {configFile}";
```
pavel-mikula-sonarsource, line 88: This should use `PackagesFolder` instead, shouldn't it? The fallback for caching would not work without the env variable.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -31,29 +31,29 @@ namespace SonarAnalyzer.UnitTest.MetadataReferences
 {
     public static partial class NuGetMetadataFactory
     {
-        private const string PackagesFolderRelativePath = @"..\..\..\..\..\packages\";
+        private static string PackagesFolder = Environment.GetEnvironmentVariable("NUGET_PACKAGES") ?? @"..\..\..\..\..\packages\";
```
pavel-mikula-sonarsource, line 34: This line should be after the next `const` line now.

pavel-mikula-sonarsource, line 34: To deal with the codesmell & to preserve the `const` behavior.
```suggestion
        private static readonly string PackagesFolder = Environment.GetEnvironmentVariable("NUGET_PACKAGES") ?? @"..\..\..\..\..\packages\";
```

pavel-mikula-sonarsource (changes requested):
One line in `Package` seems suspicious. The rest is just minor polishing.

pavel-mikula-sonarsource (approved):
LGTM :shipit:

Issue #4119, opened 2021-03-04
Cache the ProjectConfig and FilesToAnalyze content in the AnalysisContext


andrei-epure-sonarsource:
Related to #3951.

After #3951, we may want to prepare for the future and avoid reading the settings for each rule for each compilation.

See #2453 where we are using the `AnalysisContext` cache. There we really needed it because we verify for each rule whether it should run on generated code or not.

PR #4123, opened 2021-03-05
Add FilesToAnalyzeProvider

### LINKED ISSUE #4116: S5753 subtask - create a class to allow to do file queries from the FilesToAnalyze.txt
This will be complementary to #4110.

It needs the `FilesToAnalyze.txt` path, parse & loads the content in memory.

It allows to get files from the list based on some pattern or filename:
- filename based
- regex based

This issue is a sub-task for #3951 and should not be part of the milestone and should be merged to PR#3938 (on the feature branch).

csaba-sagi-sonarsource:
Fixes #4116


analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeRetriever.cs
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace SonarAnalyzer.Helpers
+{
+    public class FilesToAnalyzeRetriever : IFilesToAnalyzeRetriever
```
pavel-mikula-sonarsource, line 28: As it reads the content from the disk
```suggestion
    public class FilesToAnalyzeReader : IFilesToAnalyzeRetriever
```

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeRetriever.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        public IEnumerable<string> RetrieveFilesToAnalyze(string filePath)
+        {
+            try
+            {
+                return File.ReadAllLines(filePath);
+            }
+            catch (Exception ex) when (ex is IOException || ex is ArgumentNullException || ex is ArgumentException)
```
pavel-mikula-sonarsource, line 36: Do we need the filter? If this fails for whatever reason, we can return empty.

pavel-mikula-sonarsource, line 36: Catching explicitly `ArgumentNullExcetion` looks like a exception-driven logic (a bad pattern). You should check for empty filepath and return empty collection instead, if there's an null input or the file does not exists.

analyzers/src/SonarAnalyzer.Common/Helpers/IFilesToAnalyzeRetriever.cs
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+
+namespace SonarAnalyzer.Helpers
+{
+    public interface IFilesToAnalyzeRetriever
```
pavel-mikula-sonarsource, line 25: Am I missing something or is this interface with one method implemented only once? In that case, we don't need an interface.

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeRetriever.cs
```diff
@@ -0,0 +1,43 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace SonarAnalyzer.Helpers
+{
+    public class FilesToAnalyzeRetriever : IFilesToAnalyzeRetriever
+    {
+        public IEnumerable<string> RetrieveFilesToAnalyze(string filePath)
```
pavel-mikula-sonarsource, line 30: Following the comment on the interface, this can be just a small private method inside `FilesToAnalyzeStorage`, right? Where the `Storage` could be `Provider` to deal with name/responsibility issue.

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeStorage.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace SonarAnalyzer.Helpers
+{
+    public class FilesToAnalyzeStorage
+    {
+        private readonly Lazy<IEnumerable<string>> filestoAnalyze;
```
pavel-mikula-sonarsource, line 32:
```suggestion
        private readonly Lazy<IEnumerable<string>> filesToAnalyze;
```
or just
```suggestion
        private readonly Lazy<IEnumerable<string>> files;
```
or
```suggestion
        private readonly Lazy<IEnumerable<string>> allFiles;
```

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeStorage.cs
```diff
@@ -0,0 +1,48 @@
...
+        [ExcludeFromCodeCoverage]
+        public FilesToAnalyzeStorage(string filePath) : this(filePath, new FilesToAnalyzeRetriever()) { }
+
+        internal FilesToAnalyzeStorage(string filePath, IFilesToAnalyzeRetriever filesToAnalyzeRetriever)
+        {
+            filestoAnalyze = new Lazy<IEnumerable<string>>(() => filesToAnalyzeRetriever.RetrieveFilesToAnalyze(filePath));
+        }
+
+        public IEnumerable<string> FilesToAnalyze(string fileName) =>
+            filestoAnalyze.Value.Where(x => Path.GetFileName(x).IndexOf(fileName, StringComparison.OrdinalIgnoreCase) >= 0);
```
pavel-mikula-sonarsource, line 43: This would return many unexpected files if I search for "a". Searching for "web.config" would also return "oldweb.config.bak"
```suggestion
            filestoAnalyze.Value.Where(x => Path.GetFileName(x).Equals(fileName, StringComparison.OrdinalIgnoreCase));
```

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeStorage.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        [ExcludeFromCodeCoverage]
+        public FilesToAnalyzeStorage(string filePath) : this(filePath, new FilesToAnalyzeRetriever()) { }
+
+        internal FilesToAnalyzeStorage(string filePath, IFilesToAnalyzeRetriever filesToAnalyzeRetriever)
+        {
+            filestoAnalyze = new Lazy<IEnumerable<string>>(() => filesToAnalyzeRetriever.RetrieveFilesToAnalyze(filePath));
+        }
+
+        public IEnumerable<string> FilesToAnalyze(string fileName) =>
```
pavel-mikula-sonarsource, line 42:
```suggestion
        public IEnumerable<string> FindFiles(string fileName) =>
```

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeStorage.cs
```diff
@@ -0,0 +1,48 @@
...
+            filestoAnalyze.Value.Where(x => Path.GetFileName(x).IndexOf(fileName, StringComparison.OrdinalIgnoreCase) >= 0);
+
+        public IEnumerable<string> FilesToAnalyze(Regex fileFilter) =>
```
pavel-mikula-sonarsource, line 45: Method: To be aligned with previous one.
Parameter: To be explicit that we'll search against the full path (contrary the previous method that declares `fileName` only)
```suggestion
        public IEnumerable<string> FindFiles(Regex fullPathRegex) =>
```

analyzers/tests/SonarAnalyzer.UnitTest/ResourceTests/FilesToAnalyze/FilesToAnalyze.txt
```diff
@@ -0,0 +1,5 @@
+﻿C:\Projects/DummyProj/wEB.config
```
andrei-epure-sonarsource, line 1: maybe also add a test case with file with some numbers or non-path text? (like `foo`, `123`)

or add it here directly if the class supports random stuff inside

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeStorage.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        private readonly Lazy<IEnumerable<string>> filestoAnalyze;
+
+        [ExcludeFromCodeCoverage]
+        public FilesToAnalyzeStorage(string filePath) : this(filePath, new FilesToAnalyzeRetriever()) { }
+
+        internal FilesToAnalyzeStorage(string filePath, IFilesToAnalyzeRetriever filesToAnalyzeRetriever)
+        {
+            filestoAnalyze = new Lazy<IEnumerable<string>>(() => filesToAnalyzeRetriever.RetrieveFilesToAnalyze(filePath));
+        }
```
pavel-mikula-sonarsource, line 40: I believe this could be simplified a lot with removing the interface and the other class.
Let's don't use Lazy here and directly load all values. The state&caching will be handled by the user (Andrei's PR)

pavel-mikula-sonarsource (changes requested):
This should be pretty small and simple class. I didn't go through the tests as I asked to delete a lot of code 🙂

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeStorage.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace SonarAnalyzer.Helpers
+{
+    public class FilesToAnalyzeStorage
+    {
+        private readonly Lazy<IEnumerable<string>> filestoAnalyze;
+
+        [ExcludeFromCodeCoverage]
```
andrei-epure-sonarsource, line 34: this only works at class level

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeStorage.cs
andrei-epure-sonarsource: I think the main idea I have is not to do the I/O operation in the constructor, because operations that can throw exceptions should be avoided in the constructor.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/FilesToAnalyzeRetrieverTest.cs
```diff
@@ -0,0 +1,64 @@
...
+        [TestMethod]
+        public void EmptyFile_RetrieveFilesToAnalyze_ReturnsEmptyEnumerable()
+        {
+            var sut = new FilesToAnalyzeRetriever();
+
+            sut.RetrieveFilesToAnalyze(@"ResourceTests\FilesToAnalyze\EmptyFilesToAnalyze.txt").Should().BeEmpty();
+        }
+
+        [DataTestMethod]
+        [DataRow("")]
```
andrei-epure-sonarsource, line 47: maybe also add a non-empty string

The `null` and empty can be filtered with `IsNullOrEmpty`, however a non-empty string may cause an I/O exception ("file not found")

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/FilesToAnalyzeRetrieverTest.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            sut.RetrieveFilesToAnalyze(filePath).Should().BeEmpty();
+        }
+
+        [TestMethod]
+        public void ValidFile_RetrieveFilesToAnalyze_ReturnsRetrievedLines()
+        {
+            var sut = new FilesToAnalyzeRetriever();
+
+            sut.RetrieveFilesToAnalyze(@"ResourceTests\FilesToAnalyze\FilesToAnalyze.txt").Should().HaveCount(5);
```
andrei-epure-sonarsource, line 61: could you also check the contents of the list?

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeStorage.cs
csaba-sagi-sonarsource: WDYT about removing the interface and the other class and leaving the Lazy here. I would say it can do ho harm anyway. And remove the lazy stuff eventually when the caching is implemented in a different place.

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeStorage.cs
csaba-sagi-sonarsource: I wanted to separate concerns and  use  the other constructor and to have unit tests with only this unit but the  other class is removed as is the other constructor so this one is now tested.

analyzers/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
```diff
@@ -178,6 +178,15 @@
...
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="ResourceTests\FilesToAnalyze\EmptyFilesToAnalyze.txt">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="ResourceTests\FilesToAnalyze\FilesToAnalyze.txt">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Can you remove this entire `ItemGroup` now, when it's rebased? There are 3 `None` items currently.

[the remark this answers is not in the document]
csaba-sagi-sonarsource: Yes I rebased to be able to remove this but I totally forgot about it. Thanks for pointing out.

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeStorage.cs
pavel-mikula-sonarsource: My thoughts about the situation:
* I'm 100% sure that caching will be implemented else where (#4119, `ProjectConfig` will be properly cached and will hold a lazy reference to this class).
* Throwing exception from constructor is acceptable in our scenarios I think (easy to understand, easy to debug on current and future version). Same as throwing `ArgumentNullException` for example.
* It's also subject to [RFC-1925 6a)](https://tools.ietf.org/html/rfc1925) (If you don't know, the RFC itself is one of April's fool day jokes, but the content is a valid warning).
* The intended usage in this and the future scenario will be create&use directly. So throwing an exception from constructor is not a problem. 

I'm for removing the lazyness from this.

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace SonarAnalyzer.Helpers
+{
+    public class FilesToAnalyzeProvider
+    {
+        private readonly Lazy<IEnumerable<string>> allFiles;
+
+        public FilesToAnalyzeProvider(string filePath)
+        {
+            allFiles = new Lazy<IEnumerable<string>>(() => RetrieveFilesToAnalyze(filePath));
+        }
```
pavel-mikula-sonarsource: Could be arrow method
```suggestion
        public FilesToAnalyzeProvider(string filePath) =>
            allFiles = new Lazy<IEnumerable<string>>(() => RetrieveFilesToAnalyze(filePath));
```

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeStorage.cs
csaba-sagi-sonarsource: Indeed if we implement #4119  it does not make sense to have a lazy implementation here. I`ll remove it as it looks like we will have the time to implement #4119

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        public IEnumerable<string> FindFiles(string fileName) =>
+            allFiles.Value.Where(x => Path.GetFileName(x).Equals(fileName, StringComparison.OrdinalIgnoreCase));
+
+        public IEnumerable<string> FindFiles(Regex fullPathRegex) =>
+            allFiles.Value.Where(x => fullPathRegex.IsMatch(x));
+
+        private static IEnumerable<string> RetrieveFilesToAnalyze(string filePath)
```
pavel-mikula-sonarsource, line 44: I don't like the dog 😆 
```suggestion
        private static IEnumerable<string> LoadFilesToAnalyze(string filePath)
```
And we don't need to know that it's about `FilesToAnalyze` at this stage. The method content is not bound to it.
```suggestion
        private static IEnumerable<string> ReadFileLines(string filePath)
```
or
```suggestion
        private static IEnumerable<string> ReadLines(string filePath)
```

analyzers/tests/SonarAnalyzer.UnitTest/ResourceTests/FilesToAnalyze/InvalidValuesFileToAnalyze.txt
```diff
@@ -0,0 +1,4 @@
+﻿foo
+123
+C:\Projects\^^^\$\Controllers\HomeController.vb
```
pavel-mikula-sonarsource, line 3: I believe this is perfectly valid path. To get some invalid or unexpected stuff:
First one has invalid directory character in the way `:`
Second one is relative path from another drive (valid path, but unexpected). I'd like to see what `Path.GetFileName` actually does in that case
```suggestion
C:\Projects\Controllers:web.config
C:web.config
```

analyzers/tests/SonarAnalyzer.UnitTest/ResourceTests/FilesToAnalyze/InvalidValuesFileToAnalyze.txt
```diff
@@ -0,0 +1,4 @@
...
+C:\Projects/DummyProj/Views\Web.confiG
```
pavel-mikula-sonarsource, line 4: Why is this line in `Invalid` file? There's already a mixed `/\` case in valid file:
```
C:\Projects/DummyProj/Views\Web.confiG
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/FilesToAnalyzeProviderTest.cs
```diff
@@ -0,0 +1,112 @@
...
+            var results = sut.FindFiles("Web.config");
+            results.Should().BeEmpty();
+        }
+
+        [TestMethod]
+        public void EmptyFile_FindFiles_ReturnsEmptyEnumerable()
+        {
+            var sut = new FilesToAnalyzeProvider(@"ResourceTests\FilesToAnalyze\EmptyFilesToAnalyze.txt");
+
+            var results = sut.FindFiles("Web.config");
```
pavel-mikula-sonarsource, line 96: Let's search for all to validate the emptyness.
```suggestion
            var results = sut.FindFiles(new Regex(".*"));
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/FilesToAnalyzeProviderTest.cs
```diff
@@ -0,0 +1,112 @@
...
+            results.Should().HaveCount(2);
+            results.Should().Contain(MixedSlashesWebConfigPath1);
+            results.Should().Contain(MixedSlashesWebConfigPath2);
+        }
+
+        [TestMethod]
+        public void FileNameWithMixedCapitalizationAndMixedSlashes_FindFilesWithRegex_ReturnsAllWebConfigFiles()
+        {
+            var fileNamePattern = new Regex("web\\.config$", RegexOptions.IgnoreCase);
```
pavel-mikula-sonarsource, line 50: Let's demonstrate the intended full-path usage for the regex variant.
```suggestion
            var fileNamePattern = new Regex("[\\\/]web\\.config$", RegexOptions.IgnoreCase);
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/FilesToAnalyzeProviderTest.cs
```diff
@@ -0,0 +1,112 @@
...
+            var sut = new FilesToAnalyzeProvider(FilesToAnalyzePath);
+
+            var results = sut.FindFiles(fileNamePattern);
+            results.Should().HaveCount(2);
+            results.Should().Contain(MixedSlashesWebConfigPath1);
+            results.Should().Contain(MixedSlashesWebConfigPath2);
+        }
+
+        [TestMethod]
+        public void FileWithInvalidvalues_FindFilesWithFileName_ReturnsValidValue()
```
pavel-mikula-sonarsource, line 61: _Nitpick_
```suggestion
        public void FileWithInvalidValues_FindFilesWithFileName_ReturnsValidValue()
```

analyzers/tests/SonarAnalyzer.UnitTest/ResourceTests/FilesToAnalyze/InvalidValuesFileToAnalyze.txt
```diff
@@ -0,0 +1,4 @@
+﻿foo
```
pavel-mikula-sonarsource, line 1: This file should be renamed to `InvalidValuesFilesToAnalyze.txt` (missing `s`, it was hard to find with filter in Solution explorer)

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/FilesToAnalyzeProviderTest.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            var sut = new FilesToAnalyzeProvider(InvalidFilesToAnalyzePath);
```
pavel-mikula-sonarsource, line 63: Let's add test searching for `123` on this file. As it should find it. And at the same time, we'll test that it works for file without extension as well.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/FilesToAnalyzeProviderTest.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            var results = sut.FindFiles("Web.config");
+            results.Should().HaveCount(1);
+            results.Should().Contain(MixedSlashesWebConfigPath2);
+        }
+
+        [TestMethod]
+        public void FileWithInvalidvalues_FindFilesWithRegex_ReturnsValidValue()
```
pavel-mikula-sonarsource, line 71: _Nitpick 2_ (copy/pasted)
```suggestion
        public void FileWithInvalidValues_FindFilesWithRegex_ReturnsValidValue()
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/FilesToAnalyzeProviderTest.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        [DataTestMethod]
+        [DataRow("")]
+        [DataRow(null)]
+        [DataRow("invalidPath")]
+        public void InvalidPath_FindFiles_ReturnsEmptyEnumerable(string filePath)
+        {
+            var sut = new FilesToAnalyzeProvider(filePath);
+
+            var results = sut.FindFiles("Web.config");
```
pavel-mikula-sonarsource, line 108: Same as above for empty
```suggestion
            var results = sut.FindFiles(new Regex(".*"));
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/FilesToAnalyzeProviderTest.cs
```diff
@@ -0,0 +1,112 @@
...
+            results.Should().HaveCount(1);
+            results.Should().Contain(MixedSlashesWebConfigPath2);
+        }
+
+        [TestMethod]
+        public void NonExistentFile_FindFiles_ReturnsEmptyEnumerable()
+        {
+            var sut = new FilesToAnalyzeProvider(@"ResourceTests\FilesToAnalyze\NonExistingFile.txt");
+
+            var results = sut.FindFiles("Web.config");
```
pavel-mikula-sonarsource, line 87: Same the other one for empty
```suggestion
            var results = sut.FindFiles(new Regex(".*"));
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/FilesToAnalyzeProviderTest.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            var results = sut.FindFiles(fileNamePattern);
+            results.Should().HaveCount(1);
+            results.Should().Contain(MixedSlashesWebConfigPath2);
+        }
+
+        [TestMethod]
+        public void NonExistentFile_FindFiles_ReturnsEmptyEnumerable()
+        {
+            var sut = new FilesToAnalyzeProvider(@"ResourceTests\FilesToAnalyze\NonExistingFile.txt");
```
pavel-mikula-sonarsource, line 85: This `@"ResourceTests\FilesToAnalyze\NonExistingFile.txt"` can be added as a `DataRow` in the last test to save some lines, right?

pavel-mikula-sonarsource (changes requested):
Polishing round, minor comments

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/FilesToAnalyzeProviderTest.cs
```diff
@@ -0,0 +1,112 @@
...
+    [TestClass]
+    public class FilesToAnalyzeProviderTest
+    {
+        private const string MixedSlashesWebConfigPath1 = @"C:\Projects/DummyProj/wEB.config";
+        private const string MixedSlashesWebConfigPath2 = @"C:\Projects/DummyProj/Views\Web.confiG";
+        private const string FilesToAnalyzePath = @"ResourceTests\FilesToAnalyze\FilesToAnalyze.txt";
+        private const string InvalidFilesToAnalyzePath = @"ResourceTests\FilesToAnalyze\InvalidValuesFileToAnalyze.txt";
+
+        [TestMethod]
+        public void FileNameWithMixedCapitalizationAndMixedSlashes_FindFilesWithFileName_ReturnsAllWebConfigFiles()
```
pavel-mikula-sonarsource, line 37: Let's add a test that will search for regex `.*` and that will validate the returned collection to contain expected values (all lines from the file)

analyzers/tests/SonarAnalyzer.UnitTest/ResourceTests/FilesToAnalyze/InvalidValuesFileToAnalyze.txt
csaba-sagi-sonarsource: I`ll rename the file. To something like FilesToAnalyzeWithInvalidValues. The idea was to check that the file is fully parsed even with invalid values.

analyzers/tests/SonarAnalyzer.UnitTest/ResourceTests/FilesToAnalyze/InvalidValuesFileToAnalyze.txt
csaba-sagi-sonarsource: I renamed the file to FilesToAnalyzeWithInvalidValues

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+
+        private static bool FilterByFileName(string fullPath, string fileName)
+        {
+            try
+            {
+                return Path.GetFileName(fullPath).Equals(fileName, StringComparison.OrdinalIgnoreCase);
+            }
+            catch (ArgumentException)
```
pavel-mikula-sonarsource, line 66: And with new test cases, new stuff appears 😄

pavel-mikula-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.Common/Helpers/FilesToAnalyzeStorage.cs
pavel-mikula-sonarsource: Should we create an issue to configure it in our build then?

csaba-sagi-sonarsource: I think the answer is yes, as there are a few places where it is used on method level.

PR #4128, opened 2021-03-08
Avoid stale config when SonarLint.xml changes

costin-zaharia-sonarsource:
This will fix also the intermittent ITs failures we have on the CI pipeline.

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
@@ -21,7 +21,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Xml.Linq;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Helpers;
 
@@ -41,9 +40,11 @@ namespace SonarAnalyzer.Common
         public static IAnalyzerConfiguration AlwaysEnabled { get; } =
             new AlwaysEnabledConfiguration();
 
+        public static IRuleLoader RuleLoader { get; } = new RuleLoader();
+
         private class AlwaysEnabledConfiguration : IAnalyzerConfiguration
         {
-            public void Initialize(AnalyzerOptions options)
+            public void Initialize(AnalyzerOptions options, IRuleLoader ruleLoader)
             {
                 // Ignore options because we always return true for IsEnabled
             }
@@ -56,6 +57,10 @@ namespace SonarAnalyzer.Common
         /// </summary>
         private class HotspotConfiguration : IAnalyzerConfiguration
         {
+            // Hotspot configuration is cached at the assembly level and the MsBuild process
+            // can reuse the already loaded assembly when multiple projects are analyzed one after the other.
+            // Due to this we have to check the current configuration path to see if a reload is needed.
+            private string loadedSonarLintXmlPath;
             private bool isInitialized;
             private HashSet<string> enabledRules;
 
@@ -76,38 +81,35 @@ namespace SonarAnalyzer.Common
                     && enabledRules.Contains(ruleKey);
             }
 
-            public void Initialize(AnalyzerOptions options)
+            public void Initialize(AnalyzerOptions options, IRuleLoader ruleLoader)
             {
-                if (isInitialized)
+                var currentSonarLintXmlPath = GetSonarLintXmlPath(options);
+                if (isInitialized && loadedSonarLintXmlPath == currentSonarLintXmlPath)
                 {
                     return;
                 }
 
                 lock (@lock)
                 {
-                    if (isInitialized)
+                    if (isInitialized && loadedSonarLintXmlPath == currentSonarLintXmlPath)
                     {
                         return;
                     }
 
                     isInitialized = true;
+                    loadedSonarLintXmlPath = currentSonarLintXmlPath;
 
-                    var sonarLintXml = options.AdditionalFiles
-                        .FirstOrDefault(f => ParameterLoader.IsSonarLintXml(f.Path));
-
-                    if (sonarLintXml == null)
+                    if (loadedSonarLintXmlPath == null)
                     {
                         return;
                     }
 
-                    var document = XDocument.Load(sonarLintXml.Path);
-
-                    enabledRules = document.Descendants("Rule")
-                        .Select(r => r.Element("Key")?.Value)
-                        .WhereNotNull()
-                        .ToHashSet();
+                    enabledRules = ruleLoader.GetEnabledRules(loadedSonarLintXmlPath);
                 }
             }
+
+            private static string GetSonarLintXmlPath(AnalyzerOptions options) =>
+                options.AdditionalFiles.FirstOrDefault(f => ParameterLoader.IsSonarLintXml(f.Path))?.Path;
         }
     }
 }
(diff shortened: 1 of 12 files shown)


analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
```diff
@@ -76,38 +81,35 @@ public bool IsEnabled(string ruleKey)
...
+            public void Initialize(AnalyzerOptions options, IRuleLoader ruleLoader)
             {
-                if (isInitialized)
+                var currentSonarLintXmlPath = GetSonarLintXmlPath(options);
+                if (isInitialized && loadedSonarLintXmlPath == currentSonarLintXmlPath)
                 {
                     return;
                 }
 
                 lock (@lock)
```
csaba-sagi-sonarsource, line 92: Can we rename the lock? I do not like @keyword identifier names... and besides that it validates S1311 that says static readonly fields should begin with upper-case letter.

analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
```diff
@@ -76,38 +81,35 @@ public bool IsEnabled(string ruleKey)
                     && enabledRules.Contains(ruleKey);
             }
 
-            public void Initialize(AnalyzerOptions options)
```
csaba-sagi-sonarsource, line 79: IsEnabled can be simplified to a form of  an expression bodied method.

analyzers/src/SonarAnalyzer.Common/Common/RuleLoader.cs
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Xml.Linq;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Common
+{
+    public class RuleLoader : IRuleLoader
```
csaba-sagi-sonarsource, line 28: I think we can remove the interface and the class itself as it is only used in the AnalyzerConfiguration.cs  it can be a private static method in that class. And for the classes we can use SonarLint.xml files from the Resources instead of mocking.

analyzers/src/SonarAnalyzer.Common/Common/RuleLoader.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        public HashSet<string> GetEnabledRules(string path) =>
```
csaba-sagi-sonarsource, line 30: Can we get invalid Sonarlint.xml files? If so can we have any exceptions during the parsing?

analyzers/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
```diff
@@ -176,6 +176,12 @@
     </ProjectReference>
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="Common\Resources\SonarLint.xml">
```
csaba-sagi-sonarsource, line 180: I would say this can be in the same itemgroup with <None Include="Common\Resources\input.txt"> and maybe they can be merged with some Common\Resources\*  pattern.  Besides that I am not sure an always copy is needed. maybe preserveNewest would be good enough.

csaba-sagi-sonarsource (changes requested):
I left a few comments, some of which we already talked about.

analyzers/src/SonarAnalyzer.Common/Common/RuleLoader.cs
costin-zaharia-sonarsource: The RuleLoader together with the interface helps to decouple the IO operation and isolate it. 

In this way we can test the `HotspotConfiguration.Initialize` in isolation without accessing the disk and loading an XML file.

analyzers/src/SonarAnalyzer.Common/Common/RuleLoader.cs
costin-zaharia-sonarsource: These are generated by us and we didn't had problems so far. If an exception is raised then the user will get an `AD0001` which I think is what we want.

analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
```diff
@@ -76,38 +81,35 @@ public bool IsEnabled(string ruleKey)
...
+            public void Initialize(AnalyzerOptions options, IRuleLoader ruleLoader)
             {
-                if (isInitialized)
+                var currentSonarLintXmlPath = GetSonarLintXmlPath(options);
+                if (isInitialized && loadedSonarLintXmlPath == currentSonarLintXmlPath)
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Aren't parallel builds are handled by different processes?

analyzers/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
costin-zaharia-sonarsource: preserveNewest is good enough indeed. I also moved it near `Common\Resources\input.txt`

csaba-sagi-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I don't know. We've seen static state persisted state between analysis. Might worth to try to verify it locally with some logging.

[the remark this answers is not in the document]
costin-zaharia-sonarsource: For a solution with multiple project this should not be a problem since we have the same SonarLint.xml configuration for all the projects. 

For building multiple solutions in parallel, I think we don't need to support this scenario. According to [our documentation](https://docs.sonarqube.org/latest/analysis/scan/sonarscanner-for-msbuild/):
```
Concurrent analyses (i.e. parallel analysis of two solutions on the same build machine using a unique service account) are not supported by default by the Scanner for .NET.
```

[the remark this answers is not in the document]
costin-zaharia-sonarsource: We clarified a bit this directly and we see here two scenarios:
- when `AnalysisContext.EnableConcurrentExecution` is used: this case is happening in the same process and the `SonarLint.xml` should be the same for all the analyzers (it's the same project)
- when a user has a solution with multiple projects and it has manually configured different `SonarLint.xml` files for each project and it enables parallel builds. In this case, if the projects are compiled in parallel, they are compiled by different work processes and theoretically the list of enabled rules is loaded once per process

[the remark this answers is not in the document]
pavel-mikula-sonarsource: As a babystep, this should be fine for now. In long term, we need to rework how `SonarLint.xml` is used - it's read on different places in a different ways and not cached in `AnalysisContext` as it should be.

PR #4132, opened 2021-03-09
Updates compliant and non compliant examples for S3928

### LINKED ISSUE #4127: S3928: Fix documentation
### Description

https://rules.sonarsource.com/csharp/RSPEC-3928

Hi! I noticed two issues in the "Compliant Solution" section of this spec.

1. Typo in `nameOf(a)`
`throw new ArgumentNullException(nameOf(a));`
should be
`throw new ArgumentNullException(nameof(a));`
2. Wrong order of arguments. `"a"` is supposed to be the `paramName` here (not the `message`)
`throw new ArgumentNullException("My error message", "a");`
should be
`throw new ArgumentNullException("a", "My error message");`
Remarks: The analysis itself seems to be fine.

For reference:
- `public ArgumentNullException (string? paramName, string? message);`
- [Documentation](https://docs.microsoft.com/en-us/dotnet/api/system.argumentnullexception.-ctor?view=net-5.0#System_ArgumentNullException__ctor_System_String_System_String_)

I can also provide a PR if you agree with this changes :)

Greetings!

csaba-sagi-sonarsource:
Fixes #4127

### DIFF
--- a/analyzers/rspec/cs/S3928_c#.html
+++ b/analyzers/rspec/cs/S3928_c#.html
@@ -16,6 +16,7 @@ public void Foo(Bar a, int[] b)
   throw new ArgumentOutOfRangeException("c");
   throw new ArgumentOutOfRangeException("c", "My error message"); // Noncompliant
   throw new ArgumentOutOfRangeException("c", b, "My error message"); // Noncompliant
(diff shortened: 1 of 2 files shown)


costin-zaharia-sonarsource (approved):
LGTM!

PR #4133, opened 2021-03-09
Cache the ProjectConfig and FilesToAnalyze content in the AnalysisContext

### LINKED ISSUE #4119: Cache the ProjectConfig and FilesToAnalyze content in the AnalysisContext

pavel-mikula-sonarsource:
Fixes #4119

This is a draft, one more commit will arrive with integrating `FilesToAnalyzeProvider`

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
@@ -72,7 +72,7 @@ namespace SonarAnalyzer.Helpers
             }
 
             // ... but for compatibility purposes we need to keep handling the old-fashioned way. Old SonarLint can be used with latest NuGet.
-            if (SonarAnalysisContext.AreAnalysisScopeMatching(reportingContext.Compilation, new[] { reportingContext.Diagnostic.Descriptor }) &&
+            if (reportingContext.Compilation.AreAnalysisScopeMatching(new[] { reportingContext.Diagnostic.Descriptor }) &&
                 !VbcHelper.IsTriggeringVbcError(reportingContext.Diagnostic) &&
                 (SonarAnalysisContext.ShouldDiagnosticBeReported?.Invoke(reportingContext.SyntaxTree, reportingContext.Diagnostic) ?? true))
             {
--- a/analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
@@ -20,10 +20,8 @@
 
 using System;
 using System.IO;
-using System.Linq;
 using System.Xml.Serialization;
-using Microsoft.CodeAnalysis;
-using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Text;
 
 namespace SonarAnalyzer.Helpers
 {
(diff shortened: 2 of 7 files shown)


analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
```diff
@@ -72,7 +72,7 @@ private static void ReportDiagnostic(ReportingContext reportingContext)
             }
 
             // ... but for compatibility purposes we need to keep handling the old-fashioned way. Old SonarLint can be used with latest NuGet.
-            if (SonarAnalysisContext.AreAnalysisScopeMatching(reportingContext.Compilation, new[] { reportingContext.Diagnostic.Descriptor }) &&
+            if (reportingContext.Compilation.AreAnalysisScopeMatching(new[] { reportingContext.Diagnostic.Descriptor }) &&
```
csaba-sagi-sonarsource, line 75: It is either AreAnalysisScopesMatching or IsAnalysisScopeMatching. looking at the body of the method I would prefer the second.

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
```diff
@@ -32,54 +30,40 @@ namespace SonarAnalyzer.Helpers
...
+        public string AnalysisConfigPath => projectConfig.AnalysisConfigPath;
+        public string FilesToAnalyzePath => projectConfig.FilesToAnalyzePath;
+        public string OutPath => projectConfig.OutPath;
+        public string ProjectPath => projectConfig.ProjectPath;
+        public string TargetFramework => projectConfig.TargetFramework;
         public ProjectType ProjectType => projectType.Value;
+        public FilesToAnalyzeProvider FilesToAnalyze => filesToAnalyze.Value;
 
-        public ProjectConfigReader(AnalyzerOptions options)
+        public ProjectConfigReader(SourceText sonarProjectConfig, string logFileName)
```
csaba-sagi-sonarsource, line 45: I would say the logFileName is always SonarProjectConfigFileName, so maybe it is more appropriate to name it like that. This way it is a bit confusing which file it is.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
```diff
@@ -103,49 +95,38 @@ public void WhenHasUnexpectedProjectType_FallsBackToProduct()
...
+
+            var sut = new ProjectConfigReader(config, null);
+            var files = sut.FilesToAnalyze.FindFiles("web.config");
+
+            files.Should().HaveCount(2);
+            files.Should().BeEquivalentTo(new[] { @"C:\Projects/DummyProj/wEB.config", @"C:\Projects/DummyProj/Views\Web.confiG" });
         }
+
+        private ProjectConfigReader CreateProjectConfigReader(string relativePath) =>
+            new ProjectConfigReader(SourceText.From(File.ReadAllText(relativePath)), "LogNameFor-SonarProjectConfig.xml");
```
csaba-sagi-sonarsource, line 130: New expression can be simplified

```suggestion
            new (SourceText.From(File.ReadAllText(relativePath)), "LogNameFor-SonarProjectConfig.xml");
```

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
pavel-mikula-sonarsource: It will always have that value, but `ProjectConfigReader` doesn't know that anymore for sure. It's coming from outside, because it can't be read from `SourceText`.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
```diff
@@ -103,49 +95,38 @@ public void WhenHasUnexpectedProjectType_FallsBackToProduct()
...
-        [DataRow(null)]
-        [DataRow("/foo/bar/do-not-exit")]
-        [DataRow("/foo/bar/x.xml")]
-        public void WhenNoFileFound_ReturnsNull(string folder)
-        {
-            var options = TestHelper.CreateOptions(folder);
-
-            var sut = new ProjectConfigReader(options);
-
-            sut.AnalysisConfigPath.Should().BeNull();
```
csaba-sagi-sonarsource, line 123: When there are multiple assertions I prefere using `AssertionScope`s  and that way be able to catch all the issues in a single run if there are more than one.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
```diff
@@ -103,49 +95,38 @@ public void WhenHasUnexpectedProjectType_FallsBackToProduct()
...
-        }
-
         [DataTestMethod]
         [DataRow("Invalid_DifferentClassName")]
         [DataRow("Invalid_DifferentNamespace")]
         [DataRow("Invalid_Xml")]
         public void WhenInvalid_FilesToReturnPath_ThrowsException(string folder)
         {
-            var options = TestHelper.CreateOptions($"ResourceTests\\SonarProjectConfig\\{folder}\\SonarProjectConfig.xml");
+            Action a = () => CreateProjectConfigReader($"ResourceTests\\SonarProjectConfig\\{folder}\\SonarProjectConfig.xml");
```
csaba-sagi-sonarsource, line 109: With Assert.ThrowsException it can be simplified.
```suggestion
Assert.ThrowsException<InvalidOperationException>(() => CreateProjectConfigReader($"ResourceTests\\SonarProjectConfig\\{folder}\\SonarProjectConfig.xml"))
                                                             .Message.Should().Be("File LogNameFor-SonarProjectConfig.xml could not be parsed.");
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
```diff
@@ -88,9 +82,7 @@ public void WhenHasMissingFilesToAnalyzePath_ReturnsCorrectValue()
         [TestMethod]
         public void WhenHasUnexpectedProjectType_FallsBackToProduct()
         {
-            var options = TestHelper.CreateOptions($"ResourceTests\\SonarProjectConfig\\UnexpectedProjectTypeValue\\SonarProjectConfig.xml");
-
-            var sut = new ProjectConfigReader(options);
+            var sut = CreateProjectConfigReader($"ResourceTests\\SonarProjectConfig\\UnexpectedProjectTypeValue\\SonarProjectConfig.xml");
```
csaba-sagi-sonarsource, line 85: Wherever the variable is used only once I am not sure it is needed. From the class name ProjectConfigReaderTests it is clear that the sut is ProjectConfigReader

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -145,5 +148,92 @@ public void SonarAnalysis_WhenReportDiagnosticActionNotNull_AllowToContolWhether
...
+        [DataRow("/foo/bar/do-not-exit")]
+        [DataRow("/foo/bar/x.xml")]
+        public void ProjectConfiguration_WhenAdditionalFileNotPresent_ReturnsEmptyConfig(string folder)
+        {
+            var options = TestHelper.CreateOptions(folder);
+
+            var sut = new SonarAnalysisContext(null, Enumerable.Empty<DiagnosticDescriptor>());
+            var config = sut.ProjectConfiguration(options);
+
+            config.AnalysisConfigPath.Should().BeNull();
```
csaba-sagi-sonarsource, line 197: AssertionScope?

csaba-sagi-sonarsource (changes requested):
Overall it LGTM, there are only a few minor changes I would like to introduce.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ProjectConfigReaderTests.cs
pavel-mikula-sonarsource: That's +3 LOC for not too much of a value in this simple test (and it's siblings above). It will either fail  all because some common cause, or individually due to some renaming. It's easy to understand in both cases.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
pavel-mikula-sonarsource: Same as above

csaba-sagi-sonarsource (approved):
LGTM

PR #4138, opened 2021-03-10
Use SonarProjectConfig.xml instead of ProjectOutFolderPath.txt

### LINKED ISSUE #4118: Use SonarProjectConfig.xml instead of ProjectOutFolderPath.txt
This is follow up of #4110

Information from `ProjectOutFolderPath.txt` will be stored in `SonarProjectConfig.xml` / `OutPath`.

We'll use this as a primary source of information and fall back to old file only when config XML file does not exist for users with S4MSB 5.0 or lower.

Also align naming of that property across our analyzer: https://github.com/SonarSource/sonar-dotnet/blob/96ad8b26b3c9bce808b747bbac9d3be07dd6940e/analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs#L71

pavel-mikula-sonarsource:
Fixes #4118

Review per commit is suggested. There's a bunch of refactoring on the way, as it was hard to understand the logic. I tried to keep individual commits as small as possible.

### DIFF
--- a/analyzers/its/SonarAnalyzer.Testing.ImportBefore.targets
+++ b/analyzers/its/SonarAnalyzer.Testing.ImportBefore.targets
@@ -5,28 +5,35 @@
 
     <!-- Copy project-specific SonarLint.xml to enable project-specific settings, such as file headers -->
     <PropertyGroup>
-      <ProjectSpecificSonarLintXml Condition="Exists('$(MSBuildStartupDirectory)\config\$(PROJECT)\SonarLint.xml')">
-        $(MSBuildStartupDirectory)\config\$(PROJECT)\SonarLint.xml
-      </ProjectSpecificSonarLintXml>
-      <ProjectSpecificSonarLintXml Condition="!Exists('$(MSBuildStartupDirectory)\config\$(PROJECT)\SonarLint.xml')">
-        $(MSBuildStartupDirectory)\config\SonarLint.xml
-      </ProjectSpecificSonarLintXml>
+      <ProjectSpecificSonarLintXml Condition="Exists('$(MSBuildStartupDirectory)\config\$(PROJECT)\SonarLint.xml')">$(MSBuildStartupDirectory)\config\$(PROJECT)\SonarLint.xml</ProjectSpecificSonarLintXml>
+      <ProjectSpecificSonarLintXml Condition="!Exists('$(MSBuildStartupDirectory)\config\$(PROJECT)\SonarLint.xml')">$(MSBuildStartupDirectory)\config\SonarLint.xml</ProjectSpecificSonarLintXml>
     </PropertyGroup>
 
     <PropertyGroup>
       <SonarAnalyzerRegressionTestRunning Condition="Exists('$(MSBuildStartupDirectory)\binaries\SonarAnalyzer.dll')">true</SonarAnalyzerRegressionTestRunning>
       <SonarAnalyzerRegressionTestRunning Condition="$(SonarAnalyzerRegressionTestRunning) == ''">false</SonarAnalyzerRegressionTestRunning>
       <SonarProjectOutputPath>$(MSBuildStartupDirectory)\output\$(PROJECT)\$(AssemblyName)</SonarProjectOutputPath>
-      <ProjectConfFilePath>$(SonarProjectOutputPath)\ProjectOutFolderPath.txt</ProjectConfFilePath>
+      <SonarProjectConfigFilePath>$(SonarProjectOutputPath)\SonarProjectConfig.xml</SonarProjectConfigFilePath>
     </PropertyGroup>
+
+    <!-- This section builds content of SonarProjectConfig.xml additional file -->
+    <ItemGroup>
+      <SonarProjectConfigLine Include="&lt;?xml version=&quot;1.0&quot; encoding=&quot;utf-8&quot;?&gt;" />
+      <SonarProjectConfigLine Include="&lt;SonarProjectConfig xmlns=&quot;http://www.sonarsource.com/msbuild/analyzer/2021/1&quot;&gt;" />
+      <SonarProjectConfigLine Include="  &lt;AnalysisConfigPath&gt;&lt;/AnalysisConfigPath&gt;" />
+      <SonarProjectConfigLine Include="  &lt;ProjectPath&gt;$(MSBuildProjectFullPath)&lt;/ProjectPath&gt;" />
+      <SonarProjectConfigLine Include="  &lt;FilesToAnalyzePath&gt;&lt;/FilesToAnalyzePath&gt;" />
+      <SonarProjectConfigLine Include="  &lt;OutPath&gt;$(SonarProjectOutputPath)&lt;/OutPath&gt;" />
+      <SonarProjectConfigLine Include="  &lt;TargetFramework&gt;$(TargetFramework)&lt;/TargetFramework&gt;" />
+      <SonarProjectConfigLine Include="&lt;/SonarProjectConfig&gt;" />
+    </ItemGroup>
 
     <MakeDir Condition="$(SonarAnalyzerRegressionTestRunning)"
             Directories="$(SonarProjectOutputPath)"/>
     <WriteLinesToFile Condition="$(SonarAnalyzerRegressionTestRunning)"
-            File="$(SonarProjectOutputPath)\ProjectOutFolderPath.txt"
-            Lines="$(SonarProjectOutputPath)"
-            Overwrite="true"
-            Encoding="Unicode" />
+            File="$(SonarProjectConfigFilePath)"
+            Lines="@(SonarProjectConfigLine)"
+            Overwrite="true" />
 
     <Error Text="Could not find '$(MSBuildStartupDirectory)\binaries\SonarAnalyzer.dll'" Condition="!$(SonarAnalyzerRegressionTestRunning)" />
 
--- a/analyzers/its/SonarAnalyzer.Testing.ImportBefore.targets
+++ b/analyzers/its/SonarAnalyzer.Testing.ImportBefore.targets
@@ -15,8 +15,8 @@
       <SonarProjectOutputPath>$(MSBuildStartupDirectory)\output\$(PROJECT)\$(AssemblyName)</SonarProjectOutputPath>
       <SonarProjectConfigFilePath>$(SonarProjectOutputPath)\SonarProjectConfig.xml</SonarProjectConfigFilePath>
       <!-- Only projects with explicit SonarQubeTestProject=true will be classified as test ones. -->
-      <SonarProjectType condition="$(SonarQubeTestProject) != 'true'">Product</SonarProjectType>
-      <SonarProjectType condition="$(SonarQubeTestProject) == 'true'">Test</SonarProjectType>
+      <SonarProjectType Condition="$(SonarQubeTestProject) != 'true'">Product</SonarProjectType>
+      <SonarProjectType Condition="$(SonarQubeTestProject) == 'true'">Test</SonarProjectType>
     </PropertyGroup>
 
     <!-- This section builds content of SonarProjectConfig.xml additional file -->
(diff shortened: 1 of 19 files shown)


pavel-mikula-sonarsource:
@andrei-epure-sonarsource The failing IT is about to be removed in #4142 

It should be also quite easy to provide scaffolding in the targets to provide `FilesToAnalyze.txt` (either by adding manually crafted `FilesToAnalyze.txt` in each project or by simple task for `dir -with -some -args > FilesToAnalyze.txt` and reference it.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -32,9 +32,9 @@ namespace SonarAnalyzer.Rules
 {
     public abstract class UtilityAnalyzerBase : SonarDiagnosticAnalyzer
     {
-        internal const string ConfigurationAdditionalFile = "ProjectOutFolderPath.txt";
-        internal const string IgnoreHeaderCommentsCSharp = "sonar.cs.ignoreHeaderComments";
-        internal const string IgnoreHeaderCommentsVisualBasic = "sonar.vbnet.ignoreHeaderComments";
+        internal const string ProjectOutFolderPathFileName = "ProjectOutFolderPath.txt"; // Only for backward compatibility with S4MSB <= 5.0
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Comment was moved elsewhere in later commits.
What's with `5.1`? Version `5.1` is the one released this week with new `SonarProjectConfig.xml` file. So `ProjectOutfolderPath.txt` is checked only for `<= 5.0`.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -119,29 +119,21 @@ public abstract class UtilityAnalyzerBase<TMessage> : UtilityAnalyzerBase
...
-                    {
-                        // Make sure the folder exists
-                        Directory.CreateDirectory(WorkDirectoryBasePath);
-                        using var metricsStream = File.Create(pathToWrite);
-                        foreach (var message in messages)
-                        {
-                            message.WriteDelimitedTo(metricsStream);
+                            // Make sure the folder exists
+                            Directory.CreateDirectory(WorkDirectoryBasePath);
+                            using var metricsStream = File.Create(Path.Combine(WorkDirectoryBasePath, FileName));
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: In general yes, and `Path.Combine` is very cheap compared to the IO stuff so I moved it in.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/DisablingRequestValidationTest.cs
```diff
@@ -73,13 +75,24 @@ public class DisablingRequestValidationTest
...
             var compilation = SolutionBuilder.Create().AddProject(AnalyzerLanguage.CSharp).GetCompilation();
             var languageVersion = compilation.LanguageVersionString();
-
-            var allDiagnostics = DiagnosticVerifier.GetDiagnostics(
-                compilation,
-                new CS.DisablingRequestValidation(AnalyzerConfiguration.AlwaysEnabled, rootDirectory),
-                CompilationErrorBehavior.Default).ToList();
+            var oldCulture = Thread.CurrentThread.CurrentCulture;
+            var newCulture = (CultureInfo)oldCulture.Clone();
+            newCulture.NumberFormat.NumberDecimalSeparator = ",";   // decimal.TryParse from the implementation might not recognize 1.2 under different culture
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: It was failing locally on my system => could cause FNs on (non English) build machines.
I needed to change the implementation => I needed to cover it with stable reproducer.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -51,30 +51,31 @@ public abstract class UtilityAnalyzerBase : SonarDiagnosticAnalyzer
...
-                AnalyzeGeneratedCode = PropertiesHelper.ReadAnalyzeGeneratedCodeProperty(settings, language);
-                OutPath = File.ReadAllLines(projectOutputAdditionalFile.Path).FirstOrDefault(l => !string.IsNullOrEmpty(l));
-                if (!string.IsNullOrEmpty(OutPath))
+                lock (parameterReadLock)
                 {
-                    OutPath = Path.Combine(OutPath, language == LanguageNames.CSharp ? "output-cs" : "output-vbnet");
-                    IsAnalyzerEnabled = true;
+                    IgnoreHeaderComments = PropertiesHelper.ReadIgnoreHeaderCommentsProperty(settings, c.Compilation.Language);
+                    AnalyzeGeneratedCode = PropertiesHelper.ReadAnalyzeGeneratedCodeProperty(settings, c.Compilation.Language);
+                    if (!string.IsNullOrEmpty(outPath))
```
andrei-epure-sonarsource, line 69: you are checking outPath for null twice...

if between the first `if` and the `lock`, the value for `outPath` can change, that means that the value for settings can change, too - so both should be checked again inside the `lock`, right?

Also, could you please use the same check in line 63 and line 69 (`IsNullOrEmpty`), otherwise, it makes the code a bit confusing

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/UtilityAnalyzerBaseTests.cs
```diff
@@ -21,31 +21,49 @@
...
+        }
+
+        [TestMethod]
         [TestCategory("Utility")]
-        public void ReadParameters_OutputPath(string language, string expectedWorkDirectoryPath)
+        public void ReadConfig_OutPath_FromSonarProjectConfig_HasPriority()
         {
             // We do not test what is read from the SonarLint file, but we need it
-            var utilityAnalyzer = new TestUtilityAnalyzer(language, @"ResourceTests\SonarLint.xml", @"ResourceTests\ProjectOutFolderPath.txt");
+            var utilityAnalyzer = new TestUtilityAnalyzer(LanguageNames.CSharp, @"ResourceTests\SonarLint.xml", DefaultProjectOutFolderPath, DefaultSonarProjectConfig);
```
andrei-epure-sonarsource, line 64: it's a nitpick, but could you also provide the additional files in a different order, too (make this a `DataTestMethod`)? to make sure it doesn't count?

andrei-epure-sonarsource (commented):
comments for the "Implement" commit

analyzers/its/SonarAnalyzer.Testing.ImportBefore.targets
```diff
@@ -5,28 +5,35 @@
...
+      <SonarProjectConfigLine Include="&lt;/SonarProjectConfig&gt;" />
+    </ItemGroup>
+
     <MakeDir Condition="$(SonarAnalyzerRegressionTestRunning)"
             Directories="$(SonarProjectOutputPath)"/>
     <WriteLinesToFile Condition="$(SonarAnalyzerRegressionTestRunning)"
-            File="$(SonarProjectOutputPath)\ProjectOutFolderPath.txt"
-            Lines="$(SonarProjectOutputPath)"
-            Overwrite="true"
-            Encoding="Unicode" />
```
andrei-epure-sonarsource, line 30: why remove the encoding?

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
pavel-mikula-sonarsource: This is actually a nice catch that I missed and that pop up after deobfuscation.

pavel-mikula-sonarsource: While looking at it, I don't actually see a reason for the `lock` itself as well.

analyzers/its/SonarAnalyzer.Testing.ImportBefore.targets
```diff
@@ -5,46 +5,53 @@
...
-      <ProjectConfFilePath>
-        $(MSBuildStartupDirectory)\output\$(PROJECT)\$(AssemblyName)\ProjectOutFolderPath.txt;$(ProjectSpecificSonarLintXml)
-      </ProjectConfFilePath>
+      <SonarProjectOutputPath>$(MSBuildStartupDirectory)\output\$(PROJECT)\$(AssemblyName)</SonarProjectOutputPath>
+      <SonarProjectConfigFilePath>$(SonarProjectOutputPath)\SonarProjectConfig.xml</SonarProjectConfigFilePath>
     </PropertyGroup>
 
+    <!-- This section builds content of SonarProjectConfig.xml additional file -->
+    <ItemGroup>
+      <SonarProjectConfigLine Include="&lt;?xml version=&quot;1.0&quot; encoding=&quot;utf-8&quot;?&gt;" />
```
andrei-epure-sonarsource, line 21: ProjectType seems to be missing? if it's intentional, please add a comment to mention it

andrei-epure-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/DisablingRequestValidationTest.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: Could you open an issue for this problem?

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
andrei-epure-sonarsource: Yes, it's a read-only operation. 

Even more (and that's for a separate PR/issue, but now we're in the context) - I'm not sure the `FileWriteLock` in the generic class is needed. It's a compilation end method. And each analyzer writes to its specific protobuf file. And we don't expect multiple analyzers to run in parallel.

And regarding this class, we have a performance hotspot in [TokenClassifierBase](https://github.com/SonarSource/sonar-dotnet/blob/b8ec1a6e30f781b949941b92542f7088c39b676f/analyzers/src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs#L67) because we query the SemanticModel for each token, so we would probably improve a lot with some caching. But that's another topic.

analyzers/its/SonarAnalyzer.Testing.ImportBefore.targets
pavel-mikula-sonarsource: It was intentional to let analyzer take the decision. But that's not how it will work after MMF-2298 anyway so it's ready for the future now.

andrei-epure-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/MetricsAnalyzerBase.cs
```diff
@@ -40,13 +40,10 @@ public abstract class MetricsAnalyzerBase : UtilityAnalyzerBase<MetricsInfo>
 
         protected abstract MetricsBase GetMetrics(SyntaxTree syntaxTree, SemanticModel semanticModel);
 
-        protected sealed override MetricsInfo GetMessage(SyntaxTree syntaxTree, SemanticModel semanticModel)
+        protected sealed override MetricsInfo CreateMessage(SyntaxTree syntaxTree, SemanticModel semanticModel)
```
andrei-epure-sonarsource, line 43: clean as you code: should add a test for this, you can create a small syntax tree and check the results, it shouldn't take long...

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/DisablingRequestValidationTest.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Do you mean to review the rest of the codebase? Because it's now fixed & tested for this rule.

PR #4141, opened 2021-03-10
Cleanup around NuGetMetadataFactory

costin-zaharia-sonarsource:
- remove separator at the end of the path
- add comments
- Path.GetFullPath() to always provide full path (useful when debugging)


pavel-mikula-sonarsource (approved):
LGTM

PR #4142, opened 2021-03-10
Use FilesToAnalyze.txt in S5753

### LINKED ISSUE #4111: S5753 subtask -  update rule to get web.config from FilesToAnalyze.txt

csaba-sagi-sonarsource:
Fixes #4111


pavel-mikula-sonarsource (changes requested):
Some minor improvements

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/DisablingRequestValidationTest.cs
```diff
@@ -39,33 +39,41 @@ public class DisablingRequestValidationTest
...
+        public void DisablingRequestValidation_CS_WebConfig(string root)
+        {
+            WriteSonarProjectConfigToDisk(root);
             DiagnosticVerifier.VerifyExternalFile(
                 SolutionBuilder.Create().AddProject(AnalyzerLanguage.CSharp).GetCompilation(),
-                new CS.DisablingRequestValidation(AnalyzerConfiguration.AlwaysEnabled, root),
-                File.ReadAllText(Path.Combine(root, WebConfig)));
+                new CS.DisablingRequestValidation(AnalyzerConfiguration.AlwaysEnabled),
+                File.ReadAllText(Path.Combine(root, WebConfig)),
+                Path.Combine(root, SonarProjectConfig));
```
pavel-mikula-sonarsource, line 75: `WriteSonarProjectConfigToDisk(root);` does `Path.Combine(filesToAnalyzeDirectory, SonarProjectConfig)` inside and then is always followed by `Path.Combine(root, SonarProjectConfig)`. That's redundant.

Let's orchestrate the test like this and let the `Create` method return the full path.
```
DiagnosticVerifier.VerifyExternalFile(
                SolutionBuilder.Create().AddProject(AnalyzerLanguage.CSharp).GetCompilation(),
                new CS.DisablingRequestValidation(AnalyzerConfiguration.AlwaysEnabled),
                File.ReadAllText(Path.Combine(root, WebConfig)),
                CreateSonarProjectConfig(root));
```

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/DisablingRequestValidationTest.cs
```diff
@@ -39,33 +39,41 @@ public class DisablingRequestValidationTest
     {
         private const string AspNetMvcVersion = "5.2.7";
         private const string WebConfig = "Web.config";
+        private const string SonarProjectConfig = "SonarProjectConfig.xml";
```
pavel-mikula-sonarsource, line 42: This can be removed after addressing the other comment, as it will be needed only on a single place.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -144,9 +146,11 @@ private static void DumpActualDiagnostics(string languageVersion, IEnumerable<Di
             var compilationOptions = compilation.Options.WithSpecificDiagnosticOptions(supportedDiagnostics);
             var actualToken = cancellationToken ?? CancellationToken.None;
 
+            var analyzerOptions = string.IsNullOrWhiteSpace(sonarProjectConfigPath)? null : TestHelper.CreateOptions(sonarProjectConfigPath);
```
pavel-mikula-sonarsource, line 149: _Double nitpick:_ EOL before is not needed, previous block build opions. There's a space missing before `?` operator
```suggestion
            var analyzerOptions = string.IsNullOrWhiteSpace(sonarProjectConfigPath) ? null : TestHelper.CreateOptions(sonarProjectConfigPath);
```

pavel-mikula-sonarsource (changes requested):
Final line removal, I mean, polishing :D

pavel-mikula-sonarsource (approved):
LGTM!

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DisablingRequestValidationBase.cs
```diff
@@ -73,7 +65,7 @@ protected override void Initialize(SonarAnalysisContext context)
                         a.ConstructorArguments.Length == 1
                         && a.ConstructorArguments[0].Kind == TypedConstantKind.Primitive
                         && a.ConstructorArguments[0].Value is bool b
-                        && b == false
+                        && !b
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: I confirm it was intentional (I prefer explicit `==false` in a complex expression), but now when I look at it it's not that bad either.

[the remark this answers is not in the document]
andrei-epure-sonarsource: also, in the context of this rule, the "b" is actually the value that appears in tests (literal "false"), not a general boolean code condition.
So this makes the brain twist a little, too.

PR #4143, opened 2021-03-10
Add IT with raised issues on missing files

costin-zaharia-sonarsource:


its/src/test/java/com/sonar/it/csharp/IssuesOnMissingFilesTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.sonarqube.ws.Issues;
+
+import java.util.List;
+
+import static com.sonar.it.csharp.Tests.ORCHESTRATOR;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class IssuesOnMissingFilesTest {
```
andrei-epure-sonarsource, line 35: it needs to be included in the `Tests` list (:

its/src/test/java/com/sonar/it/csharp/IssuesOnMissingFilesTest.java
```diff
@@ -0,0 +1,56 @@
...
+  @ClassRule
+  public static final TemporaryFolder temp = TestUtils.createTempFolder();
+
+  private static final String PROJECT = "ProjectWithIssuesOnMissingFiles";
+  private static BuildResult buildResult;
+
+  @BeforeClass
+  public static void init() throws Exception {
+    TestUtils.reset(ORCHESTRATOR);
+    buildResult = Tests.analyzeProject(temp, PROJECT, null);
```
andrei-epure-sonarsource, line 45: could you also add `/warnaserror:AD0001` in `runMsBuild` in https://github.com/SonarSource/sonar-dotnet/blob/3f0918e3a32c8dc67079b7c5bd7f61ddb858df8d/its/src/test/java/com/sonar/it/shared/TestUtils.java#L148 ?

It would make sure that we never get analysis warnings in our Java ITs - and like this, this test would ensure that there are no uncaught errors for this scenario

andrei-epure-sonarsource (changes requested):
looks good, I have a small suggestion

andrei-epure-sonarsource (approved):
LGTM

PR #4146, opened 2021-03-11
Delete local cache at each Orchestrator reset

andrei-epure-sonarsource:

### DIFF
--- a/its/src/test/java/com/sonar/it/shared/TestUtils.java
+++ b/its/src/test/java/com/sonar/it/shared/TestUtils.java
@@ -86,7 +86,7 @@ public class TestUtils {
       "Read more about how the SonarScanner for .NET detects test projects: https://github.com/SonarSource/sonar-scanner-msbuild/wiki/Analysis-of-product-projects-vs.-test-projects");
   }
 
-  public static Ce.Task getAnalysisWarningsTask(Orchestrator orchestrator, BuildResult buildResult) {
+  private static Ce.Task getAnalysisWarningsTask(Orchestrator orchestrator, BuildResult buildResult) {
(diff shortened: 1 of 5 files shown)


its/src/test/java/com/sonar/it/shared/TestUtils.java
```diff
@@ -84,7 +84,7 @@ public static void verifyGuiTestOnlyProjectAnalysisWarning(Orchestrator orchestr
       "Read more about how the SonarScanner for .NET detects test projects: https://github.com/SonarSource/sonar-scanner-msbuild/wiki/Analysis-of-product-projects-vs.-test-projects");
   }
 
-  public static Ce.Task getAnalysisWarningsTask(Orchestrator orchestrator, BuildResult buildResult) {
+  private static Ce.Task getAnalysisWarningsTask(Orchestrator orchestrator, BuildResult buildResult) {
```
andrei-epure-sonarsource, line 87: I don't want to change the order based on visibility in this file, there are many methods that don't respect it and it would make this PR messy

pavel-mikula-sonarsource (approved):
LGTM

pavel-mikula-sonarsource:
Coming back to this one, each test now does it's reset and the unzip happens cca 30x.
The analyzer changes only when this `addPlugin` is executed. Putting the reset close to this build of orchestrator would do it just 2x and save us little bit of CI performance.

https://github.com/SonarSource/sonar-dotnet/blob/711358e01daf80fef8bb7b0fc76313ae7238de81/its/src/test/java/com/sonar/it/csharp/Tests.java#L72-L83

pavel-mikula-sonarsource:
..just thinking out loud

PR #4147, opened 2021-03-12
S3330 analyze web.config

### LINKED ISSUE #4083: S3330: httpOnlyCookies configuration in web.config
Related issue: https://peach.sonarsource.com/security_hotspots?id=security-expected-issues-dotnet-framework-48&hotspots=AXbxjTvd-8SghKUvbDgK

Code snippet:

```
    public class RSPEC3330HttpOnlyCookiesController : Controller
    {
        [Route("RSPEC3330/HttpOnlyCookies/")]
        public ActionResult Index()
        {
            // By default httpOnly flag is not set to newly created cookies
            // unless if on the web.config file the property httpOnlyCookies is set to true
            // like it is the case for this solution
            // <httpCookies httpOnlyCookies="true" requireSSL="true" />
            HttpCookie MyCookie1 = new HttpCookie("TestHTTPOnlyFlag1");
            Response.Cookies.Add(MyCookie1); // Compliant
            HttpCookie MyCookie2 = new HttpCookie("TestHTTPOnlyFlag2");
            MyCookie2.HttpOnly = false; // Sensitive
            Response.Cookies.Add(MyCookie2);
            return View("~/Views/RSPEC3330/index.cshtml");
        }
    }
```
Setting the attribute can be done either programmatically, or globally via configuration files.
The later one is not taken into consideration

csaba-sagi-sonarsource:
Fixes #4083


csaba-sagi-sonarsource:
Remove .Where(File.Exists) from DisablingRequestValidationBase and move it to Files.ToAnalyze.FindFiles

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/CookieShouldBeHttpOnly.cs
```diff
@@ -50,12 +54,40 @@ public sealed class CookieShouldBeHttpOnly : ObjectShouldBeInitializedCorrectlyB
...
+            foreach (var pages in document.XPathSelectElements("configuration/system.web/httpCookies"))
+            {
+                if (pages.GetAttributeIfBoolValueIs("httpOnlyCookies", true) is { } _)
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
```
pavel-mikula-sonarsource, line 91: This can be singlelined (if I haven't missed something)
```
document.XPathSelectElements("configuration/system.web/httpCookies").Any(x => x.GetAttributeIfBoolValueIs("httpOnlyCookies", true) is { } _);
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/WebConfig/Foo/FilesToAnalyze.txt
```diff
@@ -1 +0,0 @@
-﻿TestCases\WebConfig\Foo\Web.config
```
pavel-mikula-sonarsource, line 1: Moving the files into a rule specific directory made a mess in the commit history. Can you re-do this to get proper git `Rename` flag on those? In one step, do the move. In second step, fix the content inside.

Or maybe even better, remove `FilesToAnalyze.txt` and write them to disk in the test scaffolding phase when there's only one file that exists (at least on them points to a file that doesn't exist.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -101,6 +101,12 @@ public static void VerifyNoExceptionThrown(string path, IEnumerable<DiagnosticAn
         public static void VerifyAnalyzerFromCSharp9Library(string path, DiagnosticAnalyzer diagnosticAnalyzer, IEnumerable<MetadataReference> additionalReferences = null) =>
             VerifyAnalyzer(new[] { path }, new[] { diagnosticAnalyzer }, ParseOptionsHelper.FromCSharp9, CompilationErrorBehavior.Default, OutputKind.DynamicallyLinkedLibrary, additionalReferences);
 
+        public static void VerifyAnalyzer(string path,
+                                          DiagnosticAnalyzer diagnosticAnalyzer,
+                                          IEnumerable<MetadataReference> additionalReferences,
+                                          string sonarProjectConfigPath) =>
+            VerifyAnalyzer(new[] { path }, new[] { diagnosticAnalyzer }, null, CompilationErrorBehavior.Default, OutputKind.DynamicallyLinkedLibrary, additionalReferences, sonarProjectConfigPath);
```
pavel-mikula-sonarsource, line 108: Why not just add optional argument to the next one? It should not break the main logic of these overloads related to easy-to-use `additionalReferences`.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/CookieShouldBeHttpOnly.cs
```diff
@@ -50,12 +54,40 @@ public sealed class CookieShouldBeHttpOnly : ObjectShouldBeInitializedCorrectlyB
...
 
+        protected override Func<bool> DetermineIfDefaultConstructorIsSafe(SonarAnalysisContext context, AnalyzerOptions options)
+        {
+            foreach (var fullPath in context.ProjectConfiguration(options).FilesToAnalyze.FindFiles("web.config"))
+            {
+                var webConfig = File.ReadAllText(fullPath);
+                if (webConfig.Contains("<system.web>") && XmlHelper.ParseXDocument(webConfig) is { } doc
+                    && IsHttpOnlyCookies(doc))
+                {
+                    return () => true;
```
pavel-mikula-sonarsource, line 65: I don't see a reason for this lambda indirection. From the point the decision is taken based on current context, it's either `true` or `false`. `()=>true` is the same state information as `true` so we should be able to remove the indirection.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/CookieShouldBeHttpOnly.cs
```diff
@@ -50,12 +54,40 @@ public sealed class CookieShouldBeHttpOnly : ObjectShouldBeInitializedCorrectlyB
 
         internal CookieShouldBeHttpOnly(IAnalyzerConfiguration analyzerConfiguration) : base(analyzerConfiguration, DiagnosticId, MessageFormat) { }
 
+        protected override Func<bool> DetermineIfDefaultConstructorIsSafe(SonarAnalysisContext context, AnalyzerOptions options)
```
pavel-mikula-sonarsource, line 57: This should do the trick as well
```suggestion
        protected override bool DefaultConstructorIsSafe(SonarAnalysisContext context, AnalyzerOptions options)
```

analyzers/src/SonarAnalyzer.CSharp/Rules/ObjectShouldBeInitializedCorrectlyBase.cs
```diff
@@ -47,6 +51,8 @@ protected override void Initialize(SonarAnalysisContext context)
             context.RegisterCompilationStartAction(
                 ccc =>
                 {
+                    var isDefaultConstructorSafe = DetermineIfDefaultConstructorIsSafe(context, ccc.Options);
```
pavel-mikula-sonarsource, line 54: We should not waste performance investigating `web.config` files when the rule is not active. Let's move this line after the `IsEnabled` check.

analyzers/src/SonarAnalyzer.CSharp/Trackers/CSharpObjectInitializationTracker.cs
```diff
@@ -94,12 +94,12 @@ public class CSharpObjectInitializationTracker
...
 
-        internal bool ShouldBeReported(ObjectCreationExpressionSyntax objectCreation, SemanticModel semanticModel) =>
+        internal bool ShouldBeReported(ObjectCreationExpressionSyntax objectCreation, SemanticModel semanticModel, Func<bool> isDefaultConstructorSafe = null) =>
             IsTrackedType(objectCreation, semanticModel)
-            && !ObjectCreatedWithAllowedValue(objectCreation, semanticModel)
+            && !ObjectCreatedWithAllowedValue(objectCreation, semanticModel, isDefaultConstructorSafe)
             && !IsLaterAssignedWithAllowedValue(objectCreation, semanticModel);
 
-        internal bool ShouldBeReported(AssignmentExpressionSyntax assignment, SemanticModel semanticModel) =>
+        internal bool ShouldBeReported(AssignmentExpressionSyntax assignment, SemanticModel semanticModel, Func<bool> isDefaultConstructorSafe = null) =>
```
pavel-mikula-sonarsource, line 102: This new argument is not used.

analyzers/src/SonarAnalyzer.CSharp/Trackers/CSharpObjectInitializationTracker.cs
```diff
@@ -153,7 +153,7 @@ private bool IsAllowedValue(ExpressionSyntax expression, SemanticModel semanticM
         /// <remarks>
         /// Currently we do not handle the situation with default and named arguments.
         /// </remarks>
-        private bool ObjectCreatedWithAllowedValue(ObjectCreationExpressionSyntax objectCreation, SemanticModel semanticModel)
+        private bool ObjectCreatedWithAllowedValue(ObjectCreationExpressionSyntax objectCreation, SemanticModel semanticModel, Func<bool> isDefaultConstructorSafe = null)
```
pavel-mikula-sonarsource, line 156: Let's make it bool as suggested in other comments.
Let's make it non-optional on all places to force user to think about it and provide explicit value.
```suggestion
        private bool ObjectCreatedWithAllowedValue(ObjectCreationExpressionSyntax objectCreation, SemanticModel semanticModel, bool isDefaultConstructorSafe)
```

analyzers/src/SonarAnalyzer.CSharp/Trackers/CSharpObjectInitializationTracker.cs
```diff
@@ -172,8 +172,8 @@ private bool ObjectCreatedWithAllowedValue(ObjectCreationExpressionSyntax object
             }
             else
             {
-                // if no tracked properties are being explicitly set or passed as arguments, the constructor may still be safe by design
-                return constructorIsSafe;
+                // if no tracked properties are being explicitly set or passed as arguments, check if the default constructor is safe
+                return isDefaultConstructorSafe != null ? isDefaultConstructorSafe() : constructorIsSafe;
```
pavel-mikula-sonarsource, line 176: I want to challenge `constructorIsSafe` field and constructor argument. It looks redundant & confusing now. Is it really still needed in this class now? With the new parameters, I'd say it should not be here.

pavel-mikula-sonarsource (changes requested):
First round, I haven't went through all files yes

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
csaba-sagi-sonarsource: That is what I tried at first but then it is ambiguous between that one and ` public static void VerifyAnalyzer(string path,
                                          DiagnosticAnalyzer diagnosticAnalyzer,
                                          IEnumerable<ParseOptions> options = null,
                                          CompilationErrorBehavior checkMode = CompilationErrorBehavior.Default,
                                          OutputKind outputKind = OutputKind.DynamicallyLinkedLibrary,
                                          IEnumerable<MetadataReference> additionalReferences = null`

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/CookieShouldBeHttpOnly.cs
pavel-mikula-sonarsource: I also wanted to simplify the method name a little with this

pavel-mikula-sonarsource: The tracker have `isDefaultConstructorSafe` argument. Let's unify the wording to `IsDefaultConstructorSafe` here.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/CookieShouldBeHttpOnly.cs
```diff
@@ -50,12 +55,30 @@ public sealed class CookieShouldBeHttpOnly : ObjectShouldBeInitializedCorrectlyB
...
         protected override void Initialize(TrackerInput input)
         {
             var t = CSharpFacade.Instance.Tracker.ObjectCreation;
             t.Track(input,
                 t.MatchConstructor(KnownType.Nancy_Cookies_NancyCookie),
                 t.ExceptWhen(t.ArgumentIsBoolConstant("httpOnly", true)));
         }
+
+        private static bool IsHttpOnlyCookies(XDocument document) =>
+            document.XPathSelectElements("configuration/system.web/httpCookies").Any(x => x.GetAttributeIfBoolValueIs("httpOnlyCookies", true) is { } _);
```
pavel-mikula-sonarsource, line 82: This is the same in less obfuscated way
```suggestion
            document.XPathSelectElements("configuration/system.web/httpCookies").Any(x => x.GetAttributeIfBoolValueIs("httpOnlyCookies", true) != null);
```

pavel-mikula-sonarsource (changes requested):
I reviewed everything except the `TestCases` structure as it's WIP

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestHelper.cs
```diff
@@ -199,5 +204,18 @@ public static AnalyzerOptions CreateOptions(string relativePath)
...
         }
+
+        public static string CreateSonarProjectConfig(string soanrProjectConfigDirectory)
+        {
+            var sonarProjectConfigPath = Path.Combine(soanrProjectConfigDirectory, "SonarProjectConfig.xml");
+            File.WriteAllText(sonarProjectConfigPath, string.Format(ProjectConfigTemplate, soanrProjectConfigDirectory));
+            return sonarProjectConfigPath;
+        }
+
+        public static void CreateFilesToAnalyze(string filesToAnalyzeDirectory, string[] filesToAnalyze)
```
pavel-mikula-sonarsource, line 215: This will save you a bunch of `new [] {`
```suggestion
        public static void CreateFilesToAnalyze(string filesToAnalyzeDirectory, params string[] filesToAnalyze)
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestHelper.cs
```diff
@@ -199,5 +204,18 @@ public static AnalyzerOptions CreateOptions(string relativePath)
             additionalText.Setup(x => x.GetText(default)).Returns(text);
             return new AnalyzerOptions(ImmutableArray.Create(additionalText.Object));
         }
+
+        public static string CreateSonarProjectConfig(string soanrProjectConfigDirectory)
```
pavel-mikula-sonarsource, line 208: Typo
```suggestion
        public static string CreateSonarProjectConfig(string sonarProjectConfigDirectory)
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestHelper.cs
pavel-mikula-sonarsource, line 215: With current state, there's implicit implication that `FilesToAnalyze.txt` is in the same directory. While this was fine in single test, it's not so obvious when it's a helper.

`CreateFilesToAnalyze` should return string with full file name.
`CreateSonarProjectConfig` should receive `filesToAnalyzePath`.

pavel-mikula-sonarsource (changes requested):
Last round of polishing. I'll take a look at changes in TestCases now.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/WebConfig/CookieShouldBeHttpOnly/HttpOnlyCookiesConfig/CookieShouldBeHttpOnly.cs
```diff
@@ -0,0 +1,98 @@
+﻿using System;
```
pavel-mikula-sonarsource, line 1: Also, let's move this file to `TestCases/Hotsposts` to keep all of them together.

pavel-mikula-sonarsource (changes requested):
Last rename/move

pavel-mikula-sonarsource (approved):
LGTM

PR #4151, opened 2021-03-12
Extend S5753 to support Web.{Debug,Release}.config transformations

### LINKED ISSUE #3948: Extend S5753 to support Web.{Debug,Release}.config transformations
RE: #3185 and https://jira.sonarsource.com/browse/RSPEC-5753

Currently it only detects the security hotspot on `web.config` files. However, developers may change the configuration for Release via XDT transformations, and make the web app vulnerable.

Probably we should ignore Debug as in development a less secure config may be used.

https://docs.microsoft.com/en-us/aspnet/web-forms/overview/deployment/visual-studio-web-deployment/web-config-transformations#default-transformation-files

pavel-mikula-sonarsource:
Fixes #3948


csaba-sagi-sonarsource (approved):
LGTM, but please fix the code smell

PR #3608, opened 2020-09-14
bump version for 8.14

andrei-epure-sonarsource:

PR #4162, opened 2021-03-17
Bump version to 8.21.0

costin-zaharia-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.Vsix/source.extension.vsixmanifest
+++ b/analyzers/src/SonarAnalyzer.Vsix/source.extension.vsixmanifest
@@ -1,11 +1,11 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
     <Metadata>
-        <Identity Id="SonarAnalyzer.99d84a73-1512-4999-adcc-29d37452719a" Version="8.20.0.0" Language="en-US" Publisher="SonarSource SA" />
+        <Identity Id="SonarAnalyzer.99d84a73-1512-4999-adcc-29d37452719a" Version="8.21.0.0" Language="en-US" Publisher="SonarSource SA" />
         <DisplayName>SonarAnalyzer for CSharp and VB.NET</DisplayName>
         <Description xml:space="preserve">This extension provides on-the-fly feedback to developers on new bugs and quality issues injected into C# and VB.Net code.</Description>
         <MoreInfo>http://vs.sonarlint.org</MoreInfo>
-        <ReleaseNotes>https://github.com/SonarSource/sonar-dotnet/releases/tag/8.20.0.0</ReleaseNotes>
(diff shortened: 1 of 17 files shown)


analyzers/src/SonarAnalyzer.Vsix/source.extension.vsixmanifest
```diff
@@ -1,11 +1,11 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
     <Metadata>
-        <Identity Id="SonarAnalyzer.99d84a73-1512-4999-adcc-29d37452719a" Version="8.20.0.0" Language="en-US" Publisher="SonarSource SA" />
+        <Identity Id="SonarAnalyzer.99d84a73-1512-4999-adcc-29d37452719a" Version="8.21.0.0" Language="en-US" Publisher="SonarSource SA" />
         <DisplayName>SonarAnalyzer for CSharp and VB.NET</DisplayName>
         <Description xml:space="preserve">This extension provides on-the-fly feedback to developers on new bugs and quality issues injected into C# and VB.Net code.</Description>
         <MoreInfo>http://vs.sonarlint.org</MoreInfo>
-        <ReleaseNotes>https://github.com/SonarSource/sonar-dotnet/releases/tag/8.20.0.0</ReleaseNotes>
+        <ReleaseNotes>https://github.com/SonarSource/sonar-dotnet/releases/tag/8.21.0.0</ReleaseNotes>
```
andrei-epure-sonarsource, line 8: I realize now that this link is broken - I'm not sure we care though for this VSIX project
But maybe we should remove this ?

andrei-epure-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.Vsix/source.extension.vsixmanifest
pavel-mikula-sonarsource: This reminds me similar question here: https://github.com/SonarSource/sonar-dotnet/pull/3608#discussion_r487819710

costin-zaharia-sonarsource: As far as I can tell this is not used at all since we don't deploy the Vsix project. Do you remember if it was used in the past?

pavel-mikula-sonarsource: Let's merge it as is, no harm is done.
Let's spawn a tooling issue to remove it in the future (and see if something breaks without blocking development).

costin-zaharia-sonarsource: Actually, since this is not used, I've removed the release notes link and updated the script which changes the version.

andrei-epure-sonarsource: I think it's fine to remove it

andrei-epure-sonarsource: I don't know what/if it was used for; I would imagine this VSIX was cloned from SLVS and this bit wasn't used at all

PR #4167, opened 2021-03-18
Read ProjectType from the scanner ProjectConfig file

### LINKED ISSUE #4156: Consider `ProjectType` given by the Scanner as authoritative
Currently, we decide in [ProjectTypeHelper](https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/src/SonarAnalyzer.Common/Helpers/ProjectTypeHelper.cs) if the project is MAIN or TEST code, both in SL and when invoked by the Scanner for .NET.

This means the Scanner may decide the project is PRODUCT/MAIN and the analyzer can say it's actually test.

We want to change this and:
- read the ProjectType property from SonarProjectConfig.xml file inside the analyzer and use it as authoritative information about project type. No further detection will be done on analyzer side in this case.
- Analyzer will perform it's own detection based on project references only when SonarProjectConfig.xml is not present (fallback for old scanner and SL).

This is part of https://jira.sonarsource.com/browse/MMF-2298

Tasks:

- [x] Implement with UTs
- [x] Update existing IT 2 projects: one with test assemblies, one without; and the scanner deciding it's PRODUCT/MAIN code for both (with SonarQubeTestProject set to `false`)

andrei-epure-sonarsource:
Fixes #4156


analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
```diff
@@ -44,8 +45,17 @@ internal class ProjectConfigReader
 
         public ProjectConfigReader(SourceText sonarProjectConfig, string logFileName)
```
andrei-epure-sonarsource, line 46: tests need to be updated

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -92,6 +92,35 @@ internal SonarAnalysisContext(AnalysisContext context, IEnumerable<DiagnosticDes
         public bool ShouldAnalyzeGenerated(Compilation c, AnalyzerOptions options) =>
             ShouldAnalyzeGenerated(context, c, options);
 
+        public bool IsTestProject(Compilation c, AnalyzerOptions options) =>
+            IsTestProject(context.TryGetValue, c, options);
+
+        // FIXME: replace with caching at higher level
```
andrei-epure-sonarsource, line 98: @pavel-mikula-sonarsource  - this is only used when reporting issues (invoked in `AnalysisContextExtensions`), and we have a problem.

SyntaxNodeAnalysisContext, SyntaxTreeAnalysisContext, SymbolAnalysisContext, CodeBlockAnalysisContext don't have the `TryGetValue` in them. So we cannot use the cache. 

And this makes the current process _extremely_ inefficient, because we read the config file each time we report via SyntaxNodeAnalysisContext, SyntaxTreeAnalysisContext, SymbolAnalysisContext, CodeBlockAnalysisContext.

I think the best way would be to read the file in SonarDiagnosticAnalyzer.Initialize, but we are missing the `AnalyzerOptions` at that point. 

The other option is a bit messy: we could do this computation in each rule ... in the registration callback... and get whether it's a test project from the cache at that point, and use it later when reporting issues... but this means changing _every_single_rule.

analyzers/src/SonarAnalyzer.Common/Rules/EmptyMethodBase.cs
```diff
@@ -36,6 +36,7 @@ public abstract class EmptyMethodBase<TLanguageKindEnum> : EmptyMethodBase
     {
         protected override void Initialize(SonarAnalysisContext context)
         {
+            // FIXME - could have a lambda to capture the context and use the caching mechanism, with fallback on the Compilation
```
andrei-epure-sonarsource, line 39: This is only for the `EmptyBase` rule...

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -41,14 +43,35 @@ private class TestSetup
         {
             public string Path { get; set; }
             public SonarDiagnosticAnalyzer Analyzer { get; set; }
+            public IEnumerable<MetadataReference> AdditionalReferences { get; set; } = new MetadataReference[] { };
         }
 
+        // Various classes that invoke all the `ReportDiagnosticWhenActive` methods in AnalysisContextExtensions
+        // We mention in comments the type of Context that is used to invoke (directly or indirectly) the `ReportDiagnosticWhenActive` method
         private readonly List<TestSetup> testCases = new List<TestSetup>(new[]
```
andrei-epure-sonarsource, line 51: I don't really like how this test is structured, I'll think about how to refactor it using `[DataTestMethod]`

andrei-epure-sonarsource:
@pavel-mikula-sonarsource at this stage, it's really a rough draft:
- lots of FIXMEs in the code
- tests are needed for the instance method `SonarAnalysisContext.IsTestProject`
- I'm sure lots of styling errors
- many refactorings need to be done to extract static methods from `SonarAnalysisContext` to other classes and make the code more readable
- the .NET ITs are failing (haven't checked them yet)

I did a manual test with a simple project + SonarQube and it worked, however also a Java IT is needed.

So, please avoid comments regarding what is mentioned above.

I left a question of how to deal with reading the config file at ReportDiagnostic level, where we have some problems. That's the main design problem where I need some advice at the comment.

Thanks!

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectType.cs
```diff
@@ -25,6 +25,7 @@ namespace SonarAnalyzer.Helpers
     /// </summary>
     public enum ProjectType
     {
+        None,
```
pavel-mikula-sonarsource, line 28: While reading the usage, I was little puzzled by the meaning of `None`. I suggest this to be clear about it.
```suggestion
        Unknown,
```

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -160,6 +189,31 @@ private static IEnumerable<XElement> ParseXmlSettings(SourceText sourceText)
...
+        private static bool IsTestProject(TryGetValueDelegate<ProjectConfigReader> tryGetValue, Compilation c, AnalyzerOptions options)
+        {
+            ProjectType projectType = ProjectType.None;
+            if (options.AdditionalFiles.FirstOrDefault(IsSonarProjectConfig) is { } sonarProjectConfigXml)
+            {
+                var projectConfigReader = sonarProjectConfigXml.GetText() is { } sourceText
+                    // TryGetValue catches all exceptions from SourceTextValueProvider and returns false when thrown
+                    && tryGetValue(sourceText, ProjectConfigProvider, out var cachedProjectConfigReader)
+                    ? cachedProjectConfigReader
+                    : throw new InvalidOperationException($"File {Path.GetFileName(sonarProjectConfigXml.Path)} has been added as an AdditionalFile but could not be read and parsed.");
```
pavel-mikula-sonarsource, line 201: This whole section should be replaced with something like
```
var config = ProjectConfiguration(options);
if(config.IsScanner){

}
```
I was tempted to add something like `IsScanner => !string.IsNullOrEmpty(OutPath)` but there weren't enough usages before (I refactored the CBDE one differently). 

I'm not very happy with `IsScanner` name, but it might inspire you :)

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
pavel-mikula-sonarsource: Let's do baby steps. Keep it as it is for now. The XML file is intentionally very small => not so expensive to load.

I'll try to find a caching mechanism meanwhile. I'll investigate Roslyn APIs first, but some kid of file-path based logic would do the trick as well.

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
pavel-mikula-sonarsource: That should simplify the whole logic to
```
return config.IsScanner 
    ? config.... 
    :  c?.IsTest() ?? true
```

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
```diff
@@ -44,8 +45,17 @@ internal class ProjectConfigReader
...
         {
-            projectConfig = sonarProjectConfig == null ? ProjectConfig.Empty : ReadContent(sonarProjectConfig, logFileName);
-            projectType = new Lazy<ProjectType>(() => ParseProjectType());
+            if (sonarProjectConfig != null)
```
pavel-mikula-sonarsource, line 48: Let's swap it to positive condition

pavel-mikula-sonarsource (commented):
I didn't take a look at the tests yet, the production part looks good given the current structure.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
pavel-mikula-sonarsource: That won't let you use dynamic stuff like those references :/

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
andrei-epure-sonarsource: Ok.

I thought about it in the past few days and one option is to drop this check from reporting. From our previous discussion, this is only done for the case where one class implements multiple rules, and one rule may be TEST-scope, and another MAIN-scope. And the class gets registered, but when reporting issues, we need to do another check.

Doing this check for _ALL_ rules (most of which are in a single class) seems a waste of logic and an unnecessary complexity. This reporting-level extra check can be done just by the classes which implement multiple rules (and it's only a handful of them), and the other ones can use reporting without checking the project type.

pavel-mikula-sonarsource: This idea LGTM.

analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
```diff
@@ -49,19 +49,21 @@ public static SyntaxTree GetSyntaxTree<TLanguageKindEnum>(this CodeBlockStartAna
         public static SyntaxTree GetSyntaxTree(this SemanticModelAnalysisContext context) =>
             context.SemanticModel.SyntaxTree;
 
-
+        // ToDo: by default, do not verify if the project is Test or not when reporting. This is already being done at rule registration.
```
andrei-epure-sonarsource, line 52: I opened https://github.com/SonarSource/sonar-dotnet/issues/4173 for this

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
andrei-epure-sonarsource: Added card https://github.com/SonarSource/sonar-dotnet/projects/1#card-57504009

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
andrei-epure-sonarsource: I will do this after I get the 2nd round of reviews for the rest of the PR

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -92,6 +92,25 @@ internal SonarAnalysisContext(AnalysisContext context, IEnumerable<DiagnosticDes
         public bool ShouldAnalyzeGenerated(Compilation c, AnalyzerOptions options) =>
             ShouldAnalyzeGenerated(context, c, options);
 
+        public bool IsTestProject(Compilation c, AnalyzerOptions options) =>
+            IsTestProject(context.TryGetValue, c, options);
+
+        public static bool IsTestProjectNotCached(Compilation c, AnalyzerOptions options)
```
pavel-mikula-sonarsource, line 98: `IsTestProjectNotCached` confuses me. It's like `not TestProject cached`.
```suggestion
        public static bool IsTestProjectNoCache(Compilation c, AnalyzerOptions options)
```

analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
pavel-mikula-sonarsource, line 52: Can you manually measure compile time and the impact of this check on some IT project? To measure if it is a problem or not. (Out of scope of this specific PR)

pavel-mikula-sonarsource, line 52: Let's link #4173 in the ToDo comment.

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -92,6 +92,25 @@ internal SonarAnalysisContext(AnalysisContext context, IEnumerable<DiagnosticDes
...
+                sonarProjectConfigXml.GetText() is { } sourceText)
+            {
+                var projectConfigReader =  new ProjectConfigReader(sourceText, SonarProjectConfigFileName);
+                projectType = projectConfigReader.ProjectType;
+            }
+
+            return IsTest(projectType, c);
+        }
+
+        public static bool IsTestProject(CompilationAnalysisContext analysisContext) =>
```
pavel-mikula-sonarsource, line 111: _Code orientation:_ Can you swap this method and the one above it? To have non-static `IsTestProject` followed by this static `IsTestProject` and then the static `IsTestProjectNoCache`?

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -160,6 +166,32 @@ private static IEnumerable<XElement> ParseXmlSettings(SourceText sourceText)
             }
         }
 
+        private static bool IsTestProject(TryGetValueDelegate<ProjectConfigReader> tryGetValue, Compilation c, AnalyzerOptions options) =>
```
pavel-mikula-sonarsource, line 169: _Code orientation:_ Let's group this `IsTestProject` with the others above.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -41,14 +43,35 @@ private class TestSetup
         {
             public string Path { get; set; }
             public SonarDiagnosticAnalyzer Analyzer { get; set; }
+            public IEnumerable<MetadataReference> AdditionalReferences { get; set; } = new MetadataReference[] { };
```
pavel-mikula-sonarsource, line 46: `Enumerable.Empty<...>()` instead?

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -41,14 +43,35 @@ private class TestSetup
...
 
+        // Various classes that invoke all the `ReportDiagnosticWhenActive` methods in AnalysisContextExtensions
+        // We mention in comments the type of Context that is used to invoke (directly or indirectly) the `ReportDiagnosticWhenActive` method
         private readonly List<TestSetup> testCases = new List<TestSetup>(new[]
         {
+            // SyntaxNodeAnalysisContext
+            // S3244 - MAIN and TEST
             new TestSetup { Path = @"TestCases\AnonymousDelegateEventUnsubscribe.cs", Analyzer = new AnonymousDelegateEventUnsubscribe() },
+            // S2699 - TEST only
+            new TestSetup { Path = @"TestCases\TestMethodShouldContainAssertion.MsTest.cs", Analyzer = new TestMethodShouldContainAssertion(), AdditionalReferences = TestMethodShouldContainAssertionTest.GetMsTestReferences(Constants.NuGetLatestVersion)},
```
pavel-mikula-sonarsource, line 57: Explicit constructor in `TestSetup` class can remove a lot of repetitive pain with named arguments.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -224,6 +304,9 @@ public void ProjectConfiguration_WhenInvalidXml_ThrowException()
                 .WithMessage("File SonarProjectConfig.xml has been added as an AdditionalFile but could not be read and parsed.");
         }
 
+        private static IEnumerable<MetadataReference> ConcatReferences(IEnumerable<MetadataReference> additionalReferences) =>
```
pavel-mikula-sonarsource, line 307: ~~The `Concat` was puzzling on the usage side, as it takes only one argument.~~
_Edit:_ See next message in this thread.
```suggestion
        private static IEnumerable<MetadataReference> BuildReferences(IEnumerable<MetadataReference> additionalReferences) =>
```
or
```suggestion
        private static IEnumerable<MetadataReference> CreateReferences(IEnumerable<MetadataReference> additionalReferences) =>
```
or
```suggestion
        private static IEnumerable<MetadataReference> BuildAllReferences(IEnumerable<MetadataReference> additionalReferences) =>
```
or
```suggestion
        private static IEnumerable<MetadataReference> CreateAllReferences(IEnumerable<MetadataReference> additionalReferences) =>
```

pavel-mikula-sonarsource: Let's remove the method from this class and do the magic inside the suggested `TestSetup` constructor. That will make `TestSetup.AdditionalReferences` usable right away.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -83,7 +106,62 @@ public void ByDefault_ExecuteRule()
                 Verifier.VerifyAnalyzer(testCase.Path,
                                         testCase.Analyzer,
                                         ParseOptionsHelper.FromCSharp8,
-                                        MetadataReferenceFacade.SystemComponentModelPrimitives);
+                                        ConcatReferences(testCase.AdditionalReferences));
+            }
+        }
+
+        [TestMethod]
+        public void WhenProjectType_IsTest_RunsOnlyTestRules()
```
pavel-mikula-sonarsource, line 114: Can we improve this name? It currently implies that Main&Test rules are not run. Maybe `WhenProjectType_IsTest_RunRulesWithTestScope`

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -83,7 +106,62 @@ public void ByDefault_ExecuteRule()
...
+                                                   testCase.Analyzer,
+                                                   ParseOptionsHelper.FromCSharp8,
+                                                   ConcatReferences(testCase.AdditionalReferences),
+                                                   sonarProjectConfig);
+                }
+            }
+        }
+
+        [TestMethod]
+        public void WhenProjectType_IsProduct_RunsOnlyMainRules()
```
pavel-mikula-sonarsource, line 142: Same as above

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -83,7 +106,62 @@ public void ByDefault_ExecuteRule()
...
+            }
+        }
+
+        [TestMethod]
+        public void WhenProjectType_IsTest_RunsOnlyTestRules()
+        {
+            var sonarProjectConfig = TestHelper.CreateSonarProjectConfig(Directory.CreateDirectory(@"TestCases\WhenProjectType_IsTest_RunsOnlyTestRules").FullName, ProjectType.Test);
+            foreach (var testCase in testCases)
+            {
+                var isTest = testCase.Analyzer.SupportedDiagnostics.Any(d => d.CustomTags.Contains(DiagnosticDescriptorBuilder.TestSourceScopeTag));
```
pavel-mikula-sonarsource, line 119: `has` to describe the fact that it might be Main&Test and not exclusively Test.
`Scope` to differentiate between rule scope from RSPEC and compilation/scanner property.
```suggestion
                var hasTestScope = testCase.Analyzer.SupportedDiagnostics.Any(d => d.CustomTags.Contains(DiagnosticDescriptorBuilder.TestSourceScopeTag));
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -83,7 +106,62 @@ public void ByDefault_ExecuteRule()
...
+            }
+        }
+
+        [TestMethod]
+        public void WhenProjectType_IsProduct_RunsOnlyMainRules()
+        {
+            var sonarProjectConfig = TestHelper.CreateSonarProjectConfig(Directory.CreateDirectory(@"TestCases\WhenProjectType_IsTest_RunsOnlyTestRules").FullName, ProjectType.Product);
+            foreach (var testCase in testCases)
+            {
+                var isProduct = testCase.Analyzer.SupportedDiagnostics.Any(d => d.CustomTags.Contains(DiagnosticDescriptorBuilder.MainSourceScopeTag));
```
pavel-mikula-sonarsource, line 147: Same as above
```suggestion
                var hasProductScope = testCase.Analyzer.SupportedDiagnostics.Any(d => d.CustomTags.Contains(DiagnosticDescriptorBuilder.MainSourceScopeTag));
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -83,7 +106,62 @@ public void ByDefault_ExecuteRule()
...
+        [TestMethod]
+        public void WhenProjectType_IsTest_RunsOnlyTestRules()
+        {
+            var sonarProjectConfig = TestHelper.CreateSonarProjectConfig(Directory.CreateDirectory(@"TestCases\WhenProjectType_IsTest_RunsOnlyTestRules").FullName, ProjectType.Test);
+            foreach (var testCase in testCases)
+            {
+                var isTest = testCase.Analyzer.SupportedDiagnostics.Any(d => d.CustomTags.Contains(DiagnosticDescriptorBuilder.TestSourceScopeTag));
+                if (isTest)
+                {
+                    // the rules with TEST scope should run
```
pavel-mikula-sonarsource, line 122: Similar comment is not on the test below and it's fine.
```suggestion
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -83,7 +106,62 @@ public void ByDefault_ExecuteRule()
...
+                    Verifier.VerifyAnalyzer(testCase.Path,
+                                            testCase.Analyzer,
+                                            ParseOptionsHelper.FromCSharp8,
+                                            ConcatReferences(testCase.AdditionalReferences),
+                                            sonarProjectConfig);
+                }
+                else
+                {
+                    // MAIN-only rules
```
pavel-mikula-sonarsource, line 131: _Nitpick:_ Just to align with the comment from the test below
```suggestion
                    // MAIN-only rule
```

pavel-mikula-sonarsource (changes requested):
Pack of minor improvements

analyzers/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs
```diff
@@ -125,6 +125,8 @@ private static void CheckTypeName(SyntaxNodeAnalysisContext context)
             var isNameValid = IsTypeNameValid(identifier.ValueText,
                 requireInitialI: typeDeclaration is InterfaceDeclarationSyntax,
                 allowInitialI: typeDeclaration.Modifiers.Any(SyntaxKind.StaticKeyword),
+                // ToDo https://github.com/SonarSource/sonar-dotnet/issues/4172
+                // use SonarAnalysisContext.IsTestProject(c.Compilation, c.Options);
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: Yes, but they behave differently. For TEST code, they allow underscores in the name.

analyzers/src/SonarAnalyzer.Common/Helpers/ProjectConfigReader.cs
```diff
@@ -44,8 +44,17 @@ internal class ProjectConfigReader
...
+            {
+                projectConfig = ProjectConfig.Empty;
+                projectType = new Lazy<ProjectType>(() => ProjectType.Unknown);
+            }
+            else
+            {
+                projectConfig = ReadContent(sonarProjectConfig, logFileName);
+                projectType = new Lazy<ProjectType>(() => ParseProjectType());
+            }
+            // FilesToAnalyzerProvider has the fallback inside
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: nice catch

analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
andrei-epure-sonarsource: I should as a next step

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -161,6 +161,20 @@ internal void RegisterSyntaxNodeAction<TLanguageKindEnum>(Action<SyntaxNodeAnaly
         internal ProjectConfigReader ProjectConfiguration(AnalyzerOptions options) =>
             ProjectConfiguration(context.TryGetValue, options);
 
+        internal static bool IsAnalysisScopeMatching(Compilation compilation, bool isTestProject, IEnumerable<DiagnosticDescriptor> diagnostics)
+        {
+            if (compilation == null)
```
andrei-epure-sonarsource, line 166: I wanted to remove this but the `LineContinuationTest` VB test started failing (no issue raised)

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestHelper.cs
```diff
@@ -208,7 +214,14 @@ public static AnalyzerOptions CreateOptions(string relativePath)
...
             var sonarProjectConfigPath = Path.Combine(sonarProjectConfigDirectory, "SonarProjectConfig.xml");
-            File.WriteAllText(sonarProjectConfigPath, string.Format(ProjectConfigTemplate, filesToAnalyzePath));
+            File.WriteAllText(sonarProjectConfigPath, string.Format(ProjectConfigTemplateWithFilesToAnalyze, filesToAnalyzePath));
+            return sonarProjectConfigPath;
+        }
+
+        public static string CreateSonarProjectConfig(string sonarProjectConfigDirectory, ProjectType projectType)
+        {
+            var sonarProjectConfigPath = Path.Combine(sonarProjectConfigDirectory, "SonarProjectConfig.xml");
+            File.WriteAllText(sonarProjectConfigPath, string.Format(ProjectConfigTemplateWithProjectType, projectType));
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Let's move the `string.Format(ProjectConfigTemplate, name, value)` inside the new private helper method. I wanted only single line of code to be responsible for correct argument handling for this formatting.

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
pavel-mikula-sonarsource: Ah, I missed that it's private. But let's keep as it is. We'll fix that in the future.

pavel-mikula-sonarsource (approved):
LGTM, please address the last opened conversation before merging (improve UT helper a little)

analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
andrei-epure-sonarsource: I did some tests with `NodaTime`, which is running Scanner BEGIN and then `MSBuild.exe /t:"Restore,Rebuild" .\src\NodaTime.sln > out.txt`

This PR
>Time Elapsed 00:00:47.27
Time Elapsed 00:00:46.98
Time Elapsed 00:00:44.83

Master
>Time Elapsed 00:00:51.10
Time Elapsed 00:00:50.30
Time Elapsed 00:00:50.15

I guess I should check with other projects as well, but from an initial test, it seems that (read&cache during initialization + read during reporting) performs better than (iterating through assembly references all the time).

PR #4175, opened 2021-03-23
Update conding style and AZP URL

pavel-mikula-sonarsource:


docs/coding-style.md
```diff
@@ -1,4 +1,4 @@
-# Coding Style
+﻿# Coding Style
```
pavel-mikula-sonarsource, line 1: BOM

docs/coding-style.md
```diff
@@ -22,6 +22,16 @@ Ordering of class members should be the following:
 
 Furthermore, each of these categories should be ordered from higher to lower accessibility level (public, internal, protected, private).
 
+Static fields and properties should be placed before instance ones. 
+
+Static methods are preferred to be after instance methods.
+
+## Naming conventions
+
+Single variable lambdas should use `x` as the variable name (based on lambda calculus λx). Multi variable lambdas should use descriptive names, where `x` can be used for the main iterated item like `(x, index) => ...`.
```
costin-zaharia-sonarsource, line 31: If I remember correctly the agreement was with "can" not "must" 😄

docs/coding-style.md
```diff
@@ -22,6 +22,16 @@ Ordering of class members should be the following:
...
+
+Variable name `sut` (System Under Test) is allowed in unit tests that really tests a single unit (contrary to our usual rule integration unit tests).
```
costin-zaharia-sonarsource, line 33: Instead of `allowed` I would say `recommended`.

costin-zaharia-sonarsource (changes requested):

docs/coding-style.md
pavel-mikula-sonarsource: I don't see `must` in here... ?

docs/coding-style.md
pavel-mikula-sonarsource: (I hijacked your comment to do the change in GitHub UI 😄 )

docs/coding-style.md
costin-zaharia-sonarsource: Ah, sorry "should"

pavel-mikula-sonarsource: No, the suggestion was to prefer x (but not force it with `must`). RFC 2119 says that `should` is like `recommended` that matches the situation.

costin-zaharia-sonarsource: Nice reference 👍. 

Relative to that document I think `May` is the word we want. In the sense that it's preferable to have a descriptive name but in some contexts, when the logic is simple, `x` may be used. At least this is my understanding about the topic. @andrei-epure-sonarsource, @csaba-sagi-sonarsource: WDYT?

csaba-sagi-sonarsource: My opinion is the following (not a strong one.). I would prefer the word recommended with an explanation when it is OK not to use the `x` variable name.

pavel-mikula-sonarsource: I removed the line from this PR and put it in the next "How we work" to clarify.

costin-zaharia-sonarsource (approved):
LGTM!

docs/coding-style.md
andrei-epure-sonarsource: I think it's a matter on agreeing on a vocabulary. I find that "should" doesn't have the nuance of "when it makes sense to do it", because it relies whether the context is clear or not. A bit like comments on the same line - it's fine for a short line. 

but we'll discuss it anyway

Issue #3823, opened 2020-12-09
Remove rule S2758 


costin-zaharia-sonarsource:
It is is deprecated and replaced by [RSPEC-3923](https://jira.sonarsource.com/browse/RSPEC-3923).

See: https://discuss.sonarsource.com/t/c-s2758-deprecated-but-in-sonar-way/6569

PR #4177, opened 2021-03-24
Disable analysis during dev build (keep only live source analysis in the IDE)

andrei-epure-sonarsource:
- currently StyleCop warnings pollute the logs, better no warnings at all
- maven batch mode avoids some unnecessary logs when downloading (see https://stackoverflow.com/a/48929096)

### DIFF
--- a/scripts/build/dev-build.ps1
+++ b/scripts/build/dev-build.ps1
@@ -70,7 +70,6 @@ try {
 
         Invoke-MSBuild $msbuildVersion $solutionName /t:"Restore,Rebuild" `
             /consoleloggerparameters:Summary `
-            /p:WarningLevel=0 `
             /m `
             /p:configuration=$buildConfiguration `
             /p:DeployExtension=false `
@@ -103,7 +102,7 @@ try {
             $skipTests = ""
         }
         Invoke-InLocation ".." {
-            Exec { & mvn clean install --batch-mode -P local-analyzer -D analyzer.configuration=$buildConfiguration $skipTests }
+            Exec { & mvn clean install -P local-analyzer -D analyzer.configuration=$buildConfiguration $skipTests }
         }
     }
 
@@ -115,7 +114,7 @@ try {
 
     $EndDate=(Get-Date)
     $Time=(New-TimeSpan -Start $StartDate -End $EndDate)
-    Write-Host "Time spent: $Time (start at $StartDate, end at $EndDate)"
+    Write-Host "Time spent: $Time"
 
     Write-Host -ForegroundColor Green "SUCCESS: script was successful!"
     exit 0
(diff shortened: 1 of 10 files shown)


scripts/build/dev-build.ps1
```diff
@@ -100,7 +103,7 @@ try {
             $skipTests = ""
         }
         Invoke-InLocation ".." {
-            Exec { & mvn clean install -P local-analyzer -D analyzer.configuration=$buildConfiguration $skipTests }
+            Exec { & mvn clean install --batch-mode -P local-analyzer -D analyzer.configuration=$buildConfiguration $skipTests }
```
pavel-mikula-sonarsource, line 106: I compared the logs before & after and I don't see any change when running in `-buildJava` mode.
What's the difference? 
If it doesn't work as expected, let's remove it.

pavel-mikula-sonarsource (changes requested):

scripts/build/dev-build.ps1
andrei-epure-sonarsource: the difference is when it downloads JARs from the cache  (the "Downloading..." message)- when they are already in the cache, you won't see it.

anyhow, I can remove it as locally we usually have cached assemblies

andrei-epure-sonarsource:
I tested this in  #4186 and it works:

image

scripts/build/dev-build.ps1
```diff
@@ -60,6 +60,8 @@ try {
     Write-Host "Bin folder to use: $binPath"
     Write-Host "MSBuild: ${msbuildVersion}"
 
+    $StartDate=(Get-Date)
```
pavel-mikula-sonarsource, line 63: It works 🎉 
The dev-build shows only 42 warnings now (CS0618, MSB3277).

To come back to the original idea of having clean build logs, we might consider putting the `/p:WarningLevel=0` argument back, as you originally did, to silence the useless noise.

pavel-mikula-sonarsource (approved):
LAwesomeTM!

scripts/build/dev-build.ps1
andrei-epure-sonarsource: I would keep the build warnings, because ideally we should not have them. If we totally remove them from our logs, we'll forget they exist.

Most of the warnings are `'This rule is deprecated.` -related, and this is a good reminder that we should eventually just delete deprecated rules. Like we recently [removed S2758 ](https://github.com/SonarSource/sonar-dotnet/issues/3823), which had been deprecated for two years.

andrei-epure-sonarsource: I opened a [trello card](https://trello.com/c/RF2TaxeP/164-remove-rules-which-have-been-deprecated-before-sq-79-lts) to discuss it in the weekly.

PR #4179, opened 2021-03-24
Refactor ReportDiagnosticWhenActive methods to not check diagnostic scope by default

### LINKED ISSUE #4173: Refactor ReportDiagnosticWhenActive methods to not check diagnostic scope by default
See rationale in this comment: https://github.com/SonarSource/sonar-dotnet/pull/4167#discussion_r598560726

WHAT

The ReportDiagnosticWhenActive* should not verify the scope of the rule by default ( see `analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs`).

Only specific overloads should do it, which should be used by classed with multiple diagnostics (where one diagnostic may be MAIN-only and one TEST-only, so the class logic will be registered all the time).

pavel-mikula-sonarsource:
Fixes #4173

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
@@ -56,27 +56,31 @@ namespace SonarAnalyzer.Helpers
         public static SyntaxTree GetSyntaxTree(this SemanticModelAnalysisContext context) =>
             context.SemanticModel.SyntaxTree;
 
-        // ToDo: by default, do not verify if the project is Test or not when reporting. This is already being done at rule registration.
-        // Only check if it's a Test project for classes with multiple rules.
-        // https://github.com/SonarSource/sonar-dotnet/issues/4173
-        public static void ReportDiagnosticWhenActive(this SyntaxNodeAnalysisContext context, Diagnostic diagnostic) =>
-            ReportDiagnostic(new ReportingContext(context, diagnostic), SonarAnalysisContext.IsTestProjectNoCache(context.Compilation, context.Options));
+        /// <param name="verifyScopeContext">Provide value for this argument only if the class has more than one SupportedDiagnostics.</param>
+        public static void ReportDiagnosticWhenActive(this SyntaxNodeAnalysisContext context, Diagnostic diagnostic, SonarAnalysisContext verifyScopeContext = null) =>
+            ReportDiagnostic(new ReportingContext(context, diagnostic, verifyScopeContext));
 
         public static void ReportDiagnosticWhenActive(this SyntaxTreeAnalysisContext context, Diagnostic diagnostic) =>
-            ReportDiagnostic(new ReportingContext(context, diagnostic), SonarAnalysisContext.IsTestProjectNoCache(null, context.Options));
+            ReportDiagnostic(new ReportingContext(context, diagnostic, null) { IsTestProject = SonarAnalysisContext.IsTestProjectNoCache(null, context.Options) });
 
         public static void ReportDiagnosticWhenActive(this CompilationAnalysisContext context, Diagnostic diagnostic) =>
-            ReportDiagnostic(new ReportingContext(context, diagnostic), SonarAnalysisContext.IsTestProject(context));
+            ReportDiagnostic(new ReportingContext(context, diagnostic, null) { IsTestProject = SonarAnalysisContext.IsTestProject(context)});
 
         public static void ReportDiagnosticWhenActive(this SymbolAnalysisContext context, Diagnostic diagnostic) =>
-            ReportDiagnostic(new ReportingContext(context, diagnostic), SonarAnalysisContext.IsTestProjectNoCache(context.Compilation, context.Options));
+            ReportDiagnostic(new ReportingContext(context, diagnostic, null) { IsTestProject = SonarAnalysisContext.IsTestProjectNoCache(context.Compilation, context.Options)});
 
         public static void ReportDiagnosticWhenActive(this CodeBlockAnalysisContext context, Diagnostic diagnostic) =>
-            ReportDiagnostic(new ReportingContext(context, diagnostic), SonarAnalysisContext.IsTestProjectNoCache(context.SemanticModel?.Compilation, context.Options));
+            ReportDiagnostic(new ReportingContext(context, diagnostic, null) { IsTestProject = SonarAnalysisContext.IsTestProjectNoCache(context.SemanticModel?.Compilation, context.Options)});
 
-        private static void ReportDiagnostic(ReportingContext reportingContext, bool isTestProject)
+        private static void ReportDiagnostic(ReportingContext reportingContext)
         {
-            // This is the new way SonarLint will handle how and what to report...
+            if (reportingContext.IsTestProject.HasValue
+                && !SonarAnalysisContext.IsAnalysisScopeMatching(reportingContext.Compilation, reportingContext.IsTestProject.Value, new[] { reportingContext.Diagnostic.Descriptor }))
+            {
+                return;
+            }
+
+            // This is the current way SonarLint will handle how and what to report...
             if (SonarAnalysisContext.ReportDiagnostic != null)
             {
                 Debug.Assert(SonarAnalysisContext.ShouldDiagnosticBeReported == null, "Not expecting SonarLint to set both the old and the new delegates.");
@@ -85,9 +89,8 @@ namespace SonarAnalyzer.Helpers
             }
 
             // ... but for compatibility purposes we need to keep handling the old-fashioned way. Old SonarLint can be used with latest NuGet.
-            if (SonarAnalysisContext.IsAnalysisScopeMatching(reportingContext.Compilation, isTestProject, new[] { reportingContext.Diagnostic.Descriptor }) &&
-                !VbcHelper.IsTriggeringVbcError(reportingContext.Diagnostic) &&
-                (SonarAnalysisContext.ShouldDiagnosticBeReported?.Invoke(reportingContext.SyntaxTree, reportingContext.Diagnostic) ?? true))
+            if (!VbcHelper.IsTriggeringVbcError(reportingContext.Diagnostic)
+                && (SonarAnalysisContext.ShouldDiagnosticBeReported?.Invoke(reportingContext.SyntaxTree, reportingContext.Diagnostic) ?? true))
             {
                 reportingContext.ReportDiagnostic(reportingContext.Diagnostic);
             }
--- a/analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
@@ -65,13 +65,13 @@ namespace SonarAnalyzer.Helpers
             ReportDiagnostic(new ReportingContext(context, diagnostic, null));
 
         public static void ReportDiagnosticWhenActive(this CompilationAnalysisContext context, Diagnostic diagnostic) =>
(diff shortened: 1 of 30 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/Rules/EmptyMethodTest.cs
```diff
@@ -78,45 +78,20 @@ public class EmptyMethodTest
...
-            Verifier.VerifyAnalyzer(@"TestCases\EmptyMethod.OverrideVirtual.cs",
-                new CS.EmptyMethod(),
-                options: ParseOptionsHelper.FromCSharp8,
-                additionalReferences: null,
-                sonarProjectConfigPath: TestHelper.CreateSonarProjectConfig(
-                    nameof(EmptyMethod_WithVirtualOverride_RaisesIssueForMainAndUnknownProject_CS),
-                    projectType));
+        [TestMethod]
+        public void EmptyMethod_WithVirtualOverride_RaisesIssueForMainProject_CS() =>
+            Verifier.VerifyAnalyzer(@"TestCases\EmptyMethod.OverrideVirtual.cs", new CS.EmptyMethod());
```
pavel-mikula-sonarsource, line 83: _Idea:_ We should use UT for the `Product` vs `Unknown` behavior, we don't need full end-to-end integration for each rule. Like this, it will run in a `Product` mode. There will never be `Unknown` for the rule.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/EmptyMethodTest.cs
```diff
@@ -78,45 +78,20 @@ public class EmptyMethodTest
...
 
         [TestMethod]
         public void EmptyMethod_WithVirtualOverride_DoesNotRaiseIssuesForTestProject_CS() =>
-            Verifier.VerifyNoIssueReported(@"TestCases\EmptyMethod.OverrideVirtual.cs",
-                new CS.EmptyMethod(),
-                ParseOptionsHelper.FromCSharp8,
-                sonarProjectConfigPath: TestHelper.CreateSonarProjectConfig(
-                    nameof(EmptyMethod_WithVirtualOverride_DoesNotRaiseIssuesForTestProject_CS),
-                    ProjectType.Test));
+            Verifier.VerifyNoIssueReportedInTest(@"TestCases\EmptyMethod.OverrideVirtual.cs", new CS.EmptyMethod());
```
pavel-mikula-sonarsource, line 87: We already have this `VerifyNoIssueReportedInTest` method that we can use directly. For the rule, it's not important how it was detected as a test.

If really needed, I can do the SonarProjectConfig magic inside `VerifyNoIssueReportedInTest`, but it will behave the same.

analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
```diff
@@ -56,27 +56,34 @@ public static SyntaxTree GetSyntaxTree<TLanguageKindEnum>(this CodeBlockStartAna
...
+            ReportDiagnostic(new ReportingContext(context, diagnostic, verifyScopeContext));
 
+        // SyntaxTreeAnalysisContext doesn't have a Compilation => verifyScopeContext is never needed, because IsAnalysisScopeMatching returns always true.
         public static void ReportDiagnosticWhenActive(this SyntaxTreeAnalysisContext context, Diagnostic diagnostic) =>
-            ReportDiagnostic(new ReportingContext(context, diagnostic), SonarAnalysisContext.IsTestProjectNoCache(null, context.Options));
+            ReportDiagnostic(new ReportingContext(context, diagnostic, null));
 
         public static void ReportDiagnosticWhenActive(this CompilationAnalysisContext context, Diagnostic diagnostic) =>
-            ReportDiagnostic(new ReportingContext(context, diagnostic), SonarAnalysisContext.IsTestProject(context));
+            ReportDiagnostic(new ReportingContext(context, diagnostic, null) { IsTestProject = SonarAnalysisContext.IsTestProject(context) });
```
pavel-mikula-sonarsource, line 68: I don't like this dual approach (assign IsTestProject property vs provide `verifyScopeContext`), but given current APIs we have, this is the most simple way to do it.

I'll definitely improve this in the future.

pavel-mikula-sonarsource: Alternative way would be to make `SonarAnalysisContext.TryGetValueDelegate` public and change respective `IsTestProject` delegate to public, but that's not nice either.

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
```diff
@@ -175,6 +151,16 @@ internal static bool IsAnalysisScopeMatching(Compilation compilation, bool isTes
             return diagnostics.Any(d => d.CustomTags.Contains(matchingScopeTag));
         }
 
+        private static bool IsTestProject(TryGetValueDelegate<ProjectConfigReader> tryGetValue, Compilation compilation, AnalyzerOptions options)
+        {
+            var projectType = ProjectConfiguration(tryGetValue, options).ProjectType;
+            return projectType == ProjectType.Unknown
+                // SonarLint, NuGet or Scanner <= 5.0. We run the rule if the compilation is null, becase we can't detect test references.
+                ? compilation?.IsTest() ?? true
```
pavel-mikula-sonarsource, line 159: I have a doubt about this `true` fallback. It's answering `IsTestProject` question so it should be true.
I came across this during preparation for the more UTs. As the comment about running the rule is also not relevant to this case (seems like an artifact from the `IsAnalysisScopeMatching` comment)

pavel-mikula-sonarsource:
Rebased to fix the merge conflict

analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
andrei-epure-sonarsource: why not pass directly `bool? isTestProject` to the `ReportingContext` constructor?

the decision can be done inside the `AnalysisContextExtensions` class

Like this, all these decisions are taken here, and `ReportingContext` doesn't need to know about `SonarAnalysisContext`. It at least simplifies the interface, and the logic will be in only one file (currently, it's in two).

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
andrei-epure-sonarsource: Indeed it's from `IsAnalysisScopeMatching`

It shouldn't be `false` or `true`, is should just match the analysis scope. But it's better to be `false`, indeed.

andrei-epure-sonarsource (changes requested):
I suggest a small simplification

analyzers/tests/SonarAnalyzer.UnitTest/Rules/ClassAndMethodNameTest.cs
```diff
@@ -57,22 +58,27 @@ public class ClassAndMethodNameTest
...
+        public void ClassName_VB(ProjectType projectType) =>
+            Verifier.VerifyAnalyzer(@"TestCases\ClassName.vb", new VB.ClassName(), TestHelper.ProjectTypeReference(projectType));
 
-        [TestMethod]
+        [DataTestMethod]
+        [DataRow(ProjectType.Product)]
+        [DataRow(ProjectType.Test)]
         [TestCategory("Rule")]
-        public void MethodName() =>
+        public void MethodName(ProjectType projectType) =>
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: I'm not sure where the `ClassName_InTestProject_CS` went, together with `Verifier.VerifyAnalyzerInTest` which was used inside. `ClassName.Tests.cs` has test-specific behavior which needs to be tested.

Same for `ClassName_TopLevelStatement_InTestProject_CS` and `TestCases\ClassName.TopLevelStatement.Test.cs`

I imagine it's a conflict resolution glitch? 

See them in https://github.com/SonarSource/sonar-dotnet/pull/4181/files#diff-69eaed1afca5f988850cb65d80ca32a8987bbff0cb1b99851a988fb6900c78caR51

andrei-epure-sonarsource (changes requested):
Some files need to get back

analyzers/src/SonarAnalyzer.Common/Helpers/AnalysisContextExtensions.cs
pavel-mikula-sonarsource: I wanted to avoid repetition of the `IsTestProject` evaluation as it's having 5 different set of arguments, but but it's waaay cleaner like this.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/ClassAndMethodNameTest.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: `ClassName_InTestProject_CS`: Test is there with original functionality. I removed `VerifyAnalyzerInTest` (as there were two very complex overloads used only once after this PR) and replaced it with just a `, NuGetMetadataReference.MSTestTestFrameworkV1)` directly inside the test itself.

`ClassName_TopLevelStatement_InTestProject_CS` - this PR didn't do any changes with that test. It's just not visible (collapsed) in GitHub UI as it doesn't have any modification on lines 60-63.

analyzers/src/SonarAnalyzer.Common/Helpers/SonarAnalysisContext.cs
pavel-mikula-sonarsource: Update is on it's way in #4185

analyzers/tests/SonarAnalyzer.UnitTest/Rules/ClassAndMethodNameTest.cs
```diff
@@ -49,7 +49,7 @@ public class ClassAndMethodNameTest
         [TestMethod]
         [TestCategory("Rule")]
         public void ClassName_InTestProject_CS() =>
-            Verifier.VerifyAnalyzerInTest(@"TestCases\ClassName.Tests.cs", new CS.ClassAndMethodName(), ParseOptionsHelper.FromCSharp8);
+            Verifier.VerifyAnalyzer(@"TestCases\ClassName.Tests.cs", new CS.ClassAndMethodName(), ParseOptionsHelper.FromCSharp8, NuGetMetadataReference.MSTestTestFrameworkV1);
```
andrei-epure-sonarsource, line 52: I suggest to use `TestHelper.ProjectTypeReference(ProjectType.Test)`, to keep the decision about the reference centralized

pavel-mikula-sonarsource: That would be parametrized indirection (unneeded complexity). In this specific case, we're testing test-only scenario that doesn't apply to product code and we don't need to go through the specific logic inside `ProjectTypeReference`

analyzers/tests/SonarAnalyzer.UnitTest/Rules/ClassAndMethodNameTest.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: my bad, I opened it locally but it was something maybe out of sync

andrei-epure-sonarsource (approved):
LGTM

andrei-epure-sonarsource:
For the uncovered `ReportDiagnosticWhenActive` cases (I expect when `verifyScopeContext` is null`), would it be easy to add tests?

pavel-mikula-sonarsource:
@andrei-epure-sonarsource No, it's too difficult to scaffold. Let's merge it as is.

PR #4187, opened 2021-03-25
Fix S5944 FP: With NameOf

### LINKED ISSUE #4159: Fix S5944 FP: With NameOf
### Description

S5944 False positive when using NameOf, to avoid using a magic string in the first place.

### Repro steps

    Public Shared Function TestReturn() As String
        Dim result = NameOf(TestReturn)
        Return result
    End Function

### Expected behavior

No warning reported

### Actual behavior

S5944 reported

### Related information

* C#/VB.NET Plugins version 4.33.0.28535
* Visual Studio version 16.9.1
* MSBuild / dotnet version 16.9.0.11203
* Operating System Windows 10 20H2

pavel-mikula-sonarsource:
Fixes #4159

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
@@ -60,7 +60,7 @@ Public Class ImplicitReturnStatementsAreNoncompliant
     ' https//github.com/SonarSource/sonar-dotnet/issues/4159
     Public Function Repro_4159() As String
         Repro_4159 = NameOf(Exception)  ' Noncompliant
-        Return NameOf(Repro_4159)       ' Complant
+        Return NameOf(Repro_4159)       ' Compliant
     End Function
 
 End Class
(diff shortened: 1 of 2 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
```diff
@@ -57,6 +57,12 @@ Public Class ImplicitReturnStatementsAreNoncompliant
     Private Sub WithExplicitArgumentName(Something As Integer)
     End Sub
 
+    ' https//github.com/SonarSource/sonar-dotnet/issues/4159
+    Public Function Repro_4159() As String
+        Repro_4159 = NameOf(Exception)  ' Noncompliant
+        Return NameOf(Repro_4159)       ' Complant
```
andrei-epure-sonarsource, line 63:
```suggestion
        Return NameOf(Repro_4159)       ' Compliant
```

andrei-epure-sonarsource (approved):
LGTM, minor comment to be included before merge

PR #4190, opened 2021-03-25
Add back maven cache

costin-zaharia-sonarsource:
- check if the problem was fixed in the meantime

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -262,10 +262,12 @@ stages:
           artifact: RuleDescriptorBin
           targetPath: analyzers\src\SonarAnalyzer.RuleDescriptorGenerator\bin
 
+      - powershell: .\scripts\build\set-azp-project-version-variable.ps1
+
       - task: Cache@2
         displayName: "Cache Maven local repo"
         inputs:
-          key: 'maven | "$(Agent.OS)" | **/pom.xml'
+          key: 'maven | "$(Agent.OS)" | "$(SONAR_PROJECT_VERSION)" | **/pom.xml'
           path: $(MAVEN_CACHE_FOLDER)
 
       - task: DownloadSecureFile@1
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -262,6 +262,15 @@ stages:
           artifact: RuleDescriptorBin
           targetPath: analyzers\src\SonarAnalyzer.RuleDescriptorGenerator\bin
 
+      - task: Cache@2
+        displayName: "Cache Maven local repo"
+        inputs:
+          key: 'maven | "$(Agent.OS)" | **/pom.xml'
+          restoreKeys: |
+            maven | "$(Agent.OS)"
+            maven
```
pavel-mikula-sonarsource, line 271: This should not be here, as we don't need it (we never populate it with something useful).
```suggestion
```

azure-pipelines.yml
```diff
@@ -262,6 +262,15 @@ stages:
           artifact: RuleDescriptorBin
           targetPath: analyzers\src\SonarAnalyzer.RuleDescriptorGenerator\bin
 
+      - task: Cache@2
+        displayName: "Cache Maven local repo"
+        inputs:
+          key: 'maven | "$(Agent.OS)" | **/pom.xml'
```
pavel-mikula-sonarsource, line 268: We should include the version in the key to avoid infinite grow. With help of the script you refactored recently.

pavel-mikula-sonarsource (changes requested):
Let's add the improvements we prepared in the investigation.

azure-pipelines.yml
```diff
@@ -262,6 +262,14 @@ stages:
           artifact: RuleDescriptorBin
           targetPath: analyzers\src\SonarAnalyzer.RuleDescriptorGenerator\bin
 
+      - powershell: .\scripts\build\set-azp-project-version-variable.ps1
+
+      - task: Cache@1
```
pavel-mikula-sonarsource, line 267: Is it version 1 on purpose? Did you try to play with v2? I don't know what the difference is.

pavel-mikula-sonarsource (approved):
LGTM, there's just a question about version intention. :shipit:

azure-pipelines.yml
costin-zaharia-sonarsource: Yes, I've tried again with version 2 and it failed.

pavel-mikula-sonarsource (commented):
~~LGTM~~ (_edit:_ Bad PR)

PR #4191, opened 2021-03-26
Move RspecResources to LanguageFacade

pavel-mikula-sonarsource:
Refactoring to clean the code base.

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/EnumNameHasEnumSuffix.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/EnumNameHasEnumSuffix.cs
@@ -31,7 +31,5 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class EnumNameHasEnumSuffix : EnumNameHasEnumSuffixBase<SyntaxKind>
     {
         protected override ILanguageFacade<SyntaxKind> Language { get; } = CSharpFacade.Instance;
-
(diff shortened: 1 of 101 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/EnumNameHasEnumSuffix.cs
```diff
@@ -31,7 +31,5 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class EnumNameHasEnumSuffix : EnumNameHasEnumSuffixBase<SyntaxKind>
     {
         protected override ILanguageFacade<SyntaxKind> Language { get; } = CSharpFacade.Instance;
-
-        public EnumNameHasEnumSuffix() : base(RspecStrings.ResourceManager) { }
     }
 }
```
pavel-mikula-sonarsource, line 35: _Definition:_ This is the optimum amount of code needed for a rule with common base 😄  

I promise to stop deleting code from rules that reach the optimum 👼 🤣

csaba-sagi-sonarsource: The optimum would be to use ISourceGenerator and generate all these "boilerplate" classes with just the LanguageFacade. :)

csaba-sagi-sonarsource (approved):
LGTM! Very nicely done. It would be great to add some more common code in the base analyzers like the ILanguageFacade and that way remove a lot more code.

PR #4196, opened 2021-03-26
Clean up the UTs

csaba-sagi-sonarsource:
The aim is to remove as much lines as possible :)

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/CBDE/PreservingEncodingTest.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/CBDE/PreservingEncodingTest.cs
@@ -18,17 +18,18 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+extern alias csharp;
 using System;
 using System.Text;
+using csharp::SonarAnalyzer.CBDE;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
-using CS = SonarAnalyzer.CBDE;
 
 namespace SonarAnalyzer.UnitTest.CBDE
 {
     [TestClass]
     public class PreservingEncodingTest
     {
-        private readonly Encoding encoder = System.Text.Encoding.GetEncoding("ASCII", new CS.PreservingEncodingFallback(), DecoderFallback.ExceptionFallback);
+        private readonly Encoding encoder = System.Text.Encoding.GetEncoding("ASCII", new PreservingEncodingFallback(), DecoderFallback.ExceptionFallback);
 
         public string Encode(string name)
         {
--- a/analyzers/tests/SonarAnalyzer.UnitTest/CBDE/PreservingEncodingTest.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/CBDE/PreservingEncodingTest.cs
@@ -18,19 +18,17 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
- extern alias csharp;
-
 using System;
 using System.Text;
-using csharp::SonarAnalyzer.CBDE;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using CS = SonarAnalyzer.CBDE;
 
 namespace SonarAnalyzer.UnitTest.CBDE
 {
     [TestClass]
     public class PreservingEncodingTest
     {
-        private readonly Encoding encoder = System.Text.Encoding.GetEncoding("ASCII", new PreservingEncodingFallback(), DecoderFallback.ExceptionFallback);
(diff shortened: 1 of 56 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/CBDE/PreservingEncodingTest.cs
```diff
@@ -18,55 +18,44 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
- extern alias csharp;
-
 using System;
 using System.Text;
-using csharp::SonarAnalyzer.CBDE;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using CS = SonarAnalyzer.CBDE;
```
costin-zaharia-sonarsource, line 24: I think this is confusing. CS is usually used for C# but in this case is just an alias for CBDE namespace.

I prefer the initial version.

analyzers/tests/SonarAnalyzer.UnitTest/Common/AnalyzerConfigurationTests.cs
```diff
@@ -27,7 +27,6 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using SonarAnalyzer.Common;
-
```
costin-zaharia-sonarsource, line 30: I think it makes sense to separate static namespaces to make them more visible since the behavior is different.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/DiagnosticVerifierTest.cs
```diff
@@ -18,11 +18,10 @@
 * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 */
 
-extern alias csharp;
 using System;
-using csharp::SonarAnalyzer.Rules.CSharp;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using CS = SonarAnalyzer.Rules.CSharp;
```
costin-zaharia-sonarsource, line 24: The alias is not needed. You can have directly `using SonarAnalyzer.Rules.CSharp;` and remove the prefix from constructor invocations bellow. Please check the other files as well.

costin-zaharia-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/EquivalenceCheckerTest.cs
```diff
@@ -146,9 +146,9 @@ public void EqualityComparer_Node_CS()
             result.Should().BeFalse();
 
             var hashSet = new HashSet<CSharp.BlockSyntax>(new[] { methods_CS.Method1, methods_CS.Method2, methods_CS.Method3 }, comparer);
-            hashSet.Count.Should().Be(2);
-            hashSet.Contains(methods_CS.Method1).Should().BeTrue();
-            hashSet.Contains(methods_CS.Method4).Should().BeFalse();
+            hashSet.Should().HaveCount(2);
+            hashSet.Should().Contain(methods_CS.Method1);
+            hashSet.Should().NotContain(methods_CS.Method4);
```
costin-zaharia-sonarsource, line 151: Nice!

analyzers/tests/SonarAnalyzer.UnitTest/ResourceTests/GeneratedResourcesTest.cs
```diff
@@ -91,32 +91,26 @@ public void ThereShouldBeRuleDetailsForAllVbNetRuleClasses()
...
+            !typeof(UtilityAnalyzerBase).IsAssignableFrom(arg);
 
         private static string[] GetRulesFromResources(string relativePath)
         {
             var resourcesRoot = Path.GetFullPath(Path.Combine(Directory.GetCurrentDirectory(), relativePath));
 
             var resources = Directory.GetFiles(resourcesRoot);
 
-            var ruleNames = resources.Where(IsHtmlFile)
+            return resources.Where(IsHtmlFile)
```
costin-zaharia-sonarsource, line 103: We can move the other method calls to align with `.Where()` call

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/DiagnosticVerifierTest.cs
csaba-sagi-sonarsource: I fixed all non Rule class cases. For the rule class tests the rspec generator should be changed too, so I put it on how we work to align what we want.

costin-zaharia-sonarsource (approved):
LGTM!

Issue #4122, opened 2021-03-05
False Positive: [FromServices]-Attribute produces csharpsquid:S3900 (Arguments of public methods should be validated against null)


fibach:
### Description

The [FromServices] Attribute (https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.fromservicesattribute?view=aspnetcore-5.0) from AspNetCore.Mvc will inject a service via dependency injection into a method of the controller. 

If you forgot to register the service in the DI container, you would get another exception ("InvalidOperationException: No service for type 'XXX' has been registered.")

### Repro steps

Basic setup of dotnet Core MVC project (https://docs.microsoft.com/en-us/aspnet/core/tutorials/first-web-api?view=aspnetcore-5.0&tabs=visual-studio)

- Create Class ProductModelRequestService 
public class ProductModelRequestService
	{
		public string DoIt()
		{
			return "hello world";
		}
	}
- Register in DI container (e.g. in ApiStartup.cs
service.AddTransient<ProductModelRequestService>();
- Add a new Route in the Controller
[HttpGet]
public ProductModel GetProduct([FromServices] ProductModelRequestService productModelRequest)
{
    return productModelRequest.DoIt();
}

### Expected behavior

No Sonarqube warning will csharpsquid:S3900

### Actual behavior
"Arguments of public methods should be validated against null" code smell occures

### Known workarounds

-
EDIT:
Additional Info:
Sonarqube version: Enterprise EditionVersion 8.1 (build 31237)

costin-zaharia-sonarsource:
Hi @fibach, 

Thanks for pointing this out. I can confirm the false positive.

fibach:
@costin-zaharia-sonarsource thanks for tackling this issue. 
It seems you reopened it, but it seems the bug is still on done in the backlog.

KR
florian

csaba-sagi-sonarsource:
Hi @fibach,
The issue was closed by mistake, and that is why it got reopened (and by automation it went to the done column in the backlog).
I moved it back to the false positives column in the backlog.

Thank you very much for reporting the issue in the first place and then for pointing out it was in the wrong column in the backlog.

PR #4197, opened 2021-03-26
Add repro for #4122

costin-zaharia-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PublicMethodArgumentsShouldBeCheckedForNull.CSharp9.cs
```diff
@@ -95,3 +96,19 @@ public partial class Partial
         arg.ToString();   // Noncompliant
     }
 }
+
+namespace UsingAttributes
```
csaba-sagi-sonarsource, line 100: Can you add  the link to the issue as a comment?

csaba-sagi-sonarsource (changes requested):
Please add the link to the issue, otherwise everything is OK.

PR #4202, opened 2021-03-29
S3443: Treat "typeof(Type).GetType()" as an exception

### LINKED ISSUE #4201: Improve S3443: Add exception for ‘typeof(Type).GetType()’ and update rule spec to add exception
### Description

- Fix [rule documentation](https://rules.sonarsource.com/csharp/RSPEC-3443) since `typeof(x).GetType()` returns `System.RuntimeType` not `System.Type`.
- Add exception for `typeof(Type).GetType()`. We can consider this a convention to retrieve `System.RuntimeType` which cannot be directly referenced in the source code.

Source: https://community.sonarsource.com/t/c-s3443-is-wrong/37824

costin-zaharia-sonarsource:
Fixes #4201

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/TypeExaminationOnSystemType.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/TypeExaminationOnSystemType.cs
@@ -25,7 +25,6 @@ using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
-using SonarAnalyzer.Extensions;
 using SonarAnalyzer.Helpers;
 
 namespace SonarAnalyzer.Rules.CSharp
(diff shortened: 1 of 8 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/TypeExaminationOnSystemType.cs
```diff
@@ -33,40 +33,33 @@ namespace SonarAnalyzer.Rules.CSharp
...
-        internal const string MessageGetType = "Remove this use of 'GetType' on a 'System.Type'.";
-        internal const string MessageIsInstanceOfType = "Pass an argument that is not a 'System.Type' or consider using 'IsAssignableFrom'.";
-        internal const string MessageIsInstanceOfTypeWithGetType = "Consider removing the 'GetType' call, it's suspicious in an 'IsInstanceOfType' call.";
+        private const string MessageGetType = "Remove this use of 'GetType' on a 'System.Type'.";
+        private const string MessageIsInstanceOfType = "Pass an argument that is not a 'System.Type' or consider using 'IsAssignableFrom'.";
+        private const string MessageIsInstanceOfTypeWithGetType = "Consider removing the 'GetType' call, it's suspicious in an 'IsInstanceOfType' call.";
 
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        private static readonly DiagnosticDescriptor rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
```
csaba-sagi-sonarsource, line 42: static readonly fields should begin with an upper-case letter.
```suggestion
        private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
```

analyzers/src/SonarAnalyzer.CSharp/Rules/TypeExaminationOnSystemType.cs
```diff
@@ -103,23 +95,22 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            invokedMethod.Name == nameof(Type.GetType)
+            && !invokedMethod.IsStatic
+            && invokedMethod.ContainingType != null
+            && IsObjectOrType(invokedMethod.ContainingType);
+
+        private static bool IsObjectOrType(ITypeSymbol namedType) =>
+            namedType.SpecialType == SpecialType.System_Object
+            || namedType.Is(KnownType.System_Type);
 
         internal static bool IsGetTypeCall(InvocationExpressionSyntax invocation, SemanticModel semanticModel)
```
csaba-sagi-sonarsource, line 115: It should be possible to simplify to something like this:
```suggestion
        internal static bool IsGetTypeCall(InvocationExpressionSyntax invocation, SemanticModel semanticModel) =>
            invocation != null && semanticModel.GetSymbolInfo(invocation).Symbol is IMethodSymbol methodSymbol && IsGetTypeCall(methodSymbol);
```
And It would be nice to consider moving the method to a helper, as other rules use this method too.

csaba-sagi-sonarsource (changes requested):
Overall it is OK, just two minor changes.

csaba-sagi-sonarsource (approved):
LGTM!

PR #4207, opened 2021-03-30
Rule for using a testable (date) time provider

Corniel:
As `DateTime.Now`, `DateTime.UtcNow` and `DateTime.Today` are not stub-able, usage should be discouraged. Detecting the usage is not the challenge obviously, writing down what to do instead in such a way that it helps developers might be.

-------------------------
# Description
One of the principles of a unit test is that it must have full control of the system under test. This is problematic when production code includes calls to the static references `DateTime.Now`, `DateTime.UtcNow` and `DateTime.Today`. There are two different approaches to solves this problem.

## A testable static reference
A static reference that can be adjusted, ideally supports an `IDisposable` method, that not only adjusts the time behaviour for the current thread only, but also for scope of the `using`.
``` C#
public static class Clock
{
    public static DateTime UtcNow();
    public static IDisposable SetTimeForCurrentThread(Func<DateTime> time);
}
```

## Interface
An interface that relies on the (UTC) time provided:
``` C#
public interface IClock
{
    DateTime UtcNow();
}
```

## Compliant code with static reference
``` C#
public class Cookie
{
    public DateTime? ExpirationDate { get; set; }
    public bool HasExpired() => ExpirationDate .HasValue && ExpirationDate .Value < Clock.UtcNow();
}

public class CookieTest
{
    [Test]
    public void Cookie_has_expired_when_expiration_date_in_past()
    {
        using (Clock.SetTimeForCurrentThread(() => new DateTime(2017, 06, 11)))
        {
             Assert.IsTrue(new Cookie { ExpirationDate = new DateTime(2017, 06, 10) }.HasExpired());
        }
    }
}
```

## Compliant code with interface
``` C#
public class Cookie
{
    public DateTime? ExpirationDate { get; set; }
    public bool HasExpired(IClock clock) => ExpirationDate .HasValue && ExpirationDate .Value < clock.UtcNow();
}

public class CookieTest
{
    [Test]
    public void Cookie_has_expired_when_expiration_date_in_past()
    {
        var clock = new TestClock(new DateTime(2017, 06, 11));
        Assert.IsTrue(new Cookie { ExpirationDate = new DateTime(2017, 06, 10) }.HasExpired(clock));
    }
}
```

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Helpers/KnownType.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Helpers/KnownType.cs
@@ -577,6 +577,8 @@ namespace SonarAnalyzer.Helpers
         public string TypeName { get; }
         public string ShortName => shortName.Value;
 
+        public static KnownType Microsoft_Net_Http_Headers_HeaderNames { get; internal set; }
+
         private KnownType(string typeName) : this(SpecialType.None, typeName) { }
 
         private KnownType(SpecialType specialType, string typeName)
--- a/analyzers/src/SonarAnalyzer.Common/Helpers/KnownType.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Helpers/KnownType.cs
@@ -87,7 +87,7 @@ namespace SonarAnalyzer.Helpers
         internal static readonly KnownType Microsoft_Extensions_Logging_EventSourceLoggerFactoryExtensions = new KnownType("Microsoft.Extensions.Logging.EventSourceLoggerFactoryExtensions");
         internal static readonly KnownType Microsoft_Extensions_Logging_ILoggerFactory = new KnownType("Microsoft.Extensions.Logging.ILoggerFactory");
         internal static readonly KnownType Microsoft_Extensions_Primitives_StringValues = new KnownType("Microsoft.Extensions.Primitives.StringValues");
-        internal static readonly KnownType Microsoft_Net_Http_Headers_HeaderNames = new KnownType("Microsoft.Net.Http.Headers.HeaderNames");
+        internal static readonly KnownType Microsoft_VisualBasic_DateAndTime = new KnownType("Microsoft.VisualBasic.DateAndTime");
         internal static readonly KnownType Microsoft_VisualBasic_Interaction = new KnownType("Microsoft.VisualBasic.Interaction");
         internal static readonly KnownType Microsoft_VisualStudio_TestTools_UnitTesting_Assert = new KnownType("Microsoft.VisualStudio.TestTools.UnitTesting.Assert");
         internal static readonly KnownType Microsoft_VisualStudio_TestTools_UnitTesting_AssertFailedException = new KnownType("Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException");
--- a/analyzers/src/SonarAnalyzer.Common/Helpers/KnownType.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Helpers/KnownType.cs
@@ -87,7 +87,7 @@ namespace SonarAnalyzer.Helpers
         internal static readonly KnownType Microsoft_Extensions_Logging_EventSourceLoggerFactoryExtensions = new KnownType("Microsoft.Extensions.Logging.EventSourceLoggerFactoryExtensions");
         internal static readonly KnownType Microsoft_Extensions_Logging_ILoggerFactory = new KnownType("Microsoft.Extensions.Logging.ILoggerFactory");
         internal static readonly KnownType Microsoft_Extensions_Primitives_StringValues = new KnownType("Microsoft.Extensions.Primitives.StringValues");
-        internal static readonly KnownType Microsoft_VisualBasic_DateAndTime = new KnownType("Microsoft.VisualBasic.DateAndTime");
+        internal static readonly KnownType Microsoft_Net_Http_Headers_HeaderNames = new KnownType("Microsoft.Net.Http.Headers.HeaderNames");
         internal static readonly KnownType Microsoft_VisualBasic_Interaction = new KnownType("Microsoft.VisualBasic.Interaction");
         internal static readonly KnownType Microsoft_VisualStudio_TestTools_UnitTesting_Assert = new KnownType("Microsoft.VisualStudio.TestTools.UnitTesting.Assert");
         internal static readonly KnownType Microsoft_VisualStudio_TestTools_UnitTesting_AssertFailedException = new KnownType("Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException");
@@ -576,8 +576,6 @@ namespace SonarAnalyzer.Helpers
 
         public string TypeName { get; }
         public string ShortName => shortName.Value;
-
-        public static KnownType Microsoft_Net_Http_Headers_HeaderNames { get; internal set; }
 
         private KnownType(string typeName) : this(SpecialType.None, typeName) { }
 
--- a/analyzers/src/SonarAnalyzer.Common/Rules/UseTestableTimeProviderBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/UseTestableTimeProviderBase.cs
@@ -28,8 +28,8 @@ namespace SonarAnalyzer.Rules
     public abstract class UseTestableTimeProviderBase<TSyntaxKind> : SonarDiagnosticAnalyzer
          where TSyntaxKind : struct
     {
-        protected const string DiagnosticId = "S6354";
-        protected const string MessageFormat = "Use a testable (date) time provider instead.";
+        private const string DiagnosticId = "S6354";
+        private const string MessageFormat = "Use a testable (date) time provider instead.";
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
         private readonly DiagnosticDescriptor rule;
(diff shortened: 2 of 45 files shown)


Corniel:
**Labels**: bad-practice testability (not sure that is a existing label) 
**Message**: Use a testable (date) time provider instead.
**Default Severity**: Major
**Impact**: :Low/Medium (depends on the solution: interface medium)
**Likelihood**: High
**Default Quality Profiles**: Sonar way (I would argue so)
**Covered Languages**: C#, VB.Net	
**Remediation Function**: ?
**Constant Cost**: 5min/15min (depending on the solution, interface will take more time)
**Analysis Level**: ?
**Syntactic Analysis Analysis Scope**: Main Sources, Test Sources

Corniel:
I did a rebase. @andrei-epure-sonarsource Any update about the prio?

andrei-epure-sonarsource:
sorry @Corniel , I missed this, I will put it on my plate again

analyzers/src/SonarAnalyzer.Common/Rules/UseTestableTimeProviderBase.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        protected abstract ILanguageFacade<TSyntaxKind> Language { get; }
+
+        protected UseTestableTimeProviderBase() =>
+            rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, Language.RspecResources);
+
+        protected sealed override void Initialize(SonarAnalysisContext context) =>
+            context.RegisterSyntaxNodeActionInNonGenerated(Language.GeneratedCodeRecognizer, c =>
+            {
+                if (c.SemanticModel.GetSymbolInfo(c.Node).Symbol is IMethodSymbol identifier
```
andrei-epure-sonarsource, line 45: this will be very expensive in terms of performance - we usually avoid doing symbol retrieval (which triggers a compilation beneath) at the beginning of a rule. 

This rule can easily be syntax-based. You can check the name of the Identifier and MemberAccess (and ignore the edge case of static using), and eventually do a sanity symbol check after verifying it's `DateTime.X`

andrei-epure-sonarsource (commented):
I left a comment for the implementation. 

The rule is useful, although not SonarWay because it would create a lot of noise for developers not interested in making the time provider testable. 

I'll work on the rspec hopefully next week.

analyzers/src/SonarAnalyzer.Common/Rules/UseTestableTimeProviderBase.cs
Corniel: You're right. I think I saw it being done this way elsewhere, but I'll rewrite in such a way that it first checks for the naming only.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseTestableTimeProvider.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
```
andrei-epure-sonarsource, line 1: could you also add a test with a `DateTime` class in a `Foo.Test` namespace and using it (should be compliant)?

Corniel: Good scenarios to add. Did that.

Corniel:
> The rule is useful, although not SonarWay because it would create a lot of noise for developers not interested in making the time provider testable.

I gave this remark some thoughts: why do you think that? I see two options:
1. There is hardly any `DateTime.Now` dependency, so per definition, only limited noise.
2. There is a lot of `DateTime.Now` dependency, so developers should be interested in making the time provider testable.

But you might have a different view?

andrei-epure-sonarsource:
To give an update: I have started working to create a rule spec in the system. It is generic enough to cover most OO languages we support. After I create the rule spec I will ask internal review and post an update here with a link to the PR.

Re: SonarWay. We've discussed internally and the overall agreement was that for large codebases where the good approach has not been used from the beginning, the cost may be considered too high to change to a dependency-injection approach and the issues will be considered noise.

Thanks for the patience.

Corniel:
> To give an update: I have started working to create a rule spec in the system. It is generic enough to cover most OO languages we support. After I create the rule spec I will ask internal review and post an update here with a link to the PR.

I agree that this is brother than just .NET. Nice to see that is addressed in a brother sense.

> Re: SonarWay. We've discussed internally and the overall agreement was that for large codebases where the good approach has not been used from the beginning, the cost may be considered too high to change to a dependency-injection approach and the issues will be considered noise. So we won't include it in SonarWay.

Dependency injection is (agreed) a too big burden for large exciting code bases. Although an approach like this is not: [Clock.cs](https://github.com/Qowaiv/Qowaiv/blob/master/src/Qowaiv/Clock.cs). But I understand, that in the end, that discussion is not up to me, and that is fine.

Corniel:
@andrei-epure-sonarsource Any update on this?

andrei-epure-sonarsource:
Apologies @Corniel , I put this aside for too much time. Will create the RSPEC files for you to pick them up this week.

Corniel:
> Apologies @Corniel , I put this aside for too much time. Will create the RSPEC files for you to pick them up this week.

We all know how those thinks work. On top of that, you intended to add them for non-.NET languages too, what is/was admiral, but also more work.

Corniel:
@andrei-epure-sonarsource Did a rebase, and simplified reading the property name by using `Language.Syntax.NodeIdentifier(Node)`.

analyzers/src/SonarAnalyzer.Common/Helpers/KnownType.cs
```diff
@@ -576,6 +575,8 @@ internal sealed class KnownType
         public string TypeName { get; }
         public string ShortName => shortName.Value;
 
+        public static KnownType Microsoft_Net_Http_Headers_HeaderNames { get; internal set; }
```
andrei-epure-sonarsource, line 578: revert the changes on this file?

andrei-epure-sonarsource (changes requested):
you can pick up the RSPEC files from #5171

Thank you for the patience on this.

Corniel:
Did a rebase.

andrei-epure-sonarsource:
I've updated branch `andrei/s6354-rspec` based on your last comment on the RSPEC @Corniel .

Corniel:
> I've updated branch `andrei/s6354-rspec` based on your last comment on the RSPEC @Corniel .

Can/should I cherry-pick from that branch now?

Corniel:
@andrei-epure-sonarsource Did a re-base and cherrypicked from your branch.

Corniel:
@andrei-epure-sonarsource can we finish this quickly?

Corniel:
@andrei-epure-sonarsource Did a rebase, and updated the tests to use the `VerifyBuilder` (@pavel-mikula-sonarsource).

andrei-epure-sonarsource:
apologies @Corniel I got behind with my github email notifications

andrei-epure-sonarsource:
There are 2 UTs failing and ITs:

- b7d4e8af6183341e506b7446a5bae0ad0d8b9328 has the fix for ITs ([instructions how to run ITs](https://github.com/SonarSource/sonar-dotnet/blob/master/docs/contributing-analyzer.md#integration-tests) - this should work also external contributors, it does not use internal resources)
- bea7ba28351f82e26747209a7ca0e87ff70e49ff has the fix for the UTs (some metadata which gets generated by the [rspec.ps1](https://github.com/SonarSource/sonar-dotnet/blob/master/scripts/rspec/rspec.ps1) script)

both commits on branch `andrei/update-its`.

You'll need to cherry-pick these as well.

I apologize for the delay. I'll keep an eye on this PR in the following days.

Corniel:
@andrei-epure-sonarsource Is this something I did/can fix?

andrei-epure-sonarsource:
You need to cherry-pick the commits I gave in which I fixed the tests.
- the UTs were failing due to missing metadata
- the regression ITs are ran on some open-source projects and tell us when new issues are found (so that we can spot differences when changing a rule) - I created the "expected" json files for the new rule

After you add those commits with fixes on your PR, the CI should be green and ready to merge.

Corniel:
> You need to cherry-pick the commits I gave in which I fixed the tests.
> 
>     * the UTs were failing due to missing metadata for the new rule that you added (we have a script to create a new rule, however that's not usable by external contributors)
> 
>     * the regression ITs are ran on some open-source projects and tell us when new issues are found (so that we can spot differences when changing a rule) - I created the "expected" json files for the new rule
> 
> 
> After you add those commits with fixes on your PR, the CI should be green and ready to merge.

Thanks for you effort. I missed (strangely enough) your first respond. Have been cherry picked (actually fast forwarded, but with the same result).

andrei-epure-sonarsource (approved):
LGTM

AFAICT to have 100% ccov you'd need to add a test case where the property is not named `Now`, `UtcNow` or `Today`

I'm fine to merge as is as well, the test would be a bit superfluous given that anyhow you check the type and there's a regression test for that

andrei-epure-sonarsource:
thanks for your patience and contribution, @Corniel

Corniel:
@andrei-epure-sonarsource You should know me by now. ;)

d-aguilera:
Hello @Corniel 
I'm trying to fix this issue reported by SonarLint, but it seems there is no way around it. Do you have a recommended implementation of `IClock` that would not violate this rule?
It looks like any occurrence of `DateTime.Now` triggers the warning?

Corniel:
@d-aguilera: It is not possible not to trigger this rule if you use `DateTime.Now`. If you have a *home grown* `IClock` interface, you would just have to suppress this once, and I think that should be perfectly fine. The goal is to prevent code that can not be tested.

In 99% of my projects I use a static [Clock](https://github.com/Qowaiv/Qowaiv/blob/master/src/Qowaiv/Clock.cs) , that provides all. As you can see, I also suppress this rule once. 

There are ideas to solve this #5927. However, as explained [here](https://community.sonarsource.com/t/s6354-description-for-testable-clock-provides-no-compliant-implementation/68448/8?u=corniel) I don't think it is possible.

PR #4208, opened 2021-04-01
Import analysis results for test only projects

csaba-sagi-sonarsource:


its/src/test/java/com/sonar/it/shared/TestUtils.java
```diff
@@ -78,14 +78,6 @@ public static void verifyNoGuiWarnings(Orchestrator orchestrator, BuildResult bu
     assertThat(task.getWarningsList()).isEmpty();
   }
 
-  // Verify an AnalysisWarning is raised inside the SQ GUI (on the project dashboard)
-  public static void verifyGuiTestOnlyProjectAnalysisWarning(Orchestrator orchestrator, BuildResult buildResult, String language) {
```
andrei-epure-sonarsource, line 82: I believe this warning should still be kept. If a SQ project (i.e., MSBuild solution) has only TEST code, there probably is a problem with the configuration.

If we don't give this warning, users will be missing MAIN-only expected issues (e.g., vulnerabilities), and will think it's a problem in our analyzer.

its/src/test/java/com/sonar/it/csharp/TestProjectTest.java
```diff
@@ -28,36 +28,50 @@
...
-    assertThat(getComponent(PROJECT).getName()).isEqualTo("TestOnlyProject");
+    assertThat(getComponent(CSHARP_ONLY_TEST_PROJECT).getName()).isEqualTo("TestOnlyProject");
 
     assertThat(getComponent("TestOnlyProject:UnitTest1.cs").getName()).isEqualTo("UnitTest1.cs");
   }
 
   @Test
-  public void logsContainInfoAndWarning() {
-    assertThat(buildResult.getLogs()).contains(
-      "This C# sensor will be skipped, because the current solution contains only TEST files and no MAIN files. " +
```
andrei-epure-sonarsource, line 56: Like the comment for the GUI warning, we should keep a warning in the console when our product thinks there's only test code in a solution.

Example text: "The scanner detected only TEST files in the current solution. Your SonarQube/SonarCloud project will be missing many MAIN-code related issues. Read more..."

its/src/test/java/com/sonar/it/csharp/TestProjectTest.java
```diff
@@ -28,36 +28,50 @@
...
+    assertThat(Tests.getComponent("TestOnlyProject:UnitTest1.cs")).isNotNull();
+    assertThat(getMeasureAsInt(CSHARP_ONLY_TEST_PROJECT, "files")).isNull();
+    assertThat(getMeasureAsInt(CSHARP_ONLY_TEST_PROJECT, "lines")).isNull();
+    assertThat(getMeasureAsInt(CSHARP_ONLY_TEST_PROJECT, "ncloc")).isNull();
+
+    assertThat(buildResult.getLogsLines(l -> l.contains("INFO"))).contains("INFO: Found 1 MSBuild C# project: 1 TEST project.");
+  }
+
+  @Test
+  public void with_html_and_csharp_code_explicitly_marked_as_test_should_not_populate_metrics() throws Exception {
```
andrei-epure-sonarsource, line 69: same here: should have warnings in console and GUI

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
```diff
@@ -76,11 +76,12 @@ public void describe(SensorDescriptor descriptor) {
   public void execute(SensorContext context) {
     FileSystem fs = context.fileSystem();
     boolean hasMainFiles = SensorContextUtils.hasFilesOfType(fs, Type.MAIN, pluginMetadata.languageKey());
+    boolean hasTestFiles = SensorContextUtils.hasFilesOfType(fs, Type.TEST, pluginMetadata.languageKey());
     boolean hasProjects = projectTypeCollector.hasProjects();
-    if (hasMainFiles && hasProjects) {
+    if ((hasMainFiles || hasTestFiles) && hasProjects) {
```
andrei-epure-sonarsource, line 77: Is this possible, to "have projects" and not have code at all? I guess that would be possible with an empty solution... can you add an integration test for this case? Or maybe we have one already?

In any case, I think this is really an edge case (which should be logged inside "importResults", if there are no results (or the results are empty).

It would be simpler to check `boolean hasFiles = SensorContextUtils.hasFilesOfType(fs, pluginMetadata.languageKey());`  and then `if (hasFiles && hasProjects)`.

Maybe it will be easier to talk directly and decide the logic here should be like. Because:
- we'd like to log if in the whole solution there's only TEST files (probably a config error)
- we'd like to log a debug if there are no C#/VB files (probably not scanner for .NET, or a solution with empty projects, or just TypeScript code for example?)

andrei-epure-sonarsource (commented):
First batch of comments

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
andrei-epure-sonarsource: In the end, we may want to keep `(hasMainFiles || hasTestFiles)`, but let's think how to make the code more readable.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/DotNetSensorTest.java
```diff
@@ -210,30 +210,33 @@ public void whenThereAreOnlyMainFilesInAnotherLanguage_logOnlySkipSensor() {
   }
 
   @Test
-  public void whenThereAreOnlyTestFilesInPluginLanguage_andNoMainFilesInAnyLanguage_logConsoleAndAnalysisWarnings() {
+  public void whenThereAreOnlyTestFilesInPluginLanguage_andNoMainFilesInAnyLanguage_resultsAreImportedWithoutWarnings() {
```
andrei-epure-sonarsource, line 206: this should still warn, as mentioned [here](https://github.com/SonarSource/sonar-dotnet/pull/4208#discussion_r607678662)

andrei-epure-sonarsource (changes requested):

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
csaba-sagi-sonarsource: Regarding the analysis of Empty projects. It is not possible to analyze empty projects.  "No analysable projects were found. SonarQube analysis will not be performed."

its/src/test/java/com/sonar/it/csharp/TestProjectTest.java
```diff
@@ -28,36 +28,68 @@
...
 
     assertThat(getComponent("TestOnlyProject:UnitTest1.cs").getName()).isEqualTo("UnitTest1.cs");
   }
 
   @Test
   public void logsContainInfoAndWarning() {
     assertThat(buildResult.getLogs()).contains(
-      "This C# sensor will be skipped, because the current solution contains only TEST files and no MAIN files. " +
-        "Your SonarQube/SonarCloud project will not have results for C# files. " +
+      "The scanner detected only TEST files and no MAIN files in the current solution. " +
```
andrei-epure-sonarsource, line 59:
```suggestion
      "SonarScanner for .NET detected only TEST files and no MAIN files in the current solution. " +
```

its/src/test/java/com/sonar/it/csharp/TestProjectTest.java
```diff
@@ -28,36 +28,68 @@
...
+        "Your SonarQube/SonarCloud project will be missing many C# MAIN-code related issues. " +
```
andrei-epure-sonarsource, line 60: Although I like this warning, I would keep the same message from the GUI. And maybe add 

```suggestion
        "Only TEST-code related results will be imported to your SonarQube/SonarCloud project. Many of our rules (e.g. vulnerabilities) are raised only on MAIN code." +
```

to both. Unfortunately we don't have documentation on this aspect to send the users to (see [internal thread](https://discuss.sonarsource.com/t/rules-sonarsource-com-and-in-product-docs-should-contain-the-rule-scope-main-test/7379/2)), and I think it would be useful to educate them via this message.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/DotNetSensorTest.java
```diff
@@ -124,7 +124,6 @@ public void whenNoProtobufFiles_shouldNotFail() {
     assertThat(logTester.logs(LoggerLevel.DEBUG)).isEmpty();
     assertThat(logTester.logs(LoggerLevel.INFO)).containsExactly("TEST PROJECTS SUMMARY");
     assertThat(logTester.logs(LoggerLevel.WARN)).containsExactly("No protobuf reports found. The " + SHORT_LANG_NAME + " files will not have highlighting and metrics. You can get help on the community forum: https://community.sonarsource.com");
-    verify(analysisWarnings, never()).addUnique(any());
```
andrei-epure-sonarsource, line 127: should bring back? here and everywhere `verify(analysisWarnings, never()).addUnique(any());` was used

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
```diff
@@ -109,41 +115,38 @@ private void importResults(SensorContext context) {
...
       logDebugNoFiles();
     }
   }
 
   private static void warnThatProjectContainsOnlyTestCode(FileSystem fs, AnalysisWarnings analysisWarnings, String languageName) {
-    LOG.warn("This {} sensor will be skipped, because the current solution contains only TEST files and no MAIN files. " +
-        "Your SonarQube/SonarCloud project will not have results for {} files. {}",
-      languageName, languageName, READ_MORE_MESSAGE);
+    LOG.warn("The scanner detected only TEST files and no MAIN files in the current solution. " +
+        "Your SonarQube/SonarCloud project will be missing many {} MAIN-code related issues. {}",
```
andrei-epure-sonarsource, line 141:
```suggestion
        "Only TEST-code related results will be imported to your SonarQube/SonarCloud project. Many of our rules (e.g. vulnerabilities) are raised only on MAIN code. {}",
```

andrei-epure-sonarsource: Unfortunately we don't have documentation on this aspect to send the users to (see [internal thread](https://discuss.sonarsource.com/t/rules-sonarsource-com-and-in-product-docs-should-contain-the-rule-scope-main-test/7379/2)), and I think it would be useful to educate them via this message.

Especially as we know that our security offering is evaluated.

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
```diff
@@ -109,41 +115,38 @@ private void importResults(SensorContext context) {
...
+      languageName, READ_MORE_MESSAGE);
 
     // Before outputting a warning in the User Interface, we want to make sure it's worth the user attention.
     // There can be cases where a project written in language X has tests written in languages X, Y and Z.
     // In this case, the fact that there is only test code for languages Y and Z should not trigger a UI warning.
     if (!SensorContextUtils.hasAnyMainFiles(fs)) {
-      analysisWarnings.addUnique(String.format("Your project contains only TEST code for language %s and no MAIN code for any language, so no results have been imported. %s",
+      analysisWarnings.addUnique(
+        String.format("Your project contains only TEST code for language %s and no MAIN code for any language, so only TEST-code related results are imported. %s",
```
andrei-epure-sonarsource, line 149:
```suggestion
        String.format("Your project contains only TEST code for language %s and no MAIN code for any language, so only TEST-code related results are imported. Many of our rules (e.g. vulnerabilities) are raised only on MAIN code. %s",
```

andrei-epure-sonarsource (changes requested):

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/DotNetSensor.java
andrei-epure-sonarsource: (and also it's good to keep messages similar between console warnings and GUI warnings)

its/src/test/java/com/sonar/it/vbnet/VbMainCodeCsTestCodeTest.java
```diff
@@ -65,14 +66,26 @@ public void hasIssues() {
...
+
     assertThat(buildResult.getLogsLines(l -> l.contains("INFO"))).contains(
       "INFO: Found 1 MSBuild VB.NET project: 1 MAIN project.",
       "INFO: Found 1 MSBuild C# project: 1 TEST project."
     );
     TestUtils.verifyNoGuiWarnings(ORCHESTRATOR, buildResult);
   }
 
+  @Test
+  public void metrics_are_imported_only_for_maint_proj()throws Exception {
```
andrei-epure-sonarsource, line 82:
```suggestion
  public void metrics_are_imported_only_for_main_proj()throws Exception {
```

its/src/test/java/com/sonar/it/vbnet/VbMainCodeCsTestCodeTest.java
```diff
@@ -65,14 +66,26 @@ public void hasIssues() {
...
+    assertThat(getMeasureAsInt("VbMainCsTest:VbMain", "files")).isNotNull();
```
andrei-epure-sonarsource, line 83: it would be more readable to pass the actual values for Test code, rather than "isNotNull" (initially I was wondering what the difference is)

sonar-dotnet-shared-library/src/main/java/org/sonarsource/dotnet/shared/plugins/SensorContextUtils.java
```diff
@@ -43,6 +43,11 @@ public static boolean hasFilesOfType(FileSystem fs, Type fileType, String langua
     return fs.inputFiles(p.and(p.hasType(fileType), p.hasLanguage(languageKey))).iterator().hasNext();
   }
 
+  public static boolean hasFilesOfLanguage(FileSystem fs, String languageKey) {
```
andrei-epure-sonarsource, line 46: this isn't tested in `SensorContextUtilsTest.java`, right? could you add a UT?

andrei-epure-sonarsource (changes requested):

andrei-epure-sonarsource (approved):
LGTM

PR #4211, opened 2021-04-06
Skip CopyPasteToken and Metrics utility analyzers for test projects

csaba-sagi-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -189,6 +190,19 @@ static IEnumerable<TMessage> ReadProtobuf(string path)
             }
         }
 
+        public static void VeriFyUtilityAnalyzerIsNotRun(IEnumerable<string> paths,
```
andrei-epure-sonarsource, line 193:
```suggestion
        public static void VerifyUtilityAnalyzerIsNotRun(IEnumerable<string> paths,
```

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -64,6 +65,7 @@ protected void ReadParameters(SonarAnalysisContext context, CompilationAnalysisC
                 AnalyzeGeneratedCode = PropertiesHelper.ReadAnalyzeGeneratedCodeProperty(settings, c.Compilation.Language);
                 OutPath = Path.Combine(outPath, c.Compilation.Language == LanguageNames.CSharp ? "output-cs" : "output-vbnet");
                 IsAnalyzerEnabled = true;
+                IsTestProject = context.IsTestProject(c.Compilation, c.Options);
```
andrei-epure-sonarsource, line 68: this change also needs an integration test - that's the only way to verify it E2E

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/TokenTypeAnalyzerTest.cs
```diff
@@ -74,24 +74,33 @@ public void Verify(string fileName, Action<IReadOnlyList<TokenTypeInfo.Types.Tok
...
-                });
+                VerifyProtoBuf);
+
+            Verifier.VerifyUtilityAnalyzer<TokenTypeInfo>(
+                new[] { Root + fileName },
+                new TestTokenTypeAnalyzer(testRoot, false),
+                @$"{testRoot}\token-type.pb",
+                VerifyProtoBuf);
+
+            void VerifyProtoBuf(IReadOnlyList<TokenTypeInfo> messages)
```
andrei-epure-sonarsource, line 87:
```suggestion
            void VerifyProtobuf(IReadOnlyList<TokenTypeInfo> messages)
```

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/SymbolReferenceAnalyzerTest.cs
```diff
@@ -125,38 +125,53 @@ public class SymbolReferenceAnalyzerTest
...
-                });
+                VerifyProtoBuf);
+
+            Verifier.VerifyUtilityAnalyzer<SymbolReferenceInfo>(
+                new[] { Root + fileName },
+                testAnalyzer,
+                @$"{testRoot}\symrefs.pb",
+                VerifyProtoBuf);
+
+            void VerifyProtoBuf(IReadOnlyList<SymbolReferenceInfo> messages)
```
andrei-epure-sonarsource, line 148:
```suggestion
            void VerifyProtobuf(IReadOnlyList<SymbolReferenceInfo> messages)
```

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/SymbolReferenceAnalyzerTest.cs
```diff
@@ -125,38 +125,53 @@ public class SymbolReferenceAnalyzerTest
         public void Verify(string fileName, Action<IReadOnlyList<SymbolReferenceInfo.Types.SymbolReference>> verifyReference)
         {
             var testRoot = Root + TestContext.TestName;
-            UtilityAnalyzerBase analyzer = fileName.EndsWith(".cs")
-                ? new TestSymbolReferenceAnalyzer_CS(testRoot)
-                : new TestSymbolReferenceAnalyzer_VB(testRoot);
+            UtilityAnalyzerBase mainAnalyzer = fileName.EndsWith(".cs")
```
andrei-epure-sonarsource, line 128: what's the difference between `mainAnalyzer` and `testAnalyzer`?

Also, if the test scenario is going to fail, all tests are going to fail. This is not a great test setup. A possible refactoring would be to use `DataRow` and merge all `Verify_*_CS` methods into a `Verify_ProductCode_CS` with given parameters.

And then you could have a `Verify_TestCode_CS` with the same parameters, but invoking for test code

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/FileMetadataAnalyzerTest.cs
```diff
@@ -134,27 +136,37 @@ public void CreateMessage_NoEncoding_SetsEmptyString()
...
+                parseOptions);
+
+            Verifier.VerifyUtilityAnalyzer<FileMetadataInfo>(
+                projectFiles,
+                new TestFileMetadataAnalyzer(testRoot, false),
+                @$"{testRoot}\file-metadata.pb",
+                VerifyProtoBuf,
                 parseOptions);
+
+            void VerifyProtoBuf(IReadOnlyList<FileMetadataInfo> messages)
```
andrei-epure-sonarsource, line 151:
```suggestion
            void VerifyProtobuf(IReadOnlyList<FileMetadataInfo> messages)
```

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/FileMetadataAnalyzerTest.cs
```diff
@@ -134,27 +136,37 @@ public void CreateMessage_NoEncoding_SetsEmptyString()
...
-                    }
-                },
+                VerifyProtoBuf,
+                parseOptions);
+
+            Verifier.VerifyUtilityAnalyzer<FileMetadataInfo>(
+                projectFiles,
+                new TestFileMetadataAnalyzer(testRoot, false),
+                @$"{testRoot}\file-metadata.pb",
+                VerifyProtoBuf,
```
andrei-epure-sonarsource, line 148: These lines seem to be the same, except `true/false`, so it can be a local method, to make it easier to read

andrei-epure-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/SymbolReferenceAnalyzerTest.cs
andrei-epure-sonarsource: But feel free to leave the tests as they are if you think it's ok and no refactoring is necessary. I don't have a strong opinion on this one.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/FileMetadataAnalyzerTest.cs
```diff
@@ -114,47 +120,53 @@ public void NotAutogenerated()
...
-                    {
-                        var message = messages[i];
-                        message.IsGenerated.Should().BeTrue();
-                        message.FilePath.Should().Be(autogeneratedFiles[i]);
-                    }
-                },
+                VerifyProtobuf,
                 parseOptions);
+
+            void VerifyProtobuf(IReadOnlyList<FileMetadataInfo> messages)
```
andrei-epure-sonarsource, line 151: this can be static, or you can bring it back inline as before

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/SymbolReferenceAnalyzerTest.cs
```diff
@@ -63,100 +67,125 @@ public class SymbolReferenceAnalyzerTest
...
-                messages =>
-                {
-                    messages.Should().HaveCount(1);
-                    var info = messages.Single();
-                    info.FilePath.Should().Be(fileName);
-                    verifyReference(info.Reference);
-                });
+                VerifyProtobuf);
+
+            void VerifyProtobuf(IReadOnlyList<SymbolReferenceInfo> messages)
```
andrei-epure-sonarsource, line 162: can be static or inline?

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/TokenTypeAnalyzerTest.cs
```diff
@@ -48,50 +50,57 @@ public class TokenTypeAnalyzerTest
...
-                messages =>
-                {
-                    messages.Should().HaveCount(1);
-                    var info = messages.Single();
-                    info.FilePath.Should().Be(fileName);
-                    verifyTokenInfo(info.TokenInfo);
-                });
+                VerifyProtobuf);
+
+            void VerifyProtobuf(IReadOnlyList<TokenTypeInfo> messages)
```
andrei-epure-sonarsource, line 87: static or inline?

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
andrei-epure-sonarsource: as discussed offline, can be done separately as this PR doesn't change the functional behavior

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/SymbolReferenceAnalyzerBase.cs
```diff
@@ -52,6 +52,7 @@ public abstract class SymbolReferenceAnalyzerBase : UtilityAnalyzerBase<SymbolRe
         internal abstract SyntaxNode GetBindableParent(SyntaxToken token);
         protected abstract bool IsIdentifier(SyntaxToken token);
 
+        protected sealed override bool SkipAnalysisForTestProject => false;
```
pavel-mikula-sonarsource, line 55: Can this default value be in the base class and overriden only when needed for those two?

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -84,7 +87,8 @@ public abstract class UtilityAnalyzerBase<TMessage> : UtilityAnalyzerBase
             context.RegisterCompilationAction(c =>
                 {
                     ReadParameters(context, c);
-                    if (!IsAnalyzerEnabled)
+                    if (!IsAnalyzerEnabled ||
+                        (SkipAnalysisForTestProject && IsTestProject))
```
pavel-mikula-sonarsource, line 91: Shouldn't we write an empty file to avoid the warning on the plugin side in this case? To declare we've met this stage and intentionally decided not to produce the data.

andrei-epure-sonarsource (approved):
LGTM

andrei-epure-sonarsource:
putting in progress after Pavel's comments

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -87,12 +87,20 @@ public abstract class UtilityAnalyzerBase<TMessage> : UtilityAnalyzerBase
...
                     ReadParameters(context, c);
-                    if (!IsAnalyzerEnabled ||
-                        (SkipAnalysisForTestProject && IsTestProject))
+                    if (!IsAnalyzerEnabled)
                     {
                         return;
                     }
 
+                    // The results of Metrics and CopyPasteToken analyzers are not needed for Test projects yet the plugin side expects the protobuf files, so we create empty ones.
+                    if (SkipAnalysisForTestProject && IsTestProject)
```
pavel-mikula-sonarsource, line 96: I'd swap the order to first define the context and then the special condition on it.
```suggestion
                    if (IsTestProject && SkipAnalysisForTestProject)
```

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -87,12 +87,20 @@ public abstract class UtilityAnalyzerBase<TMessage> : UtilityAnalyzerBase
...
+                    {
+                        // Make sure the folder exists
+                        Directory.CreateDirectory(OutPath);
+                        File.WriteAllBytes(Path.Combine(OutPath, FileName), new byte[] { });
```
pavel-mikula-sonarsource, line 100: I don't like the bytes there, as it raises the questions what does it mean to write something empty. Let's just create the file.
```suggestion
                        File.Create(Path.Combine(OutPath, FileName)).Dispose();
```

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -87,12 +87,20 @@ public abstract class UtilityAnalyzerBase<TMessage> : UtilityAnalyzerBase
...
-                        (SkipAnalysisForTestProject && IsTestProject))
+                    if (!IsAnalyzerEnabled)
                     {
                         return;
                     }
 
+                    // The results of Metrics and CopyPasteToken analyzers are not needed for Test projects yet the plugin side expects the protobuf files, so we create empty ones.
+                    if (SkipAnalysisForTestProject && IsTestProject)
+                    {
+                        // Make sure the folder exists
```
pavel-mikula-sonarsource, line 98: These 3 lines are here two times and they are not trivial. Let's move them to a method/local function to get the stream. This one will just dispose it, the other one will feed it.

pavel-mikula-sonarsource (changes requested):
Some polishing to avoid duplicates

pavel-mikula-sonarsource (approved):
LGTM

PR #4212, opened 2021-04-08
Fix RegEx hotspots in DotCoverReportParser

### LINKED ISSUE #4194: Fix RegEx hotspots in DotCoverReportParser
Remove backtracking from Regex in `DotCoverReportParser`, as found on SonarCloud:
https://sonarcloud.io/project/security_hotspots?id=org.sonarsource.dotnet%3Asonar-dotnet&sinceLeakPeriod=true#

```
private static final Pattern TITLE_PATTERN = Pattern.compile(".*?<title>(.*?)</title>.*", Pattern.DOTALL);
private static final Pattern COVERED_LINES_PATTERN_1 = Pattern.compile(
    ".*<script type=\"text/javascript\">\\s*+highlightRanges\\(\\[(.*?)\\]\\);\\s*+</script>.*",
    Pattern.DOTALL);
```

andrei-epure-sonarsource:
Fixes #4194

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
@@ -32,10 +32,10 @@ import org.sonar.api.utils.log.Loggers;
 public class DotCoverReportParser implements CoverageParser {
 
   private static final String TITLE_START = "<title>";
-  private static final Pattern COVERED_LINES_PATTERN_1 = Pattern.compile(
+  private static final Pattern FILE_COVERAGE_PATTERN = Pattern.compile(
     ".*<script type=\"text/javascript\">\\s*+highlightRanges\\(\\[([\\d\\[\\],]*?)\\]\\);\\s*+</script>.*",
     Pattern.DOTALL);
-  private static final Pattern COVERED_LINES_PATTERN_2 = Pattern.compile("\\[(\\d++),\\d++,(\\d++),\\d++,(\\d++)\\]");
+  private static final Pattern SEQUENCE_POINT_PATTERN = Pattern.compile("\\[(\\d++),\\d++,(\\d++),\\d++,(\\d++)\\]");
 
   private static final Logger LOG = Loggers.get(DotCoverReportParser.class);
   private final FileService fileService;
@@ -92,11 +92,11 @@ public class DotCoverReportParser implements CoverageParser {
     }
 
     private void collectCoverage(String fileCanonicalPath, String contents) {
-      Matcher matcher = COVERED_LINES_PATTERN_1.matcher(contents);
+      Matcher matcher = FILE_COVERAGE_PATTERN.matcher(contents);
       checkMatches(matcher);
       String highlightedContents = matcher.group(1);
 
-      matcher = COVERED_LINES_PATTERN_2.matcher(highlightedContents);
+      matcher = SEQUENCE_POINT_PATTERN.matcher(highlightedContents);
 
       while (matcher.find()) {
         int lineStart = Integer.parseInt(matcher.group(1));
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
@@ -32,8 +32,6 @@ import org.sonar.api.utils.log.Loggers;
 public class DotCoverReportParser implements CoverageParser {
 
   private static final String TITLE_START = "<title>";
-  private static final String TITLE_END = "</title>";
-  private static final String NO_TAG_MESSAGE = "The report does not contain a '%s' tag.";
   // the pattern for the information about a sequence point - (lineStart, columnStart, lineEnd, columnEnd, hits)
   private static final Pattern SEQUENCE_POINT = Pattern.compile("\\[(\\d++),\\d++,\\d++,\\d++,(\\d++)]");
   private static final String SEQUENCE_POINTS_GROUP_NAME = "SequencePoints";
@@ -85,7 +83,9 @@ public class DotCoverReportParser implements CoverageParser {
 
     @Nullable
     private String extractFileCanonicalPath(String contents) {
-      String lowerCaseAbsolutePath = getTitlePath(contents);
+      int indexOfTitleStart = getIndexOf(contents, TITLE_START, 0);
+      int indexOfTitleEnd = getIndexOf(contents, "</title>", indexOfTitleStart);
+      String lowerCaseAbsolutePath = contents.substring(indexOfTitleStart + TITLE_START.length(), indexOfTitleEnd);
       try {
         return new File(lowerCaseAbsolutePath).getCanonicalPath();
       } catch (IOException e) {
@@ -114,17 +114,12 @@ public class DotCoverReportParser implements CoverageParser {
       }
     }
 
-    // fileContent could contain <title>foo</title> or </title>foo<title>
-    private String getTitlePath(String fileContent) {
-      int indexOfTitleStart = fileContent.indexOf(TITLE_START);
-      if (indexOfTitleStart == -1) {
-        throw new IllegalArgumentException(String.format(NO_TAG_MESSAGE, TITLE_START));
+    private int getIndexOf(String fileContent, String tag, int startIndex) {
+      int index = fileContent.indexOf(tag, startIndex);
+      if (index == -1) {
+        throw new IllegalArgumentException("The report does not contain a vald '<title>...</title>' tag.");
       }
-      int indexOfTitleEnd = fileContent.indexOf(TITLE_END, indexOfTitleStart);
-      if (indexOfTitleEnd == -1) {
-        throw new IllegalArgumentException(String.format(NO_TAG_MESSAGE, TITLE_END));
-      }
-      return fileContent.substring(indexOfTitleStart + TITLE_START.length(), indexOfTitleEnd);
+      return index;
     }
   }
 }
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
@@ -31,9 +31,8 @@ import org.sonar.api.utils.log.Loggers;
 
 public class DotCoverReportParser implements CoverageParser {
 
-  private static final Pattern TITLE_PATTERN = Pattern.compile(".*?<title>(.*?)</title>.*", Pattern.DOTALL);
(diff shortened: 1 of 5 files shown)


sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
```diff
@@ -31,9 +31,8 @@
 
 public class DotCoverReportParser implements CoverageParser {
 
-  private static final Pattern TITLE_PATTERN = Pattern.compile(".*?<title>(.*?)</title>.*", Pattern.DOTALL);
   private static final Pattern COVERED_LINES_PATTERN_1 = Pattern.compile(
-    ".*<script type=\"text/javascript\">\\s*+highlightRanges\\(\\[(.*?)\\]\\);\\s*+</script>.*",
+    ".*<script type=\"text/javascript\">\\s*+highlightRanges\\(\\[([\\d\\[\\],]*?)\\]\\);\\s*+</script>.*",
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: TBH I don't understand the purpose of `COVERED_LINES_PATTERN_1` and `COVERED_LINES_PATTERN_2`, better naming could help. But the v2 seems prettly similar to what my `FRAGMENT` does.

pavel-mikula-sonarsource (changes requested):
Some improvements suggested

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
```diff
@@ -32,10 +32,14 @@
 public class DotCoverReportParser implements CoverageParser {
 
   private static final String TITLE_START = "<title>";
+  // the pattern for the information about a sequence point - (lineStart, columnStart, lineEnd, columnEnd, hits)
+  private static final Pattern SEQUENCE_POINT_PATTERN = Pattern.compile("\\[(\\d++),\\d++,(\\d++),\\d++,(\\d++)]");
```
andrei-epure-sonarsource, line 36: I realize now that we don't really use the 2nd captured number - it's lineEnd but we don't use it

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
```diff
@@ -93,27 +96,39 @@ private String extractFileCanonicalPath(String contents) {
...
 
         coverage.addHits(fileCanonicalPath, lineStart, hits);
 
         LOG.trace("dotCover parser: found coverage for line '{}', hits '{}' when analyzing the path '{}'.",
-            lineStart, hits, fileCanonicalPath);
+          lineStart, hits, fileCanonicalPath);
+      }
+    }
+
+    private int getIndexOf(String fileContent, String tag) {
```
pavel-mikula-sonarsource, line 118: We could use `public int indexOf(String str, int fromIndex)` overload.
It will start from 0 for the `<title>`
It will start from `<title>` to search for `</title>`, ensuring we have consistent state.

Like this, we can remove `getTitlePath`, as the check will be naturally embedded in the system.
```suggestion
    private int getIndexOf(String fileContent, String tag, int startIndex = 0) {
```

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
```diff
@@ -31,11 +31,15 @@
...
-  private static final Pattern COVERED_LINES_PATTERN_1 = Pattern.compile(
-    ".*<script type=\"text/javascript\">\\s*+highlightRanges\\(\\[(.*?)\\]\\);\\s*+</script>.*",
+  private static final String TITLE_START = "<title>";
+  // the pattern for the information about a sequence point - (lineStart, columnStart, lineEnd, columnEnd, hits)
+  private static final Pattern SEQUENCE_POINT_PATTERN = Pattern.compile("\\[(\\d++),\\d++,(\\d++),\\d++,(\\d++)]");
+  private static final String SEQUENCE_POINTS_GROUP_NAME = "SequencePoints";
+  // the file coverage has a list of sequence points
+  // we use SEQUENCE_POINT_PATTERN below to avoid non-determinism in the regular expression, due to the `[` and `]` characters appearing multiple times
+  private static final Pattern FILE_COVERAGE_PATTERN = Pattern.compile(
+    ".*<script type=\"text/javascript\">\\s*+highlightRanges\\(\\[(?<" + SEQUENCE_POINTS_GROUP_NAME + ">" + SEQUENCE_POINT_PATTERN + "(," + SEQUENCE_POINT_PATTERN + ")*)]\\);\\s*+</script>.*",
```
pavel-mikula-sonarsource, line 41: I think the trailing `]` should be escaped
```suggestion
    ".*<script type=\"text/javascript\">\\s*+highlightRanges\\(\\[(?<" + SEQUENCE_POINTS_GROUP_NAME + ">" + SEQUENCE_POINT_PATTERN + "(," + SEQUENCE_POINT_PATTERN + ")*)\\]\\);\\s*+</script>.*",
```

pavel-mikula-sonarsource (changes requested):
Regex nitpick and small improvement suggestion

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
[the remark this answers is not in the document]
pavel-mikula-sonarsource: The new version and naming is awesome! It's pretty clear now and it actually all fits together now.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
andrei-epure-sonarsource: I fixed a code smell by not escaping it. and the unit tests pass.

could you help with a test case to cover adding it back?

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/DotCoverReportParserTest.java
```diff
@@ -55,17 +53,56 @@ public void prepare() {
 
   @Test
   public void no_title() {
-    thrown.expect(IllegalArgumentException.class);
-    thrown.expectMessage("The report contents does not match the following regular expression: .*?<title>(.*?)</title>.*");
-    new DotCoverReportParser(alwaysTrue).accept(new File("src/test/resources/dotcover/no_title.html"), mock(Coverage.class));
+    DotCoverReportParser parser = new DotCoverReportParser(alwaysTrue);
```
pavel-mikula-sonarsource, line 56: Codesmell: SL complains about merging these 3 tests to a single, parametrized one.

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
pavel-mikula-sonarsource: Hmm, it shouldn't be there indeed. I've learned something new 🙂

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
pavel-mikula-sonarsource: I meant it like this to respect previous comment about `getIndexOf`.
```
    private int getIndexOf(String fileContent, String tag, int startIndex) {
      int index = fileContent.indexOf(tag, startIndex);
      if (index == -1) {
        throw new IllegalArgumentException(String.format("The report does not contain a '%s' tag.", tag));
      }
      return index;
    }
```
and usage like
```
      int indexOfTitleStart = getIndexOf(contents, TITLE_START);
      int indexOfTitleEnd = getIndexOf(contents, "</title>", indexOfTitleStart);
      String lowerCaseAbsolutePath = fileContent.substring(indexOfTitleStart + TITLE_START.length(), indexOfTitleEnd);

```
We won't need so many constants as well, as stuff would not be repeated.

pavel-mikula-sonarsource: The </title><title> is such a corner case, that we can live with current error message. We could also update it to something like hardcoded `..not contain valid <title> tag..`

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/DotCoverReportParserTest.java
andrei-epure-sonarsource: I know, but I'd have to update to JUnit5 for this and I don't want to do it in this PR.

The JUnit4 way of dealing with parameterized tests is depressing. I opened #4223 for updating to JUnit5.

andrei-epure-sonarsource: (tried in some PR before to update the JUnit version, either for UTs or ITs, and there was a problem at that point - maybe a small one... so it's better to have a separate ticket)

pavel-mikula-sonarsource (changes requested):
I've meanwhile added a comment to the `indexOf` thread above

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
andrei-epure-sonarsource: (sorry, I misread your comment initially, makes sense)

sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/DotCoverReportParser.java
```diff
@@ -93,27 +95,31 @@ private String extractFileCanonicalPath(String contents) {
...
       }
     }
 
-    private void checkMatches(Matcher matcher) {
-      if (!matcher.matches()) {
-        throw new IllegalArgumentException("The report contents does not match the following regular expression: " + matcher.pattern().pattern());
+    private int getIndexOf(String fileContent, String tag, int startIndex) {
+      int index = fileContent.indexOf(tag, startIndex);
+      if (index == -1) {
+        throw new IllegalArgumentException("The report does not contain a vald '<title>...</title>' tag.");
```
pavel-mikula-sonarsource, line 120: Typo
```suggestion
        throw new IllegalArgumentException("The report does not contain a valid '<title>...</title>' tag.");
```

pavel-mikula-sonarsource (approved):
LGTM, please fix the last typo before merging

PR #4213, opened 2021-04-08
Stop checking "generated" in region names for generated code detection

### LINKED ISSUE #2228: Region with word 'generated' disables analysis for entire file
### Description

A region with the word generated anywhere in the title causes sonarlint to skip analysis for the entire file.

```csharp
namespace SonarLint_Analysis_Failure
{
    public static class ShadowedMember
    {
        public static int MyProperty { get; set; }
        public class Initializer
        {
            public int MyProperty { get; set; } // S3218 should be thrown here
        }

        #region Requests generated from remote systems need to be tested here
        #endregion
    }
}
```

### Related information

* SonarC# Version: Latest
* Visual Studio Version: 15.9.4

andrei-epure-sonarsource:
Fixes #2228


its/src/test/java/com/sonar/it/csharp/AutoGeneratedTest.java
```diff
@@ -33,61 +33,62 @@
 
 public class AutoGeneratedTest {
 
-    @ClassRule
-    public static TemporaryFolder temp = TestUtils.createTempFolder();
+  @ClassRule
```
andrei-epure-sonarsource, line 36: sorry for the whitespace change noise, you can enable "hide whitespace changes" in GH for easy review

csaba-sagi-sonarsource:
Why did you opt in the end not to look for "Windows Form Designer generated code"?

csaba-sagi-sonarsource (approved):
LGTM. 
Nitpick:  
GeneratedCodeRecognizer - reorder the class members to comply with our coding style
CSharpGeneratedCodeRecognizer - lazy -> Lazy
VisualBasicGeneratedCodeRecognizer - lazy -> Lazy, Helpers. is surplus.

andrei-epure-sonarsource:
@csaba-sagi-sonarsource - RE: not keeping the "Windows Form Designer generated code" check
- there's a performance cost of verifying all regions, for each SyntaxTree / file
- the perf cost is not that big, however, in my understanding, we'd do this for some pre-2010 legacy way of generating Windows Forms code - it's a bit too much of an edge case

I think that if we will receive feedback on this, we can re-visit the decision. I believe we won't.

Another "safer" strategy would be to keep "Windows Form Designer generated code" for SQ LTS, and remove it afterwards. But I'm not sure it's worth the effort.

PR #4215, opened 2021-04-08
Automatically dectected test cases for rules

Corniel:
The test solution contains a lot of tests, which in most cases have the following form:

``` C#
[TestMethod]
[TestCategory("Rule")]
public void ArrayCovariance() =>
    Verifier.VerifyAnalyzer(@"TestCases\MyAnalyzer.cs", new MyAnalyzer());
```
That does not only leads to a lot of extra files - as each analyzer requires both a test case file, and a test file - but those test files, also contain a lot of duplicated code.

I think there is a better way, and this pull request is an attempt to convince you.

So the general approach is that by using custom attributes (I used NUnit to achieve this, it might be possible with MS Test too, but that would be less elegant, and and has less extension points) to decorate one method to basically load generated all rules tests at once:

``` C#
public class Verify
{
    [VerifyAssembly("TestCases", typeof(SonarAnalyzer.Rules.CSharp.ArrayCovariance))]
    [VerifyAssembly("TestCases", typeof(SonarAnalyzer.Rules.VisualBasic.ArrayCreationLongSyntax))]
    public void Assembly(
        string testCase,
        DiagnosticAnalyzer analyzer,
        ParseOptions[] options,
        IEnumerable<MetadataReference> additionalReferences) =>
        Verifier.VerifyAnalyzer(testCase, analyzer, options, additionalReferences);
}
```
The `VerifyAssembly` attribute checks for all `DiagnosticAnalyzer`s in the assembly, and checks for all test cases files on the given location (directory).

For each language version that is support by a rule (can be detected by a part of the file name like: `AssignmentInsideSubExpression.CSharp9.cs` or `MethodOverrideChangedDefaultValue.FromCSharp8.cs`
and the external packages can be defined by adding a line in the test files:
``` C#
// <PackageReference Include="Moq" Version="4.13.1" />
```

It automatically adds tags/catagory attributes, and as every language version has its own test method, so it is easy to run a test case for just one language version. A test name looks like:
> S101: [Visual Basic 12.0] ClassName

Something that does not work yet as I would like is that within visual studio, MS Tests groups some test cases (as it says a test has multiple result outcomes). Do you think this can be beneficial?


pavel-mikula-sonarsource:
Hi @Corniel, this is interesting idea that might remove quite a lot of code. Even individual steps (file autodetection, versioning, reference resolution) can improve the layout a lot.

One question I have is: How do I run single rule against single testcase file when fixing FP and single rule against all of it's testcase files to verify I haven't break anything?

Corniel:
> One question I have is: How do I run single rule against single testcase file when fixing FP and single rule against all of it's testcase files to verify I haven't break anything?

The idea is that in your test explorer every testcase, analyzer, language, version combination is a single test, so that should be easy.

Corniel:
@pavel-mikula-sonarsource I added an example for tests that need more test setup than that can be provided by `[VerifyAssembly]`.

pavel-mikula-sonarsource:
I like this idea and I'll schedule a team discussion to discuss this. We have some plans to reorganize our unit test projects and this can help with that. If approved, we'll probably split this PR into several smaller steps to deliver it incrementally.

I like the rule ID as a part of the test name, but the quick search `Search Test Explorer` in `Search Explorer` windows doesn't work for me when searching per rule ID or analyzer name. Do you have it the same? Can it be fixed?

pavel-mikula-sonarsource:
We'll also need some opt-in mechanism for the beginning that might turn to opt-out later. Because some tests are more complicated:
* we can have one analyzer class with two rules, both rules can share the test cases file (just with different messages)
* we can have one analyzer class with two rules and each rule ID has it's own test case file with different name
* some classes have complicated setup scenarios `XmlExternalEntityShouldNotBeParsedTest`
* some can have complicated code fixes 
 
and so on

Corniel:
> I like the rule ID as a part of the test name, but the quick search `Search Test Explorer` in `Search Explorer` windows doesn't work for me when searching per rule ID or analyzer name. Do you have it the same? Can it be fixed?

I never used the functionality so far, but I can confirm that it does not work. I have no clue (yet) why, so it is worth investigation. Fortunately, it is ordered by name, so by ID. (The MS test explorer works in mysterious ways )

Corniel:
> We'll also need some opt-in mechanism for the beginning that might turn to opt-out later. Because some tests are more complicated (..)

See the introduction:

``` C#
public class LineLength
{
    [Verify(typeof(CS.LineLength), scenario: "maximum 127")]
    public void Verify_CSharp(CS.LineLength analyzer, ParseOptions[] options)
    {
        analyzer.Maximum = 127;
        Verifier.VerifyAnalyzer(@"Cases.Complex\LineLength.cs", analyzer, options);
    }
}
```
So, you can, with a `[Verify]` attribute, specify everything you want. The attribute generates multiple tests that only differ in `ParseOptions`, and use the same naming format as `[VerifyAssembly]`. In the test explorer overview, both types of tests show up in the same list.

Corniel:
That is a valid outcome of a discussion. I'd liked doing the POC anyway.

PR #4222, opened 2021-04-09
Extract GetBindableParent to SyntaxTokenExtensions

costin-zaharia-sonarsource:
- to avoid the creation of unnecessary instances of `SymbolReferenceAnalyzer`


costin-zaharia-sonarsource:
> As it requires a different implementation for C# an VB.NET might it be worth putting it to the language facade?

The implementation is actually identical (`GetBindableParent`). I'm working to move it in a shared project to reduce duplication.

analyzers/SonarAnalyzer.Shared/SonarAnalyzer.Shared.projitems
```diff
@@ -0,0 +1,14 @@
...
+  <PropertyGroup>
+    <MSBuildAllProjects Condition="'$(MSBuildVersion)' == '' Or '$(MSBuildVersion)' &lt; '16.0'">$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+    <HasSharedItems>true</HasSharedItems>
+    <SharedGUID>892cf3fa-3be2-42e8-a7e6-76ae72864dec</SharedGUID>
+  </PropertyGroup>
+  <PropertyGroup Label="Configuration">
+    <Import_RootNamespace>SonarAnalyzer.Shared</Import_RootNamespace>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="$(MSBuildThisFileDirectory)Extensions\SyntaxTokenExtensions.cs" />
```
pavel-mikula-sonarsource, line 12: This list [tends to grow](https://github.com/SonarSource/sonar-security/blob/master/its/projects/internal/InternalCSharp/Shared.All/Shared.All.projitems) and therefore needs to be maintained.

Can we use some `*` trick here like [this](https://github.com/SonarSource/sonar-dotnet/blob/c0a120782f45321265c499f668c5f2b3ee885364/analyzers/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj#L50-L54)?

Something like
```suggestion
    <Compile Include="$(MSBuildThisFileDirectory)**\*" />
```

analyzers/SonarAnalyzer.Shared/Extensions/SyntaxTokenExtensions.cs
```diff
@@ -0,0 +1,86 @@
...
+                    return parent;
+                }
+
+                // If this node is not parented by a name, we're done.
+                if (!(parent is NameSyntax))
+                {
+                    return node;
+                }
+
+                node = parent;
```
pavel-mikula-sonarsource, line 80: To clean as we code, I'd prefer `if/else` here with positive condition.
```suggestion
                // If this node is not parented by a name, we're done.
                if (parent is NameSyntax)
                {
                    node = parent;
                }
                else
                {
                    return node;
                }                
```

analyzers/src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
```diff
@@ -56,11 +56,19 @@
     <NoWarn>NU1605</NoWarn>
     <RunAnalyzersDuringBuild>false</RunAnalyzersDuringBuild>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
+    <DefineConstants>TRACE;CS</DefineConstants>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+    <DefineConstants>TRACE;CS</DefineConstants>
+  </PropertyGroup>
```
pavel-mikula-sonarsource, line 64: This could be more straight forward approach. But it needs testing if it plays well with the default behavior.
```suggestion
  <PropertyGroup>
    <DefineConstants>$(DefineConstants);CS</DefineConstants>
  </PropertyGroup>
```

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/SyntaxTokenExtensionsTests.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    [TestClass]
+    public class SyntaxTokenExtensionsTests
+    {
+        [TestMethod]
+        public void GivenEmptyToken_GetBindableParent_ReturnsNull()
+        {
+            SyntaxToken empty = default;
+
+            Assert.IsNull(empty.GetBindableParent());
```
pavel-mikula-sonarsource, line 39: Is there a reason to not use `.Should().BeNull()` instead?

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/SyntaxTokenExtensionsTests.cs
```diff
@@ -0,0 +1,85 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using static csharp::SonarAnalyzer.Extensions.SyntaxTokenExtensions;
+
+namespace SonarAnalyzer.UnitTest.Extensions
+{
+    [TestClass]
+    public class SyntaxTokenExtensionsTests
+    {
+        [TestMethod]
+        public void GivenEmptyToken_GetBindableParent_ReturnsNull()
```
pavel-mikula-sonarsource, line 35: I'd put the method name `GetBindableParent_` as a prefix for all tests. If we add more methods to the extension class, it's easier to navigate tests prefixed by the tested method name.

pavel-mikula-sonarsource, line 35: I'd use `For` instead of `Given`
Both suggestion will result in `GetBindableParent_ForEmptyToken_ReturnsNull` that I find easier to understand.

pavel-mikula-sonarsource (changes requested):
Some minor improvements suggested

analyzers/SonarAnalyzer.Shared/SonarAnalyzer.Shared.projitems
costin-zaharia-sonarsource: This has the side effect of adding additional files like `SonarAnalyzer.Shared.projitems`. I've updated the pattern to `**\*.cs`

analyzers/src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
costin-zaharia-sonarsource: I prefer to let this as it is since it's generated by the project properties editor.

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/SyntaxTokenExtensionsTests.cs
costin-zaharia-sonarsource: 😄 not really. I've change it to be similar with the other assertions.

analyzers/SonarAnalyzer.Shared/Extensions/SyntaxTokenExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+            var node = token.Parent;
+
+            while (node != null)
+            {
+                var parent = node.Parent;
+
+                // If this node is on the left side of a member access expression, don't ascend
+                // further or we'll end up binding to something else.
+                if (parent is MemberAccessExpressionSyntax memberAccess &&
+                    memberAccess.Expression == node)
```
pavel-mikula-sonarsource, line 45: To fit our coding style (&& should be at the beginnig of the line). Conditions are short so let's make them single line.
```suggestion
                if (parent is MemberAccessExpressionSyntax memberAccess &&                    memberAccess.Expression == node)
```

pavel-mikula-sonarsource: The first 4 looks like a candidate for some switch expression, but I didn't tried that so I don't know how that would play with the rest.

pavel-mikula-sonarsource (approved):
LGTM. There's one last suggestion that I wrote before, but it didn't got submitted to the review for some reason (probably issue between keyboard and chair)

analyzers/SonarAnalyzer.Shared/Extensions/SyntaxTokenExtensions.cs
costin-zaharia-sonarsource: The problem is the branch at line 81 which does not return. It updates the parent to re-enter the loop.

I've reorganized the code a bit with 2 switch statements.

pavel-mikula-sonarsource: Nice :shipit:

PR #4224, opened 2021-04-09
Remove unused code

costin-zaharia-sonarsource:


pavel-mikula-sonarsource:
PR stats: `+4 −213` Oh yeah 😄

Corniel (approved):

analyzers/src/SonarAnalyzer.Common/Helpers/DotWriter.cs
```diff
@@ -59,16 +59,6 @@ public void WriteNode(string id, string header, params string[] items)
             this.writer.WriteLine("]");
         }
 
-        internal void WriteEdge(string startId, string endId, string label)
```
pavel-mikula-sonarsource, line 62: I suggest to keep this one, even when not currently used.
It was very useful during our Roslyn CFG investigation sprint (I didn't have to spent too much time figuring out how to do it). I used it a lot there.

pavel-mikula-sonarsource: This whole class is relevant only for debug output. So it offers help when we need it.

pavel-mikula-sonarsource: And it will actually most likely be needed when we reimplement the `CfgSerializer` for the new CFG.

pavel-mikula-sonarsource (approved):
LGTM, I just left one comment for preserving one unused debug method

PR #4226, opened 2021-04-12
Fix codesmells in UT proj

csaba-sagi-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
```diff
@@ -979,11 +985,11 @@ public void Cfg_For()
             var cfg = Build("for (var i = 0; true; i++) { var x = 10; }");
             VerifyForStatement(cfg);
             VerifyAllInstructions(cfg.EntryBlock, "0", "i = 0");
-            var condition = cfg.EntryBlock.SuccessorBlocks.First();
+            var condition = cfg.EntryBlock.SuccessorBlocks[0];
```
pavel-mikula-sonarsource, line 988: Why this change? It's not a code smell and `.First()` is easier to understand than the `[0]`

pavel-mikula-sonarsource (changes requested):
One question for now

analyzers/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs
csaba-sagi-sonarsource: It is a fix for this rolsyn based info level code smell: https://sonarcloud.io/project/issues?id=sonaranalyzer-dotnet&issues=AXi2n7WIzHMUitPsOt8i&open=AXi2n7WIzHMUitPsOt8i

PR #4227, opened 2021-04-12
Fix S4143 FP with Indexers

### LINKED ISSUE #4178: Fix S4143 FP: Indexers
### Description

S4143 False positive when using indexers.

### Repro steps

```cs
var first = new Curve();
var second = new Curve();

first.Points[0] = new Point();
second.Points[0] = new Point();
```
S4143 is reported, although the index in accessed on separate objects.

### Expected behavior

No warning reported.

### Actual behavior

S4143 is reported.

csaba-sagi-sonarsource:
Fixes #4178

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotOverwriteCollectionElements.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotOverwriteCollectionElements.cs
@@ -32,14 +32,9 @@ namespace SonarAnalyzer.Rules.CSharp
 {
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     [Rule(DiagnosticId)]
-    public sealed class DoNotOverwriteCollectionElements : DoNotOverwriteCollectionElementsBase<ExpressionStatementSyntax>
+    public sealed class DoNotOverwriteCollectionElements : DoNotOverwriteCollectionElementsBase<SyntaxKind, ExpressionStatementSyntax>
     {
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-
-        protected override DiagnosticDescriptor Rule { get; } = rule;
-
-        private static readonly HashSet<SyntaxKind> identifierOrLiteral = new HashSet<SyntaxKind>
+        private static readonly HashSet<SyntaxKind> IdentifierOrLiteral = new HashSet<SyntaxKind>
         {
             SyntaxKind.IdentifierName,
             SyntaxKind.StringLiteralExpression,
@@ -49,6 +44,8 @@ namespace SonarAnalyzer.Rules.CSharp
             SyntaxKind.TrueLiteralExpression,
             SyntaxKind.FalseLiteralExpression,
         };
+
+        protected override ILanguageFacade<SyntaxKind> Language => CSharpFacade.Instance;
 
         protected override void Initialize(SonarAnalysisContext context) =>
             context.RegisterSyntaxNodeActionInNonGenerated(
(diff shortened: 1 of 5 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotOverwriteCollectionElements.cs
```diff
@@ -300,4 +300,22 @@ public void Foo()
...
+            var first = new ClassWithListProperty();
+            var second = new ClassWithListProperty();
+
+            first.IntList[0] = 1;
+            second.IntList[0] = 2;
+        }
+
+        private class ClassWithListProperty
+        {
+            public List<int> IntList { get; set; }
```
andrei-epure-sonarsource, line 318: could you also add IntList2, and then the test above should be:

```
            first.IntList[0] = 1;
            first.IntList2[0] = 11;
            second.IntList[0] = 2;
            second.IntList2[0] = 22;
```
and no issue should be raised

Also, could you also add separate tests with various combinations of correctly raised issues like

```
            first.IntList[0] = 1;
            first.IntList2[0] = 2;
            first.IntList[0] = 3; // Noncompliant
            first.IntList2[1] = 4; // ok, different index
```

and 

```
            first.IntList[0] = 1;
            first.IntList2[1] = 2;
            first.IntList[1] = 3; 
            first.IntList2[1] = 4; // Noncompliant
```

and more compliant test cases like:
```
            first.IntList[0] = 1;
            first.IntList2[1] = 2;
            first.IntList[1] = 3; 
            first.IntList2[0] = 4;
```

andrei-epure-sonarsource: also, can you add some tests with nested objects, like

```
private class ClassWithNested
{
            public ClassWithListProperty Alpha { get; set; }
            public ClassWithListProperty Beta { get; set; }
}
```

and something like

```
            first.Alpha.IntList[0] = 1;
            first.Alpha.IntList2[0] = 11;
            first.Beta.IntList[0] = 1111;
            first.Beta.IntList2[0] = 11111;
            second.Alpha.IntList[0] = 2;
             second.Alpha.IntList2[0] = 22;
            second.Beta.IntList[0] = 222;
            second.Beta.IntList2[0] = 2222;
```

and a noncompliant case with such nested structure

andrei-epure-sonarsource (changes requested):

andrei-epure-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotOverwriteCollectionElements.cs
```diff
@@ -155,9 +152,9 @@ private static SyntaxNode GetIdentifier(ExpressionSyntax expression)
                     // ElementAccess    // <-- we are here
                     //   Arguments: index
                     //   Expression: SimpleMemberAccess
-                    //                 Expression: a
-                    //                 Name: b  // <-- we need this
-                    return ((MemberAccessExpressionSyntax)expression).Name;
+                    //                 Expression: a // <-- we need this
+                    //                 Name: b
+                    return ((MemberAccessExpressionSyntax)expression);
```
pavel-mikula-sonarsource, line 157: We don't need to change the type. And there's extra pair of parenthesis.
```
return expression;
```

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotOverwriteCollectionElements.cs
```diff
@@ -155,9 +152,9 @@ private static SyntaxNode GetIdentifier(ExpressionSyntax expression)
                     // ElementAccess    // <-- we are here
                     //   Arguments: index
                     //   Expression: SimpleMemberAccess
-                    //                 Expression: a
-                    //                 Name: b  // <-- we need this
-                    return ((MemberAccessExpressionSyntax)expression).Name;
+                    //                 Expression: a // <-- we need this
+                    //                 Name: b
```
pavel-mikula-sonarsource, line 156: This comment is not correct. We don't need `a`. We need `a.b` and that matches the implementation.

Issue #4010, opened 2021-02-01
Rule S2115: Databases should be password-protected (C# and web.config detection)


pierre-loup-tristant-sonarsource:
https://jira.sonarsource.com/browse/MMF-2255
https://jira.sonarsource.com/browse/RSPEC-2115
https://github.com/SonarSource/security-expected-issues/pull/265

Steps:
- [x] C# code
- [x] web.config
- [ ] appsettings.json
- [ ] support constant propagation

andrei-epure-sonarsource:
I closing this as being treated by #4249 and #4271.
I opened  #4298 as a follow-up for app.json

Issue #3995, opened 2021-01-26
Rule S5122: Having a permissive Cross-Origin Resource Sharing policy is security-sensitive


costin-zaharia-sonarsource:
https://jira.sonarsource.com/browse/MMF-2255
https://jira.sonarsource.com/browse/RSPEC-5122
https://github.com/SonarSource/security-expected-issues/pull/255

PR #4228, opened 2021-04-13
[C#] Rule: S4502 Disabling CSRF protections is security-sensitive

### LINKED ISSUE #4005: Rule: S4502 Disabling CSRF protections is security-sensitive
https://jira.sonarsource.com/browse/MMF-2255
https://jira.sonarsource.com/browse/RSPEC-4502
https://github.com/SonarSource/security-expected-issues/pull/266

costin-zaharia-sonarsource:
Fixes #4005


analyzers/src/SonarAnalyzer.CSharp/Extensions/AttributeSyntaxExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Extensions
+{
+    internal static class AttributeSyntaxExtensions
+    {
+        public static bool IsKnownType(this AttributeSyntax attribute, KnownType knownType, SemanticModel semanticModel) =>
+            attribute.Name.GetName().Contains(knownType.ShortNameWithoutAttributeSuffix)
```
costin-zaharia-sonarsource, line 30: I've added this overload on `AttributeSyntax` to be able to do a check first at the syntax level, for performance reasons.

analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -327,8 +327,8 @@ public static SimpleNameSyntax GetIdentifier(this ExpressionSyntax expression)
             expression switch
             {
                 MemberAccessExpressionSyntax memberAccess => memberAccess.Name.Identifier.ValueText,
-                IdentifierNameSyntax identifierName => identifierName.Identifier.ValueText,
-                GenericNameSyntax genericNameSyntax => genericNameSyntax.Identifier.ValueText,
+                SimpleNameSyntax simpleNameSyntax => simpleNameSyntax.Identifier.ValueText,
```
costin-zaharia-sonarsource, line 330: `SimpleNameSyntax` is the base class for both `IdentifierNameSyntax` and `GenericNameSyntax`.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DisablingCsrfProtection.cs
```diff
@@ -0,0 +1,86 @@
...
+namespace SonarAnalyzer.Rules.Hotspots
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public sealed class DisablingCsrfProtection : HotspotDiagnosticAnalyzer
+    {
+        private const string DiagnosticId = "S4502";
+        private const string MessageFormat = "Make sure disabling CSRF protection is safe here.";
+        private const string IgnoreAntiforgeryTokenAttributeName = "IgnoreAntiforgeryTokenAttribute";
+        private const SyntaxKind ImplicitObjectCreationExpression = (SyntaxKind)8659;
```
costin-zaharia-sonarsource, line 39: This is needed in order to be able to handle the target-typed new expressions from C# 9.

analyzers/tests/SonarAnalyzer.UnitTest/Common/SecurityHotspotTest.cs
```diff
@@ -67,7 +67,7 @@ private static string GetTestCaseFileName(DiagnosticAnalyzer analyzer)
 
             return typeName switch
             {
-                "ConfiguringLoggers" => "ConfiguringLoggers_AspNetCore",
+                "ConfiguringLoggers" => "ConfiguringLoggers_Log4Net",
```
costin-zaharia-sonarsource, line 70: We only need a test suite to check if hotspot rule is disabled by default. `ConfiguringLoggers_AspNetCore` makes sense only on .Net or .Net Core and I prefer to use `ConfiguringLoggers_Log4Net` instead since it works for all runtimes.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/CoreMetadataReference.cs
```diff
@@ -27,6 +27,12 @@ namespace SonarAnalyzer.UnitTest.MetadataReferences
 {
     internal static class CoreMetadataReference
     {
+        internal static MetadataReference MicrosoftAspNetCoreMvc { get; } = CreateReference(typeof(Microsoft.Extensions.DependencyInjection.MvcServiceCollectionExtensions).Assembly.Location);
+        internal static MetadataReference MicrosoftAspNetCoreMvcAbstractions { get; } = CreateReference(typeof(Microsoft.AspNetCore.Mvc.IActionResult).Assembly.Location);
+        internal static MetadataReference MicrosoftAspNetCoreMvcCore { get; } = CreateReference(typeof(Microsoft.AspNetCore.Mvc.ControllerBase).Assembly.Location);
+        internal static MetadataReference MicrosoftAspNetCoreMvcViewFeatures { get; } = CreateReference(typeof(Microsoft.AspNetCore.Mvc.Controller).Assembly.Location);
+        internal static MetadataReference MicrosoftExtensionsDependencyInjectionAbstractions { get; } =
+            CreateReference(typeof(Microsoft.Extensions.DependencyInjection.IServiceCollection).Assembly.Location);
```
costin-zaharia-sonarsource, line 35: I don't know a better way to provide the location without hardcoding the path.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/KnownTypeTest.cs
```diff
@@ -0,0 +1,18 @@
+﻿using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.UnitTest.Helpers
+{
+    [TestClass]
+    public class KnownTypeTest
```
andrei-epure-sonarsource, line 8: what if the KnownType isn't an Attribute? can you add a test for "X" - it will throw...

analyzers/src/SonarAnalyzer.Common/Helpers/KnownType.cs
```diff
@@ -547,6 +547,10 @@ private KnownType(SpecialType specialType, string typeName)
 
         public string ShortName => shortName.Value;
 
+        public string ShortNameWithoutAttributeSuffix => shortName.Value == nameof(Attribute)
```
andrei-epure-sonarsource, line 553: you are adding an attribute-specific property in a class which is meant for any kind of type... I have an itch about it

analyzers/src/SonarAnalyzer.CSharp/Extensions/AttributeSyntaxExtensions.cs
andrei-epure-sonarsource: it looks like "short name without attribute suffix" should be a private method in this class - it's only necessary here, right?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DisablingCSRFProtection.cs
```diff
@@ -0,0 +1,53 @@
...
+    [IgnoreAntiforgeryToken] // Noncompliant
+    public class S4502Controller : Controller
+    {
+        private readonly string path = string.Empty;
+
+        [HttpPost]
+        [IgnoreAntiforgeryToken] // Noncompliant
+        public IActionResult ChangeEmail_Noncompliant(string model) => View(path);
+
+        [HttpPost, IgnoreAntiforgeryTokenAttribute] // Noncompliant
```
andrei-epure-sonarsource: can you also add a test with `[HttpGet, IgnoreAntiforgeryTokenAttribute, AnotherAttribute]` ?

and one with only `[IgnoreAntiforgeryToken]` ?

is it related only to HttpPost?

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Common/Helpers/KnownType.cs
andrei-epure-sonarsource: see my comment on AttributeSyntaxExtensions.cs   - probably it should be there

andrei-epure-sonarsource:
RE: Razor - you said that it will be done after we finish the implementation for these ones, right?

analyzers/src/SonarAnalyzer.CSharp/Extensions/AttributeSyntaxExtensions.cs
costin-zaharia-sonarsource: I was in a doubt where to add this. It's used only here but it acts and needs only a known type to return the value.

I think I can move it here for now since it makes sense only for attributes.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/KnownTypeTest.cs
costin-zaharia-sonarsource: I've added a test for this in `AttributeSyntaxExtensionsTest`.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DisablingCSRFProtection.cs
costin-zaharia-sonarsource: > can you also add a test with [HttpGet, IgnoreAntiforgeryTokenAttribute, AnotherAttribute] ?
> and one with only [IgnoreAntiforgeryToken] ?

As far as I can tell we don't need these cases. We register for `AttributeSyntax` and we will get triggered for each occurrence.

> is it related only to HttpPost?

No, it's independent. I've added [HttpPost] just to cover the case when the attribute is in a list and not alone like on the line 40.

costin-zaharia-sonarsource:
> RE: Razor - you said that it will be done after we finish the implementation for these ones, right?

Yes, I think the razor implementation is optional and should be done at the end of the sprint, after we finish with #3995 and #4010. 

We could also consider to do this in a different sprint since it could be a big effort. If I remember well this was the reason why it wasn't part of the initial spec.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DisablingCSRFProtection.cs
andrei-epure-sonarsource: I was asking for these from a black-box testing point of view, from a functional point of view. It's good to have edge-case scenarios.

The edge cases I meant were:
- a scenario without "HttpPost" on the method. If I come after one year, I may be tempted to change the implementation to check if "HttpPost" is present on the method, and tests will be green. From the current documentation (i.e. tests), I see that this rule should only raise on "HttpPost" methods.
-  a scenario where "IgnoreAntiforgeryTokenAttribute" is not the last attribute in a list of attributes. Maybe that's exaggerated, but I try to cover edge cases. And I don't see major downsides.

The implementation can change. This is a simple rule, but in general, implementation can change and existing test cases should ensure a good enough safety net for refactoring.

andrei-epure-sonarsource (changes requested):
could you please add a test with a method which is doesn't have the "HttpPost" attribute, just to make it clear it's not related? (see [large comment](https://github.com/SonarSource/sonar-dotnet/pull/4228#discussion_r613314241))

andrei-epure-sonarsource (approved):
LGTM

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DisablingCSRFProtection.cs
costin-zaharia-sonarsource: I usually prefer a red-green-refactor approach in order to be able to add only the tests and code we need.

As discussed offline I've removed the `[HttpPost]` attribute from line 39 to have a case with just an attribute.

PR #4229, opened 2021-04-13
Add IT with nunit and xunit analyzers as external issues

### LINKED ISSUE #4214: Add IT with nunit and xunit analyzers as external issues
External issues should be imported for test code and we should have integration tests for it.

pavel-mikula-sonarsource:
Fixes #4214

### DIFF
--- a/its/projects/ExternalIssues.TestProject.CS/.editorconfig
+++ b/its/projects/ExternalIssues.TestProject.CS/.editorconfig
@@ -1,0 +1,6 @@
+﻿root = true
+
+[*.cs]
+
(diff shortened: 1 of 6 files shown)


csaba-sagi-sonarsource (approved):
LGTM. IT would be nice to have ITs verifying the same functionality for VB, but as the importing part is in the shared library I do not have a strong opinion.

pavel-mikula-sonarsource:
Unfortunately, there's o xUnit nor NUnit VB analyzer available. I would add it otherwise.

PR #4234, opened 2021-04-13
Fixes S2699 FP with using static directive

### LINKED ISSUE #4233: S2699 is raised if assertion is used with using static directive
### Description

S2699 raises issue if the assertion is used with using static directive.

### Repro steps
```
using static Microsoft.VisualStudio.TestTools.UnitTesting.Assert;

namespace ReproNameSpace
{
    [TestClass]
    public class Repro
    {
        [TestMethod]
        public void TestWithUsingStatic(int b)
        {
            int a = 5;
            IsTrue(a == b);
        }
    }
}
```

### Expected behavior

Issue is not raised.

### Actual behavior

S2699 issue is raised. 

### Known workarounds

-

### Related information

* C#/VB.NET Plugins version 8.20

csaba-sagi-sonarsource:
Fixes #4233

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotOverwriteCollectionElements.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotOverwriteCollectionElements.cs
@@ -152,9 +152,10 @@ namespace SonarAnalyzer.Rules.CSharp
                     // ElementAccess    // <-- we are here
                     //   Arguments: index
                     //   Expression: SimpleMemberAccess
-                    //                 Expression: a // <-- we need this
+                    //                 Expression: a
                     //                 Name: b
-                    return ((MemberAccessExpressionSyntax)expression);
+                    // We need a.b
+                    return expression;
                 case SyntaxKind.IdentifierName:
                     // a[index]
                     // ElementAccess    // <-- we are here
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
@@ -38,11 +38,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private const string MessageFormat = "Add at least one assertion to this test case.";
         private const string CustomAssertionAttributeName = "AssertionMethodAttribute";
 
-        private static readonly DiagnosticDescriptor rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
-
-        private static readonly Dictionary<string, KnownType> knownAssertions = new Dictionary<string, KnownType>
+        private static readonly Dictionary<string, KnownType> KnownAssertions = new Dictionary<string, KnownType>
         {
             {"DidNotReceive", KnownType.NSubstitute_SubstituteExtensions},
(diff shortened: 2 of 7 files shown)


analyzers/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
```diff
@@ -151,5 +102,28 @@ public static bool IsMsTestOrNUnitTestIgnored(this IMethodSymbol method)
         public static KnownType FindFirstTestMethodType(this IMethodSymbol method) =>
             KnownTestMethodAttributes.FirstOrDefault(known =>
                     method.GetAttributes().Any(att => att.AttributeClass.Is(known)));
+
+        private static readonly ImmutableArray<KnownType> KnownTestMethodAttributes = ImmutableArray.Create(
```
pavel-mikula-sonarsource, line 106: Fields should be before methods

analyzers/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs
```diff
@@ -24,81 +24,47 @@
 
 // Note: useful comparison of the differing syntax across unit test frameworks
 // at https://xunit.github.io/docs/comparisons
-
```
pavel-mikula-sonarsource, line 27: The link above doesn't work. If you can find replacement in 2 minutes, we can update it (to a single line version). Otherwise we don't need the comment.

analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
```diff
@@ -117,8 +122,11 @@ private static bool IsTestIgnored(IMethodSymbol method)
 
         private static bool IsKnownAssertion(ISymbol methodSymbol)
         {
-            var type = knownAssertions.GetValueOrDefault(methodSymbol.Name);
-            return (type != null) && methodSymbol.ContainingType.ConstructedFrom.Is(type);
+            var type = KnownAssertions.GetValueOrDefault(methodSymbol.Name);
+            return type != null
+                ? methodSymbol.ContainingType.ConstructedFrom.Is(type)
+                : methodSymbol.ContainingType is { } containingType
+                  && containingType.DerivesFromAny(KnownAssertionTypes);
```
pavel-mikula-sonarsource, line 129: `type != null` can be `is {}`
`ContainingType` cannot be null and we can access it directly.
With that, we can simplify this to an arrow method.
```suggestion
            (KnownAssertions.GetValueOrDefault(methodSymbol.Name) is { } type && methodSymbol.ContainingType.ConstructedFrom.Is(type))
            || methodSymbol.ContainingType.DerivesFromAny(KnownAssertionTypes);
```

pavel-mikula-sonarsource (changes requested):
Minor improvements

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotOverwriteCollectionElements.cs
```diff
@@ -152,9 +152,10 @@ private static SyntaxNode GetIdentifier(ExpressionSyntax expression)
                     // ElementAccess    // <-- we are here
                     //   Arguments: index
                     //   Expression: SimpleMemberAccess
-                    //                 Expression: a // <-- we need this
+                    //                 Expression: a
                     //                 Name: b
-                    return ((MemberAccessExpressionSyntax)expression);
+                    // We need a.b
+                    return expression;
```
pavel-mikula-sonarsource, line 158: This is just minor improvement to cleanup another PR.

pavel-mikula-sonarsource (approved):
LGTM

PR #4237, opened 2021-04-14
Remove /p:RunAnalyzers=true

pavel-mikula-sonarsource:
This should not be needed anymore. https://github.com/SonarSource/sonar-scanner-msbuild/issues/829 was fixed and we have S4NET 5.2 on AZP already. 

Functionality is verified in #4238 that's based on this branch - new artificial issue failed the QG


andrei-epure-sonarsource (approved):
lgtm

PR #4243, opened 2021-04-14
Fix S2699 FP: Inspect invoked methods to search for assertions

### LINKED ISSUE #4241: Fix S2699 FP: Inspect invoked methods to search for assertions
Assertion can be placed in a helper method invoked by several tests. We should try to do some shallow inspection of invoked methods.

pavel-mikula-sonarsource:
Fixes #4241

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
@@ -85,21 +85,7 @@ namespace SonarAnalyzer.Rules.CSharp
                         return;
                     }
 
-                    if (ContainsAssertion(methodDeclaration, c.SemanticModel, new HashSet<IMethodSymbol>(), 0))
-                    {
-                        return;
-                    }
-
-                    var hasAnyAssertion = methodDeclaration.DescendantNodes()
-                        .OfType<InvocationExpressionSyntax>()
-                        .Any(invocation => IsAssertion(invocation));
-
-                    var hasAnyThrownAssertionExceptions = methodDeclaration.DescendantNodes()
-                        .OfType<ThrowStatementSyntax>()
-                        .Any(throwStatement => throwStatement.Expression is { } expression
-                                               && c.SemanticModel.GetTypeInfo(expression).Type.DerivesFromAny(KnownAsertionExceptionTypes));
-
-                    if (!hasAnyAssertion && !hasAnyThrownAssertionExceptions)
+                    if (!ContainsAssertion(methodDeclaration, c.SemanticModel, new HashSet<IMethodSymbol>(), 0))
                     {
                         c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, methodDeclaration.Identifier.GetLocation()));
                     }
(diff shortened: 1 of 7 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
```diff
@@ -98,6 +95,36 @@ protected override void Initialize(SonarAnalysisContext context)
                 SyntaxKind.MethodDeclaration);
         }
 
+        private static bool ContainsAssertion(MethodDeclarationSyntax methodDeclaration, SemanticModel previousSemanticModel, ISet<IMethodSymbol> visitedSymbols, int level)
```
csaba-sagi-sonarsource, line 98: hasAnyAssertion part should be also part of ContainsAssertion. Because this way we do not recognize things like x.Should().Be(y) if it is in a helper method.

csaba-sagi-sonarsource: Also, now that the assert exception part is implemented, it would be nice to rebase on top of master and put hte hasAnyThrownAssertionExceptions  part inside the ContainsAssertion check too.

analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
```diff
@@ -98,6 +95,36 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            if (level == MaxInvocationDepth)
+            {
+                return false;
+            }
+            foreach (var symbol in invokedSymbols.Where(x => !visitedSymbols.Contains(x)))
+            {
+                visitedSymbols.Add(symbol);
+                foreach (var invokedDeclaration in symbol.DeclaringSyntaxReferences.Select(x => x.GetSyntax()).OfType<MethodDeclarationSyntax>())
+                {
+                    if (ContainsAssertion(invokedDeclaration, currentSemanticModel, visitedSymbols, level + 1))
```
csaba-sagi-sonarsource, line 119: If I remember correctly we talked about BFS implementation and this one looks like DFS. Or am I missing something?

csaba-sagi-sonarsource (changes requested):
After addressing the comments, it would be possible to remove the AssertionMethodAttribute Attribute from our unit tests, as we will be able to detect the assertions without the custom attribute.

pavel-mikula-sonarsource:
> After addressing the comments, it would be possible to remove the AssertionMethodAttribute Attribute from our unit tests, as we will be able to detect the assertions without the custom attribute.

And after release to SC :)

analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
pavel-mikula-sonarsource: Nice catch! I can simplify the logic and remove redundant `DescendantNodes()` evaluation (that can be expensive).

analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
pavel-mikula-sonarsource: I've skipped the BFS idea to keep it simple, because I've chosen the limit to be 2 instead of the original idea of 3

csaba-sagi-sonarsource (changes requested):
Can we add new UT cases with  assert exception throwing  and  fluent assertion with Should?

csaba-sagi-sonarsource (approved):
LGTM

PR #4247, opened 2021-04-14
Don't run rules with Main&Test scope on Test projects

### LINKED ISSUE #4245: Don't run rules with Main&Test scope on Test projects
Rules that have a scope `Main Sources` and `Test Sources` should not run on projects classified as `Test Source`.

pavel-mikula-sonarsource:
Fixes #4245


pavel-mikula-sonarsource:
Let's merge it with the coverage issue, I've bumped into an old problem there

csaba-sagi-sonarsource (approved):
LGTM. In SonarAnalysisContext.Test.cs it would be nice to use everywhere the hasMainScope (in some tests it is called hasProductScope). And to add to all if else branches what is happening with the mixed rules. All the comments are correct, yet I needed some time to understand which rules are run where. I think the comments would help with that.

PR #4253, opened 2021-04-16
[C#] Rule S5122: Having a permissive Cross-Origin Resource Sharing policy is security-sensitive

### LINKED ISSUE #3995: Rule S5122: Having a permissive Cross-Origin Resource Sharing policy is security-sensitive

costin-zaharia-sonarsource:
Fixes #3995


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/PermissiveCors.Net.cs
```diff
@@ -0,0 +1,85 @@
...
+    [Route("[controller]")]
+    public class CorsEnabledManualAddedHeadersController : Controller
+    {
+        private const string Star = "*";
+
+        [HttpGet]
+        public void Index(string header, string headerValue)
+        {
+            // The Access-Control-Allow-Origin response header indicates whether the response can be shared with requesting code from the given origin.
+            Response.Headers.Add("Access-Control-Allow-Origin", "*"); // Noncompliant
```
andrei-epure-sonarsource, line 20: how about a combination of header names? e.g. from [SO answer](https://stackoverflow.com/a/55016585/9908638)
```
context.Response.Headers.Add(
  CorsConstants.AccessControlAllowHeaders,
  $"{HeaderNames.Authorization}, {HeaderNames.ContentType}, {HeaderNames.AcceptLanguage}, {HeaderNames.Accept}");
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/PermissiveCors.Net.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    public class Setup
+    {
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddCors(options =>
+            {
+                options.AddDefaultPolicy(builder =>
+                {
+                    builder.WithOrigins("*"); // Noncompliant
```
andrei-epure-sonarsource, line 55: can you add a test with a constant
```
private const string ALLOW_ALL = "*";
//...
builder.WithOrigins(ALLOW_ALL);
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/PermissiveCors.Net.cs
costin-zaharia-sonarsource: I think the linked example is not correct.

As far as I can tell `context.Response.Headers` is `IDictionary<string, StringValues>` which expects the format "HEADER", "VALUES" (e.g. `context.Response.Headers.Add("Access-Control-Allow-Origin", "*");`).

We could add a test case with concatenated values instead but I think this is not used a lot since the concatenation happens by default due to `StringValues`.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/PermissiveCors.Net.cs
costin-zaharia-sonarsource: I've added already a few tests with constant at lines 21, 35 and 41 but I think I can add one more here as well.

analyzers/src/SonarAnalyzer.CSharp/RspecStrings.resx
```diff
@@ -12553,6 +12553,33 @@
   <data name="S5042_Type" xml:space="preserve">
     <value>SECURITY_HOTSPOT</value>
   </data>
+  <data name="S5122_Category" xml:space="preserve">
+    <value>Minor Security Hotspot</value>
+  </data>
+  <data name="S5122_Description" xml:space="preserve">
+    <value>Having a permissive Cross-Origin Resource Sharing policy is security-sensitive. It has led in the past to the following vulnerabilities:</value>
```
andrei-epure-sonarsource, line 12560: this description is a bit abrupt. Maybe there's something to improve in our tooling, to ensure we don't include sentences ending in `:`?

analyzers/tests/SonarAnalyzer.UnitTest/Common/SecurityHotspotTest.cs
```diff
@@ -87,14 +87,16 @@ analyzerName switch
...
+                "PermissiveCors" => "PermissiveCors.Net",
 #endif
                 _ => analyzerName
             };
 
         private static bool IsTestValid(string analyzerName)
         {
 #if NETFRAMEWORK
-            return analyzerName != nameof(DisablingCsrfProtection);
+            return analyzerName != nameof(DisablingCsrfProtection)
```
andrei-epure-sonarsource, line 98: given that `ASP.NET MVC` is part of the spec, shouldn't there be a `ASP.NET MVC` test case for .NET Framework, and  `ASP.NET Core MVC` for Netcore? basically splitting  the existing test case in two?

or isn't this already the case for `PermissiveCors.WebApi.cs`? shouldn't that be in this test as well? the same as we do for `LooseFilePermissions`?

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/PermissiveCorsTest.cs
```diff
@@ -0,0 +1,78 @@
...
+                CoreMetadataReference.MicrosoftAspNetCoreMvcViewFeatures,
+                CoreMetadataReference.MicrosoftExtensionsDependencyInjectionAbstractions,
+                CoreMetadataReference.MicrosoftExtensionsPrimitives,
+                CoreMetadataReference.MicrosoftNetHttpHeadersHeaderNames,
+            };
+#else
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void PermissiveCors_AspNet_WebApi() =>
+            Verifier.VerifyAnalyzer(@"TestCases\Hotspots\PermissiveCors.WebApi.cs",
```
andrei-epure-sonarsource, line 65: wouldn't it be better to name it `PermissiveCors.NetFramework.cs`? (there is AspNetCore WebApi as well...)

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PermissiveCors.cs
```diff
@@ -0,0 +1,145 @@
...
+                if (IsEnabled(compilationContext.Options))
+                {
+                    context.RegisterSyntaxNodeActionInNonGenerated(VisitAttribute, SyntaxKind.Attribute);
+                }
+            });
+        }
+
+        protected override void Initialize(TrackerInput input)
+        {
+            var iTracker = Language.Tracker.Invocation;
```
andrei-epure-sonarsource, line 65: `iTracker` sounds like an interface... maybe we can find a better name?

`invocationTracker` may be longer, but it doesn't sound like an interface :)

or if you split the Invocation tracking from Object Creation tracking in two methods, you'll be able to use just tracker, or just `it`

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PermissiveCors.cs
```diff
@@ -0,0 +1,145 @@
...
+        private static bool ContainsStar(IEnumerable<ExpressionSyntax> expressions, SemanticModel semanticModel) =>
+            expressions.Any(expression => ContainsStar(semanticModel.GetConstantValue(expression)));
+
+        private static bool ContainsStar(Optional<object> constantValue) =>
+            constantValue is {HasValue: true, Value: StarConstant};
+
+        private static bool ContainsStar(ObjectCreationExpressionSyntax objectCreation, SemanticModel semanticModel) =>
+            (objectCreation.IsKnownType(KnownType.Microsoft_Extensions_Primitives_StringValues, semanticModel)
+             || objectCreation.IsKnownType(KnownType.Microsoft_AspNetCore_Cors_Infrastructure_CorsPolicyBuilder, semanticModel))
+            && objectCreation.ArgumentList is { }
```
andrei-epure-sonarsource, line 142: this can only be null for VB, right?

andrei-epure-sonarsource (commented):
(intermediate comments )

analyzers/its/sources/ManuallyAddedNoncompliantIssues/NetFramework48/packages.config
```diff
@@ -0,0 +1,14 @@
...
+  <package id="Microsoft.AspNet.WebApi.Core" version="5.2.7" targetFramework="net48" />
+  <package id="Microsoft.AspNet.WebApi.Cors" version="5.2.7" targetFramework="net48" />
+  <package id="Microsoft.Extensions.Primitives" version="2.2.0" targetFramework="net48" />
+  <package id="Microsoft.Net.Http.Headers" version="2.2.8" targetFramework="net48" />
+  <package id="Newtonsoft.Json" version="6.0.4" targetFramework="net48" />
+  <package id="System.Buffers" version="4.5.0" targetFramework="net48" />
+  <package id="System.Memory" version="4.5.1" targetFramework="net48" />
+  <package id="System.Numerics.Vectors" version="4.4.0" targetFramework="net48" />
+  <package id="System.Runtime.CompilerServices.Unsafe" version="4.5.1" targetFramework="net48" />
+</packages>
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: this is automatically generated and managed by the IDE. I prefer to let it as it is.

analyzers/src/SonarAnalyzer.CSharp/RspecStrings.resx
costin-zaharia-sonarsource: I think this feedback is for the appsec team. @pierre-loup-tristant-sonarsource

analyzers/tests/SonarAnalyzer.UnitTest/Common/SecurityHotspotTest.cs
costin-zaharia-sonarsource: Most of the test cases are common between `ASP.NET MVC` and `Web Api` since the API is defined in a NuGet package. I will rename the file to `PermissiveCors.NetFramework.cs` to reflect that. I will also add more test cases

analyzers/src/SonarAnalyzer.CSharp/RspecStrings.resx
pierre-loup-tristant-sonarsource: Can you elaborate @andrei-epure-sonarsource? Is it not correct to end the sentence with `:`?

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PermissiveCors.cs
costin-zaharia-sonarsource: I've checked the code and it is defined as nullable reference also in the `CSharp` namespace:
image
I prefer to let this as it is.

analyzers/src/SonarAnalyzer.CSharp/RspecStrings.resx
andrei-epure-sonarsource: Hmm nevermind. This won't show up in SonarLint. This is the description that shows up in SonarLint AFAIK, but security hotspots don't show up there.

The problem I raised initially is that if a users sees in the IDE `It has led in the past to the following vulnerabilities:` without the actual list, it's weird.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PermissiveCors.cs
andrei-epure-sonarsource: do you have a test case for this?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/PermissiveCors.Net.cs
```diff
@@ -0,0 +1,91 @@
...
+            Response.Headers.Append(HeaderNames.AccessControlAllowOrigin, "https://trustedwebsite.com");
+            Response.Headers.Append("OtherString", "*");
+            Response.Headers.Append(HeaderNames.Age, "*");
+
+            Response.Headers.Append(HeaderNames.AccessControlAllowOrigin, new StringValues("*")); // Noncompliant
+            Response.Headers.Append(HeaderNames.AccessControlAllowOrigin, new StringValues(Star)); // Noncompliant
+            Response.Headers.Append(HeaderNames.AccessControlAllowOrigin, new StringValues("https://trustedwebsite.com"));
+            Response.Headers.Append(HeaderNames.AccessControlAllowOrigin, new StringValues(new [] {"*", "https://trustedwebsite.com"})); // Noncompliant
+
+            Response.Headers.Append(HeaderNames.AccessControlAllowOrigin, "*, https://trustedwebsite.com"); // Compliant - FN
```
andrei-epure-sonarsource, line 45: please leave just `"// FN"` - if it's compliant, it's not an FN
same for below line

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PermissiveCors.cs
costin-zaharia-sonarsource: I've added one now.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/PermissiveCors.Net.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                options.AddDefaultPolicy(builder =>
+                {
+                    builder.WithOrigins("*"); // Noncompliant
+                    builder.WithOrigins(Star); // Noncompliant
+                });
+
+                options.AddPolicy("EnableAllPolicy", builder =>
+                {
+                    builder.WithOrigins("https://trustedwebsite.com", "*"); // Noncompliant
```
andrei-epure-sonarsource, line 66: can you also add a case with `builder.WithOrigins("*", "*", "*");` ?

(I know it doesn't make sense, but to document it's raised only once)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/PermissiveCors.Net.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        private const string Star = "*";
+
+        [HttpGet]
+        public void Index(string header, string headerValue)
+        {
+            // The Access-Control-Allow-Origin response header indicates whether the response can be shared with requesting code from the given origin.
+            Response.Headers.Add("Access-Control-Allow-Origin", "*"); // Noncompliant
+            Response.Headers.Add("Access-Control-Allow-Origin", Star); // Noncompliant
+            Response.Headers.Add("Access-Control-Allow-Origin", "https://trustedwebsite.com");
```
andrei-epure-sonarsource, line 22: could you also add a case `Response.Headers.Add("Access-Control-Allow-Origin", "*****");` ?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/PermissiveCors.Net.cs
andrei-epure-sonarsource: and another case, which should be compliant ([source](https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-5.0#set-the-allowed-origins)):
```
options.AddPolicy("MyAllowSubdomainPolicy",
    builder =>
    {
        builder.WithOrigins("https://*.example.com")
            .SetIsOriginAllowedToAllowWildcardSubdomains();
    });
```

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PermissiveCors.cs
```diff
@@ -0,0 +1,148 @@
...
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public sealed class PermissiveCors : TrackerHotspotDiagnosticAnalyzer<SyntaxKind>
+    {
+        private const string DiagnosticId = "S5122";
+        private const string MessageFormat = "Make sure this permissive CORS policy is safe here.";
+        private const string AccessControlAllowOriginHeader = "Access-Control-Allow-Origin";
+        private const string AccessControlAllowOriginPropertyName = "AccessControlAllowOrigin";
+        private const string StarConstant = "*";
+        private const int ParameterCount = 2;
```
andrei-epure-sonarsource, line 43: this is just for `Add`/`Append`, not for `WithOrigins` or `AllowAnyOrigin` - so it's not "the parameter count for all methods verified in this rule"... 

and this could be a `const int` just inside `SetupInvocationTracker`

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PermissiveCors.cs
```diff
@@ -0,0 +1,148 @@
...
+                               && IsSecondArgumentStarString((InvocationExpressionSyntax)c.Node, c.SemanticModel),
+                          tracker.IsIHeadersDictionary());
+
+            tracker.Track(input,
+                          tracker.MatchMethod(new MemberDescriptor(KnownType.Microsoft_AspNetCore_Cors_Infrastructure_CorsPolicyBuilder, "WithOrigins")),
+                          c => ContainsStar(((InvocationExpressionSyntax)c.Node).ArgumentList.Arguments.Select(a => a.Expression), c.SemanticModel));
+
+            tracker.Track(input, tracker.MatchMethod(new MemberDescriptor(KnownType.Microsoft_AspNetCore_Cors_Infrastructure_CorsPolicyBuilder, "AllowAnyOrigin")));
+
+            tracker.Track(input,
```
andrei-epure-sonarsource, line 85: semantically, it makes sense to move this after line 78, to keep Add/Append methods close

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PermissiveCors.cs
```diff
@@ -0,0 +1,148 @@
...
+                                              new MemberDescriptor(KnownType.System_Web_HttpResponse, "AppendHeader"),
+                                              new MemberDescriptor(KnownType.System_Web_HttpResponseBase, "AddHeader"),
+                                              new MemberDescriptor(KnownType.System_Collections_Specialized_NameValueCollection, "Add"),
+                                              new MemberDescriptor(KnownType.System_Net_Http_Headers_HttpHeaders, "Add")),
+                          tracker.MethodHasParameters(ParameterCount),
+                          c => IsFirstArgumentAccessControlAllowOrigin((InvocationExpressionSyntax)c.Node, c.SemanticModel)
+                               && IsSecondArgumentStarString((InvocationExpressionSyntax)c.Node, c.SemanticModel));
+        }
+
+        private static void SetupObjectCreationTracker(ObjectCreationTracker<SyntaxKind> tracker, TrackerInput input) =>
```
andrei-epure-sonarsource, line 96: this is going to be called for all object creations - can you do an early-return based on a simple string check inside the argument list, if it contains "*"?

because it's going to get the symbol for all object creations, and that's a lot of symbols to check

or maybe do the arguments check before?... that would work if inside `ContainsStar` the argumnets check would be done before the method symbol check

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PermissiveCors.cs
```diff
@@ -0,0 +1,148 @@
...
+            };
+
+        private static bool ContainsStar(IEnumerable<ExpressionSyntax> expressions, SemanticModel semanticModel) =>
+            expressions.Any(expression => ContainsStar(semanticModel.GetConstantValue(expression)));
+
+        private static bool ContainsStar(Optional<object> constantValue) =>
+            constantValue is {HasValue: true, Value: StarConstant};
+
+        private static bool ContainsStar(ObjectCreationExpressionSyntax objectCreation, SemanticModel semanticModel) =>
+            (objectCreation.IsKnownType(KnownType.Microsoft_Extensions_Primitives_StringValues, semanticModel)
```
andrei-epure-sonarsource, line 143: nitpick: wouldn't the argument check be better to be done before? 

do we know if `GetConstantValue` is cheaper than `GetSymbolInfo`? (I would expect it to be cheaper)

andrei-epure-sonarsource (changes requested):
I left more comments. My main worry is performance around object creation verification

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PermissiveCors.cs
costin-zaharia-sonarsource: Yes, [WithOrigins](https://docs.microsoft.com/en-us/dotnet/api/Microsoft.AspNetCore.Cors.Infrastructure.CorsPolicyBuilder.WithOrigins?view=aspnetcore-5.0&viewFallbackFrom=netcore-5.0) needs `params string[] origins` [AllowAnyOrigin](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.cors.infrastructure.corspolicybuilder.allowanyorigin?view=aspnetcore-5.0) does not has any parameters

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PermissiveCors.cs
costin-zaharia-sonarsource: We now check first for the attribute list to be != null (syntax level), then check the name of the type (syntax level) and only later get the type. I expect that filtering by name will reduce a lot the number of types to get.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/PermissiveCors.cs
costin-zaharia-sonarsource: I've added `ObjectCreationExpressionSyntaxExtensions.IsKnownType` which checks first the name of the type before retrieving the symbol.

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/ObjectCreationExpressionSyntaxExtensionsTest.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace SonarAnalyzer.UnitTest.Extensions
+{
+    [TestClass]
+    public class ObjectCreationExpressionSyntaxExtensionsTest
+    {
+        [TestMethod]
+        [DataRow("new System.DateTime()", true)]
+        [DataRow("using System; class T { DateTime field = new DateTime(); }", true)]
+        [DataRow("new double()", false)]
```
andrei-epure-sonarsource, line 39: Also with 
```
using DT = System.DateTime; class T { DT field = new DT(); }
```
and
```
"using System; class T { DateTime field = new(); }"
```

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Extensions/ObjectCreationExpressionSyntaxExtensions.cs
```diff
@@ -33,5 +33,9 @@ internal static class ObjectCreationExpressionSyntaxExtensions
 
         public static IEnumerable<ArgumentSyntax> GetArgumentsOfKnownType(this ObjectCreationExpressionSyntax objectCreation, KnownType knownType, SemanticModel semanticModel) =>
             objectCreation?.ArgumentList?.Arguments.GetArgumentsOfKnownType(knownType, semanticModel) ?? Enumerable.Empty<ArgumentSyntax>();
+
+        public static bool IsKnownType(this ObjectCreationExpressionSyntax objectCreation, KnownType knownType, SemanticModel semanticModel) =>
+            objectCreation.Type.GetName().EndsWith(knownType.ShortName)
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: I will not work. Target-type new expressions are of type `ImplicitObjectCreationExpression`. We need to handle this separately. We could have a PR to fix this for all the rules which are using the ObjectCreationTracker and see then if we need a tracker or not.

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/ObjectCreationExpressionSyntaxExtensionsTest.cs
costin-zaharia-sonarsource: > "using System; class T { DateTime field = new(); }"

`new()` in this case is `ImplicitObjectCreationExpression` so this case is not relevant.

analyzers/src/SonarAnalyzer.CSharp/RspecStrings.resx
pierre-loup-tristant-sonarsource: It looks like all the Security-Hotspots description are truncated in this file and only the first line appear.

> This is the description that shows up in SonarLint AFAIK, but security hotspots don't show up there
    
I think they do now: https://community.sonarsource.com/t/sonarlint-for-vscode-1-20-1-released-review-security-hotspots-in-the-ide/38162

andrei-epure-sonarsource: To be checked with SonarLint if they query the HTML from SonarQube/SC - @duncanp-sonar ?

andrei-epure-sonarsource (approved):
LGTM

Issue #4255, opened 2021-04-16
S4583: FP with BeginInvoke / EndInvoke - Endinvoke as delegate


csaba-sagi-sonarsource:
### Description

If Action.Endinvoke is used directly as an argument of BeginInvoke, an issue is raised. (S4583)

### Repro steps

```
Action a = () => { };
a.BeginInvoke(a.EndInvoke, null);
```

### Expected behavior

An issue is not raised.

### Actual behavior

An issue is raised.

### Known workarounds

```
Action a = () => { };
a.BeginInvoke(ar => a.EndInvoke(ar), null);
```

### Related information

* C#/VB.NET Plugins version : 8.20

PR #4256, opened 2021-04-16
Add Repro for S4583 with action.EndInvoke as callback

csaba-sagi-sonarsource:
Reproduces #4255


costin-zaharia-sonarsource (approved):
LGTM!

PR #4263, opened 2021-04-19
SecurityHotspotTest: link only required references

costin-zaharia-sonarsource:
When running the hotspots, link only the required references instead of doing a union of all the references needed by all hotspots

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/CookieShouldBeHttpOnlyTest.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/CookieShouldBeHttpOnlyTest.cs
@@ -18,11 +18,10 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-#if NET
 using System.Collections.Generic;
-using Microsoft.CodeAnalysis;
-#endif
 using System.IO;
+using System.Linq;
+using Microsoft.CodeAnalysis;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.UnitTest.MetadataReferences;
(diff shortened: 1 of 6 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/CookieShouldBeHttpOnlyTest.cs
```diff
@@ -99,5 +98,8 @@ public void CookiesShouldBeHttpOnly_WithWebConfigValueSetToFalse()
         private static IEnumerable<MetadataReference> GetAdditionalReferences_NetCore() =>
             NuGetMetadataReference.MicrosoftAspNetCoreHttpFeatures(Constants.NuGetLatestVersion);
 #endif
+
+        internal static IEnumerable<MetadataReference> AdditionalReferences =>
```
andrei-epure-sonarsource, line 102: nitpick internal should be above private (you can make this change in your next PR though)

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/ExecutingSqlQueriesTest.cs
```diff
@@ -40,26 +40,26 @@ public class ExecutingSqlQueriesTest
...
 
         internal static IEnumerable<MetadataReference> GetReferencesNet46(string sqlServerCeVersion) =>
-            Enumerable.Empty<MetadataReference>()
-                .Concat(FrameworkMetadataReference.SystemData)
-                .Concat(FrameworkMetadataReference.SystemDataOracleClient)
-                .Concat(NuGetMetadataReference.SystemDataSqlServerCe(sqlServerCeVersion)
-                .Concat(NuGetMetadataReference.MySqlData("8.0.22"))
-                .Concat(NuGetMetadataReference.MicrosoftDataSqliteCore())
-                .Concat(NuGetMetadataReference.SystemDataSQLiteCore()));
+            NetStandardMetadataReference.Netstandard
```
andrei-epure-sonarsource, line 56: for my education: why was it working before?

andrei-epure-sonarsource (approved):
LGTM!

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/CookieShouldBeHttpOnlyTest.cs
andrei-epure-sonarsource: I realize now that the private one is inside a preprocessor directive, so it doesn't apply

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/ExecutingSqlQueriesTest.cs
costin-zaharia-sonarsource: In this test file we use `Verifier.VerifyAnalyzer` which adds the Netstandard if missing. See https://github.com/SonarSource/sonar-dotnet/blob/f75a387985c158f654dee052b644bf5b60c41b24/analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs#L77-L81

`Verifier.VerifyNoIssueReported` which is used by `SecurityHotspotTest` is not doing this so we have to add it manually. I think in the long term we will have to unify a bit the behavior.

PR #4265, opened 2021-04-19
[VB.NET] Rule S4581: Use 'Guid.NewGuid()' or 'Guid.Empty'

Corniel:
Implemented S4581 for VB.NET.

(Partly to experiment with Language facade).

Question: does this rule also imply that this is not desired?

``` C#
var guid = default(Guid);
Guid other = default;
```


pavel-mikula-sonarsource:
Hi @Corniel,

Thank you for giving it a try. I've crated the VB.NET RSPEC and branch https://github.com/SonarSource/sonar-dotnet/tree/Pavel/Corniel-S4581-Scaffolding with a scaffolding for you to implement the VB.NET part.

I would not use trackers in this rule. It can be done in a easy fashion without them. They are more suitable for more complex scenarios related mainly to security hotspots.

The C# `default` case is a good extension of the rule, I've update the RSPEC and created another branch https://github.com/SonarSource/sonar-dotnet/tree/Pavel/Corniel-S4581-CSharp-default for another PR if you'd like to implement it (probably after this PR is merged to avoid rebasing issues). We'll have to validate if it isn't too noisy once it's ready.

analyzers/src/SonarAnalyzer.CSharp/Rules/NewGuidShouldNotBeUsed.cs
```diff
@@ -30,30 +29,11 @@ namespace SonarAnalyzer.Rules.CSharp
...
-                        c.SemanticModel.GetSymbolInfo(objectCreationSyntax).Symbol is IMethodSymbol methodSymbol &&
-                        methodSymbol.ContainingType.Is(KnownType.System_Guid))
-                    {
-                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, objectCreationSyntax.GetLocation()));
-                    }
-                },
-                SyntaxKind.ObjectCreationExpression);
-        }
+        protected override int? ConstructorArgumentListCount(SyntaxNode node) =>
+             (node as ObjectCreationExpressionSyntax)?.ArgumentList?.Arguments.Count;
```
pavel-mikula-sonarsource, line 37: We should stick with the original approach with direct cast. We're registering for `SyntaxKind.ObjectCreationExpression` so there couldn't be anything else than `ObjectCreationExpressionSyntax`. The `?` part would not be coverable by tests.
```suggestion
            ((ObjectCreationExpressionSyntax)c.Node).ArgumentList?.Arguments.Count
```

analyzers/src/SonarAnalyzer.VisualBasic/Rules/NewGuidShouldNotBeUsed.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace SonarAnalyzer.Rules.VisualBasic
+{
+    [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
+    [Rule(DiagnosticId)]
+    public sealed class NewGuidShouldNotBeUsed : NewGuidShouldNotBeUsedBase<SyntaxKind>
+    {
+        protected override ILanguageFacade<SyntaxKind> Language => VisualBasicFacade.Instance;
+
+        protected override int? ConstructorArgumentListCount(SyntaxNode node) =>
+            (node as ObjectCreationExpressionSyntax)?.ArgumentList?.Arguments.Count;
```
pavel-mikula-sonarsource, line 37: Similar to C# - We should stick with the original approach with direct cast. We're registering for `SyntaxKind.ObjectCreationExpression` so there couldn't be anything else than `ObjectCreationExpressionSyntax`. The `?` part would not be coverable by tests.
```suggestion
            ((ObjectCreationExpressionSyntax)c.Node).ArgumentList?.Arguments.Count
```

pavel-mikula-sonarsource, line 37: _Hint:_ The `.ArgumentList?` part is coverable with `New Guid()` vs `new Guid`, but VS will not help with that. I'd use something like Notepad to do the edit before committing.

pavel-mikula-sonarsource (changes requested):
I left some improvements suggestions

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NewGuidShouldNotBeUsed.vb
```diff
@@ -0,0 +1,20 @@
+﻿Imports System
+
+Public Class NewGuid
+
+    Public Sub WithoutArguments()
+
+        Dim id As Guid = New Guid()  ' Noncompliant {{Use 'Guid.NewGuid()' or 'Guid.Empty' or add arguments to this Guid instantiation.}}
+        '                ^^^^^^^^^^
+    End Sub
```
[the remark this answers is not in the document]
Corniel: There are some challenges with `Nothing` (and `default`, `default(Guid)`) assignments: 
1. There are two syntax kinds for C# and one for VB (does that still fit the Language.SyntaxKind pattern)?
2. Even if you use the NothingLiteralExpression and Default(Literal)Expression, what is the best way forward from there?
3. What about this:

``` C#
private static readonly Guid empty;
var id = empty;
```
Are both lines invalid, or just the top one? Or is this allowed?

Corniel:
@pavel-mikula-sonarsource  So, I added the scaffolding branch and the update of the C# related resources.

The question is: which kind of `Guid id = default` should be allowed, and which should not? (see comments above)

Corniel:
I took up the challenge, and started implementing dealing with `default` and `Nothing`. The C#-only construct `default(Guid)` was easy to tackle, the other part is work in progress now. I rewrote the test scenarios to better suit the extended functionality.

It might be worth to rename the files to something like `GuidEmptyShouldBeExplicit`, as that is was this rule enforces.

Corniel:
Continuing in a fresh branch: #4814

PR #4269, opened 2021-04-20
Refactor MemberDescriptor to check name before symbol

### LINKED ISSUE #4267: Defer symbol access in MemberDescriptor.MatchesAny
See https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/src/SonarAnalyzer.Common/Helpers/MemberDescriptor.cs#L50

Now the symbol is retrieved before AST-based name checks:
```
            memberSymbol.Value is { } symbol
            && members.Any(x => HasSameName(memberName, x.Name, nameComparison) && x.HasSameContainingType(symbol, checkOverriddenMethods));
```

Should be : 

```
            members.Any(x => HasSameName(memberName, x.Name, nameComparison))
            && memberSymbol.Value is { } symbol
            && members.Any(x => x.HasSameContainingType(symbol, checkOverriddenMethods));

```

andrei-epure-sonarsource:
- MatchesAny - query semantic model only if the name matches
- IsMatch micro-optimization (1 string check is less expensive than Type.Is

Fixes #4267


analyzers/src/SonarAnalyzer.Common/Helpers/MemberDescriptor.cs
```diff
@@ -50,8 +51,9 @@ public bool IsMatch<TSymbolType>(string memberName, Lazy<TSymbolType> memberSymb
         public static bool MatchesAny<TSymbolType>(string memberName, Lazy<TSymbolType> memberSymbol, bool checkOverriddenMethods, StringComparison nameComparison, params MemberDescriptor[] members)
             where TSymbolType : class, ISymbol =>
             memberName != null
+            && members.Any(x => memberName.Equals(x.Name, nameComparison))
```
andrei-epure-sonarsource, line 54: HasSameName does null-check inside, already done here

costin-zaharia-sonarsource: Doesn't this change the functionality?

Previously we checked that the member with the same name had the correct containing type now we check if we have a name witch matches and a symbol witch containing type (but the symbol can be different).

costin-zaharia-sonarsource (changes requested):

costin-zaharia-sonarsource:
From what I can see the gate is green although the coverage is 80% (one branch is not covered). Do we have a threshold configured for a minimum number of lines?

analyzers/src/SonarAnalyzer.Common/Helpers/MemberDescriptor.cs
andrei-epure-sonarsource: very, very good catch, kudos for this! 🙏

andrei-epure-sonarsource:
RE: coverage - yes @costin-zaharia-sonarsource , I believe there is a minimum number of lines for the QG to be affected (but not sure where it's configured)

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/MethodSignatureHelperTest.cs
```diff
@@ -146,7 +146,38 @@ public void DoStuff(XmlNode node, XmlDocument doc)
         }
 
         [TestMethod]
-        public void IsMatch_AndCheckingOverrides_DoesMatchOverrides_VB()
+        public void MatchesAny_MethodAndTypeCombination_FindsCorrectOne()
+        {
+            var code = @"
```
costin-zaharia-sonarsource, line 151: code can be a constant

analyzers/src/SonarAnalyzer.Common/Helpers/MemberDescriptor.cs
```diff
@@ -39,7 +39,8 @@ internal MemberDescriptor(KnownType containingType, string name)
             $"{ContainingType.ShortName}.{Name}";
 
         public bool IsMatch(string memberName, ITypeSymbol containingType, StringComparison nameComparison) =>
-            containingType.Is(ContainingType) && HasSameName(memberName, Name, nameComparison);
+            HasSameName(memberName, Name, nameComparison)
```
costin-zaharia-sonarsource, line 42: I think we should cover also the missing branch for this method.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/MethodSignatureHelperTest.cs
```diff
@@ -124,7 +124,7 @@ End Namespace
         }
 
         [TestMethod]
-        public void IsMatch_AndCheckingOverrides_DoesMatchOverrides_CS()
+        public void MatchesAny_AndCheckingOverrides_DoesMatchOverrides_CS()
```
costin-zaharia-sonarsource, line 127: - I think we can rename the test file to `MemberDescriptorTest` since it appears to test only the `MemberDescriptor`.
 - I've checked and none of the tests from `MethodSignatureHelperTest.cs` calls `IsMatch`. It seems to be covered only indirectly by other tests.

costin-zaharia-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/MethodSignatureHelperTest.cs
andrei-epure-sonarsource: (I'm not sure the names are perfect now either, but I think they're ok-ish)

analyzers/src/SonarAnalyzer.Common/Helpers/MemberDescriptor.cs
```diff
@@ -50,8 +51,11 @@ public bool IsMatch<TSymbolType>(string memberName, Lazy<TSymbolType> memberSymb
         public static bool MatchesAny<TSymbolType>(string memberName, Lazy<TSymbolType> memberSymbol, bool checkOverriddenMethods, StringComparison nameComparison, params MemberDescriptor[] members)
             where TSymbolType : class, ISymbol =>
             memberName != null
```
andrei-epure-sonarsource, line 53: I can add a test for this, too

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/MemberDescriptorTest.cs
```diff
@@ -34,12 +35,88 @@
...
+}
+";
+            var snippet = new SnippetCompiler(code, MetadataReferenceFacade.SystemXml);
+            xmlNodeCloneNodeInvocationContext = CreateContextForMethod("XmlNode.CloneNode", snippet);
+        }
+
+        [TestMethod]
+        public void IsMatch_WhenMethodNameIsNull_ReturnsFalse()
+        {
+            var underTest = new MemberDescriptor(KnownType.System_Xml_XmlNode, "CloneNode");
```
costin-zaharia-sonarsource, line 66: `underTest`, I though we agreed to use `sut`

costin-zaharia-sonarsource (approved):
LGTM! There is only one comment regarding naming.

analyzers/src/SonarAnalyzer.Common/Helpers/MemberDescriptor.cs
```diff
@@ -64,7 +64,7 @@ private bool HasSameContainingType<TSymbolType>(TSymbolType memberSymbol, bool c
             where TSymbolType : class, ISymbol
         {
             var containingType = memberSymbol.ContainingType?.ConstructedFrom;
-            return checkOverriddenMethods ? containingType.DerivesOrImplements(ContainingType) : containingType.Is(ContainingType);
+            return containingType != null && checkOverriddenMethods ? containingType.DerivesOrImplements(ContainingType) : containingType.Is(ContainingType);
```
andrei-epure-sonarsource, line 67: there are null checks inside `DerivesOrImplements` and `Is`, but this is better for readability

costin-zaharia-sonarsource: Since **the method is private** we can accept that the parameter sanitization is done by the caller. From my point of view this is fine anyway.

costin-zaharia-sonarsource (approved):

PR #4270, opened 2021-04-21
Debug log when adding issues

pavel-mikula-sonarsource:


andrei-epure-sonarsource (approved):
LGTM

andrei-epure-sonarsource:
(code smells are tolerable)

Issue #4276, opened 2021-04-22
Rule S3264: FP when using +=


costin-zaharia-sonarsource:
### Description

[Rule S3264](https://rules.sonarsource.com/csharp/RSPEC-3264): a false positive is raised when chaining events. 

### Repro steps

```
namespace TestCases
{
    using System;

    public class EventChainingDirect
    {
        public event EventHandler Logging;

        public void RunTest()
        {
            EventTestBase eventTestBase = new EventTestBase();
            eventTestBase.Logging += Logging;
            eventTestBase.RunSomethingThatSendAnEvent();
        }
    }

    public class EventChainingLambda
    {
        public event EventHandler Logging;

        public void RunTest()
        {
            EventTestBase eventTestBase = new EventTestBase();
            eventTestBase.Logging += (s, e) => Logging(s, e);
            eventTestBase.RunSomethingThatSendAnEvent();
        }
    }

    public class EventTestBase
    {
        public event EventHandler Logging;

        public void RunSomethingThatSendAnEvent() => Logging(this, EventArgs.Empty);
    }
}
```

### Expected behavior

The issue should not be raised in this context.

### Actual behavior

`EventChainingDirect.Logging` is considered non-compliant although the event is invoked by `RunSomethingThatSendAnEvent`

### Known workarounds

Using a lambda function like in `EventChainingLambda`

### Related information

* C#/VB.NET Plugins version: 8.21

MrDaedra:
We are facing the same issue but not even with direct chaining. For `ProcessMessageAsync` event we get S3264: Remove the unused event 'ProcessErrorAsync' or invoke it.
```C#
internal class ServiceBusMessageProcessor
{
    public event Func<ProcessBrokerMessageEventArgs, Task> ProcessMessageAsync;

    public ServiceBusMessageProcessor(ServiceBusClient serviceBusClient, string topic, string subscriptionName)
    {
        var processor = serviceBusClient.CreateProcessor(topic, subscriptionName, new ServiceBusProcessorOptions());
        processor.ProcessMessageAsync += OnProcessMessageAsync;
    }

    private async Task OnProcessMessageAsync(ProcessMessageEventArgs arg)
    {
        await ProcessMessageAsync?.Invoke(new ProcessServiceBusMessageEventArgs(arg));
    }
}
```

pavel-mikula-sonarsource:
Internal ticket [NET-1914](https://sonarsource.atlassian.net/browse/NET-1914)

[...]

PR #4277, opened 2021-04-22
Add repro for #4276

costin-zaharia-sonarsource:
Related to #4276

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UninvokedEventDeclaration.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UninvokedEventDeclaration.cs
@@ -227,9 +227,16 @@ namespace Tests.Diagnostics
         }
     }
 
+    public class EventTestBase
+    {
+        public event EventHandler Logging;
+
+        public void RunSomethingThatSendAnEvent() => Logging(this, EventArgs.Empty);
+    }
+
     public class EventChainingDirect
     {
-        public event EventHandler Logging; // Noncompliant - FP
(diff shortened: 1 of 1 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UninvokedEventDeclaration.cs
```diff
@@ -226,4 +226,35 @@ public void SomeMethod()
             }
         }
     }
+
+    public class EventChainingDirect
+    {
+        public event EventHandler Logging; // Noncompliant - FP
```
andrei-epure-sonarsource, line 232: could you also add a comment with a link to the issue, please?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UninvokedEventDeclaration.cs
```diff
@@ -226,4 +226,35 @@ public void SomeMethod()
...
+
+        public void RunTest()
+        {
+            EventTestBase eventTestBase = new EventTestBase();
+            eventTestBase.Logging += (s, e) => Logging(s, e);
+            eventTestBase.RunSomethingThatSendAnEvent();
+        }
+    }
+
+    public class EventTestBase
```
andrei-epure-sonarsource, line 254: could you please move this above `EventChainingDirect`? because I was confused if `EventChainingLambda` had something to do with the repro or not...

But I understand now that `EventChainingDirect` and `EventChainingLambda` are just to separate use cases with `EventTestBase`, and they don't interact with each other.

Thanks!

andrei-epure-sonarsource (approved):
LGTM, I left some comments to address before merge

PR #4278, opened 2021-04-22
Improve unit test report error message

### LINKED ISSUE #4259: Improve unit test report error message
As reported in [internal forum](https://discuss.sonarsource.com/t/nunit-file-import-failures-difficult-to-troubleshoot/7346), [the root cause](https://github.com/SonarSource/sonar-dotnet/blob/1af2e34548c0873aa437a11d02f9d51373ed7177/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/NUnitTestResultsFileParser.java#L58) of any UT report import error is [eaten](https://github.com/SonarSource/sonar-dotnet/blob/1af2e34548c0873aa437a11d02f9d51373ed7177/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsAggregator.java#L105), exception message is ignored and the situation is hard to diagnose.

We should include the exception message in the warn log.

pavel-mikula-sonarsource:
Fixes #4259

### DIFF
--- a/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsAggregator.java
+++ b/sonar-dotnet-shared-library/src/main/java/org/sonar/plugins/dotnet/tests/UnitTestResultsAggregator.java
@@ -102,7 +102,7 @@ public class UnitTestResultsAggregator {
           try {
             parser.accept(reportFile, unitTestResults);
           } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


csaba-sagi-sonarsource (approved):

csaba-sagi-sonarsource (approved):
LGTM

PR #4280, opened 2021-04-22
Revert "Fix RegEx hotspots in DotCoverReportParser (#4212)"

pavel-mikula-sonarsource:
Revert for master branch


andrei-epure-sonarsource:
Related to https://community.sonarsource.com/t/stackoverflowerror-error-during-sonarscanner-execution/42089 . 

Commit c0a1207 was fixing security hotspots without a functional need. Removing this from master is safer until we properly understand what the root cause is.

andrei-epure-sonarsource (approved):
LGTM

PR #4287, opened 2021-04-22
Don't publish CFG in 8.21 branch

pavel-mikula-sonarsource:
Avoiding repox issues:

>[!] [409] Server Error
Promoting nuget SonarAnalyzer.CFG.CSharp.8.21.1.30749.nupkg from sonarsource-nuget-qa to sonarsource-nuget-dev
##[error]The process 'C:\hostedtoolcache\windows\Python\3.9.2\x64\python.exe' failed because one or more lines were written to the STDERR stream
Finishing: promoteRepox


pavel-mikula-sonarsource:
Not valid, problem is with the version of nuget package

PR #4288, opened 2021-04-22
Fix nuspec version

pavel-mikula-sonarsource:
We have a problem with set-version script


pavel-mikula-sonarsource:
No we don't 🤦‍♂️

PR #4293, opened 2021-04-23
Update SLVS config for connected mode

andrei-epure-sonarsource:
The QP on SonarCloud updated, we need to change the auto-generated SLVS files to avoid conflicts.

### DIFF
--- a/analyzers/.sonarlint/SonarAnalyzer.slconfig
+++ b/analyzers/.sonarlint/SonarAnalyzer.slconfig
@@ -9,7 +9,7 @@
   "Profiles": {
     "CSharp": {
       "ProfileKey": "AWT63_RpV-oLu4FdMlyI",
-      "ProfileTimestamp": "2020-12-28T10:06:24Z"
+      "ProfileTimestamp": "2021-03-26T13:27:24Z"
     }
   }
 }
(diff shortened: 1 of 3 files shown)


analyzers/.sonarlint/SonarAnalyzer.slconfig
```diff
@@ -9,7 +9,7 @@
   "Profiles": {
     "CSharp": {
       "ProfileKey": "AWT63_RpV-oLu4FdMlyI",
-      "ProfileTimestamp": "2020-12-28T10:06:24Z"
+      "ProfileTimestamp": "2021-03-26T13:27:24Z"
```
andrei-epure-sonarsource, line 12: @csaba-sagi-sonarsource I now realize that these are auto-generated files by SonarLint and maybe we should just ignore the SLVS folder.

WDYT?

csaba-sagi-sonarsource: Indeed, I agree with you.

csaba-sagi-sonarsource (approved):
LGTM. In a separate PR the SLVS folder can be ignored if the others agree with it too.

analyzers/.sonarlint/SonarAnalyzer.slconfig
andrei-epure-sonarsource: the only problem is that it's referenced in each project... : `    <AdditionalFiles Include="..\..\.sonarlint\sonaranalyzer-dotnet\CSharp\SonarLint.xml" Link="Properties\SonarLint.xml" />`

PR #4299, opened 2021-04-23
Fix S2589 FP: Tuple Assignment

### LINKED ISSUE #3239: Fix S2589 FP: Loop with Tuple Assignment
### Description

When using a flag that is only updated through a tuple assignment, S2589 will be raised on conditions using that flag.

### Repro steps

```csharp
var tmp = 0;
var flag = true;
while(flag)
    (flag, tmp) = (false, 5);

Console.WriteLine(tmp);
```

### Expected behavior

No warning raised.

### Actual behavior

Raised warning: S2589: Change this condition so that it does not always evaluate to true.

### Related information

* SonarC# Version 8.5.0.15942
* Visual Studio Version: 16.4.6
* dotnet version: netcoreapp 3.1
* Operating System: Win10

### LINKED ISSUE #4257: Fix S2583 FP: Value assigned with Tuple
### Description

"S2583 Change this condition so that it does not always evaluate to 'false'; some subsequent code is never executed" when we are using Tuple.

### Repro steps

```
public void Testing()
{
	MemoryStream myVariable = null;
	string myString = null;

	(myVariable, myString) = GetData();

	if (myVariable != null) // S2583
	{
		// some code
	}
}

public (MemoryStream, string) GetData()
{
	return (null, null);
}
```

### Expected behavior

S2583 should not be reported in this case

### Actual behavior

S2583 warning is given

### Known workarounds

Replace:

`MemoryStream myVariable = null;`

with 

`MemoryStream myVariable;`

### Related information

* C#/VB.NET Plugins version: 9.0 / 4.33.1.28913
* Visual Studio version: 16.10.0.Preview 1.0
* MSBuild / dotnet version: net5.0
* SonarScanner for .NET version (if used)
* Operating System: Win 10

pavel-mikula-sonarsource:
Fixes #3239
Fixes #4257

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ConditionEvaluatesToConstant.cs
@@ -142,22 +142,25 @@ namespace SonarAnalyzer.Rules.CSharp
                 return Enumerable.Empty<Diagnostic>()
                     .Union(conditionTrue
                         .Except(conditionFalse)
-                        .Where(x => !IsInsideCatchOrFinallyBlock(x) && !IsConditionOfLoopWithBreak(x))
+                        .Where(x => !IsMuted(x) && !IsInsideCatchOrFinallyBlock(x) && !IsConditionOfLoopWithBreak(x))
                         .Select(x => GetNodeDiagnostics(x, true)))
                     .Union(conditionFalse
                         .Except(conditionTrue)
-                        .Where(x => !IsInsideCatchOrFinallyBlock(x))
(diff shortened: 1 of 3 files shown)


pavel-mikula-sonarsource:
This is a draft to avoid merging to 8.22 milestone.

csaba-sagi-sonarsource (approved):
LGTM. In my opinion MutedSyntaxWalker is a bit misleading name as it does not mute anything. I would  give it a name that reflects that it  checks for tuple assignments, and  at the IsMuted usage I would check for IsInTupleAssignment and than add a comment there that we  mute it currently.

pavel-mikula-sonarsource:
I plan to add more scenarios to the Walker in next PRs, that why I've chosen more generic name. And the class should go away in the future.

PR #4300, opened 2021-04-23
Fix S1751 FP: Triggered when returning a property inside try catch block

### LINKED ISSUE #2522: Fix S1751 FP: triggered when returning a property
### Description

returning a property in a try catch confuses the analyser

### Repro steps

```csharp
class C
{
	int P => throw new Exception();
	int M()
	{
		while (true)
		{
			try
			{
				return P;
			}
			catch (Exception ex)
			{
			}
		}
	}
}
```

### Expected behavior

No Exception

### Actual behavior

S1751: Refactor the containing loop to do more than one iteration

### Related information

* SonarC# Version: 7.14.0.8411
* Visual Studio Version: 16.2

csaba-sagi-sonarsource:
Fixes #2522

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/UnconditionalJumpStatement.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/UnconditionalJumpStatement.cs
@@ -80,6 +80,20 @@ namespace SonarAnalyzer.Rules.CSharp
             {
             }
 
+            protected override bool IsAccessToClassMember(StatementSyntax node)
+            {
+                var returnStatementExpression = ((ReturnStatementSyntax)node).Expression;
+                if (returnStatementExpression is IdentifierNameSyntax identifier
+                    && semanticModel.GetSymbolInfo(identifier) is { } symbol
+                    && symbol.Symbol.Kind == SymbolKind.Property)
+                {
+                    return true;
+                }
+
+                // We are checking for memberAccessExpression to catch NullReferenceException.
+                return returnStatementExpression is MemberAccessExpressionSyntax memberAccessExpression;
+            }
+
             public override void Visit()
             {
                 var csWalker = new CsLoopwalker(this);
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/UnconditionalJumpStatement.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/UnconditionalJumpStatement.cs
@@ -19,7 +19,6 @@
  */
 
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -36,11 +35,8 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class UnconditionalJumpStatement : UnconditionalJumpStatementBase<StatementSyntax, SyntaxKind>
     {
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        protected override ILanguageFacade<SyntaxKind> Language => CSharpFacade.Instance;
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
-        protected override GeneratedCodeRecognizer GeneratedCodeRecognizer => CSharpGeneratedCodeRecognizer.Instance;
         protected override ISet<SyntaxKind> LoopStatements { get; } = new HashSet<SyntaxKind>
         {
             SyntaxKind.ForEachStatement,
@@ -51,7 +47,6 @@ namespace SonarAnalyzer.Rules.CSharp
 
         protected override LoopWalkerBase<StatementSyntax, SyntaxKind> GetWalker(SyntaxNodeAnalysisContext context)
             => new LoopWalker(context, LoopStatements);
-
 
         private class LoopWalker : LoopWalkerBase<StatementSyntax, SyntaxKind>
         {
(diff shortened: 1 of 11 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UnconditionalJumpStatement.vb
```diff
@@ -380,5 +380,39 @@ Namespace Tests.Diagnostics
             Throw New NotImplementedException()
         End Sub
     End Class
+
+    Public Class ReproducerClass
+        ReadOnly Property FirstProperty As ReproducerClass
```
pavel-mikula-sonarsource, line 385: _Nitpick:_ There should be EOL around `Class` and `End Class`
```suggestion
    Public Class ReproducerClass
    
        ReadOnly Property FirstProperty As ReproducerClass
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UnconditionalJumpStatement.vb
```diff
@@ -380,5 +380,39 @@ Namespace Tests.Diagnostics
...
+        Public Function MethodWithMemberAccess() As Int32
+              While True
+                Try
+                    Return FirstProperty.SecondProperty ' Compliant
+                Catch E As Exception
+
+                End Try
+            End While
+        End Function
+    End Class
```
pavel-mikula-sonarsource, line 416: _Nitpick:_ NL before and after `End Class`
```suggestion
        End Function
        
    End Class
    
```

analyzers/src/SonarAnalyzer.Common/Rules/UnconditionalJumpStatementBase.cs
```diff
@@ -19,29 +19,36 @@
...
     {
         protected const string DiagnosticId = "S1751";
-        protected const string MessageFormat = "Refactor the containing loop to do more than one iteration.";
+        private const string MessageFormat = "Refactor the containing loop to do more than one iteration.";
 
-        protected abstract GeneratedCodeRecognizer GeneratedCodeRecognizer { get; }
+        private readonly DiagnosticDescriptor rule;
 
-        protected abstract LoopWalkerBase<TStatementSyntax, TLanguageKindEnum> GetWalker(SyntaxNodeAnalysisContext context);
+        protected abstract ILanguageFacade<TSyntaxKind> Language { get; }
```
pavel-mikula-sonarsource, line 39: This and the next abstract property `LoopStatements` should be together. And they won't need empty line between them.

analyzers/src/SonarAnalyzer.Common/Rules/UnconditionalJumpStatementBase.cs
```diff
@@ -99,16 +110,16 @@ public List<TStatementSyntax> GetRuleViolations()
         {
             var ancestors = node
                 .Ancestors()
-                .TakeWhile(n => !this.rootExpression.Equals(n))
+                .TakeWhile(n => !rootExpression.Equals(n))
```
pavel-mikula-sonarsource, line 113: Line 108/109: Empty line in the middle of method signature makes it hard to read

analyzers/src/SonarAnalyzer.CSharp/Rules/UnconditionalJumpStatement.cs
```diff
@@ -75,15 +70,43 @@ private class LoopWalker : LoopWalkerBase<StatementSyntax, SyntaxKind>
...
 
-            public override void Visit()
+            protected override bool IsAccessToClassMember(StatementSyntax node)
             {
-                var csWalker = new CsLoopwalker(this);
-                csWalker.SafeVisit(this.rootExpression);
+                var returnStatementExpression = ((ReturnStatementSyntax)node).Expression;
+                if (returnStatementExpression is IdentifierNameSyntax identifier
+                    && semanticModel.GetSymbolInfo(identifier) is { } symbolInfo
+                    && symbolInfo.Symbol is { } symbol
```
pavel-mikula-sonarsource, line 85: `SymbolInfo` is a struct so it cannot be `null`. You can go directly for it's symbol
```suggestion
                    && semanticModel.GetSymbolInfo(identifier).Symbol is { } symbol
```

analyzers/src/SonarAnalyzer.CSharp/Rules/UnconditionalJumpStatement.cs
```diff
@@ -75,15 +70,43 @@ private class LoopWalker : LoopWalkerBase<StatementSyntax, SyntaxKind>
...
+                    && symbol.Kind == SymbolKind.Property)
+                {
+                    return true;
+                }
+
+                // We are checking for memberAccessExpression to catch NullReferenceException.
+                return returnStatementExpression is MemberAccessExpressionSyntax memberAccessExpression;
```
pavel-mikula-sonarsource, line 92: Unused variable
```suggestion
                return returnStatementExpression is MemberAccessExpressionSyntax;
```

pavel-mikula-sonarsource, line 92: We should also check for `ElementAccessExpression` for `return arr[0]`

pavel-mikula-sonarsource, line 92: We should check `returnStatementExpression.DescendantNodesAndSelf().Any(x => x.IsAnyKind(...` to cover all nested cases. Like `(((arr[0])))` or other unexpected syntaxes like `return x = member.Item`

pavel-mikula-sonarsource (changes requested):
Improvements suggested

csaba-sagi-sonarsource:
WDYT about introducing ISyntaxHelper to the LanguageFacade to remove the duplication?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UnconditionalJumpStatement.vb
pavel-mikula-sonarsource: I wrote 
>and after `End Class`
🙂

analyzers/src/SonarAnalyzer.CSharp/Rules/UnconditionalJumpStatement.cs
```diff
@@ -75,15 +74,34 @@ private class LoopWalker : LoopWalkerBase<StatementSyntax, SyntaxKind>
...
             {
+                var csWalker = new CsLoopwalker(this);
+                csWalker.SafeVisit(rootExpression);
             }
 
-            public override void Visit()
+            protected override bool IsPropertyAccess(StatementSyntax node) =>
+                node.DescendantNodes().Any(x => x.IsKind(SyntaxKind.IdentifierName)
+                                                && x is IdentifierNameSyntax identifier
+                                                && semanticModel.GetSymbolInfo(identifier).Symbol is { } symbol
```
pavel-mikula-sonarsource, line 87: We don't need `IsKind` check before `is IdentifierNameSyntax` in this case. That can simplify the syntax like this:
```suggestion
                node.DescendantNodes().OfType<IdentifierNameSyntax>().Any(x => semanticModel.GetSymbolInfo(x).Symbol is { } symbol
```

analyzers/src/SonarAnalyzer.CSharp/Rules/UnconditionalJumpStatement.cs
```diff
@@ -75,15 +74,34 @@ private class LoopWalker : LoopWalkerBase<StatementSyntax, SyntaxKind>
...
+                                                && symbol.Kind == SymbolKind.Property);
+
+            protected override bool IsAnyKind(SyntaxNode node, ISet<SyntaxKind> syntaxKinds) => node.IsAnyKind(syntaxKinds);
+            protected override bool IsReturnStatement(SyntaxNode node) => node.IsKind(SyntaxKind.ReturnStatement);
```
pavel-mikula-sonarsource, line 91: Formatting (EOL between them, as they are not properties and they are not trivial)
```suggestion
            protected override bool IsAnyKind(SyntaxNode node, ISet<SyntaxKind> syntaxKinds) => 
                node.IsAnyKind(syntaxKinds);
            
            protected override bool IsReturnStatement(SyntaxNode node) => 
                node.IsKind(SyntaxKind.ReturnStatement);
```

analyzers/src/SonarAnalyzer.VisualBasic/Rules/UnconditionalJumpStatement.cs
```diff
@@ -83,15 +79,36 @@ private class LoopWalker : LoopWalkerBase<StatementSyntax, SyntaxKind>
...
 
             public override void Visit()
             {
                 var vbWalker = new VbLoopwalker(this);
-                vbWalker.SafeVisit(this.rootExpression);
+                vbWalker.SafeVisit(rootExpression);
+            }
+
+            protected override bool IsPropertyAccess(StatementSyntax node) =>
+                node.DescendantNodes().Any(x => x.IsKind(SyntaxKind.IdentifierName)
```
pavel-mikula-sonarsource, line 91: Same as C# in this file

pavel-mikula-sonarsource (changes requested):
Last round

pavel-mikula-sonarsource (approved):
LGTM

PR #4301, opened 2021-04-23
Fix S5332 FP: Don't raise for known schema and example domains

### LINKED ISSUE #4171: Fix S5332 FP: Don't raise for known schema and example domains
From [internal discussion](https://discuss.sonarsource.com/t/ignore-http-protocol-on-xmlns-fields/7229) as defined for other languages:

Improvement similar to #4124 

## No XML namespaces
No issue should be raised if the domain is one of the following as they are commonly used for XML or similar namespaces:

www.w3.org
xml.apache.org
schemas.xmlsoap.org
schemas.openxmlformats.org
rdfs.org
purl.org
xmlns.com
schemas.google.com
a9.com
ns.adobe.com
ltsc.ieee.org
docbook.org
graphml.graphdrawing.org
json-schema.org

## No example domains
No issue should be raised if the domain is one of the following. Also, no issue should be raised on all sub-domains of this domains:

example.com
example.org
test.com

dmitry-golovinov-sonarsource:
Fixes #4171

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
@@ -94,7 +94,10 @@ namespace SonarAnalyzer.Rules.CSharp
         public ClearTextProtocolsAreSensitive(IAnalyzerConfiguration analyzerConfiguration) : base(analyzerConfiguration)
         {
             const string allSubdomainsPattern = @"([^/?#]+\.)?";
-            var domainsList = localhostAddresses.Concat(commonlyUsedXmlDomains).Concat(commonlyUsedExampleDomains.Select(x => allSubdomainsPattern + x));
+            var domainsList = localhostAddresses
+                .Concat(commonlyUsedXmlDomains)
+                    .Select(Regex.Escape)
+                    .Concat(commonlyUsedExampleDomains.Select(x => allSubdomainsPattern + Regex.Escape(x)));
             var validServerPattern = domainsList.JoinStr("|");
 
             httpRegex = CompileRegex(@$"^http:\/\/(?!{validServerPattern}).");
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
@@ -55,7 +55,7 @@ namespace SonarAnalyzer.Rules.CSharp
             {"clear-text SMTP", "SMTP over SSL/TLS or SMTP with STARTTLS" }
         };
 
-        private readonly List<string> commonlyUsedXmlDomains = new List<string>
+        private readonly string[] commonlyUsedXmlDomains =
         {
             "www.w3.org",
             "xml.apache.org",
@@ -73,8 +73,8 @@ namespace SonarAnalyzer.Rules.CSharp
             "json-schema.org"
         };
 
-        private readonly List<string> commonlyUsedExampleDomains = new List<string> {"example.com", "example.org", "test.com"};
-        private readonly List<string> localhostAddresses = new List<string> {"localhost", "127.0.0.1", "::1"};
+        private readonly string[] commonlyUsedExampleDomains = {"example.com", "example.org", "test.com"};
+        private readonly string[] localhostAddresses = {"localhost", "127.0.0.1", "::1"};
 
         private readonly CSharpObjectInitializationTracker objectInitializationTracker =
             new CSharpObjectInitializationTracker(constantValue => constantValue is bool value && value,
@@ -93,8 +93,8 @@ namespace SonarAnalyzer.Rules.CSharp
 
         public ClearTextProtocolsAreSensitive(IAnalyzerConfiguration analyzerConfiguration) : base(analyzerConfiguration)
         {
-            const string allSubdomainsPattern = "([^/]+\\.)?";
-            var domainsList = commonlyUsedXmlDomains.Concat(commonlyUsedExampleDomains.Select(x => allSubdomainsPattern + x)).Concat(localhostAddresses);
+            const string allSubdomainsPattern = @"([^/?#]+\.)?";
+            var domainsList = localhostAddresses.Concat(commonlyUsedXmlDomains).Concat(commonlyUsedExampleDomains.Select(x => allSubdomainsPattern + x));
             var validServerPattern = domainsList.JoinStr("|");
 
             httpRegex = CompileRegex(@$"^http:\/\/(?!{validServerPattern}).");
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
@@ -20,6 +20,7 @@
 
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Linq;
 using System.Text.RegularExpressions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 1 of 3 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
```diff
@@ -61,16 +55,54 @@ public sealed class ClearTextProtocolsAreSensitive : HotspotDiagnosticAnalyzer
             {"clear-text SMTP", "SMTP over SSL/TLS or SMTP with STARTTLS" }
         };
 
+        private readonly List<string> commonlyUsedXmlDomains = new List<string>
```
pavel-mikula-sonarsource, line 58: We'll never do `.Add` operation => we don't need a list. This can be an array. Same for others.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
```diff
@@ -61,16 +55,54 @@ public sealed class ClearTextProtocolsAreSensitive : HotspotDiagnosticAnalyzer
...
+        private readonly Regex validServerRegex;
+
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(DefaultRule, EnableSslRule);
 
         public ClearTextProtocolsAreSensitive() : this(AnalyzerConfiguration.Hotspot) { }
 
-        public ClearTextProtocolsAreSensitive(IAnalyzerConfiguration analyzerConfiguration) : base(analyzerConfiguration) { }
+        public ClearTextProtocolsAreSensitive(IAnalyzerConfiguration analyzerConfiguration) : base(analyzerConfiguration)
+        {
+            const string allSubdomainsPattern = "([^/]+\\.)?";
```
pavel-mikula-sonarsource, line 96: Let's use verbatim syntax to make it readable (eternal pain of C-like languages)
```suggestion
            const string allSubdomainsPattern = @"([^/]+\.)?";
```

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
```diff
@@ -61,16 +55,54 @@ public sealed class ClearTextProtocolsAreSensitive : HotspotDiagnosticAnalyzer
...
+            var domainsList = commonlyUsedXmlDomains.Concat(commonlyUsedExampleDomains.Select(x => allSubdomainsPattern + x)).Concat(localhostAddresses);
```
pavel-mikula-sonarsource, line 97: I'd swap the order to make it easier to read (simple stuff together, more complicated at the end)
```suggestion
            var domainsList = commonlyUsedXmlDomains.Concat(localhostAddresses).Concat(commonlyUsedExampleDomains.Select(x => allSubdomainsPattern + x));
```

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
```diff
@@ -61,16 +55,54 @@ public sealed class ClearTextProtocolsAreSensitive : HotspotDiagnosticAnalyzer
...
+            var validServerPattern = domainsList.JoinStr("|");
```
pavel-mikula-sonarsource, line 98: We need to escape the input data (and we have a very useful overload for that)
```suggestion
            var validServerPattern = domainsList.JoinStr("|", x => Regex.Escape(x));
```

pavel-mikula-sonarsource (changes requested):
First round, I haven't reviewed all yet

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
pavel-mikula-sonarsource, line 96: What's the idea behind `\` in `([^/]+\.)?`? To avoid overrunning to path part of uri in cases like `http://badidea.com/example.com` or `http://badidea.com/path?domain=example.com`? 
Let's add unit tests for such scenarios.

pavel-mikula-sonarsource: If yes, what about `http://badidea.com?domain=example.com`? We should avoid `/?#@:`. Let's add UTs for all of them

pavel-mikula-sonarsource (changes requested):
One more suggestion

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
pavel-mikula-sonarsource: Update: This is even better. localhost, then common domains, than example domains (to star simple and keep domains together)
```suggestion
            var domainsList = localhostAddresses.Concat(commonlyUsedXmlDomains).Concat(commonlyUsedExampleDomains.Select(x => allSubdomainsPattern + x));
```

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
dmitry-golovinov-sonarsource: I do not see why @ and : are bad. Basically, this way you can specify the login and password inside the url.

Seems to be valid "example" url `http://user:passwd@example.com`. Wdyt?

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
dmitry-golovinov-sonarsource: We cannot do that, since some domains contain regular expressions that should be interpreted as regular expressions

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
pavel-mikula-sonarsource: You're right, `:` and `@` are fine

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
pavel-mikula-sonarsource: We need to encode it selectively then. In more complicated way. Something like
```
var domainsList = localhostAddresses
    .Concat(commonlyUsedXmlDomains)
	.Select(Regex.Escape)
	.Concat(commonlyUsedExampleDomains.Select(x => allSubdomainsPattern + Regex.Escape(x)));
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/ClearTextProtocolsAreSensitive.cs
```diff
@@ -148,14 +148,66 @@ public void Ftp()
...
+            "http://subdomain.rdfs.org", // Noncompliant
+            "http://subdomain.purl.org", // Noncompliant
+            "http://subdomain.xmlns.com", // Noncompliant
+            "http://subdomain.schemas.google.com", // Noncompliant
+            "http://subdomain.a9.com", // Noncompliant
+            "http://subdomain.ns.adobe.com", // Noncompliant
+            "http://subdomain.ltsc.ieee.org", // Noncompliant
+            "http://subdomain.docbook.org", // Noncompliant
+            "http://subdomain.graphml.graphdrawing.org", // Noncompliant
+            "http://subdomain.json-schema.org", // Noncompliant
```
pavel-mikula-sonarsource, line 195: Indenting the annotations, for example like this, makes it easier to read.
```suggestion
            "http://subdomain.rdfs.org", 		// Noncompliant
            "http://subdomain.purl.org", 		// Noncompliant
            "http://subdomain.xmlns.com", 		// Noncompliant
            "http://subdomain.schemas.google.com", // Noncompliant
            "http://subdomain.a9.com", 			// Noncompliant
            "http://subdomain.ns.adobe.com", 	// Noncompliant
            "http://subdomain.ltsc.ieee.org", 	// Noncompliant
            "http://subdomain.docbook.org", 	// Noncompliant
            "http://subdomain.graphml.graphdrawing.org", // Noncompliant
            "http://subdomain.json-schema.org", // Noncompliant
```
We don't need to ident all of them to far end, but indenting them in a reasonable groups helps a lot to separate testcase from it's expected result without optical mix of action (comment) and data (syntax) on nearby rows.

pavel-mikula-sonarsource (changes requested):
Final polishing, mainly due to the missing escape

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/ClearTextProtocolsAreSensitive.cs
```diff
@@ -148,14 +148,66 @@ public void Ftp()
...
+            "http://example.org",
+            "http://test.com",
+            "http://subdomain.example.com",
+            "http://subdomain.example.org",
+            "http://subdomain.test.com",
+            "http://email:password@subdomain.example.com",
+            "http://domain.com/example.com",               // Noncompliant
+            "http://domain.com/path?domain=.example.com",  // Noncompliant
+            "http://domain.com?q=.example.com",            // Noncompliant
+            "http://domain.com#.example.com",              // Noncompliant
```
pavel-mikula-sonarsource, line 210: There are few unneeded `.` here

pavel-mikula-sonarsource (approved):
LGTM. I left one optional nitpick comment

PR #4303, opened 2021-04-23
S1751 FP fix with Anonymous Method Expression

### LINKED ISSUE #2441: S1751 False positive: delegate returning value wrongly considered as affecting loop
**Repro steps**

```
public delegate T FalsePositive1_Delegate<out T>();

/// <summary>
/// The return in a delegate is wrongly considered as a return in the owner loop.
/// </summary>
private static void FalsePositive()
{
	List<object> entries = new List<object>() { new object() };

	foreach (object entry in entries)
	{
		FalsePositive_Sub<int>(
			delegate () { return -1; }  // Return in delegate is considered as return in the foreach.
			);
	}

	// Went past the loop.
	Console.WriteLine("Went past the loop");        //NOSONAR.
}

private static void FalsePositive_Sub<T>(FalsePositive1_Delegate<T> action)
{
	// Nothing to do.
}
```

**Expected behavior**

Rule S1751 should not be triggered

**Actual Behavior**

The return in the delegate is wrongly considered as early exit of the loop

**Related information**

Sonar C# version  7.11

csaba-sagi-sonarsource:
Fixes #2441

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/UnconditionalJumpStatement.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/UnconditionalJumpStatement.cs
@@ -38,12 +38,12 @@ namespace SonarAnalyzer.Rules.CSharp
         protected override ILanguageFacade<SyntaxKind> Language => CSharpFacade.Instance;
 
         protected override ISet<SyntaxKind> LoopStatements { get; } = new HashSet<SyntaxKind>
-            {
-                SyntaxKind.ForEachStatement,
-                SyntaxKind.ForStatement,
-                SyntaxKind.WhileStatement,
-                SyntaxKind.DoStatement
-            };
+        {
+            SyntaxKind.ForEachStatement,
+            SyntaxKind.ForStatement,
+            SyntaxKind.WhileStatement,
+            SyntaxKind.DoStatement
+        };
 
         protected override LoopWalkerBase<StatementSyntax, SyntaxKind> GetWalker(SyntaxNodeAnalysisContext context)
             => new LoopWalker(context, LoopStatements);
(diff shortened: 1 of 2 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/UnconditionalJumpStatement.cs
```diff
@@ -42,12 +42,12 @@ public sealed class UnconditionalJumpStatement : UnconditionalJumpStatementBase<
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
         protected override GeneratedCodeRecognizer GeneratedCodeRecognizer => CSharpGeneratedCodeRecognizer.Instance;
         protected override ISet<SyntaxKind> LoopStatements { get; } = new HashSet<SyntaxKind>
-        {
-            SyntaxKind.ForEachStatement,
-            SyntaxKind.ForStatement,
-            SyntaxKind.WhileStatement,
-            SyntaxKind.DoStatement
-        };
+            {
```
costin-zaharia-sonarsource, line 45: pedantic: I think we don't need to shift this with 4 spaces. The other field you changed is aligned with the visibility modifier.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UnconditionalJumpStatement.cs
```diff
@@ -544,4 +544,32 @@ static void Throw()
...
+            {
+                Delegate1 d1 = delegate () { return -1; }; // Compliant
+            }
+        }
+
+        Func<int> myFunc3()
+        {
+            while (true)
+            {
+                return (Func<int>)delegate () { return -1; }; // Noncompliant
```
costin-zaharia-sonarsource, line 571: You can add a link to the issue.

costin-zaharia-sonarsource (changes requested):
It looks good. I've added just a two minor comments.

costin-zaharia-sonarsource (approved):
LGTM!

PR #4306, opened 2021-04-26
Fix S2368 FP: Params keyword on jagged arrays

### LINKED ISSUE #4225: Don't raise S2368 which complains about multidimensional arrays when params keyword is declared on jagged arrays
### Description

The rule S2368 currently fires when the `params` keyword is used on jagged (nested) arrays.

However, at least for two-level jagged arrays declared with `params`, the caller of the function usually just experiences them as a variadic number of distinct arguments of the same type, so the warning should not be triggered in this case.

Additionally, the diagnostics should be fixed to stop calling jagged arrays "multidimensional", as multidimensional arrays (like `byte[,]`) are something different as jagged (nested) arrays (like `byte[][]`). 

### Repro steps

```
namespace SonarLintBugS2368
{
    using System;

    public static class GuidHelper
    {
        public static Guid CreateNameBasedGuid(params byte[][] nameComponents)
        {
            return Guid.Empty;
        }
    }
}
```

### Expected behavior

S2368 does not raise on this method.

### Actual behavior

S2368 is raised on this method, and additionally falsely claims the jagged array to be a multidimensional array.

### Known workarounds

Declare the method as private, and provide N public wrapper methods with 1, 2, 3, ... n parameters of type byte[].

### Related information

* C# Version 8.0 (but works with older versions, too).
* Visual Studio version 2019
* MSBuild / dotnet version: .NET 4.8 / .NET Core 3.1
* SonarScanner for .NET version (if used): SonarScanner for MSBuild 4.9/SonarScanner 4.3.0.2102 (Also reproducible with SonarLint for VS 2019 Version 4.33.1.28913)

csaba-sagi-sonarsource:
Fixes #4225

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Rules/PublicMethodWithMultidimensionalArrayBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/PublicMethodWithMultidimensionalArrayBase.cs
@@ -28,7 +28,7 @@ namespace SonarAnalyzer.Rules.Common
     public abstract class PublicMethodWithMultidimensionalArrayBase : SonarDiagnosticAnalyzer
     {
         protected const string DiagnosticId = "S2368";
-        protected const string MessageFormat = "Make this method private or simplify its parameters to not use multidimensional arrays.";
+        protected const string MessageFormat = "Make this method private or simplify its parameters to not use multidimensional/jagged arrays.";
 
         protected abstract GeneratedCodeRecognizer GeneratedCodeRecognizer { get; }
     }
--- a/analyzers/src/SonarAnalyzer.Common/Rules/PublicMethodWithMultidimensionalArrayBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/PublicMethodWithMultidimensionalArrayBase.cs
@@ -60,13 +60,15 @@ namespace SonarAnalyzer.Rules.Common
                 SyntaxKindsOfInterest.ToArray());
         }
 
-        private static bool MethodHasMultidimensionalArrayParameters(IMethodSymbol methodSymbol)
-        {
-            return methodSymbol.Parameters
-                .Select(param => param.Type as IArrayTypeSymbol)
-                .WhereNotNull()
-                .Any(type => type.Rank > 1 || type.ElementType is IArrayTypeSymbol);
-        }
+        private static bool MethodHasMultidimensionalArrayParameters(IMethodSymbol methodSymbol) =>
+            methodSymbol.Parameters.Any(param => param.Type is IArrayTypeSymbol arrayType
+                                                 && (arrayType.Rank > 1
+                                                     || IsJaggedArrayParam(param, arrayType)));
+
+        private static bool IsJaggedArrayParam(IParameterSymbol param, IArrayTypeSymbol arrayType) =>
+            param.IsParams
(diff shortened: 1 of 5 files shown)


csaba-sagi-sonarsource:
Do you agree to update the message of the rule from "Make this method private or simplify its parameters to not use multidimensional arrays." to "Make this method private or simplify its parameters to not use multidimensional or jagged arrays." ?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PublicMethodWithMultidimensionalArray.vb
```diff
@@ -38,6 +38,16 @@ Namespace Tests.Diagnostics
...
         End Sub
+
+        Public Sub Method7(ParamArray a As Integer()()) ' Compliant
+        End Sub
+
+        Public Sub Method8(ParamArray a As Integer()()()) ' Noncompliant
+        End Sub
+
+        Sub Method9(ParamArray a As Integer()(,)) ' Compliant
+        End Sub
```
andrei-epure-sonarsource, line 49: can you add a test with a parameterless declaration? (null param array)

```
    Sub Main
        Console.WriteLine("Hello, world!")
    End Sub
```

analyzers/src/SonarAnalyzer.Common/Rules/PublicMethodWithMultidimensionalArrayBase.cs
```diff
@@ -60,13 +60,15 @@ protected sealed override void Initialize(SonarAnalysisContext context)
                 SyntaxKindsOfInterest.ToArray());
         }
 
-        private static bool MethodHasMultidimensionalArrayParameters(IMethodSymbol methodSymbol)
-        {
-            return methodSymbol.Parameters
-                .Select(param => param.Type as IArrayTypeSymbol)
-                .WhereNotNull()
```
andrei-epure-sonarsource, line 67: why did you remove the null-check?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PublicMethodWithMultidimensionalArray.cs
```diff
@@ -36,6 +36,18 @@ public void MethodX(int[] a)
         public void Method5(int[,] a) //Compliant, implements interface
         {
         }
+
+        void Method6(params int[][,] a) // Compliant
+        {
+        }
+
+        public void Method7(params int[][] a) // Compliant
```
andrei-epure-sonarsource, line 44: can you also add with
```
public void Method7(int[][] a, params int[][] b) {}
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PublicMethodWithMultidimensionalArray.cs
```diff
@@ -36,6 +36,18 @@ public void MethodX(int[] a)
         public void Method5(int[,] a) //Compliant, implements interface
         {
         }
+
+        void Method6(params int[][,] a) // Compliant
```
andrei-epure-sonarsource, line 40: and also with `public void Method61(int[,][] a) {}`?

andrei-epure-sonarsource (changes requested):
I suggest to add more tests, otherwise looks very nice!

andrei-epure-sonarsource (approved):
LGTM

PR #4308, opened 2021-04-26
Fix S5332 FP: Don't raise http warning for xml namespace

### LINKED ISSUE #4124: Fix S5332 FP: Don't raise http warning for xml namespace
As seen on SonarCloud for S4MSB:
https://sonarcloud.io/project/security_hotspots?id=sonarscanner-msbuild&pullRequest=937&hotspots=AXgCSAIHik2AsAwb1Bpy

HTTP should not trigger the issue when assigned to field/variable/property/argument with "Namespace" in it's name:

```C#
[XmlRoot(ElementName = "SonarProjectConfig", Namespace = "http://www.sonarsource.com/msbuild/analyzer/2021/1")]
public class ProjectConfig
{
}
```

Same FP will happen in #4120 in our repo.

dmitry-golovinov-sonarsource:
Fixes #4124

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Text.RegularExpressions;
@@ -123,8 +124,7 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private void VisitStringExpressions(SyntaxNodeAnalysisContext c)
         {
-            var text = GetText(c.Node);
-            if (GetUnsafeProtocol(text, c.Node) is {} unsafeProtocol)
+            if (GetUnsafeProtocol(c.Node) is {} unsafeProtocol)
             {
                 c.ReportDiagnosticWhenActive(Diagnostic.Create(DefaultRule, c.Node.GetLocation(), unsafeProtocol, recommendedProtocols[unsafeProtocol]));
             }
@@ -134,8 +134,9 @@ namespace SonarAnalyzer.Rules.CSharp
             objectCreation.ArgumentList?.Arguments.Count > 0
             && validServerRegex.IsMatch(GetText(objectCreation.ArgumentList.Arguments[0].Expression));
 
-        private string GetUnsafeProtocol(string text, SyntaxNode node)
+        private string GetUnsafeProtocol(SyntaxNode node)
         {
+            var text = GetText(node);
             if (httpRegex.IsMatch(text) && !IsNamespace(node.Parent))
             {
                 return "http";
@@ -176,7 +177,7 @@ namespace SonarAnalyzer.Rules.CSharp
             };
 
         private static bool TokenContainsNamespace(SyntaxToken token) =>
-            token.Text.Contains("Namespace");
+            token.Text.IndexOf("Namespace", StringComparison.OrdinalIgnoreCase) != -1;
 
         private static Regex CompileRegex(string pattern, bool ignoreCase = true) =>
             new Regex(pattern, ignoreCase
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
@@ -124,7 +124,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private void VisitStringExpressions(SyntaxNodeAnalysisContext c)
         {
             var text = GetText(c.Node);
-            if (GetUnsafeProtocol(text) is {} unsafeProtocol)
+            if (GetUnsafeProtocol(text, c.Node) is {} unsafeProtocol)
             {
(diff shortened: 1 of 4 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
```diff
@@ -134,10 +134,32 @@ private void VisitStringExpressions(SyntaxNodeAnalysisContext c)
             objectCreation.ArgumentList?.Arguments.Count > 0
             && validServerRegex.IsMatch(GetText(objectCreation.ArgumentList.Arguments[0].Expression));
 
-        private string GetUnsafeProtocol(string text)
+        private string GetUnsafeProtocol(string text, SyntaxNode node)
         {
             if (httpRegex.IsMatch(text))
             {
+                switch (node.Parent)
```
andrei-epure-sonarsource, line 141: this looks like it could be a method "isNamespace", arrow-bodies with a switch expression

and then the `if` condition on line 139 would be `if (httpRegex.IsMatch(text) && !isNamespace())`

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
```diff
@@ -124,7 +124,7 @@ private static void VisitAssignments(SyntaxNodeAnalysisContext context)
         private void VisitStringExpressions(SyntaxNodeAnalysisContext c)
         {
             var text = GetText(c.Node);
-            if (GetUnsafeProtocol(text) is {} unsafeProtocol)
+            if (GetUnsafeProtocol(text, c.Node) is {} unsafeProtocol)
```
andrei-epure-sonarsource, line 127: now that you are passing `c.Node` to `GetUnsafeProtocol`, wouldn't it be better to call `GetUnsafeProtocol(c.Node)` and retrieve the text inside the method?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/ClearTextProtocolsAreSensitive.cs
```diff
@@ -158,6 +159,58 @@ public void Ftp()
...
+            var NamespaceLikeVar = "ftp://a:a@foo.com/";  // Noncompliant
+            NamespaceLikeArgument = "ftp://a:a@foo.com/";  // Noncompliant
+            NamespaceLikeField = "ftp://a:a@foo.com/";  // Noncompliant
+            NamespaceLikeProperty = "ftp://a:a@foo.com/";  // Noncompliant
+
+            var namespaceLikeVar = "http://www.sonarsource.com/msbuild/analyzer/2021/1";  // Noncompliant
+            string[] NamespaceArray = new string[1];
+            NamespaceArray[0] = "http://www.sonarsource.com/msbuild/analyzer/2021/1";  // Noncompliant
+
+            var NaaamespaceLikeVar = "http://www.sonarsource.com/msbuild/analyzer/2021/1";  // Noncompliant
```
andrei-epure-sonarsource, line 202: I find the "Naaamespace" a bit difficult to read in the test (I ask myself if it's intended or not, if it's a typo...) 

it could just be "foo" or just "namefoo" if you want to test that the full "namespace" string is checked

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/ClearTextProtocolsAreSensitive.cs
```diff
@@ -158,6 +159,58 @@ public void Ftp()
...
+        private string NaaamespaceLikeProperty { get; set; }
+
+        public void Foo(string NaaamespaceLikeArgument, string NamespaceLikeArgument = "ftp://a:a@foo.com/")  // Noncompliant
+        {
+            var NamespaceLikeVar = "ftp://a:a@foo.com/";  // Noncompliant
+            NamespaceLikeArgument = "ftp://a:a@foo.com/";  // Noncompliant
+            NamespaceLikeField = "ftp://a:a@foo.com/";  // Noncompliant
+            NamespaceLikeProperty = "ftp://a:a@foo.com/";  // Noncompliant
+
+            var namespaceLikeVar = "http://www.sonarsource.com/msbuild/analyzer/2021/1";  // Noncompliant
```
andrei-epure-sonarsource, line 198: why is this Noncompliant? shouldn't the check ignore casing? it's about making sure we don't raise on any sort of namespace

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/ClearTextProtocolsAreSensitive.cs
```diff
@@ -158,6 +159,58 @@ public void Ftp()
...
+
+    [XmlRoot(ElementName = "SonarProjectConfig", Namespace = "http://www.sonarsource.com/msbuild/analyzer/2021/1")]
+    public class NamespaceLikeAssignment
+    {
+        private string NamespaceLikeField = "http://www.sonarsource.com/msbuild/analyzer/2021/1";
+        private string NamespaceLikeProperty { get; set; }
+
+        public void Foo(string NamespaceLikeArgument)
+        {
+            var NamespaceLikeVar = "http://www.sonarsource.com/msbuild/analyzer/2021/1";
```
andrei-epure-sonarsource, line 171: could you also add tests with `fooNamespace` and `FOO_NAMESPACE`?

And add some test cases:

```
using System;
using System.Xml.Serialization;

[XmlRoot(Namespace=XML_NAMESPACE)]
public class C {
    public const string XML_NAMESPACE = "http://x";
}

public static class Constants {
    public const String NAMESPACE1 = "http://x";
}
```

andrei-epure-sonarsource (changes requested):
I left some comments

andrei-epure-sonarsource (approved):
LGTM

PR #4309, opened 2021-04-26
Fix S4275 FP: WPF DependencyProperty naming pattern (#3442)

### LINKED ISSUE #3442: S4275 FP with WPF DependencyProperty naming pattern

costin-zaharia-sonarsource:
Fixes #3442


analyzers/src/SonarAnalyzer.CSharp/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -119,6 +121,18 @@ protected override bool ShouldIgnoreAccessor(IMethodSymbol accessorMethod)
             || (property.GetMethod?.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax() is AccessorDeclarationSyntax getter
                 && getter.DescendantNodes().Any());
 
+        private static bool ContainsGetOrSetOnDependencyProperty(SyntaxNode syntaxNode, Compilation compilation)
+        {
+            var semanticModel = compilation.GetSemanticModel(syntaxNode.SyntaxTree);
```
andrei-epure-sonarsource: does this have a cost? could you first retrieve the `GetValue`/`SetValue` syntax nodes, and just after get, if the result is non-empty, get the semantic model for the compilation and do the check?

This should also separate the AST-only logic from the semantic logic. WDYT?

costin-zaharia-sonarsource: I was curios about this and I've run the profiler:
image

As far as I can tell `GetSemanticModel` is not time consuming at all. Almost all execution time for this function is spent while running `DerivesFrom` (18ms from 18ms).

I would prefer to let this as it is since the code is easy to read and maintain.

costin-zaharia-sonarsource: `IsMatch` looks like it could be improved but since it's just 1.33% from total rule execution time (18ms from 1346ms) I think that currently it doesn't worth the effort.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PropertiesAccessCorrectField.NetFramework.vb
```diff
@@ -0,0 +1,34 @@
...
+
+    Public Shared ReadOnly BackgroundColorProperty As DependencyProperty = DependencyProperty.Register("BackgroundColor", GetType(Color), GetType(Color), New PropertyMetadata(""))
+    Public Shared ReadOnly ForegroundColorProperty As DependencyProperty = DependencyProperty.Register("ForegroundColor", GetType(Color), GetType(Color), New PropertyMetadata(""))
+
+    Private backgroundColor_ As Color
+    Private foregroundColor_ As Color
+
+    Public Property BackgroundColor As Color
+        Get
+            Return GetValue(BackgroundColorProperty)
```
andrei-epure-sonarsource, line 16: could you add a test with GETVALUE, too? it's fine to not support it to keep things simple, but to document that VB is case insensitive

analyzers/src/SonarAnalyzer.VisualBasic/Rules/PropertiesAccessCorrectField.cs
```diff
@@ -160,9 +172,9 @@ private static ExpressionSyntax SingleInvocation(StatementSyntax body)
             if (semanticModel != null && argument.Parent is ArgumentListSyntax argList)
             {
                 var argumentIndex = argList.Arguments.IndexOf(argument);
-                if (semanticModel.GetSymbolInfo(argList.Parent).Symbol is IMethodSymbol methodSymbol &&
-                    argumentIndex < methodSymbol?.Parameters.Length &&
-                    methodSymbol?.Parameters[argumentIndex]?.RefKind != RefKind.None)
```
andrei-epure-sonarsource, line 165: are you confident on the code coverage of the VB test cases for this?

andrei-epure-sonarsource (approved):
LGTM, I suggest to document case insensitivity for VB

but can also be done in a subsequent PR

PR #4310, opened 2021-04-26
S2583 and S2589 False positive when a variable is initialized inside local function

### LINKED ISSUE #3161: S2583 and S2589 False positive when a variable is initialized inside local function
### Description

When a variable is initialized inside local function, S2583 is raised.

### Repro steps
```
public void Test()
{
        string buffer = null;
        Local();
        if (buffer != null)
        {
        }

        void Local()
        {
             buffer = new string('a', 1);
        }
}
```
### Expected behavior

No warning.

### Actual behavior

warning S2583: Change this condition so that it does not always evaluate to 'false'; some subsequent code is never executed.

### Related information

SonarAnalyzer.CSharp.8.4.0.15306
VisualStudio 2019 16.4.4

pavel-mikula-sonarsource:
Fixes #3161


csaba-sagi-sonarsource (approved):
LGTM!

PR #4311, opened 2021-04-26
Fix S1854 FP: When a variable is used inside local function

### LINKED ISSUE #3126: Fix S1854 FP: When a variable is used inside local function

### LINKED ISSUE #3719: Fix S1854 FP: Variable used in local function as delegate

pavel-mikula-sonarsource:
Fixes #3126
Fixes #3719

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Helpers/MutedSyntaxWalker.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Helpers/MutedSyntaxWalker.cs
@@ -35,20 +35,20 @@ namespace SonarAnalyzer.Helpers
         // All kinds that SonarAnalysisContextExtensions.RegisterExplodedGraphBasedAnalysis registers for
         private static readonly SyntaxKind[] RootKinds = new[]
         {
-                SyntaxKind.ConstructorDeclaration,
-                SyntaxKind.DestructorDeclaration,
-                SyntaxKind.ConversionOperatorDeclaration,
-                SyntaxKind.OperatorDeclaration,
-                SyntaxKind.MethodDeclaration,
-                SyntaxKind.PropertyDeclaration,
-                SyntaxKind.GetAccessorDeclaration,
-                SyntaxKind.SetAccessorDeclaration,
-                SyntaxKind.AddAccessorDeclaration,
-                SyntaxKind.RemoveAccessorDeclaration,
-                SyntaxKind.AnonymousMethodExpression,
-                SyntaxKind.SimpleLambdaExpression,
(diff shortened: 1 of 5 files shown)


pavel-mikula-sonarsource:
I don't want to fix the old coverage issue in this case.

analyzers/src/SonarAnalyzer.CSharp/Helpers/MutedSyntaxWalker.cs
```diff
@@ -0,0 +1,102 @@
...
+                SyntaxKind.MethodDeclaration,
+                SyntaxKind.PropertyDeclaration,
+                SyntaxKind.GetAccessorDeclaration,
+                SyntaxKind.SetAccessorDeclaration,
+                SyntaxKind.AddAccessorDeclaration,
+                SyntaxKind.RemoveAccessorDeclaration,
+                SyntaxKind.AnonymousMethodExpression,
+                SyntaxKind.SimpleLambdaExpression,
+                SyntaxKind.ParenthesizedLambdaExpression
+            };
```
csaba-sagi-sonarsource, line 51: NITPICK: indentation does not align with the opening curly bracket.

csaba-sagi-sonarsource:
I know we said we will not cover formatting stuff now, but please fix `IsAllowedNumericInitialization` in DeadStores.cs the method is  unreadable.

csaba-sagi-sonarsource (approved):
LGTM

PR #4314, opened 2021-04-26
Fix S3927 FP: should not treat public methods having EditorBrowsable(EditorBrowsableState.Never) as bug

### LINKED ISSUE #2689: S3927 should not treat public methods as bug
### Description

I have a few false positive bugs with sonaranalyzer S3927.
See https://sonarcloud.io/project/issues?id=OpenRIAServices_OpenRiaServices&resolutions=FALSE-POSITIVE

The rule requires methods to be private, but that is not a requirement from the framework.
See source at https://referencesource.microsoft.com/#mscorlib/system/runtime/serialization/serializationeventscache.cs,41

On the contrary in medium trust scenarios (not that common anymore) and some older implementations the methods cannot be private but must be public (or internal with internal access granted to different framework assemblies). However in that scenario it is recommended to hide it using editorbrowsable attribute just as in the code refered to below.

If you really want to warn against public methods you should change it to a code smell.

Another side issue (not important to me personally) is that **virtual is not allowed for the methods**, which would probably be better to raise bugs for.

I understand if this is not a top priority for you, but I want to record the issue so that you are aware of it. Also adding a check for virtual might be nice to have and be very easy for your developers to do.

### Repro steps

See code at 
https://sonarcloud.io/project/issues?id=OpenRIAServices_OpenRiaServices&open=AWf-_5guXONnvcWDclJp&resolutions=FALSE-POSITIVE

repository is availible at github OpenRiaServices/OpenRiaServices

### Expected behavior

Please provide a description of the behavior you expect.

Having public serialization methods with `[EditorBrowsable(EditorBrowsableState.Never)]` should not be a bug. It is a with acceptable with a code smell however.

Or at least update documentation to say that it is bad practice but not a bug to have it public.

### Actual behavior

Please provide a description of the actual behavior you observe.

The usage is reported as a bug.

### Known workarounds

Please provide a description of any known workarounds.

*

### Related information

Last sonarcloud and azure devops sonarcloud plugin as per 2019-09-28

dmitry-golovinov-sonarsource:
Fixes #2689

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementSerializationMethodsCorrectly.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementSerializationMethodsCorrectly.cs
@@ -48,5 +48,12 @@ namespace SonarAnalyzer.Rules.CSharp
                 .FirstOrDefault()
                 ?.Identifier
                 .GetLocation();
+
+        protected override bool ContainsMatchingAttributes(IMethodSymbol methodSymbol) =>
+            base.ContainsMatchingAttributes(methodSymbol) && !HiddenByEditorBrowsableAttribute(methodSymbol);
+
+        private static bool HiddenByEditorBrowsableAttribute(IMethodSymbol methodSymbol) =>
+            methodSymbol.GetAttributes(KnownType.System_ComponentModel_EditorBrowsableAttribute)
+                .Any(x => x.ConstructorArguments.Any(a => a.Type.Is(KnownType.System_ComponentModel_EditorBrowsableState) && 1.CompareTo(a.Value) == 0));
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementSerializationMethodsCorrectly.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementSerializationMethodsCorrectly.cs
@@ -54,6 +54,6 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static bool HiddenByEditorBrowsableAttribute(IMethodSymbol methodSymbol) =>
             methodSymbol.GetAttributes(KnownType.System_ComponentModel_EditorBrowsableAttribute)
-                .Any(x => x.ConstructorArguments.Any(a => a.Type.Is(KnownType.System_ComponentModel_EditorBrowsableState) && 1.CompareTo(a.Value) == 0));
+                .Any(x => x.ConstructorArguments.Any(a => 1.CompareTo(a.Value) == 0));
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementSerializationMethodsCorrectly.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementSerializationMethodsCorrectly.cs
@@ -48,12 +48,5 @@ namespace SonarAnalyzer.Rules.CSharp
                 .FirstOrDefault()
                 ?.Identifier
                 .GetLocation();
-
-        protected override bool ContainsMatchingAttributes(IMethodSymbol methodSymbol) =>
-            base.ContainsMatchingAttributes(methodSymbol) && !HiddenByEditorBrowsableAttribute(methodSymbol);
-
-        private static bool HiddenByEditorBrowsableAttribute(IMethodSymbol methodSymbol) =>
-            methodSymbol.GetAttributes(KnownType.System_ComponentModel_EditorBrowsableAttribute)
-                .Any(x => x.ConstructorArguments.Any(a => 1.CompareTo(a.Value) == 0));
     }
 }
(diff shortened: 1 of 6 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ImplementSerializationMethodsCorrectly.CSharp9.cs
```diff
@@ -55,6 +56,38 @@ public record Foo
...
+    public void OnDeserializing3(StreamingContext context) { }          // Compliant
+
+    [OnSerializing]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public void OnSerializing3(StreamingContext context) { }            // Compliant
+
+    [OnSerialized]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    int OnSerialized3(StreamingContext context) { return 1; }           // Compliant
+
```
costin-zaharia-sonarsource, line 90: Do we need to duplicate the cases from `ImplementSerializationMethodsCorrectly.cs` aslo here?

analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementSerializationMethodsCorrectly.cs
```diff
@@ -48,5 +48,12 @@ public sealed class ImplementSerializationMethodsCorrectly : ImplementSerializat
                 .FirstOrDefault()
                 ?.Identifier
                 .GetLocation();
+
+        protected override bool ContainsMatchingAttributes(IMethodSymbol methodSymbol) =>
+            base.ContainsMatchingAttributes(methodSymbol) && !HiddenByEditorBrowsableAttribute(methodSymbol);
+
+        private static bool HiddenByEditorBrowsableAttribute(IMethodSymbol methodSymbol) =>
+            methodSymbol.GetAttributes(KnownType.System_ComponentModel_EditorBrowsableAttribute)
+                .Any(x => x.ConstructorArguments.Any(a => 1.CompareTo(a.Value) == 0));
```
costin-zaharia-sonarsource, line 57: Can't we check the editor attributes in the base class to have this issue fixed also for VB?

costin-zaharia-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementSerializationMethodsCorrectly.cs
dmitry-golovinov-sonarsource: You are right, initially I was not able to find that it is also applicable to VB, that's why I've moved the implementation to CS part. Will fix it!

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ImplementSerializationMethodsCorrectly.CSharp9.cs
dmitry-golovinov-sonarsource: So basically in CSharp9 we want to test only C#9 specific syntax only, right?

analyzers/src/SonarAnalyzer.Common/Helpers/KnownType.cs
```diff
@@ -197,6 +197,8 @@ internal sealed class KnownType
         internal static readonly KnownType System_Collections_Specialized_NameObjectCollectionBase = new KnownType("System.Collections.Specialized.NameObjectCollectionBase");
         internal static readonly KnownType System_Collections_Specialized_NameValueCollection = new KnownType("System.Collections.Specialized.NameValueCollection");
         internal static readonly KnownType System_ComponentModel_DefaultValueAttribute = new KnownType("System.ComponentModel.DefaultValueAttribute");
+        internal static readonly KnownType System_ComponentModel_EditorBrowsableAttribute = new KnownType("System.ComponentModel.EditorBrowsableAttribute");
+        internal static readonly KnownType System_ComponentModel_EditorBrowsableState = new KnownType("System.ComponentModel.EditorBrowsableState");
```
pavel-mikula-sonarsource, line 201: This one isn't used
```suggestion
```

analyzers/src/SonarAnalyzer.Common/Rules/ImplementSerializationMethodsCorrectlyBase.cs
```diff
@@ -71,6 +71,10 @@ public abstract class ImplementSerializationMethodsCorrectlyBase : SonarDiagnost
                 },
                 SymbolKind.Method);
 
+        private static bool HiddenByEditorBrowsableAttribute(IMethodSymbol methodSymbol) =>
+            methodSymbol.GetAttributes(KnownType.System_ComponentModel_EditorBrowsableAttribute)
+                .Any(x => x.ConstructorArguments.Any(a => 1.CompareTo(a.Value) == 0));
```
pavel-mikula-sonarsource, line 76: This should be dejavalized to remove the indirection `1.CompareTo(a.Value) == 0`

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ImplementSerializationMethodsCorrectly.CSharp9.cs
pavel-mikula-sonarsource: Yes

pavel-mikula-sonarsource (approved):
LGTM

costin-zaharia-sonarsource (approved):
LGTM!

PR #4316, opened 2021-04-26
Fix S1854, S2583 and S2589 FPs related to Try/Catch

### LINKED ISSUE #2590: Fix S2583/S2589 FP: On null check if instance is used in try block
### Description

A S2583 is detected when checking for null when a variable is assinged using a function and the object is used inside a try block. Similarly a S2589 is detected when checking for not null when a variable is assinged using a function and the object is used inside a try block

### Repro steps

```
static class Program
{
    static void Main()
    {
        var foo = GetFoo(Guid.Empty);

        try
        {
            foo.Write();
        }
        catch
        {
            // Do nothing
        }

        if (foo == null) // S2583
        {
            Console.WriteLine("Foo is null");
        }

        if (foo != null) // 2589
        {
            Console.WriteLine("Foo is not null");
        }
    }

    static Foo GetFoo(Guid id)
    {
        return id != Guid.Empty ? new Foo(id) : null;
    }

    class Foo
    {
        private Guid Id { get; }

        public Foo(Guid id)
        {
            Id = id;
        }

        public void Write()
        {
            Console.WriteLine(Id);
        }
    }
}
```

### Expected behavior

No warnings

### Actual behavior

A S2583 warning on `if (foo == null)` and a 2589 warning on `if (foo != null)`

### Related information

* SonarC# Version 7.16.0.8981
* Visual Studio Version 16.1.5
* .NET Core 2.2

### LINKED ISSUE #2596: Fix S1854 FP: Try/Catch statement with constant value propagation
In the following code, SonarAnalyzer.CSharp 7.16.0.8981 flags `S1854: Remove this useless assignment to local variable 'length'.` Is this a false positive?

```
using System;
using System.IO;

namespace S1854_False_Positive
{
    public static class C
    {
        public static long GetFileLength(string path)
        {
            const int unknownfilelength = -1;
            long length = unknownfilelength;
            try
            {
                length = new FileInfo(path).Length;
            }
            catch (Exception e)
            {
                Console.WriteLine(e);
            }
            return length;
        }
    }
}
```

### LINKED ISSUE #2598: Fix S1854 FP: Try/Catch with cast `(string) null`
### Description

if you initialize string `s = null` and `var s = default(string);` [S1854 dead stores](https://rules.sonarsource.com/csharp/RSPEC-1854) it won't raise. 

There is a bug here because we don't support the cast (string) null

### Repro steps

```

        public static string WithCastedNull(string path)
        {
            string length = (string)null; // Noncompliant FP
            try
            {
                length = new System.IO.FileInfo(path).Length.ToString();
            }
            catch (Exception e)
            {
                Console.WriteLine(e);
            }
            return length;
        }

```

### Expected behavior

No issue, just as with 

```
            string length = default(string);
            string length = null;
```

### Actual behavior

Raises FP

### Related information

* SonarC# Version 7.16

Initially reported by  @antoinebj  as a comment to #2596

### LINKED ISSUE #2600: Fix S2589 / S1854 FP: There is only one block for the `try` body in the CFG
### Description

See below. It's a tricky one because, to fix this, we should make a block for every statement inside the try block that can either continue or throw.

### Repro steps

```
        public int Start()
        {
            const int x = -1;
            int exitCode = x; // Noncompliant FP - if Archive throws, it will be returned
            try
            {
                Archive(); // can throw here

                exitCode = 1;
            }
            catch (SystemException e)
            {

            }

            return exitCode;
        }
```

### Expected behavior

No issue

### Related information

* SonarC# Version 7.16

This was reported on the [community forum](https://community.sonarsource.com/t/s1854-false-positive/13263)

### LINKED ISSUE #2761: S2583 FP in try/catch
### Description

The S2583 warning sometimes occurs as a false positive (note, a lot of other fixes since 4.12 has made it in, thankyou! so this is new since 4.12).

### Repro steps

Consider the following code block:

```csharp
public static void CreateDirectory(string directory)
{
    string fullPath = GetFullPath(directory, WorkDirectory);

    if (Directory.Exists(fullPath)) return;

    bool created = false;
    int attempts = CopyWaitAttempts;

    DeleteFile(fullPath);
    do {
        try {
            Directory.CreateDirectory(fullPath);
            created = true;
        } catch (AccessViolationException) {
            if (attempts == 0) throw;
        } catch (UnauthorizedAccessException) {
            // On windows occurs if the file is already open.
            if (attempts == 0) throw;
        }

        if (!created) {
            // If the copy failed, it's because it's probably already open somewhere else. So we
            // wait 250ms and try again.
            --attempts;
            Thread.Sleep(CopyWaitInterval);
        }
    } while (!created);
}
```

It complains that the variable `created` always evaluates to `false`. This is not the case, as we see that in the try/catch block it is set to true if successful, and only on some specific exceptions with some specific conditions, the catch blocks fall through and the variable remains `false`. So come to the `if (!created)` line, the variable `created` is true if the directory was created, or false if there was an exception.

### Expected behavior

S2583 is able to analyze that an exception can occur anywhere within the try section, fall through a catch section, and so a variable may change, or it may not change.

### Actual behavior

See the description, false warning.

### Known workarounds

None.

### Related information

* SonarC# Version 4.13.0.1687
* Visual Studio Version 2019 16.3.6

### LINKED ISSUE #2766: S1854 FP: Variable assignment in catch block is not recognized
### Description

Variable assignment in `catch` should be allowed.

### Repro steps

https://github.com/sm-g/AnalyzerIssues/blob/master/AnalyzerIssues/S1854.cs

Notice, that there is no false positive when `try` without `return`.

### Related information

* SonarC# Version 8.0.0.9566
* Visual Studio Version 16.3.6

pavel-mikula-sonarsource:
Fixes #2590
Fixes #2596
Fixes #2598
Fixes #2600
Fixes #2761
Fixes #2766

First squash commit doesn't need a review

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Helpers/MutedSyntaxWalker.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Helpers/MutedSyntaxWalker.cs
@@ -54,6 +54,8 @@ namespace SonarAnalyzer.Helpers
         private readonly SyntaxNode root;
         private readonly ISymbol[] symbols;
         private bool isMuted;
+        private bool isInTryOrCatch;
+        private bool isOutsideTryCatch;
 
         public MutedSyntaxWalker(SemanticModel semanticModel, SyntaxNode node)
             : this(semanticModel, node, node.DescendantNodesAndSelf().OfType<IdentifierNameSyntax>().Select(x => semanticModel.GetSymbolInfo(x).Symbol).WhereNotNull().ToArray()) { }
@@ -71,7 +73,7 @@ namespace SonarAnalyzer.Helpers
             {
                 Visit(root);
             }
-            return isMuted;
+            return isMuted || (isInTryOrCatch && isOutsideTryCatch);
         }
 
         public override void Visit(SyntaxNode node)
@@ -87,6 +89,15 @@ namespace SonarAnalyzer.Helpers
             if (symbols.FirstOrDefault(x => node.NameIs(x.Name) && x.Equals(semanticModel.GetSymbolInfo(node).Symbol)) is { } symbol)
             {
                 isMuted = IsInTupleAssignmentTarget() || IsUsedInLocalFunction(symbol) || IsInUnsupportedExpression();
+                if (HasAncestor(SyntaxKind.TryStatement))
+                {
+                    // We're only interested in "try" and "catch" blocks. Don't count "finally" block
+                    isInTryOrCatch = isInTryOrCatch || !HasAncestor(SyntaxKind.FinallyClause);
+                }
+                else
+                {
+                    isOutsideTryCatch = true;
+                }
             }
             base.VisitIdentifierName(node);
 
@@ -96,10 +107,13 @@ namespace SonarAnalyzer.Helpers
             bool IsUsedInLocalFunction(ISymbol symbol) =>
                 // We don't mute it if it's declared and used in local function
                 !(symbol.ContainingSymbol is IMethodSymbol containingSymbol && containingSymbol.MethodKind == MethodKindEx.LocalFunction)
-                && node.FirstAncestorOrSelf<SyntaxNode>(x => x.IsKind(SyntaxKindEx.LocalFunctionStatement)) != null;
+                && HasAncestor(SyntaxKindEx.LocalFunctionStatement);
 
             bool IsInUnsupportedExpression() =>
                 node.FirstAncestorOrSelf<SyntaxNode>(x => x.IsAnyKind(SyntaxKindEx.IndexExpression, SyntaxKindEx.RangeExpression)) != null;
+
+            bool HasAncestor(SyntaxKind kind) =>
+                node.FirstAncestorOrSelf<SyntaxNode>(x => x.IsKind(kind)) != null;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


analyzers/src/SonarAnalyzer.CSharp/Helpers/MutedSyntaxWalker.cs
```diff
@@ -0,0 +1,130 @@
...
+            root = node.Ancestors().FirstOrDefault(x => x.IsAnyKind(RootKinds));
+        }
+
+        public bool IsMuted()
+        {
+            if (symbols.Any() && root != null)
+            {
+                Visit(root);
+            }
+            return isMuted || (isInTryOrCatch && isOutsideTryCatch);
```
csaba-sagi-sonarsource, line 76: It would be nice to add a comment explaining why we look for both in and outside of trycatch.

csaba-sagi-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.CSharp/Helpers/MutedSyntaxWalker.cs
pavel-mikula-sonarsource: It's quite clear from the variable names IMO

Issue #2429, opened 2019-05-21
Utility analyzers execution time is too long (performance problem)


duncanp-sonar:
### Description

See this community thread:
https://community.sonarsource.com/t/help-figuring-out-net-project-scan-taking-significantly-longer/9202/26

To summarise, the `TokenTypeAnalyzer` and `SymbolReferenceAnalyzer` took nearly eight minutes each to run against a specific test project of around 10k lines of code:
```
448.313 50 SonarAnalyzer.Rules.CSharp.TokenTypeAnalyzer
445.228 49 SonarAnalyzer.Rules.CSharp.SymbolReferenceAnalyzer
```

The time taken is excessive for what isn't a large number of lines of code.

Note: the project being analyzed was a test project so only a subset of the utility analyzers are being run. It's possible that the same issue exists with all of the utility analyzers.

### Known workarounds
Exclude the project from analysis by adding the `SonarQubeExclude` property:

```
  <PropertyGroup>
    <SonarQubeExclude>true</SonarQubeExclude> 
  </PropertyGroup>
```

### Related information

* SonarC# Version: v7.13

andrei-epure-sonarsource:
Might help: https://benchmarkdotnet.org/

andrei-epure-sonarsource:
another [community ticket](https://community.sonarsource.com/t/sonarscanner-msbuild-sonar-showprofiling-not-working/15582/3)

andrei-epure-sonarsource:
[This community ticket](https://community.sonarsource.com/t/net-analysis-time-during-msbuild-increased-from-15-min-to-40-min/32898/8) has the same problem

```
33%      SonarAnalyzer.Rules.CSharp.TokenTypeAnalyzer
22%      SonarAnalyzer.Rules.CSharp.SymbolReferenceAnalyzer
18%      SonarAnalyzer.Rules.CSharp.MetricsAnalyzer
12%      SonarAnalyzer.Rules.CSharp.CookieShouldBeHttpOnly
8%      SonarAnalyzer.Rules.CSharp.CopyPasteTokenAnalyzer
````

andrei-epure-sonarsource:
This has been raised in another community topic ([public thread](https://community.sonarsource.com/t/azure-dev-ops-c-slow-build-times/33007) , [private thread](https://community.sonarsource.com/t/re-azure-dev-ops-c-slow-build-times/33108)).

The user sent us a reproducer which we can use to find the perf hotspot in our code.

[...]

PR #4318, opened 2021-04-27
Avoid generating metrics for files with a large number of identifier tokens

### LINKED ISSUE #4273: Skip Utility Analyzers for complex files
Related to  #2429. Especially [this comment ](https://github.com/SonarSource/sonar-dotnet/issues/2429#issuecomment-808374949) and [this one](https://github.com/SonarSource/sonar-dotnet/issues/2429#issuecomment-816534924).

Repro: https://github.com/costin-zaharia-sonarsource/repro-2429

## TL;DR

We should skip generating metadata for files which lead to very expensive semantic model calls. We already  have the [IsToComplex()](https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/src/SonarAnalyzer.CFG/ControlFlowGraph/CSharpControlFlowGraphBuilder.cs#L500) method, and we don't generate CFGs for complex methods. 

We should do the same for Utility Analyzers.

## Longer story

> This has been raised in another community topic ([public thread](https://community.sonarsource.com/t/azure-dev-ops-c-slow-build-times/33007), and a private thread as well).
>
> The user sent us a reproducer which we can use to find the perf hotspot in our code.
>
> The file has 5800 LOC which instantiate a huge dictionary, and our analysis is put to its knees due to a single dictionary instantiation.
>
> This is likely due to us using the Semantic Model for generating the token information - we should find a way to cache this for single files, or just stop when we detect too nested syntaxes, just like we've done in [SONARSEC-1603](https://jira.sonarsource.com/browse/SONARSEC-1603).

```
  public class FooCollection<K, V> : Dictionary<K, KeyValuePair<int, FooCollection<K, V>>>
  {
    // stuff
  };

public static class Bar
  {
    public static readonly ReadOnlyDictionary<string, KeyValuePair<int, FooCollection<string, string>>> Foo = new ReadOnlyDictionary<string, KeyValuePair<int, FooCollection<string, string>>>(new FooCollection<string, string>()
    { 
   // lots and lots of nested initializations 
    }; // end of 5800 LOC
```

costin-zaharia-sonarsource:
Affected utility analyzers:
- `TokenTypeAnalyzer`
- `SymbolReferenceAnalyzer`

Fixes #4273


analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -92,34 +93,39 @@ public abstract class UtilityAnalyzerBase<TMessage> : UtilityAnalyzerBase
...
                     }
                 });
 
-        private bool ShouldGenerateMetrics(SyntaxTree tree) =>
+        protected virtual bool ShouldGenerateMetrics(SyntaxTree tree) =>
             FileExtensionWhitelist.Contains(Path.GetExtension(tree.FilePath))
              && (AnalyzeGeneratedCode || !GeneratedCodeRecognizer.IsGenerated(tree));
 
+        protected bool HasTooManyIdentifiers(SyntaxTree syntaxTree, int syntaxKindRaw) =>
+            syntaxTree.GetRoot().DescendantTokens().Count(token => token.RawKind == syntaxKindRaw) > TokenCountThreshold;
```
pavel-mikula-sonarsource, line 127: We had an performance issue with `.DescendantNodes()` for large files in `sonar-security` SONARSEC-1603.

If that proves problematic, we can use `CSharpSyntaxWalker` to walk it, count and stop after it's too big.

pavel-mikula-sonarsource: It's related to 8,3 MB file `Models\GlassMapperItemClasses.cs` from the reproducer

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -76,6 +76,7 @@ protected void ReadParameters(SonarAnalysisContext context, CompilationAnalysisC
     public abstract class UtilityAnalyzerBase<TMessage> : UtilityAnalyzerBase
         where TMessage : IMessage, new()
     {
+        private const int TokenCountThreshold = 7_000;
```
costin-zaharia-sonarsource, line 79: I've done a few tests on my PC to see how long it takes to process large number of tokens:

<table>
<tr>
	<td>15000
	<td>~30sec
<tr>
	<td>8500
	<td>~12sec
<tr>
	<td>7000
	<td>~8sec
<tr>
	<td>5500
	<td>~5.5sec
</table>

pavel-mikula-sonarsource: It's not a token count, but identifier count, isn't it?

Does it make sense to highlight tokens if there's more then 1000 of them in a single file from user experience point of view?

costin-zaharia-sonarsource: We count only tokens. The code should be generic enough to cover `SymbolReferenceAnalyzer` too. Which retrieves symbols for all kind of node.

costin-zaharia-sonarsource: > Does it make sense to highlight tokens if there's more then 1000 of them in a single file from user experience point of view?

Probably on large classes it would be more helpful since the code is more complex and hard to read. On the long term I think it makes sense to implement the suggestions from https://github.com/SonarSource/sonar-dotnet/issues/2429#issuecomment-816641431 to be able to increase this limit.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -76,9 +76,11 @@ protected void ReadParameters(SonarAnalysisContext context, CompilationAnalysisC
     public abstract class UtilityAnalyzerBase<TMessage> : UtilityAnalyzerBase
         where TMessage : IMessage, new()
     {
+        private const int TokenCountThreshold = 10_000;
```
pavel-mikula-sonarsource, line 79: At some point, the implementation was counting only identifiers. Does the measurement comment in this PR reflects `identifiers` or `all tokens`?

Because 7 000 identifiers could be something like 35 000 tokens...?

pavel-mikula-sonarsource, line 79: We should have some UT or IT for this. This trick can help with that:
https://github.com/SonarSource/sonar-dotnet/blob/f6feccca65fb24d565fd3a9a614cc66b69125fdc/analyzers/tests/SonarAnalyzer.UnitTest/ControlFlowGraph/ControlFlowGraphTest.cs#L5240

pavel-mikula-sonarsource: For example, `analyzers\tests\SonarAnalyzer.UnitTest\ControlFlowGraph\ControlFlowGraphTest.cs` has 34 000 tokens (12 000 identifiers)

I think we should increate the number to something like 40 000-50 000 tokens.

pavel-mikula-sonarsource (changes requested):
I have some clarification questions.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
costin-zaharia-sonarsource: > At some point, the implementation was counting only identifiers. Does the measurement comment in this PR reflects identifiers or all tokens?

As mentioned in comment the numbers are for tokens.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs
```diff
@@ -39,6 +39,8 @@ public abstract class TokenTypeAnalyzerBase : UtilityAnalyzerBase<TokenTypeInfo>
 
         protected abstract TokenClassifierBase GetTokenClassifier(SyntaxToken token, SemanticModel semanticModel);
 
+        protected override bool SkipAnalysisForLargeFiles => true;
```
pavel-mikula-sonarsource, line 42: This will completely disable the highlighting for the file. Even for known keywords that are cheap to detect.

We should just avoid `ClassifyIdentifier()` when the file is known to be too large.

pavel-mikula-sonarsource (changes requested):
One more thing

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs
costin-zaharia-sonarsource: Indeed this would be nice but I did not want to have a different behavior per analyzer type.

pavel-mikula-sonarsource: This will introduce significant decrease of UX for LTS users for next 1,5 years :( The files without highlighting are way harder to read

andrei-epure-sonarsource: one thing to consider: are we introducing a performance regression for all codebases just to avoid the edge case of huge files?

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
andrei-epure-sonarsource: one thing to consider: are we introducing a performance regression for all codebases just to avoid the edge case of huge files? or is the overall change negligible for normal codebases?

andrei-epure-sonarsource:
I tend to think that, given this will possibly impact UX, maybe it's better not to ship this with LTS? And first deploy to SC, see if we get feedback, and then ship with the next SQ version, to have more confidence?

I think it will be hard to validate the change in UX ... not very sure...

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
andrei-epure-sonarsource: actually, never mind that : on some codebases anyway this is already taking 20-30% of analysis time. So what we'll do here should decrease it.

andrei-epure-sonarsource:
On the other hand, those projects with large files cannot be analyzed without this. And we cannot exclude a single file from analysis. 

So this is quite necessary.

andrei-epure-sonarsource (dismissed):
Lgtm

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs
```diff
@@ -58,10 +67,17 @@ protected sealed override TokenTypeInfo CreateMessage(SyntaxTree syntaxTree, Sem
             return tokenTypeInfo;
         }
 
+        protected override void Initialize(SyntaxTree tree) =>
+            SkipIdentifierTokens = HasTooManyIdentifierTokens(tree, identifierTokenKind);
+
+        private static bool HasTooManyIdentifierTokens(SyntaxTree syntaxTree, int tokenKind) =>
+            syntaxTree.GetRoot().DescendantTokens().Count(token => token.RawKind == tokenKind) > IdentifierTokenCountThreshold;
```
pavel-mikula-sonarsource, line 74:
```suggestion
```
We can avoid all of this and this additional `DescendantTokens()` with better state propagation - see `CreateMessage` above.
It already gathers all tokens. We can do this
```
var skipIdentifierTokens = tokens.Count(token => token.RawKind == tokenKind) > IdentifierTokenCountThreshold
foreach(...
    spans.AddRange(GetTokenClassifier(token, semanticModel, skipIdentifierTokens).Spans);
```
Like that, we can remove `Initialize`, `HasTooManyIdentifierTokens` and `SkipIdentifierTokens`

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs
```diff
@@ -30,13 +30,22 @@ namespace SonarAnalyzer.Rules
 {
     public abstract class TokenTypeAnalyzerBase : UtilityAnalyzerBase<TokenTypeInfo>
     {
-        protected const string DiagnosticId = "S9999-token-type";
+        protected bool SkipIdentifierTokens;
```
pavel-mikula-sonarsource, line 33: This should be after constants, but it can be removed with the other comment

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -92,31 +92,35 @@ public abstract class UtilityAnalyzerBase<TMessage> : UtilityAnalyzerBase
                         return;
                     }
 
-                    // The results of Metrics and CopyPasteToken analyzers are not needed for Test projects yet the plugin side expects the protobuf files, so we create empty ones.
                     if (IsTestProject && SkipAnalysisForTestProject)
                     {
-                        EnsureDirectoryExistsAndCreateFile().Dispose();
+                        // The results of Metrics and CopyPasteToken analyzers are not needed for Test projects yet the plugin side expects the protobuf files, so we create empty ones.
```
pavel-mikula-sonarsource, line 97: With this new structure and `ShouldGenerateMetrics` logic, we can remove this duplication and embed the `IsTestProject && SkipAnalysisForTestProject` condition in `ShouldGenerateMetrics`

pavel-mikula-sonarsource: After that, the recently created `EnsureDirectoryExistsAndCreateFile` can be embedded directly in this method again

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/SymbolReferenceAnalyzerBase.cs
```diff
@@ -148,7 +151,10 @@ private SymRefInfo GetSymRefInfo(SyntaxToken token, SemanticModel semanticModel)
             return null;
         }
 
-        public class SymRefInfo
+        private static bool HasTooManyTokens(SyntaxTree syntaxTree) =>
+            syntaxTree.GetRoot().DescendantTokens().Count() > TokenCountThreshold;
```
pavel-mikula-sonarsource, line 155: `CreateMessage` above already does `DescendantTokens`. We can avoid 2nd evaluation if we put the condition there and don't run the foreach...
```
var tokens = syntaxTree.GetRoot().DescendantTokens();
if(tokens.Length <= TokenCountThreshold)
{
  foreach...

}
return symbolReferenceInfo;
```

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/SymbolReferenceAnalyzerBase.cs
```diff
@@ -78,12 +79,14 @@ protected sealed override SymbolReferenceInfo CreateMessage(SyntaxTree syntaxTre
             return symbolReferenceInfo;
         }
 
+        protected override bool ShouldGenerateMetrics(SyntaxTree tree) =>
```
pavel-mikula-sonarsource, line 82: This can be reverted and made `private` in base class after addressing the comment about `HasTooManyTokens`

pavel-mikula-sonarsource (changes requested):
We can simplify the state and avoid additional `DescendantTokens()` calls.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/SymbolReferenceAnalyzerBase.cs
andrei-epure-sonarsource: good catch! from this POV, we should do perf testing with this and before this before merging on `master`. To see the effect on "normal" projects.

costin-zaharia-sonarsource: FYI: I've already did some tests to check the performance for `DescendantTokens` and it seems to be relatively cheap. For a tree with 334009 tokens it takes ~00:00:00.0591375

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs
costin-zaharia-sonarsource: FYI: We need to iterate `DescendantTokens()` two times: once to know how many tokens there are and a second pass to create the spans.

pavel-mikula-sonarsource: Yes, and `DescendantTokens` builds the list from walking the entire tree, while the identifier count would be iteration over flat array.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/SymbolReferenceAnalyzerBase.cs
andrei-epure-sonarsource: Thanks.

costin-zaharia-sonarsource: I prefer not to update this method. If we do the check here then we will create empty `SymbolReferenceInfo` which we will not need (or return null which is not great).

Doing the check in `ShouldGenerateMetrics` is better in my opinion since it will prevent the `CreateMessage` calls if they are not needed.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs
```diff
@@ -30,30 +30,42 @@ namespace SonarAnalyzer.Rules
...
         private const string TokenTypeFileName = "token-type.pb";
+        private const int IdentifierTokenCountThreshold = 1_000;
+        private readonly int identifierTokenKind;
 
         private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetUtilityDescriptor(DiagnosticId, Title);
+
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
-        protected abstract TokenClassifierBase GetTokenClassifier(SyntaxToken token, SemanticModel semanticModel);
+        protected TokenTypeAnalyzerBase(int identifierTokenKind)
```
pavel-mikula-sonarsource, line 43: _Optional nitpick:_ Could be arrow method

pavel-mikula-sonarsource:
Changes looks good, let's add UTs to verify them and we'll be good to go

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs
```diff
@@ -30,30 +30,39 @@ namespace SonarAnalyzer.Rules
 {
     public abstract class TokenTypeAnalyzerBase : UtilityAnalyzerBase<TokenTypeInfo>
     {
-        protected const string DiagnosticId = "S9999-token-type";
+        private const string DiagnosticId = "S9999-token-type";
         private const string Title = "Token type calculator";
         private const string TokenTypeFileName = "token-type.pb";
+        private const int IdentifierTokenCountThreshold = 4_000;
```
costin-zaharia-sonarsource, line 36: It takes around 3min 21sec on my PC to generate the messages for 4k identifier tokens

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/SymbolReferenceAnalyzerTest.cs
```diff
@@ -139,6 +139,12 @@ public class SymbolReferenceAnalyzerTest
             // This path is unreachable for VB code
             new TestSymbolReferenceAnalyzer_VB(null, isTestProject).TestGetSetKeyword(null).Should().BeNull();
 
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void Verify_TokenThreshold() =>
+            // In TokenThreshold.cs there are 4009 tokens which is more than the current limit of 40000
```
pavel-mikula-sonarsource, line 145: Typo? 
```suggestion
            // In TokenThreshold.cs there are 40009 tokens which is more than the current limit of 40000
```
or
```suggestion
            // In TokenThreshold.cs there are 40090 tokens which is more than the current limit of 40000
```
I don't know

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/SymbolReferenceAnalyzerTest.cs
```diff
@@ -161,10 +170,13 @@ private void Verify(string fileName, ProjectType projectType, Action<IReadOnlyLi
...
-                    var info = messages.Single();
-                    info.FilePath.Should().Be(fileName);
-                    verifyReference(info.Reference);
+                    messages.Should().HaveCount(expectedMessageCount);
+                    var info = messages.FirstOrDefault();
+                    if (info != null)
+                    {
+                        info.FilePath.Should().Be(fileName);
+                        verifyReference(info.Reference);
+                    }
```
pavel-mikula-sonarsource, line 179: There's implicit expectation that `expectedMessageCount` will be only 0 or 1. I'd prefer it this way to throw on `Single()` in case somebody breaks the expectation. He will have to rethink the test structure.
```suggestion
                    if (expectedMessageCount != 0)
                    {
                        var info = messages.Single();
                        info.FilePath.Should().Be(fileName);
                        verifyReference(info.Reference);
                    }
```

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/TokenTypeAnalyzerTest.cs
```diff
@@ -76,6 +76,15 @@ public class TokenTypeAnalyzerTest
                 info.Should().ContainSingle(x => x.TokenType == TokenType.Keyword);
             });
 
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void Verify_IdentifierTokenThreshold() =>
+            Verify("IdentifierTokenThreshold.cs", ProjectType.Product, tokenInfo =>
+            {
+                // IdentifierTokenThreshold.cs has 4001 identifiers which exceeds current threshold of 4000. Due to this, the identifiers are not classified
+                tokenInfo.Where(token => token.TokenType == TokenType.TypeName).Should().BeEmpty();
```
pavel-mikula-sonarsource, line 85: `TokenType.TypeName` vs empty - nice!

pavel-mikula-sonarsource (approved):
LGTM, there's typo and one improvement suggestion to consider before merging

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/SymbolReferenceAnalyzerTest.cs
costin-zaharia-sonarsource: 😄

It's `40009`

PR #4319, opened 2021-04-27
Fix S4457 FP with async void methods

### LINKED ISSUE #2665: Update S4457: FP for async void method (event handlers)
### Description

There is no way of splitting async void method into two without breaking another sonar rules. 

### Repro steps

```
private async void OnButtonPressed( object sender, EventArgs e)
{
  if(e == null)
    throw new ArgumentException(nameof(e));

  await Task.Delay(0);
}
```

### Expected behavior

the S4457 should not be raised for async void methods

### Actual behavior

S4457 is raised

### Related information

* SonarLint for VS2017 4.12.0.0
* VS 2017

andrei-epure-sonarsource:
Fixes #2665

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ParameterValidationInAsyncShouldBeWrapped.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ParameterValidationInAsyncShouldBeWrapped.cs
@@ -48,7 +48,7 @@ namespace SonarAnalyzer.Rules.CSharp
                 {
                     var methodDeclaration = (MethodDeclarationSyntax)c.Node;
                     if (!methodDeclaration.Modifiers.Any(SyntaxKind.AsyncKeyword)
-                        || c.SemanticModel.GetTypeInfo(methodDeclaration.ReturnType).Type.SpecialType == SpecialType.System_Void)
+                        || methodDeclaration.ReturnType.ToString() == "void")
                     {
                         return;
                     }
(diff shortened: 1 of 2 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/ParameterValidationInAsyncShouldBeWrapped.cs
```diff
@@ -42,13 +42,13 @@ public sealed class ParameterValidationInAsyncShouldBeWrapped : SonarDiagnosticA
...
-        protected override void Initialize(SonarAnalysisContext context)
-        {
+        protected override void Initialize(SonarAnalysisContext context) =>
             context.RegisterSyntaxNodeActionInNonGenerated(
                 c =>
                 {
                     var methodDeclaration = (MethodDeclarationSyntax)c.Node;
-                    if (!methodDeclaration.Modifiers.Any(SyntaxKind.AsyncKeyword))
+                    if (!methodDeclaration.Modifiers.Any(SyntaxKind.AsyncKeyword)
+                        || c.SemanticModel.GetTypeInfo(methodDeclaration.ReturnType).Type.SpecialType == SpecialType.System_Void)
```
pavel-mikula-sonarsource, line 51: Better API
```suggestion
                        || c.SemanticModel.GetDeclaredSymbol(methodDeclaration).ReturnsVoid)
```

pavel-mikula-sonarsource (approved):
LGTM, please use the other API before merging

analyzers/src/SonarAnalyzer.CSharp/Rules/ParameterValidationInAsyncShouldBeWrapped.cs
andrei-epure-sonarsource: I avoided on purpose getting the symbol, because that can trigger a compilation; whereas I only need the type info for this purpose. 

I will try and do a small benchmark

pavel-mikula-sonarsource: From a peek to Roslyn, it uses `GetSemanticInfoSymbolInNonMemberContext` anyway

andrei-epure-sonarsource: I changed it to remove SemanticModel

Issue #4320, opened 2021-04-27
S3928 should treat nameof() correctly


andrei-epure-sonarsource:
Follow-up of #4180

Should ensure that `nameof` is about the argument and not about a field, a local var etc.

andrei-epure-sonarsource:
actually, `nameof` isn't treated correctly only when the type is not solvable

PR #4321, opened 2021-04-27
Fix S3928 FP with nameof

### LINKED ISSUE #4180: S3928 reports the parameter name as '' when the referenced argument type is not resolvable
Split this off of https://github.com/SonarSource/sonar-dotnet/issues/653#issuecomment-803505601  per @pavel-mikula-sonarsource request

> This rule also triggers a false positive in this scenario as well. I had not yet installed the nuget package with the IMyLogger in it and the S3928 pops up, apparently not knowing the parameter list. (do these not analyze the source code?)
image

andrei-epure-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
```diff
@@ -101,7 +101,13 @@ private static void CheckForIssue(SyntaxNodeAnalysisContext analysisContext)
             }
 
             var methodArgumentNames = GetMethodArgumentNames(objectCreationSyntax);
-            if (methodArgumentNames.Contains(TakeOnlyBeforeDot(parameterNameValue)))
+            var parameterName = TakeOnlyBeforeDot(parameterNameValue);
+            if (string.IsNullOrWhiteSpace(parameterName))
```
andrei-epure-sonarsource, line 105: this is just avoiding the FP - treating `nameof` should be done in #4320

andrei-epure-sonarsource: (also, this class must be refactored, but outside of scope of this PR)

csaba-sagi-sonarsource (approved):
LGTM

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CheckArgumentException.cs
```diff
@@ -171,7 +171,7 @@ void ComplexCase(int a)
         void Bar2(int a)
         {
             // See https://github.com/SonarSource/sonar-dotnet/issues/1867
-            throw new ArgumentNullException(null, string.Empty); // Noncompliant
+            throw new ArgumentNullException(null, string.Empty); // rare FN introduced by fixing https://github.com/SonarSource/sonar-dotnet/issues/4180
```
csaba-sagi-sonarsource, line 174: Please add two more test cases.
```suggestion
            throw new ArgumentNullException(null, string.Empty); // rare FN introduced by fixing https://github.com/SonarSource/sonar-dotnet/issues/4180
            throw new ArgumentNullException(string.Empty, string.Empty); // rare FN introduced by fixing https://github.com/SonarSource/sonar-dotnet/issues/4180
            throw new ArgumentNullException("      ", string.Empty); // Noncompliant
```

analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
csaba-sagi-sonarsource, line 105: We should use IsNullOrEmpty to create less FNs
```suggestion
            if (string.IsNullOrEmpty(parameterName))
```

csaba-sagi-sonarsource (changes requested):
Just one minor change with 2 additional test cases for it.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CheckArgumentException.cs
andrei-epure-sonarsource: nice catch. actually, having `throw new ArgumentNullException("", "")` and forgetting to update it would be quite a common "forget" scenario... so that's a pretty bad FN...

andrei-epure-sonarsource: and I read better the issue description: `"when the referenced argument type is not resolvable"`

that means that `nameof` is handled ok when the type is solvable. This is really an edge-case

andrei-epure-sonarsource:
closing this stale PR

PR #4322, opened 2021-04-27
Fix S927 FP: When implementing generic interfaces it might be more useful to name the argument depending on the type name

### LINKED ISSUE #3047: S927 FP
### Description

False positive for **S927**. When implementing generic interfaces it might be more useful to name the argument depending on the type name than using the generic name from the interface.

### Repro steps

```
private class ConsoleProgressReporter : System.IProgress<ProjectLoadProgress>
{
    public void Report(ProjectLoadProgress loadProgress)
    {

    }
}
```

### Expected behavior

Do not raise the rule.

### Actual behavior

In the previous example I think that **loadProgress** provides more information than *value* used in the generic interface (`void Report(T value);`).

Warning	S927	Rename parameter 'loadProgress' to 'value' to match the interface declaration. (The name of a parameter in an externally visible method override does not match the name of the parameter in the base declaration of the method, or the name of the parameter in the interface declaration of the method or the name of any other partial definition.)

### Related information

* SonarC# Version: 8.2.0.14119
* Visual Studio Version 16.4.3

dmitry-golovinov-sonarsource:
Fixes #3047


analyzers/src/SonarAnalyzer.Common/Rules/ParameterNameMatchesOriginalBase.cs
```diff
@@ -64,7 +65,7 @@ public abstract class ParameterNameMatchesOriginalBase<TSyntaxKind, TMethodDecla
                         }
                         else if (methodSymbol.GetInterfaceMember() is { } interfaceMember)
                         {
-                            VerifyParameters(c, methodSyntax, interfaceMember.Parameters, "interface");
+                            VerifyInterfaceMemberParameters(c, methodSymbol, methodSyntax, interfaceMember, "interface");
```
csaba-sagi-sonarsource, line 68: As we discussed in an offline talk, let`s try to cover the following cases:

```
    public abstract class BaseClass<T>
    {
        public abstract void SomeMethod(T someParameter);
    }

    public class ClassOne : BaseClass<int>
    {
        public override void SomeMethod(int renamedParam)
        {
        }
    }

    public interface IGenericInterface<T>
    {
        void SomeMethod(T someParameter);
    }

    public class Implementation : IGenericInterface<int>
    {
        public void SomeMethod(int renamedParam)
        {
        }
    }
```

And as a bonus if it is a low hanging fruit :

  ```
  public abstract record BaseRecord<T>
    {
        public abstract void SomeMethod(T someParameter);
    }

    public record RecordOne : BaseRecord<int>
    {
        public override void SomeMethod(int renamedParam)
        {
        }
    }
```

File: analyzers/src/SonarAnalyzer.Common/Rules/ParameterNameMatchesOriginalBase.cs [hunk 2]
```diff
@@ -79,5 +80,41 @@ private void VerifyParameters(SyntaxNodeAnalysisContext context, TMethodDeclarat
...
+
+            for (var i = 0; i < parameters.Count; i++)
+            {
+                var parameter = parameters[i];
+                var parameterType = methodSymbol.Parameters[i].Type;
+                var interfaceMethodParameter = interfaceMethodParameters[i];
+                if (!parameter.ValueText.Equals(interfaceMethodParameter.Name, Language.NameComparison))
+                {
+                    var matchingGenericInterfaceArg = genericInterfaceTypeArguments.FirstOrDefault(x => parameterType.DerivesOrImplements(x));
+                    if (matchingGenericInterfaceArg == null || matchingGenericInterfaceArg.Name.IndexOf(parameter.ValueText, StringComparison.OrdinalIgnoreCase) != 0)
```
csaba-sagi-sonarsource, line 112: Also I would remove any restrictions on the renaming in this case. I would not force the user to use the type name in the parameter name.

csaba-sagi-sonarsource (changes requested):
Overall it looks good, let`s try to be more precise and cover some more cases.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ParameterNameMatchesOriginal.cs
```diff
@@ -93,4 +91,53 @@ public partial interface IParameterNamesInPartialMethods
...
+    }
+
+    public abstract class AbstractClassWithGenericMethod
+    {
+        abstract public void Foo<T>(T val);
+    }
+
+    public class InheritedClassWithDefinition : AbstractClassWithGenericMethod
+    {
+        public override void Foo<T>(T myNewName) { }
```
csaba-sagi-sonarsource, line 141: I would say this is a FN, as  the overridden method is still generic, so I do not see why would the parameter name change here. If we opt to solve it, it can be fixed by checking that the overridden parameter type kind is not generic anymore.

analyzers/src/SonarAnalyzer.Common/Rules/ParameterNameMatchesOriginalBase.cs
```diff
@@ -60,11 +60,11 @@ public abstract class ParameterNameMatchesOriginalBase<TSyntaxKind, TMethodDecla
                         }
                         else if (methodSymbol.OverriddenMethod != null)
                         {
-                            VerifyParameters(c, methodSyntax, methodSymbol.OverriddenMethod.Parameters, "base class");
+                            VerifyGenericParameters(c, methodSyntax, methodSymbol.OverriddenMethod.OriginalDefinition.Parameters, "base class");
```
csaba-sagi-sonarsource, line 63: Please add a few test cases with deeper inheritance, just to be sure that the originalDefinition does the work in more "complex" cases.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ParameterNameMatchesOriginal.cs
```diff
@@ -93,4 +91,53 @@ public partial interface IParameterNamesInPartialMethods
         {
         }
     }
+
+    public interface IGenericInterface<A>
```
csaba-sagi-sonarsource, line 95: We can also add a test case with a method without parameters, as there is no such example.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ParameterNameMatchesOriginal.vb
```diff
@@ -60,3 +60,58 @@ Public Class IgnoreCaseChange
     End Sub
 
 End Class
+
```
csaba-sagi-sonarsource, line 63: Seems to me there are no cases for VB where the parameter name is only different by the case. for example: somePARAMETER and SOMEparameter.

csaba-sagi-sonarsource (changes requested):
A few test cases should be added and maybe fix the FN if you agree that it is a FN. Otherwise looks nice.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ParameterNameMatchesOriginal.vb
```diff
@@ -60,3 +60,125 @@ Public Class IgnoreCaseChange
...
+End Class
+
+Public Class AnotherParameterClass
+End Class
+
+Public Class Implementation
+    Implements IGenericInterface(Of ParameterClass)
+    Public Sub DoSomething() Implements IGenericInterface(Of ParameterClass).DoSomething
+    End Sub
+    Public Sub DoSomething(ByVal parameter As ParameterClass) Implements IGenericInterface(Of ParameterClass).DoSomething
```
pavel-mikula-sonarsource, line 84: This is not readable, there should be empty line between procedures
```suggestion
    Implements IGenericInterface(Of ParameterClass)
    
    Public Sub DoSomething() Implements IGenericInterface(Of ParameterClass).DoSomething
    End Sub
    
    Public Sub DoSomething(ByVal parameter As ParameterClass) Implements IGenericInterface(Of ParameterClass).DoSomething
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ParameterNameMatchesOriginal.vb
```diff
@@ -60,3 +60,125 @@ Public Class IgnoreCaseChange
...
+        '                           ^^^^^^^^^^^^^^^^
+    End Sub
+    Public Sub DoSomethingCaseSensitive(ByVal Value As ParameterClass, ByVal IntValue As Integer) Implements IGenericInterface(Of ParameterClass).DoSomethingCaseSensitive
+    End Sub
+End Structure
+
+Public MustInherit Class BaseClass(Of T)
+    Public MustOverride Sub SomeMethod(ByVal someParameter As T)
+    Public MustOverride Sub SomeMethod(ByVal someParameter As T, ByVal intParam As Integer)
+End Class
```
pavel-mikula-sonarsource, line 129: The whitespace should be like this:
```suggestion
Public MustInherit Class BaseClass(Of T)

    Public MustOverride Sub SomeMethod(ByVal someParameter As T)
    Public MustOverride Sub SomeMethod(ByVal someParameter As T, ByVal intParam As Integer)

End Class
```

analyzers/src/SonarAnalyzer.Common/Rules/ParameterNameMatchesOriginalBase.cs
```diff
@@ -79,5 +79,21 @@ private void VerifyParameters(SyntaxNodeAnalysisContext context, TMethodDeclarat
...
+
+        private void VerifyGenericParameters(SyntaxNodeAnalysisContext context, TMethodDeclarationSyntax methodSyntax, IList<IParameterSymbol> actualParameters,
+            IList<IParameterSymbol> expectedParameters, string expectedLocation)
+        {
+            var parameters = ParameterIdentifiers(methodSyntax).ToList();
+            for (var i = 0; i < parameters.Count; i++)
+            {
+                var parameter = parameters[i];
+                var expectedParameter = expectedParameters[i];
+                if (!parameter.ValueText.Equals(expectedParameter.Name, Language.NameComparison)
```
pavel-mikula-sonarsource, line 91: This line LGTM. It's usual (and desired) to rename `value As Integer` from interface to `Value As Integer` as VB.NET doesn't suffer from the need to do a variables lower case prefixed.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ParameterNameMatchesOriginal.vb
```diff
@@ -60,3 +60,125 @@ Public Class IgnoreCaseChange
     End Sub
 
 End Class
+
+Public Interface IGenericInterface(Of A)
+    Sub DoSomething()
+    Sub DoSomething(ByVal value As A)
```
pavel-mikula-sonarsource, line 66: All `ByVal` can be removed for better readability

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ParameterNameMatchesOriginal.vb
```diff
@@ -60,3 +60,125 @@ Public Class IgnoreCaseChange
...
+    Sub DoSomething(ByVal value As A, ByVal intValue As Integer)
```
pavel-mikula-sonarsource, line 67: _Nitpick:_ Example like this is more authentic
```suggestion
    Sub DoSomething(ByVal Value As A, ByVal IntValue As Integer)
```

csaba-sagi-sonarsource (approved):
LGTM! Please address the comments that Pavel put there.

PR #4323, opened 2021-04-27
Fix S2376 FP with getter in base class

### LINKED ISSUE #2390: S2376 FP: should not raise for overridden property with base getter
### Description

If a virtual property has a getter and a derived class overrides the property, but does not specify a getter, Sonar raises a false positive S2376 for the overridden property.

### Repro steps

```
	public class A
	{
		protected int m = 5;
		public virtual int M
		{
			get { return m; }
			set { m = value; }
		}
	}
	public class B : A
	{
		public override int M
		{
			set { m = value + 1; }
		}
	}
	public class Program
	{
		public static void Main(string[] args)
		{
			B b = new B();
			Console.WriteLine(b.M);
		}
	}
```
### Expected behavior

S2376 is not raised for B.M.

### Actual behavior

S2376 is raised for B.M.

### Known workarounds

```
		public override int M
		{
			get{ return base.M; }
			set { m = value + 1; }
		}
```

### Related information

* SonarC# 7.13
* Visual Studio 2017 (15.9.7)
* If running through the Scanner for MSBuild, its version: 4.6.0.1930-net46

andrei-epure-sonarsource:
Fixes #2390

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyWriteOnly.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyWriteOnly.cs
@@ -51,7 +51,9 @@ namespace SonarAnalyzer.Rules.CSharp
                 return false;
             }
 
-            return accessors.Accessors.First().IsKind(SyntaxKind.SetAccessorDeclaration);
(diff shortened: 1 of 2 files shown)


csaba-sagi-sonarsource (approved):
LGTM

PR #4326, opened 2021-04-27
Rename .NET ITs

pavel-mikula-sonarsource:
Cleanup to align project names into a regular form.

Review per commit should be easy (renamed directories, updated root paths in JSONs)

This is a draft to avoid merging before release


andrei-epure-sonarsource (approved):
LGTM

PR #4328, opened 2021-04-28
Update parent pom and sonar-analyzer-commons

andrei-epure-sonarsource:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -54,6 +54,8 @@
     <artifactsToDownload>${project.groupId}:SonarAnalyzer.CSharp:nupkg,${project.groupId}:SonarAnalyzer.VisualBasic:nupkg</artifactsToDownload>
     <!-- We are ignoring java doc warnings - this is because we are using JDK 11. Ideally we should not do that. -->
     <doclint>none</doclint>
+    <sonar.analyzer.commons>1.14.1.690</sonar.analyzer.commons>
+    <sonar.version>7.9</sonar.version>
   </properties>
 
   <profiles>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.sonarsource.parent</groupId>
     <artifactId>parent</artifactId>
-    <version>58.0.22</version>
+    <version>59.0.29</version>
   </parent>
 
   <groupId>org.sonarsource.dotnet</groupId>
--- a/sonar-csharp-plugin/pom.xml
+++ b/sonar-csharp-plugin/pom.xml
@@ -40,13 +40,16 @@
     <maven.test.redirectTestOutputToFile>true</maven.test.redirectTestOutputToFile>
     <sonarAnalyzer.workDirectory>${project.build.directory}/analyzer</sonarAnalyzer.workDirectory>
     <documentationDirectory>${project.basedir}/src/main/resources</documentationDirectory>
+    <sonar.analyzer.commons>1.14.0.681</sonar.analyzer.commons>
   </properties>
 
   <dependencies>
+    <!-- provided at runtime -->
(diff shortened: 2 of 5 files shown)


sonar-csharp-plugin/pom.xml
```diff
@@ -40,30 +40,34 @@
     <maven.test.redirectTestOutputToFile>true</maven.test.redirectTestOutputToFile>
     <sonarAnalyzer.workDirectory>${project.build.directory}/analyzer</sonarAnalyzer.workDirectory>
     <documentationDirectory>${project.basedir}/src/main/resources</documentationDirectory>
+    <sonar.analyzer.commons>1.14.1.690</sonar.analyzer.commons>
```
pavel-mikula-sonarsource, line 43: Does this have to be on 3 different places? Can we define the property in the root pom once?

sonar-dotnet-shared-library/pom.xml
```diff
@@ -20,34 +20,47 @@
...
     <dependency>
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
       <version>${protobuf.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.sonarsource.analyzer-commons</groupId>
+      <artifactId>sonar-analyzer-commons</artifactId>
+      <version>${sonar.analyzer.commons}</version>
+      <scope>compile</scope>
```
pavel-mikula-sonarsource, line 56: _For consistency:_ The other occurrences of compiled dependencies do not use explicit value (like the one just above this, or the modified ones in other two POMs)
```suggestion
```

sonar-dotnet-shared-library/pom.xml
```diff
@@ -20,34 +20,47 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.sonarsource.analyzer-commons</groupId>
+      <artifactId>sonar-xml-parsing</artifactId>
+      <version>${sonar.analyzer.commons}</version>
+      <scope>compile</scope>
```
pavel-mikula-sonarsource, line 62: Same
```suggestion
```

pavel-mikula-sonarsource (approved):
LGTM, my comments/questions are optional and we can do them later (if we can do them)

pom.xml
```diff
@@ -54,6 +54,8 @@
     <artifactsToDownload>${project.groupId}:SonarAnalyzer.CSharp:nupkg,${project.groupId}:SonarAnalyzer.VisualBasic:nupkg</artifactsToDownload>
     <!-- We are ignoring java doc warnings - this is because we are using JDK 11. Ideally we should not do that. -->
     <doclint>none</doclint>
+    <sonar.analyzer.commons>1.14.1.690</sonar.analyzer.commons>
+    <sonar.version>7.9</sonar.version>
```
andrei-epure-sonarsource, line 58: note: I tried with 8.x (latest and even 8.0) and they removed the `import org.sonar.api.internal.google.common.annotations.VisibleForTesting;` API

I will ask if it's a problem at this point or we can ship it.

pavel-mikula-sonarsource: We have 
```
    <sonar.version>7.9</sonar.version>
    <sonarLTS.version>7.9</sonarLTS.version>
```
I think we should update them both after LTS release

pavel-mikula-sonarsource: Why not to move `sonarLTS.version` as well when moving `sonar.version`? They were next to each other

andrei-epure-sonarsource: sonarLTS isn't in sonar-dotnet-shared-library, so I didn't move it in the parent.

pavel-mikula-sonarsource (approved):
Still LGTM

PR #4333, opened 2021-04-28
Fix S2699 throws AD0001 exception

### LINKED ISSUE #4332: S2699 throws AD0001 when invoked via AnalyzerRunner.exe
happens (at least) for `nodatime`, `n2cms`, `efcore`

```
  AD0001: 16 instances
error AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.TestMethodShouldContainAssertion' threw an exception of type 'System.ArgumentException' with message 'SyntaxTree is not part of the compilation
Parameter name: syntaxTree'.
Exception occurred with following context:
Compilation: NodaTime.Test
SyntaxTree: C:\Workspace\external-projects\PERF_SUITE\projects\nodatime\src\NodaTime.Test\Globalization\NodaFormatInfoTest.cs
SyntaxNode: [Test] [TestCaseSource(typeof(Cultures ... [MethodDeclarationSyntax]@[756..967) (21,8)-(26,9)

System.ArgumentException: SyntaxTree is not part of the compilation
Parameter name: syntaxTree
   at Microsoft.CodeAnalysis.CSharp.CSharpCompilation.GetSemanticModel(SyntaxTree syntaxTree, Boolean ignoreAccessibility) in C:\Workspace\external-projects\roslyn\src\Compilers\CSharp\Portable\Compilation\CSharpCompilation.cs:line 2147
   at Microsoft.CodeAnalysis.CSharp.CSharpCompilation.CommonGetSemanticModel(SyntaxTree syntaxTree, Boolean ignoreAccessibility) in C:\Workspace\external-projects\roslyn\src\Compilers\CSharp\Portable\Compilation\CSharpCompilation.cs:line 3229
   at Microsoft.CodeAnalysis.Compilation.GetSemanticModel(SyntaxTree syntaxTree, Boolean ignoreAccessibility) in C:\Workspace\external-projects\roslyn\src\Compilers\Core\Portable\Compilation\Compilation.cs:line 211
   at SonarAnalyzer.Rules.CSharp.TestMethodShouldContainAssertion.ContainsAssertion(MethodDeclarationSyntax methodDeclaration, SemanticModel previousSemanticModel, ISet`1 visitedSymbols, Int32 level)
   at SonarAnalyzer.Rules.CSharp.TestMethodShouldContainAssertion.ContainsAssertion(MethodDeclarationSyntax methodDeclaration, SemanticModel previousSemanticModel, ISet`1 visitedSymbols, Int32 level)
   at SonarAnalyzer.Rules.CSharp.TestMethodShouldContainAssertion.<>c.<Initialize>b__11_0(SyntaxNodeAnalysisContext c)
   at SonarAnalyzer.Helpers.DiagnosticAnalyzerContextHelper.<>c__DisplayClass1_0`1.<RegisterSyntaxNodeActionInNonGenerated>b__0(SyntaxNodeAnalysisContext c)
   at SonarAnalyzer.Helpers.SonarAnalysisContext.<>c__DisplayClass46_0`1.<RegisterContextAction>b__0(TContext c)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.<>c__62`1.<ExecuteSyntaxNodeAction>b__62_0(ValueTuple`2 data) in C:\Workspace\external-projects\roslyn\src\Compilers\Core\Portable\DiagnosticAnalyzer\AnalyzerExecutor.cs:line 918
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock[TArg](DiagnosticAnalyzer analyzer, Action`1 analyze, TArg argument, Nullable`1 info) in C:\Workspace\external-projects\roslyn\src\Compilers\Core\Portable\DiagnosticAnalyzer\AnalyzerExecutor.cs:line 1596
-----

Suppress the following diagnostics to disable this analyzer: S2699
```

dmitry-golovinov-sonarsource:
Fixes #4332

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
@@ -64,8 +64,11 @@ namespace SonarAnalyzer.Helpers
 
         public override void Visit(SyntaxNode node)
         {
-            semanticModel = node.EnsureCorrectSemanticModel(semanticModel ?? compilation.GetSemanticModel(node.SyntaxTree));
-            base.Visit(node);
+            semanticModel = node.EnsureCorrectSemanticModelOrDefault(semanticModel ?? compilation.GetSemanticModel(node.SyntaxTree));
+            if (semanticModel != null)
+            {
+                base.Visit(node);
+            }
         }
 
         // TupleExpression "(a, b) = qix"
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/MethodParameterUnused.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/MethodParameterUnused.cs
@@ -235,8 +235,9 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static bool IsMethodUsedAsEventHandlerFunctionOrActionInExpression(IMethodSymbol methodSymbol, ExpressionSyntax expression, SemanticModel semanticModel) =>
             !expression.IsKind(SyntaxKind.InvocationExpression)
+            && semanticModel != null
             && IsStandaloneExpression(expression)
-            && methodSymbol.Equals(semanticModel?.GetSymbolInfo(expression).Symbol?.OriginalDefinition);
+            && methodSymbol.Equals(semanticModel.GetSymbolInfo(expression).Symbol?.OriginalDefinition);
 
         private static bool IsStandaloneExpression(ExpressionSyntax expression)
         {
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/MethodParameterUnused.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/MethodParameterUnused.cs
@@ -226,7 +226,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private static bool IsUsedAsEventHandlerFunctionOrAction(MethodContext declaration) =>
             declaration.Symbol.ContainingType.DeclaringSyntaxReferences
             .Select(r => r.GetSyntax())
-            .Any(n => IsMethodUsedAsEventHandlerFunctionOrActionWithinNode(declaration.Symbol, n, n.EnsureCorrectSemanticModel(declaration.Context.SemanticModel)));
+            .Any(n => IsMethodUsedAsEventHandlerFunctionOrActionWithinNode(declaration.Symbol, n, n.EnsureCorrectSemanticModelOrDefault(declaration.Context.SemanticModel)));
 
         private static bool IsMethodUsedAsEventHandlerFunctionOrActionWithinNode(IMethodSymbol methodSymbol, SyntaxNode typeDeclaration, SemanticModel semanticModel) =>
             typeDeclaration.DescendantNodes()
@@ -236,7 +236,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private static bool IsMethodUsedAsEventHandlerFunctionOrActionInExpression(IMethodSymbol methodSymbol, ExpressionSyntax expression, SemanticModel semanticModel) =>
             !expression.IsKind(SyntaxKind.InvocationExpression)
             && IsStandaloneExpression(expression)
-            && methodSymbol.Equals(semanticModel.GetSymbolInfo(expression).Symbol?.OriginalDefinition);
+            && methodSymbol.Equals(semanticModel?.GetSymbolInfo(expression).Symbol?.OriginalDefinition);
 
         private static bool IsStandaloneExpression(ExpressionSyntax expression)
         {
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
@@ -93,7 +93,11 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static bool ContainsAssertion(MethodDeclarationSyntax methodDeclaration, SemanticModel previousSemanticModel, ISet<IMethodSymbol> visitedSymbols, int level)
         {
-            var currentSemanticModel = methodDeclaration.EnsureCorrectSemanticModel(previousSemanticModel);
+            var currentSemanticModel = methodDeclaration.EnsureCorrectSemanticModelOrDefault(previousSemanticModel);
+            if (currentSemanticModel == null)
+            {
+                return false;
+            }
             var descendantNodes = methodDeclaration.DescendantNodes();
             var invocations = descendantNodes.OfType<InvocationExpressionSyntax>().ToArray();
             if (invocations.Any(x => IsAssertion(x))
--- a/analyzers/src/SonarAnalyzer.Common/Trackers/ConstantValueFinder.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Trackers/ConstantValueFinder.cs
@@ -46,10 +46,19 @@ namespace SonarAnalyzer.Helpers
         public object FindConstant(SyntaxNode node) =>
             FindConstant(node, null);
 
-        private object FindConstant(SyntaxNode node, HashSet<SyntaxNode> visitedVariables) =>
-            node == null || node.RawKind == nullLiteralExpressionSyntaxKind  // Performance shortcut
-            ? null
-            : node.EnsureCorrectSemanticModel(semanticModel).GetConstantValue(node).Value ?? FindAssignedConstant(node, visitedVariables);
+        private object FindConstant(SyntaxNode node, HashSet<SyntaxNode> visitedVariables)
+        {
+            if (node == null || node.RawKind == nullLiteralExpressionSyntaxKind) // Performance shortcut
+            {
+                return null;
+            }
+            var nodeSemanticModel = node.EnsureCorrectSemanticModelOrDefault(semanticModel);
+            if (nodeSemanticModel == null)
+            {
+                return null;
+            }
+            return nodeSemanticModel.GetConstantValue(node).Value ?? FindAssignedConstant(node, visitedVariables);
+        }
 
         private object FindAssignedConstant(SyntaxNode node, HashSet<SyntaxNode> visitedVariables)
         {
@@ -59,7 +68,7 @@ namespace SonarAnalyzer.Helpers
 
             SyntaxNode FindFieldInitializer()
             {
-                if (identifier.EnsureCorrectSemanticModel(semanticModel).GetSymbolInfo(identifier).Symbol is IFieldSymbol fieldSymbol
+                if (identifier.EnsureCorrectSemanticModelOrDefault(semanticModel)?.GetSymbolInfo(identifier).Symbol is IFieldSymbol fieldSymbol
                     && VariableDeclarator(fieldSymbol.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax()) is { } variable
                     && (visitedVariables == null || !visitedVariables.Contains(variable)))
                 {
--- a/analyzers/src/SonarAnalyzer.Common/Trackers/ConstantValueFinder.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Trackers/ConstantValueFinder.cs
@@ -53,11 +53,11 @@ namespace SonarAnalyzer.Helpers
                 return null;
             }
             return node.EnsureCorrectSemanticModelOrDefault(semanticModel) is { } nodeSemanticModel
-                ? nodeSemanticModel.GetConstantValue(node).Value ?? FindAssignedConstant(node, visitedVariables, nodeSemanticModel)
+                ? nodeSemanticModel.GetConstantValue(node).Value ?? FindAssignedConstant(node, nodeSemanticModel, visitedVariables)
                 : null;
         }
 
-        private object FindAssignedConstant(SyntaxNode node, HashSet<SyntaxNode> visitedVariables, SemanticModel nodeSemanticModel)
+        private object FindAssignedConstant(SyntaxNode node, SemanticModel nodeSemanticModel, HashSet<SyntaxNode> visitedVariables)
         {
             return node is TIdentifierNameSyntax identifier
                 ? FindConstant(assignmentFinder.FindLinearPrecedingAssignmentExpression(IdentifierName(identifier), node, FindFieldInitializer), visitedVariables)
--- a/analyzers/src/SonarAnalyzer.Common/Trackers/ConstantValueFinder.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Trackers/ConstantValueFinder.cs
@@ -52,12 +52,9 @@ namespace SonarAnalyzer.Helpers
             {
                 return null;
             }
-            var nodeSemanticModel = node.EnsureCorrectSemanticModelOrDefault(semanticModel);
-            if (nodeSemanticModel == null)
-            {
-                return null;
-            }
-            return nodeSemanticModel.GetConstantValue(node).Value ?? FindAssignedConstant(node, visitedVariables);
+            return node.EnsureCorrectSemanticModelOrDefault(semanticModel) is { } nodeSemanticModel
+                ? nodeSemanticModel.GetConstantValue(node).Value ?? FindAssignedConstant(node, visitedVariables)
+                : null;
         }
 
         private object FindAssignedConstant(SyntaxNode node, HashSet<SyntaxNode> visitedVariables)
--- a/analyzers/src/SonarAnalyzer.Common/Trackers/ConstantValueFinder.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Trackers/ConstantValueFinder.cs
@@ -53,11 +53,11 @@ namespace SonarAnalyzer.Helpers
                 return null;
             }
             return node.EnsureCorrectSemanticModelOrDefault(semanticModel) is { } nodeSemanticModel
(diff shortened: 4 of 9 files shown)


analyzers/src/SonarAnalyzer.Common/Trackers/ConstantValueFinder.cs
```diff
@@ -46,10 +46,19 @@ protected ConstantValueFinder(SemanticModel semanticModel, AssignmentFinder assi
...
+            if (node == null || node.RawKind == nullLiteralExpressionSyntaxKind) // Performance shortcut
+            {
+                return null;
+            }
+            var nodeSemanticModel = node.EnsureCorrectSemanticModelOrDefault(semanticModel);
+            if (nodeSemanticModel == null)
+            {
+                return null;
+            }
+            return nodeSemanticModel.GetConstantValue(node).Value ?? FindAssignedConstant(node, visitedVariables);
```
pavel-mikula-sonarsource, line 60:
```suggestion
            return node.EnsureCorrectSemanticModelOrDefault(semanticModel) is { } nodeSemanticModel
                ? nodeSemanticModel.GetConstantValue(node).Value ?? FindAssignedConstant(node, visitedVariables)
                : null;
```

analyzers/src/SonarAnalyzer.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -236,7 +236,7 @@ private static IImmutableList<IParameterSymbol> GetUnusedParameters(MethodContex
         private static bool IsMethodUsedAsEventHandlerFunctionOrActionInExpression(IMethodSymbol methodSymbol, ExpressionSyntax expression, SemanticModel semanticModel) =>
             !expression.IsKind(SyntaxKind.InvocationExpression)
             && IsStandaloneExpression(expression)
-            && methodSymbol.Equals(semanticModel.GetSymbolInfo(expression).Symbol?.OriginalDefinition);
+            && methodSymbol.Equals(semanticModel?.GetSymbolInfo(expression).Symbol?.OriginalDefinition);
```
pavel-mikula-sonarsource, line 239: Let's add `&& semanticModel != null` as 2nd condition for performance reason. We don't need to evaluate `IsStandaloneExpression` in that case

pavel-mikula-sonarsource (changes requested):
Improvements suggested

pavel-mikula-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
```diff
@@ -93,7 +93,11 @@ public sealed class TestMethodShouldContainAssertion : SonarDiagnosticAnalyzer
 
         private static bool ContainsAssertion(MethodDeclarationSyntax methodDeclaration, SemanticModel previousSemanticModel, ISet<IMethodSymbol> visitedSymbols, int level)
         {
-            var currentSemanticModel = methodDeclaration.EnsureCorrectSemanticModel(previousSemanticModel);
+            var currentSemanticModel = methodDeclaration.EnsureCorrectSemanticModelOrDefault(previousSemanticModel);
+            if (currentSemanticModel == null)
+            {
+                return false;
```
andrei-epure-sonarsource, line 99: won't this mean we're going to raise issues when there is this problem?  If `!ContainsAssertion` , we raise.

We should be silent if there's no semantic model. Or am I missing something?

andrei-epure-sonarsource (dismissed):

andrei-epure-sonarsource:
Let's not merge this before SQ LTS. And try to come up with a tests to cover the logic.

The `analyzerRunner.exe` scenario is an edge case, no need to hurry.

analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
andrei-epure-sonarsource: from offline discussion:
>It's in crossmethod inspection, we return false when we dont't inspect deeper (or we can't tell). As it's not for real life scenario, it should be fine as is

analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSymbolUsageCollector.cs
```diff
@@ -64,8 +64,11 @@ public CSharpSymbolUsageCollector(Compilation compilation, IEnumerable<ISymbol>
 
         public override void Visit(SyntaxNode node)
         {
-            semanticModel = node.EnsureCorrectSemanticModel(semanticModel ?? compilation.GetSemanticModel(node.SyntaxTree));
-            base.Visit(node);
+            semanticModel = node.EnsureCorrectSemanticModelOrDefault(semanticModel ?? compilation.GetSemanticModel(node.SyntaxTree));
+            if (semanticModel != null)
```
andrei-epure-sonarsource, line 68: For the C# usage symbol collector, you can create two different "snippet" compilations in the UT, and give compilation1 to the constructor, and give a SyntaxNode from the second compilation to Visit. Then, it should not do the visiting.

See ConstantValueFinderTest for example how to do that

And then check `UsedSymbols` or some other property, that it collects symbols when the semanticModel is resolved, and that it doesn't when the SemanticModel doesn't get resolved.

pavel-mikula-sonarsource (changes requested):
Let's add UT as discussed. I'm adding `Request changes` to keep the history explicit.

analyzers/tests/SonarAnalyzer.UnitTest/Trackers/ConstantValueFinderTest.cs
```diff
@@ -57,5 +57,34 @@ public int Method()
...
+        return Field;
+    }
+}";
+            var compilation1 = SolutionBuilder.Create().AddProject(AnalyzerLanguage.CSharp, createExtraEmptyFile: false)
+                .AddSnippet(code1)
+                .GetCompilation();
+            var compilation2 = SolutionBuilder.Create().AddProject(AnalyzerLanguage.CSharp, createExtraEmptyFile: false)
+                .AddSnippet(code2)
+                .GetCompilation();
+            var tree = compilation2.SyntaxTrees.Single(x => x.GetRoot().DescendantNodes().OfType<MethodDeclarationSyntax>().Any());
```
andrei-epure-sonarsource, line 84: I suggest to improve the naming a bit, to make the differences obvious and easier to follow:
- compilation1  -> firstCompilation
- compilation2 -> secondCompilation`

You don't need `tree` from what I see, you can merge that into the LINQ query for returnExpression
Also:
-  returnExpression -> secondCompilationReturnExpression

And then you have:
```
firstCompilationFinder.FindConstant(secondCompilationReturnExpression).Should().BeNull();
```

which is much clearer

analyzers/tests/SonarAnalyzer.UnitTest/Trackers/ConstantValueFinderTest.cs
```diff
@@ -57,5 +57,34 @@ public int Method()
             var finder = new CSharpConstantValueFinder(compilation.GetSemanticModel(tree));
             finder.FindConstant(returnExpression).Should().Be(42);
         }
+
+        [TestMethod]
+        public void WrongCompilationBeingUsed()
+        {
+            const string code1 = @"
+public partial class Sample
```
andrei-epure-sonarsource, line 65: using `partial class` and the same name makes the test hard to follow, because the two partial classes are in two different solutions, so they're totally different.

IMO for the sake of testing this scenario you can have just two classes `Foo` and `Bar`  and pass the wrong Compilation to the class, and it will be easy to understand what this test is doing

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/CSharpSymbolUsageCollectorTest.cs
```diff
@@ -0,0 +1,86 @@
...
+}";
+            const string code2 = @"
+public partial class Sample
+{
+    public int Bar()
+    {
+        return Field;
+    }
+}";
+            var compilation1 = SolutionBuilder.Create().AddProject(AnalyzerLanguage.CSharp, createExtraEmptyFile: false)
```
andrei-epure-sonarsource, line 58: same here: I suggest to use "firstCompilation" and "secondCompilation" (and "firstTree" and "secondTree") because it's easier do differentiate the names

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/CSharpSymbolUsageCollectorTest.cs
```diff
@@ -0,0 +1,86 @@
...
+namespace SonarAnalyzer.UnitTest.Helpers
+{
+    [TestClass]
+    public class CSharpSymbolUsageCollectorTest
+    {
+        [TestMethod]
+        public void VerifyUsagesBeingCollectedOnMatchingSyntaxNodes()
+        {
+            const string code1 = @"
+public partial class Sample
```
andrei-epure-sonarsource, line 40: same comment as on `WrongCompilationBeingUsed` - why use `partial class Sample` in both ?

either way, in real life compilation2  wouldn't compile in its own solution, right?

andrei-epure-sonarsource (changes requested):
Great to see these tests, I have some readability suggestions

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/CSharpSymbolUsageCollectorTest.cs
dmitry-golovinov-sonarsource: Renamed the classes and removed partial

analyzers/tests/SonarAnalyzer.UnitTest/Trackers/ConstantValueFinderTest.cs
```diff
@@ -57,5 +57,33 @@ public int Method()
...
+            const string firstSnippet = @"
+public class Foo
+{
+    private static int Original = 42;
+    private int Field = Original;
+}";
+            const string secondSnippet = @"
+public class Bar
+{
+    public int Method()
```
andrei-epure-sonarsource, line 73: this code still doesn't compile

So even if you use a secondCompilationFinder, it will still return `null`, because the `Field` symbol does not have any symbol.

If I add these to the test, it fails, because there's no symbol:
```
var secondCompilationFinder = new CSharpConstantValueFinder(secondCompilation.GetSemanticModel(secondCompilation.SyntaxTrees.Single()));
secondCompilationFinder.FindConstant(secondCompilationReturnExpression).Should().NotBeNull();
```

This brings confusion to the test (is it just testing that the `firstCompilationFinder` cannot find symbols because the wrong compilation is used, or because the `secondSnippet` doesn't compile , so it has no symbols even for its own compilation?).

I suggest to make the `Bar` class compilable.

```
public class Bar
{
    private int Field = 42;
    public int Method()
    {
      return Field;
     }
}
```
And then these assertions will both work:
```
firstCompilationFinder.FindConstant(secondCompilationReturnExpression).Should().BeNull();
secondCompilationFinder.FindConstant(secondCompilationReturnExpression).Should().NotBeNull();
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/CSharpSymbolUsageCollectorTest.cs
```diff
@@ -0,0 +1,85 @@
...
+    public int FooMethod(int arg)
+    {
+        Field += arg;
+        return Field;
+    }
+}";
+            const string secondSnippet = @"
+public class Bar
+{
+    public int BarMethod()
```
andrei-epure-sonarsource, line 53: same here, please add `private int Field = 42;` above `BarMethod` so that the code can compile

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/CSharpSymbolUsageCollectorTest.cs
```diff
@@ -0,0 +1,85 @@
...
+            var secondCompilation = SolutionBuilder.Create().AddProject(AnalyzerLanguage.CSharp, createExtraEmptyFile: false)
+                .AddSnippet(secondSnippet)
+                .GetCompilation();
+
+            var firstTree = firstCompilation.SyntaxTrees.Single();
+            var fooMethodDecl = firstTree.GetRoot().DescendantNodes().OfType<MethodDeclarationSyntax>().Single();
+            var semanticModel = firstCompilation.GetSemanticModel(firstTree);
+
+            var fieldSymbol = semanticModel.GetSymbolInfo(fooMethodDecl.DescendantNodes().OfType<ReturnStatementSyntax>().Single().Expression).Symbol;
+            var knownSymbolsFirstCompilation = new List<ISymbol> { fieldSymbol };
```
andrei-epure-sonarsource, line 70: nitpick: if you use `firstCompilationKnownSymbols` I can tell from the beginning that this refers to `first...` - the same for other names. It's easier to navigate the test, I can see `var firstCompilationXyz` and I don't have to read what's on the rest of the line to tell what the `Xyz` is about.

andrei-epure-sonarsource (changes requested):
last few comments

analyzers/tests/SonarAnalyzer.UnitTest/Trackers/ConstantValueFinderTest.cs
dmitry-golovinov-sonarsource: Haven't got this from the beginning. Indeed, those 2 are not compilable!

andrei-epure-sonarsource (approved):
LGTM

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/CSharpSymbolUsageCollectorTest.cs
```diff
@@ -0,0 +1,85 @@
...
+}";
+            var firstCompilation = SolutionBuilder.Create().AddProject(AnalyzerLanguage.CSharp, createExtraEmptyFile: false)
+                .AddSnippet(firstSnippet)
+                .GetCompilation();
+            var secondCompilation = SolutionBuilder.Create().AddProject(AnalyzerLanguage.CSharp, createExtraEmptyFile: false)
+                .AddSnippet(secondSnippet)
+                .GetCompilation();
+
+            var firstTree = firstCompilation.SyntaxTrees.Single();
+            var fooMethodDecl = firstTree.GetRoot().DescendantNodes().OfType<MethodDeclarationSyntax>().Single();
```
pavel-mikula-sonarsource, line 66: _Nitpick:_ This should be enough to name it, as there's no other `fooMethodSymbol` or `fooMethodBody` that would collide. We usually don't use half-words like the Java `Impl`
```suggestion
            var fooMethod = firstTree.GetRoot().DescendantNodes().OfType<MethodDeclarationSyntax>().Single();
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/CSharpSymbolUsageCollectorTest.cs
```diff
@@ -0,0 +1,85 @@
...
+            var firstCompilationSemanticModel = firstCompilation.GetSemanticModel(firstTree);
+
+            var firstCompilationFieldSymbol = firstCompilationSemanticModel.GetSymbolInfo(fooMethodDecl.DescendantNodes().OfType<ReturnStatementSyntax>().Single().Expression).Symbol;
```
pavel-mikula-sonarsource, line 69: _Nitpick:_ I don't see a reason for this empty line. It's a sequence of `firstCompilationXxxx =` initializations.
```suggestion
            var firstCompilationSemanticModel = firstCompilation.GetSemanticModel(firstTree);
            var firstCompilationFieldSymbol = firstCompilationSemanticModel.GetSymbolInfo(fooMethodDecl.DescendantNodes().OfType<ReturnStatementSyntax>().Single().Expression).Symbol;
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/CSharpSymbolUsageCollectorTest.cs
```diff
@@ -0,0 +1,85 @@
...
+            var firstCompilationKnownSymbols = new List<ISymbol> { firstCompilationFieldSymbol };
+
+            // compilation matches semantic model and syntax node
+            var firstCompilationUsageCollector = new CSharpSymbolUsageCollector(firstCompilation, firstCompilationKnownSymbols);
+            firstCompilationUsageCollector.Visit(fooMethodDecl);
+            firstCompilationUsageCollector.UsedSymbols.Should().NotBeEmpty();
+            var firstCompilationUsedSymbols = firstCompilationUsageCollector.UsedSymbols;
+
+            var secondTree = secondCompilation.SyntaxTrees.Single();
+            var barMethodDecl = secondTree.GetRoot().DescendantNodes().OfType<MethodDeclarationSyntax>().Single();
```
pavel-mikula-sonarsource, line 79: Same as above
```suggestion
            var barMethod = secondTree.GetRoot().DescendantNodes().OfType<MethodDeclarationSyntax>().Single();
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/CSharpSymbolUsageCollectorTest.cs
```diff
@@ -0,0 +1,85 @@
...
+            // compilation doesn't match syntax node, since it belongs to another compilation
```
pavel-mikula-sonarsource, line 80: _Nitpick:_ I'd move this comment 2 lines higher to have a regular structure (first comment, var, visit, assert, second comment, var, visit, assert).

pavel-mikula-sonarsource (approved):
LGTM. I left some nitpicks

PR #4334, opened 2021-04-29
Update release workflow to v3

### LINKED ISSUE #4294: Update release workflow to v3
https://jira.sonarsource.com/browse/BUILD-909?focusedCommentId=246823&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-246823

costin-zaharia-sonarsource:
Fixes #4294

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -271,6 +271,12 @@ stages:
         inputs:
           secureFile: 'maven-settings.xml'
 
+      - task: DownloadSecureFile@1
+        displayName: 'Download the sign key'
+        name: signKey
+        inputs:
+          secureFile: 'sign-key.asc'
+
       - template: update-maven-version-steps.yml
         parameters:
           mavenSettingsFilePath: $(mavenSettings.secureFilePath)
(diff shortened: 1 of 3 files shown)


andrei-epure-sonarsource:
FYI I am updating the parent pom in #4328

pom.xml
```diff
@@ -78,5 +78,30 @@
         <maven.compiler.release>8</maven.compiler.release>
       </properties>
     </profile>
+    <profile>
```
andrei-epure-sonarsource, line 81: @costin-zaharia-sonarsource is this really needed if the parent pom is updated?
I looked at e,g, sonar-php whcih got released, it doesn't have this

andrei-epure-sonarsource: this is already in the parent pom https://github.com/SonarSource/parent-oss/blob/e7e21714eee9a861d2371b174f7ab804f7a79f06/pom.xml#L577

Please rebase on master and remove this

costin-zaharia-sonarsource: The agent runs as a `NetworkService` and for this user we don't have a default `.m2` folder from where the script can take the signing key.

In order to fix this we need to override the profile and set the path to the key file.

azure-pipelines.yml
```diff
@@ -280,10 +286,11 @@ stages:
         env:
           ARTIFACTORY_DEPLOY_USERNAME: $(ARTIFACTORY_DEPLOY_USERNAME)
           ARTIFACTORY_DEPLOY_PASSWORD: $(ARTIFACTORY_DEPLOY_PASSWORD)
+          PGP_PASSPHRASE: $(PGP_PASSPHRASE)
```
costin-zaharia-sonarsource, line 289: This is not propagated by default. We have to copy it by hand (similar with `ARTIFACTORY_DEPLOY_USERNAME` and `ARTIFACTORY_DEPLOY_PASSWORD`).

tomverin (approved):

PR #4335, opened 2021-04-29
Fix S4462 FP when using Task.WhenAll

### LINKED ISSUE #3452: False positive on S4462 when using WhenAll
### Description

To get a better performance when reading several web services, I kick off several tasks. I use the WhenAll function to wait until all tasks are completed. Then I take the result out of each task.

The Sonar marks the read of the result as a S4462 violation which is a false positive; it is safe to read the contents of the task because each task is completed.

### Repro steps

```
using System;
using System.Threading.Tasks;

namespace S4462_reproduce_step
{
    public static class Program
    {
        static async Task Main(string[] args)
        {
            var getValue1Task = GetValue1Async();
            var getValue2Task = GetValue2Async();

            await Task.WhenAll(getValue1Task, getValue2Task).ConfigureAwait(false);

            var result1 = getValue1Task.Result;
            var result2 = getValue2Task.Result;

            Console.WriteLine($"Result of task 1: {result1}, result of task 2: {result2}.");
        }

        public static Task<int> GetValue1Async()
        {
            return Task.FromResult(1);
        }
        public static Task<int> GetValue2Async()
        {
            return Task.FromResult(2);
        }

    }
}
```

### Expected behavior

No Sonar S4462 violation (blocker)

### Actual behavior

Sonar S4462 violation (blocker)

### Known workarounds

Suppress the warning, but that should not be done, because of a false positive.

### Related information

* C#
* Visual Studio Version: 2019 Enterprise
* MSBuild / dotnet version: .NET Core 3.1
* Sonar Qube 6.7 (build 33306)
* Operating System: Win10 Pro 1909

dmitry-golovinov-sonarsource:
Fixes #3452

### DIFF
--- a/analyzers/src/SonarAnalyzer.Shared/Extensions/StatementSyntaxExtensions.cs
+++ b/analyzers/src/SonarAnalyzer.Shared/Extensions/StatementSyntaxExtensions.cs
@@ -29,7 +29,7 @@ using Microsoft.CodeAnalysis.VisualBasic.Syntax;
 
 namespace SonarAnalyzer.Extensions
 {
-    internal static class StatementSyntaxEstensions
+    internal static class StatementSyntaxExtensions
     {
         /// <summary>
         /// Returns all statements before the specified statement within the containing method.
--- a/analyzers/src/SonarAnalyzer.Shared/Extensions/StatementSyntaxExtensions.cs
+++ b/analyzers/src/SonarAnalyzer.Shared/Extensions/StatementSyntaxExtensions.cs
@@ -1,0 +1,47 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2021 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+using System.Linq;
+
+#if CS
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+#else
+using Microsoft.CodeAnalysis.VisualBasic.Syntax;
+#endif
+
+namespace SonarAnalyzer.Extensions
+{
+    internal static class StatementSyntaxEstensions
+    {
+        /// <summary>
+        /// Returns all statements before the specified statement within the containing method.
+        /// This method recursively traverses all parent blocks of the provided statement.
+        /// </summary>
+        public static IEnumerable<StatementSyntax> GetPreviousStatements(this StatementSyntax statement)
+        {
+            var previousStatements = statement.GetPreviousStatementsCurrentBlock();
+
+            return statement.Parent is StatementSyntax parentStatement
+                ? previousStatements.Union(GetPreviousStatements(parentStatement))
+                : previousStatements;
+        }
+    }
+}
--- a/analyzers/src/SonarAnalyzer.Shared/Extensions/SyntaxNodeExtensions.cs
+++ b/analyzers/src/SonarAnalyzer.Shared/Extensions/SyntaxNodeExtensions.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.Collections.Generic;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using SonarAnalyzer.Helpers;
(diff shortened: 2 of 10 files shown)


analyzers/src/SonarAnalyzer.Shared/Extensions/StatementSyntaxExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+
+#if CS
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+#else
+using Microsoft.CodeAnalysis.VisualBasic.Syntax;
+#endif
+
+namespace SonarAnalyzer.Extensions
+{
+    internal static class StatementSyntaxEstensions
```
costin-zaharia-sonarsource, line 32: typo: `StatementSyntaxExtensions`

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -161,14 +161,66 @@ public override string ToString()
...
+        static async Task AwaitAfterAccess(string[] args)
+        {
+            var getValue1Task = Task.FromResult(1);
+            var getValue2Task = Task.FromResult(2);
+            var result1 = getValue1Task.Result;  // Noncompliant
+//                        ^^^^^^^^^^^^^^^^^^^^
+            var result2 = getValue2Task.Result;  // Noncompliant
+//                        ^^^^^^^^^^^^^^^^^^^^
+
+            await Task.WhenAll(getValue1Task, getValue2Task).ConfigureAwait(false);
```
costin-zaharia-sonarsource, line 218: Can we add also a test case with a FN due to branches in code?
```
if ()
{
 await Task.WhenAll(getValue1Task, getValue2Task).ConfigureAwait(false);
}
else
{
  var result2 = getValue2Task.Result; // FN
}
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -161,14 +161,66 @@ public override string ToString()
...
+        }
+
+        public class TaskLike
```
costin-zaharia-sonarsource, line 221: There are a few more methods which will wait for task execution:
- [Wait](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.wait?view=net-5.0)
- [WaitAll](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.waitall?view=net-5.0)
- [RunSynchronously](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.runsynchronously?view=net-5.0)

Can we add some testcases to document their behavior when accessing the result?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -161,14 +161,66 @@ public override string ToString()
...
+//                        ^^^^^^^^^^^^^^^^^^^^
+            var result2 = getValue2Task.Result;  // Noncompliant
+//                        ^^^^^^^^^^^^^^^^^^^^
+        }
+
+        static async Task AwaitAfterAccess(string[] args)
+        {
+            var getValue1Task = Task.FromResult(1);
+            var getValue2Task = Task.FromResult(2);
+            var result1 = getValue1Task.Result;  // Noncompliant
```
costin-zaharia-sonarsource, line 213: I think this is actually a false positive since the `Result` is already there `Task.FromResult` returns an already completed task. See: https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.fromresult?view=net-5.0

costin-zaharia-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CallToAsyncMethodShouldNotBeBlocking.cs
dmitry-golovinov-sonarsource: Actually, we are able to correctly detect this case. Added test to prove that.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CallToAsyncMethodShouldNotBeBlocking.cs
dmitry-golovinov-sonarsource: Turned it to FP, as discussed.

analyzers/src/SonarAnalyzer.Shared/Extensions/SyntaxNodeExtensions.cs
```diff
@@ -43,5 +44,13 @@ public static bool ContainsGetOrSetOnDependencyProperty(this SyntaxNode syntaxNo
                    .Where(invocation => invocation.Expression.NameIs("GetValue") || invocation.Expression.NameIs("SetValue"))
                    .Any(invocation => semanticModel.GetSymbolInfo(invocation).Symbol.ContainingType.DerivesFrom(KnownType.System_Windows_DependencyObject));
         }
+
+        public static IEnumerable<StatementSyntax> GetPreviousStatementsCurrentBlock(this SyntaxNode expression)
+        {
+            var statement = expression.FirstAncestorOrSelf<StatementSyntax>();
```
costin-zaharia-sonarsource, line 50: I think we should add a test for the uncovered branch. We can add `SyntaxNodeExtensionsTest` similar with [SyntaxTokenExtensionsTests](https://github.com/SonarSource/sonar-dotnet/blob/1711a5f773013925e079d11e32bf0a3f592a6706/analyzers/tests/SonarAnalyzer.UnitTest/Extensions/SyntaxTokenExtensionsTests.cs).

costin-zaharia-sonarsource (approved):
LGTM! Would be great though to add the test I've mentioned.

analyzers/src/SonarAnalyzer.Shared/Extensions/SyntaxNodeExtensions.cs
dmitry-golovinov-sonarsource: Tests were added

PR #4341, opened 2021-04-30
Fix S3876 FP System.Index and System.Range are valid indices

### LINKED ISSUE #4236: S3876 - should consider `System.Index` and `System.Range` as valid indeces
https://jira.sonarsource.com/browse/RSPEC-3876

See [SC repro](https://sonarcloud.io/project/issues?id=ranges&open=AXjMHwcmLiYmPkS7za6M&resolved=false&types=CODE_SMELL).

```
public class Foo
{
	int[] bar = new int[] { 1, 2, 3, 4, 5 };
	public int[] this[Range range] => bar[range]; // Noncompliant FP {{Use a string or integral type here, or refactor this into a method. }}
}

public class Bar
{
	int[] bar = new int[] { 1, 2, 3, 4, 5 };
	public int this[Index index] => bar[index]; // Noncompliant FP {{Use a string or integral type here, or refactor this into a method. }}
}
```

Note: RSPEC-3876 is not part of the default profile SonarWay.

dmitry-golovinov-sonarsource:
Fixes #4236

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/StringOrIntegralTypesForIndexers.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/StringOrIntegralTypesForIndexers.cs
@@ -34,10 +34,10 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class StringOrIntegralTypesForIndexers : SonarDiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S3876";
-        private const string MessageFormat = "Use string or an integral type here, or refactor this indexer into a method.";
+        private const string MessageFormat = "Use string, integral, index or a range type here, or refactor this indexer into a method.";
 
         private static readonly ImmutableArray<KnownType> allowedIndexerTypes =
-            new[] { KnownType.System_Object, KnownType.System_String }
+            new[] { KnownType.System_Object, KnownType.System_String, KnownType.System_Index, KnownType.System_Range }
             .Concat(KnownType.IntegralNumbers)
             .ToImmutableArray();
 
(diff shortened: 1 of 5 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/StringOrIntegralTypesForIndexers.cs
```diff
@@ -34,10 +34,10 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class StringOrIntegralTypesForIndexers : SonarDiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S3876";
-        private const string MessageFormat = "Use string or an integral type here, or refactor this indexer into a method.";
+        private const string MessageFormat = "Use string, integral, index or a range type here, or refactor this indexer into a method.";
```
csaba-sagi-sonarsource, line 37: If we change the message format in code we have to update it it on JIRA too.

csaba-sagi-sonarsource:
In general we tend to clean up a file if we touch it, so please fix the two stylecop issues in StringOrIntegralTypesForIndexers. `allowedIndexerTypes` and `rule`. It would be also nice to change IsgenericTypeParameter and Initialize methods to be expression bodied methods. And last but not least align the if on line 61 with our coding style.

csaba-sagi-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.CSharp/Rules/StringOrIntegralTypesForIndexers.cs
csaba-sagi-sonarsource: Maybe I would remove a from a range type? And  go with USe string, integral, index or range type here, ... WDYT?

PR #4342, opened 2021-04-30
S2053 FP fix and message improvement

### LINKED ISSUE #4289: Rule S2053: issue message is not correct when salt is too short
In [smsportal-msbuild14](https://peach.sonarsource.com/issues?projects=smsportal-msbuild14&resolved=false&rules=csharpsquid%3AS2053) we raise 2 TPs but the fix suggested by the issue message is incorrect:

>Make this salt longer.

Making the salt longer will not fix the issue as the main problem is that the salt is hardcoded ans therefore predictable.
We should only advice to use a longer salt in the case where the size is byte array size is shorter that 16 *and* it's populated from a cryptographically secure random function.

Also the message has been updated the to "* Make this salt at least 16 bytes.".

```csharp
var shortSalt = new byte[8];
RandomNumberGenerator.Create().GetBytes(shortSalt);
var pdbShortSalt = new PasswordDeriveBytes(passwordString, shortSalt);          // Noncompliant: Make this salt at least 16 bytes

var pdbPredictableSalt = new PasswordDeriveBytes(passwordString, new byte[8]);          // Noncompliant: Make this salt unpredictable.
```

### LINKED ISSUE #4290: Rule S2053: FP when salt is greater or equal to 16
In [eventstore](https://peach.sonarsource.com/issues?open=AXZuEvXsxqXer6FNyCKn&projects=eventstore&resolved=false&rules=csharpsquid%3AS2053) there is a FN:

```csharp
using System.Security.Cryptography;
using EventStore.Core.Authentication;

namespace EventStore.Core.Services.Transport.Http.Authentication
{
    public class Rfc2898PasswordHashAlgorithm : PasswordHashAlgorithm
    {
        private const int HashSize = 20;
        private const int SaltSize = 16;

        public override void Hash(string password, out string hash, out string salt)
        {
            var salt_ = new byte[SaltSize];
            var randomProvider = new RNGCryptoServiceProvider();
            randomProvider.GetBytes(salt_);
            var hash_ = new Rfc2898DeriveBytes(password, salt_).GetBytes(HashSize); // Compliant: Salt is random and of length 16
        }
    }
}
```

No issue should be raised because the salt is made of 16 random bytes.

csaba-sagi-sonarsource:
Fixes #4290
Fixes #4289

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/HashesShouldHaveUnpredictableSalt.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/HashesShouldHaveUnpredictableSalt.cs
@@ -84,7 +84,7 @@ namespace Tests.Diagnostics
 
             var shortSalt = new byte[15];
             rng.GetBytes(shortSalt);
-            var pdb3 = new PasswordDeriveBytes(passwordBytes, shortSalt); // Noncompliant {{Make this salt longer.}}
+            var pdb3 = new PasswordDeriveBytes(passwordBytes, shortSalt); // Noncompliant {{Make this salt at least 16 bytes.}}
             var pbkdf3 = new Rfc2898DeriveBytes(passwordString, shortSalt); // Noncompliant
         }
 
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/HashesShouldHaveUnpredictableSalt.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/HashesShouldHaveUnpredictableSalt.cs
@@ -12,7 +12,7 @@ namespace Tests.Diagnostics
 
         public void ShortSaltIsNotCompliant()
         {
-            var shortSalt = new byte[31];
+            var shortSalt = new byte[15];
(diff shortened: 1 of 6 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/HashesShouldHaveUnpredictableSalt.cs
```diff
@@ -182,26 +182,26 @@ public void ByteArrayCases(byte[] passwordBytes)
             new PasswordDeriveBytes(passwordBytes, a5); // Compliant
         }
 
-        private byte[] GetSalt() => new byte[16];
+        private byte[] GetSalt() => new byte[1];
```
pavel-mikula-sonarsource, line 185: I'm not sure why 1 as it's bound to compliant cases? It's SE rule and cross-procedural case. Let's return `null;` instead to demonstrate that we don't care.

If we need a number, let's use 8.

pavel-mikula-sonarsource (approved):
LGTM. I left one comment about semantic of a test case

PR #4343, opened 2021-04-30
Configure dependabot.yml

pavel-mikula-sonarsource:
GitHub Dependabot is adding PRs to our repo once upon a time, like #4317

By default, it creates `dependency` label and adds it to the PR. This should prevent the label from appearing.

Changes are based on [docs for dependabot](https://docs.github.com/en/code-security/supply-chain-security/configuration-options-for-dependency-updates) and [labels section](https://docs.github.com/en/code-security/supply-chain-security/configuration-options-for-dependency-updates#labels)

### DIFF
--- a/.github/GitHub.shproj
+++ b/.github/GitHub.shproj
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- This is a minimal Shared project to show directory structure in Solution Explorer. This project should not be build nor referenced. -->
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.CSharp.targets" />
+  <ItemGroup>
+    <None Include="**\*" />
+  </ItemGroup>
+</Project>
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+# https://docs.github.com/en/code-security/supply-chain-security/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels: [ ]     # Do not add default "dependency" label
+
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels: [ ]     # Do not add default "dependency" label
+
+  - package-ecosystem: "nuget"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels: [ ]     # Do not add default "dependency" label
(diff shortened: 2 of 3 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
+# https://docs.github.com/en/code-security/supply-chain-security/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels: [ ]     # Do not add default "dependency" label
```
costin-zaharia-sonarsource, line 10: Can we add the `tooling` label?

.github/GitHub.shproj
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
```
costin-zaharia-sonarsource, line 1: This seems to be a misuse of the shared project. We can link just the yml file. Or add a pattern in the *.sln file to include in Common solution items. WDYT?

costin-zaharia-sonarsource (changes requested):

.github/dependabot.yml
pavel-mikula-sonarsource: Yes we can, but updating dependencies for security reasons isn't tooling. More of an `improvement`, but in general, we don't put labels on standalone PRs

pavel-mikula-sonarsource: Ah, this is in github action dependencies (not visible in UI 🙂 ), so yes, we could, be we're not putting labels on standalone PRs so we can be without it

.github/GitHub.shproj
pavel-mikula-sonarsource: I admit it's little bit of cheating, but it doesn't get build nor referenced. I haven't found a way how to achieve this in other cost-less way - I want to see entire content of `.github` directory without a need to update SLN file everytime something changes to keep maintenance cost low.

costin-zaharia-sonarsource: Adding a pattern there does not help?

pavel-mikula-sonarsource: I couldn't find any docs for this format in SLN file. This [SO answer](https://stackoverflow.com/a/47637436/1486185) claims it as `relative path to file = relative path to file`. I don't have a guess how to inject pattern there.
```
	ProjectSection(SolutionItems) = preProject
		..\.editorconfig = ..\.editorconfig
		src\AssemblyInfo.Shared.cs = src\AssemblyInfo.Shared.cs
		..\azure-pipelines-external.yml = ..\azure-pipelines-external.yml
		..\azure-pipelines.yml = ..\azure-pipelines.yml
		..\scripts\build\build-utils.ps1 = ..\scripts\build\build-utils.ps1
		..\scripts\build\dev-build.ps1 = ..\scripts\build\dev-build.ps1
		Directory.Build.props = Directory.Build.props
		NuGet.Config = NuGet.Config
		src\Signing.cs = src\Signing.cs
		src\SonarAnalyzer.ruleset = src\SonarAnalyzer.ruleset
		tests\SonarAnalyzer.Test.ruleset = tests\SonarAnalyzer.Test.ruleset
		StyleCopAnalyzers.targets = StyleCopAnalyzers.targets
		..\scripts\utils.ps1 = ..\scripts\utils.ps1
	EndProjectSection
```

costin-zaharia-sonarsource (approved):
LGTM!

PR #4345, opened 2021-04-30
Fix S3900 FP: should not raise when argument is annotated with [FromServices] attribute

### LINKED ISSUE #4122: False Positive: [FromServices]-Attribute produces csharpsquid:S3900 (Arguments of public methods should be validated against null)

dmitry-golovinov-sonarsource:
Fixes #4122

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/PublicMethodArgumentsShouldBeCheckedForNull.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/PublicMethodArgumentsShouldBeCheckedForNull.cs
@@ -52,7 +52,8 @@ namespace SonarAnalyzer.Rules.CSharp
         {
             if (args.Symbol is IParameterSymbol
                 && !semanticModel.IsExtensionMethod(args.Identifier.Parent)
-                && !args.Symbol.HasConstraint(ObjectConstraint.NotNull, args.ProgramState))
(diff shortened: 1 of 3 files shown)


costin-zaharia-sonarsource (approved):
LGTM!

PR #4354, opened 2021-05-03
Update the shim layer

### LINKED ISSUE #4348: Update the shim layer
We need this in order to be able to support C# 9.

costin-zaharia-sonarsource:
Fixes #4348


analyzers/src/SonarAnalyzer.ShimLayer.CodeGeneration/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
```
csaba-sagi-sonarsource, line 1: Should we add the shared assembly to the project too?

costin-zaharia-sonarsource: This is not needed since this assembly will not be shipped. It is used only during the project compilation.

analyzers/src/SonarAnalyzer.CFG/ShimLayer/IOperationWrapper.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace StyleCop.Analyzers.Lightup
+{
+    using Microsoft.CodeAnalysis;
+
+    public interface IOperationWrapper
+    {
+        IOperation? WrappedOperation { get; }
+
+        ////IOperationWrapper Parent { get; }
```
csaba-sagi-sonarsource, line 14: In general I saw a lot of commented code. Should we remove it?

analyzers/src/SonarAnalyzer.CFG/ShimLayer/INamedTypeSymbolExtensions.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
-// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
```
csaba-sagi-sonarsource, line 2: Can we check if we should change this to MIT or not?

analyzers/global.json
```diff
@@ -0,0 +1,7 @@
+{
```
csaba-sagi-sonarsource, line 1: Is this really necessary?

analyzers/src/SonarAnalyzer.CFG/ShimLayer/IOperationWrapper.cs
costin-zaharia-sonarsource: I think we should keep the manual changes on this code as few as possible. Since we don't actually maintain the it the commented code should not be a problem.

analyzers/global.json
costin-zaharia-sonarsource: I've removed the file

analyzers/src/SonarAnalyzer.CFG/ShimLayer/INamedTypeSymbolExtensions.cs
costin-zaharia-sonarsource: As far as I understand we need to update the header with license it since it's part of the file.

I've added back the ReadMe file which I deleted by mistake and I've also included the license file in the shim layer folder.

csaba-sagi-sonarsource: Should we also add the license to the CodeGeneration proj?

csaba-sagi-sonarsource (approved):
LGTM

PR #4355, opened 2021-05-03
S3903: DeclareTypesInNamespaces - Records support

csaba-sagi-sonarsource:


analyzers/src/SonarAnalyzer.Common/Rules/DeclareTypesInNamespacesBase.cs
```diff
@@ -18,34 +18,45 @@
...
+            rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, Language.RspecResources);
+
+        protected override void Initialize(SonarAnalysisContext context) =>
+          context.RegisterSyntaxNodeActionInNonGenerated(Language.GeneratedCodeRecognizer,
+              c =>
+              {
+                  var declaration = c.Node;
 
-                var identifier = GetTypeIdentifier(declaration);
+                  if (IsInnerTypeOrWithinNamespace(declaration, c.SemanticModel) || c.ContainingSymbol.Kind != SymbolKind.NamedType)
```
costin-zaharia-sonarsource, line 52: I think it's faster to do the symbol kind check first since we already have the symbol.

costin-zaharia-sonarsource (approved):
LGTM! Please don't forget to switch the conditions order.

PR #4365, opened 2021-05-03
S4136: MethodOverloadsShouldBeGrouped- Records support

csaba-sagi-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -18,21 +18,31 @@
...
 {
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     [Rule(DiagnosticId)]
-    public sealed class MethodOverloadsShouldBeGrouped : MethodOverloadsShouldBeGroupedBase<MemberDeclarationSyntax>
+    public sealed class MethodOverloadsShouldBeGrouped : MethodOverloadsShouldBeGroupedBase<SyntaxKind, MemberDeclarationSyntax>
     {
-        public MethodOverloadsShouldBeGrouped() : base(RspecStrings.ResourceManager) { }
+        protected override ILanguageFacade<SyntaxKind> Language => CSharpFacade.Instance;
+
+        protected override SyntaxKind[] SyntaxKinds { get; } = new[]
```
costin-zaharia-sonarsource, line 39: pedantic: `new []` is redundant

analyzers/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -51,28 +61,13 @@ protected override MemberInfo CreateMemberInfo(SyntaxNodeAnalysisContext c, Memb
...
-                return methodDeclaration.ExplicitInterfaceSpecifier == null;
-            }
-            return true;
-        }
+        protected override IEnumerable<MemberDeclarationSyntax> GetMemberDeclarations(SyntaxNode node) =>
+            ((TypeDeclarationSyntax)node).Members;
 
-        private static bool IsStatic(BaseMethodDeclarationSyntax declaration) => declaration.Modifiers.Any(x => x.Kind() == SyntaxKind.StaticKeyword);
+        private static bool IsValidMemberForOverload(MemberDeclarationSyntax member) =>
+            !(member is MethodDeclarationSyntax methodDeclaration) || methodDeclaration.ExplicitInterfaceSpecifier == null;
```
costin-zaharia-sonarsource, line 68: I think this is a bit more concise and easier to read:
```suggestion
            (member as MethodDeclarationSyntax)?.ExplicitInterfaceSpecifier == null;
```

analyzers/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -51,28 +61,13 @@ protected override MemberInfo CreateMemberInfo(SyntaxNodeAnalysisContext c, Memb
...
 
+        private static bool IsStatic(BaseMethodDeclarationSyntax declaration) =>
```
costin-zaharia-sonarsource, line 70: I'm pretty sure I've seen a similar check in another rule. Could you please extract an extension method and move it to `BaseMethodDeclarationSyntaxExtensions`?

Edit: there is one usage also on the vb rule, we could try to extract it to the shared project.

analyzers/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -18,49 +18,57 @@
...
-        }
+                if (c.ContainingSymbol.Kind != SymbolKind.NamedType)
+                {
+                    return;
+                }
+
+                foreach (var misplacedOverload in GetMisplacedOverloads(c, GetMemberDeclarations(c.Node)))
+                {
+                    var firstName = misplacedOverload.First().NameSyntax;
+                    var secondaryLocations = misplacedOverload.Skip(1).Select(x => new SecondaryLocation(x.NameSyntax.GetLocation(), "Non-adjacent overload"));
```
costin-zaharia-sonarsource, line 63: The `secondaryLocations` enumerable is iterated twice on report diagnostic call which means the `SecondaryLocation` classes are created twice. Could you please make it an array or a list?

analyzers/src/SonarAnalyzer.VisualBasic/Rules/MethodOverloadsShouldBeGrouped.cs
```diff
@@ -30,9 +31,16 @@ namespace SonarAnalyzer.Rules.VisualBasic
 {
     [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
     [Rule(DiagnosticId)]
-    public sealed class MethodOverloadsShouldBeGrouped : MethodOverloadsShouldBeGroupedBase<StatementSyntax>
+    public sealed class MethodOverloadsShouldBeGrouped : MethodOverloadsShouldBeGroupedBase<SyntaxKind, StatementSyntax>
     {
-        public MethodOverloadsShouldBeGrouped() : base(RspecStrings.ResourceManager) { }
+        protected override ILanguageFacade<SyntaxKind> Language => VisualBasicFacade.Instance;
+
+        protected override SyntaxKind[] SyntaxKinds { get; } = new[]
```
costin-zaharia-sonarsource, line 38: pedantic: `new []` is redundant

analyzers/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
```diff
@@ -160,46 +168,37 @@ symbol switch
...
-            }
+            public bool NameEquals(MemberInfo other) =>
+                NameSyntax.ValueText.Equals(other?.NameSyntax.ValueText, isCaseSensitive ? StringComparison.InvariantCulture : StringComparison.InvariantCultureIgnoreCase);
 
-            public override bool Equals(object obj)
-            {
+            public override bool Equals(object obj) =>
                 // Groups that should be together are defined by accessibility, abstract, static and member name #4136
-                return obj is MemberInfo other
+                obj is MemberInfo other
```
costin-zaharia-sonarsource, line 194: I see that there are 2 uncovered branches. I assume one is when `obj is MemberInfo other` is `false`. Is the other the `accessibility` check? We might be able to easily cover it.

costin-zaharia-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/MethodOverloadsShouldBeGrouped.cs
csaba-sagi-sonarsource: I moved it to the extensions. The SyntaxKinds are different so it can not go into the shared project.

analyzers/src/SonarAnalyzer.Common/Rules/MethodOverloadsShouldBeGroupedBase.cs
csaba-sagi-sonarsource: The other uncovered path is `NameEquals(other) == false`, as NameEquals is already called before on that path. Yet I would not remove that line of code from the equals as it could create some confusion.

costin-zaharia-sonarsource (approved):
LGTM!

PR #4367, opened 2021-05-03
Define &, |, and ! operators for Tracker Conditions

### LINKED ISSUE #4305: Define &, |, and ! operators for Tracker Conditions
``` C#
public readonly struct Condition<TContext> where TContext : BaseContext
{
    public delegate bool Lamda(TContext context);

    private Lamda Lamba { get; }
    private Condition(Lamda lamda) => Lamba = lamda;

    public bool Invoke(TContext context) => !(Lamba is null) && Lamba(context);

    private Condition<TContext> Or(Condition<TContext> other)
    {
        var lambda = Lamba;
        return new Condition<TContext>((context) => lambda(context) || other.Lamba(context));
    }
    private Condition<TContext> And(Condition<TContext> other)
    {
        var lambda = Lamba;
        return new Condition<TContext>((context) => lambda(context) && other.Lamba(context));
    }
    private Condition<TContext> Not()
    {
        var lambda = Lamba;
        return new Condition<TContext>((context) => !lambda(context));
    }

    public static implicit operator Condition<TContext>(Lamda lamda) => new Condition<TContext>(lamda);
    public static Condition<TContext> operator |(Condition<TContext> l, Condition<TContext> r) => l.Or(r);
    public static Condition<TContext> operator &(Condition<TContext> l, Condition<TContext> r) => l.And(r);
    public static Condition<TContext> operator ~(Condition<TContext> l) => l.Not();
}
```

Would allow:

``` C#
using Condition = Condition<BaseContext>;
public class Conditions
{
    public static readonly Condition True = (Condition)(context => true);
    public static readonly Condition False = (Condition)(context => false);

    [Test]
    public void True_and_True_is_True()
    {
        var combined = True & True;
        Assert.That(combined.Invoke(null), Is.True);
    }

    [Test]
    public void True_and_False_is_False()
    {
        var combined = True & False;
        Assert.That(combined.Invoke(null), Is.False);
    }

    [Test]
    public void False_or_False_is_False()
    {
        var combined = False | False;
        Assert.That(combined.Invoke(null), Is.False);
    }

    [Test]
    public void True_or_False_is_True()
    {
        var combined = True | False;
        Assert.That(combined.Invoke(null), Is.True);
    }

    [Test]
    public void Not_True_is_False()
    {
        var negated = ~True;
        Assert.That(negated.Invoke(null), Is.False);
    }
}
```

I think it would increase the readability of complex conditions. Is see to (small) downsides:
1. The compiler allows to define an implicit cast, but gives a CS1660 if you apply it (so it has to be explicit).
2. You can have to use `.Invoke(context)` instead of `(context)`.

What do you think?

Corniel:
Fixes #4305.

- [ ] There are still places left where the delegates have to be converted to a `Condition` instance.

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Helpers/BuilderPatternDescriptor.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Helpers/BuilderPatternDescriptor.cs
@@ -40,7 +40,7 @@ namespace SonarAnalyzer.Helpers
         }
 
         public bool IsMatch(InvocationContext context) =>
-            invocationConditions.All(x => x(context));
+            invocationConditions.All(x => x.Invoke(context));
 
         public bool IsValid(TInvocationSyntax invocation) =>
             isValid(invocation);
--- a/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CommandPathBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/CommandPathBase.cs
@@ -41,18 +41,18 @@ namespace SonarAnalyzer.Rules
             var inv = Language.Tracker.Invocation;
             inv.Track(input,
                 inv.MatchMethod(new MemberDescriptor(KnownType.System_Diagnostics_Process, "Start")),
-                c => IsInvalid(FirstArgument(c)));
+                new Condition(c => IsInvalid(FirstArgument(c))));
 
             var pa = Language.Tracker.PropertyAccess;
             pa.Track(input,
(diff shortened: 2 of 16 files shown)


pavel-mikula-sonarsource:
As briefly discussed, the `new Condition` is all over the place, unfortunately. And we'd need `new TrackerBase<TSyntaxKind, InvocationContext>.Condition` in `CommandPathBase` and other custom places.

Let's abandon this idea, as the compiler support is not good enough to bring value.

Corniel:
Too bad, but agreed.

Issue #4370, opened 2021-05-04
S927 FP 


dmitry-golovinov-sonarsource:
### Description

False positive for S927. When the type of the argument is generic interface it will be more useful to provide name based on the actual substituted type, since it can better express what the code is about

### Repro steps

```
    public interface SomeInterface<A> { }

    public interface BaseInterface<A>
    {
        void Apply(SomeInterface<A> param);
    }

    public class BasicImplementation : BaseInterface<int>
    {
        public void Apply(SomeInterface<int> intValue) { }   //Noncompliant FP
    }
```

### Expected behavior

Do not raise the rule.

### Actual behavior

In the previous example issue is being raised, since type kind of parameter **param** in **Apply** method of **BaseInterface**  
does not have the kind of **TypeParameter**.

### Related information

* SonarC# Version: 8.22
* Visual Studio 16.9.4

PR #4371, opened 2021-05-04
Documenting S927 FP found on Peach

dmitry-golovinov-sonarsource:
Providing example for #4370


csaba-sagi-sonarsource (approved):
LGTM. It would be nice to add the reproducer to the VB test cases too.

PR #4372, opened 2021-05-04
Rule S110: Add support for records

costin-zaharia-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveInheritance.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveInheritance.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
@@ -28,6 +29,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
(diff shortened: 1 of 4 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveInheritance.cs
```diff
@@ -36,89 +38,108 @@ namespace SonarAnalyzer.Rules.CSharp
...
+
+        private class ObjectTypeInfo
+        {
+            public SyntaxToken Identifier { get; }
+
+            public INamedTypeSymbol Symbol { get; }
+
+            public string Name { get; }
+
+            public ObjectTypeInfo(SyntaxNode node, SemanticModel model)
```
dmitry-golovinov-sonarsource, line 127: I am personally not a fan of creating such constructors that incorporate such kind of logic. If we will move this to some common space, may be lets create a factory?

analyzers/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveInheritance.cs
```diff
@@ -36,89 +38,108 @@ namespace SonarAnalyzer.Rules.CSharp
     public class AvoidExcessiveInheritance : ParameterLoadingDiagnosticAnalyzer
     {
         private const string DiagnosticId = "S110";
-        private const string MessageFormat = "This class has {0} parents which is greater than {1} authorized.";
+        private const string MessageFormat = "This {0} has {1} parents which is greater than {2} authorized.";
```
dmitry-golovinov-sonarsource, line 41: Not quite sure, but probably makes sense to state the change of the message here https://jira.sonarsource.com/browse/RSPEC-3914

dmitry-golovinov-sonarsource (commented):
Please, check my comments.

analyzers/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveInheritance.cs
```diff
@@ -36,89 +38,108 @@ namespace SonarAnalyzer.Rules.CSharp
...
         }
 
-        private static Regex WilcardPatternToRegularExpression(string pattern)
+        private static Regex WildcardPatternToRegularExpression(string pattern)
         {
             var regexPattern = string.Concat("^", Regex.Escape(pattern).Replace("\\*", ".*"), "$");
             return new Regex(regexPattern, RegexOptions.Compiled);
         }
+
+        private class ObjectTypeInfo
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: I would postpone this until actually needed. We will be able to take a better decision then.

analyzers/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveInheritance.cs
```diff
@@ -36,89 +38,108 @@ namespace SonarAnalyzer.Rules.CSharp
...
 
-                    var thisTypeRootNamespace = GetRootNamespace(symbol);
+                    var thisTypeRootNamespace = GetRootNamespace(objectTypeInfo.Symbol);
 
-                    var baseTypesCount = symbol.GetSelfAndBaseTypes()
+                    var baseTypesCount = objectTypeInfo.Symbol.GetSelfAndBaseTypes()
                         .Skip(1) // remove the class itself
                         .TakeWhile(s => GetRootNamespace(s) == thisTypeRootNamespace)
                         .Select(nts => nts.OriginalDefinition.ToDisplayString())
-                        .TakeWhile(className => this.filteredClassesRegex.All(regex => !regex.IsMatch(className)))
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: I will not change the property name since it's already used and this will be a breaking change for the users. This is not that bad since the Records are in the end immutable classes.

analyzers/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveInheritance.cs
costin-zaharia-sonarsource: I want to keep things as simple as possible. I think we can create the factory when the need arises.

analyzers/src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveInheritance.cs
```diff
@@ -42,21 +42,21 @@ public class AvoidExcessiveInheritance : ParameterLoadingDiagnosticAnalyzer
...
+        private static readonly DiagnosticDescriptor Rule =
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager, false);
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
         [RuleParameter("max", PropertyType.Integer, "Maximum depth of the inheritance tree. (Number)", MaximumDepthDefaultValue)]
         public int MaximumDepth { get; set; } = MaximumDepthDefaultValue;
 
-        private ICollection<Regex> filteredClassesRegex = new List<Regex>();
+        private ICollection<Regex> filters = new List<Regex>();
```
dmitry-golovinov-sonarsource, line 52: On lines 52 and 54 we have 2 fields that are mixed with property definitions. Let's put fields before properties.

dmitry-golovinov-sonarsource (approved):
Small comment that you might want to address before merging PR. Everything else looks good.

PR #4373, opened 2021-05-04
Rule S1121: Add support for records

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/AssignmentInsideSubExpression.cs
```diff
@@ -72,11 +72,15 @@ public sealed class AssignmentInsideSubExpression : SonarDiagnosticAnalyzer
 
         private static bool IsNonCompliantSubExpression(AssignmentExpressionSyntax assignment, ExpressionSyntax topParenthesizedExpression) =>
             IsInsideExpression(topParenthesizedExpression)
+            && !IsInWithInitializerExpression(topParenthesizedExpression)
             && !IsCompliantAssignmentInsideExpression(assignment, topParenthesizedExpression);
 
         private static bool IsInsideExpression(ExpressionSyntax expression) =>
             expression.Parent.FirstAncestorOrSelf<ExpressionSyntax>() != null;
 
+        private static bool IsInWithInitializerExpression(ExpressionSyntax expression) =>
```
csaba-sagi-sonarsource, line 81: With this implementation we have inconsistency as with ObjectInitializerExpression we have FNs and with WIthInitializerExpression we raise as expected. Also rename this method and use IsAnyKind with both InitializerExpressions. In addition if we remove ObjectInitializerExpression from the AllowedParentKinds we should be able to cover the following case: `var v7 = new MyClass { MyField = i = 42 }; // Noncompliant`

csaba-sagi-sonarsource (changes requested):
The record part looks good, and we can improve the rule more. Also please take a look at our [coding style](https://github.com/SonarSource/sonar-dotnet/blob/master/docs/coding-style.md)  and align the member orders as described there.

csaba-sagi-sonarsource (approved):
LGTM!

PR #4374, opened 2021-05-05
Rule S3168: Add support for records

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/AsyncVoidMethod.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/AsyncVoidMethod.cs
@@ -75,7 +75,8 @@ namespace SonarAnalyzer.Rules.CSharp
             || HasAnyMsTestV1AllowedAttribute(methodSymbol);
 
         private static bool IsUsedAsEventHandler(MethodDeclarationSyntax methodDeclaration) =>
-            methodDeclaration.FirstAncestorOrSelf((SyntaxNode x) => x.IsKind(SyntaxKind.ClassDeclaration) || x.IsKind(SyntaxKindEx.RecordDeclaration)) is { } parentDeclaration
+            methodDeclaration.FirstAncestorOrSelf<TypeDeclarationSyntax>(x =>
+                x.IsAnyKind(SyntaxKind.ClassDeclaration, SyntaxKindEx.RecordDeclaration, SyntaxKind.InterfaceDeclaration)) is { } parentDeclaration
             && parentDeclaration.DescendantNodes()
                 .OfType<AssignmentExpressionSyntax>()
                 .Where(x => x.IsKind(SyntaxKind.AddAssignmentExpression))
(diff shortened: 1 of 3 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/AsyncVoidMethod.cs
```diff
@@ -77,22 +75,22 @@ public sealed class AsyncVoidMethod : SonarDiagnosticAnalyzer
             || HasAnyMsTestV1AllowedAttribute(methodSymbol);
 
         private static bool IsUsedAsEventHandler(MethodDeclarationSyntax methodDeclaration) =>
-            methodDeclaration.FirstAncestorOrSelf<ClassDeclarationSyntax>() is ClassDeclarationSyntax parentClass
-            && parentClass.DescendantNodes()
+            methodDeclaration.FirstAncestorOrSelf((SyntaxNode x) => x.IsKind(SyntaxKind.ClassDeclaration) || x.IsKind(SyntaxKindEx.RecordDeclaration)) is { } parentDeclaration
```
csaba-sagi-sonarsource, line 78:
```suggestion
            methodDeclaration.FirstAncestorOrSelf<TypeDeclarationSyntax>(x => x.IsAnyKind(SyntaxKind.ClassDeclaration, SyntaxKindEx.RecordDeclaration)) is { } parentDeclaration
```
I think this can also do the job and I expect it to do a better job as only for TypeDeclarationSyntax will it do the checks.

csaba-sagi-sonarsource (changes requested):
Only one minor change regarding the records support, but lets add test cases for interfaces and structs. Structs should work out of the box, and it would be nice to cover interfaces too if it is not complicated.

analyzers/src/SonarAnalyzer.CSharp/Rules/AsyncVoidMethod.cs
```diff
@@ -77,22 +75,23 @@ public sealed class AsyncVoidMethod : SonarDiagnosticAnalyzer
             || HasAnyMsTestV1AllowedAttribute(methodSymbol);
 
         private static bool IsUsedAsEventHandler(MethodDeclarationSyntax methodDeclaration) =>
-            methodDeclaration.FirstAncestorOrSelf<ClassDeclarationSyntax>() is ClassDeclarationSyntax parentClass
-            && parentClass.DescendantNodes()
+            methodDeclaration.FirstAncestorOrSelf<TypeDeclarationSyntax>(x =>
+                x.IsAnyKind(SyntaxKind.ClassDeclaration, SyntaxKindEx.RecordDeclaration, SyntaxKind.InterfaceDeclaration)) is { } parentDeclaration
```
csaba-sagi-sonarsource, line 79: Thanks to the `SyntaxKind.InterfaceDeclaration` here we now support the following case:

 ```
 public interface ISomeInterface
      {
        event EventHandler<bool> MyEvent;

        public void SomeMethod()
        {
            MyEvent += EventHandlerCases_MyEvent;
        }

        private async void EventHandlerCases_MyEvent(object sender, bool e)
        {
            await Task.Run(() => Console.WriteLine("test"));
        }
    }
```
And it would be nice to also cover it in the UTs.

dmitry-golovinov-sonarsource: Nice one, thank you!

PR #4375, opened 2021-05-05
Rule S4830: Add support for records

dmitry-golovinov-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CertificateValidationCheck.CSharp9.cs
```diff
@@ -28,7 +28,7 @@ record InnerAssignmentRecord
 
             public void InitAsArgument(RemoteCertificateValidationCallback callback)
             {
-                CreateRQ().ServerCertificateValidationCallback += callback; // FN
+                CreateRQ().ServerCertificateValidationCallback += callback;  //Noncompliant
```
csaba-sagi-sonarsource, line 31: Please add test cases with positional records.

csaba-sagi-sonarsource (changes requested):
Your changes look good! Yet before merging let`s try to bump up the coverage a bit if possible.

csaba-sagi-sonarsource (approved):
LGTM

PR #4376, opened 2021-05-05
Rules S100, S101: Add support for records

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/ClassAndMethodName.cs
```diff
@@ -70,10 +71,60 @@ public sealed class ClassAndMethodName : SonarDiagnosticAnalyzer
...
+            {
+                yield return currentWord.ToString();
+            }
+        }
+
         protected override void Initialize(SonarAnalysisContext context)
         {
             context.RegisterSyntaxNodeAction(c =>
                 {
+                    if (c.ContainingSymbol.Kind != SymbolKind.NamedType)
```
csaba-sagi-sonarsource, line 124: Let`s use RegisterSyntaxNodeAction instead of RegisterSyntaxNodeAction so we can remove the second if with ShouldAnalyze.

csaba-sagi-sonarsource (approved):
LGTM

PR #4384, opened 2021-05-05
Rule S1694: Add support for records

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/ClassNotInstantiatable.cs
```diff
@@ -21,28 +21,31 @@
...
             context.RegisterSymbolAction(CheckClassWithOnlyUnusedPrivateConstructors, SymbolKind.NamedType);
-        }
 
-        private static void CheckClassWithOnlyUnusedPrivateConstructors(SymbolAnalysisContext context)
+        protected override bool IsTypeDeclaration(SyntaxNode node) =>
+            node.IsAnyKind(SyntaxKind.ClassDeclaration, SyntaxKindEx.RecordDeclaration);
+
+        private void CheckClassWithOnlyUnusedPrivateConstructors(SymbolAnalysisContext context)
         {
             var namedType = context.Symbol as INamedTypeSymbol;
```
costin-zaharia-sonarsource, line 50: pedantic: we should use here direct cast as we know that it's `INamedTypeSymbol` and we don't check for null after this assignment.

costin-zaharia-sonarsource (approved):
LGTM! Just a minor comment which is up to you.

PR #4390, opened 2021-05-06
Rule S1694: add support for records

costin-zaharia-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ClassShouldNotBeAbstract.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ClassShouldNotBeAbstract.cs
@@ -26,6 +26,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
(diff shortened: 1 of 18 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/ClassShouldNotBeAbstract.cs
```diff
@@ -26,105 +26,113 @@
...
+        protected override void Initialize(SonarAnalysisContext context) =>
             context.RegisterSymbolAction(
                 c =>
                 {
-                    var symbol = c.Symbol as INamedTypeSymbol;
-                    if (!symbol.IsClass() ||
-                        !symbol.IsAbstract ||
-                        ClassHasInheritedAbstractMembers(symbol))
+                    var symbol = (INamedTypeSymbol)c.Symbol;
+                    if (symbol.Kind != SymbolKind.NamedType
```
csaba-sagi-sonarsource, line 52: I think this check is uncovered as it is redundant. Is it possible to get a NamedTypeSymbol which is not of symbolKind namedtype?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ClassShouldNotBeAbstract.CSharp9.cs
```diff
@@ -58,17 +58,22 @@ public void flipSwitch()
...
     {
         public abstract string Content { get; }
         public abstract string Content1 { get; }
         public string Content2 { get; }
     }
 
-    public abstract record Record(string X);
+    public abstract record Record(string X); // Noncompliant {{Convert this 'abstract' record to a concrete type with a protected constructor.}}
+
+    public abstract record Record2(string X) // Compliant, this record has a propery X which is concrete
```
csaba-sagi-sonarsource, line 75: The comment about the concrete X property is a bit confusing, as the  case is Compliant due to the Content abstract property.

csaba-sagi-sonarsource (changes requested):
Overall it looks good. Let`s try to remove that one more line if possible.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ClassShouldNotBeAbstract.CSharp9.cs
costin-zaharia-sonarsource: In this case the X property is relevant, without it we would suggest to use an interface since all methods are abstract.

csaba-sagi-sonarsource: Yes, indeed. I missed that fact. Thanks for pointing it out.

csaba-sagi-sonarsource (approved):
LGTM!

PR #4391, opened 2021-05-06
Rule S4002: add support for records

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/DisposableTypesNeedFinalizers.cs
```diff
@@ -26,56 +26,50 @@
...
             {
-                var classDeclaration = (ClassDeclarationSyntax)c.Node;
-                var classSymbol = c.SemanticModel.GetDeclaredSymbol(classDeclaration);
+                var declaration = (TypeDeclarationSyntax)c.Node;
+                var classSymbol = c.SemanticModel.GetDeclaredSymbol(declaration);
 
-                if (classSymbol.Implements(KnownType.System_IDisposable) &&
-                    HasNativeHandleFields(classDeclaration, c.SemanticModel) &&
-                    !HasFinalizer(classDeclaration))
+                if (classSymbol.Implements(KnownType.System_IDisposable)
```
dmitry-golovinov-sonarsource, line 56: _c.ContainingSymbol.Kind != SymbolKind.NamedType_ check is missing here

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DisposableTypesNeedFinalizers.CSharp9.cs
```diff
@@ -1,7 +1,7 @@
 ﻿using System;
 using System.Runtime.InteropServices;
 
-public record Foo : IDisposable // FN
+public record Foo : IDisposable // Noncompliant {{Implement a finalizer that calls your 'Dispose' method.}}
```
dmitry-golovinov-sonarsource, line 4: Please add test with positional record declaration

dmitry-golovinov-sonarsource (changes requested):
Please, see my comments.

analyzers/src/SonarAnalyzer.CSharp/Rules/DisposableTypesNeedFinalizers.cs
costin-zaharia-sonarsource: Nice find. I've forgot about this.

dmitry-golovinov-sonarsource (approved):
LGTM!

PR #4392, opened 2021-05-07
Rule S4015: add support for records

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotDecreaseMemberVisibility.cs
```diff
@@ -27,30 +27,29 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
-        protected override void Initialize(SonarAnalysisContext context)
-        {
+        protected override void Initialize(SonarAnalysisContext context) =>
             context.RegisterSyntaxNodeActionInNonGenerated(c =>
                 {
-                    var classDeclaration = (ClassDeclarationSyntax)c.Node;
+                    var classDeclaration = (TypeDeclarationSyntax)c.Node;
                     var classSymbol = c.SemanticModel.GetDeclaredSymbol(classDeclaration);
```
csaba-sagi-sonarsource, line 48: Let`s give it a try, but my understanding is that the containingSymbol is actually the DeclaredSymbol for Class and Record declaration syntax nodes. So maybe we can work with that, without retrieving the declared symbol. (Only in case the kind is NamedType.)

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotDecreaseMemberVisibility.cs
```diff
@@ -135,28 +127,30 @@ private Diagnostic FindMethodIssue(MemberDeclarationSyntax memberDeclaration, IM
 
             private Diagnostic FindPropertyIssue(MemberDeclarationSyntax memberDeclaration, IPropertySymbol propertySymbol)
```
csaba-sagi-sonarsource, line 128: FindMethodIssue and FindPopertyIssue methods are very similar, maybe they could be unified?

csaba-sagi-sonarsource (changes requested):
Looks good, one more case left to cover.

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotDecreaseMemberVisibility.cs
costin-zaharia-sonarsource: That's very nice! This was already here so I'm pretty sure that we can make this improvement in other places too.

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotDecreaseMemberVisibility.cs
costin-zaharia-sonarsource: Indeed they look very similar but I would let this exercise for the next time we will touch the class.

csaba-sagi-sonarsource (approved):
LGTM!

PR #4393, opened 2021-05-07
Rule S4019: add support for records

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotHideBaseClassMethods.cs
```diff
@@ -27,91 +27,91 @@
...
                 {
-                    var classDeclaration = (ClassDeclarationSyntax)c.Node;
-                    var classSymbol = c.SemanticModel.GetDeclaredSymbol(classDeclaration);
+                    if (c.ContainingSymbol.Kind != SymbolKind.NamedType)
+                    {
+                        return;
+                    }
 
-                    if (classSymbol == null || classDeclaration.Identifier.IsMissing)
+                    var declarationSyntax = (TypeDeclarationSyntax)c.Node;
```
dmitry-golovinov-sonarsource, line 52: Small suggestion, we can make it to line 47 and do a check there. But probably it won't save that much time, so it's up to you

dmitry-golovinov-sonarsource (approved):
LGTM!

PR #4395, opened 2021-05-07
Rule S3249: add support for records

costin-zaharia-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/GetHashCodeEqualsOverride.CSharp9.cs
```diff
@@ -9,5 +9,5 @@ public record Noncompliant
 {
     private readonly int x;
 
-    public override int GetHashCode() => x.GetHashCode() ^ base.GetHashCode(); //Noncompliant - FP, this rule should not apply for records since Equals and GetHashCode are value-based
```
csaba-sagi-sonarsource, line 12: Please rename the record as it is not noncompliant and also leave there a comment for the future to know why it is compliant.
Also it would be nice to verify that we work with Positional records.

analyzers/src/SonarAnalyzer.CSharp/Rules/GetHashCodeEqualsOverride.cs
```diff
@@ -153,11 +139,8 @@ private static bool IfStatementWithSingleReturnTrue(IfStatementSyntax ifStatemen
                 return false;
             }
 
-            return returnStatement.Expression != null &&
-                CSharpEquivalenceChecker.AreEquivalent(returnStatement.Expression, CSharpSyntaxHelper.TrueLiteralExpression);
+            return returnStatement.Expression != null
+                   && CSharpEquivalenceChecker.AreEquivalent(returnStatement.Expression, CSharpSyntaxHelper.TrueLiteralExpression);
```
csaba-sagi-sonarsource, line 143: It can be changed to the following return statement and than the previous if can be removed.
```suggestion
            return returnStatement != null
                   && returnStatement.Expression != null
                   && CSharpEquivalenceChecker.AreEquivalent(returnStatement.Expression, CSharpSyntaxHelper.TrueLiteralExpression);
```

analyzers/src/SonarAnalyzer.CSharp/Rules/GetHashCodeEqualsOverride.cs
```diff
@@ -85,30 +86,32 @@ protected override void Initialize(SonarAnalysisContext context)
...
                             var location = locations.First(loc => loc.SourceSpan.Start == firstPosition);
-                            c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, location, methodSymbol.Name));
+                            c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, location, methodSymbol.Name));
                         });
                 });
-        }
 
-        private static bool TryGetLocationFromInvocationInsideMethod(SyntaxNodeAnalysisContext context,
-            IMethodSymbol methodSymbol, out Location location)
+        internal static bool IsEqualsCallInGuardCondition(InvocationExpressionSyntax invocation, IMethodSymbol invokedMethod) =>
```
csaba-sagi-sonarsource, line 93: Please fix the code smell. (Reorder members)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/GuardConditionOnEqualsOverride.cs
```diff
@@ -1,7 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-
-namespace Tests.Diagnostics
+﻿namespace Tests.Diagnostics
```
csaba-sagi-sonarsource, line 1: Why aren`t there any  test cases for records for this rule?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/GetHashCodeEqualsOverride.CSharp9.cs
costin-zaharia-sonarsource: For this rule all records are ignored. It does not matter if they are positional or not.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/GuardConditionOnEqualsOverride.cs
csaba-sagi-sonarsource: I have added the rule to the excel sheet

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/GetHashCodeEqualsOverride.CSharp9.cs
csaba-sagi-sonarsource: I know, I just wanted test cases to have a bigger safety net.

csaba-sagi-sonarsource (approved):
LGTM

PR #4403, opened 2021-05-10
Rule S3881: Add support for records

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/its/expected/Net5/Net5--S3881.json
+++ b/analyzers/its/expected/Net5/Net5--S3881.json
@@ -1,0 +1,37 @@
+{
+"issues":  [
+{
+"id":  "S3881",
+"message":  "Fix this implementation of 'IDisposable' to conform to the dispose pattern.",
+"location":  [
+{
+"uri":  "sources\Net5\Net5\Records.cs",
+"region":  {
+"startLine":  101,
+"startColumn":  19,
+"endLine":  101,
+"endColumn":  35
+}
+},
+{
+"uri":  "sources\Net5\Net5\Records.cs",
+"region":  {
+"startLine":  101,
+"startColumn":  19,
+"endLine":  101,
+"endColumn":  35
+}
+},
+{
+"uri":  "sources\Net5\Net5\Records.cs",
+"region":  {
+"startLine":  103,
+"startColumn":  21,
+"endLine":  103,
+"endColumn":  28
+}
+}
+]
+}
+]
+}
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
@@ -37,110 +37,155 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class ImplementIDisposableCorrectly : SonarDiagnosticAnalyzer
     {
-        internal const string DiagnosticId = "S3881";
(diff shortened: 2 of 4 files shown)


analyzers/its/expected/Net5/Net5--S3881.json
```diff
@@ -0,0 +1,37 @@
...
+}
+},
+{
+"uri":  "sources\Net5\Net5\Records.cs",
+"region":  {
+"startLine":  101,
+"startColumn":  19,
+"endLine":  101,
+"endColumn":  35
+}
```
costin-zaharia-sonarsource, line 23: It looks that the issue is reported twice. Can you please double check that we don't have a problem here?

analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
```diff
@@ -37,110 +37,157 @@ namespace SonarAnalyzer.Rules.CSharp
...
+                        checker = new DisposableChecker(classDeclaration.BaseList, declarationIdentifier, c.SemanticModel.GetDeclaredSymbol(classDeclaration), "class", c.SemanticModel);
+                    }
+                    else
+                    {
+                        var recordDeclaration = (RecordDeclarationSyntaxWrapper)c.Node;
+                        declarationIdentifier = recordDeclaration.Identifier;
+                        checker = new DisposableChecker(recordDeclaration.BaseList, declarationIdentifier, c.SemanticModel.GetDeclaredSymbol(recordDeclaration), "record", c.SemanticModel);
+                    }
+
+                    var locations = checker.GetIssueLocations();
```
costin-zaharia-sonarsource, line 76: `locations` is enumerable and it's iterated 3 times. We should update GetIssueLocations to return a List or array.

analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
```diff
@@ -37,110 +37,157 @@ namespace SonarAnalyzer.Rules.CSharp
...
 
-                return this.secondaryLocations;
+                return secondaryLocations;
             }
 
-            private void AddSecondaryLocation(Location location, string message)
-            {
-                this.secondaryLocations.Add(new SecondaryLocation(location, message));
-            }
+            private static bool IsDisposeBool(IMethodSymbol method) =>
```
costin-zaharia-sonarsource, line 158: Please do not move these. We prefer to have the static methods at the end.

analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
```diff
@@ -37,110 +37,119 @@ namespace SonarAnalyzer.Rules.CSharp
...
-        }
+                    if (c.ContainingSymbol.Kind != SymbolKind.NamedType)
+                    {
+                        return;
+                    }
+
+                    var typeDeclarationSyntax = (TypeDeclarationSyntax)c.Node;
+                    var declarationIdentifier = typeDeclarationSyntax.Identifier;
+                    var checker = new DisposableChecker(typeDeclarationSyntax.BaseList,
+                        declarationIdentifier,
```
costin-zaharia-sonarsource, line 64: pedantic: the parameters should be aligned one under the other. Rider has a setting for this which is very helpful.

costin-zaharia-sonarsource (approved):
LGTM! Please see the comment with alignment before merging.

PR #4412, opened 2021-05-11
Rule S3925: Add support for records

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementISerializableCorrectly.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementISerializableCorrectly.cs
@@ -53,23 +53,10 @@ namespace SonarAnalyzer.Rules.CSharp
                         return;
                     }
 
-                    SyntaxToken typeKeyword;
-                    SyntaxToken typeIdentifier;
-                    INamedTypeSymbol typeSymbol;
-                    if (c.Node.IsKind(SyntaxKind.ClassDeclaration))
-                    {
-                        var classDeclaration = (ClassDeclarationSyntax)c.Node;
-                        typeKeyword = classDeclaration.Keyword;
-                        typeIdentifier = classDeclaration.Identifier;
-                        typeSymbol = c.SemanticModel.GetDeclaredSymbol(classDeclaration);
-                    }
-                    else
-                    {
-                        var recordDeclaration = (RecordDeclarationSyntaxWrapper)c.Node;
-                        typeKeyword = recordDeclaration.Keyword;
-                        typeIdentifier = recordDeclaration.Identifier;
-                        typeSymbol = c.SemanticModel.GetDeclaredSymbol(recordDeclaration);
-                    }
+                    var typeDeclarationSyntax = (TypeDeclarationSyntax)c.Node;
+                    var typeKeyword = typeDeclarationSyntax.Keyword;
+                    var typeIdentifier = typeDeclarationSyntax.Identifier;
+                    var typeSymbol = c.SemanticModel.GetDeclaredSymbol(typeDeclarationSyntax);
 
                     if (!ImplementsISerializable(typeSymbol))
                     {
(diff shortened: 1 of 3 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementISerializableCorrectly.cs
```diff
@@ -28,28 +28,35 @@
...
 {
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     [Rule(DiagnosticId)]
     public class ImplementISerializableCorrectly : SonarDiagnosticAnalyzer
     {
-        internal const string DiagnosticId = "S3925";
-        internal const string MessageFormat = "Update this implementation of 'ISerializable' to conform to the " +
-            "recommended serialization pattern.";
+        private const string DiagnosticId = "S3925";
+        private const string MessageFormat = "Update this implementation of 'ISerializable' to conform to the " +
```
csaba-sagi-sonarsource, line 40: Nitpick: This can be single line.

analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementISerializableCorrectly.cs
```diff
@@ -28,28 +28,35 @@
...
-                    var typeSymbol = c.SemanticModel.GetDeclaredSymbol(classDeclaration);
+                    if (c.ContainingSymbol.Kind != SymbolKind.NamedType)
+                    {
+                        return;
+                    }
+
+                    var typeDeclarationSyntax = (TypeDeclarationSyntax)c.Node;
+                    var typeKeyword = typeDeclarationSyntax.Keyword;
+                    var typeIdentifier = typeDeclarationSyntax.Identifier;
+                    var typeSymbol = c.SemanticModel.GetDeclaredSymbol(typeDeclarationSyntax);
```
csaba-sagi-sonarsource, line 59: c.ContainingSymbol should be the same as the symbol returned by GetDeclaredSymbol, so you can use that instead.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ImplementISerializableCorrectly.CSharp9.cs
```diff
@@ -170,33 +170,44 @@ public new void GetObjectData(SerializationInfo info, StreamingContext context)
...
-    public partial record Partial_SerializableDerived_CallingBase_GetObjectData_Record : SerializableRecord
+    public record Serializable_NoConstructor_Positional(string Value) : ISerializable
+//                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Noncompliant {{Update this implementation of 'ISerializable' to conform to the recommended serialization pattern.}}
+//         ^^^^^^ Secondary@-1 {{Add a 'protected' constructor 'Serializable_NoConstructor_Positional(SerializationInfo, StreamingContext)'.}}
+    {
+        public Serializable_NoConstructor_Positional() : this("") { }
+        public virtual void GetObjectData(SerializationInfo info, StreamingContext context) { }
+    }
+
+// See https://github.com/SonarSource/sonar-dotnet/issues/4411 Applicable to all raised issues below
```
csaba-sagi-sonarsource, line 181: Put the cases in a separate namespace. This way if somebody adds a non related case below it can create confusion. And also please add at least one test case with classes.

csaba-sagi-sonarsource (changes requested):
Nice catch of the FP! Let`s just add a few more things and it will be ready to merge! 🎉

analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementISerializableCorrectly.cs
dmitry-golovinov-sonarsource: Good to know!

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ImplementISerializableCorrectly.CSharp9.cs
dmitry-golovinov-sonarsource: Actually, last case is the case with classes, it also doesn't work for the same reason - partial methods are not supported. Added comment in it back and extracted everything in another namespace.

csaba-sagi-sonarsource: I missed the fact that the last was a class. Thank you anyway for putting them into a namespace.

csaba-sagi-sonarsource (approved):
LGTM!

PR #4413, opened 2021-05-11
Rule S4039: add support for records

costin-zaharia-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/InterfaceMethodsShouldBeCallableByChildTypes.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/InterfaceMethodsShouldBeCallableByChildTypes.cs
@@ -71,34 +71,34 @@ namespace SonarAnalyzer.Rules.CSharp
                 return;
             }
 
-            var classDeclaration = memberDeclaration.FirstAncestorOrSelf<ClassDeclarationSyntax>();
-            if (classDeclaration == null
-                || classDeclaration.Identifier.IsMissing
-                || !IsClassTracked(classDeclaration, analysisContext.SemanticModel))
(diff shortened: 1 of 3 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/InterfaceMethodsShouldBeCallableByChildTypes.cs
```diff
@@ -74,65 +71,54 @@ protected override void Initialize(SonarAnalysisContext context)
...
 
-        private static bool IsClassTracked(ClassDeclarationSyntax classDeclaration, SemanticModel semanticModel)
+        private static bool IsDeclarationTracked(BaseTypeDeclarationSyntax declaration, SemanticModel semanticModel)
         {
-            var classSymbol = semanticModel.GetDeclaredSymbol(classDeclaration);
+            var symbol = semanticModel.GetDeclaredSymbol(declaration);
 
-            return classSymbol != null &&
-                !classSymbol.IsSealed &&
-                !classSymbol.IsStatic &&
```
dmitry-golovinov-sonarsource, line 104: This check has disappeared in the resulting code. Is that somewhat expected?

dmitry-golovinov-sonarsource (commented):
Overall looks good to me. I've left one question, could you please provide the answer?

analyzers/src/SonarAnalyzer.CSharp/Rules/InterfaceMethodsShouldBeCallableByChildTypes.cs
costin-zaharia-sonarsource: Yes, this branch was never true since we check first for explicit interface specifiers and these cannot be present for static types (static types cannot implement interfaces).

dmitry-golovinov-sonarsource (approved):
LGTM!

PR #4418, opened 2021-05-12
S1939: RedundantInheritanceList - Records support

csaba-sagi-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantInheritanceList.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantInheritanceList.cs
@@ -39,10 +39,10 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class RedundantInheritanceList : SonarDiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S1939";
-        internal const string MessageEnum = "'int' should not be explicitly used as the underlying type.";
-        internal const string MessageObjectBase = "'Object' should not be explicitly extended.";
-        internal const string MessageAlreadyImplements = "'{0}' implements '{1}' so '{1}' can be removed from the inheritance list.";
         internal const string RedundantIndexKey = "redundantIndex";
+        private const string MessageEnum = "'int' should not be explicitly used as the underlying type.";
+        private const string MessageObjectBase = "'Object' should not be explicitly extended.";
+        private const string MessageAlreadyImplements = "'{0}' implements '{1}' so '{1}' can be removed from the inheritance list.";
         private const string MessageFormat = "{0}";
 
         private static readonly DiagnosticDescriptor Rule =
@@ -61,7 +61,7 @@ namespace SonarAnalyzer.Rules.CSharp
         {
             var typeDeclaration = (TypeDeclarationSyntax)context.Node;
             if (context.ContainingSymbol.Kind != SymbolKind.NamedType
-                || IsBaseListNullEmpty(typeDeclaration))
+                || IsBaseListNullOrEmpty(typeDeclaration))
             {
                 return;
             }
@@ -87,7 +87,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private static void CheckInterface(SyntaxNodeAnalysisContext context)
         {
             var interfaceDeclaration = (InterfaceDeclarationSyntax)context.Node;
-            if (IsBaseListNullEmpty(interfaceDeclaration))
+            if (IsBaseListNullOrEmpty(interfaceDeclaration))
             {
                 return;
             }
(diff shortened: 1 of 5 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantInheritanceList.cs
```diff
@@ -200,9 +200,9 @@ private static bool CanInterfacebeRemovedbasedOnMembers(INamedTypeSymbol declare
             foreach (var interfaceMember in allMembersOfInterface)
             {
                 var classMember = declaredType.FindImplementationForInterfaceMember(interfaceMember);
-                if (classMember != null &&
-                    (classMember.ContainingType.Equals(declaredType) ||
-                    !classMember.ContainingType.Interfaces.SelectMany(i => i.AllInterfaces.Concat(new[] { i })).Contains(interfaceType)))
+                if (classMember != null
```
dmitry-golovinov-sonarsource, line 203: I feel that we can improve the name casing of this method.

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantInheritanceList.cs
```diff
@@ -229,5 +229,8 @@ private static Location GetLocationWithToken(TypeSyntax type, SeparatedSyntaxLis
 
             return Location.Create(type.SyntaxTree, new TextSpan(start, end - start));
         }
+
+        private static bool IsBaseListNullEmpty(BaseTypeDeclarationSyntax baseTypeDelcaration) =>
+            baseTypeDelcaration.BaseList == null || !baseTypeDelcaration.BaseList.Types.Any();
```
dmitry-golovinov-sonarsource, line 234:
```suggestion
        private static bool IsBaseListNullEmpty(BaseTypeDeclarationSyntax baseTypeDeclaration) =>
            baseTypeDeclaration.BaseList == null || !baseTypeDeclaration.BaseList.Types.Any();
```

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantInheritanceListCodeFixProvider.cs
```diff
@@ -67,12 +58,19 @@ protected override Task RegisterCodeFixesAsync(SyntaxNode root, CodeFixContext c
             return TaskHelper.CompletedTask;
         }
 
+        internal static bool HasLineEnding(SyntaxNode node) =>
```
dmitry-golovinov-sonarsource, line 61: internal should go before protected

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantInheritanceListCodeFixProvider.cs
```diff
@@ -35,17 +35,8 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class RedundantInheritanceListCodeFixProvider : SonarCodeFixProvider
     {
         internal const string Title = "Remove redundant declaration";
-        public override ImmutableArray<string> FixableDiagnosticIds
```
dmitry-golovinov-sonarsource, line 38: _Title_ could be made _private_

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantInheritanceListCodeFixProvider.cs
```diff
@@ -67,12 +58,19 @@ protected override Task RegisterCodeFixesAsync(SyntaxNode root, CodeFixContext c
...
         }
 
+        internal static bool HasLineEnding(SyntaxNode node) =>
+            node.HasTrailingTrivia
+            && node.GetTrailingTrivia().Last().IsKind(SyntaxKind.EndOfLineTrivia);
+
+        private static bool HasLineEnding(SyntaxToken token) =>
+            token.HasTrailingTrivia
+            && token.TrailingTrivia.Last().IsKind(SyntaxKind.EndOfLineTrivia);
+
```
dmitry-golovinov-sonarsource, line 68: Do not need to break line 69 after the second argument

dmitry-golovinov-sonarsource (approved):
LGTM! Please check my minor suggestions

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantInheritanceListCodeFixProvider.cs
csaba-sagi-sonarsource: Indeed, yet in this case because it is static it is OK to leave it like this. According to the [coding style](https://github.com/SonarSource/sonar-dotnet/blob/master/docs/coding-style.md) : Static methods are preferred to be after instance methods.

PR #4425, opened 2021-05-14
S4050 fix FPs and add test cases for records

### LINKED ISSUE #4424: Fix S4050 FP: OperatorsShouldBeOverloadedConsistently - Align implementation with rule description
### Description

Overwriting the + or - operator forces the user to overwrite the other operator too. And that should not happen by the design of the rule.

csaba-sagi-sonarsource:
Fixes #4424

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/OperatorsShouldBeOverloadedConsistently.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/OperatorsShouldBeOverloadedConsistently.cs
@@ -60,10 +60,9 @@ namespace SonarAnalyzer.Rules.CSharp
             context.RegisterSyntaxNodeActionInNonGenerated(c =>
             {
                 var classDeclaration = (ClassDeclarationSyntax)c.Node;
-                var classSymbol = c.SemanticModel.GetDeclaredSymbol(classDeclaration);
+                var classSymbol = (INamedTypeSymbol)c.ContainingSymbol;
 
-                if (classSymbol == null
-                    || classDeclaration.Identifier.IsMissing
+                if (classDeclaration.Identifier.IsMissing
                     || !classSymbol.IsPubliclyAccessible())
                 {
                     return;
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/OperatorsShouldBeOverloadedConsistently.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/OperatorsShouldBeOverloadedConsistently.cs
@@ -37,14 +37,17 @@ namespace SonarAnalyzer.Rules.CSharp
         internal const string DiagnosticId = "S4050";
         private const string MessageFormat = "Provide an implementation for: {0}.";
 
-        private static readonly DiagnosticDescriptor rule =
+        private static readonly DiagnosticDescriptor Rule =
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
         private static class MethodName
         {
             public const string OperatorPlus = "operator+";
             public const string OperatorMinus = "operator-";
+            public const string OperatorMultiply = "operator*";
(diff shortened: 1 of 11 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/OperatorsShouldBeOverloadedConsistently.cs
```diff
@@ -37,14 +37,17 @@ public sealed class OperatorsShouldBeOverloadedConsistently : SonarDiagnosticAna
         internal const string DiagnosticId = "S4050";
```
costin-zaharia-sonarsource, line 37: can be private

analyzers/src/SonarAnalyzer.CSharp/Rules/OperatorsShouldBeOverloadedConsistently.cs
```diff
@@ -37,14 +37,17 @@ public sealed class OperatorsShouldBeOverloadedConsistently : SonarDiagnosticAna
...
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
         private static class MethodName
         {
             public const string OperatorPlus = "operator+";
             public const string OperatorMinus = "operator-";
+            public const string OperatorMultiply = "operator*";
+            public const string OperatorDevide = "operator/";
```
costin-zaharia-sonarsource, line 49: typo: "divide"

analyzers/src/SonarAnalyzer.CSharp/Rules/OperatorsShouldBeOverloadedConsistently.cs
```diff
@@ -57,22 +60,22 @@ protected override void Initialize(SonarAnalysisContext context)
             context.RegisterSyntaxNodeActionInNonGenerated(c =>
```
costin-zaharia-sonarsource, line 60: pedantic: can be arrow method

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/OperatorsShouldBeOverloadedConsistently.CSharp9.cs
```diff
@@ -1,15 +1,33 @@
...
     public static object operator -(Compliant a, Compliant b) => new object();
+    public static object operator *(Compliant a, Compliant b) => new object();
+    public static object operator /(Compliant a, Compliant b) => new object();
+    public static object operator %(Compliant a, Compliant b) => new object();
 }
 
-public record OnlyPlus // Compliant - FN, should implement `-` too
+public record OnlyPlus // Compliant
 {
     public static object operator +(OnlyPlus a, OnlyPlus b) => new object();
```
costin-zaharia-sonarsource, line 12: Don't we want to suggest `-` in this case? `==` cannot be overloaded but `-` can.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/OperatorsShouldBeOverloadedConsistently.cs
```diff
@@ -71,4 +88,35 @@ public class Foo6
     {
         public static object operator ==(Foo6 a, Foo6 b) => new object(); // Error [CS0216] - requires != operator
     }
+
+    public class Foo8 // Noncompliant
```
costin-zaharia-sonarsource, line 92: For these it would be nice to check also the display message

costin-zaharia-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/OperatorsShouldBeOverloadedConsistently.CSharp9.cs
csaba-sagi-sonarsource: Before these modifications the rule suggested to implement -, but that implementation was not aligned with the rule description. So, I changed the rule to suggest just overloading the == operator as the rule originally suggested. I am open to any discussion if you think we should go back to the previous implementation.

costin-zaharia-sonarsource (approved):
LGTM!

costin-zaharia-sonarsource (approved):
LGTM!

PR #4431, opened 2021-05-17
Rule S3459: Add support for records

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpRemovableDeclarationCollector.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpRemovableDeclarationCollector.cs
@@ -23,6 +23,7 @@ using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Helpers
 {
@@ -36,11 +37,11 @@ namespace SonarAnalyzer.Helpers
         {
         }
 
-        public static bool IsNodeStructOrClassDeclaration(SyntaxNode node) =>
-            node.IsKind(SyntaxKind.ClassDeclaration) || node.IsKind(SyntaxKind.StructDeclaration);
+        public static bool IsNodeStructOrClassOrRecordDeclaration(SyntaxNode node) =>
+            node.IsAnyKind(SyntaxKind.ClassDeclaration, SyntaxKind.StructDeclaration, SyntaxKindEx.RecordDeclaration);
 
         public static bool IsNodeContainerTypeDeclaration(SyntaxNode node) =>
-            IsNodeStructOrClassDeclaration(node) || node.IsKind(SyntaxKind.InterfaceDeclaration);
+            IsNodeStructOrClassOrRecordDeclaration(node) || node.IsKind(SyntaxKind.InterfaceDeclaration);
 
         protected override IEnumerable<SyntaxNode> SelectMatchingDeclarations(
             SyntaxNodeAndSemanticModel<BaseTypeDeclarationSyntax> container, ISet<SyntaxKind> kinds) =>
--- a/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpRemovableDeclarationCollector.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpRemovableDeclarationCollector.cs
@@ -37,19 +37,13 @@ namespace SonarAnalyzer.Helpers
         {
         }
 
-        public static bool IsNodeStructOrClassOrRecordDeclaration(SyntaxNode node) =>
-            node.IsAnyKind(SyntaxKind.ClassDeclaration, SyntaxKind.StructDeclaration, SyntaxKindEx.RecordDeclaration);
-
-        public static bool IsNodeContainerTypeDeclaration(SyntaxNode node) =>
-            IsNodeStructOrClassOrRecordDeclaration(node) || node.IsKind(SyntaxKind.InterfaceDeclaration);
-
         protected override IEnumerable<SyntaxNode> SelectMatchingDeclarations(
             SyntaxNodeAndSemanticModel<BaseTypeDeclarationSyntax> container, ISet<SyntaxKind> kinds) =>
             container.SyntaxNode.DescendantNodes(IsNodeContainerTypeDeclaration)
                 .Where(node => kinds.Contains(node.Kind()));
 
         public override IEnumerable<SyntaxNodeSymbolSemanticModelTuple> GetRemovableFieldLikeDeclarations(
-    ISet<SyntaxKind> kinds, Accessibility maxAcessibility)
+    ISet<SyntaxKind> kinds, Accessibility maxAccessibility)
         {
             var fieldLikeNodes = TypeDeclarations
                 .SelectMany(typeDeclaration => SelectMatchingDeclarations(typeDeclaration, kinds)
@@ -63,10 +57,16 @@ namespace SonarAnalyzer.Helpers
             return fieldLikeNodes
                 .SelectMany(fieldLikeNode => fieldLikeNode.SyntaxNode.Declaration.Variables
                     .Select(variable => SelectNodeTuple(variable, fieldLikeNode.SemanticModel))
-                    .Where(tuple => IsRemovable(tuple.Symbol, maxAcessibility)));
+                    .Where(tuple => IsRemovable(tuple.Symbol, maxAccessibility)));
         }
 
         internal override BaseTypeDeclarationSyntax GetOwnerOfSubnodes(BaseTypeDeclarationSyntax node) =>
             node;
+
+        public static bool IsNodeContainerTypeDeclaration(SyntaxNode node) =>
+            IsNodeStructOrClassOrRecordDeclaration(node) || node.IsKind(SyntaxKind.InterfaceDeclaration);
+
+        private static bool IsNodeStructOrClassOrRecordDeclaration(SyntaxNode node) =>
+            node.IsAnyKind(SyntaxKind.ClassDeclaration, SyntaxKind.StructDeclaration, SyntaxKindEx.RecordDeclaration);
     }
 }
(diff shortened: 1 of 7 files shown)


analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpRemovableDeclarationCollector.cs
```diff
@@ -36,11 +37,11 @@ public CSharpRemovableDeclarationCollector(INamedTypeSymbol namedType, Compilati
         {
         }
 
-        public static bool IsNodeStructOrClassDeclaration(SyntaxNode node) =>
-            node.IsKind(SyntaxKind.ClassDeclaration) || node.IsKind(SyntaxKind.StructDeclaration);
+        public static bool IsNodeStructOrClassOrRecordDeclaration(SyntaxNode node) =>
```
costin-zaharia-sonarsource, line 40: This method is used only here and should be made private.

analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpRemovableDeclarationCollector.cs
```diff
@@ -36,11 +37,11 @@ public CSharpRemovableDeclarationCollector(INamedTypeSymbol namedType, Compilati
...
+            node.IsAnyKind(SyntaxKind.ClassDeclaration, SyntaxKind.StructDeclaration, SyntaxKindEx.RecordDeclaration);
 
         public static bool IsNodeContainerTypeDeclaration(SyntaxNode node) =>
-            IsNodeStructOrClassDeclaration(node) || node.IsKind(SyntaxKind.InterfaceDeclaration);
+            IsNodeStructOrClassOrRecordDeclaration(node) || node.IsKind(SyntaxKind.InterfaceDeclaration);
 
         protected override IEnumerable<SyntaxNode> SelectMatchingDeclarations(
             SyntaxNodeAndSemanticModel<BaseTypeDeclarationSyntax> container, ISet<SyntaxKind> kinds) =>
```
costin-zaharia-sonarsource, line 47: pedantic: there is a type bellow `maxAccessibility`

analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -42,42 +42,37 @@ public sealed class NotAssignedPrivateMember : SonarDiagnosticAnalyzer
...
-        internal const string DiagnosticId = "S3459";
+        private const string DiagnosticId = "S3459";
         private const string MessageFormat = "Remove unassigned {0} '{1}', or set its value.";
 
-        private static readonly DiagnosticDescriptor rule =
+        private static readonly DiagnosticDescriptor Rule =
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        private static readonly Accessibility MaxAccessibility = Accessibility.Private;
```
costin-zaharia-sonarsource, line 51: this can be a constant

analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -42,42 +42,37 @@ public sealed class NotAssignedPrivateMember : SonarDiagnosticAnalyzer
...
-                        .Where(tuple => !IsInitializedOrFixed((VariableDeclaratorSyntax)tuple.SyntaxNode) &&
-                                        !HasStructLayoutAttribute(tuple.Symbol.ContainingType));
-
-                    var candidateProperties = removableDeclarationCollector.GetRemovableDeclarations(
-                        new HashSet<SyntaxKind> { SyntaxKind.PropertyDeclaration }, maxAccessibility)
-                        .Where(tuple => IsAutoPropertyWithNoInitializer((PropertyDeclarationSyntax)tuple.SyntaxNode) &&
-                                        !HasStructLayoutAttribute(tuple.Symbol.ContainingType));
-
-                    var allCandidateMembers = candidateFields.Concat(candidateProperties).ToList();
+                    var allCandidateMembers = RetrieveFielAndPropertyDeclarations(removableDeclarationCollector);
```
costin-zaharia-sonarsource, line 75: typo: "field"
I would rename the method to `GetFieldAndPropertyDeclatations` or `GetCandidateDeclarations`.

analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -87,56 +82,65 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                        var field = candidateMember.SyntaxNode as VariableDeclaratorSyntax;
+                        var property = candidateMember.SyntaxNode as PropertyDeclarationSyntax;
 
-                            var memberType = field != null ? "field" : "auto-property";
+                        var memberType = field != null ? "field" : "auto-property";
 
-                            var location = field != null
-                                ? field.Identifier.GetLocation()
-                                : property.Identifier.GetLocation();
+                        var location = field != null
```
costin-zaharia-sonarsource, line 94: we should invert the condition

analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -87,56 +82,65 @@ protected override void Initialize(SonarAnalysisContext context)
...
-
-                        if (!assignedMemberSymbols.Contains(candidateMember.Symbol))
-                        {
-                            var field = candidateMember.SyntaxNode as VariableDeclaratorSyntax;
-                            var property = candidateMember.SyntaxNode as PropertyDeclarationSyntax;
+                        var field = candidateMember.SyntaxNode as VariableDeclaratorSyntax;
+                        var property = candidateMember.SyntaxNode as PropertyDeclarationSyntax;
 
-                            var memberType = field != null ? "field" : "auto-property";
+                        var memberType = field != null ? "field" : "auto-property";
```
costin-zaharia-sonarsource, line 92: we should invert the condition
```suggestion
                        var memberType = field == null ? "auto-property" : "field";
```

analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -87,56 +82,65 @@ protected override void Initialize(SonarAnalysisContext context)
...
-                            c.ReportDiagnosticIfNonGenerated(Diagnostic.Create(rule, location, memberType, candidateMember.Symbol.Name));
-                        }
+                        c.ReportDiagnosticIfNonGenerated(Diagnostic.Create(Rule, location, memberType, candidateMember.Symbol.Name));
                     }
                 },
                 SymbolKind.NamedType);
+
+        private static List<SyntaxNodeSymbolSemanticModelTuple<SyntaxNode, ISymbol>> RetrieveFielAndPropertyDeclarations(CSharpRemovableDeclarationCollector removableDeclarationCollector)
+        {
+            var candidateFields = removableDeclarationCollector.GetRemovableFieldLikeDeclarations(
```
costin-zaharia-sonarsource, line 105: these are a bit hard to read. I suggest to move the parameters on the same line with the method call and align the `.Where` under the `.GetRemovableFieldLikeDeclarations`. Also && should be on the next line.

Same for `candidateProperties` below.

analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -87,56 +82,65 @@ protected override void Initialize(SonarAnalysisContext context)
...
+
+            var candidateProperties = removableDeclarationCollector.GetRemovableDeclarations(
+                        new HashSet<SyntaxKind> { SyntaxKind.PropertyDeclaration }, MaxAccessibility)
+                        .Where(tuple => IsAutoPropertyWithNoInitializer((PropertyDeclarationSyntax)tuple.SyntaxNode) &&
+                                        !HasStructLayoutAttribute(tuple.Symbol.ContainingType));
+
+            return candidateFields.Concat(candidateProperties).ToList();
         }
 
+        private static bool TypeDefinitionShouldBeSkipped(INamedTypeSymbol namedType) =>
```
costin-zaharia-sonarsource, line 118: the parameter can be of type `ITypeSymbol`

costin-zaharia-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
dmitry-golovinov-sonarsource: Actually, I want to bring up this alignment topic on our next how we work meeting. I don't feel it comfortable reading those aligned lines, where this dot part goes almost to the right part of the screen.

dmitry-golovinov-sonarsource: This also made me turning off type hints functionality of Rider completely, to make it at least readable

dmitry-golovinov-sonarsource: (basically, with Inlay hints on this alignment makes no sense, since it's not aligned in any case.

costin-zaharia-sonarsource: I find the inlay hints rarely useful. For me they make the code harder to read since they change indentation.

dmitry-golovinov-sonarsource: For me the thing is not only in them, rather reading the code in the right part of the screen. So I bet this is a good topic to discuss within our team.

costin-zaharia-sonarsource: We had a discussion about this in the past but sure, we can bring it up and see how we can improve this.

PR #4432, opened 2021-05-17
Rule S4428: add support for records

costin-zaharia-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Rules/PartCreationPolicyShouldBeUsedWithExportAttributeBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/PartCreationPolicyShouldBeUsedWithExportAttributeBase.cs
@@ -25,16 +25,15 @@ using SonarAnalyzer.Helpers;
 
 namespace SonarAnalyzer.Rules
 {
-    public abstract class PartCreationPolicyShouldBeUsedWithExportAttributeBase<TAttributeSyntax, TClassSyntax> : SonarDiagnosticAnalyzer
+    public abstract class PartCreationPolicyShouldBeUsedWithExportAttributeBase<TAttributeSyntax, TDeclarationSyntax> : SonarDiagnosticAnalyzer
         where TAttributeSyntax : SyntaxNode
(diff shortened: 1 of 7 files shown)


analyzers/src/SonarAnalyzer.Common/Rules/PartCreationPolicyShouldBeUsedWithExportAttributeBase.cs
```diff
@@ -44,30 +43,26 @@ protected void AnalyzeNode(SyntaxNodeAnalysisContext c)
...
-            c.ReportDiagnosticWhenActive(
-                Diagnostic.Create(
-                    SupportedDiagnostics[0],
-                    attribute.GetLocation()));
+            c.ReportDiagnosticWhenActive(Diagnostic.Create(SupportedDiagnostics[0], attribute.GetLocation()));
 
             bool IsPartCreationPolicyAttribute(TAttributeSyntax attributeSyntax) =>
-                c.SemanticModel.GetSymbolInfo(attributeSyntax).Symbol is IMethodSymbol attributeSymbol &&
-                attributeSymbol.ContainingType.Is(KnownType.System_ComponentModel_Composition_PartCreationPolicyAttribute);
+                c.SemanticModel.GetSymbolInfo(attributeSyntax).Symbol is IMethodSymbol attributeSymbol
```
dmitry-golovinov-sonarsource, line 64: AFAIK _c.SemanticModel.GetSymbolInfo(c.Node).Symbol_ returns the same thing we already have in _c.ContainingSymbol_

dmitry-golovinov-sonarsource (approved):
LGTM. Please check my minor suggestion before merging.

analyzers/src/SonarAnalyzer.Common/Rules/PartCreationPolicyShouldBeUsedWithExportAttributeBase.cs
costin-zaharia-sonarsource: In this case no. `c.ContainingSymbol` for this attribute is the compilation unit

PR #4434, opened 2021-05-17
Rule S3253: add support for records

costin-zaharia-sonarsource:


csaba-sagi-sonarsource (approved):
LGTM! I am wondering if it would make sense to add a test case for the code fixer for the CSharp9 test case?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundancyInConstructorDestructorDeclaration.BaseCall.Fixed.cs
```diff
@@ -93,4 +93,9 @@ public MyClass6() : base(10)
         {
         }
     }
+
+    class LambdaCtor
+    {
+        private int i;
+        LambdaCtor() => i++;     }
```
costin-zaharia-sonarsource, line 100: See: https://github.com/SonarSource/sonar-dotnet/issues/4439

PR #4435, opened 2021-05-17
Rule S1192: add support for records

costin-zaharia-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
@@ -26,6 +26,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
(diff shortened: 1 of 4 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
```diff
@@ -38,7 +39,8 @@ public sealed class StringLiteralShouldNotBeDuplicated : StringLiteralShouldNotB
         protected override SyntaxKind[] SyntaxKinds { get; } =
         {
             SyntaxKind.ClassDeclaration,
-            SyntaxKind.StructDeclaration
+            SyntaxKind.StructDeclaration,
+            SyntaxKindEx.RecordDeclaration
         };
 
         public StringLiteralShouldNotBeDuplicated() : base(RspecStrings.ResourceManager) { }
```
csaba-sagi-sonarsource, line 46: Can we introduce ILanguageFacade to the base class? That would remove this constructor and the GeneratedCodeRecognizer property. Both from the Csharp and the VbNet implementations.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/StringLiteralShouldNotBeDuplicated.CSharp9.cs
```diff
@@ -7,39 +7,48 @@
 
 record Record
```
csaba-sagi-sonarsource, line 8: Can you please add a positional record example? As the base class does not do the containingSymbol check currently, I expect that with the current implementation we raise twice the issues.

csaba-sagi-sonarsource (changes requested):
Please check the behaviour for Positional Records.

csaba-sagi-sonarsource (approved):
LGTM!

PR #4437, opened 2021-05-17
Rule S1128: fix false positive with top level statements

costin-zaharia-sonarsource:


csaba-sagi-sonarsource (approved):
LGTM! It would be nice to fix the two stylecop issues. Converting rule-> Rule and necessaryNamespaces ->NecessaryNamespaces.

PR #4438, opened 2021-05-17
Rule S3994: add support for records

costin-zaharia-sonarsource:


csaba-sagi-sonarsource (approved):
LGTM! Nice catch for the property declaration for positional records!

PR #4440, opened 2021-05-18
S3241: UnusedReturnValue - records support

csaba-sagi-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/UnusedReturnValue.cs
```diff
@@ -35,105 +35,89 @@ namespace SonarAnalyzer.Rules.CSharp
...
+                || localFunctionSymbol.ReturnsVoid
+                || topmostContainingMethod == null)
             {
-                return false;
+                return;
             }
 
-            return semanticModel.GetSymbolInfo(lambda).Symbol is IMethodSymbol symbol && symbol.ReturnsVoid;
+            var matchingInvocations = GetLocalMatchingInvocations(topmostContainingMethod, localFunctionSymbol, context.SemanticModel);
+            // Method invocation is noncompliant when there is at least 1 inovcation of the method, and no invocation is using the return value. The case of 0 invocation is handled by S1144.
```
dmitry-golovinov-sonarsource, line 96:
```suggestion
            // Method invocation is noncompliant when there is at least 1 invocation of the method, and no invocation is using the return value. The case of 0 invocation is handled by S1144.
```

analyzers/src/SonarAnalyzer.CSharp/Rules/UnusedReturnValue.cs
```diff
@@ -35,105 +35,89 @@ namespace SonarAnalyzer.Rules.CSharp
...
+            if (matchingInvocations.Any() && !matchingInvocations.Any(x => IsReturnValueUsed(x)))
```
dmitry-golovinov-sonarsource, line 97: It's debatable, but in cases like this I would better convert matchingInvocations to list

analyzers/src/SonarAnalyzer.CSharp/Rules/UnusedReturnValue.cs
```diff
@@ -143,39 +127,33 @@ private static bool IsExpressionStatement(SyntaxNode node)
...
+                        .Select(node =>
+                            new SyntaxNodeSymbolSemanticModelTuple<MethodDeclarationSyntax, IMethodSymbol>
+                            {
+                                SyntaxNode = node,
+                                SemanticModel = container.SemanticModel,
+                                Symbol = container.SemanticModel.GetDeclaredSymbol(node)
+                            }))
+                        .Where(node =>
+                            node.Symbol != null
+                            && !node.Symbol.ReturnsVoid
```
dmitry-golovinov-sonarsource, line 156:
```suggestion
                            node.Symbol is { ReturnsVoid: false }
```

analyzers/src/SonarAnalyzer.CSharp/Rules/UnusedReturnValue.cs
dmitry-golovinov-sonarsource: The same way you do this on line 74

dmitry-golovinov-sonarsource (commented):
See my comments. Also, WDYT about adding C#9 test with interface having a default implementation with such a problem?

dmitry-golovinov-sonarsource (approved):
LGTM. You can try to improve the coverage by testing broken method declaration, that should lead to the syntax with no symbol

Issue #4446, opened 2021-05-20
Fix S1121 FP: Issue should not be raised in the condition inside a while loop


dmitry-golovinov-sonarsource:
### Description

S1121 should not trigger the issue in while loop's condition

### Repro steps

```
interface ICustomMsgQueue
{
    string? Pop();
}

class MessageQueueUseCase
{
    void Process(ICustomMsgQueue queue)
    {
        string msg;
        while ((msg = queue.Pop()) is not null)  // Noncompliant FP
        {
            Console.WriteLine(msg);
        }
    }
}
```

### Expected behavior

Assigned variable in while loop condition, whose value is checked inside the condition or used inside the loop, should not raise an issue

### Actual behavior

Issue is being raised.

### Related information

* C#/VB.NET Plugins version 8.22

andrei-epure-sonarsource:
This got fixed in #5044

PR #4447, opened 2021-05-20
Documenting S1121 FP

dmitry-golovinov-sonarsource:
Repro for #4446


csaba-sagi-sonarsource (approved):
LGTM! Please add the following description on the PR: 
Repro for #4446

PR #4448, opened 2021-05-20
Update rspec

costin-zaharia-sonarsource:


analyzers/rspec/cs/S4581_c#.html
```diff
@@ -10,6 +10,8 @@ <h2>Noncompliant Code Example</h2>
 public void Foo()
 {
     var g = new Guid(); // Noncompliant - what's the intent?
+    var guid = default(Guid);
```
csaba-sagi-sonarsource, line 13: I would not update this, as we do not raise issue for these two cases. And I assume we should also create an issue for that.

analyzers/src/SonarAnalyzer.Utilities/Rules.Description/S4581.html
```diff
@@ -10,6 +10,8 @@ <h2>Noncompliant Code Example</h2>
 public void Foo()
 {
     var g = new Guid(); // Noncompliant - what's the intent?
+    var guid = default(Guid);
```
csaba-sagi-sonarsource, line 13: Same as I mentioned already.

csaba-sagi-sonarsource (approved):
LGTM!

PR #4451, opened 2021-05-20
Cleanup SymbolHelper

costin-zaharia-sonarsource:
- moved methods to respect our convention
- use arrow methods where possible
- fix alignment
- use switch expressions
- re-arrange operators where needed


dmitry-golovinov-sonarsource (approved):
Nice refactoring!

PR #4458, opened 2021-05-21
Rule S4432: Add support for records

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Helpers/Wrappers/ObjectCreationFactory.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Helpers/Wrappers/ObjectCreationFactory.cs
@@ -1,0 +1,84 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2021 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using StyleCop.Analyzers.Lightup;
+
+namespace SonarAnalyzer.Helpers.Wrappers
+{
+    public interface IObjectCreation
+    {
+        InitializerExpressionSyntax GetInitializer();
+        ArgumentListSyntax GetArgumentList();
+        ExpressionSyntax GetExpression();
+        string GetTypeAsString(SemanticModel semanticModel);
+    }
+
+    public class ObjectCreationFactory
+    {
+        public static IObjectCreation Create(SyntaxNode node) =>
+            node is ObjectCreationExpressionSyntax objectCreationExpressionSyntax
+                ? (IObjectCreation)new ObjectCreation(objectCreationExpressionSyntax)
+                : new ImplicitObjectCreation((ImplicitObjectCreationExpressionSyntaxWrapper)node);
+
+        private class ObjectCreation : IObjectCreation
+        {
+            private readonly ObjectCreationExpressionSyntax objectCreation;
+            public ObjectCreation(ObjectCreationExpressionSyntax objectCreationExpressionSyntax)
+            {
+                objectCreation = objectCreationExpressionSyntax;
+            }
+
+            public InitializerExpressionSyntax GetInitializer() =>
+                objectCreation.Initializer;
+
+            public ArgumentListSyntax GetArgumentList() =>
+                objectCreation.ArgumentList;
+
+            public ExpressionSyntax GetExpression() =>
+                objectCreation;
+
+            public string GetTypeAsString(SemanticModel semanticModel) =>
+                objectCreation.Type.ToString();
+        }
+
+        private class ImplicitObjectCreation : IObjectCreation
+        {
+            private readonly ImplicitObjectCreationExpressionSyntaxWrapper objectCreation;
+            public ImplicitObjectCreation(ImplicitObjectCreationExpressionSyntaxWrapper wrapper)
+            {
+                objectCreation = wrapper;
+            }
+
+            public InitializerExpressionSyntax GetInitializer() =>
+                objectCreation.Initializer;
+
+            public ArgumentListSyntax GetArgumentList() =>
+                objectCreation.ArgumentList;
+
+            public ExpressionSyntax GetExpression() =>
+                objectCreation.SyntaxNode;
+
+            public string GetTypeAsString(SemanticModel semanticModel) =>
+                semanticModel.GetTypeInfo(objectCreation).Type.Name;
+        }
+    }
+}
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
@@ -31,6 +31,8 @@ using SonarAnalyzer.Common;
 using SonarAnalyzer.Extensions;
 using SonarAnalyzer.Helpers;
 using SonarAnalyzer.Helpers.Trackers;
+using SonarAnalyzer.Helpers.Wrappers;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
@@ -117,22 +119,22 @@ namespace SonarAnalyzer.Rules.CSharp
                 }
 
                 context.RegisterSyntaxNodeActionInNonGenerated(VisitStringExpressions, SyntaxKind.StringLiteralExpression, SyntaxKind.InterpolatedStringExpression);
-                context.RegisterSyntaxNodeActionInNonGenerated(VisitObjectCreation, SyntaxKind.ObjectCreationExpression);
+                context.RegisterSyntaxNodeActionInNonGenerated(VisitObjectCreation, SyntaxKind.ObjectCreationExpression, SyntaxKindEx.ImplicitObjectCreationExpression);
                 context.RegisterSyntaxNodeActionInNonGenerated(VisitInvocationExpression, SyntaxKind.InvocationExpression);
                 context.RegisterSyntaxNodeActionInNonGenerated(VisitAssignments, SyntaxKind.SimpleAssignmentExpression);
             });
 
         private void VisitObjectCreation(SyntaxNodeAnalysisContext context)
         {
-            var objectCreation = (ObjectCreationExpressionSyntax)context.Node;
+            var objectCreation = ObjectCreationFactory.Create(context.Node);
 
-            if (telnetRegexForIdentifier.IsMatch(objectCreation.Type.ToString()))
+            if (!IsServerSafe(objectCreation) && objectInitializationTracker.ShouldBeReported(objectCreation, context.SemanticModel, false))
             {
-                context.ReportDiagnosticWhenActive(Diagnostic.Create(DefaultRule, objectCreation.GetLocation(), TelnetKey, recommendedProtocols[TelnetKey]));
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(EnableSslRule, objectCreation.GetExpression().GetLocation()));
             }
-            else if (!IsServerSafe(objectCreation) && objectInitializationTracker.ShouldBeReported(objectCreation, context.SemanticModel, false))
+            else if (telnetRegexForIdentifier.IsMatch(objectCreation.GetTypeAsString(context.SemanticModel)))
             {
-                context.ReportDiagnosticWhenActive(Diagnostic.Create(EnableSslRule, objectCreation.GetLocation()));
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(DefaultRule, objectCreation.GetExpression().GetLocation(), TelnetKey, recommendedProtocols[TelnetKey]));
             }
         }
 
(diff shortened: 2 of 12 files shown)


costin-zaharia-sonarsource:
We need to update the name, right? 😄

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/ClearTextProtocolsAreSensitive.CSharp9.cs
```diff
@@ -16,6 +16,7 @@ void Method()
     var a = "http://foo.com"; // Noncompliant
     var b = "https://foo.com";
     var httpProtocolScheme = "http://"; // It's compliant when standalone
+    Telnet c = new(); // Noncompliant
```
pavel-mikula-sonarsource, line 19: What about this?
```
Telnet d;
d = new();
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/CookieShouldBeHttpOnly_Net.cs
```diff
@@ -20,7 +20,7 @@ class Program
 
     void InitializerSetsNotAllowedValue()
     {
-        CookieOptions c1 = new () { HttpOnly = false }; // FN
-        CookieOptions c2 = new () { Secure = true }; // FN, as HttpOnly is not set
+        CookieOptions c1 = new () { HttpOnly = false }; // Noncompliant
+        CookieOptions c2 = new () { Secure = true }; // Noncompliant
```
pavel-mikula-sonarsource, line 24: _Nitpick (for next time):_ I'd align these comments when I'm touching old code (only these two, not those above)

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/ClearTextProtocolsAreSensitive.cs
```diff
@@ -117,22 +119,22 @@ public ClearTextProtocolsAreSensitive(IAnalyzerConfiguration analyzerConfigurati
                 }
 
                 context.RegisterSyntaxNodeActionInNonGenerated(VisitStringExpressions, SyntaxKind.StringLiteralExpression, SyntaxKind.InterpolatedStringExpression);
-                context.RegisterSyntaxNodeActionInNonGenerated(VisitObjectCreation, SyntaxKind.ObjectCreationExpression);
+                context.RegisterSyntaxNodeActionInNonGenerated(VisitObjectCreation, SyntaxKind.ObjectCreationExpression, SyntaxKindEx.ImplicitObjectCreationExpression);
```
pavel-mikula-sonarsource, line 122: Shouldn't the title of this PR mention S5332 as well? This new registration is a new feature implementation.

analyzers/src/SonarAnalyzer.CSharp/Rules/XmlExternalEntityShouldNotBeParsed.cs
```diff
@@ -63,20 +65,20 @@ public sealed class XmlExternalEntityShouldNotBeParsed : SonarDiagnosticAnalyzer
...
                             {
-                                c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, objectCreation.GetLocation()));
+                                c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, objectCreation.Expression.GetLocation()));
                             }
 
-                            VerifyXPathDocumentConstructor(c);
+                            VerifyXPathDocumentConstructor(c, objectCreation);
                         },
-                        SyntaxKind.ObjectCreationExpression);
+                        SyntaxKind.ObjectCreationExpression, SyntaxKindEx.ImplicitObjectCreationExpression);
```
pavel-mikula-sonarsource, line 81: Shouldn't S2755 be part of this PR title as well? There's this new registration.

analyzers/src/SonarAnalyzer.CSharp/Helpers/Wrappers/ObjectCreationFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+using StyleCop.Analyzers.Lightup;
+
+namespace SonarAnalyzer.Helpers.Wrappers
+{
+    public interface IObjectCreation
+    {
+        InitializerExpressionSyntax Initializer { get; }
+        ArgumentListSyntax ArgumentList { get; }
+        ExpressionSyntax Expression { get; }
+        string GetTypeAsString(SemanticModel semanticModel);
```
pavel-mikula-sonarsource, line 32: I'm generally avoiding `Get` prefix wherever possible, as it's redundant.

analyzers/src/SonarAnalyzer.CSharp/Helpers/Wrappers/ObjectCreationFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public class ObjectCreationFactory
+    {
+        public static IObjectCreation Create(SyntaxNode node) =>
+            node is ObjectCreationExpressionSyntax objectCreationExpressionSyntax
```
pavel-mikula-sonarsource, line 38: Let's use switch expression instead that will explicitly throw invalid operation exception under the `_` branch.

analyzers/src/SonarAnalyzer.CSharp/Helpers/Wrappers/ObjectCreationFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+                ? (IObjectCreation)new ObjectCreation(objectCreationExpressionSyntax)
+                : new ImplicitObjectCreation((ImplicitObjectCreationExpressionSyntaxWrapper)node);
+
+        private class ObjectCreation : IObjectCreation
+        {
+            private readonly ObjectCreationExpressionSyntax objectCreation;
+            public InitializerExpressionSyntax Initializer => objectCreation.Initializer;
```
pavel-mikula-sonarsource, line 45: I'm not fan of empty lines, but I'd put one between block of fields and block of properties to separate state and contract (same below).
```suggestion
            private readonly ObjectCreationExpressionSyntax objectCreation;
            
            public InitializerExpressionSyntax Initializer => objectCreation.Initializer;
```

analyzers/src/SonarAnalyzer.CSharp/Helpers/Wrappers/ObjectCreationFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+            public ArgumentListSyntax ArgumentList => objectCreation.ArgumentList;
+            public ExpressionSyntax Expression => objectCreation;
+
+            public ObjectCreation(ObjectCreationExpressionSyntax objectCreationExpressionSyntax)
+            {
+                objectCreation = objectCreationExpressionSyntax;
+            }
```
pavel-mikula-sonarsource, line 52: Can be arrow method (same below)
```suggestion
            public ObjectCreation(ObjectCreationExpressionSyntax objectCreationExpressionSyntax) =>
                objectCreation = objectCreationExpressionSyntax;
```

analyzers/src/SonarAnalyzer.CSharp/Helpers/Wrappers/ObjectCreationFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+            public ArgumentListSyntax ArgumentList => objectCreation.ArgumentList;
+            public ExpressionSyntax Expression => objectCreation.SyntaxNode;
+
+            public ImplicitObjectCreation(ImplicitObjectCreationExpressionSyntaxWrapper wrapper)
+            {
+                objectCreation = wrapper;
+            }
+
+            public string GetTypeAsString(SemanticModel semanticModel) =>
+                semanticModel.GetTypeInfo(objectCreation).Type.Name;
```
pavel-mikula-sonarsource, line 71: Are you sure that there's not going to be `null` anywhere on the path?
Let's add some test case with missing type name and `// Error` flag to make it compilable.

analyzers/src/SonarAnalyzer.CSharp/Helpers/Wrappers/ObjectCreationFactory.cs
pavel-mikula-sonarsource: I'd consider explicit `null` branch with appropriate exception as first one, because not-null is implicit expectation in the factory implementation below. It will make the `InvalidOperationException("Unexpected type: " + c.Node.GetType().Name)` easier

pavel-mikula-sonarsource: We'll need a UT file for this class then.

analyzers/src/SonarAnalyzer.CSharp/Helpers/Wrappers/ObjectCreationFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using StyleCop.Analyzers.Lightup;
+
+namespace SonarAnalyzer.Helpers.Wrappers
```
pavel-mikula-sonarsource, line 25: Let's move `Wrappers` directory to the root and change the namespace. Basically everything except `Rules` could be inside `Helpers` but there's no reason for it to be there.

pavel-mikula-sonarsource (changes requested):
Some minor improvements

analyzers/src/SonarAnalyzer.CSharp/Helpers/Wrappers/ObjectCreationFactory.cs
dmitry-golovinov-sonarsource: There is no way to obtain null here. Added test that proves this behavior.

analyzers/src/SonarAnalyzer.CSharp/Wrappers/ObjectCreationFactory.cs
```diff
@@ -0,0 +1,77 @@
...
+        ExpressionSyntax Expression { get; }
+        string TypeAsString(SemanticModel semanticModel);
+    }
+
+    public class ObjectCreationFactory
+    {
+        public static IObjectCreation Create(SyntaxNode node) =>
+            node switch
+            {
+                null => throw new ArgumentNullException(nameof(node), "Argument should not be null"),
```
pavel-mikula-sonarsource, line 41: This will work just as fine, as there's nothing additional to be said in the error
```suggestion
                null => throw new ArgumentNullException(nameof(node)),
```

analyzers/src/SonarAnalyzer.CSharp/Wrappers/ObjectCreationFactory.cs
```diff
@@ -0,0 +1,77 @@
...
+                ObjectCreationExpressionSyntax objectCreation => new ObjectCreation(objectCreation),
+                { } wrapper when ImplicitObjectCreationExpressionSyntaxWrapper.IsInstance(wrapper) => new ImplicitObjectCreation((ImplicitObjectCreationExpressionSyntaxWrapper)wrapper),
```
pavel-mikula-sonarsource, line 43: We don't need to rename it. And the instance is not actually wrapper but rather the node that will be wrapped.
```suggestion
                { } when ImplicitObjectCreationExpressionSyntaxWrapper.IsInstance(node) => new ImplicitObjectCreation((ImplicitObjectCreationExpressionSyntaxWrapper)node),
```

analyzers/tests/SonarAnalyzer.UnitTest/Wrappers/ObjectCreationFactoryTest.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        [TestMethod]
+        public void GivenNull_ThrowsException()
+        {
+            Action action = () => { ObjectCreationFactory.Create(null); };
+#if NET
+            action.Should().Throw<ArgumentNullException>().WithMessage("Argument should not be null (Parameter 'node')");
+#else
+            action.Should().Throw<ArgumentNullException>().WithMessage("Argument should not be null\nParameter name: node");
+#endif
```
pavel-mikula-sonarsource, line 81: I would not fight with the framework. Let's keep it simple:
```suggestion
            action.Should().Throw<ArgumentNullException>();
```

analyzers/tests/SonarAnalyzer.UnitTest/Wrappers/ObjectCreationFactoryTest.cs
```diff
@@ -0,0 +1,100 @@
...
+        public void GivenNonConstructor_ThrowsException()
+        {
+            var compilation = CreateCompilation("public class A{}");
+            var syntaxTree = compilation.SyntaxTrees.First();
+            var classDeclaration = syntaxTree.GetRoot().DescendantNodes().OfType<ClassDeclarationSyntax>().First();
+            Action action = () => { ObjectCreationFactory.Create(classDeclaration); };
+            action.Should().Throw<InvalidOperationException>().WithMessage("Unexpected type: ClassDeclarationSyntax");
+        }
+
+        private static CSharpCompilation CreateCompilation(string code) =>
```
pavel-mikula-sonarsource, line 94: This shouldn't be needed. You an use either `SnippetCompiler` or `TestHelper.Compile` to achieve the same with existing helpers.

analyzers/tests/SonarAnalyzer.UnitTest/Wrappers/ObjectCreationFactoryTest.cs
```diff
@@ -0,0 +1,100 @@
...
+
+namespace SonarAnalyzer.UnitTest.Wrappers
+{
+    [TestClass]
+    public class ObjectCreationFactoryTest
+    {
+        [TestMethod]
+        public void ObjectCreationSyntax()
+        {
+            var compilation = CreateCompilation("public class A{public int X;public A(int y){}} public class B{void Foo(){var bar =new A(1){X = 2};}}");
```
pavel-mikula-sonarsource, line 40: There's quite a lot of stuff going on there. Because we have two classes, I'd use verbatim multiline string instead (with as few new lines as possible)
```
  var code = @"
public class A
{
    public int X;
    public A(int y) { }
}
public class B
{
    void Foo()
    {
        var bar = new A(1) { X = 2 };
    }
}";
```

analyzers/tests/SonarAnalyzer.UnitTest/Wrappers/ObjectCreationFactoryTest.cs
```diff
@@ -0,0 +1,100 @@
...
+            wrapper.Expression.Should().BeEquivalentTo(objectCreation);
+            wrapper.Initializer.Should().BeEquivalentTo(objectCreation.Initializer);
+            wrapper.ArgumentList.Should().BeEquivalentTo(objectCreation.ArgumentList);
+            wrapper.TypeAsString(compilation.GetSemanticModel(syntaxTree)).Should().Be("A");
+        }
+
+        [TestMethod]
+        public void ImplicitObjectCreationSyntax()
+        {
+            var compilation = CreateCompilation("public class A{public int X;public A(int y){}} public class B{void Foo(){A bar =new(1){X = 2};}}");
```
pavel-mikula-sonarsource, line 53: Same as above

pavel-mikula-sonarsource (changes requested):
Last round

pavel-mikula-sonarsource (approved):
LGTM

PR #4471, opened 2021-05-25
Rules S2278, S5547, S2070: Add support for target typed new expressions

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/InsecureEncryptionAlgorithm.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/InsecureEncryptionAlgorithm.cs
@@ -59,10 +59,8 @@ namespace SonarAnalyzer.Rules.CSharp
             ((LiteralExpressionSyntax)argument.Expression).Token.ValueText;
 
         protected override Location Location(SyntaxNode objectCreation) =>
-            objectCreation switch
-            {
-                ObjectCreationExpressionSyntax objectCreationExpression => objectCreationExpression.Type.GetLocation(),
-                _ => ((ImplicitObjectCreationExpressionSyntaxWrapper)objectCreation).SyntaxNode.GetLocation()
-            };
+            objectCreation is ObjectCreationExpressionSyntax objectCreationExpression
+                ? objectCreationExpression.Type.GetLocation()
+                : ((ImplicitObjectCreationExpressionSyntaxWrapper)objectCreation).SyntaxNode.GetLocation();
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/InsecureEncryptionAlgorithm.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/InsecureEncryptionAlgorithm.cs
@@ -25,7 +25,6 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
-using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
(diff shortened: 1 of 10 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/InsecureHashAlgorithm.cs
```diff
@@ -82,8 +84,10 @@ public sealed class InsecureHashAlgorithm : DoNotCallInsecureSecurityAlgorithmBa
                 KnownType.System_Security_Cryptography_HMACRIPEMD160
             );
 
-        protected override Location Location(ObjectCreationExpressionSyntax objectCreation) =>
-            objectCreation.Type.GetLocation();
+        protected override Location Location(SyntaxNode objectCreation) =>
+            objectCreation is ObjectCreationExpressionSyntax objectCreationExpression
```
pavel-mikula-sonarsource, line 88: Why not to reuse the factory here as well? This `if` logic (including error handling) is already implemented there.

analyzers/src/SonarAnalyzer.CSharp/Rules/InsecureHashAlgorithm.cs
```diff
@@ -27,21 +27,23 @@
...
     {
-        internal const string DiagnosticId = "S2070";
+        private const string DiagnosticId = "S2070";
         private const string MessageFormat = "Use a stronger hashing/asymmetric algorithm.";
 
         private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
         protected override ILanguageFacade<SyntaxKind> Language { get; } = CSharpFacade.Instance;
+        protected override SyntaxKind[] ObjectCreationExpressionKinds { get; } = { SyntaxKind.ObjectCreationExpression, SyntaxKindEx.ImplicitObjectCreationExpression };
```
pavel-mikula-sonarsource, line 46: If there's at least one more rule that will need this, I'd put it in the facade.

analyzers/src/SonarAnalyzer.CSharp/Rules/InsecureEncryptionAlgorithm.cs
```diff
@@ -56,7 +58,9 @@ public sealed class InsecureEncryptionAlgorithm : InsecureEncryptionAlgorithmBas
         protected override string StringLiteralValue(ArgumentSyntax argument) =>
             ((LiteralExpressionSyntax)argument.Expression).Token.ValueText;
 
-        protected override Location Location(ObjectCreationExpressionSyntax objectCreation) =>
-            objectCreation.Type.GetLocation();
+        protected override Location Location(SyntaxNode objectCreation) =>
+            objectCreation is ObjectCreationExpressionSyntax objectCreationExpression
```
pavel-mikula-sonarsource, line 62: Same here

pavel-mikula-sonarsource (changes requested):
Some questions

analyzers/src/SonarAnalyzer.CSharp/Rules/InsecureHashAlgorithm.cs
dmitry-golovinov-sonarsource: We agreed to do it in slightly different way, since there is no Type part in `ImplicitObjectCreationException`

analyzers/src/SonarAnalyzer.CSharp/Rules/InsecureEncryptionAlgorithm.cs
dmitry-golovinov-sonarsource: We agreed to do it in slightly different way, since there is no Type part in `ImplicitObjectCreationException`

pavel-mikula-sonarsource (approved):
LGTM

PR #4474, opened 2021-05-26
Rule S3254: Add support for target typed new expressions

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantArgument.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantArgument.cs
@@ -55,19 +55,13 @@ namespace SonarAnalyzer.Rules.CSharp
                         ? invocationExpression.ArgumentList
                         : ((ImplicitObjectCreationExpressionSyntaxWrapper)c.Node).ArgumentList;
                     var methodParameterLookup = new CSharpMethodParameterLookup(argumentList, c.SemanticModel);
-                    var argumentMappings = methodParameterLookup.GetAllArgumentParameterMappings().ToList();
 
-                    var methodSymbol = methodParameterLookup.MethodSymbol;
-                    if (methodSymbol == null)
+                    if (methodParameterLookup.MethodSymbol != null)
                     {
-                        return;
-                    }
-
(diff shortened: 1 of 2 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantArgument.cs
```diff
@@ -52,10 +51,11 @@ protected override void Initialize(SonarAnalysisContext context)
...
-                    var argumentMappings = methodParameterLookup.GetAllArgumentParameterMappings()
-                        .ToList();
+                    var argumentList = c.Node is InvocationExpressionSyntax invocationExpression
+                        ? invocationExpression.ArgumentList
+                        : ((ImplicitObjectCreationExpressionSyntaxWrapper)c.Node).ArgumentList;
+                    var methodParameterLookup = new CSharpMethodParameterLookup(argumentList, c.SemanticModel);
+                    var argumentMappings = methodParameterLookup.GetAllArgumentParameterMappings().ToList();
 
                     var methodSymbol = methodParameterLookup.MethodSymbol;
                     if (methodSymbol == null)
```
pavel-mikula-sonarsource, line 61: While we're touching this rule, let's remove the return and make it shorter
```
if(methodParameterLookup.MethodSymbol is {} methodSymbol)
{
    // Move var argumentMappings = .. here as a performance optimization - it's not needed before the condition
}

pavel-mikula-sonarsource (approved):
LGTM, I left one more refactoring suggestion

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantArgument.cs
pavel-mikula-sonarsource: Or even better - rename the foreach variable to `x` to make it short. And then we don't really need `argumentMappings` variable nor the `ToList()`

pavel-mikula-sonarsource: I'd be also fine with removing `argument` and `parameter` as it's just a rename (waste of lines)

PR #4475, opened 2021-05-26
Add configuration for StyleCop rules

andrei-epure-sonarsource:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -269,8 +269,8 @@ dotnet_diagnostic.SA1028.severity = warning
 
 # Ordering Rules (SA1200-) https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/1.1.118/documentation/OrderingRules.md
 
-dotnet_diagnostic.SA1200.severity = none # needs configuration to be "outsideNamespace"
-dotnet_diagnostic.SA1201.severity = none # needs configuration for our coding style order
+dotnet_diagnostic.SA1200.severity = warning
+dotnet_diagnostic.SA1201.severity = warning # needs monitoring, might need tuning inside the stylecop.json configuration
 dotnet_diagnostic.SA1202.severity = warning
 dotnet_diagnostic.SA1203.severity = warning
 dotnet_diagnostic.SA1204.severity = none # Doesn't match our coding style for private static methods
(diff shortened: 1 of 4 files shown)


.editorconfig
```diff
@@ -269,8 +269,8 @@ dotnet_diagnostic.SA1028.severity = warning
 
 # Ordering Rules (SA1200-) https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/1.1.118/documentation/OrderingRules.md
 
-dotnet_diagnostic.SA1200.severity = none # needs configuration to be "outsideNamespace"
-dotnet_diagnostic.SA1201.severity = none # needs configuration for our coding style order
+dotnet_diagnostic.SA1200.severity = warning
+dotnet_diagnostic.SA1201.severity = warning # needs monitoring, might need tuning inside the stylecop.json configuration
```
pavel-mikula-sonarsource, line 273: This one complains about our constructors => too noisy

analyzers/stylecop.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://raw.githubusercontent.com/DotNetAnalyzers/StyleCopAnalyzers/master/StyleCop.Analyzers/StyleCop.Analyzers/Settings/stylecop.schema.json",
+  "settings": {
+    "orderingRules": {
+      "systemUsingDirectivesFirst": true,
```
pavel-mikula-sonarsource, line 5: `true` is default => we should not need it
```suggestion
```

analyzers/stylecop.json
```diff
@@ -0,0 +1,19 @@
...
+      "usingDirectivesPlacement": "outsideNamespace",
+      "elementOrder": [
+        "constant",
+        "static",
```
pavel-mikula-sonarsource, line 9: We should ignore the `static` configuration, as it's noisy for private static methods in our code base.
```suggestion
```

analyzers/stylecop.json
```diff
@@ -0,0 +1,19 @@
...
+        "accessibility",
+        "readonly",
+        "kind"
```
pavel-mikula-sonarsource, line 12: Open question: Does it make sense to have `kind` configured when we can't use SA1201 (yet)?

pavel-mikula-sonarsource (changes requested):

analyzers/stylecop.json
andrei-epure-sonarsource: yes, and I will also remove accessibility, because we can have `private static` at the top

pavel-mikula-sonarsource (approved):
LGTM

PR #4477, opened 2021-05-26
Rule S4005: Add support for target typed new expressions

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseUriInsteadOfString.CSharp9.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseUriInsteadOfString.CSharp9.cs
@@ -1,15 +1,20 @@
-﻿S3994 p = new("www.sonarsource.com"); // Compliant - FN
+﻿using System;
+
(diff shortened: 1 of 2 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseUriInsteadOfString.CSharp9.cs
```diff
@@ -1,15 +1,20 @@
...
+S3994 p = new("www.sonarsource.com"); // Noncompliant
+
+S3994 q = new("www.sonarsource.com", false); // Compliant
 
 string GetUrl(string url) => ""; // Compliant - FN
 
 public record S3994
 {
-    public S3994(string uri) { } // Noncompliant {{Either change this parameter type to 'System.Uri' or provide an overload which takes a 'System.Uri' parameter.}}
+    public S3994(string uri) { }
```
pavel-mikula-sonarsource, line 11: You've removed one of the old test cases for purpose of the new one. Can you create a new record just with this constructor to preserve the old test case?

pavel-mikula-sonarsource (approved):
LGTM, let's restore the test case before merging

PR #4480, opened 2021-05-28
Configure SA11xx rules

pavel-mikula-sonarsource:


.editorconfig
```diff
@@ -267,6 +267,49 @@ dotnet_diagnostic.SA1026.severity = warning
...
+dotnet_diagnostic.SA1122.severity = warning
+dotnet_diagnostic.SA1123.severity = warning
+dotnet_diagnostic.SA1124.severity = none    # We need regions sometimes
+dotnet_diagnostic.SA1125.severity = warning
+dotnet_diagnostic.SA1126.severity = none    # Deprecated / not implemented rule
+dotnet_diagnostic.SA1127.severity = none    # Noisy for single-line method declarations
+dotnet_diagnostic.SA1128.severity = none    # Doesn't match our code base
+dotnet_diagnostic.SA1129.severity = warning
+dotnet_diagnostic.SA1130.severity = warning
+dotnet_diagnostic.SA1131.severity = warning
```
andrei-epure-sonarsource, line 303: I am curious on what this will find in our current codebase - we should also adapt our coding style to reflect this, right?

.editorconfig
```diff
@@ -267,6 +267,49 @@ dotnet_diagnostic.SA1026.severity = warning
...
+dotnet_diagnostic.SA1132.severity = warning
+dotnet_diagnostic.SA1133.severity = warning
+dotnet_diagnostic.SA1134.severity = warning
+dotnet_diagnostic.SA1135.severity = none    # Noisy for generics
```
andrei-epure-sonarsource, line 307: > A violation of this rule occurs when a using directive is contained within a namespace and is not qualified.

but we don't put usings inside namespaces at all, right? maybe update the comment, to reflect this?

andrei-epure-sonarsource (changes requested):
nice! this will help us a lot 😄 

I left some comments

.editorconfig
```diff
@@ -267,6 +267,49 @@ dotnet_diagnostic.SA1026.severity = warning
...
+
+dotnet_diagnostic.SA1100.severity = warning
+dotnet_diagnostic.SA1101.severity = none    # Doesn't match our coding style, we don't use "this." when not neded
+dotnet_diagnostic.SA1102.severity = warning
+dotnet_diagnostic.SA1103.severity = warning
+dotnet_diagnostic.SA1104.severity = warning
+dotnet_diagnostic.SA1105.severity = warning
+dotnet_diagnostic.SA1106.severity = warning
+dotnet_diagnostic.SA1107.severity = warning
+dotnet_diagnostic.SA1108.severity = none    # Noisy for short comments on short lines (if, for, foreach)
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: https://github.com/SonarSource/sonar-dotnet/blob/36b38ad5a4f08f580e3ccd90e0b1d960d3d2c701/analyzers/src/SonarAnalyzer.Common/Trackers/ConstantValueFinder.cs#L51

https://github.com/SonarSource/sonar-dotnet/blob/36b38ad5a4f08f580e3ccd90e0b1d960d3d2c701/analyzers/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs#L246

.editorconfig
pavel-mikula-sonarsource: Quite a few actually, as we had C++ and Java devs contributing and something got copy-paste spreaded :)
https://github.com/SonarSource/sonar-dotnet/blob/36b38ad5a4f08f580e3ccd90e0b1d960d3d2c701/analyzers/src/SonarAnalyzer.CSharp/CBDE/MlirExporter.cs#L289

https://github.com/SonarSource/sonar-dotnet/blob/36b38ad5a4f08f580e3ccd90e0b1d960d3d2c701/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/CookieShouldBeSecure.cs#L45

.editorconfig
```diff
@@ -267,6 +267,49 @@ dotnet_diagnostic.SA1026.severity = warning
...
 dotnet_diagnostic.SA1028.severity = warning
 
+# Readability Rules (SA1100-) https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/1.1.118/DOCUMENTATION.md
+
+dotnet_diagnostic.SA1100.severity = warning
+dotnet_diagnostic.SA1101.severity = none    # Doesn't match our coding style, we don't use "this." when not neded
+dotnet_diagnostic.SA1102.severity = warning
+dotnet_diagnostic.SA1103.severity = warning
+dotnet_diagnostic.SA1104.severity = warning
+dotnet_diagnostic.SA1105.severity = warning
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: We don't use it, but I left them in case somebody ever uses it in the future. They should not cost us anything, as the syntax will not be present.

.editorconfig
pavel-mikula-sonarsource: This rule goes crazy on some special cases that I find useful:

https://github.com/SonarSource/sonar-dotnet/blob/36b38ad5a4f08f580e3ccd90e0b1d960d3d2c701/analyzers/src/SonarAnalyzer.CSharp/Rules/FieldShouldBeReadonly.cs#L35-L36

https://github.com/SonarSource/sonar-dotnet/blob/36b38ad5a4f08f580e3ccd90e0b1d960d3d2c701/analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.cs#L38-L39

pavel-mikula-sonarsource: It's not fully qualified, but it's way more readable than the fully qualified version.

.editorconfig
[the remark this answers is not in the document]
andrei-epure-sonarsource: ah ok so it raised also when it's on the same line

.editorconfig
andrei-epure-sonarsource: before enabling this one, shouldn't we discuss it and add it to our coding style? 

I agree to enable it and have a consistent way of doing it, but I am thinking that I myself might have added sometimes constants on the left, so I'd favor a discussion before to make sure we're all aligned before this pops up on PRs

 (so I suggest to disable it in this PR until we discuss)

.editorconfig
andrei-epure-sonarsource: Ok. This is an interesting pattern, I haven't seen it before I think. I agree

andrei-epure-sonarsource (approved):
LGTM

RE: SA1131 - I would still favor discussing it before enabling it

.editorconfig
pavel-mikula-sonarsource: This one is quite obvious/no brainer even without discussion, as we're doing it like that anyway. The rule will just help with that so I'm for merging it as is.

andrei-epure-sonarsource: if you're confident we're aligned on this, please update the coding style as well

andrei-epure-sonarsource: I'm not sure "coding style" is "obvious"

PR #4482, opened 2021-05-28
Rule S3257: Add support for target typed new expressions

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
@@ -37,26 +37,12 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class RedundantDeclaration : SonarDiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S3257";
+        internal const string DiagnosticTypeKey = "diagnosticType";
         private const string MessageFormat = "Remove the {0}; it is redundant.";
 
+        private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
-
-        internal const string DiagnosticTypeKey = "diagnosticType";
-
-        internal enum RedundancyType
-        {
-            LambdaParameterType,
-            ArraySize,
-            ArrayType,
-            ExplicitDelegate,
-            ExplicitNullable,
-            ObjectInitializer,
-            DelegateParameterList
-        }
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
         protected override void Initialize(SonarAnalysisContext context)
         {
@@ -103,33 +89,29 @@ namespace SonarAnalyzer.Rules.CSharp
                 return;
             }
 
-            var newParameterList = SyntaxFactory.ParameterList(
-                SyntaxFactory.SeparatedList(lambda.ParameterList.Parameters.Select(p => SyntaxFactory.Parameter(p.Identifier))));
+            var newParameterList = SyntaxFactory.ParameterList(SyntaxFactory.SeparatedList(lambda.ParameterList.Parameters.Select(p => SyntaxFactory.Parameter(p.Identifier))));
             var newLambda = lambda.WithParameterList(newParameterList);
 
             newLambda = ChangeSyntaxElement(lambda, newLambda, context.SemanticModel, out var newSemanticModel);
 
-            if (!(newSemanticModel.GetSymbolInfo(newLambda).Symbol is IMethodSymbol newSymbol) ||
-                ParameterTypesDoNotMatch(symbol, newSymbol))
+            if (!(newSemanticModel.GetSymbolInfo(newLambda).Symbol is IMethodSymbol newSymbol) || ParameterTypesDoNotMatch(symbol, newSymbol))
             {
                 return;
             }
 
             foreach (var parameter in lambda.ParameterList.Parameters)
             {
-                context.ReportDiagnosticWhenActive(Diagnostic.Create(rule, parameter.Type.GetLocation(),
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, parameter.Type.GetLocation(),
                     ImmutableDictionary<string, string>.Empty.Add(DiagnosticTypeKey, RedundancyType.LambdaParameterType.ToString()),
                     "type specification"));
             }
         }
 
-        private static bool IsParameterListModifiable(ParenthesizedLambdaExpressionSyntax lambda)
-        {
-            return lambda.ParameterList != null &&
-                lambda.ParameterList.Parameters.All(
-                    p => p.Type != null &&
-                    p.Modifiers.All(m => !RefOutKeywords.Contains(m.Kind())));
-        }
+        private static bool IsParameterListModifiable(ParenthesizedLambdaExpressionSyntax lambda) =>
+            lambda.ParameterList != null
+            && lambda.ParameterList.Parameters.All(
+                p => p.Type != null
+                     && p.Modifiers.All(m => !RefOutKeywords.Contains(m.Kind())));
 
         private static bool ParameterTypesDoNotMatch(IMethodSymbol method1, IMethodSymbol method2)
         {
@@ -155,9 +137,9 @@ namespace SonarAnalyzer.Rules.CSharp
                 return;
             }
 
-            if (IsInNotVarDeclaration(objectCreation) ||
-                IsInAssignmentOrReturnValue(objectCreation) ||
-                IsInArgumentAndCanBeChanged(objectCreation, context.SemanticModel))
+            if (IsInNotVarDeclaration(objectCreation)
+                || IsInAssignmentOrReturnValue(objectCreation)
+                || IsInArgumentAndCanBeChanged(objectCreation, context.SemanticModel))
             {
                 ReportIssueOnRedundantObjectCreation(context, objectCreation, "explicit nullable type creation", RedundancyType.ExplicitNullable);
             }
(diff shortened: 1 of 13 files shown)


pavel-mikula-sonarsource (changes requested):
Round one: Only the refactoring commit

analyzers/src/SonarAnalyzer.CSharp/Wrappers/ObjectCreationFactory.cs
```diff
@@ -31,6 +31,7 @@ public interface IObjectCreation
         ArgumentListSyntax ArgumentList { get; }
         ExpressionSyntax Expression { get; }
         string TypeAsString(SemanticModel semanticModel);
+        ISymbol TypeSymbol(SemanticModel semanticModel);
```
pavel-mikula-sonarsource, line 34: Aren't there one or two other recent PRs that can benefit from this functionality as well?

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
```diff
@@ -149,49 +133,45 @@ private static bool ParameterTypesDoNotMatch(IMethodSymbol method1, IMethodSymbo
...
-            return parent is AssignmentExpressionSyntax ||
-                parent is ReturnStatementSyntax ||
-                parent is LambdaExpressionSyntax;
+            return parent is AssignmentExpressionSyntax
+                   || parent is ReturnStatementSyntax
+                   || parent is LambdaExpressionSyntax;
         }
 
-        private static bool IsInNotVarDeclaration(ObjectCreationExpressionSyntax objectCreation)
+        private static bool IsInNotVarDeclaration(SyntaxNode objectCreation)
```
pavel-mikula-sonarsource, line 169: We need a C# 9 specific compliant test case that will enter this section and find variable declaration, because it's using the `.Parent.Parent.Parent`. It should work if I did the finger counting correctly, but let's be safe.

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
```diff
@@ -270,16 +248,15 @@ private static void ReportRedundantArrayTypeSpecifier(SyntaxNodeAnalysisContext
...
-            var objectCreation = (ObjectCreationExpressionSyntax)context.Node;
+            var objectCreation = ObjectCreationFactory.Create(context.Node);
             if (objectCreation.ArgumentList == null)
             {
                 return;
             }
 
-            if (objectCreation.Initializer != null &&
-                !objectCreation.Initializer.Expressions.Any())
+            if (objectCreation.Initializer != null && !objectCreation.Initializer.Expressions.Any())
```
pavel-mikula-sonarsource, line 257: Let's join the conditions of this and the previous `if` into a single `if` statement.

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
```diff
@@ -291,66 +268,56 @@ private static void ReportOnRedundantObjectInitializer(SyntaxNodeAnalysisContext
...
 
-        private static bool IsDynamic(ArgumentSyntax argument, SemanticModel semanticModel)
-        {
-            return semanticModel.GetTypeInfo(argument.Expression).Type is IDynamicTypeSymbol;
-        }
+        private static bool IsDynamic(ArgumentSyntax argument, SemanticModel semanticModel) =>
+            semanticModel.GetTypeInfo(argument.Expression).Type is IDynamicTypeSymbol;
 
-        private static bool IsInDeclarationNotVarNotDelegate(ObjectCreationExpressionSyntax objectCreation, SemanticModel semanticModel)
+        private static bool IsInDeclarationNotVarNotDelegate(SyntaxNode objectCreation, SemanticModel semanticModel)
```
pavel-mikula-sonarsource, line 296: Similar as the previous: Let's add a C# 9 compliant test case that will enter this section and find the variable declarations. Just to be sure it works as expected

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
```diff
@@ -407,18 +371,18 @@ private static void ReportOnRedundantParameterList(SyntaxNodeAnalysisContext con
...
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, anonymousMethod.ParameterList.GetLocation(),
                     ImmutableDictionary<string, string>.Empty.Add(DiagnosticTypeKey, RedundancyType.DelegateParameterList.ToString()),
                     "parameter list"));
             }
         }
 
         #endregion
 
-        private static bool IsInArgumentAndCanBeChanged(ObjectCreationExpressionSyntax objectCreation, SemanticModel semanticModel,
+        private static bool IsInArgumentAndCanBeChanged(IObjectCreation objectCreation, SemanticModel semanticModel,
```
pavel-mikula-sonarsource, line 382: Could be single line

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
```diff
@@ -445,15 +409,17 @@ private static void ReportOnRedundantParameterList(SyntaxNodeAnalysisContext con
...
 
-            return newSemanticModel.GetSymbolInfo(newInvocation).Symbol is IMethodSymbol newMethodSymbol &&
-                methodSymbol.ToDisplayString() == newMethodSymbol.ToDisplayString();
+            return newSemanticModel.GetSymbolInfo(newInvocation).Symbol is IMethodSymbol newMethodSymbol
+                   && methodSymbol.ToDisplayString() == newMethodSymbol.ToDisplayString();
         }
 
         private static void ReportIssueOnRedundantObjectCreation(SyntaxNodeAnalysisContext context,
-            ObjectCreationExpressionSyntax objectCreation, string message, RedundancyType redundancyType)
+            SyntaxNode node, string message, RedundancyType redundancyType)
```
pavel-mikula-sonarsource, line 417: Method declaration could be single line

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
```diff
@@ -291,66 +268,56 @@ private static void ReportOnRedundantObjectInitializer(SyntaxNodeAnalysisContext
...
+            if (type == null || type.IsVar)
             {
                 return false;
             }
 
             var typeInformation = semanticModel.GetTypeInfo(type).Type;
 
-            return typeInformation != null &&
-                !typeInformation.Is(KnownType.System_Delegate);
+            return typeInformation != null && !typeInformation.Is(KnownType.System_Delegate);
```
pavel-mikula-sonarsource, line 310: Coverage: `typeInformation` will be `ErrorType` and will never be null if I'm not mistaken. Let's remove it to bump coverage and add a test case.

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
```diff
@@ -375,8 +340,7 @@ private static bool IsDelegateCreation(ObjectCreationExpressionSyntax objectCrea
 
             var typeInformation = semanticModel.GetTypeInfo(assignment.Left).Type;
 
-            return typeInformation != null &&
-                !typeInformation.Is(KnownType.System_Delegate);
+            return typeInformation != null && !typeInformation.Is(KnownType.System_Delegate);
```
pavel-mikula-sonarsource, line 343: _Coverage:_ `typeInformation` cannot be null (if I'm not mistaken)

pavel-mikula-sonarsource (changes requested):
Second round for the rest

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
```diff
@@ -149,49 +133,45 @@ private static bool ParameterTypesDoNotMatch(IMethodSymbol method1, IMethodSymbo
...
 
-
-        private static bool IsInAssignmentOrReturnValue(ObjectCreationExpressionSyntax objectCreation)
+        private static bool IsInAssignmentOrReturnValue(SyntaxNode objectCreation)
         {
             var parent = objectCreation.GetFirstNonParenthesizedParent();
-            return parent is AssignmentExpressionSyntax ||
-                parent is ReturnStatementSyntax ||
-                parent is LambdaExpressionSyntax;
+            return parent is AssignmentExpressionSyntax
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: No strict opinion on this one. Converted it to switch

analyzers/src/SonarAnalyzer.CSharp/Wrappers/ObjectCreationFactory.cs
dmitry-golovinov-sonarsource: Many of them use CSharpObjectInitializationTracker.IsTrackedType, which is also used by non-object creation assignments to check if the type is being tracked. Fixed in a few places I could

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantDeclaration.CSharp9.cs
```diff
@@ -10,6 +10,10 @@
 Record Local() => new Record(); // Compliant - FN
 static Record LocalStatic(Record r) => r;
 
+int? xx = ((new int?(5))); // Noncompliant {{Remove the explicit nullable type creation; it is redundant.}}
+
+var myEvent = new EventHandler((a, b) => { });
```
pavel-mikula-sonarsource, line 15: Shouldn't this be C# 9 alternative?
```suggestion
EventHandler myEvent = new ((a, b) => { });
```

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
```diff
@@ -149,49 +141,38 @@ private static bool ParameterTypesDoNotMatch(IMethodSymbol method1, IMethodSymbo
...
-            return parent is AssignmentExpressionSyntax ||
-                parent is ReturnStatementSyntax ||
-                parent is LambdaExpressionSyntax;
-        }
+        private static bool IsInAssignmentOrReturnValue(SyntaxNode objectCreation) =>
+            objectCreation.GetFirstNonParenthesizedParent() switch
+            {
+                AssignmentExpressionSyntax assignment => true,
+                ReturnStatementSyntax @return => true,
+                LambdaExpressionSyntax lambda => true,
```
pavel-mikula-sonarsource, line 166: We don't need to name them, please use discard to avoid unused variables. We should have a rule about this (or update existing one)
```suggestion
                AssignmentExpressionSyntax _ => true,
                ReturnStatementSyntax _ => true,
                LambdaExpressionSyntax _ => true,
```

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
pavel-mikula-sonarsource: This one doesn't hit the breakpoint for the new `new()` syntax, only for the old one - probably related to the comment in C# 9 test case.

pavel-mikula-sonarsource (changes requested):
Final polishing

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
dmitry-golovinov-sonarsource: Fixed to _. QG didn't complain about that.

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantDeclaration.cs
dmitry-golovinov-sonarsource: Verified that we stop on this case now with implicit syntax.

pavel-mikula-sonarsource (approved):
LGTM

PR #4485, opened 2021-05-31
Rule S3453: Add support for target typed new expressions

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.Common/Rules/ClassNotInstantiatableBase.cs
```diff
@@ -31,20 +31,18 @@ public abstract class ClassNotInstantiatableBase : SonarDiagnosticAnalyzer
         protected const string MessageFormat = "This {0} can't be instantiated; make {1} 'public'.";
 
         protected abstract bool IsTypeDeclaration(SyntaxNode node);
+        protected abstract bool IsObjectCreation(SyntaxNode node);
```
pavel-mikula-sonarsource, line 34: Let's say you provide `ILanguageFacade` instead to this class. Like that, you can use `IsAnyKind` (from the facade) with the `ObjectCreationSyntaxKinds`.

analyzers/src/SonarAnalyzer.CSharp/Rules/ClassNotInstantiatable.cs
```diff
@@ -45,6 +43,9 @@ public sealed class ClassNotInstantiatable : ClassNotInstantiatableBase
         protected override bool IsTypeDeclaration(SyntaxNode node) =>
             node.IsAnyKind(SyntaxKind.ClassDeclaration, SyntaxKindEx.RecordDeclaration);
 
+        protected override bool IsObjectCreation(SyntaxNode node) =>
+            node.IsAnyKind(SyntaxKind.ObjectCreationExpression, SyntaxKindEx.ImplicitObjectCreationExpression);
+
         private void CheckClassWithOnlyUnusedPrivateConstructors(SymbolAnalysisContext context)
```
pavel-mikula-sonarsource, line 49: Let's refactor this rule to remove duplications. 
`Initialize` can be moved to base class.
`Rule` and `SupportedDiagnostics` can be moved to base class (rule instance will be created in the constructor)
`CheckClassWithOnlyUnusedPrivateConstructors` can be moved to base class.

There will be just few minor things to deal with as overridables - `CollectRemovableDeclarations` - it can return pairs of `name` and `typeDeclaration` for the error message.

pavel-mikula-sonarsource: See `DoNotLockWeakIdentityObjectsBase` (for example) for inspiration/snippets

analyzers/src/SonarAnalyzer.Common/Rules/ClassNotInstantiatableBase.cs
```diff
@@ -31,20 +31,18 @@ public abstract class ClassNotInstantiatableBase : SonarDiagnosticAnalyzer
...
+            where TBaseTypeSyntax : SyntaxNode =>
             typeDeclarations
                 .Select(typeDeclaration => new
                 {
                     typeDeclaration.SemanticModel,
                     DescendantNodes = typeDeclaration.SyntaxNode.DescendantNodes().ToList()
                 })
                 .Any(descendants =>
-                    IsAnyConstructorToCurrentType<TObjectCreationSyntax>(descendants.DescendantNodes, namedType, descendants.SemanticModel) ||
+                    IsAnyConstructorToCurrentType(descendants.DescendantNodes, namedType, descendants.SemanticModel) ||
```
pavel-mikula-sonarsource, line 45: Let's fix the `||` formatting, it's modified line and it's the only occurrence in this file

pavel-mikula-sonarsource (changes requested):
Let's refactor the rule structure

analyzers/src/SonarAnalyzer.CSharp/Rules/ClassNotInstantiatable.cs
```diff
@@ -31,54 +33,33 @@ namespace SonarAnalyzer.Rules.CSharp
...
+    public sealed class ClassNotInstantiatable : ClassNotInstantiatableBase<BaseTypeDeclarationSyntax, SyntaxKind>
     {
-        private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
-
-        protected override void Initialize(SonarAnalysisContext context) =>
-            context.RegisterSymbolAction(CheckClassWithOnlyUnusedPrivateConstructors, SymbolKind.NamedType);
+        protected override ILanguageFacade<SyntaxKind> Language => CSharpFacade.Instance;
+        protected override GeneratedCodeRecognizer CodeRecognizer => CSharpGeneratedCodeRecognizer.Instance;
```
pavel-mikula-sonarsource, line 39: This is available in `Language.GeneratedCodeRecognizer`

analyzers/src/SonarAnalyzer.CSharp/Rules/ClassNotInstantiatable.cs
```diff
@@ -31,54 +33,33 @@ namespace SonarAnalyzer.Rules.CSharp
...
-                {
-                    var declarationKind = typeDeclaration.SyntaxNode.IsKind(SyntaxKind.ClassDeclaration)
-                        ? "class"
-                        : "record";
-                    context.ReportDiagnosticIfNonGenerated(
-                        Diagnostic.Create(Rule, typeDeclaration.SyntaxNode.Identifier.GetLocation(), declarationKind, message));
-                }
-            }
+            return typeDeclarations
+                .Select(x => new Tuple<SyntaxNodeAndSemanticModel<BaseTypeDeclarationSyntax>, Diagnostic>(x, Diagnostic.Create(rule,
```
pavel-mikula-sonarsource, line 54: The diagnostics can be created in the base class. I wanted to return pairs of `(x, DeclarationKind(x.SyntaxNode))`

analyzers/src/SonarAnalyzer.CSharp/Rules/ClassNotInstantiatable.cs
```diff
@@ -31,54 +33,33 @@ namespace SonarAnalyzer.Rules.CSharp
...
+                                                                                                                               x.SyntaxNode.Identifier.GetLocation(),
+                                                                                                                               DeclarationKind(x.SyntaxNode),
+                                                                                                                               message)));
         }
+
+        private static string DeclarationKind(SyntaxNode node) =>
+            node.IsKind(SyntaxKind.ClassDeclaration)
```
pavel-mikula-sonarsource, line 61: This is as short and simple that it can be single line. But I'd swap the logic to fall back to `class` in case something new and unexpected appears

analyzers/src/SonarAnalyzer.CSharp/Rules/ClassNotInstantiatable.cs
dmitry-golovinov-sonarsource: There are different ways how to craft the location for VB and C#, so I cannot unite them like this.

pavel-mikula-sonarsource: :(

analyzers/src/SonarAnalyzer.CSharp/Rules/ClassNotInstantiatable.cs
pavel-mikula-sonarsource: I also mentioned that it can be single line, because it's short and simple

analyzers/src/SonarAnalyzer.CSharp/Rules/ClassNotInstantiatable.cs
```diff
@@ -32,52 +33,32 @@ namespace SonarAnalyzer.Rules.CSharp
...
-                {
-                    var declarationKind = typeDeclaration.SyntaxNode.IsKind(SyntaxKind.ClassDeclaration)
-                        ? "class"
-                        : "record";
-                    context.ReportDiagnosticIfNonGenerated(
-                        Diagnostic.Create(Rule, typeDeclaration.SyntaxNode.Identifier.GetLocation(), declarationKind, message));
-                }
-            }
+            return typeDeclarations
+                .Select(x => new Tuple<SyntaxNodeAndSemanticModel<BaseTypeDeclarationSyntax>, Diagnostic>(x, Diagnostic.Create(rule,
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Well, it actually doesn't have to be tuple. We can just go with a class with two readonly properties and constructor. As `.Item2` doesn't look nice.

pavel-mikula-sonarsource (changes requested):
Let's remove a bit more code :)

analyzers/src/SonarAnalyzer.CSharp/Rules/ClassNotInstantiatable.cs
```diff
@@ -32,52 +33,32 @@ namespace SonarAnalyzer.Rules.CSharp
...
-
-            var members = namedType.GetMembers();
-            var constructors = GetConstructors(members).Where(x => !x.IsImplicitlyDeclared).ToList();
+            var typeDeclarations = new CSharpRemovableDeclarationCollector(namedType, compilation).TypeDeclarations;
 
-            if (!HasOnlyCandidateConstructors(constructors) || HasOnlyStaticMembers(members.Except(constructors).ToList()))
-            {
-                return;
-            }
+            var message = count > 1
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I wanted the argument to be `string` and to move the ternary operator logic to the base class. The `int count` name wasn't an issue :D

analyzers/src/SonarAnalyzer.Common/Facade/ILanguageFacade.cs
```diff
@@ -41,5 +41,6 @@ public interface ILanguageFacade<TSyntaxKind> : ILanguageFacade
         SyntaxFacade<TSyntaxKind> Syntax { get; }
         ISyntaxKindFacade<TSyntaxKind> SyntaxKind { get; }
         ITrackerFacade<TSyntaxKind> Tracker { get; }
+        bool IsAnyKind(SyntaxNode node, params TSyntaxKind[] syntaxKinds);
```
pavel-mikula-sonarsource, line 44: This should be in `SyntaxFacade` type, right next to the existing `IsAnyKind` overload.

analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxKindFacade.cs
```diff
@@ -27,6 +27,7 @@ internal sealed class CSharpSyntaxKindFacade : ISyntaxKindFacade<SyntaxKind>
     {
         public SyntaxKind InvocationExpression => SyntaxKind.InvocationExpression;
         public SyntaxKind[] ObjectCreationExpression => new[] {SyntaxKind.ObjectCreationExpression, SyntaxKindEx.ImplicitObjectCreationExpression};
+        public SyntaxKind[] ClassDeclaration => new[] {SyntaxKind.ClassDeclaration, SyntaxKindEx.RecordDeclaration};
```
pavel-mikula-sonarsource, line 30: Let's revert this. Member `ClassDeclaration` should return only `SyntaxKind.ClassDeclaration` for those, that would like explicitly subscribe to that single syntax kind.

Let's:
* Use the original approach with abstract method in the rule base, if this is not reusable
* Add new `SyntaxKind[] ClassAndRecordDeclaration` if we can use it at least 3 times directly in existing rules that already contains `ILanguageFacade`.

pavel-mikula-sonarsource (changes requested):
Final rearrangement

analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxKindFacade.cs
dmitry-golovinov-sonarsource: Actually, we don't have to do 1) if we do 2). I've found 2 other rules that benefited from that change

analyzers/src/SonarAnalyzer.Common/Rules/ClassNotInstantiatableBase.cs
```diff
@@ -18,56 +18,79 @@
...
+    public abstract class ClassNotInstantiatableBase<TBaseTypeSyntax, TSyntaxKind> : SonarDiagnosticAnalyzer
+        where TBaseTypeSyntax : SyntaxNode
+        where TSyntaxKind : struct
     {
         protected const string DiagnosticId = "S3453";
-        protected const string MessageFormat = "This {0} can't be instantiated; make {1} 'public'.";
+        private const string MessageFormat = "This {0} can't be instantiated; make {1} 'public'.";
+
+        protected readonly DiagnosticDescriptor rule;
+        private readonly HashSet<TSyntaxKind> objectCreationSyntaxKinds;
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: The key of this update was to remove this field :) 
Constructor can be arrow method then.
The IsAnyKind will directly ask for the syntax kinds and will not have the `ToHashSet`

pavel-mikula-sonarsource (approved):
LGTM. There are few last lines to delete before merging (see the last open thread above)

Issue #4223, opened 2021-04-09
Java ITs: Migrate to jUnit 5


andrei-epure-sonarsource:
In order to easily do parameterized tests and benefit of better test APIs. 

E.g. see [this code smell](https://sonarcloud.io/project/issues?id=org.sonarsource.dotnet%3Asonar-dotnet&issues=AXTHxlaBaN_sgxWhE7U2&open=AXTHxlaBaN_sgxWhE7U2)

* [x] Bump dependency to v5
* [x] Update ITs to jUnit5
* [x] Solve Orchestrator startup
* [x] Remove suite

pavel-mikula-sonarsource:
This is now required to bump Orchestrator to 4.x series, see #7274

PR #4487, opened 2021-05-31
Fix RegEx hotspots in DotCoverReportParser v2

### LINKED ISSUE #4194: Fix RegEx hotspots in DotCoverReportParser

pavel-mikula-sonarsource:
Fixes #4194

First commit is a cherry-pick from the original #4212 
Second commits fixes the SO issue by removing the regex.


sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/DotCoverReportParserTest.java
```diff
@@ -101,6 +147,18 @@ public void valid() throws Exception {
...
 
+  @Test
+  public void valid_big() throws Exception {
+    Coverage coverage = new Coverage();
+    new DotCoverReportParser(alwaysTrue).accept(new File("src/test/resources/dotcover/valid_big.html"), coverage);
+
+    String filePath = new File("mylibrary\\calc.cs").getCanonicalPath();
+    assertThat(coverage.files()).containsOnly(filePath);
+    assertThat(coverage.hits(filePath)).hasSize(10000);
+    assertThat(logTester.logs(LoggerLevel.INFO).get(0)).startsWith("Parsing the dotCover report ");
```
andrei-epure-sonarsource, line 158: could you check the number of `TRACE` statements? and also the number of `INFO` statements (to be clear if there's only one or more TRACE/INFO log lines)

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/DotCoverReportParserTest.java
```diff
@@ -101,6 +147,18 @@ public void valid() throws Exception {
       .startsWith("dotCover parser: found coverage for line '12', hits '0' when analyzing the path '");
   }
 
+  @Test
+  public void valid_big() throws Exception {
+    Coverage coverage = new Coverage();
+    new DotCoverReportParser(alwaysTrue).accept(new File("src/test/resources/dotcover/valid_big.html"), coverage);
```
andrei-epure-sonarsource, line 153: do you think it's a good idea to, inside the `valid_big.html` file, replace the contents inside `Add10000()` with a placeholder, and generate the `value = value + 1;` statements here in the test? and create a temporary file from the template + code?

because having a 10K LOC file is difficult to handle: 
-  both to load (the IDE + plugins may be very slow), also github
- and to review historical diffs

andrei-epure-sonarsource (changes requested):

sonar-dotnet-shared-library/src/test/java/org/sonar/plugins/dotnet/tests/DotCoverReportParserTest.java
pavel-mikula-sonarsource: The file review is one-off task. I don't expect any changes there in a future. So I don't think we need complicated scaffolding in this case. The plain file is simple and 100% sure to be big.

pavel-mikula-sonarsource:
I tried to fix the code smell, but introducing parameterized tests goes into a Java dependency hell rabbit hole 🤯 So I'm not able to do it in reasonable time.

andrei-epure-sonarsource:
RE: parameterized - JUnit4 parameterized tests are messy. If we are to invest time, we should upgrade to JUnit5 which makes it as easy as in .NET to write parameterized tests. We have an issue for it: https://github.com/SonarSource/sonar-dotnet/issues/4223

andrei-epure-sonarsource (approved):
LGTM

PR #4494, opened 2021-06-02
Fix S2068 FP: Don't propagate constants through self-modifying assignments

### LINKED ISSUE #4493: Fix S2068 FP: Don't propagate constants through self-modifying assignments
Current AssignmentFinder doesn't distinguish between `=` and other `+=`, `-=`, `|=`... operators when tracking constant values. Value modified with `+=` operator is not constant anymore.
```
arg += "Literal";
var a = "Server = localhost; Database = Test; User = SA; Password = " + arg;                    // Noncompliant FP
```

```
Arg &= "Literal"
Dim a As String = "Server = localhost; Database = Test; User = SA; Password = " & Arg                     ' Noncompliant FP
```

pavel-mikula-sonarsource:
Fixes #4493

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Trackers/CSharpAssignmentFinder.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Trackers/CSharpAssignmentFinder.cs
@@ -30,13 +30,15 @@ namespace SonarAnalyzer.Helpers
         protected override SyntaxNode GetTopMostContainingMethod(SyntaxNode node) =>
             node.GetTopMostContainingMethod();
 
-        protected override bool IsAssignmentToIdentifier(SyntaxNode node, string identifierName, out SyntaxNode rightExpression)
+        protected override bool IsAssignmentToIdentifier(SyntaxNode node, string identifierName, bool anyAssignmentKind, out SyntaxNode rightExpression)
         {
             if (node is ExpressionStatementSyntax statement)
             {
                 node = statement.Expression;
             }
-            if (node is AssignmentExpressionSyntax assignment && assignment.Left.NameIs(identifierName))
+            if ((anyAssignmentKind || node.IsKind(SyntaxKind.SimpleAssignmentExpression))
+                && node is AssignmentExpressionSyntax assignment
+                && assignment.Left.NameIs(identifierName))
             {
                 rightExpression = assignment.Right;
                 return true;
--- a/analyzers/src/SonarAnalyzer.CSharp/Trackers/CSharpAssignmentFinder.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Trackers/CSharpAssignmentFinder.cs
@@ -30,6 +30,7 @@ namespace SonarAnalyzer.Helpers
         protected override SyntaxNode GetTopMostContainingMethod(SyntaxNode node) =>
             node.GetTopMostContainingMethod();
 
+        /// <param name="anyAssignmentKind">'true' will find any AssignmentExpressionSyntax like =, +=, -=, &=. 'false' will find only '=' SimpleAssignmentExpression.</param>
         protected override bool IsAssignmentToIdentifier(SyntaxNode node, string identifierName, bool anyAssignmentKind, out SyntaxNode rightExpression)
         {
             if (node is ExpressionStatementSyntax statement)
--- a/analyzers/src/SonarAnalyzer.Common/Trackers/AssignmentFinder.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Trackers/AssignmentFinder.cs
@@ -27,7 +27,7 @@ namespace SonarAnalyzer.Helpers
     public abstract class AssignmentFinder
     {
         protected abstract SyntaxNode GetTopMostContainingMethod(SyntaxNode node);
-        protected abstract bool IsAssignmentToIdentifier(SyntaxNode node, string identifierName, out SyntaxNode rightExpression);
+        protected abstract bool IsAssignmentToIdentifier(SyntaxNode node, string identifierName, bool anyAssignmentKind, out SyntaxNode rightExpression);
         protected abstract bool IsIdentifierDeclaration(SyntaxNode node, string identifierName, out SyntaxNode initializer);
         protected abstract bool IsLoop(SyntaxNode node);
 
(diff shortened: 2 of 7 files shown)


pavel-mikula-sonarsource:
Rebased

analyzers/src/SonarAnalyzer.CSharp/Trackers/CSharpAssignmentFinder.cs
```diff
@@ -30,13 +30,15 @@ public class CSharpAssignmentFinder : AssignmentFinder
...
 
-        protected override bool IsAssignmentToIdentifier(SyntaxNode node, string identifierName, out SyntaxNode rightExpression)
+        protected override bool IsAssignmentToIdentifier(SyntaxNode node, string identifierName, bool anyAssignmentKind, out SyntaxNode rightExpression)
         {
             if (node is ExpressionStatementSyntax statement)
             {
                 node = statement.Expression;
             }
-            if (node is AssignmentExpressionSyntax assignment && assignment.Left.NameIs(identifierName))
+            if ((anyAssignmentKind || node.IsKind(SyntaxKind.SimpleAssignmentExpression))
```
andrei-epure-sonarsource, line 39: could you add a comment above to say what this logic is for? (maybe give an example of another SyntaxKind which is an AssignmentExpressionSyntax but not simple assignment

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/DoNotHardcodeCredentials_DefaultValues.cs
```diff
@@ -121,18 +121,20 @@ public void Concatenations(string arg)
...
-            a = someMethod() + " Database = Test; User = SA; Password = " + secretConst + " suffix";                // Noncompliant
+            a = SomeMethod() + " Database = Test; User = SA; Password = " + secretConst + " suffix";                // Noncompliant
             a = "Server = localhost; Database = Test; User = SA; Password = " + secretConst + arg + " suffix";      // Noncompliant
             a = "Server = localhost; Database = Test; User = SA; Password = " + arg + secretConst + " suffix";      // Compliant
             a = secretConst + "Server = localhost; Database = Test; User = SA; Password = " + arg;                  // Compliant
-            a = "Server = localhost; Database = Test; User = SA; " + someMethod() + secretConst;                    // Compliant
+            a = "Server = localhost; Database = Test; User = SA; " + SomeMethod() + secretConst;                    // Compliant
 
             // Reassigned
+            arg += "Literal";
```
andrei-epure-sonarsource, line 131: could you please also add a test with `??=`?
```
var a = SomeMethod();
a ??= "hardcoded";
...
```

I'm not sure how this should behave. It's the opposite of initializeToConstant then assignValueFromMethod

could you check with AppSec before merging this, please? it's a Security Hotspot so it should show possibly vulnerable cases, and this looks like one.

analyzers/src/SonarAnalyzer.Common/Trackers/AssignmentFinder.cs
```diff
@@ -43,7 +43,7 @@ public SyntaxNode FindLinearPrecedingAssignmentExpression(string identifierName,
 
                 foreach (var statement in current.Parent.ChildNodes().TakeWhile(x => x != current).Reverse())
                 {
-                    if (IsAssignmentToIdentifier(statement, identifierName, out var right))
+                    if (IsAssignmentToIdentifier(statement, identifierName, false, out var right))
```
andrei-epure-sonarsource, line 46: I think with this change this method needs some comments describing what it does, and why. Some simple ones to give a direction (example) on what to look at

like give a sample of code where it will find one, and give a sample of code where it won't (depending on the assignment types)

andrei-epure-sonarsource (approved):
LGTM as implementation

it's a security hotspot, so I'd advise to check w/ appsec for the `??=`  scenario

I left suggestions for commnets

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/DoNotHardcodeCredentials_DefaultValues.cs
pavel-mikula-sonarsource: I've added a test case. The `??=` is semantically same as this:
```
string SomeMethod() => null;

var a = SomeMethod();
a += "hardcoded";
```
I don't think we need to check with appsec. This is not SE rule and such constant propagation detection wasn't available until recently. So killing the noise for this and other rules should be priority.

analyzers/src/SonarAnalyzer.Common/Trackers/AssignmentFinder.cs
pavel-mikula-sonarsource: I've added a documentation comment on the `IsAssignmentToIdentifier` method for the `anyAssignmentKind` argument. That should make it clear.

pavel-mikula-sonarsource:
Re-requesting review, I'm not sure if the comment on all 3 places is a good idea

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/DoNotHardcodeCredentials_DefaultValues.cs
andrei-epure-sonarsource: I don't think it has the same semantics. The `+=` "enhances" a non-hardcoded password (like adding a salt to it), whereas "??=" replaces sets a hardcoded value to it.

But I agree we it's the same as
```
var password = "hardcoded";
var b = someMethod();
if (b != null) 
{
  password = b;
}
```
so I drop my case.

andrei-epure-sonarsource: I drop my case especially as indeed it's out of scope to change the implementation.

I think you can send an FYI to AppSec just for them to know of this behavior (after merging this PR)

andrei-epure-sonarsource (approved):
LGTM
RE: comments in multiple places - TBH I don't know how it's best, I put an action on our next "how we work" meeting

I think it won't hurt to leave it like this, we can re-evaluate this in the future

andrei-epure-sonarsource:
RE: S3874 - I think we can merge, and also that we can keep this rule. It's a good reminder to reconsider design.

PR #4496, opened 2021-06-02
Fix S1172 FP: Should not raise for event handlers

### LINKED ISSUE #4406: Fix S1172 FP: Should not raise for event handlers
### Description

S1172 raises issue to remove parameters from event handlers where the method signature cannot be changed.

It should not apply to methods with `Handles Source.Event` syntax.

### Repro steps

```
Public Class Source

    Public Event Dirty(Sender As Object, Name As String, Count As Integer)

End Class

Public Class Consumer

    Private WithEvents fSource As Source

    Private Sub fSource_Dirty(Sender As Object, Name As String, Count As Integer) Handles fSource.Dirty 'FP
        Dim S As String = Name
    End Sub

End Class
```

pavel-mikula-sonarsource:
Fixes #4406

### DIFF
--- a/analyzers/src/SonarAnalyzer.VisualBasic/Rules/MethodParameterUnused.cs
+++ b/analyzers/src/SonarAnalyzer.VisualBasic/Rules/MethodParameterUnused.cs
@@ -36,26 +36,23 @@ namespace SonarAnalyzer.Rules.VisualBasic
     {
         private const string MessageFormat = "Remove this unused procedure parameter '{0}'.";
 
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+        private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
-        protected override void Initialize(SonarAnalysisContext context)
-        {
-            context.RegisterSyntaxNodeActionInNonGenerated(
-                c =>
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        protected override void Initialize(SonarAnalysisContext context) =>
+            context.RegisterSyntaxNodeActionInNonGenerated(c =>
                 {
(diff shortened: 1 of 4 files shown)


analyzers/src/SonarAnalyzer.VisualBasic/Rules/MethodParameterUnused.cs
```diff
@@ -66,92 +63,71 @@ protected override void Initialize(SonarAnalysisContext context)
...
         private static bool HasAnyAttribute(MethodBlockBaseSyntax method) =>
-            method.BlockStatement.AttributeLists.Count > 0;
+            method.BlockStatement.AttributeLists.Any();
 
         private static bool OnlyThrowsNotImplementedException(MethodBlockBaseSyntax method, SemanticModel semanticModel) =>
             method.Statements.Count == 1
             && method.Statements
                 .OfType<ThrowStatementSyntax>()
-                .Select(tss => tss.Expression)
+                .Select(x => x.Expression)
```
andrei-epure-sonarsource, line 108: yes here `x` is much better than `tss`, `oces` or `s` :)

analyzers/src/SonarAnalyzer.VisualBasic/Rules/MethodParameterUnused.cs
```diff
@@ -67,7 +67,6 @@ public sealed class MethodParameterUnused : MethodParameterUnusedBase
                     // Bail-out if this is not a method we want to report on (only based on symbols checks)
                     var methodSymbol = c.SemanticModel.GetDeclaredSymbol(methodBlock);
                     if (methodSymbol == null
-                        || methodSymbol.IsAbstract
```
andrei-epure-sonarsource: this seems like a change in bevahior? is it intended? shouldn't a test be added for abstract methods?

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.VisualBasic/Rules/MethodParameterUnused.cs
pavel-mikula-sonarsource: It's for coverage. This condition is unreachable because of syntax checks above. And there's case in UTs for it as well (`MustOverride`).

analyzers/src/SonarAnalyzer.VisualBasic/Rules/MethodParameterUnused.cs
pavel-mikula-sonarsource: I found `oces` especially hard to understand. Even when it became obvious when you finally find it :)

andrei-epure-sonarsource (approved):
LGTM

PR #4497, opened 2021-06-02
Fix S5944 FP: When namespace or type is same as function name

### LINKED ISSUE #4347: Fix S5944 FP: When namespace or type is same as function name
### Description

Rule raises issue when namespace is the same as function name.

Rule raises issue when return type is the same as function name.

Rule raises issue when declared type is same as function name

### Repro steps

```
Public Function Data() As Data.Container 'S5944	Do not make use of the implicit return value.
    Dim D As Data.Container    'S5944	Do not make use of the implicit return value.
End Function    

Public Function Data() As Data  'S5944	Do not make use of the implicit return value.
    Dim D As Data  'S5944	Do not make use of the implicit return value.
End Function
```

* SonarLint 4.33.1
* Visual Studio version 2019

pavel-mikula-sonarsource:
Fixes #4347

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
@@ -144,8 +144,12 @@ Public Class DoesNotApplyOn
         Dim Ret As OtherType                    ' Noncompliant FP
     End Function
 
-    Public Function NamespaceName() As NamespaceName.Something  'Noncompliant FP
-        Dim Ret As NamespaceName.Something                      'Noncompliant FP
+    Public Function NamespaceName() As NamespaceName.Something  'Compliant
+        Dim Ret As NamespaceName.Something                      'Compliant
+    End Function
+
+    Public Function Something() As NamespaceName.Something      'Compliant
+        Dim Ret As NamespaceName.Something                      'Compliant
     End Function
 
     Public Sub SomeEvent() Handles fSource.SomeEvent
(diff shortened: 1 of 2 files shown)


pavel-mikula-sonarsource:
The PR is missing [coverage info on SC](https://sonarcloud.io/summary/new_code?id=sonaranalyzer-dotnet&pullRequest=4497) for some reason

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
```diff
@@ -124,6 +139,20 @@ Public Class DoesNotApplyOn
...
+        Dim Ret As OtherType                    ' Noncompliant FP
+    End Function
+
+    Public Function NamespaceName() As NamespaceName.Something  'Noncompliant FP
+        Dim Ret As NamespaceName.Something                      'Noncompliant FP
+    End Function
+
+    Public Sub SomeEvent() Handles fSource.SomeEvent
+        Dim S As Source
+        AddHandler S.SomeEvent, AddressOf SomeEvent
```
andrei-epure-sonarsource, line 158: I don't get it - is this part of the repro?

andrei-epure-sonarsource (approved):
LGTM. I left a comment for me to better understand. Not sure it requires a change or not.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UseReturnStatement.vb
pavel-mikula-sonarsource: No, this is just another unrelated syntax that came to my mind while working on it and I wanted to be sure we don't have another FP there.

PR #4499, opened 2021-06-02
Fix S3264 perf regression

### LINKED ISSUE #4498: Fix performance S3264 perf regression
It seems that  #3932 introduced a performance regression. 

Previously `invokedSymbols` and `possiblyCopiedSymbols` were cached before the LINQ `Where()` queries, and after the change they were moved inside the methods invoked in the LINQ `Where()` queries (thus executing for each iteration).

For particular projects reported [here](https://discuss.sonarsource.com/t/chsarp-performance-problem-post-upgrade-to-8-9/7836) this lead to an increase of analysis time for each two projects (3000 seconds increase for each, i.e. 100 extra minutes).

andrei-epure-sonarsource:
Fixes  #4498

- doing the same method call inside a loop is not optimal.
- this wouldn't have lead to a huge perf increase, but `GetInvokedEventSymbols()` does a `tuple.SemanticModel.GetSymbolInfo(tuple.SyntaxNode).Symbol` inside, so it can explode if there's lots of events inside a codebase

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/UninvokedEventDeclaration.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/UninvokedEventDeclaration.cs
@@ -73,8 +73,7 @@ namespace SonarAnalyzer.Rules.CSharp
             var possiblyCopiedSymbols = GetPossiblyCopiedSymbols(removableDeclarationCollector);
 
             removableEventFields
-                .Where(x => !invokedSymbols.Contains(x.Symbol))
-                .Where(x => !possiblyCopiedSymbols.Contains(x.Symbol))
+                .Where(x => !invokedSymbols.Contains(x.Symbol) && !possiblyCopiedSymbols.Contains(x.Symbol))
                 .ToList()
                 .ForEach(x => context.ReportDiagnosticIfNonGenerated(
                     Diagnostic.Create(Rule, GetLocation(x.SyntaxNode), x.Symbol.Name)));
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/UninvokedEventDeclaration.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/UninvokedEventDeclaration.cs
@@ -69,9 +69,12 @@ namespace SonarAnalyzer.Rules.CSharp
                 return;
             }
 
+            var invokedSymbols = GetInvokedEventSymbols(removableDeclarationCollector);
+            var possiblyCopiedSymbols = GetPossiblyCopiedSymbols(removableDeclarationCollector);
+
             removableEventFields
-                .Where(IsNotInvoked)
-                .Where(IsNotCopied)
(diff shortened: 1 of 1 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/UninvokedEventDeclaration.cs
```diff
@@ -69,9 +69,12 @@ private void RaiseOnUninvokedEventDeclaration(SymbolAnalysisContext context)
...
             }
 
+            var invokedSymbols = GetInvokedEventSymbols(removableDeclarationCollector);
+            var possiblyCopiedSymbols = GetPossiblyCopiedSymbols(removableDeclarationCollector);
+
             removableEventFields
-                .Where(IsNotInvoked)
-                .Where(IsNotCopied)
+                .Where(tuple => !invokedSymbols.Contains(tuple.Symbol))
+                .Where(tuple => !possiblyCopiedSymbols.Contains(tuple.Symbol))
```
pavel-mikula-sonarsource, line 77: I'd go for `x` as it's pretty flat case and it's obvious that we're iterating `removableEventFields`. There's also `x` below in `ForEach`
```suggestion
                .Where(x => !invokedSymbols.Contains(x.Symbol))
                .Where(x => !possiblyCopiedSymbols.Contains(x.Symbol))
```

pavel-mikula-sonarsource (approved):
LGTM, nice catch

andrei-epure-sonarsource:
re-requesting review as I've merged two loops which is more than a name change (but should be fine)

analyzers/src/SonarAnalyzer.CSharp/Rules/UninvokedEventDeclaration.cs
```diff
@@ -73,8 +73,7 @@ private void RaiseOnUninvokedEventDeclaration(SymbolAnalysisContext context)
             var possiblyCopiedSymbols = GetPossiblyCopiedSymbols(removableDeclarationCollector);
 
             removableEventFields
-                .Where(x => !invokedSymbols.Contains(x.Symbol))
-                .Where(x => !possiblyCopiedSymbols.Contains(x.Symbol))
+                .Where(x => !invokedSymbols.Contains(x.Symbol) && !possiblyCopiedSymbols.Contains(x.Symbol))
```
pavel-mikula-sonarsource, line 76: _While seeing it like this:_ Do we need to different sets to do the same thing twice? Let imagine
* Variable `usedSymbols` instead
* `GetInvokedEventSymbols` to return `IEnumerable` 
* `GetPossiblyCopiedSymbols` to return `IEnumerable` 
* `usedSymbols = GetInvokedEventSymbols(..).Concat(GetPossiblyCopiedSymbols(..)).ToHashSet()`

pavel-mikula-sonarsource (approved):
Change itself LGTM, I left one more simplification suggestion to consider.

pavel-mikula-sonarsource:
:shipit:

PR #4503, opened 2021-06-03
Fix S3928 FP and FN: Parameters of local functions

### LINKED ISSUE #4423: Fix S3928 FP: Parameters of local functions
### Description

S3928 is reported inside local functions during arguments' validation.

### Repro steps

```csharp
private void Foo(int i)
{
    Bar(i);

    void Bar(int j)
    {
        throw new ArgumentOutOfRangeException(nameof(j), j, null); // <- S3928: the parameter name 'j' in not declared in the argument list
    }
}
```

### Expected behavior

The diagnostic is not reported.

### Additional information

Nothing is reported when I use `nameof(i)` (i.e. argument name from a containing method).

### Related information

* SonarAnalyzer.CSharp 8.2.0-RC1

pavel-mikula-sonarsource:
Fixes #4423

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
@@ -115,22 +115,24 @@ namespace SonarAnalyzer.Rules.CSharp
                 || ancestor is IndexerDeclarationSyntax
                 || LocalFunctionStatementSyntaxWrapper.IsInstance(ancestor));
 
-            if (node is SimpleLambdaExpressionSyntax simpleLambda)
+            return node switch
             {
-                return new[] { simpleLambda.Parameter.Identifier.ValueText };
-            }
-            else if (node is BaseMethodDeclarationSyntax method)
-            {
-                return IdentifierNames(method.ParameterList);
-            }
-            else if (node is ParenthesizedLambdaExpressionSyntax lambda)
-            {
-                return IdentifierNames(lambda.ParameterList);
-            }
-            else if (node is AccessorDeclarationSyntax accessor)
+                SimpleLambdaExpressionSyntax simpleLambda => new[] { simpleLambda.Parameter.Identifier.ValueText },
+                BaseMethodDeclarationSyntax method => IdentifierNames(method.ParameterList),
+                ParenthesizedLambdaExpressionSyntax lambda => IdentifierNames(lambda.ParameterList),
+                AccessorDeclarationSyntax accessor => AccessorIdentifierNames(accessor),
+                IndexerDeclarationSyntax indexerDeclaration => IdentifierNames(indexerDeclaration.ParameterList),
+                { } when LocalFunctionStatementSyntaxWrapper.IsInstance(node) => IdentifierNames(((LocalFunctionStatementSyntaxWrapper)node).ParameterList),
+                _ => Enumerable.Empty<string>()
+            };
+
+            static IEnumerable<string> IdentifierNames(BaseParameterListSyntax parameterList) =>
+                    parameterList.Parameters.Select(x => x.Identifier.ValueText);
+
+            static IEnumerable<string> AccessorIdentifierNames(AccessorDeclarationSyntax accessor)
             {
                 var arguments = new List<string>();
-                if (node.FirstAncestorOrSelf<IndexerDeclarationSyntax>() is { } indexer)
+                if (accessor.FirstAncestorOrSelf<IndexerDeclarationSyntax>() is { } indexer)
                 {
                     arguments.AddRange(IdentifierNames(indexer.ParameterList));
                 }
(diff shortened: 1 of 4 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CheckArgumentException.CSharp9.cs
```diff
@@ -71,3 +71,18 @@ public string Request
...
+// https://github.com/SonarSource/sonar-dotnet/issues/4423
+public class Repro_4423
+{
+    public void InsideLocalFunction_Static(string methodArg)
+    {
+        Something(null);
+
+        static void Something(string localArg)
+        {
+            throw new ArgumentNullException(nameof(localArg));   // Compliant
```
costin-zaharia-sonarsource, line 84: In C# 9 we can have also top level functions. I think it would be great to have a test case to cover that.

analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
```diff
@@ -40,37 +41,32 @@ public class CheckArgumentException : SonarDiagnosticAnalyzer
...
-                as IMethodSymbol;
-            if (methodSymbol?.ContainingType == null ||
-                !methodSymbol.ContainingType.IsAny(ArgumentExceptionTypesToCheck))
+            var methodSymbol = analysisContext.SemanticModel.GetSymbolInfo(objectCreationSyntax).Symbol as IMethodSymbol;
+            if (methodSymbol?.ContainingType == null || !methodSymbol.ContainingType.IsAny(ArgumentExceptionTypesToCheck))
             {
                 return;
             }
 
             if (objectCreationSyntax.ArgumentList.Arguments.Count == 0)
```
costin-zaharia-sonarsource, line 67: This check is much faster than retrieving the method symbol and verifying it. Could we move it on top?

costin-zaharia-sonarsource, line 67: Depending on the syntax the argument list can be missing. I think we need to check for null.

analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
```diff
@@ -100,61 +95,44 @@ private static void CheckForIssue(SyntaxNodeAnalysisContext analysisContext)
...
+                    : string.Format(InvalidParameterName, parameterNameValue.Value);
+                analysisContext.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, objectCreationSyntax.GetLocation(), message));
             }
-
-            analysisContext.ReportDiagnosticWhenActive(Diagnostic.Create(rule, objectCreationSyntax.GetLocation(),
-                    string.Format(InvalidParameterName, parameterNameValue.Value)));
         }
 
-        private static ISet<string> GetMethodArgumentNames(ObjectCreationExpressionSyntax creationSyntax)
+        private static IEnumerable<string> GetMethodArgumentNames(ObjectCreationExpressionSyntax creationSyntax)
```
costin-zaharia-sonarsource, line 108: That's one very long method. I think it would be better to split it.

costin-zaharia-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
pavel-mikula-sonarsource: No, we need to report on the expected exceptions. So the symbol loading is needed first. Otherwise this would report on any `new AnyTypeWithNoArgs()`

pavel-mikula-sonarsource: Good catch!

analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
pavel-mikula-sonarsource: I intentionally joined them, because there was hidden dependency between them. The first part lists some types and the second one processes that exact list of types.

costin-zaharia-sonarsource: And instead of having one very long method we could extract one method `GetMethodDeclaration` (responsibility will be clear, there will be one single reason for change in this case) and `GetIdentifierNames` responsible for providing the names. The second can call directly the first or receive the SyntaxNode as parameter. One other thing we can improve is the second part which seems to be replaceable with a switch expression or statement.

costin-zaharia-sonarsource (changes requested):
I think that splitting `GetMethodArgumentNames` will make the code more readable

analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
pavel-mikula-sonarsource: Two methods like this was the original state. While they were easy to read individually, the hidden dependency made them hard to work with.

The switch expression removed quite a lot of noise. Please take a look if if fixes the readability issue.

costin-zaharia-sonarsource: I think it's certainly better since it reduced a lot the size of the method 😄

costin-zaharia-sonarsource (approved):
LGTM!

PR #4504, opened 2021-06-03
Rule S4507: Add exception for StartupDevelopment

### LINKED ISSUE #4398: Fix S4507 FP: Do not raise inside StartupDevelopment class
### Description

If we have an override for `Startup.cs` with `StartupDevelopment.cs`, the overridden file is expected to run only on Development environment. (ASP.Net automatically does that based on ASPNETCORE_ENVIRONMENT, when it equals "Development"). In production, it won't be called.

image

This is a false-positive for FP S4507. 

### Repro steps

Create ASP.Net core app with `Startup.cs` and override it with `StartupDevelopment.cs`. Inside the overrode call `app.UseDeveloperExceptionPage();`

### Expected behavior

If `app.UseDeveloperExceptionPage();` is called from a `StartupDevelopment.cs` file its should not trigger FP S4507

### Actual behavior

FP S4507 is raised as security bug

costin-zaharia-sonarsource:
Fixes #4398

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DeliveringDebugFeaturesInProduction.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DeliveringDebugFeaturesInProduction.cs
@@ -38,16 +38,14 @@ namespace SonarAnalyzer.Rules.CSharp
 
         internal /*for testing*/ DeliveringDebugFeaturesInProduction(IAnalyzerConfiguration configuration) : base(configuration) { }
 
-        protected override TrackerBase<SyntaxKind, InvocationContext>.Condition IsInvokedConditionally() =>
-            context =>
-                context.Node.FirstAncestorOrSelf<StatementSyntax>() is { } invocationStatement
-                && invocationStatement.Ancestors().Any(node => IsDevelopmentCheck(node, context.SemanticModel));
+        protected override bool IsInvokedConditionally(SyntaxNode node, SemanticModel semanticModel) =>
+            node.FirstAncestorOrSelf<StatementSyntax>() is { } invocationStatement
+            && invocationStatement.Ancestors().Any(x => IsDevelopmentCheck(x, semanticModel));
 
-        protected override TrackerBase<SyntaxKind, InvocationContext>.Condition IsInDevelopmentContext() =>
-            context => context.Node
-                              .Ancestors()
-                              .OfType<ClassDeclarationSyntax>()
-                              .Any(declaration => declaration.Identifier.Text == StartupDevelopment);
+        protected override bool IsInDevelopmentContext(SyntaxNode node) =>
+            node.Ancestors()
+                .OfType<ClassDeclarationSyntax>()
+                .Any(x => x.Identifier.Text == StartupDevelopment);
 
         private bool IsDevelopmentCheck(SyntaxNode node, SemanticModel semanticModel) =>
             node is IfStatementSyntax ifStatement
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DeliveringDebugFeaturesInProduction.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DeliveringDebugFeaturesInProduction.cs
@@ -43,6 +43,12 @@ namespace SonarAnalyzer.Rules.CSharp
                 context.Node.FirstAncestorOrSelf<StatementSyntax>() is { } invocationStatement
                 && invocationStatement.Ancestors().Any(node => IsDevelopmentCheck(node, context.SemanticModel));
 
+        protected override TrackerBase<SyntaxKind, InvocationContext>.Condition IsInDevelopmentContext() =>
+            context => context.Node
+                              .Ancestors()
+                              .OfType<ClassDeclarationSyntax>()
+                              .Any(declaration => declaration.Identifier.Text == StartupDevelopment);
+
         private bool IsDevelopmentCheck(SyntaxNode node, SemanticModel semanticModel) =>
             node is IfStatementSyntax ifStatement
             && ifStatement.Condition.RemoveParentheses() is InvocationExpressionSyntax condition
--- a/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DeliveringDebugFeaturesInProductionBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DeliveringDebugFeaturesInProductionBase.cs
@@ -31,7 +31,8 @@ namespace SonarAnalyzer.Rules
         where TSyntaxKind : struct
     {
         protected const string DiagnosticId = "S4507";
-        protected const string MessageFormat = "Make sure this debug feature is deactivated before delivering the code in production.";
+        protected const string StartupDevelopment = "StartupDevelopment";
+        private const string MessageFormat = "Make sure this debug feature is deactivated before delivering the code in production.";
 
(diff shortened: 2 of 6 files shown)


analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DeliveringDebugFeaturesInProductionBase.cs
```diff
@@ -40,17 +41,19 @@ public abstract class DeliveringDebugFeaturesInProductionBase<TSyntaxKind> : Tra
...
         {
             var t = Language.Tracker.Invocation;
             t.Track(input,
-                t.MatchMethod(
-                    new MemberDescriptor(KnownType.Microsoft_AspNetCore_Builder_DeveloperExceptionPageExtensions, "UseDeveloperExceptionPage"),
-                    new MemberDescriptor(KnownType.Microsoft_AspNetCore_Builder_DatabaseErrorPageExtensions, "UseDatabaseErrorPage")),
-                t.ExceptWhen(IsInvokedConditionally()));
+                    t.MatchMethod(new MemberDescriptor(KnownType.Microsoft_AspNetCore_Builder_DeveloperExceptionPageExtensions, "UseDeveloperExceptionPage"),
+                                  new MemberDescriptor(KnownType.Microsoft_AspNetCore_Builder_DatabaseErrorPageExtensions, "UseDatabaseErrorPage")),
+                    t.And(t.ExceptWhen(IsInvokedConditionally()),
```
pavel-mikula-sonarsource, line 55: You can remove the `t.And` to achieve the same. The `Track` method has `params` conditions that behaves as `and` mode.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DeliveringDebugFeaturesInProductionBase.cs
```diff
@@ -31,7 +31,8 @@ public abstract class DeliveringDebugFeaturesInProductionBase<TSyntaxKind> : Tra
         where TSyntaxKind : struct
     {
         protected const string DiagnosticId = "S4507";
-        protected const string MessageFormat = "Make sure this debug feature is deactivated before delivering the code in production.";
+        protected const string StartupDevelopment = "StartupDevelopment";
+        private const string MessageFormat = "Make sure this debug feature is deactivated before delivering the code in production.";
```
pavel-mikula-sonarsource, line 35: There's stylecope formatting issue on line 40 with `);` that can be easily fixed (clean as you code)

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DeliveringDebugFeaturesInProductionBase.cs
```diff
@@ -40,17 +41,19 @@ public abstract class DeliveringDebugFeaturesInProductionBase<TSyntaxKind> : Tra
 
         protected abstract TrackerBase<TSyntaxKind, InvocationContext>.Condition IsInvokedConditionally();
 
+        protected abstract TrackerBase<TSyntaxKind, InvocationContext>.Condition IsInDevelopmentContext();
```
pavel-mikula-sonarsource, line 44: These two methods have very complicated signature that's propagated to inheriting classes. It's always puzzling to understand the lambda indirection here. I suggest update both of them like this:
```
protected abstract bool IsInDevelopmentContext(SyntaxNode node);
```
and use them like this:
```
t.ExceptWhen(c => IsInvokedConditionally(c.Node))
```

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/DeliveringDebugFeaturesInProduction.cs
```diff
@@ -43,6 +43,12 @@ public sealed class DeliveringDebugFeaturesInProduction : DeliveringDebugFeature
                 context.Node.FirstAncestorOrSelf<StatementSyntax>() is { } invocationStatement
                 && invocationStatement.Ancestors().Any(node => IsDevelopmentCheck(node, context.SemanticModel));
 
+        protected override TrackerBase<SyntaxKind, InvocationContext>.Condition IsInDevelopmentContext() =>
+            context => context.Node
+                              .Ancestors()
+                              .OfType<ClassBlockSyntax>()
+                              .Any(classBlockSyntax => classBlockSyntax.ClassStatement.Identifier.Text == StartupDevelopment);
```
pavel-mikula-sonarsource, line 50: This is the case where `x` will make it shorter and more readable. It's clear that it's `ClassBlockSyntax` from the previous line.
```suggestion
                              .Any(x => x.ClassStatement.Identifier.Text == StartupDevelopment);
```

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/DeliveringDebugFeaturesInProduction.cs
```diff
@@ -43,6 +43,12 @@ public sealed class DeliveringDebugFeaturesInProduction : DeliveringDebugFeature
                 context.Node.FirstAncestorOrSelf<StatementSyntax>() is { } invocationStatement
                 && invocationStatement.Ancestors().Any(node => IsDevelopmentCheck(node, context.SemanticModel));
 
+        protected override TrackerBase<SyntaxKind, InvocationContext>.Condition IsInDevelopmentContext() =>
+            context => context.Node
+                              .Ancestors()
+                              .OfType<ClassDeclarationSyntax>()
+                              .Any(declaration => declaration.Identifier.Text == StartupDevelopment);
```
pavel-mikula-sonarsource, line 50: Same as VB - could be `x`, as the `OfType` above is quite clear about the type
```suggestion
                              .Any(x => x.Identifier.Text == StartupDevelopment);
```

pavel-mikula-sonarsource (changes requested):
Minor refactoring improvements

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/DeliveringDebugFeaturesInProductionBase.cs
pavel-mikula-sonarsource: Soo much easier to read :)

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/DeliveringDebugFeaturesInProduction.cs
```diff
@@ -51,7 +55,7 @@ public sealed class DeliveringDebugFeaturesInProduction : DeliveringDebugFeature
             node switch
             {
                 MultiLineIfBlockSyntax multiline => multiline.IfStatement.Condition,
-                SingleLineIfStatementSyntax singleline => singleline.Condition,
+                SingleLineIfStatementSyntax singleLine => singleLine.Condition,
```
pavel-mikula-sonarsource, line 58: _Nitpick: _ This one was renamed to `singleLine` but the one above is `multiLine` - looks inconsistent now

pavel-mikula-sonarsource (approved):
LGTM!

analyzers/src/SonarAnalyzer.VisualBasic/Rules/Hotspots/DeliveringDebugFeaturesInProduction.cs
costin-zaharia-sonarsource: I will fix it in a following PR. I would not restart the pipeline just for that

PR #4505, opened 2021-06-03
Change S1451 default value to generic template

### LINKED ISSUE #3913: Change S1451 default value to generic template
S1451 CheckFileLicense contains SonarSource 2013 license header as the default value for all our users. 

We should change it to some short and simple default license template with placeholders like `<YOUR COMPANY>` and `<YEAR FROM>-<YEAR TO>` to help users understand the rule and how it works.

It also makes `Happy New Year` maintenance commits harder to validate, as there should be 2013 licenses persisted in our repo / .NET ITs.

pavel-mikula-sonarsource:
Fixes #3913

It would be funny if this change actually raises some noise on user projects 😄

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Rules/CheckFileLicenseBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/CheckFileLicenseBase.cs
@@ -32,12 +32,12 @@ namespace SonarAnalyzer.Rules
     public abstract class CheckFileLicenseBase : ParameterLoadingDiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S1451";
-        internal const string HeaderFormatRuleParameterKey = "headerFormat";
         internal const string HeaderFormatPropertyKey = nameof(HeaderFormat);
-        internal const string IsRegularExpressionRuleParameterKey = "isRegularExpression";
         internal const string IsRegularExpressionPropertyKey = nameof(IsRegularExpression);
-        internal const string IsRegularExpressionDefaultValue = "false";
-        protected const string MessageFormat = "Add or update the header of this file.";
+        protected const string HeaderFormatRuleParameterKey = "headerFormat";
+        private const string IsRegularExpressionRuleParameterKey = "isRegularExpression";
+        private const string IsRegularExpressionDefaultValue = "false";
+        private const string MessageFormat = "Add or update the header of this file.";
 
         private readonly DiagnosticDescriptor rule;
 
@@ -73,7 +73,7 @@ namespace SonarAnalyzer.Rules
                     }
                 });
 
-        protected static bool IsRegexPatternValid(string pattern)
+        private static bool IsRegexPatternValid(string pattern)
         {
             try
             {
@@ -86,7 +86,7 @@ namespace SonarAnalyzer.Rules
             }
         }
 
-        protected bool HasValidLicenseHeader(SyntaxNode node)
+        private bool HasValidLicenseHeader(SyntaxNode node)
         {
             if (node == null || !node.HasLeadingTrivia)
             {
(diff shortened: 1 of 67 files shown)


analyzers/src/SonarAnalyzer.Common/Rules/CheckFileLicenseBase.cs
```diff
@@ -18,31 +18,61 @@
...
     public abstract class CheckFileLicenseBase : ParameterLoadingDiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S1451";
-        protected const string MessageFormat = "Add or update the header of this file.";
-
         internal const string HeaderFormatRuleParameterKey = "headerFormat";
         internal const string HeaderFormatPropertyKey = nameof(HeaderFormat);
-        public abstract string HeaderFormat { get; set; }
-
         internal const string IsRegularExpressionRuleParameterKey = "isRegularExpression";
```
costin-zaharia-sonarsource, line 37: clean as you code: can be private

analyzers/src/SonarAnalyzer.Common/Rules/CheckFileLicenseBase.cs
```diff
@@ -18,31 +18,61 @@
...
         internal const string IsRegularExpressionPropertyKey = nameof(IsRegularExpression);
         internal const string IsRegularExpressionDefaultValue = "false";
```
costin-zaharia-sonarsource, line 39: clean as you code: can be private

analyzers/src/SonarAnalyzer.Common/Rules/CheckFileLicenseBase.cs
```diff
@@ -18,31 +18,61 @@
...
-        [RuleParameter(IsRegularExpressionRuleParameterKey, PropertyType.Boolean,
-            "Whether the headerFormat is a regular expression.", IsRegularExpressionDefaultValue)]
+        protected const string MessageFormat = "Add or update the header of this file.";
```
costin-zaharia-sonarsource, line 40: can be private

analyzers/src/SonarAnalyzer.Common/Rules/CheckFileLicenseBase.cs
```diff
@@ -18,31 +18,61 @@
...
+            context.RegisterSyntaxTreeActionInNonGenerated(Language.GeneratedCodeRecognizer, c =>
+                {
+                    if (HeaderFormat == null)
+                    {
+                        return;
+                    }
+
+                    if (IsRegularExpression && !IsRegexPatternValid(HeaderFormat))
+                    {
+                        throw new InvalidOperationException($"Invalid regular expression: {HeaderFormat}");
```
costin-zaharia-sonarsource, line 65: Do we really want to throw an exception here? If I understand the behavior right this will raise AD001 for each project file if the regex pattern is not valid.

analyzers/src/SonarAnalyzer.Common/Rules/CheckFileLicenseBase.cs
```diff
@@ -18,31 +18,61 @@
...
+
+                    var firstNode = c.Tree.GetRoot().ChildTokens().FirstOrDefault().Parent;
+                    if (!HasValidLicenseHeader(firstNode))
+                    {
+                        var properties = CreateDiagnosticProperties();
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, Location.Create(c.Tree, TextSpan.FromBounds(0, 0)), properties));
+                    }
+                });
+
         protected static bool IsRegexPatternValid(string pattern)
```
costin-zaharia-sonarsource, line 76: can be private

costin-zaharia-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Common/Rules/CheckFileLicenseBase.cs
pavel-mikula-sonarsource: I don't remember any community problem with this so far. 
It's not SonarWay.
There's no validation on SQ/SC side for the regex.
If you want this blocker rule, you should have it configured it correctly instead of silent FNs everywhere that are hard to diagnose. 

So I'd say it's noisy, but reasonable in this case.

analyzers/src/SonarAnalyzer.Common/Rules/CheckFileLicenseBase.cs
pavel-mikula-sonarsource: Nice, almost everything is private now :)

costin-zaharia-sonarsource (approved):
LGTM!

PR #4506, opened 2021-06-03
Rule S2223: Fix FP when using ThreadStatic

### LINKED ISSUE #4407: Rule S2223 FP when using ThreadStatic
### Description

Rule [S2223](https://rules.sonarsource.com/csharp/RSPEC-2223) is raising a false positive when using ThreadStatic attributes on fields.

### Repro steps

```
using System;

class Program
{
    [ThreadStatic]
    public static int value;

    static void Main(){}
}
```

### Expected behavior

S2223 should not be raised in this case since the field is decorated with ThreadStatic attribute and it's value is not shared between threads.

### Actual behavior

S2223 should is raised.

See: https://discuss.sonarsource.com/t/s2223-and-threadstatic-attribute/7517/3

costin-zaharia-sonarsource:
Fixes #4407

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/StaticFieldVisible.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/StaticFieldVisible.cs
@@ -55,13 +55,12 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static IEnumerable<Diagnostic> GetDiagnostics(VariableDeclarationSyntax declaration, SemanticModel semanticModel) =>
             declaration.Variables
-                       .Select(declarator => new {Syntax = declarator, Symbol = semanticModel.GetDeclaredSymbol(declarator) as IFieldSymbol})
-                       .Where(f => FieldIsRelevant(f.Symbol))
-                       .Select(pair => Diagnostic.Create(Rule, pair.Syntax.Identifier.GetLocation(), pair.Syntax.Identifier.ValueText));
+                       .Where(x => FieldIsRelevant(semanticModel.GetDeclaredSymbol(x) as IFieldSymbol))
+                       .Select(x => Diagnostic.Create(Rule, x.Identifier.GetLocation(), x.Identifier.ValueText));
 
         private static bool FieldIsRelevant(IFieldSymbol fieldSymbol) =>
             fieldSymbol is {IsStatic: true, IsConst: false, IsReadOnly: false}
             && fieldSymbol.DeclaredAccessibility != Accessibility.Private
-            && !fieldSymbol.GetAttributes().Any(attribute => attribute.AttributeClass.Is(KnownType.System_ThreadStaticAttribute));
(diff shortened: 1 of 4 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/StaticFieldVisible.cs
```diff
@@ -46,21 +46,22 @@ public sealed class StaticFieldVisible : SonarDiagnosticAnalyzer
...
+                        c.ReportDiagnosticWhenActive(diagnostic);
                     }
                 },
                 SyntaxKind.FieldDeclaration);
 
+        private static IEnumerable<Diagnostic> GetDiagnostics(VariableDeclarationSyntax declaration, SemanticModel semanticModel) =>
+            declaration.Variables
+                       .Select(declarator => new {Syntax = declarator, Symbol = semanticModel.GetDeclaredSymbol(declarator) as IFieldSymbol})
+                       .Where(f => FieldIsRelevant(f.Symbol))
+                       .Select(pair => Diagnostic.Create(Rule, pair.Syntax.Identifier.GetLocation(), pair.Syntax.Identifier.ValueText));
```
pavel-mikula-sonarsource, line 60: It's weird to name the same thing `f` on one line and `pair` on another. My personal favorite is `x`, as nothing special is needed to distinguish them

analyzers/src/SonarAnalyzer.CSharp/Rules/StaticFieldVisible.cs
```diff
@@ -46,21 +46,22 @@ public sealed class StaticFieldVisible : SonarDiagnosticAnalyzer
...
-                        c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, field.Syntax.Identifier.GetLocation(), field.Syntax.Identifier.ValueText));
+                        c.ReportDiagnosticWhenActive(diagnostic);
                     }
                 },
                 SyntaxKind.FieldDeclaration);
 
+        private static IEnumerable<Diagnostic> GetDiagnostics(VariableDeclarationSyntax declaration, SemanticModel semanticModel) =>
+            declaration.Variables
+                       .Select(declarator => new {Syntax = declarator, Symbol = semanticModel.GetDeclaredSymbol(declarator) as IFieldSymbol})
+                       .Where(f => FieldIsRelevant(f.Symbol))
```
pavel-mikula-sonarsource, line 59: The `Symbol` part is used only in the `Where`, so it's causing more syntax and readability trouble than it solves.
Let's move the `GetDeclaredSymbol` part to the `Where` section to simplify the overall processing.
```suggestion
                       .Where(f => FieldIsRelevant(semanticModel.GetDeclaredSymbol(declarator) as IFieldSymbol))
```

pavel-mikula-sonarsource (changes requested):
Minor syntax improvements

pavel-mikula-sonarsource (approved):
LGTM

PR #4508, opened 2021-06-04
Rule S3330, S2092, S2077, S5122, S2755: Add support for target typed new expressions

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Trackers/CSharpObjectCreationTracker.cs
```diff
@@ -21,22 +21,24 @@
...
+using SonarAnalyzer.Wrappers;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Helpers.Trackers
 {
     public class CSharpObjectCreationTracker : ObjectCreationTracker<SyntaxKind>
     {
         protected override ILanguageFacade<SyntaxKind> Language => CSharpFacade.Instance;
-        protected override SyntaxKind[] TrackedSyntaxKinds { get; } = { SyntaxKind.ObjectCreationExpression };
+        protected override SyntaxKind[] TrackedSyntaxKinds { get; } = { SyntaxKind.ObjectCreationExpression, SyntaxKindEx.ImplicitObjectCreationExpression };
```
pavel-mikula-sonarsource, line 32: We can move this to the base class now.
And I'd rename `ISyntaxKindFacade.ObjectCreationExpression` to `ObjectCreationExpressions`

analyzers/src/SonarAnalyzer.CSharp/Rules/XXE/XmlReaderSettingsValidator.cs
```diff
@@ -111,16 +114,17 @@ public IList<Location> GetUnsafeAssignmentLocations(InvocationExpressionSyntax i
...
             invocation.DescendantNodes()
                       .Union(symbol.GetLocationNodes(invocation))
-                      .OfType<ObjectCreationExpressionSyntax>()
+                      .Where(x => x.IsAnyKind(SyntaxKind.ObjectCreationExpression, SyntaxKindEx.ImplicitObjectCreationExpression))
+                      .Select(ObjectCreationFactory.Create)
                       .FirstOrDefault(objectCreation => IsXmlReaderSettingsCreationWithInitializer(objectCreation, semanticModel));
 
-        private static bool IsXmlReaderSettingsCreationWithInitializer(ObjectCreationExpressionSyntax objectCreation, SemanticModel semanticModel) =>
-            objectCreation.Initializer != null && IsXmlReaderSettings(objectCreation, semanticModel);
+        private static bool IsXmlReaderSettingsCreationWithInitializer(IObjectCreation objectCreation, SemanticModel semanticModel) =>
```
pavel-mikula-sonarsource, line 126: Method `IsXmlReaderSettings` on line 114 is now used only once and can be embedded into it's call site instead.

pavel-mikula-sonarsource: _Just to be clear:_ I want to preserve `IsXmlReaderSettingsCreationWithInitializer` as is, because it's used on a long line in nested lambda and the name brings some readability as well.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/ExecutingSqlQueries.CSharp9.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.Data.Sqlite;
+
+string ConstQuery = "";
+
+void Foo(DbContext context, SqliteConnection connection, string query, params object[] parameters)
+{
+    context.Database.ExecuteSqlCommand(ConstQuery); // Compliant, constants are safe
+    context.Database.ExecuteSqlCommand(query); // Compliant, not concat or format
+
+    context.Database.ExecuteSqlCommand($"SELECT * FROM mytable WHERE mycol={query} AND mycol2={0}", parameters[0]); // Noncompliant, string interpolation  it is RawSqlString
```
pavel-mikula-sonarsource, line 13: The comment has weird wording

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/CookieShouldBeSecure_Net.cs
```diff
@@ -1,26 +0,0 @@
-﻿using System;
```
pavel-mikula-sonarsource, line 1: History for this file will be probably lost (I'm not sure if squash merge will recognize the rename). I'm not sure if we can do something about it, but as general, it's good to rename files in a single commit to avoid git confusion when file is renamed and modified

pavel-mikula-sonarsource (changes requested):
Few minor improvements

analyzers/src/SonarAnalyzer.CSharp/Trackers/CSharpObjectCreationTracker.cs
```diff
@@ -21,22 +21,24 @@
...
 
         internal override Condition ArgumentAtIndexIsConst(int index) =>
-            context => ((ObjectCreationExpressionSyntax)context.Node).ArgumentList is { } argumentList
+            context => ObjectCreationFactory.Create(context.Node).ArgumentList is { } argumentList
```
pavel-mikula-sonarsource, line 35: Let's try this syntax to cover the case with `ArgumentList` being `null`
```
var x = new Something {Property=42};
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/CookieShouldBeSecure_Net.cs
dmitry-golovinov-sonarsource: Ok, will put more attention into it.

analyzers/src/SonarAnalyzer.CSharp/Trackers/CSharpObjectCreationTracker.cs
dmitry-golovinov-sonarsource: Had to craft a custom test for this, since this is not invocable when constructor has no parameters in the existing rules.

pavel-mikula-sonarsource (commented):
LGTM

pavel-mikula-sonarsource (approved):
LGTM v2

PR #4511, opened 2021-06-04
Fix FP when Localizable attribute has false argument in constructor

### LINKED ISSUE #4502: S4055 FP: Localizable attribute with false argument in constructor
### Description

Rule S4055 should not raise an issue if parameters or properties have [Localizable(false)] attribute.

### Repro steps

```
public void DoWork()
{
    LogMessage(
        "Debug", // This should *not* raise S4055
        "My debug message" /* This is fine if it warns */);
}

public void LogMessage(
    [Localizable(false)]string messageType, 
    string actualMessage)
{
    switch (messageType)
    {
        case "Debug":
            // handle debug
            break;
        case "Trace":
            // etc.
            break;
    }
}
```

### Expected behavior

No issues are raised if parameters or properties have [Localizable(false)] attribute.

### Actual behavior

Issues are raised if parameters or properties have [Localizable(false)] attribute.

### Known workarounds

-

### Related information

* C#/VB.NET Plugins version : 8.23

csaba-sagi-sonarsource:
Fixes #4502

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/LiteralsShouldNotBePassedAsLocalizedParameters.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/LiteralsShouldNotBePassedAsLocalizedParameters.cs
@@ -111,20 +111,20 @@ namespace SonarAnalyzer.Rules.CSharp
         private static bool IsLocalizable(ISymbol symbol) =>
             symbol?.Name != null
             && symbol.GetAttributes(KnownType.System_ComponentModel_LocalizableAttribute) is var localizableAttributes
-            && IsLocalizable(symbol.Name, new List<AttributeData>(localizableAttributes).AsReadOnly());
+            && IsLocalizable(symbol.Name, new List<AttributeData>(localizableAttributes));
 
         private static bool IsLocalizable(string symbolName, IReadOnlyCollection<AttributeData> localizableAttributes) =>
-            localizableAttributes.Any(x => HasConstructorWithBoolValue(x, true))
+            localizableAttributes.Any(x => HasConstructorWitValue(x, true))
             || (symbolName.SplitCamelCaseToWords().Any(LocalizableSymbolNames.Contains)
-               && (!localizableAttributes.Any(x => HasConstructorWithBoolValue(x, false))));
+               && (!localizableAttributes.Any(x => HasConstructorWitValue(x, false))));
 
-        private static bool IsLocalizableStringLiteral(IParameterSymbol parameter, ArgumentSyntax argumentSyntax, SemanticModel semanticModel) =>
-            parameter != null
+        private static bool IsLocalizableStringLiteral(ISymbol symbol, ArgumentSyntax argumentSyntax, SemanticModel semanticModel) =>
+            symbol != null
             && argumentSyntax != null
-            && IsLocalizable(parameter)
+            && IsLocalizable(symbol)
             && IsStringLiteral(argumentSyntax.Expression, semanticModel);
 
-        private static bool HasConstructorWithBoolValue(AttributeData attribute, bool expectedValue) =>
+        private static bool HasConstructorWitValue(AttributeData attribute, bool expectedValue) =>
             attribute.ConstructorArguments.Any(c => c.Value is bool boolValue && boolValue == expectedValue);
     }
 }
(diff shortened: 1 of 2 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/LiteralsShouldNotBePassedAsLocalizedParameters.cs
```diff
@@ -52,84 +51,74 @@ public sealed class LiteralsShouldNotBePassedAsLocalizedParameters : SonarDiagno
...
+            symbol?.Name != null
+            && symbol.GetAttributes(KnownType.System_ComponentModel_LocalizableAttribute) is var localizableAttribute
+            && IsLocalizable(symbol.Name, localizableAttribute);
+
+        private static bool IsLocalizable(string symbolName, IEnumerable<AttributeData> localizableAttribute) =>
+            localizableAttribute.Any(x => AttributeHasConstructorArgument(x, true))
+            || (symbolName.SplitCamelCaseToWords().Any(LocalizableSymbolNames.Contains)
+               && (!localizableAttribute.Any(x => AttributeHasConstructorArgument(x, false))));
+
+        private static bool AttributeHasConstructorArgument(AttributeData attribute, bool expectedValue) =>
```
costin-zaharia-sonarsource, line 121: `AttributeHasConstructorArgument`: I think it's clear that it does a check on the attribute due to the parameter type and name. We can rename it to `HasCtorWith`.

We should make it an extension method on `AttibuteData` to enable reuse.

analyzers/src/SonarAnalyzer.CSharp/Rules/LiteralsShouldNotBePassedAsLocalizedParameters.cs
```diff
@@ -52,84 +51,74 @@ public sealed class LiteralsShouldNotBePassedAsLocalizedParameters : SonarDiagno
...
-            return symbol.Name.SplitCamelCaseToWords().Any(localizableSymbolNames.Contains) ||
-                symbol.GetAttributes(KnownType.System_ComponentModel_LocalizableAttribute)
-                    .Any(a => a.ConstructorArguments.Any(c => (c.Value as bool?) ?? false));
-        }
+        private static bool IsLocalizable(ISymbol symbol) =>
+            symbol?.Name != null
+            && symbol.GetAttributes(KnownType.System_ComponentModel_LocalizableAttribute) is var localizableAttribute
+            && IsLocalizable(symbol.Name, localizableAttribute);
+
+        private static bool IsLocalizable(string symbolName, IEnumerable<AttributeData> localizableAttribute) =>
```
costin-zaharia-sonarsource, line 116: the name should be at the plural form since there can be multiple attributes

costin-zaharia-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/LiteralsShouldNotBePassedAsLocalizedParameters.cs
```diff
@@ -103,22 +104,27 @@ private static void AnalyzeAssignments(SyntaxNodeAnalysisContext context)
...
 
-        private static bool IsLocalizable(string symbolName, IEnumerable<AttributeData> localizableAttribute) =>
-            localizableAttribute.Any(x => AttributeHasConstructorArgument(x, true))
+        private static bool IsLocalizable(string symbolName, IReadOnlyCollection<AttributeData> localizableAttributes) =>
+            localizableAttributes.Any(x => HasConstructorWithBoolValue(x, true))
             || (symbolName.SplitCamelCaseToWords().Any(LocalizableSymbolNames.Contains)
-               && (!localizableAttribute.Any(x => AttributeHasConstructorArgument(x, false))));
+               && (!localizableAttributes.Any(x => HasConstructorWithBoolValue(x, false))));
+
+        private static bool IsLocalizableStringLiteral(IParameterSymbol parameter, ArgumentSyntax argumentSyntax, SemanticModel semanticModel) =>
```
costin-zaharia-sonarsource, line 121: `parameter` can be `ISymbol`. We can also rename it to `symbol`

analyzers/src/SonarAnalyzer.CSharp/Rules/LiteralsShouldNotBePassedAsLocalizedParameters.cs
```diff
@@ -103,22 +104,27 @@ private static void AnalyzeAssignments(SyntaxNodeAnalysisContext context)
...
+            parameter != null
+            && argumentSyntax != null
+            && IsLocalizable(parameter)
+            && IsStringLiteral(argumentSyntax.Expression, semanticModel);
 
-        private static bool AttributeHasConstructorArgument(AttributeData attribute, bool expectedValue) =>
+        private static bool HasConstructorWithBoolValue(AttributeData attribute, bool expectedValue) =>
```
costin-zaharia-sonarsource, line 127: We can rename it to `HasConstructorWitValue`. `expectedValue` is already `bool` so we don't need to mention it in the method name. What do you think?

costin-zaharia-sonarsource (approved):

PR #4516, opened 2021-06-04
GitHub workflow: Improve Milestone

pavel-mikula-sonarsource:
Do not create card when there already exist one to prevent Action error and email with an error log.

Do not add card for closed issue, it's already done.

### DIFF
--- a/.github/workflows/MilestoneIssue.yml
+++ b/.github/workflows/MilestoneIssue.yml
@@ -8,6 +8,8 @@ jobs:
   create_card:
     name: Create card
     runs-on: ubuntu-latest
+    # Single quotes must be used here https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#literals
+    if: github.event.issue.state != 'closed'
 
     steps:
       # https://github.com/actions/github-script
--- a/.github/workflows/MilestoneIssue.yml
+++ b/.github/workflows/MilestoneIssue.yml
@@ -16,7 +16,30 @@ jobs:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
             const TODO_COLUMN = 4971951;
-            console.log("Creating Issue card");
-            // Action will fail if a card already exists for this issue
-            github.projects.createCard({ column_id: TODO_COLUMN, content_id: context.payload.issue.id, content_type: "Issue" });
+            const IN_PROGRESS_COLUMN = 4971952;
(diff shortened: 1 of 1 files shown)


.github/workflows/MilestoneIssue.yml
```diff
@@ -16,7 +18,30 @@ jobs:
...
             const TODO_COLUMN = 4971951;
-            console.log("Creating Issue card");
-            // Action will fail if a card already exists for this issue
-            github.projects.createCard({ column_id: TODO_COLUMN, content_id: context.payload.issue.id, content_type: "Issue" });
+            const IN_PROGRESS_COLUMN = 4971952;
+            const REVIEW_IN_PROGRESS_COLUMN = 4971953;
+            const REVIEW_APPROVED_COLUMN = 4971954;
+            const DONE_COLUMN = 4971955;
+            //
+            async function findCard(content_url) {
```
pavel-mikula-sonarsource, line 26: This method is copy/pasted from here, except different column ordering:
https://github.com/SonarSource/sonar-dotnet/blob/9bf155c48565506572195db880001524699662f4/.github/workflows/RequestReview.yml#L39-L53

pavel-mikula-sonarsource:
I'm going to ignore the failed CFG repox step

.github/workflows/MilestoneIssue.yml
```diff
@@ -16,7 +16,30 @@ jobs:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
             const TODO_COLUMN = 4971951;
-            console.log("Creating Issue card");
-            // Action will fail if a card already exists for this issue
-            github.projects.createCard({ column_id: TODO_COLUMN, content_id: context.payload.issue.id, content_type: "Issue" });
+            const IN_PROGRESS_COLUMN = 4971952;
+            const REVIEW_IN_PROGRESS_COLUMN = 4971953;
+            const REVIEW_APPROVED_COLUMN = 4971954;
```
andrei-epure-sonarsource, line 23: also declared `const VALIDATE_PEACH_COLUMN = 7838551;`?

you have it in `findCard()` but not here

andrei-epure-sonarsource (changes requested):

.github/workflows/MilestoneIssue.yml
pavel-mikula-sonarsource: Good catch 🤦‍♂️

andrei-epure-sonarsource (approved):
lgtm

PR #4517, opened 2021-06-04
GitHub workflow: await removeAssignees

pavel-mikula-sonarsource:
Old state: When card was already assigned to the desired user, `removeAssignees` and `addAssignee` happend asynchronously. So user was often added and them all users were removed.

New state: Wait for removal before adding the desired user.


andrei-epure-sonarsource (approved):
LGTM

PR #4518, opened 2021-06-04
Improve S3984 performance

andrei-epure-sonarsource:
In our [April perf evaluation](https://docs.google.com/spreadsheets/d/1uo7kSDPkbRFyoXtzBp7Piftcpda29i9mUoj-ZTluCtQ/edit#gid=1598900315) , on `nodatime`, S3984 behaved pretty badly:

<google-sheets-html-origin>
ExceptionsShouldBeUsed |  6324 ms (4%)
-- | -- |

</google-sheets-html-origin>

Also, see [internal report.](https://discuss.sonarsource.com/t/performance-problem-reported-for-csharp-post-8-9-upgrade-rules-impacted-s1128-s3984-and-s1450/7867/3). 

It's pretty obvious it will behave bad because it's querying the symbol as the first operation in the rule.

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionsShouldBeUsed.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionsShouldBeUsed.cs
@@ -32,7 +32,7 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class ExceptionsShouldBeUsed : SonarDiagnosticAnalyzer
     {
-        internal const string DiagnosticId = "S3984";
+        private const string DiagnosticId = "S3984";
         private const string MessageFormat = "Throw this exception or remove this useless statement.";
 
         private static readonly DiagnosticDescriptor Rule =
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionsShouldBeUsed.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionsShouldBeUsed.cs
@@ -45,13 +45,11 @@ namespace SonarAnalyzer.Rules.CSharp
             {
                 var objectCreation = (ObjectCreationExpressionSyntax)c.Node;
                 var parent = objectCreation.GetFirstNonParenthesizedParent();
-                if (parent.IsKind(SyntaxKind.ExpressionStatement))
+                if (parent.IsKind(SyntaxKind.ExpressionStatement)
+                    && c.SemanticModel.GetSymbolInfo(objectCreation.Type).Symbol is INamedTypeSymbol createdObjectType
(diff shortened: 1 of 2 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionsShouldBeUsed.cs
```diff
@@ -35,30 +35,25 @@ public sealed class ExceptionsShouldBeUsed : SonarDiagnosticAnalyzer
...
-                if (!createdObjectType.DerivesFrom(KnownType.System_Exception))
-                {
-                    return;
-                }
-
                 var parent = objectCreation.GetFirstNonParenthesizedParent();
                 if (parent.IsKind(SyntaxKind.ExpressionStatement))
                 {
-                    c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, objectCreation.GetLocation()));
+                    var createdObjectType = c.SemanticModel.GetSymbolInfo(objectCreation.Type).Symbol as INamedTypeSymbol;
```
Corniel, line 50: `&& c.SemanticModel.GetSymbolInfo(objectCreation.Type).Symbol is INamedTypeSymbol createdObjectType`

To have less nesting?

analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionsShouldBeUsed.cs
```diff
@@ -35,30 +35,23 @@ public sealed class ExceptionsShouldBeUsed : SonarDiagnosticAnalyzer
         internal const string DiagnosticId = "S3984";
```
costin-zaharia-sonarsource, line 35: pedantic: can be private

costin-zaharia-sonarsource (approved):
LGTM!

Corniel:
What about:

``` c#
public void NotCompiant()
{
    SomeException.FactoryMethod(); // returns an exception
}
```

(Agreed, not part of the improvement)

andrei-epure-sonarsource:
Hi @Corniel - if you think there's a false negative, you can open a new ticket. But generally I think exceptions are created simply by invoking the constructor.

Validation: 
- on nodatime, the time went down from 6324 ms to 58 ms
- on efcore, went down from 2755 ms to 796 ms
- on n2cms, went down from 471 ms to 33 ms

Corniel:
> 
> Hi @Corniel - if you think there's a false negative, you can open a new ticket. But generally I think exceptions are created simply by invoking the constructor.

I know for a fact that it is not always the case, but I agree that you should not add the issue here.

>     * on nodatime, the time went down from 6324 ms to 58 ms
>     * on efcore, went down from 2755 ms to 796 ms
>     * on n2cms, went down from 471 ms to 33 ms
Nice improvements!

PR #4524, opened 2021-06-08
Fix NullReferenceException raised by S4462

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CallToAsyncMethodShouldNotBeBlocking.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CallToAsyncMethodShouldNotBeBlocking.cs
@@ -314,7 +314,7 @@ namespace Tests.Diagnostics
             Task<long>[] arr;
             Task.WaitAll(arr[0]);   // Error [CS0165]
                                     // Noncompliant@-1 FP
-            var x = arr[0].Result;
+            var x = arr[0].Result;  // Noncompliant
         }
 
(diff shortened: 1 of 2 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -309,6 +309,13 @@ public static Task<int> GetValueTask(int num)
             return Task.Run(() => num);
         }
 
+        public void TaskIsNull()
+        {
+            Task<long>[] arr;
+            var x = arr[0].Result;  // Error [CS0165]
```
pavel-mikula-sonarsource, line 315: This doesn't reproduce the NRE on my machine with the original implementation 😢

pavel-mikula-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CallToAsyncMethodShouldNotBeBlocking.cs
dmitry-golovinov-sonarsource: My bad. Fixed and verified that it fails. I was verifying for null previously :)

pavel-mikula-sonarsource (changes requested):

pavel-mikula-sonarsource (approved):
LGTM

PR #4531, opened 2021-06-09
Update Automapper project to the latest version

dmitry-golovinov-sonarsource:


analyzers/its/sources/AutoMapper/src/IntegrationTests/AutoMapper.IntegrationTests.csproj
```diff
@@ -1,49 +1,18 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <!-- We need to manually mark this project as TEST. See analyzers\its\SonarAnalyzer.Testing.ImportBefore.targets for details.-->
-    <SonarQubeTestProject>true</SonarQubeTestProject>
```
andrei-epure-sonarsource, line 5: please add this back

analyzers/its/sources/AutoMapper/src/UnitTests/AutoMapper.UnitTests.csproj
```diff
@@ -1,52 +1,15 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <!-- We need to manually mark this project as TEST. See analyzers\its\SonarAnalyzer.Testing.ImportBefore.targets for details.-->
-    <SonarQubeTestProject>true</SonarQubeTestProject>
```
andrei-epure-sonarsource, line 5: please add this property back

andrei-epure-sonarsource (changes requested):
For future readers of this PR: in `AutoMapper.csproj` the  `<ProjectGuid>025B9F93-7953-454B-AF29-E6A55B2AA58F</ProjectGuid>` has been removed, thus a lot of files seem to have been deleted and a lot added, but in fact they have only been renamed. It's the same phenomenon for the UnitTests project (`<ProjectGuid>F0CC7EF4-B035-4FE1-A476-0676F7251200</ProjectGuid>`) and the IntegrationTests project (`<ProjectGuid>1EE3612B-FCA4-4DA5-BA9E-741A9D8DC59E</ProjectGuid>`). 

@dmitry-golovinov-sonarsource - do you think it is possible , for the changes in commit b3ba0c5 , to use `git mv` for the issues in the main project? This will help to see easily the history of the file, in case we want to look at historical differences for a particular rule (even if the issue location changed, in many cases the same issues are still present, they've just be translated inside the file after more code has been added).


Probably a script to iterate through the issues that were raised on the old code and do something like (the project guid stays the same, only `SXXXX` changes):
```
git mv AutoMapper-025B9F93-7953-454B-AF29-E6A55B2AA58F-S103.json AutoMapper--S103.json
```



And once you add back `<SonarQubeTestProject>true</SonarQubeTestProject>` to the {Unit,Integration}Tests.csproj files, it should only be one JSON file with S2699 for each, and then it will be easy to `git mv` the 2 files.

analyzers/its/sources/AutoMapper/src/AutoMapper/AutoMapper.csproj
```diff
@@ -3,55 +3,56 @@
...
-    <RepositoryUrl>https://github.com/AutoMapper/AutoMapper</RepositoryUrl>
+    <PackageIcon>icon.png</PackageIcon>
+    <PackageProjectUrl>https://automapper.org</PackageProjectUrl>
+    <Features Condition=" '$(TargetFramework)' == 'net461' ">peverify-compat</Features>
+
+    <MinVerTagPrefix>v</MinVerTagPrefix>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
-    <CodeAnalysisRuleSet>..\..\..\..\ValidationRuleset.ruleset</CodeAnalysisRuleSet>
```
andrei-epure-sonarsource, line 20: this is important in all csproj files, please bring back as well

andrei-epure-sonarsource:
Also, please make whatever changes you are going to do in separate commits, to ease the CR.

andrei-epure-sonarsource (approved):
LGTM

PR #4532, opened 2021-06-09
Add ITs scripts to solution

andrei-epure-sonarsource:


analyzers/SonarAnalyzer.sln
```diff
@@ -57,7 +57,9 @@ Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "SonarAnalyzer.Shared", "src
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "GitHub", "..\.github\GitHub.shproj", "{532AC70E-90CB-48AD-86F4-16767C22314D}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SonarAnalyzer.ShimLayer.CodeGeneration", "src\SonarAnalyzer.ShimLayer.CodeGeneration\SonarAnalyzer.ShimLayer.CodeGeneration.csproj", "{684F0AD7-BBC9-40C8-9E54-D9C2B57560D9}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SonarAnalyzer.ShimLayer.CodeGeneration", "src\SonarAnalyzer.ShimLayer.CodeGeneration\SonarAnalyzer.ShimLayer.CodeGeneration.csproj", "{684F0AD7-BBC9-40C8-9E54-D9C2B57560D9}"
+EndProject
+Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "IntegrationTests", "its\IntegrationTests.shproj", "{AF787E26-7AB1-4B08-AF05-07CB0ECDE4E5}"
```
pavel-mikula-sonarsource, line 62: `ITs` is actually a real folder in the project structure, just as `Tests`.

Let's make it look the same as `Tests` folder in the project solution - beside the `Common Solution Items`.

analyzers/its/IntegrationTests.shproj
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- This is a minimal Shared project to show directory structure in Solution Explorer. This project should not be build nor referenced. -->
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.CSharp.targets" />
+  <ItemGroup>
+    <None Include="*.ps1" />
+    <None Include="*ruleset*" />
+    <None Include="*.targets" />
```
pavel-mikula-sonarsource, line 8: Let's include just all of them, to avoid surprises/maintenance cost in the future. For example the `.gitignore` could also be useful to ahve.
```suggestion
    <None Include="*" />
```

pavel-mikula-sonarsource (changes requested):

analyzers/SonarAnalyzer.sln
```diff
@@ -57,7 +57,9 @@ Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "SonarAnalyzer.Shared", "src
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "GitHub", "..\.github\GitHub.shproj", "{532AC70E-90CB-48AD-86F4-16767C22314D}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SonarAnalyzer.ShimLayer.CodeGeneration", "src\SonarAnalyzer.ShimLayer.CodeGeneration\SonarAnalyzer.ShimLayer.CodeGeneration.csproj", "{684F0AD7-BBC9-40C8-9E54-D9C2B57560D9}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SonarAnalyzer.ShimLayer.CodeGeneration", "src\SonarAnalyzer.ShimLayer.CodeGeneration\SonarAnalyzer.ShimLayer.CodeGeneration.csproj", "{684F0AD7-BBC9-40C8-9E54-D9C2B57560D9}"
```
andrei-epure-sonarsource, line 60: not sure why this wasn't here before, but it looks like it should be in the solution

analyzers/SonarAnalyzer.sln
pavel-mikula-sonarsource: My original intention was to have structure like this
```
Solution `SonarAnalyzer`/ITs/<something>`
Solution `SonarAnalyzer`/Tests/SonarAnalyzer.UnitTest`
```
where I don't have a good idea for `<something>` (`ITs/ITs`, `ITs/Root`, `ITs/Files`, `ITs/Common`) to avoid `Common Solution Items` in the way
```
Solution `SonarAnalyzer`/Common Solution Items/its`
```

analyzers/SonarAnalyzer.sln
pavel-mikula-sonarsource, line 60: Any idea why this GUID got changed? Probably VS auto update

analyzers/SonarAnalyzer.sln
```diff
@@ -57,7 +57,9 @@ Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "SonarAnalyzer.Shared", "src
...
+EndProject
+Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "its", "its\its.shproj", "{AF787E26-7AB1-4B08-AF05-07CB0ECDE4E5}"
```
pavel-mikula-sonarsource, line 62: If we don't come up with good idea for `<something>` from the other comment, let's just make it `ITs` and leave it in `Common Solution Items`
```suggestion
Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "ITs", "its\its.shproj", "{AF787E26-7AB1-4B08-AF05-07CB0ECDE4E5}"
```

pavel-mikula-sonarsource (changes requested):

analyzers/SonarAnalyzer.sln
andrei-epure-sonarsource: I kept the spelling of the folder, based on your comment:
> ITs is actually a real folder in the project structure, just as Tests.

pavel-mikula-sonarsource: I'd still prefer `ITs` as it's easier to read. `.github` is also lowercase on the filesystem

analyzers/SonarAnalyzer.sln
andrei-epure-sonarsource: I think it's just VS auto-update, indeed. Do you think it's a problem to checkin these changes? I don't think we're going to look into the history of the GUID

pavel-mikula-sonarsource: I'm fine with it, let's keep it. I was just curious in case there's something wrong going on.

analyzers/its/expected/expected.shproj
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- This is a minimal Shared project to show the directory structure in Solution Explorer. This project should not be built nor referenced. -->
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.CSharp.targets" />
+  <ItemGroup>
+    <None Include="*\**" />
+    <None Include="*" />
```
pavel-mikula-sonarsource, line 7: This should not be needed, the previous line will take care of it
```suggestion
```

analyzers/its/config/config.shproj
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- This is a minimal Shared project to show the directory structure in Solution Explorer. This project should not be built nor referenced. -->
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.CSharp.targets" />
+  <ItemGroup>
+    <None Include="*\**" />
+    <None Include="*" />
```
pavel-mikula-sonarsource, line 7: This should not be needed, the previous line will take care of it
```suggestion
```

analyzers/SonarAnalyzer.sln
```diff
@@ -57,7 +57,17 @@ Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "SonarAnalyzer.Shared", "src
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "GitHub", "..\.github\GitHub.shproj", "{532AC70E-90CB-48AD-86F4-16767C22314D}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SonarAnalyzer.ShimLayer.CodeGeneration", "src\SonarAnalyzer.ShimLayer.CodeGeneration\SonarAnalyzer.ShimLayer.CodeGeneration.csproj", "{684F0AD7-BBC9-40C8-9E54-D9C2B57560D9}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SonarAnalyzer.ShimLayer.CodeGeneration", "src\SonarAnalyzer.ShimLayer.CodeGeneration\SonarAnalyzer.ShimLayer.CodeGeneration.csproj", "{684F0AD7-BBC9-40C8-9E54-D9C2B57560D9}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "ITs", "ITs", "{314A0C27-4768-4F15-B7FA-DC3C33C08180}"
+EndProject
+Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "config", "its\config\config.shproj", "{5D6668F5-61B7-488E-9B8A-958FCEF05E71}"
```
pavel-mikula-sonarsource, line 64: Can we make these look nice pleaaase 😃 (Git isn't friendly with case-only file renames. I'm usually creating commit like `xConfig.shproj` and second with `Config.shproj`)
```suggestion
Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "Config", "its\config\Config.shproj", "{5D6668F5-61B7-488E-9B8A-958FCEF05E71}"
```

pavel-mikula-sonarsource (approved):
LGTM, nice idea of adding more in the `ITs` folder.
There are few minor things to make it nice.

analyzers/its/expected/expected.shproj
andrei-epure-sonarsource: it didn't, this is why I added line 7
but I'll replace both with `**\**`

analyzers/SonarAnalyzer.sln
andrei-epure-sonarsource: I just renamed the projects to upper-case and re-imported them

PR #4544, opened 2021-06-14
Enable repeatable package restores using lock files

### LINKED ISSUE #4534: Add packages.lock.json for our dependencies
Bring back what was removed in  #3429. We can add lock files w/o caching. Let's try and add back packages.lock.json and see if we still have (intermittent) problems with NET 4.6.1 , and then contact the Nuget team if it still happens.

costin-zaharia-sonarsource:
Fixes #4534


costin-zaharia-sonarsource:
In order to correctly re-generate the lock files, please run the following:
```
dotnet nuget locals all --clear
git clean -xfd
git rm **/packages.lock.json -f
nuget restore -LockedMode -ConfigFile "analyzers\NuGet.Config" analyzers\SonarAnalyzer.sln
```

andrei-epure-sonarsource (approved):
LGTM

PR #4550, opened 2021-06-15
S3329: fix FN with implicitly typed arrays

### LINKED ISSUE #4274: Rule S3329: FN when IV byte array is created without a `new`expression
### Description

The following FN is based on a test case from [Juliet Test Suite for C#](https://samate.nist.gov/SARD/testsuite.php)

```csharp
using System;

using System.IO;
using System.Security.Cryptography;

namespace testcases.CWE329_Not_Using_Random_IV_with_CBC_Mode
{
  class CWE329_Not_Using_Random_IV_with_CBC_Mode__basic_01 : AbstractTestCase
  {
    public override void Bad()
    {
      /* FLAW: hardcoded initialization vector used */
      byte[] initializationVector =
      {
        0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
        0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00
      };
      using (AesCryptoServiceProvider aes = new AesCryptoServiceProvider())
      {
        ICryptoTransform encryptor = aes.CreateEncryptor(aes.Key, initializationVector); //Noncompliant; FN here
      }
    }
  }
}
```

Creating the initialization vectore byte array using the `new` keyword makes the issue appear.

andrei-epure-sonarsource:
Fixes #4274

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/Checks/ByteArrayCheck.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/Checks/ByteArrayCheck.cs
@@ -19,6 +19,7 @@
  */
 
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using SonarAnalyzer.Helpers;
 using SonarAnalyzer.SymbolicExecution.Common.Constraints;
@@ -32,10 +33,19 @@ namespace SonarAnalyzer.SymbolicExecution.Common.Checks
         public override ProgramState PostProcessInstruction(ProgramPoint programPoint, ProgramState programState) =>
             programPoint.CurrentInstruction switch
             {
+                InitializerExpressionSyntax initializerExpression => ImplicitlyTypedArrayPostProcess(initializerExpression, programState),
                 ArrayCreationExpressionSyntax arrayCreation => ArrayCreationPostProcess(arrayCreation, programState),
                 InvocationExpressionSyntax invocation => InvocationExpressionPostProcess(invocation, programState),
                 _ => programState
             };
+
+        private ProgramState ImplicitlyTypedArrayPostProcess(InitializerExpressionSyntax initializerExpression, ProgramState programState) =>
+            !initializerExpression.Parent.IsKind(SyntaxKind.ArrayCreationExpression)
+            && initializerExpression.FirstAncestorOrSelf<VariableDeclarationSyntax>() is VariableDeclarationSyntax variableDeclaration
+            && semanticModel.GetTypeInfo(variableDeclaration.Type).Type.Is(KnownType.System_Byte_Array)
+            && programState.HasValue
+                ? programState.SetConstraint(programState.PeekValue(), ByteArraySymbolicValueConstraint.Constant)
+                : programState;
 
         private ProgramState ArrayCreationPostProcess(ArrayCreationExpressionSyntax arrayCreation, ProgramState programState) =>
             semanticModel.GetTypeInfo(arrayCreation).Type.Is(KnownType.System_Byte_Array) && programState.HasValue
--- a/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/Checks/ByteArrayCheck.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/Checks/ByteArrayCheck.cs
@@ -33,22 +33,24 @@ namespace SonarAnalyzer.SymbolicExecution.Common.Checks
         public override ProgramState PostProcessInstruction(ProgramPoint programPoint, ProgramState programState) =>
             programPoint.CurrentInstruction switch
             {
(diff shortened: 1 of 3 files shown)


analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/Checks/ByteArrayCheck.cs
```diff
@@ -42,6 +44,16 @@ programPoint.CurrentInstruction switch
                 ? programState.SetConstraint(programState.PeekValue(), ByteArraySymbolicValueConstraint.Constant)
                 : programState;
 
+        private ProgramState ImplicitlyTypedArrayPostProcess(InitializerExpressionSyntax initializerExpression, ProgramState programState) =>
+            initializerExpression.IsKind(SyntaxKind.ArrayInitializerExpression)
```
andrei-epure-sonarsource, line 48: it's a bit difficult to realize which condition is not met by the coverage

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/Checks/ByteArrayCheck.cs
```diff
@@ -42,6 +44,16 @@ programPoint.CurrentInstruction switch
...
+            // when the initializer is in a typed array creation, it is handled by ArrayCreationPostProcess()
+            && !initializerExpression.Parent.IsKind(SyntaxKind.ArrayCreationExpression)
+            && initializerExpression.FirstAncestorOrSelf<VariableDeclarationSyntax>() is VariableDeclarationSyntax variableDeclaration
+            && semanticModel.GetTypeInfo(variableDeclaration.Type).Type.Is(KnownType.System_Byte_Array)
+            && programState.HasValue
+                ? programState.SetConstraint(programState.PeekValue(), ByteArraySymbolicValueConstraint.Constant)
```
pavel-mikula-sonarsource, line 54: We're putting `Constant` constraint on every initialization. Let's document FP in case like this (I don't think it's worth inspecting the values to be constant-only in this super-safe approach :) ):
```
            using var rng = new RNGCryptoServiceProvider();
            byte[] initializationVector =
            {
                Rnd(), Rnd(), Rnd(), Rnd(), Rnd(), Rnd(), Rnd(), Rnd(),
                Rnd(), Rnd(), Rnd(), Rnd(), Rnd(), Rnd(), Rnd(), Rnd()
            };

            using (AesCryptoServiceProvider aes = new AesCryptoServiceProvider())
            {
                ICryptoTransform encryptor = aes.CreateEncryptor(aes.Key, initializationVector); // Noncompliant FP, implicitly typed array initialization considers the values as constant
            }

            byte Rnd()
            {
                var buff = new byte[] { 0 };
                rng.GetBytes(buff);
                return buff[0];
            }
```

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/Checks/ByteArrayCheck.cs
pavel-mikula-sonarsource: I wonder how coverage got to 9/10. By brute force measuring, I get 7/10.
These conditions are never `false`:
`initializerExpression.IsKind(SyntaxKind.ArrayInitializerExpression)`
`initializerExpression.FirstAncestorOrSelf<VariableDeclarationSyntax>() is VariableDeclarationSyntax variableDeclaration`
`programState.HasValue`

pavel-mikula-sonarsource (approved):
LGTM. I left one (nonactionable?) coverage comment and one FP to document in UTs

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InitializationVectorShouldBeRandom.cs
```diff
@@ -260,6 +257,17 @@ public void ImplicitlyTypedArrayNonByte()
             }
         }
 
+        public void CollectionInitializer()
+        {
+            List<byte> listWithConstant = new List<byte> { 0x00 };
+            List<byte> list = new List<byte> { Rnd() };
+            using (AesCryptoServiceProvider aes = new AesCryptoServiceProvider())
+            {
+                ICryptoTransform encryptor = aes.CreateEncryptor(aes.Key, listWithConstant.ToArray()); // FN
```
andrei-epure-sonarsource, line 266: I'll send an email to @pierre-loup-tristant-sonarsource  about this FN. 
Let's keep the scope of the PR limited and only fix the initial ticket.

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/Checks/ByteArrayCheck.cs
andrei-epure-sonarsource: I am puzzled, I added a test with CollectionInitialzer and it's still at 9/10
image

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InitializationVectorShouldBeRandom.cs
andrei-epure-sonarsource: follow-up https://github.com/SonarSource/sonar-dotnet/issues/4565

PR #4551, opened 2021-06-15
Providing example for S4018 FP when type parameter is used in return value.

dmitry-golovinov-sonarsource:


costin-zaharia-sonarsource (approved):
LGTM!

PR #4553, opened 2021-06-15
GitHub workflow: Don't run CreatePullRequest for bots

pavel-mikula-sonarsource:
Dependabot is creating PRs directly from our repository, so it passes the original `if` condition.
It doesn't have enough permissions and creates errors like this https://github.com/SonarSource/sonar-dotnet/pull/4543/checks?check_run_id=2817846021

We don't need a card for bot PRs. It will be created later, when needed by other actions (request (self) review, approve review)

### DIFF
--- a/.github/workflows/CreatePullRequest.yml
+++ b/.github/workflows/CreatePullRequest.yml
@@ -8,8 +8,12 @@ jobs:
   assign_and_create_card:
     name: Assign issue to sender and create Kanban card
     runs-on: ubuntu-latest
+    # Single quotes must be used here https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#literals
     # PRs from forks don't have required token authorization
-    if: github.event.pull_request.head.repo.full_name == github.repository
+    # Dependabot works directly under our repository, but doesn't have enough priviledges to create project card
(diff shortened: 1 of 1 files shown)


.github/workflows/CreatePullRequest.yml
```diff
@@ -8,8 +8,12 @@ jobs:
   assign_and_create_card:
     name: Assign issue to sender and create Kanban card
     runs-on: ubuntu-latest
+    # Single quotes must be used here https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#literals
     # PRs from forks don't have required token authorization
-    if: github.event.pull_request.head.repo.full_name == github.repository
+    # Dependabot works directly under our repository, but doesn't have enough priviledges to create project card
+    if: |
+        github.event.pull_request.head.repo.full_name == github.repository
+        && github.event.sender.type != 'Bot'
```
pavel-mikula-sonarsource, line 16: Bots are recognizable like this, as seen here:
https://github.com/pavel-mikula-sonarsource/GitHubActionPlayground/runs/2829424837?check_suite_focus=true#step:2:461

andrei-epure-sonarsource (approved):
LGTM

PR #4557, opened 2021-06-15
GitHub workflow: Label issue to Backlog

pavel-mikula-sonarsource:
When Label `Type: xxx` is added to an open issue:

* create card in `Backlog` project and associated column 
* move card to the new column if exists

Column selection is based on the column name. Column-related API is available in preview mode (special header tag is needed to gain access).

If two or more `Type: xxx` labels are added (invalid state as per process), last queued action wins.

### Getting correct project ID 

Isn't easy: 
https://docs.github.com/en/rest/reference/projects#list-repository-projects

This script can be used (after fitting with github token instead of `FIXME`):
https://github.com/pavel-mikula-sonarsource/GitHubActionPlayground/blob/8a932277c20ffdfc97c66e6b8468bc65989621fe/index.js#L15-L26

### Demo

https://github.com/pavel-mikula-sonarsource/GitHubActionPlayground/projects/2

https://github.com/pavel-mikula-sonarsource/GitHubActionPlayground/blob/master/.github/workflows/LabelIssue.yml

### DIFF
--- a/.github/workflows/LabelIssue.yml
+++ b/.github/workflows/LabelIssue.yml
@@ -1,0 +1,68 @@
+name: Issue Labeled
+
+on:
+  issues:
+    types: ["labeled"]
+
+jobs:
+  add_to_backlog:
+    name: Log
+    runs-on: ubuntu-latest
+    # Single quotes must be used here https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#literals
+    # Only limited global functions are available in this context https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#functions
+    if: |
+        github.event.issue.state == 'Open'
+        && startsWith(github.event.label.name, 'Type: ')
+
+    steps:
+      # https://github.com/actions/github-script
+      - uses: actions/github-script@v4.0.2
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const BACKLOG_PROJECT = 3897364;
+            const mediaType = { previews: ['inertia'] }; // Column related APIs are in Alpha Preview. We need to set this HTTP Header to gain access.
+            //
+            async function loadColumns() {
+                const columns = await github.projects.listColumns({ project_id: BACKLOG_PROJECT, mediaType });
+                const ret = new Map();
+                for (let column of columns.data) {
+                    ret.set(column.name, column.id);
+                }
+                return ret;
+            }
+            //
+            const map = await loadColumns();
+            //
+            async function findCard(content_url) {
+                // Columns are searched from the most probable one
+                for (let columnId of map.values()) {
+                    let cards = await github.projects.listCards({ column_id: columnId });
+                    let card = cards.data.find(x => x.content_url.endsWith(content_url)); // "https://" is missing from event payload
+                    if (card) {
+                        return card;
+                    }
+                }
+                console.log("Card not found for: " + content_url);
+                return null;
+            }
+            //
+            let name = context.payload.label.name;
+            if (name.startsWith("Type: ")) {
+                name = name.substring(6);
+                if (map.has(name)) {
+                    const newColumn = map.get(name);
+                    const card = await findCard(context.payload.issue.url);
+                    if (card) {
+                        console.log("Moving card to column: " + name);
+                        github.projects.moveCard({ card_id: card.id, position: "bottom", column_id: newColumn });
+                    } else {
+                        console.log("Creating card in column: " + name);
+                        github.projects.createCard({ column_id: newColumn, content_id: context.payload.issue.id, content_type: "Issue" });
+                    }
+                } else {
+                    console.log("Backlog column doesn't exist: " + name);
+                }
+            } else {
+                console.log("Unexpected label name: " + name);
+            }
--- a/.github/workflows/LabelIssue.yml
+++ b/.github/workflows/LabelIssue.yml
@@ -23,21 +23,20 @@ jobs:
             const BACKLOG_PROJECT = 3897364;
             const mediaType = { previews: ['inertia'] }; // Column related APIs are in Alpha Preview. We need to set this HTTP Header to gain access.
             //
-            async function loadColumns() {
+            async function loadColumnMap() {
                 const columns = await github.projects.listColumns({ project_id: BACKLOG_PROJECT, mediaType });
                 const ret = new Map();
                 for (let column of columns.data) {
-                    ret.set(column.name, column.id);
+                    ret.set(column.name, column);
                 }
                 return ret;
             }
             //
-            const map = await loadColumns();
+            const columnMap = await loadColumnMap();
             //
             async function findCard(content_url) {
-                // Columns are searched from the most probable one
-                for (let columnId of map.values()) {
-                    let cards = await github.projects.listCards({ column_id: columnId });
+                for (let column of columnMap.values()) {
+                    let cards = await github.projects.listCards({ column_id: column.id });
                     let card = cards.data.find(x => x.content_url.endsWith(content_url)); // "https://" is missing from event payload
                     if (card) {
                         return card;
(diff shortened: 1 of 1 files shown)


.github/workflows/LabelIssue.yml
```diff
@@ -0,0 +1,68 @@
...
+        && startsWith(github.event.label.name, 'Type: ')
+
+    steps:
+      # https://github.com/actions/github-script
+      - uses: actions/github-script@v4.0.2
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const BACKLOG_PROJECT = 3897364;
+            const mediaType = { previews: ['inertia'] }; // Column related APIs are in Alpha Preview. We need to set this HTTP Header to gain access.
```
andrei-epure-sonarsource, line 24: is `inertia` some sort of project codename?

.github/workflows/LabelIssue.yml
```diff
@@ -0,0 +1,68 @@
...
+                for (let column of columns.data) {
+                    ret.set(column.name, column.id);
+                }
+                return ret;
+            }
+            //
+            const map = await loadColumns();
+            //
+            async function findCard(content_url) {
+                // Columns are searched from the most probable one
```
andrei-epure-sonarsource, line 38: this is a bit ambiguous: what does "most probable" one mean? it seems random to me.

.github/workflows/LabelIssue.yml
```diff
@@ -0,0 +1,68 @@
...
+            async function loadColumns() {
+                const columns = await github.projects.listColumns({ project_id: BACKLOG_PROJECT, mediaType });
+                const ret = new Map();
+                for (let column of columns.data) {
+                    ret.set(column.name, column.id);
+                }
+                return ret;
+            }
+            //
+            const map = await loadColumns();
```
andrei-epure-sonarsource, line 35: maybe name this backlogColumnMap?

.github/workflows/LabelIssue.yml
```diff
@@ -0,0 +1,68 @@
...
+                }
+                console.log("Card not found for: " + content_url);
+                return null;
+            }
+            //
+            let name = context.payload.label.name;
+            if (name.startsWith("Type: ")) {
+                name = name.substring(6);
+                if (map.has(name)) {
+                    const newColumn = map.get(name);
```
andrei-epure-sonarsource, line 54: maybe name `columnForType` instead of `newColumn`? because if the card wasn't in any column before, there isn't really "old" (as the word "new" suggests)

.github/workflows/LabelIssue.yml
```diff
@@ -0,0 +1,68 @@
...
+                    let card = cards.data.find(x => x.content_url.endsWith(content_url)); // "https://" is missing from event payload
+                    if (card) {
+                        return card;
+                    }
+                }
+                console.log("Card not found for: " + content_url);
+                return null;
+            }
+            //
+            let name = context.payload.label.name;
```
andrei-epure-sonarsource, line 50: suggest to name it `labelName`

.github/workflows/LabelIssue.yml
```diff
@@ -0,0 +1,68 @@
...
+            if (name.startsWith("Type: ")) {
+                name = name.substring(6);
```
andrei-epure-sonarsource, line 52: suggest to use a new variable `let typeName = labelName...`

and line 57 and line 60 and line 64 logs need to be updated with `typeName`
and line 67 with `labelName`

.github/workflows/LabelIssue.yml
```diff
@@ -0,0 +1,68 @@
...
+                const ret = new Map();
+                for (let column of columns.data) {
+                    ret.set(column.name, column.id);
+                }
+                return ret;
+            }
+            //
+            const map = await loadColumns();
+            //
+            async function findCard(content_url) {
```
andrei-epure-sonarsource, line 37: suggest to name `findCardInBacklog`

(yes this whole thing is about the backlog, but you can imagine a reader may think that this could also interact with the Kanban project )

.github/workflows/LabelIssue.yml
```diff
@@ -0,0 +1,68 @@
...
+            }
+            //
+            let name = context.payload.label.name;
+            if (name.startsWith("Type: ")) {
+                name = name.substring(6);
+                if (map.has(name)) {
+                    const newColumn = map.get(name);
+                    const card = await findCard(context.payload.issue.url);
+                    if (card) {
+                        console.log("Moving card to column: " + name);
```
andrei-epure-sonarsource, line 57:
```suggestion
                        console.log("Moving backlog card to column: " + name);
```

.github/workflows/LabelIssue.yml
```diff
@@ -0,0 +1,68 @@
...
+                        github.projects.moveCard({ card_id: card.id, position: "bottom", column_id: newColumn });
+                    } else {
+                        console.log("Creating card in column: " + name);
```
andrei-epure-sonarsource, line 60:
```suggestion
                        console.log("Creating backlog card in column: " + name);
```

andrei-epure-sonarsource (changes requested):
great to see this! 

I've left some comments

.github/workflows/LabelIssue.yml
pavel-mikula-sonarsource: I have no idea what that means :) Looks like a project codename. It's copy/paste from their [docs](https://docs.github.com/en/rest/reference/projects#list-project-columns).

.github/workflows/LabelIssue.yml
pavel-mikula-sonarsource: Copy/Paste issue. It shouldn't be there anymore.

.github/workflows/LabelIssue.yml
pavel-mikula-sonarsource: I'll rename it to `columnMap`. This fragment will be reused in other places soon (and then unified to common lib)

.github/workflows/LabelIssue.yml
pavel-mikula-sonarsource: This fragment (moveOrCreateCard) will be also moved to a lib so I'd like to have a generic name. It isn't perfect, but it's good enough. It's the column where the card will end.

.github/workflows/LabelIssue.yml
pavel-mikula-sonarsource: `columnName`

.github/workflows/LabelIssue.yml
pavel-mikula-sonarsource: I'll keep it generic, as it will be moved to a lib. The Kanban actions also don't have Kanban word in the name.

.github/workflows/LabelIssue.yml
pavel-mikula-sonarsource: This is going to be moved to a lib, so I'll keep it generic. The context is clear in this case and there will be `Moving card to column: False Positive` in most cases, so it will not be confused with Kanban

.github/workflows/LabelIssue.yml
pavel-mikula-sonarsource: Same as above

.github/workflows/LabelIssue.yml
```diff
@@ -0,0 +1,67 @@
...
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const BACKLOG_PROJECT = 3897364;
+            const mediaType = { previews: ['inertia'] }; // Column related APIs are in Alpha Preview. We need to set this HTTP Header to gain access.
+            //
+            async function loadColumnMap() {
+                const columns = await github.projects.listColumns({ project_id: BACKLOG_PROJECT, mediaType });
+                const ret = new Map();
+                for (let column of columns.data) {
+                    ret.set(column.name, column);
```
pavel-mikula-sonarsource, line 30: This change (from `column.id` to `column`) isn't needed by this PR, but will make our life easier when moving it to a common lib

pavel-mikula-sonarsource:
Changes are also reflected in the Playground repo for testing

andrei-epure-sonarsource (approved):
LGTM

PR #4558, opened 2021-06-15
Extract GetDeclarationTypeName

costin-zaharia-sonarsource:
- to enable reuse and remove code duplication


dmitry-golovinov-sonarsource (changes requested):
1. Let's get rid of code smell
2. Let's add the test to SyntaxNodeExtensionsTest.cs to have 100% coverage
3. You can also reuse it in "ImplementIDisposableCorrectly" (line 66)

dmitry-golovinov-sonarsource (approved):
LGTM!

PR #4563, opened 2021-06-16
Don't fail the parallel build if another process is writing to xml file

dmitry-golovinov-sonarsource:


andrei-epure-sonarsource (approved):
LGTM. I am running locally and if all green I will merge this :D

PR #4564, opened 2021-06-16
Enable security hotspot rules in ITs automatically

### LINKED ISSUE #1934: Enable security hotspot rules in ITs automatically
#### WHY?
The security hotspots are enabled only when their ID is present in SonarLint.xml, which is created by Scanner for MSBuild. Our C# ITs don't use Scanner for MSBuild but have a hardcoded [SonarLint.xml](https://github.com/SonarSource/sonar-csharp/blob/master/sonaranalyzer-dotnet/its/SonarLint.xml). When adding new security hotspot rules we need to manually update the aforementioned file.

#### WHAT
We need to develop automatic hotspot enabling when the ITs are run so that we don't forget to execute the rules.

pavel-mikula-sonarsource:
Fixes #1934


analyzers/its/regression-test.ps1
```diff
@@ -28,13 +28,40 @@ if ($PSBoundParameters['Verbose'] -Or $PSBoundParameters['Debug']) {
...
+
+    $SourcePath = ".\config\$ProjectName\SonarLint.xml"
+    if(-Not (Test-Path $SourcePath)){
+        $SourcePath = ".\config\SonarLint.xml"
+    }
+    $Content = Get-Content -Path $SourcePath -Raw
+
+    if($ruleId){
+        $RuleFragment = "    <Rule><Key>$ruleId</Key></Rule>"
+    } else {
```
pavel-mikula-sonarsource, line 43: This logic is new.
_Old behavior:_ When $ruleId input parameter was specified, analyzer run this rule and all security hotspots. Creating some noise and wait time on the way.
_New behavior:_ Run only $ruleId

analyzers/its/expected/Ember-MM/Ember Media Manager-{9B57D3AB-AF12-4012-B945-284C2448DC81}-S4036.json
```diff
@@ -0,0 +1,316 @@
+{
+"issues":  [
+{
+"id":  "S4036",
```
pavel-mikula-sonarsource, line 4: And the winner is... S4036 that was missing in SonarLint.xml files 😄

analyzers/its/SonarAnalyzer.Testing.ImportBefore.targets
```diff
@@ -51,7 +44,7 @@
 
       <!-- Add the SonarAnalyzer analyzer DLLs -->
       <Analyzer Include="$(MSBuildStartupDirectory)\binaries\SonarAnalyzer*.dll" />
-      <AdditionalFiles Include="$(ProjectSpecificSonarLintXml)" />
+      <AdditionalFiles Include="$(MSBuildStartupDirectory)\output\$(PROJECT)\SonarLint.xml" />
```
pavel-mikula-sonarsource, line 47: Change from `\config\` to `\output\` directory

costin-zaharia-sonarsource (approved):
LGTM!

PR #4566, opened 2021-06-16
RollForward for akka.net sdk version

andrei-epure-sonarsource:
no need to have a fixed version installed locally


dmitry-golovinov-sonarsource (approved):
LGTM!

PR #4570, opened 2021-06-17
Add lock files for IT projects

costin-zaharia-sonarsource:
And enable locked mode restore for ITs


pavel-mikula-sonarsource (approved):
LGTM

PR #4571, opened 2021-06-17
Move SonarLint.xml to Directory.Build.Props

costin-zaharia-sonarsource:


analyzers/Directory.Build.props
```diff
@@ -6,6 +6,15 @@
              If a setting should only apply to certain projects then add it at a
              lower level.
    -->
+   
+  <PropertyGroup>
+    <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
+    <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
```
dmitry-golovinov-sonarsource, line 12: Don't we need to add <RestoreLockedMode>true</RestoreLockedMode>as well?

dmitry-golovinov-sonarsource (approved):
LGTM! Please check my comment.

analyzers/Directory.Build.props
costin-zaharia-sonarsource: `RestoreLockedMode` is used during the build as an option: See here: https://github.com/SonarSource/sonar-dotnet/blob/master/azure-pipelines.yml#L42

PR #4572, opened 2021-06-17
Fix S125 FP : URL containing } at the end

### LINKED ISSUE #2772: Fix S125 FP : URL containing `}` at the end
### Description

[community issue](https://community.sonarsource.com/t/s125-with-url-in-the-comment/16188)

See [implementation](https://github.com/SonarSource/sonar-dotnet/blob/e56271226a3b64daf46ad780d15397b8d798a353/sonaranalyzer-dotnet/src/SonarAnalyzer.CSharp/Rules/CommentedOutCode.cs#L191) ) to start investigation

### Repro steps

```
//http://localhost:7071/runtime/webhooks/EventGrid?functionName={functionname} // Noncompliant
```

### Expected behavior

This is not code, it's just a URL. Should not raise.

### Actual behavior

S125 gets raised considering it code

### Related information

* SonarC# Version 8.0

pavel-mikula-sonarsource:
Fixes #2772


analyzers/src/SonarAnalyzer.CSharp/Rules/CommentedOutCode.cs
```diff
@@ -34,163 +34,115 @@ namespace SonarAnalyzer.Rules.CSharp
...
-            if (trivia.IsKind(SyntaxKind.SingleLineCommentTrivia))
+            string TriviaContent()
             {
-                if (triviaContent.StartsWith("//", StringComparison.Ordinal))
-                {
-                    triviaContent = triviaContent.Substring(2);
-                }
-
-                return triviaContent;
+                var content = trivia.ToString().Substring(CommentMarkLength);
```
pavel-mikula-sonarsource, line 102: _About these coverage changes_, Roslyn is directly going for the substrings as well, when comment kind is present
https://github.com/dotnet/roslyn/blob/main/src/Analyzers/CSharp/Analyzers/FileHeaders/CSharpFileHeaderHelper.cs#L28

analyzers/src/SonarAnalyzer.CSharp/Rules/CommentedOutCode.cs
```diff
@@ -70,57 +71,44 @@ private static void CheckTrivias(SyntaxTreeAnalysisContext context, IEnumerable<
...
                 }
             }
-        }
 
-        private static string GetTriviaContent(SyntaxTrivia trivia)
-        {
-            const int commentLength = 2;
-            var content = trivia.ToString();
-            if (trivia.IsKind(SyntaxKind.MultiLineCommentTrivia))
+            string TriviaContent()
```
andrei-epure-sonarsource, line 100: I suggest to use a static local functions instead of a context-capturing local functions (which behind the scenes need another class to be created) - especially for a one-liner function. The static local function is implemented as a private method by the compiler. Context-capturing needs another structure for the context. See on [sharplab](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEBMBGAsAKHQGYACLEgdSgBcALAYQgAdqBXAJ1gHMSBvAkoLKl0AFhIBZABQBKPgKGLY1EgA8SAXhLYA3AsWDlJAJ6aSjFhwCm9APYxqV1dVl78BofoNGLbdjftHZ1lNAD41EmQSTDdFAF8CBPwCYjJMEgA5W18Obnl3IVSxSRD+Ao8jdS1dLyUHEgAvMwBlaghqKABjABlbTogAG3QAVg77KVUZWI8SWsLsADYSI1b2rt7+odGoccq5DXD1KJivJKSgA==).

I suggest to rename it to `TrimCommentMarks` and then `TrimCommentMarks(trivia)` would be easier to read then `TriviaContent()` which is a method that doesn't contain a verb (code smell).

andrei-epure-sonarsource (changes requested):
nice refactoring!

analyzers/src/SonarAnalyzer.CSharp/Rules/CommentedOutCode.cs
```diff
@@ -34,163 +34,115 @@ namespace SonarAnalyzer.Rules.CSharp
...
                 });
-        }
 
-        private static void CheckTrivias(IEnumerable<SyntaxTrivia> trivias, SyntaxTreeAnalysisContext context)
+        private static void CheckTrivias(SyntaxTreeAnalysisContext context, IEnumerable<SyntaxTrivia> trivias)
         {
             var shouldReport = true;
             foreach (var trivia in trivias)
             {
                 // comment start is checked because of  https://github.com/dotnet/roslyn/issues/10003
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: In my understanding, the issue is still relevant as we can run on old Roslyn.
The 2nd part of the `&&` is dealing with it `... StartsWith("/**",...`
Swapping this condition is a performance optimization, because `IsKind` is super cheap and `.ToFullString()` is little bit more expensive.

analyzers/src/SonarAnalyzer.CSharp/Rules/CommentedOutCode.cs
pavel-mikula-sonarsource: Using static local function would shadow the input argument and add unneeded indirection.
The verb is represented by `()` invocation, meaning `Get`.

The MSIL overhead under the hood is pretty small compared to what we do in `TokenTypeAnalyzerBase` where we create class with several collection fields for each token in source code.

analyzers/src/SonarAnalyzer.CSharp/Rules/CommentedOutCode.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: You're right, either way the check is still the same. I've got confused.

analyzers/src/SonarAnalyzer.CSharp/Rules/CommentedOutCode.cs
andrei-epure-sonarsource: I won't insist, it's a matter of preference.

The shadowing of local input doesn't need to happen - you just use a different name.

andrei-epure-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.CSharp/Rules/CommentedOutCode.cs
andrei-epure-sonarsource: And I don't agree with the verb being represented by `()` .

Issue #4573, opened 2021-06-17
Fix AD0001 with Symbolic Execution engine


andrei-epure-sonarsource:
This snippet leads to an uncaugth exception in the symbolic execution engine:

```
public class Reproducer
{
    private DateTime? foo;

    public virtual DateTime? Foo
    {
        get { return foo; }

        set
        {
            if (value.HasValue)
            {
                value = DateTime.SpecifyKind(value.Value, DateTimeKind.Unspecified);
            }

            if (foo != value || (foo.HasValue && value.HasValue && foo.Value.Kind != value.Value.Kind))
            {
                foo = value;
                Bar("x");
            }
        }
    }

    public void Bar(string x) { }
}
```

Stack trace:
```
[...]

PR #4574, opened 2021-06-18
Add repro for AD0001 for SE

andrei-epure-sonarsource:
Reproducer for #4573

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/Rules/SymbolicExecution/ConditionEvaluatesToConstantTest.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/Rules/SymbolicExecution/ConditionEvaluatesToConstantTest.cs
@@ -74,7 +74,6 @@ namespace SonarAnalyzer.UnitTest.Rules
             Verifier.VerifyAnalyzerFromCSharp9Console(@"TestCases\ConditionEvaluatesToConstant.CSharp9.TopLevelStatements.cs", GetAnalyzer());
 #endif
 
-        // https://github.com/SonarSource/sonar-dotnet/issues/4573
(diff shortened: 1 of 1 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/Rules/SymbolicExecution/ConditionEvaluatesToConstantTest.cs
```diff
@@ -72,6 +74,45 @@ public class ConditionEvaluatesToConstantTest
...
+            }
+        }
+    }
+    public void Bar(string x) { }
+}", GetAnalyzer(), CompilationErrorBehavior.Ignore);
+
+            // https://github.com/SonarSource/sonar-dotnet/issues/4573
+            action.Should().Throw<AssertFailedException>().Where(e =>
+                e.Message.Contains("Expected diagnostics {error AD0001: Analyzer 'SonarAnalyzer.Rules.SymbolicExecution.SymbolicExecutionRunner' threw an exception of type " +
+                "'SonarAnalyzer.SymbolicExecution.SymbolicExecutionException' with message 'Error processing method: set_Foo ## Method file: snippet1.cs " +
```
andrei-epure-sonarsource, line 111: FYI Dima, because Roslyn keeps only the first line of an exception message, we concatenate all the lines in a single one where we replace newline by `##`. Like this, we can copy-paste the line locally and replace `##` with newline and have a beautiful stacktrace when examining some issue.

dmitry-golovinov-sonarsource (approved):
LGTM! Thanks for the insight!

dmitry-golovinov-sonarsource:
Not merging, since master is red.

PR #4577, opened 2021-06-18
 Fix S5693 FN: Should check web.config file

### LINKED ISSUE #4165: Fix S5693 FN: Should check web.config file
```
<configuration>
    <system.web>
        <httpRuntime maxRequestLength="81920" executionTimeout="3600" />
        <!-- Sensitive: maxRequestLength is exprimed in KB, so 81920KB = 80MB  -->
    </system.web>
    <system.webServer>
        <security>
            <requestFiltering>
                <requestLimits maxAllowedContentLength="83886080" />
                <!-- Sensitive: maxAllowedContentLength is exprimed in bytes, so 83886080B = 80MB  -->
            </requestFiltering>
        </security>
    </system.webServer>
</configuration>
```

two attributes are relevant for this rules:
* `maxRequestLength` exprimed in kilobytes
* `maxAllowedContentLength` exprimed in bytes

so if:
* `maxRequestLength` is defined and superior to 8192 then it's sensitive (> 8MB the default value for this parameterizable rule)
* `maxAllowedContentLength` is defined and superior to 8388608 then it's sensitive (> 8MB the default value for this parameterizable rule)

the default values in dotnet are:
* [4096](https://docs.microsoft.com/en-us/dotnet/api/system.web.configuration.httpruntimesection.maxrequestlength?view=netframework-4.8) KB for `maxRequestLength`
* [30000000](https://docs.microsoft.com/en-us/iis/configuration/system.webserver/security/requestfiltering/requestlimits/) for `maxAllowedContentLength` 

30000000 (28MB) is superior by default for what we recommend by default (8 MB), so we should not raise when these attributes are not defined, because there is a protection even if is more than what we recommend, we should only when explicitely defined and superior to the limit defined on this rule (8MB by default).

dmitry-golovinov-sonarsource:
Fixes #4165


analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -136,6 +144,43 @@ private bool IsEnabled(AnalyzerOptions options)
             return SupportedDiagnostics.Any(d => analyzerConfiguration.IsEnabled(d.Id));
         }
 
+        private void CheckWebConfig(SonarAnalysisContext context, CompilationAnalysisContext c)
+        {
+            foreach (var fullPath in context.GetWebConfig(c))
+            {
+                var webConfig = File.ReadAllText(fullPath);
+                if (webConfig.Contains("<system.web") && XmlHelper.ParseXDocument(webConfig) is { } doc)
```
andrei-epure-sonarsource, line 152: what if it contains only `<system.webServer`?

maybe add tests with only `system.web` and only `system.webServer`

and with attributes which have the sought names, but are not under the correct XML elements

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -136,6 +144,43 @@ private bool IsEnabled(AnalyzerOptions options)
...
+                if (requestLimit.Attribute("maxAllowedContentLength") is { } maxAllowedContentLength
+                    && IsVulnerable(maxAllowedContentLength.Value, MaxAllowedContentLength)
+                    && maxAllowedContentLength.CreateLocation(webConfigPath) is { } location)
+                {
+                    c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, location));
+                }
+            }
+        }
+
+        private static bool IsVulnerable(string value, int limit) =>
```
andrei-epure-sonarsource, line 181: if it fails to parse, you consider it vulnerable ("hotspot")? 

It's not necessarily a bad thing, I guess (it's a hotspot, not a vulnerability), but can a webapp truly work with non-integer values? and if it doesn't, should we raise a security hotspot if there's actually a bug there?

Also, I see that for this scenario you added `Noncompliant FP` - so why keep it like this?

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -136,6 +144,43 @@ private bool IsEnabled(AnalyzerOptions options)
...
+            !(int.TryParse(value, out var val) && val <= limit);
```
andrei-epure-sonarsource, line 182: I see `!(............ val <= limit)` and it's hurting my brain, because if `value <= limit`, it should be ok.

And then I see the `!` (this is why we have https://rules.sonarsource.com/csharp/RSPEC-1940 , but it didn't detect this)

I suggest:
```
! int.TryParse(value, out var val)
|| val > limit
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/WebConfig/RequestsWithExcessiveLength/Corrupt/Web.config
```diff
@@ -0,0 +1,10 @@
+﻿<configuration>
+  <system.web>
+    <httpRuntime maxRequestLength="81920" executionTimeout="3600" />
+    <!-- Sensitive: maxRequestLength is exprimed in KB, so 81920KB = 80MB  -->
```
andrei-epure-sonarsource, line 4: could you please remove the `Sensitive` comments? Because I have to think if this means an issue is found in the corrupt file

analyzers/its/expected/Nancy/Nancy.ViewEngines.Razor--net452-S5693.json
```diff
@@ -0,0 +1,30 @@
+{
+"issues":  [
+{
+"id":  "S5693",
+"message":  "Make sure the content length limit is safe here.",
+"location":  {
+"uri":  "sources\Nancy\samples\Nancy.Demo.Hosting.Owin\Web.config",
```
andrei-epure-sonarsource, line 7: I see all issues which are raised are raised on the same file: `sources\Nancy\samples\Nancy.Demo.Hosting.Owin\Web.config` , even if the projects that get analyzed are different

Could you look why this is happening? Does this mean that in SonarQube multiple issues for the same locations will be raised?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/WebConfig/RequestsWithExcessiveLength/Values/ValidValues/Web.config
```diff
@@ -0,0 +1,12 @@
+<configuration>
+  <system.web>
+    <httpRuntime maxRequestLength="8192" executionTimeout="3600" />
```
andrei-epure-sonarsource, line 3: maybe also add with smaller values than the limit? like 4 MB

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/WebConfig/RequestsWithExcessiveLength/Values/CornerCases/Web.config
```diff
@@ -0,0 +1,12 @@
+<configuration>
+  <system.web>
+    <httpRuntime maxRequestLength="819200000000000000000" executionTimeout="3600" /> <!-- Noncompliant -->
```
andrei-epure-sonarsource, line 3: other edge cases:
- `1`
- `0`
- `-1`
- `-2147483648`
- `0x7FFFFFFF`

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
dmitry-golovinov-sonarsource: I've simplified it in a different way, so that it won't raise FP

analyzers/its/expected/Nancy/Nancy.ViewEngines.Razor--net452-S5693.json
dmitry-golovinov-sonarsource: The way we build FilesToAnalyze.txt in our ITs - they are all of the same size and contain all the files for this project. That's why you see this entry even in unrelated projects.

andrei-epure-sonarsource: hm ok, we should add a comment somewhere about this...
Not very sure where - in the Target that adds the FilesToAnalyze.txt

but that would be a separate PR

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -178,8 +178,14 @@ private void ReportRequestLengthViolation(XDocument doc, string webConfigPath, C
             }
         }
 
-        private static bool IsVulnerable(string value, int limit) =>
-            !(int.TryParse(value, out var val) && val <= limit);
+        private static bool IsVulnerable(string value, int limit)
```
andrei-epure-sonarsource, line 181: you can make it an arrow function
```
int.TryParse(value, out var val) 
  ? val > limit
  : false;
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/WebConfig/RequestsWithExcessiveLength/Values/WrongPath/Web.config
```diff
@@ -0,0 +1,12 @@
+<configuration>
```
andrei-epure-sonarsource, line 1: please add comments to explain that `httpRuntime should be inside system.web` and `requestLimits should be inside requestFiltering` to help the reader quickly understand the test case

also, I would change `WrongPath` to something else like `InvalidConfig`  because WrongPath can refer to the FileSystem and would introduce confusion to the reader.

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
andrei-epure-sonarsource: what I meant was adding a file
```
<configuration>
  <system.web>
    <httpRuntime maxRequestLength="81920" executionTimeout="3600" /> <!-- Noncompliant {{Make sure the content length limit is safe here.}} -->
  </system.web>
</configuration>
```

and a file

```
<configuration>
  <system.webServer>
    <security>
      <requestFiltering>
        <requestLimits maxAllowedContentLength="83886080" />         <!-- Noncompliant {{Make sure the content length limit is safe here.}} -->
      </requestFiltering>
    </security>
  </system.webServer>
</configuration>
```

I expect this last one to fail because of this `if`.

andrei-epure-sonarsource (changes requested):
almost there

analyzers/its/expected/Nancy/Nancy.ViewEngines.Razor--net452-S5693.json
andrei-epure-sonarsource: (I am leaving this thread open for a reader of the PR to easily find this conversation in case he/she will wonder why this is happening)

dmitry-golovinov-sonarsource:
@andrei-epure-sonarsource  reply to your last comment - added and it didn't fail

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
dmitry-golovinov-sonarsource: It didn't fail, since <system.web> and <system.webServer> both start with <system.web

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
```diff
@@ -21,7 +21,10 @@
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.IO;
```
pavel-mikula-sonarsource, line 24: Don't forget to update RSPEC for the rule to ship the `web.config` part with it.

andrei-epure-sonarsource (approved):
LGTM, only the RPSEC should be updated with the web.config sample

analyzers/src/SonarAnalyzer.Common/Rules/Hotspots/RequestsWithExcessiveLengthBase.cs
andrei-epure-sonarsource: Ah nice! My bad

analyzers/rspec/cs/S5693_c#.html
```diff
@@ -59,6 +76,23 @@ <h2>Compliant Solution</h2>
     }
 }
 </pre>
+<p>In <a href="https://docs.microsoft.com/en-us/troubleshoot/aspnet/create-web-config">Web.config</a>:</p>
```
andrei-epure-sonarsource, line 79: You should check the other rules with web.config, but normally we put links at the end, all together

andrei-epure-sonarsource: Also, I don't think the troubleshoot link is the best to share

dmitry-golovinov-sonarsource: TBH the way it was done in the RSPEC I like more, since those links at the end concern only security related links. And all non-security related appear in the text where they need to be.

Updated PR, please give your feedback.

andrei-epure-sonarsource (approved):
LGTM!

_(not merging to master yet, waiting to see if it's frozen for the release or not)_

analyzers/rspec/cs/S5693_c#.html
andrei-epure-sonarsource: I agree it's debatable which is best. I favor consistency however until a cross-LT guideline is created.

analyzers/src/SonarAnalyzer.Utilities/Rules.Description/S5693_cs.html
```diff
@@ -38,6 +38,23 @@ <h2>Sensitive Code Example</h2>
     }
 }
 </pre>
+<p>In <a href="https://docs.microsoft.com/en-us/troubleshoot/aspnet/create-web-config">Web.config</a>:</p>
```
andrei-epure-sonarsource, line 41: Oops I missed this - the two html files are different.

The RSPECs should be updated in their repo https://github.com/SonarSource/rspec and then the rule-api should be invoked to generated the files.

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Utilities/Rules.Description/S5693_cs.html
dmitry-golovinov-sonarsource: Updated from RSPEC

andrei-epure-sonarsource (approved):
LGTM, great work!

_(should be merged after the version bump PR gets merged)_

PR #4578, opened 2021-06-18
Delete S1145 (deprecated)

costin-zaharia-sonarsource:


analyzers/its/AllSonarAnalyzerRules.ruleset
```diff
@@ -1152,7 +1152,6 @@
     <Rule Id="S1142" Action="Warning" />
     <Rule Id="S1143" Action="Warning" />
     <Rule Id="S1144" Action="Warning" />
-    <Rule Id="S1145" Action="Warning" />
```
andrei-epure-sonarsource, line 1155: this file just has all possible rule IDs from S1 to S10000 - so most of the IDs here do not correspond to actual rules

I think you should revert it to avoid confusion

analyzers/tests/SonarAnalyzer.UnitTest/PackagingTests/RuleTypeTest.cs
```diff
@@ -32,29 +33,37 @@ namespace SonarAnalyzer.UnitTest.PackagingTests
     [TestClass]
     public class RuleTypeTest
     {
+        private readonly HashSet<string> deletedRules = new () {"S1145"};
```
andrei-epure-sonarsource, line 36: maybe add a comment - "rules that have been deprecated and deleted" - not to be in contradiction with the `YOU SHOULD NEVER DELETE RULES!` message

analyzers/its/SingleRule.ruleset.template
```diff
@@ -1152,7 +1152,6 @@
     <Rule Id="S1142" Action="None" />
     <Rule Id="S1143" Action="None" />
     <Rule Id="S1144" Action="None" />
-    <Rule Id="S1145" Action="None" />
```
andrei-epure-sonarsource, line 1155: same comment as for `analyzers/its/AllSonarAnalyzerRules.ruleset` - this file is generated without being too smart about which rules are actually implemented

this is only used to activate a single rule during ITs - if someone activates `S9998` it won't do anything

I believe you should revert it

analyzers/tests/SonarAnalyzer.UnitTest/PackagingTests/RuleTypeTest.cs
```diff
@@ -63,8 +72,7 @@ private static void DetectTypeChanges(ResourceManager resourceManager, IImmutabl
 
             // IMPORTANT: Rules should not be deleted without careful consideration and prior
             // deprecation. We need to notify the platform team as well.
```
andrei-epure-sonarsource, line 74:
```suggestion
            // deprecation. We need to notify the product teams (SQ, SC, SL) as well.
```

andrei-epure-sonarsource (changes requested):
Nice to see we're removing some of the old deprecated rules 👍 

akka.net and EmberMM have issues in the expected results for this rule, should also remove them

analyzers/tests/SonarAnalyzer.UnitTest/PackagingTests/RuleTypeTest.cs
andrei-epure-sonarsource: IMO this comment about notification should also be put above the `deletedRules` field - I find it's a good reminder

andrei-epure-sonarsource (approved):
LGTM

PR #4579, opened 2021-06-18
Remove duplication when referencing StyleCopAnalyzers

costin-zaharia-sonarsource:
If we add the reference in `Directory.Build.props` it will be automatically available for all the projects.

Changes done:
- We introduced a problem with lock files (appeared only on a clean build): we need to reference parent `Directory.Build.props` in tests, otherwise it is not considered
- Fix the assembly name of the Vsix project (previously it was the same with SonarAnalyzer.Common which caused a conflict on NuGet restore)
- Used PackageReference instead of packages file in the VSIX project (we cannot use mixed and in `StyleCopAnalyzers.targets` package reference was used)
- Remove duplication where referencing `StyleCopAnalyzers.targets`

### DIFF
--- a/analyzers/src/SonarAnalyzer.CFG/SonarAnalyzer.CFG.csproj
+++ b/analyzers/src/SonarAnalyzer.CFG/SonarAnalyzer.CFG.csproj
@@ -63,6 +63,4 @@
     <Exec WorkingDirectory="$(OutDir)" Command="&quot;$(SIGNTOOL_PATH)&quot; sign /fd SHA256 /f $(PFX_PATH) /p $(PFX_PASSWORD) /tr http://sha256timestamp.ws.symantec.com/sha256/timestamp SonarAnalyzer.CFG.dll" />
   </Target>
 
-  <Import Project="..\..\StyleCopAnalyzers.targets" />
-
 </Project>
--- a/analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.Vsix.csproj
+++ b/analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.Vsix.csproj
@@ -1,6 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\packages\Microsoft.VSSDK.BuildTools.15.9.3032\build\Microsoft.VSSDK.BuildTools.props" Condition="Exists('..\..\packages\Microsoft.VSSDK.BuildTools.15.9.3032\build\Microsoft.VSSDK.BuildTools.props')" />
   <PropertyGroup>
     <MinimumVisualStudioVersion>15.0</MinimumVisualStudioVersion>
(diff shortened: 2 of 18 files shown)


analyzers/tests/Directory.Build.props
```diff
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
 
+  <!-- Import the higher level common properties -->
+  <Import Project="..\Directory.Build.props" />
```
costin-zaharia-sonarsource, line 5: We need to reference parent the `Directory.Build.props` in tests, otherwise it is not considered and the configuration is not applied.

analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.Vsix.csproj
```diff
@@ -19,7 +18,7 @@
     <OutputType>Library</OutputType>
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>SonarAnalyzer.Vsix</RootNamespace>
-    <AssemblyName>SonarAnalyzer</AssemblyName>
+    <AssemblyName>SonarAnalyzer.Vsix</AssemblyName>
```
costin-zaharia-sonarsource, line 21: It was the same assembly name with `SonarAnalyzer.Common`. NuGet restore can get confused due to this

analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.Vsix.csproj
```diff
@@ -122,16 +120,20 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
-    <Reference Include="Google.Protobuf, Version=3.6.1.0, Culture=neutral, PublicKeyToken=a7d26565bac4d604, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Google.Protobuf.3.6.1\lib\net45\Google.Protobuf.dll</HintPath>
```
costin-zaharia-sonarsource, line 126: I like a lot more `PackageReference`. No more relative paths.

pavel-mikula-sonarsource: 🎉 It's so much better now

analyzers/src/SonarAnalyzer.CFG/SonarAnalyzer.CFG.csproj
```diff
@@ -63,6 +63,4 @@
     <Exec WorkingDirectory="$(OutDir)" Command="&quot;$(SIGNTOOL_PATH)&quot; sign /fd SHA256 /f $(PFX_PATH) /p $(PFX_PASSWORD) /tr http://sha256timestamp.ws.symantec.com/sha256/timestamp SonarAnalyzer.CFG.dll" />
   </Target>
 
-  <Import Project="..\..\StyleCopAnalyzers.targets" />
```
pavel-mikula-sonarsource, line 66: This can be also removed from `SonarAnalyzser.UnitTest.csproj`

pavel-mikula-sonarsource (approved):
LGTM. Please delete one more line before merging 😄

PR #4580, opened 2021-06-18
Fix S3353 FP: Variable is used within an Expression

### LINKED ISSUE #4015: Fix S3353 FP: Variable is used within an Expression

pavel-mikula-sonarsource:
Fixes #4015

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
@@ -41,7 +41,13 @@ namespace SonarAnalyzer.Rules.CSharp
         private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
-        private enum DeclarationType { CannotBeConst, Value, Reference, String };
+        private enum DeclarationType
+        {
+            CannotBeConst,
+            Value,
+            Reference,
+            String
+        }
 
         protected override void Initialize(SonarAnalysisContext context) =>
             context.RegisterSyntaxNodeActionInNonGenerated(c =>
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
@@ -180,10 +180,10 @@ namespace SonarAnalyzer.Rules.CSharp
                     var lookup = new CSharpMethodParameterLookup(invocation, semanticModel);
                     return lookup.TryGetSymbol(argument, out var parameter) && parameter.IsType(KnownType.System_Linq_Expressions_Expression_T);
                 }
-                else if (lambda.Parent is AssignmentExpressionSyntax assignment // Lambda cannot be on the left side
-                    && semanticModel.GetSymbolInfo(assignment.Left).Symbol is { } assignmentTargetSymbol)
+                else if (lambda.Parent is AssignmentExpressionSyntax assignment)
                 {
-                    return assignmentTargetSymbol.GetSymbolType().Is(KnownType.System_Linq_Expressions_Expression_T);
+                    // Lambda cannot be on the left side, we don't need to check it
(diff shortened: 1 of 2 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
```diff
@@ -154,16 +160,35 @@ private static bool HasMutableUsagesInMethod(VariableDeclaratorSyntax variable,
...
+        {
+            if (identifier.FirstAncestorOrSelf<LambdaExpressionSyntax>().GetSelfOrTopParenthesizedExpression() is { } lambda)
+            {
+                if (lambda.Parent is ArgumentSyntax argument
+                    && argument.FirstAncestorOrSelf<InvocationExpressionSyntax>() is { } invocation)
+                {
+                    var lookup = new CSharpMethodParameterLookup(invocation, semanticModel);
+                    return lookup.TryGetSymbol(argument, out var parameter) && parameter.IsType(KnownType.System_Linq_Expressions_Expression_T);
+                }
+                else if (lambda.Parent is AssignmentExpressionSyntax assignment // Lambda cannot be on the left side
```
pavel-mikula-sonarsource, line 183: AFAICT 😃 or can it? 🤔

costin-zaharia-sonarsource: I think not.

analyzers/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
```diff
@@ -154,16 +160,35 @@ private static bool HasMutableUsagesInMethod(VariableDeclaratorSyntax variable,
...
+                    && semanticModel.GetSymbolInfo(assignment.Left).Symbol is { } assignmentTargetSymbol)
+                {
+                    return assignmentTargetSymbol.GetSymbolType().Is(KnownType.System_Linq_Expressions_Expression_T);
+                }
```
costin-zaharia-sonarsource, line 187: We can use `assignment.Left.IsKnownType(KnownType.System_Linq_Expressions_Expression_T, semanticModel)` and simplify the code a bit.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UnchangedLocalVariablesShouldBeConst.cs
```diff
@@ -352,13 +352,75 @@ public NormalClass NormalClassSetToNull()
...
+
+    public void Noncompliant_Assignment()
+    {
+        string a = null;    // Noncompliant
+        string b = null;    // Noncompliant
+        string c = null;    // Noncompliant
+        var repro = new Repro_4015();
+        Func<string> variableFunc;
+
+        // Compliant cases, changing the Expression<Func<T>> below changes the behavior of the code
```
costin-zaharia-sonarsource, line 413: The cases are marked as compliant but the variables are noncompliant. Did you forgot to remove the comment?

costin-zaharia-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/UnchangedLocalVariablesShouldBeConst.cs
pavel-mikula-sonarsource: I don't agree on the `a bit`. It's `quite a lot` 😄

costin-zaharia-sonarsource (approved):
LGTM!

PR #4581, opened 2021-06-18
Update RSPEC

pavel-mikula-sonarsource:
Update RSPEC before release

This is first RSPEC update after rule-api v2.0 release

I did some manual changes, please re-run the rule-api to see and validate them.


analyzers/rspec/cs/S2326_c#.json
```diff
@@ -12,5 +12,10 @@
   "defaultSeverity": "Major",
   "ruleSpecification": "RSPEC-2326",
   "sqKey": "S2326",
-  "scope": "All"
+  "scope": "All",
+  "securityStandards": {
+    "CERT": [
+      "MSC12-CPP."
+    ]
```
costin-zaharia-sonarsource, line 19: Isn't this wrong? To have a cpp standard on a C# project?

pavel-mikula-sonarsource: CERT removed manually.
There might be some more of them going away after RULEAPI-665 is fixed in further releases. There should be no harm, as they aren't user facing according to slack message.

analyzers/rspec/cs/S1117_c#.json
```diff
@@ -13,5 +13,7 @@
   "defaultSeverity": "Major",
   "ruleSpecification": "RSPEC-1117",
   "sqKey": "S1117",
-  "scope": "All"
+  "scope": "All",
+  "securityStandards": {
+  }
```
costin-zaharia-sonarsource, line 18: Would be great to have the tool updated to skip the generation of empty sections.

analyzers/rspec/cs/S2386_c#.html
```diff
@@ -33,7 +33,7 @@ <h2>Exceptions</h2>
       <li> <code>System.Collections.Immutable.IImmutableSet&lt;T&gt;</code> </li>
       <li> <code>System.Collections.Immutable.IImmutableStack&lt;T&gt;</code> </li>
       <li> <code>System.Collections.Immutable.IImmutableQueue&lt;T&gt;</code> </li>
-    </ul> </li>
+    </ul>  </li>
```
costin-zaharia-sonarsource, line 36: Strange that we get an extra space here.

analyzers/rspec/cs/S4581_c#.html
```diff
@@ -10,6 +10,8 @@ <h2>Noncompliant Code Example</h2>
 public void Foo()
 {
     var g = new Guid(); // Noncompliant - what's the intent?
+    var guid = default(Guid);
+    Guid other = default;
```
costin-zaharia-sonarsource, line 14: We don't support this yet. Needs to be reverted.

analyzers/rspec/cs/S1117_c#.json
pavel-mikula-sonarsource: That was probably just me, with manual edits.

analyzers/src/SonarAnalyzer.Utilities/Rules.Description/S4581.html
```diff
@@ -10,6 +10,8 @@ <h2>Noncompliant Code Example</h2>
 public void Foo()
 {
     var g = new Guid(); // Noncompliant - what's the intent?
+    var guid = default(Guid);
+    Guid other = default;
```
costin-zaharia-sonarsource, line 14: Here also. Probably this one will get fixed automatically when the other one is fixed.

analyzers/src/SonarAnalyzer.Utilities/Rules.Description/S4784_cs.html
```diff
@@ -85,6 +85,4 @@ <h2>See</h2>
   </li>
   <li> OWASP Regular expression Denial of Service - ReDoS </li>
 </ul>
-<h2>Deprecated</h2>
-<p>This rule is deprecated; use {rule:csharpsquid:S2631} instead.</p>
```
costin-zaharia-sonarsource, line 89: I think we should keep this message. It's relevant for the user. Another tool issue?

costin-zaharia-sonarsource (changes requested):
Please see my comments

analyzers/rspec/cs/S1117_c#.json
costin-zaharia-sonarsource: Ah, this appears in a lot of places.

analyzers/rspec/cs/S2386_c#.html
pavel-mikula-sonarsource: Source: https://raw.githubusercontent.com/SonarSource/rspec/master/rules/S2386/csharp/rule.adoc

Seems just like the way adoc serializes these nested bulletpoints.

analyzers/rspec/cs/S4581_c#.html
pavel-mikula-sonarsource: Nice catch!

analyzers/src/SonarAnalyzer.Utilities/Rules.Description/S4784_cs.html
pavel-mikula-sonarsource: This rule was marked as `ready` during migration from Jira (see this [thread](https://sonarsource.slack.com/archives/CFSLVKNBB/p1624024528158900) ) while it should remain deprecated. So the tool itself is probably fine.

I just missed the fact that this should be undone.

analyzers/rspec/cs/S1117_c#.json
pavel-mikula-sonarsource: I just rerun that 6 LOC script I used to remove the CERT :)

costin-zaharia-sonarsource (approved):
LGTM!

PR #4583, opened 2021-06-18
Remove duplication in AssemblyInfo files

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.Vsix.csproj
```diff
@@ -64,9 +64,6 @@
     <None Include="SonarAnalyzer.VisualBasic.nuspec">
```
andrei-epure-sonarsource, line 64: this didn't get propagated to the VSIX proj - maybe link it to analyzers/src/Directory.Build.props ?

andrei-epure-sonarsource (approved):
LGTM
I also checked [the artifacts](https://dev.azure.com/sonarsource/DotNetTeam%20Project/_build/results?buildId=33626&view=artifacts&pathAsName=false&type=publishedArtifacts) to have the correct version, all good.

analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.Vsix.csproj
costin-zaharia-sonarsource: It seems to be a bug in visual studio when displaying the link: https://developercommunity.visualstudio.com/t/File-links-added-in-DirectoryBuildprop/1456131?entry=problem

Actually the file is considered during build (we can easily test that by adding a duplicate attribute).

PR #4586, opened 2021-06-21
Remove duplicate property and reduce the scope for the warning suppression

costin-zaharia-sonarsource:
`DisableImplicitNuGetFallbackFolder` it is already defined in Directory.Build.props


csaba-sagi-sonarsource (approved):
LGTM!

PR #4590, opened 2021-06-21
S3247: introduce C#9 patterns support

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/CastShouldNotBeDuplicated.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/CastShouldNotBeDuplicated.cs
@@ -38,7 +38,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private const string DiagnosticId = "S3247";
         private const string MessageFormat = "{0}";
         private const string ReplaceWithAsAndNullCheckMessage = "Replace this type-check-and-cast sequence with an 'as' and a null check.";
-        private const string RemoveRedundantCaseMessage = "Remove this redundant cast.";
+        private const string RemoveRedundantCaseMessage = "Remove this cast and use the appropriate variable.";
 
         private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
@@ -59,8 +59,14 @@ namespace SonarAnalyzer.Rules.CSharp
             {
                 return;
             }
-            ProcessPatternExpression(analysisContext, casePatternSwitch.Pattern, parentSwitchStatement.Expression, parentSwitchStatement.Expression.GetLocation(), parentSwitchStatement);
+            ProcessPatternExpression(analysisContext,
+                                     casePatternSwitch.Pattern,
+                                     parentSwitchStatement.Expression,
+                                     parentSwitchStatement.Expression.GetLocation(),
+                                     parentSwitchStatement,
+                                     RemoveRedundantCaseMessage);
         }
+
         private static void SwitchExpressionArm(SyntaxNodeAnalysisContext analysisContext)
         {
             var isSwitchExpression = (SwitchExpressionArmSyntaxWrapper)analysisContext.Node;
@@ -71,7 +77,12 @@ namespace SonarAnalyzer.Rules.CSharp
                 return;
             }
             var switchExpression = (SwitchExpressionSyntaxWrapper)parent;
-            ProcessPatternExpression(analysisContext, isSwitchExpression.Pattern, switchExpression.GoverningExpression, switchExpression.GoverningExpression.GetLocation(), isSwitchExpression);
+            ProcessPatternExpression(analysisContext,
+                                     isSwitchExpression.Pattern,
+                                     switchExpression.GoverningExpression,
+                                     switchExpression.GoverningExpression.GetLocation(),
+                                     isSwitchExpression,
+                                     RemoveRedundantCaseMessage);
         }
 
         private static void IsPatternExpression(SyntaxNodeAnalysisContext analysisContext)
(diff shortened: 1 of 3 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CastShouldNotBeDuplicated.CSharp9.cs
```diff
@@ -31,23 +32,64 @@ public void Foo(Object x, Object y)
...
             }
 
-            if (x is Fruit f2)
+            if ((x, y) is (Fruit f2, Vegetable v2))   // Secondary
             {
-                var ff = (Fruit)f2; // FN
+                var ff2 = (Vegetable)v2;              // Noncompliant
             }
 
+            if ((x, y) is (Fruit f3, Vegetable v3))   // Noncompliant
```
costin-zaharia-sonarsource, line 46: I think we should add also a few cases with `is not`

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CastShouldNotBeDuplicated.CSharp9.cs
```diff
@@ -31,23 +32,64 @@ public void Foo(Object x, Object y)
...
+                var ff9 = (Fruit)x;
+            }
+
+            x is Fruit f0; // Error [CS0201]
+
             var message = x switch
             {
-                Fruit f3 => ((Fruit)f3).ToString(), // FN
-                Vegetable v3 => ((Vegetable)v3).ToString(), // FN
+                Fruit f10 => ((Fruit)f10).ToString(), // FN
```
costin-zaharia-sonarsource, line 85: We should add also a branch with a tuple pattern inside the switch

analyzers/src/SonarAnalyzer.CSharp/Rules/CastShouldNotBeDuplicated.cs
```diff
@@ -26,24 +27,59 @@
...
+        {
+            var isPatternExpression = (IsPatternExpressionSyntaxWrapper)analysisContext.Node;
+            if (!(isPatternExpression.SyntaxNode.GetFirstNonParenthesizedParent() is IfStatementSyntax parentIfStatement))
+            {
+                return;
+            }
+            var isPatternLocation = isPatternExpression.SyntaxNode.GetLocation();
+            if (isPatternExpression.Pattern.SyntaxNode.IsKind(SyntaxKindEx.RecursivePattern))
+            {
+                var recursivePattern = (RecursivePatternSyntaxWrapper)isPatternExpression.Pattern.SyntaxNode;
```
costin-zaharia-sonarsource: As far as I can tell we handle only the first level of the recursive pattern. 
I think we should either support imbrication or document the FN.
e.g.
```
void M(object x, object y)
    {
        if ((x, y) is ((int a, int b), string v3))
        {
            
        }
    }
```

dmitry-golovinov-sonarsource: Will document it as FN.

analyzers/src/SonarAnalyzer.CSharp/Rules/CastShouldNotBeDuplicated.cs
```diff
@@ -26,55 +27,162 @@
...
-                castTypeSymbol.TypeKind == TypeKind.Struct)
+        private static void CasePatternSwitchLabel(SyntaxNodeAnalysisContext analysisContext)
+        {
+            var casePatternSwitch = (CasePatternSwitchLabelSyntaxWrapper)analysisContext.Node;
+            if (!(casePatternSwitch.SyntaxNode.GetFirstNonParenthesizedParent().GetFirstNonParenthesizedParent() is SwitchStatementSyntax parentSwitchStatement))
             {
                 return;
             }
+            ProcessPatternExpression(analysisContext, casePatternSwitch.Pattern, parentSwitchStatement.Expression, parentSwitchStatement.Expression.GetLocation(), parentSwitchStatement);
+        }
```
csaba-sagi-sonarsource, line 63: Empty line between methods is missing.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CastShouldNotBeDuplicated.CSharp9.cs
```diff
@@ -13,41 +15,108 @@ class Program
...
         {
             var fruit = x as Fruit;
-            if (fruit is not Fruit) // FN
+            if (fruit is not Fruit) // Compliant, redundant condition, not related for the current rule
             {
             }
 
             object o;
-            switch (x)
+            switch (x)                            // Noncompliant [switch-st-0]
```
csaba-sagi-sonarsource, line 23: I am not sure the suggestion to use 'as' an null check is a valid one for a switch. So, maybe we should suggest somehow to use the u variable for example in line 32. Or just say to remove the redundant cast.

csaba-sagi-sonarsource:
This case should be also covered:
            if (x is Fruit { Prop: 1} tuttyFrutty)
            {
                var aRealFruit = (Fruit)tuttyFrutty;
                var anotherFruit = (Fruit)x;
            }

csaba-sagi-sonarsource:
And also:
            if (x is (Fruit or Vegetable))
            {
                var fruit = (Fruit)x;
                var vegetable = (Vegetable)x;
            }

csaba-sagi-sonarsource (changes requested):
Please cover the cases I suggested, or at least document them as FNs.

dmitry-golovinov-sonarsource:
> 
> 
> And also:
> if (x is (Fruit or Vegetable))
> {
> var fruit = (Fruit)x;
> var vegetable = (Vegetable)x;
> }

this case was added as a FN

analyzers/src/SonarAnalyzer.CSharp/Rules/CastShouldNotBeDuplicated.cs
```diff
@@ -26,55 +27,205 @@
...
+                .Where(x => x.Type.WithoutTrivia().IsEquivalentTo(castType.WithoutTrivia())
+                            && IsCastOnSameSymbol(x))
+                .Select(x => x.GetLocation()).ToList();
+
+            bool IsCastOnSameSymbol(CastExpressionSyntax castExpression) =>
+                Equals(analysisContext.SemanticModel.GetSymbolInfo(castExpression.Expression).Symbol, typeExpressionSymbol);
+        }
+
+        private static void ProcessPatternExpression(SyntaxNodeAnalysisContext analysisContext,
+            SyntaxNode isPattern,
```
csaba-sagi-sonarsource, line 139: For some methods you aligned the parameters under each other and on some places not. See line 156. As we still do not have a clear coding style for that I would only ask you to align them to be consistent. For whichever option you opt.

dmitry-golovinov-sonarsource: Done!

csaba-sagi-sonarsource (approved):
LGTM!

PR #4592, opened 2021-06-22
Rule S3240: add support for C# 9 pattern matching

costin-zaharia-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CFG/Extensions/IsPatternExpressionSyntaxWrapperExtensions.cs
+++ b/analyzers/src/SonarAnalyzer.CFG/Extensions/IsPatternExpressionSyntaxWrapperExtensions.cs
@@ -28,8 +28,9 @@ namespace SonarAnalyzer.Extensions
             isPatternWrapper.Pattern.IsNull();
 
         public static bool IsNotNull(this IsPatternExpressionSyntaxWrapper isPatternWrapper) =>
-            UnaryPatternSyntaxWrapper.IsInstance(isPatternWrapper.Pattern)
-            && ((UnaryPatternSyntaxWrapper)isPatternWrapper.Pattern) is var unaryPatternSyntaxWrapper
+            isPatternWrapper.Pattern.RemoveParentheses() is var syntaxNode
+            && UnaryPatternSyntaxWrapper.IsInstance(syntaxNode)
+            && ((UnaryPatternSyntaxWrapper)syntaxNode) is var unaryPatternSyntaxWrapper
             && unaryPatternSyntaxWrapper.IsNot()
             && unaryPatternSyntaxWrapper.Pattern.IsNull();
     }
--- a/analyzers/src/SonarAnalyzer.CFG/Extensions/IsPatternExpressionSyntaxWrapperExtensions.cs
+++ b/analyzers/src/SonarAnalyzer.CFG/Extensions/IsPatternExpressionSyntaxWrapperExtensions.cs
@@ -1,0 +1,36 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2021 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using StyleCop.Analyzers.Lightup;
+
+namespace SonarAnalyzer.Extensions
+{
+    public static class IsPatternExpressionSyntaxWrapperExtensions
+    {
+        public static bool IsNull(this IsPatternExpressionSyntaxWrapper isPatternWrapper) =>
+            isPatternWrapper.Pattern.IsNull();
+
+        public static bool IsNotNull(this IsPatternExpressionSyntaxWrapper isPatternWrapper) =>
+            UnaryPatternSyntaxWrapper.IsInstance(isPatternWrapper.Pattern)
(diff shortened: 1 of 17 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/Extensions/PatternSyntaxWrapperExtensionsTest.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace SonarAnalyzer.UnitTest.Extensions
+{
+    [TestClass]
+    public class PatternSyntaxWrapperExtensionsTest
+    {
+        [TestMethod]
+        public void IsNull_WithDifferentPatternThanConstant_ReturnsFalse()
+        {
+            var isPattern = (IsPatternExpressionSyntaxWrapper)SyntaxFactory.ParseExpression("is not 1");
+            isPattern.Pattern.IsNull().Should().BeFalse();
```
csaba-sagi-sonarsource, line 36: Shouldn`t we also cover the happy path?

analyzers/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
```diff
@@ -40,25 +40,28 @@ public sealed class ConditionalSimplification : SonarDiagnosticAnalyzer
...
+        private static readonly DiagnosticDescriptor RuleMultipleNegation = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageMultipleNegation, RspecStrings.ResourceManager);
+
+        private static readonly ISet<SyntaxKind> EqualsOrNotEquals = new HashSet<SyntaxKind>
         {
             SyntaxKind.EqualsExpression,
             SyntaxKind.NotEqualsExpression
         };
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
```
csaba-sagi-sonarsource, line 54: Shouldn`t this contain Both Rule and RuleMultipleNegation?

analyzers/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
```diff
@@ -72,6 +75,24 @@ internal static bool TryGetExpressionComparedToNull(ExpressionSyntax expression,
                     return true;
                 }
             }
+
+            if (IsPatternExpressionSyntaxWrapper.IsInstance(expression.RemoveParentheses()))
+            {
+                var isPatternWrapper = (IsPatternExpressionSyntaxWrapper)expression.RemoveParentheses();
```
csaba-sagi-sonarsource, line 81: Please add a case like:
Apple a = null;

Apple x;
if (a is (not null)) // Noncompliant {{Use the '??' operator here.}}
{
    x = a;
}
else
{
    x = b;
}

I assume that it will not work until you don`t rebase on master as there is the implementation for RemoveParentheses that removes  parenthesize for ParenthesizedPatternSyntax.

csaba-sagi-sonarsource (changes requested):
Overall looks great, just a few minor suggestions left.

csaba-sagi-sonarsource:
I am also wondering how we should handle the following case:
```
using System;

Apple a = null;
Apple b = null;

Apple x;
if (a is not not null)
{
    x = a;
}
else
{
    x = b;
}

abstract class Fruit { }
class Apple : Fruit { }
```
It currently suggests the ?: operator, yet if it would "reduce the nulls" it would suggest the ?? operator.

analyzers/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
```diff
@@ -89,16 +110,38 @@ protected override void Initialize(SonarAnalysisContext context)
...
         {
-            if (context.Node.GetFirstNonParenthesizedParent() is AssignmentExpressionSyntax assignment && IsCoalesceAssignmentSupported(context))
+            if (context.Node.GetFirstNonParenthesizedParent() is AssignmentExpressionSyntax assignment
+                && context.Compilation.IsCoalesceAssignmentSupported())
             {
                 var left = ((BinaryExpressionSyntax)context.Node).Left.RemoveParentheses();
                 if (CSharpEquivalenceChecker.AreEquivalent(assignment.Left.RemoveParentheses(), left))
                 {
-                    context.ReportDiagnosticWhenActive(Diagnostic.Create(rule, assignment.GetLocation(), BuildCodeFixProperties(context), "??="));
+                    context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, assignment.GetLocation(), BuildCodeFixProperties(context), "??="));
```
csaba-sagi-sonarsource, line 144: Maybe it would be nice to create constants for all the operator strings, not to repeat them so often.

analyzers/src/SonarAnalyzer.CFG/Extensions/IsPatternExpressionSyntaxWrapperExtensions.cs
```diff
@@ -0,0 +1,36 @@
+﻿/*
```
csaba-sagi-sonarsource, line 1: I just realised these extensions are added to SonarAnalyzer.CFG. Shouldn`t they be in the SonarAnalyzer.CSharp/Extensions?

costin-zaharia-sonarsource: I want them to stay together with the LightUp layer. They are useful this layer is used (in this way we can reuse them in security).

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/PatternSyntaxWrapperExtensionsTest.cs
costin-zaharia-sonarsource: Happy path is covered by the test cases but sure, I can add a test also here.

analyzers/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
costin-zaharia-sonarsource: Good point. Small note though: rebase will not solve this issue since we have to call `RemoveParentheses` also when handling inner patterns. 

One more note: we currently have a mess (code duplication) with methods named `RemoveParentheses` and the fix was applied in one place and not the other. We need to address this but I will do it in a separate PR after we split the code in multiple projects.

costin-zaharia-sonarsource:
> It currently suggests the ?: operator, yet if it would "reduce the nulls" it would suggest the ?? operator.

I consider it a corner case, double negation is not something which will happen often. I would leave it like this for now.

csaba-sagi-sonarsource (approved):
LGTM!

PR #4595, opened 2021-06-23
Improve SE performance for a long OR chain condition

### LINKED ISSUE #3395: Improve SE performance for a long OR chain condition
### Description

Some files take forever to be analyzed under specific patterns. It appears to be related to multiple bitwise `OR` operations inside an if condition.

### Repro steps

```csharp
namespace Test
{
    public enum Helper
    {
        A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P
    }
    public static class Test
    {
        public static void TestStuff()
        {
            var helper = Helper.A;
            if (helper == Helper.A | helper == Helper.B | helper == Helper.C | helper == Helper.D |
               helper == Helper.E | helper == Helper.F | helper == Helper.G | helper == Helper.H |
               helper == Helper.I | helper == Helper.A | helper == Helper.B | helper == Helper.C |
               helper == Helper.D | helper == Helper.E | helper == Helper.F | helper == Helper.G |
               helper == Helper.H | helper == Helper.I | helper == Helper.A | helper == Helper.B | 
               helper == Helper.C | helper == Helper.D | helper == Helper.E | helper == Helper.F |
               helper == Helper.G | helper == Helper.H | helper == Helper.I)
            {
                System.Console.WriteLine("");
            }
        }
    }
}

```

**Note:** There's a bit of repetition but the idea is the number of bitwise ORs.
This file is not being analyzed when using these rules: S2259, S3655, S1944, S2583.

### Known workarounds

- Either disabling these patterns
- Removing some of the ORs (it seems that half of these bitwise ORs doesn't fail)
- Assigning the result of the expression to a variable before the if
- Changing the bitwise OR (`\`) by an OR (`||`)

### Related information

* C#
* SonarAnalyzer.CSharp: 8.7.0.17535

pavel-mikula-sonarsource:
Fixes #3395

`OrSymbolicValue.TrySetConstraint` triggered too deep exponential tree evaluation. Same happens for other binary constraints as well.

Changes:
* Introduced `BoolBinarySymbolicValue` to extract common logic for all `BinarySymbolicValue` types that work with `BoolContraint`
* Measure nested size of operands
* Limit size of explored operands based on measurements

SE flow:
* All instructions are processed
* `VisitBinaryBranch` is called at the end of the block
* Constraint is set in `TrySetConstraint` here https://github.com/SonarSource/sonar-dotnet/blob/954aaee95a640e6de52480e5bdacdf67e7b0a687/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs#L870

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/SymbolicValues/Binary/ComparisonSymbolicValue.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/SymbolicValues/Binary/ComparisonSymbolicValue.cs
@@ -25,48 +25,27 @@ using SonarAnalyzer.SymbolicExecution.Relationships;
 
 namespace SonarAnalyzer.SymbolicExecution.SymbolicValues
 {
-    public class ComparisonSymbolicValue : BinarySymbolicValue
+    public class ComparisonSymbolicValue : BoolBinarySymbolicValue
     {
         private readonly ComparisonKind comparisonKind;
 
-        public ComparisonSymbolicValue(ComparisonKind comparisonKind, SymbolicValue leftOperand, SymbolicValue rightOperand)
-            : base(leftOperand, rightOperand)
-        {
+        public ComparisonSymbolicValue(ComparisonKind comparisonKind, SymbolicValue leftOperand, SymbolicValue rightOperand) : base(leftOperand, rightOperand) =>
             this.comparisonKind = comparisonKind;
-        }
 
-        public override IEnumerable<ProgramState> TrySetConstraint(SymbolicValueConstraint constraint, ProgramState programState)
-        {
-            if (!(constraint is BoolConstraint boolConstraint))
-            {
-                return new[] { programState };
-            }
-
-            var relationship = GetRelationship(boolConstraint);
-
-            var newProgramState = programState.TrySetRelationship(relationship);
-            if (newProgramState == null)
-            {
-                return Enumerable.Empty<ProgramState>();
-            }
-
-            return new[] { newProgramState };
-        }
+        protected override IEnumerable<ProgramState> TrySetBoolConstraint(BoolConstraint constraint, ProgramState programState) =>
+            programState.TrySetRelationship(GetRelationship(constraint)) is { } newProgramState
+            ? new[] { newProgramState }
+            : Enumerable.Empty<ProgramState>();
 
         private BinaryRelationship GetRelationship(BoolConstraint boolConstraint)
         {
-            var relationship = new ComparisonRelationship(this.comparisonKind, LeftOperand, RightOperand);
-
-            return boolConstraint == BoolConstraint.True
-                ? relationship
-                : relationship.Negate();
+            var relationship = new ComparisonRelationship(comparisonKind, LeftOperand, RightOperand);
+            return boolConstraint == BoolConstraint.True ? relationship : relationship.Negate();
         }
 
         public override string ToString()
         {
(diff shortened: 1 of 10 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/SymbolicValues/SymbolicValue_TrySetConstraint.cs
```diff
@@ -30,49 +30,6 @@ namespace SonarAnalyzer.SymbolicExecution.SymbolicValues
...
-
-                foreach (var expectedConstraint in expectedConstraints)
-                {
-                    programState.HasConstraint(sv, expectedConstraint).Should().BeTrue(
-                        $"{expectedConstraint} should be present in returned ProgramState.");
-                }
-            }
-        }
-
         public static IEnumerable<object[]> TrueConstraintData { get; } = new[]
```
costin-zaharia-sonarsource, line 33: These methods are used only here. We should make them private

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereference.cs
```diff
@@ -821,3 +821,116 @@ string DoSomethingArg(IEnumerable<object> list)
     }
 }
 
+// https://github.com/SonarSource/sonar-dotnet/issues/3395s
```
costin-zaharia-sonarsource, line 824: wrong link. Remove the `s` at the end.

costin-zaharia-sonarsource (changes requested):
Nice. I've let just a few minor comments.

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/SymbolicValues/Binary/ComparisonSymbolicValue.cs
```diff
@@ -25,48 +25,27 @@
...
+            : Enumerable.Empty<ProgramState>();
 
         private BinaryRelationship GetRelationship(BoolConstraint boolConstraint)
         {
-            var relationship = new ComparisonRelationship(this.comparisonKind, LeftOperand, RightOperand);
-
-            return boolConstraint == BoolConstraint.True
-                ? relationship
-                : relationship.Negate();
+            var relationship = new ComparisonRelationship(comparisonKind, LeftOperand, RightOperand);
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: We could remove the `{}` that I generally don't like, but I'm not convinced it will improve readability in this specific case (`if`), where we aren't forced to embed it into a longer chain of conditions or inside a lambda as perf improvement.

To have it readable, it would look like this that results in the same LOC
```
            new ComparisonRelationship(comparisonKind, LeftOperand, RightOperand) is var relationship
            && boolConstraint == BoolConstraint.True
                ? relationship
                : relationship.Negate();
```

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/SymbolicValues/SymbolicValue_TrySetConstraint.cs
pavel-mikula-sonarsource: We should have a (non SonarWay) rule about this for everyone writing non-API codebase.

costin-zaharia-sonarsource (approved):
LGTM!

PR #4597, opened 2021-06-24
AZP Build: Persist variables between stages

pavel-mikula-sonarsource:
Old: Version is read from repo. That has some consequences around version bump build when:
- PR build starts. Version is read from checked out source as `v8.n.0`, CFG artifacts are build as `v8.n.0.xxxx`
- Version bump is merged to master
- PR build continues (ITs, UTs)
- PR build promotes CFG nuget. Version is read from checked out source as `v8.n+1.0`. CFG promotion fails, because `v8.n+1.0.xxxx` doesn't exist.

New: 
- Version is read from repo and stored on the build agent file system in `/azp-variables/<variable-name>` at the beginning of the build
- `azp-variables` is published as a pipeline artifact to propagate it
- Other stages downloads the artifacts
- Version is read from the artifact file and set as environment variable in each stage

Verification:
- There's custom base branch in #4591
- I pushed squash version of this PR to #4591 to start the build
- I pushed version bump to the base branch after the `Build and Package` started


scripts/build/set-azp-variables.ps1
```diff
@@ -0,0 +1,8 @@
+# Read variables shared between stages and tasks
+$Dir = "$env:AGENT_BUILDDIRECTORY\azp-variables"
+
+$Version = Get-Conent "${Dir}\SONAR_PROJECT_VERSION"
+Write-Host "Sonar project version is '${Version}'"
+
+# Set the azure build variable (SONAR_PROJECT_VERSION) so it can be used by other tasks
+Write-Host "##vso[task.setvariable variable=SONAR_PROJECT_VERSION;]$Version"
```
pavel-mikula-sonarsource, line 8: This weird stuff is taken from the deleted `set-azp-project-verison-variable.ps1`. Git stage didn't handle he rename history nicely.

Docs: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#set-variables-in-scripts

scripts/build/store-azp-variables.ps1
```diff
@@ -0,0 +1,15 @@
+# Read the project version from the Version.props file
+$VersionFilePath = "$env:BUILD_SOURCESDIRECTORY\scripts\version\Version.props"
+Write-Host "Reading the Sonar project version from '${VersionFilePath}' ..."
+
+[xml]$VersionProps = Get-Content "$VersionFilePath"
+$Version = $VersionProps.Project.PropertyGroup.MainVersion
+Write-Host "Sonar project version is '${Version}'"
+
+# Save variables to files so they can be used by other tasks and stages
+$Dir = "$env:AGENT_BUILDDIRECTORY\azp-variables"
```
pavel-mikula-sonarsource, line 10: `AGENT_BUILDDIRECTORY`=`PIPELINE_WORKSPACE`
`BUILD_SOURCESDIRECTORY`=`PIPELINE_WORKSPACE` + `\s`

I'm putting this outside sources directory, to avoid the cleanup of `Post-job: Checkout SonarSource/sonar-dotnet@master to s` job.

Docs: https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml

scripts/build/store-azp-variables.ps1
```diff
@@ -0,0 +1,15 @@
...
+New-Item -ItemType Directory -Path $Dir | Out-Null
+
+$File = "${Dir}\SONAR_PROJECT_VERSION"
+Write-Host "Writing '${Version}' to ${File}"
+Set-Content -Path $File -Value $Version
```
pavel-mikula-sonarsource, line 15: Intention: Each variable to have it's own file for simple reading. These 3 lines would repeat (or be moved in a function in the future)

scripts/build/store-azp-variables.ps1
```diff
@@ -0,0 +1,15 @@
+# Read the project version from the Version.props file
+$VersionFilePath = "$env:BUILD_SOURCESDIRECTORY\scripts\version\Version.props"
+Write-Host "Reading the Sonar project version from '${VersionFilePath}' ..."
+
+[xml]$VersionProps = Get-Content "$VersionFilePath"
+$Version = $VersionProps.Project.PropertyGroup.MainVersion
+Write-Host "Sonar project version is '${Version}'"
```
pavel-mikula-sonarsource, line 7: This section is from the deleted `set-azp-project-version-variable.ps1`

azure-pipelines.yml
```diff
@@ -40,8 +40,13 @@ stages:
...
 
       - script: nuget restore -LockedMode -ConfigFile "analyzers\NuGet.Config" $(solution)
+        displayName: "NuGet Restore"
 
-      - powershell: .\scripts\build\set-azp-project-version-variable.ps1
+      - powershell: .\scripts\build\store-azp-variables.ps1
+        displayName: "Store AZP Variables"
+
+      - powershell: .\scripts\build\set-azp-variables.ps1
+        displayName: "Set AZP Variables"
```
pavel-mikula-sonarsource, line 49: Set `displayName` to avoid `CmdLine`, `PowerShell`, `PowerShell` default job names.

scripts/build/store-azp-variables.ps1
```diff
@@ -0,0 +1,15 @@
...
+
+# Save variables to files so they can be used by other tasks and stages
+$Dir = "$env:AGENT_BUILDDIRECTORY\Azp-Variables"
+New-Item -ItemType Directory -Path $Dir -Force | Out-Null
```
pavel-mikula-sonarsource, line 11: -Force to suppress the error message when it exists on reused agent
-Out-Null to suppress noisy output report

azure-pipelines.yml
```diff
@@ -146,7 +155,12 @@ stages:
       - task: NuGetToolInstaller@1
         displayName: "Install NuGet"
 
-      - powershell: .\scripts\build\set-azp-project-version-variable.ps1
+      - download: current
+        artifact: Azp-Variables
+        displayName: "Download AZP Variables pipeline artifact"
+
+      - powershell: .\scripts\build\set-azp-variables.ps1
+        displayName: "Set AZP Variables"
```
pavel-mikula-sonarsource, line 163: These 6 lines repeats 5 times in this file. Shall I extract it to `azp-variables-steps.yml` template? Usage would be single line step:
```
- template: azp-variables-steps.yml
```

azure-pipelines.yml
andrei-epure-sonarsource: thanks! that was annoying indeed

scripts/build/set-azp-variables.ps1
andrei-epure-sonarsource: did you use `git mv` for the rename?

azure-pipelines.yml
andrei-epure-sonarsource: yes that would be nice

scripts/build/store-azp-variables.ps1
andrei-epure-sonarsource: maybe add these comments to this script, too?
we will rarely touch it and they will help

scripts/build/store-azp-variables.ps1
andrei-epure-sonarsource: why would the lines repeat?

scripts/build/set-azp-variables.ps1
```diff
@@ -0,0 +1,8 @@
+# Read variables shared between stages and tasks
+$Dir = "$env:AGENT_BUILDDIRECTORY\Azp-Variables"
+
+$Version = Get-Content "${Dir}\SONAR_PROJECT_VERSION"
+Write-Host "Sonar project version is '${Version}'"
+
+# Set the azure build variable (SONAR_PROJECT_VERSION) so it can be used by other tasks
```
andrei-epure-sonarsource, line 7:
```suggestion
# Set the azure build variable (SONAR_PROJECT_VERSION) so it can be used by other tasks in the job
```

scripts/build/set-azp-variables.ps1
```diff
@@ -0,0 +1,8 @@
+# Read variables shared between stages and tasks
```
andrei-epure-sonarsource, line 1: between stages and tasks or between stages and jobs?

a stage is made of jobs; a job is made of tasks (steps) - not clear what this comment is about

scripts/build/set-azp-variables.ps1
```diff
@@ -0,0 +1,8 @@
...
+$Dir = "$env:AGENT_BUILDDIRECTORY\Azp-Variables"
+
+$Version = Get-Content "${Dir}\SONAR_PROJECT_VERSION"
```
andrei-epure-sonarsource, line 4: you are using `SONAR_PROJECT_VERSION` both in this storage and in `##vso[task.setvariable variable=SONAR_PROJECT_VERSION;]$Version` which from what I understand are different...

could you prefix the global one with `GLOBAL_SONAR_PROJECT_VERSION` to make it clear which is the one stored globally?

(or maybe I don't understand something and we need to sync offline to better understand)

andrei-epure-sonarsource (changes requested):
RE: the file names - I find confusing the `set-` and `store-` - they have similar meanings, however the `set-` also actually reads the stored values...

so I would rename the `set-` script `load-and-set-azp-variables` - it maybe explicit but I think that's fine for the build files

scripts/build/set-azp-variables.ps1
pavel-mikula-sonarsource: No, I stage changes before commit and that pairs the deleted/added as rename. There's no much harm for this file.

scripts/build/store-azp-variables.ps1
pavel-mikula-sonarsource: Only in case we add more variables in the future. Lines would repeat for 2nd variable.

scripts/build/set-azp-variables.ps1
pavel-mikula-sonarsource: Environment variable `SONAR_PROJECT_VERSION` (line 8) is holding the value of sonar project version for current job.
This value is persisted in `Azp-Variables\SONAR_PROJECT_VERSION` file between stages and jobs.

It's important to have exactly the same name for both.

pavel-mikula-sonarsource:
I don't want to rename the `set-azp-variables.ps` as `SET` is the operation to configure environment variable. The source of the value is implementation detail. And it had the `set-` prefix before as well.
```
SET PATH=C:\Break\Everything\Else
```

I agree that `set-` and `store-` are quite close to each other. Will renaming `store-` to something else solve it? At the same time, I'd prefer simplicity here, we don't read those names very often and we can take a look what's (well-documented) inside when we need to.

azure-pipelines.yml
```diff
@@ -146,7 +154,7 @@ stages:
       - task: NuGetToolInstaller@1
         displayName: "Install NuGet"
 
-      - powershell: .\scripts\build\set-azp-project-version-variable.ps1
+      - template: set-azp-variables-steps.yml
```
pavel-mikula-sonarsource, line 157: YML template file is relative to the outer template from @commonTempaltes. 
Having file in our repo produced error:
>/stage-with-burgr-notifications.yml@commonTemplates: File /azp-variables-steps.yml not found in repository https://dev.azure.com/sonarsource/DotNetTeam%20Project/_git/pipelines-yaml-templates branch refs/tags/v1.0.11 version 0b8ce2a631db4d551d8194bc472d2887069ba0a9.

The steps template is now here and redesigned to be reusable.
https://dev.azure.com/sonarsource/DotNetTeam%20Project/_git/pipelines-yaml-templates?path=%2Fset-azp-variables-steps.yml

azure-pipelines.yml
```diff
@@ -21,7 +21,7 @@ resources:
     - repository: commonTemplates
       type: git
       name: pipelines-yaml-templates
-      ref:  refs/tags/v1.0.11
+      #FIXME: ref:  refs/tags/v1.0.11
```
pavel-mikula-sonarsource, line 24: This PR is now draft. 
Please take a look at https://dev.azure.com/sonarsource/DotNetTeam%20Project/_git/pipelines-yaml-templates?path=%2Fset-azp-variables-steps.yml
If it's OK, I'll add tag there and reference it here before merging.

azure-pipelines.yml
```diff
@@ -40,8 +40,16 @@ stages:
...
 
-      - powershell: .\scripts\build\set-azp-project-version-variable.ps1
+      - powershell: .\scripts\build\store-azp-variables.ps1
+        displayName: "Store AZP Variables"
+
+      - publish: $(Agent.BuildDirectory)/Azp-Variables
+        artifact: Azp-Variables
+        displayName: "Publish AZP Variables as pipeline artifact"
+
+      - template: set-azp-variables-steps.yml@commonTemplates
```
pavel-mikula-sonarsource, line 52: This downloads the artifact that was just published on the previous line, but it takes only 1s so I'm keeping it as is for simplicity/consistency.

scripts/build/set-azp-variables.ps1
andrei-epure-sonarsource: ok, I didn't know `Azp-Variables\SONAR_PROJECT_VERSION` was a file

azure-pipelines.yml
andrei-epure-sonarsource: it looks fine. Once you add the tag, please also send an email to the SonarLint team as they are using AZP for all SL flavors

scripts/build/store-azp-variables.ps1
```diff
@@ -0,0 +1,17 @@
...
+$VersionFilePath = "$env:BUILD_SOURCESDIRECTORY\scripts\version\Version.props"
+Write-Host "Reading the Sonar project version from '${VersionFilePath}' ..."
+
+[xml]$VersionProps = Get-Content "$VersionFilePath"
+$Version = $VersionProps.Project.PropertyGroup.MainVersion
+Write-Host "Sonar project version is '${Version}'"
+
+# Save variables to files so they can be used by other tasks and stages
+# -Force to suppress the error message when it exists on reused agent
+# | Out-Null to suppress noisy output report
```
andrei-epure-sonarsource, line 11:
```suggestion
# - Out-Null to suppress noisy output report
```

andrei-epure-sonarsource (approved):
LGTM

scripts/build/store-azp-variables.ps1
pavel-mikula-sonarsource: This was intentional as a code fragment for output redirection, not a bullet point. Do you want to change it?

azure-pipelines.yml
pavel-mikula-sonarsource: FIXME is fixed, I'll ping SL once thisone is merged.

PR #4599, opened 2021-06-24
Delete S4142 (deprecated)

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.Utilities/Rules.Description/S1145_cs.html
```diff
@@ -1,36 +0,0 @@
-<p><code>if</code> statements with conditions that
```
csaba-sagi-sonarsource, line 1: Why is this one deleted? This PR should only delete S4142 related stuff, or am I missing something?

analyzers/src/SonarAnalyzer.Utilities/Rules.Description/S1145_vb.html
```diff
@@ -1,32 +0,0 @@
-<p><code>if</code> statements with conditions that
```
csaba-sagi-sonarsource, line 1: Same question for this file.

csaba-sagi-sonarsource (changes requested):
Seems to me that 2 files were deleted unintentionally.

analyzers/src/SonarAnalyzer.Utilities/Rules.Description/S1145_cs.html
costin-zaharia-sonarsource: I've missed these files while deleting S4142 and found them now.

analyzers/src/SonarAnalyzer.Utilities/Rules.Description/S1145_vb.html
costin-zaharia-sonarsource: See up

csaba-sagi-sonarsource (approved):
LGTM!

Issue #4537, opened 2021-06-10
S2259 FP: Null conditional combined with null coalescing


przemyslaw-piescik:
### Description

S2259 is incorrectly reported when using null conditional with null coalescing operator.

### Repro steps

```cs
        private void Sample()
        {
            string someString = null;

            if (!someString?.Contains("a") ?? true)
                Console.WriteLine("It's null or doesn't contain 'a'");
            else
                Console.WriteLine(someString.Length); //S2259 reported here
        }
```

### Expected behavior

S2259 should not be reported in this case.

### Actual behavior

S2259 is reported.

pavel-mikula-sonarsource:
Hi @przemyslaw-piescik ,

Thank you for reporting this case, I can confirm it as False Positive. We already have similar issue reported in #2528 with the same root cause that you can follow. I'll close this one.

pavel-mikula-sonarsource:
I'll actually keep this one opened, as it's not isolated from the integer constraint issue reported in the other one.

PR #4600, opened 2021-06-24
S2259: Add FP repro for #4537

pavel-mikula-sonarsource:
Reproduces #4537

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereference.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereference.cs
@@ -819,5 +819,19 @@ namespace Tests.Diagnostics
             return item.ToString(); // False Negative, item could be null here
         }
     }
+
+    // https://github.com/SonarSource/sonar-dotnet/issues/4537
+    class Repro_4537
+    {
+        private void ConditionalAccess_NullCoalescing()
+        {
+            string someString = null;
+
+            if (!someString?.Contains("a") ?? true)
+                Console.WriteLine("It's null or doesn't contain 'a'");
+            else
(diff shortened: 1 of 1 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereference.cs
```diff
@@ -819,5 +819,19 @@ string DoSomethingArg(IEnumerable<object> list)
...
+    class Repro_4537
+    {
+        private void ConditionalAccess_NullCoalescing()
+        {
+            string someString = null;
+
+            if (!someString?.Contains("a") ?? true)
+                Console.WriteLine("It's null or doesn't contain 'a'");
+            else
+                Console.WriteLine(someString.Length); // Noncompliant FP, this path is unreachable
```
andrei-epure-sonarsource, line 833: why is it unreachable? If it's not null and it does contain "a", then it will reach here

andrei-epure-sonarsource (approved):
please update the comment before merging.
lgtm otherwise

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereference.cs
pavel-mikula-sonarsource: Because `string someString = null;` and we have `NULL` constraint on that variable.

pavel-mikula-sonarsource: So SE should not reach that path and should not raise S2259.
In real code base, it should raise S2583/S2589 instead.

andrei-epure-sonarsource: ah sorry I've missed the initialization at the top

Issue #4515, opened 2021-06-04
S2583 FN: Local functions


hankovich:
### Description

```csharp
public void A()
{
    int? i = GetNullableInt();

    if (i == null && i == 3) // <- S2583 is reported here (i == 3 is always false)
    {

    }

    static void B()
    {
        int? i = GetNullableInt();

        if (i == null && i == 3) // <- but... S2583 is NOT reported here
        {

        }
    }

    static int? GetNullableInt() => 42;
}
```

### Related information

* C#/VB.NET Plugins version: 8.23.0.32424

csaba-sagi-sonarsource:
Hi @hankovich , I can confirm the false negative. Thanks for your feedback.

mary-georgiou-sonarsource:
Fixed by https://github.com/SonarSource/sonar-dotnet/pull/7478

PR #4602, opened 2021-06-25
FN reproducer with local functions S2583

csaba-sagi-sonarsource:
Reproduces #4515

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs
@@ -298,4 +298,30 @@ namespace Tests.Diagnostics
             }
         }
     }
+
+    // https://github.com/SonarSource/sonar-dotnet/issues/4515
+    public class Repro_4515
+    {
+        public void DoWork()
+        {
+            int? i = GetNullableInt();
+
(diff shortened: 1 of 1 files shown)


costin-zaharia-sonarsource:
We have already a few test cases which document that local functions are not inspected: 
See: 
https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs#L133
and 
https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs#L140.

csaba-sagi-sonarsource:
> We have already a few test cases which document that local functions are not inspected:
> See:
> https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs#L133
> and
> https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.CSharp8.cs#L140.

Indeed, yet that says StaticLocalFunctions. Yet this example I added does not work even without the static keyword.

costin-zaharia-sonarsource (approved):
LGTM!

PR #4609, opened 2021-06-29
Rule S3060: add support for C# 9 pattern matching

csaba-sagi-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
@@ -18,6 +18,8 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
@@ -36,7 +38,7 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class DoNotTestThisWithIsOperator : SonarDiagnosticAnalyzer
     {
         private const string DiagnosticId = "S3060";
-        private const string MessageFormat = "Offload the code that's conditional on this 'is' test to the appropriate subclass and remove the test.";
+        private const string MessageFormat = "Offload the code that's conditional on this type test to the appropriate subclass and remove the condition.";
 
         private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
@@ -54,47 +56,76 @@ namespace SonarAnalyzer.Rules.CSharp
         {
             if (((BinaryExpressionSyntax)context.Node).Left.RemoveParentheses() is ThisExpressionSyntax)
             {
-                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.Node.GetLocation()));
+                ReportDiagnostic(context, context.Node);
             }
         }
 
         private static void AnalyzeIsPatternExpression(SyntaxNodeAnalysisContext context)
         {
-            var isPatternExpression = (IsPatternExpressionSyntaxWrapper)context.Node;
-            if (isPatternExpression.Expression.RemoveParentheses() is ThisExpressionSyntax
-                && ContainsTypeCheckInPattern(isPatternExpression))
+            if (((IsPatternExpressionSyntaxWrapper)context.Node).Expression.RemoveParentheses() is ThisExpressionSyntax
+                && ContainsTypeCheckInPattern(context.Node))
             {
-                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.Node.GetLocation()));
+                ReportDiagnostic(context, context.Node);
             }
         }
 
         private static void AnalyzeSwitchStatement(SyntaxNodeAnalysisContext context)
         {
             var switchStatement = (SwitchStatementSyntax)context.Node;
-            if (switchStatement.Expression.RemoveParentheses() is ThisExpressionSyntax
-                && (ContainsTypeCheckInPattern(switchStatement)
-                    || ContainsTypeCheckInCaswSwitchLabel(switchStatement)))
+            var secondaryLocations = new List<SecondaryLocation>();
+            if (switchStatement.Expression.RemoveParentheses() is ThisExpressionSyntax)
             {
-                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.Node.GetLocation()));
+                foreach (var section in switchStatement.Sections)
+                {
+                    foreach (var label in section.Labels)
+                    {
+                        if (ContainsTypeCheckInPattern(label)
+                            || ContainsTypeCheckInCaswSwitchLabel(label))
+                        {
+                            secondaryLocations.Add(new SecondaryLocation(GetTypeMatchLocation(label), string.Empty));
+                        }
+                    }
+                }
+
+                if (secondaryLocations.Any())
+                {
+                    context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule,
+                                                                         switchStatement.Expression.GetLocation(),
+                                                                         additionalLocations: secondaryLocations.ToAdditionalLocations(),
+                                                                         properties: secondaryLocations.ToProperties()));
+                }
             }
         }
 
         private static void AnalyzeSwitchExpression(SyntaxNodeAnalysisContext context)
         {
             var switchExpression = (SwitchExpressionSyntaxWrapper)context.Node;
-            if (switchExpression.GoverningExpression.RemoveParentheses() is ThisExpressionSyntax
-                && ContainsTypeCheckInPattern(switchExpression))
+            var secondaryLocations = new List<SecondaryLocation>();
+            if (switchExpression.GoverningExpression.RemoveParentheses() is ThisExpressionSyntax)
             {
-                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.Node.GetLocation()));
+                foreach (var arm in switchExpression.Arms)
+                {
+                    if (ContainsTypeCheckInPattern(arm.Pattern.SyntaxNode))
+                    {
+                        secondaryLocations.Add(new SecondaryLocation(arm.Pattern.SyntaxNode.GetLocation(), string.Empty));
+                    }
+                }
+
+                if (secondaryLocations.Any())
+                {
+                    context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule,
+                                                                         switchExpression.GoverningExpression.GetLocation(),
+                                                                         additionalLocations: secondaryLocations.ToAdditionalLocations(),
+                                                                         properties: secondaryLocations.ToProperties()));
+                }
             }
         }
 
-        private static bool ContainsTypeCheckInCaswSwitchLabel(SwitchStatementSyntax switchStatement) =>
-            switchStatement.Sections.Any(x => x.Labels.Any(x => x is CaseSwitchLabelSyntax caseSwitchLabel
-                                                                && caseSwitchLabel.Value.IsKind(SyntaxKind.IdentifierName)));
+        private static bool ContainsTypeCheckInCaswSwitchLabel(SwitchLabelSyntax switchLabel) =>
+              switchLabel is CaseSwitchLabelSyntax caseSwitchLabel && caseSwitchLabel.Value.IsKind(SyntaxKind.IdentifierName);
 
         private static bool ContainsTypeCheckInPattern(SyntaxNode syntaxNode) =>
-            syntaxNode.DescendantNodes()
+            syntaxNode.DescendantNodesAndSelf()
                       .Where(x => x.IsAnyKind(SyntaxKindEx.ConstantPattern, SyntaxKindEx.DeclarationPattern, SyntaxKindEx.RecursivePattern))
                       .Any(x => IsTypeCheckOnThis(x));
 
@@ -125,5 +156,22 @@ namespace SonarAnalyzer.Rules.CSharp
                                                                   SyntaxKind.SwitchStatement,
                                                                   SyntaxKindEx.Subpattern)) is var containingNode
             && !containingNode.IsKind(SyntaxKindEx.Subpattern);
+
+        private static void ReportDiagnostic(SyntaxNodeAnalysisContext context, SyntaxNode node) =>
+            context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, node.GetLocation()));
+
+        private static Location GetTypeMatchLocation(SwitchLabelSyntax label)
+        {
+            if (label is CaseSwitchLabelSyntax caseSwitchLabel)
+            {
+                return caseSwitchLabel.Value.GetLocation();
+            }
+            else if (CasePatternSwitchLabelSyntaxWrapper.IsInstance(label))
+            {
+                return ((CasePatternSwitchLabelSyntaxWrapper)label).Pattern.SyntaxNode.GetLocation();
+            }
+
+            return Location.None;
+        }
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
@@ -18,13 +18,16 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Collections.Immutable;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
+using SonarAnalyzer.Extensions;
 using SonarAnalyzer.Helpers;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
@@ -39,15 +42,20 @@ namespace SonarAnalyzer.Rules.CSharp
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
-        protected override void Initialize(SonarAnalysisContext context) =>
-            context.RegisterSyntaxNodeActionInNonGenerated(c =>
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(c => Analyze(c, x => ((BinaryExpressionSyntax)x).Left), SyntaxKind.IsExpression);
+            context.RegisterSyntaxNodeActionInNonGenerated(c => Analyze(c, x => ((SwitchStatementSyntax)x).Expression), SyntaxKind.SwitchStatement);
+            context.RegisterSyntaxNodeActionInNonGenerated(c => Analyze(c, x => ((IsPatternExpressionSyntaxWrapper)x).Expression), SyntaxKindEx.IsPatternExpression);
+            context.RegisterSyntaxNodeActionInNonGenerated(c => Analyze(c, x => ((SwitchExpressionSyntaxWrapper)x).GoverningExpression), SyntaxKindEx.SwitchExpression);
+        }
+
+        private static void Analyze(SyntaxNodeAnalysisContext context, Func<SyntaxNode, ExpressionSyntax> retrieveExpression)
+        {
+            if (retrieveExpression(context.Node).RemoveParentheses() is ThisExpressionSyntax)
             {
-                var expression = (BinaryExpressionSyntax)c.Node;
-                if (expression.Left is ThisExpressionSyntax)
-                {
-                    c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, expression.GetLocation()));
-                }
-            },
-            SyntaxKind.IsExpression);
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.Node.GetLocation()));
+            }
+        }
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
@@ -32,25 +32,22 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class DoNotTestThisWithIsOperator : SonarDiagnosticAnalyzer
     {
-        internal const string DiagnosticId = "S3060";
+        private const string DiagnosticId = "S3060";
         private const string MessageFormat = "Offload the code that's conditional on this 'is' test to the appropriate subclass and remove the test.";
 
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
-        protected override void Initialize(SonarAnalysisContext context)
-        {
+        protected override void Initialize(SonarAnalysisContext context) =>
             context.RegisterSyntaxNodeActionInNonGenerated(c =>
+            {
+                var expression = (BinaryExpressionSyntax)c.Node;
+                if (expression.Left is ThisExpressionSyntax)
                 {
-                    var expression = (BinaryExpressionSyntax)c.Node;
-                    if (expression.Left is ThisExpressionSyntax)
-                    {
-                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, expression.GetLocation()));
-                    }
-                },
-                SyntaxKind.IsExpression);
-        }
+                    c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, expression.GetLocation()));
+                }
+            },
+            SyntaxKind.IsExpression);
     }
 }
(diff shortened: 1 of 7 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotTestThisWithIsOperator.CSharp9.cs
```diff
@@ -5,13 +5,33 @@ namespace Tests.Diagnostics
...
+            if (this is Program p) // Noncompliant
             {
             }
 
-            switch (this) // FN
+            if (this is var variable) // Noncompliant
+            {
+            }
+
+            if (this is { MyProperty: 5 }) // Noncompliant
```
pavel-mikula-sonarsource, line 30: I don't think this is noncompliant. This one doesn't test the type of `this` but uses syntax sugar to test property values.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotTestThisWithIsOperator.CSharp9.cs
```diff
@@ -5,13 +5,33 @@ namespace Tests.Diagnostics
...
+            {
+            }
+
+            if (o is not IDisposable)
+            {
+            }
+
+            if (this is Program p) // Noncompliant
             {
             }
```
pavel-mikula-sonarsource, line 24: Let's add test case with `OrPattern`

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -18,39 +18,44 @@
...
-                    }
-                },
-                SyntaxKind.IsExpression);
+            context.RegisterSyntaxNodeActionInNonGenerated(c => Analyze(c, x => ((BinaryExpressionSyntax)x).Left), SyntaxKind.IsExpression);
+            context.RegisterSyntaxNodeActionInNonGenerated(c => Analyze(c, x => ((SwitchStatementSyntax)x).Expression), SyntaxKind.SwitchStatement);
+            context.RegisterSyntaxNodeActionInNonGenerated(c => Analyze(c, x => ((IsPatternExpressionSyntaxWrapper)x).Expression), SyntaxKindEx.IsPatternExpression);
+            context.RegisterSyntaxNodeActionInNonGenerated(c => Analyze(c, x => ((SwitchExpressionSyntaxWrapper)x).GoverningExpression), SyntaxKindEx.SwitchExpression);
+        }
+
+        private static void Analyze(SyntaxNodeAnalysisContext context, Func<SyntaxNode, ExpressionSyntax> retrieveExpression)
```
pavel-mikula-sonarsource, line 53: Could be just `expression`. It's clear that it will return something when it's `Func`.
```suggestion
        private static void Analyze(SyntaxNodeAnalysisContext context, Func<SyntaxNode, ExpressionSyntax> expression)
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotTestThisWithIsOperator.CSharp9.cs
```diff
@@ -5,13 +5,33 @@ namespace Tests.Diagnostics
 {
     class Program
     {
+        public int MyProperty { get; set; }
+
+        public int MyProperty2 { get; set; }
```
pavel-mikula-sonarsource, line 10: Can be removed?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotTestThisWithIsOperator.CSharp9.cs
```diff
@@ -21,7 +41,24 @@ public Program(object o)
...
+            {
+                case IDisposable:
+                    break;
+                case IEnumerable:
+                    break;
+                default:
+                    break;
+            }
+
+            var result = (this) switch // Noncompliant
```
pavel-mikula-sonarsource, line 54: `(this)` is here to test the parenthesis removal? If yes, let's be explicit about it, because it looks like a copy-paste issue from switch statement on first look and inconsistent with the next `o switch` case.
```suggestion
            var result = (((this))) switch // Noncompliant
```

pavel-mikula-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotTestThisWithIsOperator.CSharp9.cs
```diff
@@ -5,13 +5,113 @@ namespace Tests.Diagnostics
...
+            if (this is not null) // Compliant as it does not check the type.
+            {
+
+            }
+
+            if (this is Program or IDisposable) // Noncompliant
+            {
+            }
+
+            if (this is Program and IDisposable) // Noncompliant
```
pavel-mikula-sonarsource, line 39: Let's add a test case
```
if (x is not null and (IDisposable or Program))
```

pavel-mikula-sonarsource (changes requested):
Just first part about the UTs for now. I'll take a look at the implementation later

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -19,38 +19,111 @@
...
+            {
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.Node.GetLocation()));
+            }
+        }
+
+        private static void AnalyzeIsPatternExpression(SyntaxNodeAnalysisContext context)
+        {
+            var isPatternExpression = (IsPatternExpressionSyntaxWrapper)context.Node;
+            if (isPatternExpression.Expression.RemoveParentheses() is ThisExpressionSyntax
+                && ContainsTypeCheckInPattern(isPatternExpression))
```
pavel-mikula-sonarsource, line 65: Can be simplified
```suggestion
            if (((IsPatternExpressionSyntaxWrapper)context.Node).Expression.RemoveParentheses() is ThisExpressionSyntax
                && ContainsTypeCheckInPattern(context.Node))
```

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -19,38 +19,111 @@
...
+            if (((BinaryExpressionSyntax)context.Node).Left.RemoveParentheses() is ThisExpressionSyntax)
+            {
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.Node.GetLocation()));
+            }
+        }
+
+        private static void AnalyzeIsPatternExpression(SyntaxNodeAnalysisContext context)
+        {
+            var isPatternExpression = (IsPatternExpressionSyntaxWrapper)context.Node;
+            if (isPatternExpression.Expression.RemoveParentheses() is ThisExpressionSyntax
```
pavel-mikula-sonarsource, line 64: Condition could be single line

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -19,38 +19,111 @@
...
+            context.RegisterSyntaxNodeActionInNonGenerated(AnalyzeIsPatternExpression, SyntaxKindEx.IsPatternExpression);
+            context.RegisterSyntaxNodeActionInNonGenerated(AnalyzeSwitchExpression, SyntaxKindEx.SwitchExpression);
+            context.RegisterSyntaxNodeActionInNonGenerated(AnalyzeSwitchStatement, SyntaxKind.SwitchStatement);
         }
+
+        private static void AnalyzeIsExpression(SyntaxNodeAnalysisContext context)
+        {
+            if (((BinaryExpressionSyntax)context.Node).Left.RemoveParentheses() is ThisExpressionSyntax)
+            {
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.Node.GetLocation()));
```
pavel-mikula-sonarsource, line 57: This line repeats 4 times in exactly the same (complicated) form. Let's extract it as `ReportDiagnostics(context)` that will do the trick.

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -19,38 +19,111 @@
...
+            }
+        }
+
+        private static void AnalyzeSwitchStatement(SyntaxNodeAnalysisContext context)
+        {
+            var switchStatement = (SwitchStatementSyntax)context.Node;
+            if (switchStatement.Expression.RemoveParentheses() is ThisExpressionSyntax
+                && (ContainsTypeCheckInPattern(switchStatement)
+                    || ContainsTypeCheckInCaswSwitchLabel(switchStatement)))
```
pavel-mikula-sonarsource, line 76: Let's make this single line, the EOL is hard to read there
```suggestion
                && (ContainsTypeCheckInPattern(switchStatement) || ContainsTypeCheckInCaswSwitchLabel(switchStatement)))
```

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
pavel-mikula-sonarsource: The `.RemoveParentheses() is ThisExpressionSyntax` is repetitive pattern here, let's extract it to `IsThis(SyntaxNode node)` to remove the repetitions. It will also make conditions shorter and easier to read on a single line

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -19,38 +19,111 @@
...
+        {
+            var switchExpression = (SwitchExpressionSyntaxWrapper)context.Node;
+            if (switchExpression.GoverningExpression.RemoveParentheses() is ThisExpressionSyntax
+                && ContainsTypeCheckInPattern(switchExpression))
+            {
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.Node.GetLocation()));
+            }
+        }
+
+        private static bool ContainsTypeCheckInCaswSwitchLabel(SwitchStatementSyntax switchStatement) =>
```
pavel-mikula-sonarsource, line 92: Typo
```suggestion
        private static bool ContainsTypeCheckInCaseSwitchLabel(SwitchStatementSyntax switchStatement) =>
```

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -19,38 +19,111 @@
...
+            {
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.Node.GetLocation()));
+            }
+        }
+
+        private static void AnalyzeSwitchExpression(SyntaxNodeAnalysisContext context)
+        {
+            var switchExpression = (SwitchExpressionSyntaxWrapper)context.Node;
+            if (switchExpression.GoverningExpression.RemoveParentheses() is ThisExpressionSyntax
+                && ContainsTypeCheckInPattern(switchExpression))
```
pavel-mikula-sonarsource, line 86: Can be simplified
```suggestion
            if (((SwitchExpressionSyntaxWrapper)context.Node)).GoverningExpression.RemoveParentheses() is ThisExpressionSyntax && ContainsTypeCheckInPattern(context.Node))
```

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -19,38 +19,111 @@
...
+            {
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.Node.GetLocation()));
+            }
+        }
+
+        private static bool ContainsTypeCheckInCaswSwitchLabel(SwitchStatementSyntax switchStatement) =>
+            switchStatement.Sections.Any(x => x.Labels.Any(x => x is CaseSwitchLabelSyntax caseSwitchLabel
+                                                                && caseSwitchLabel.Value.IsKind(SyntaxKind.IdentifierName)));
```
pavel-mikula-sonarsource, line 94: Let's make this a single line. The EOL is making it harder to read

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -19,38 +19,111 @@
...
+            var switchExpression = (SwitchExpressionSyntaxWrapper)context.Node;
+            if (switchExpression.GoverningExpression.RemoveParentheses() is ThisExpressionSyntax
+                && ContainsTypeCheckInPattern(switchExpression))
+            {
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.Node.GetLocation()));
+            }
+        }
+
+        private static bool ContainsTypeCheckInCaswSwitchLabel(SwitchStatementSyntax switchStatement) =>
+            switchStatement.Sections.Any(x => x.Labels.Any(x => x is CaseSwitchLabelSyntax caseSwitchLabel
```
pavel-mikula-sonarsource, line 93: The nested lambda should not use `x`, because the outer lambda does. I'm surprised it compiles, because they share the scope :/ 
```suggestion
            switchStatement.Sections.Any(section => section.Labels.Any(x => x is CaseSwitchLabelSyntax caseSwitchLabel
```

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -19,38 +19,111 @@
...
+                                                                && caseSwitchLabel.Value.IsKind(SyntaxKind.IdentifierName)));
+
+        private static bool ContainsTypeCheckInPattern(SyntaxNode syntaxNode) =>
+            syntaxNode.DescendantNodes()
+                      .Where(x => x.IsAnyKind(SyntaxKindEx.ConstantPattern, SyntaxKindEx.DeclarationPattern, SyntaxKindEx.RecursivePattern))
+                      .Any(x => IsTypeCheckOnThis(x));
```
pavel-mikula-sonarsource, line 99: Can be simplified
```suggestion
            syntaxNode.DescendantNodes().Any(x => x.IsAnyKind(SyntaxKindEx.ConstantPattern, SyntaxKindEx.DeclarationPattern, SyntaxKindEx.RecursivePattern) && IsTypeCheckOnThis(x));
```

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -19,38 +19,111 @@
...
+            {
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.Node.GetLocation()));
+            }
+        }
+
+        private static bool ContainsTypeCheckInCaswSwitchLabel(SwitchStatementSyntax switchStatement) =>
+            switchStatement.Sections.Any(x => x.Labels.Any(x => x is CaseSwitchLabelSyntax caseSwitchLabel
+                                                                && caseSwitchLabel.Value.IsKind(SyntaxKind.IdentifierName)));
+
+        private static bool ContainsTypeCheckInPattern(SyntaxNode syntaxNode) =>
```
pavel-mikula-sonarsource, line 96: This is called on the entire switch node. If there would be nested type pattern check inside the returned value, this looks like it would raise a FP. Let's add test case.

I think the `DescendantNodes` should have stopping condition to avoid nesting inside the returned value/block/section.

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -19,38 +19,111 @@
...
+                      .Where(x => x.IsAnyKind(SyntaxKindEx.ConstantPattern, SyntaxKindEx.DeclarationPattern, SyntaxKindEx.RecursivePattern))
+                      .Any(x => IsTypeCheckOnThis(x));
+
+        private static bool IsTypeCheckOnThis(SyntaxNode pattern)
+        {
+            if (ConstantPatternSyntaxWrapper.IsInstance(pattern)
+                && (ConstantPatternSyntaxWrapper)pattern is var constantPattern)
+            {
+                return constantPattern.Expression.IsKind(SyntaxKind.IdentifierName)
+                       && IsNotInSubPattern(pattern);
```
pavel-mikula-sonarsource, line 107: This is overusage of `var` pattern, becase we doesn't have to do it this way - it's not needed by other condition in the `&&` chain. Let's simplify it to avoid it and reduce empty lines at the same time.

Same for conditions below.
```suggestion
if (ConstantPatternSyntaxWrapper.IsInstance(pattern))
            {
                return ((ConstantPatternSyntaxWrapper)pattern).Expression.IsKind(SyntaxKind.IdentifierName) && IsNotInSubPattern(pattern);
```

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -19,38 +19,111 @@
...
+            }
+
+            return false;
+        }
+
+        private static bool IsNotInSubPattern(SyntaxNode node) =>
+            node.FirstAncestorOrSelf<SyntaxNode>(x => x.IsAnyKind(SyntaxKindEx.IsPatternExpression,
+                                                                  SyntaxKindEx.SwitchExpression,
+                                                                  SyntaxKind.SwitchStatement,
+                                                                  SyntaxKindEx.Subpattern)) is var containingNode
```
pavel-mikula-sonarsource, line 126: It is puzzling why we need to continue when it returns `null`
```suggestion
                                                                  SyntaxKindEx.Subpattern)) is { } containingNode
```

pavel-mikula-sonarsource: I'm not sure if `IsKind` tolerates null. If so (and I hope so), why do the indirection with variable? Let's just return directly `FirstAncestorOrSelf<>(...).IsKind(...)`

pavel-mikula-sonarsource (changes requested):
Implementation part

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
csaba-sagi-sonarsource: After the last changes I think it is not applicable.

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
csaba-sagi-sonarsource: After the last changes I think it is not applicable.

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
csaba-sagi-sonarsource: After the latest changes it is removed.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotTestThisWithIsOperator.CSharp9.cs
```diff
@@ -5,13 +5,113 @@ namespace Tests.Diagnostics
...
+
+            if (this is ({ MyProperty: 5 })) // Compliant as it does not check the type
+            {
+            }
+
+            if (this is not { MyProperty: 5 })
+            {
+            }
+
+            if (o is null)
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: We don't have the simple complicat test case now. We only have more complicated scenarios with `not`. Let's add
```
if (o is Program) {}
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotTestThisWithIsOperator.CSharp9.cs
```diff
@@ -5,13 +5,113 @@ namespace Tests.Diagnostics
 {
     class Program
     {
+        public int MyProperty { get; set; }
+        public object MyProperty2 { get; set; }
+
         public Program(object o)
         {
-            if (this is not IDisposable) // FN
+            if (this is Program and { MyProperty: 2 }) // Noncompliant
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Can you highlight only the `Program` part of it? Becase that's what's raising the issue. Preserving `{ MyProperty: 2 }` is legal and most likely desired.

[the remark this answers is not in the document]
csaba-sagi-sonarsource: It is possible, yet I am not sure it is worth the trouble. The message clearly says to remove the type check. Plus currently the rule stops at the first non compliant SubPattern, and if we want to underline only the type that would me we should not stope at the first non compliant SubPattern but retrieve all and create  SecondaryLocations for all of them.

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -18,39 +18,172 @@
...
+            if (IsThisExpressionSyntax(((BinaryExpressionSyntax)context.Node).Left))
+            {
+                ReportDiagnostic(context, context.Node);
+            }
+        }
+
+        private static void AnalyzeIsPatternExpression(SyntaxNodeAnalysisContext context)
+        {
+            if (IsThisExpressionSyntax(((IsPatternExpressionSyntaxWrapper)context.Node).Expression)
+                && ContainsTypeCheckInPattern(context.Node))
```
pavel-mikula-sonarsource, line 66: Could be single line

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -18,39 +18,172 @@
...
+            }
+        }
+
+        private static void AnalyzeSwitchStatement(SyntaxNodeAnalysisContext context)
+        {
+            var switchStatement = (SwitchStatementSyntax)context.Node;
+            if (IsThisExpressionSyntax(switchStatement.Expression))
+            {
+                var secondaryLocations = CollectSecondaryLocations(switchStatement);
+                if (secondaryLocations.Any())
```
pavel-mikula-sonarsource, line 78: We can move this condition inside the reporting method. Like that, we can do the `CollectSecondaryLocation` without the extra variable and pass it directly in.
It would also remove the logic duplication from the other method below.

pavel-mikula-sonarsource (changes requested):
Pack of nitpicks and improvement of highlighting suggested.

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -18,39 +18,172 @@
...
+                                                 .IsKind(SyntaxKindEx.Subpattern);
+
+        private static void ReportDiagnostic(SyntaxNodeAnalysisContext context, SyntaxNode node) =>
+            context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, node.GetLocation()));
+
+        private static void ReportDiagnosticWithSecondaryLocation(SyntaxNodeAnalysisContext context, SyntaxNode node, IList<SecondaryLocation> secondaryLocations) =>
+            context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule,
+                                                                 node.GetLocation(),
+                                                                 additionalLocations: secondaryLocations.ToAdditionalLocations(),
+                                                                 properties: secondaryLocations.ToProperties()));
```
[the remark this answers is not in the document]
csaba-sagi-sonarsource: It sets the Index and the message of the secondary location.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotTestThisWithIsOperator.CSharp9.cs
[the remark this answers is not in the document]
andrei-epure-sonarsource: did you open a ticket in the end for the secondary locations improvement?

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
andrei-epure-sonarsource: Did you take any action here about the stopping condition, @csaba-sagi-sonarsource ? I see you didn't reply to this comment and didn't change the code, either?

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
```diff
@@ -18,39 +18,172 @@
...
+                    if (ContainsTypeCheckInPattern(label)
+                        || ContainsTypeCheckInCaseSwitchLabel(label))
                     {
-                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, expression.GetLocation()));
+                        secondaryLocations.Add(new SecondaryLocation(GetTypeMatchLocation(label), string.Empty));
                     }
-                },
-                SyntaxKind.IsExpression);
+                }
+            }
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: for these sort of refactoring to use LINQ I suggest doing a small benchmarks , see this comment https://github.com/SonarSource/sonar-dotnet/pull/4611#discussion_r662381123

andrei-epure-sonarsource (commented):
I see you addressed most comments and I would not do the CR again - I'd like to get answer to some things though before being ready to approve. Thanks!

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotTestThisWithIsOperator.CSharp9.cs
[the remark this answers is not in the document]
csaba-sagi-sonarsource: https://github.com/SonarSource/sonar-dotnet/issues/4634

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotTestThisWithIsOperator.cs
csaba-sagi-sonarsource: The stopping conditions are the same but I added some logic and test cases and now we do not raise FPs.

andrei-epure-sonarsource (approved):
LGTM

Issue #3674, opened 2020-10-27
Update the nuget SDK to a non-deprecated version


andrei-epure-sonarsource:
We should do what we did for [SONARSEC-1190](https://jira.sonarsource.com/browse/SONARSEC-1190) on [this PR.](https://github.com/SonarSource/sonar-security/pull/1243)

The nuget SDK we use has been deprecated and is no longer supported. The last version is from 2016.

andrei-epure-sonarsource:
Also check this comment - https://github.com/SonarSource/sonar-dotnet/pull/3678#discussion_r513305308

It's not related to the nuget API, but to the Roslyn `ProjectBuilder`. If this ticket doesn't do something to fix the above comment, we should decide whether we want to stick with the current approach (deduplicating MetadataReferences) or changing the test framework and tests setup to avoid having duplicates.

andrei-epure-sonarsource:
This popped out again - the current nuget libraries we use depend on the nuget v2 package feed `<add key="nuget.org v2" value="https://www.nuget.org/api/v2/" />`. 

`NuGetMetadataFactory.CreateNetStandard21()` calls `PackageManager.InstallPackage();` which fails if we remove the V2 line from `sonar-dotnet\analyzers\tests\SonarAnalyzer.UnitTest\NuGet.config`.

(issue found in #4526)

andrei-epure-sonarsource:
I added it to [our trello](https://trello.com/c/7WBJz4BE/247-update-the-nuget-sdk-used-in-our-unit-test-framework-to-a-non-deprecated-version) to plan it. It will need a short sprint.

PR #4610, opened 2021-06-29
Update nuget.config

andrei-epure-sonarsource:

### DIFF
--- a/analyzers/NuGet.Config
+++ b/analyzers/NuGet.Config
@@ -2,7 +2,7 @@
 <configuration>
   <packageSources>
     <clear />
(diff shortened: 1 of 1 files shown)


andrei-epure-sonarsource:
before modifying the config for our UTs, we need to fix this issue https://github.com/SonarSource/sonar-dotnet/issues/3674

csaba-sagi-sonarsource (approved):
LGTM! I would just like to ask you if by this change it means that it will not even fallback to nuget official package source?

andrei-epure-sonarsource:
@csaba-sagi-sonarsource yes that's correct, we will only go through repox.

PR #4611, opened 2021-06-29
Rule S1862 and S2760: add support for C# 9 pattern matching

pavel-mikula-sonarsource:
I don't want to fix this. PR contains only some cleanup and more UTs.

`CSharpEquivalenceChecker` can be updated to replace nodes in the tree to normalize null checks, but it's not worth the trouble and performance impact on all usages.

From syntax point of view, the check is not the same. Even when the semantic meaning is.

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpEquivalenceChecker.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpEquivalenceChecker.cs
@@ -57,6 +57,11 @@ namespace SonarAnalyzer.Helpers
                 }
                 return NullCheckExpression(binary) is { } expression ? new NullCheck(expression, positive) : null;
             }
+            else if (node.IsKind(SyntaxKindEx.IsPatternExpression))
+            {
+                var isPattern = (IsPatternExpressionSyntaxWrapper)node;
+                return NullCheckPattern(isPattern.Expression, isPattern.Pattern.SyntaxNode, positive);
+            }
             else
             {
                 return null;
@@ -72,6 +77,18 @@ namespace SonarAnalyzer.Helpers
             else if (binary.Right.IsKind(SyntaxKind.NullLiteralExpression))
             {
                 return binary.Left;
+            }
+            else
+            {
+                return null;
+            }
+        }
+
+        private static NullCheck NullCheckPattern(SyntaxNode expression, SyntaxNode pattern, bool positive)
+        {
+            if (pattern.IsKind(SyntaxKindEx.ConstantPattern) && ((ConstantPatternSyntaxWrapper)pattern).Expression.IsKind(SyntaxKind.NullLiteralExpression))
+            {
+                return new NullCheck(expression, positive);
             }
             else
             {
--- a/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpEquivalenceChecker.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpEquivalenceChecker.cs
@@ -37,30 +37,31 @@ namespace SonarAnalyzer.Helpers
             Common.EquivalenceChecker.AreEquivalent(nodeList1, nodeList2, NodeComparator);
 
         private static bool NodeComparator(SyntaxNode node1, SyntaxNode node2) =>
-            NullCheckState(node1) is { } nullCheck1
-            && NullCheckState(node2) is { } nullCheck2
-            && nullCheck1.Positive == nullCheck2.Positive
+            NullCheckState(node1, true) is { } nullCheck1
+            && NullCheckState(node2, true) is { } nullCheck2
+            && nullCheck1.IsPositive == nullCheck2.IsPositive
                 ? SyntaxFactory.AreEquivalent(nullCheck1.Expression, nullCheck2.Expression)
                 : SyntaxFactory.AreEquivalent(node1, node2);
 
-        private static NullCheck NullCheckState(SyntaxNode node, bool positive = true)
+        /// <param name="isPositive">Flag indicating that current chain of null checks is positive '== null'. It's flipped for each `!` operator. '!(x != null)' is equal to 'x == null'.</param>
+        private static NullCheck NullCheckState(SyntaxNode node, bool isPositive)
         {
             if (node is PrefixUnaryExpressionSyntax unary && unary.IsKind(SyntaxKind.LogicalNotExpression))
             {
-                return NullCheckState(unary.Operand.RemoveParentheses(), !positive);
+                return NullCheckState(unary.Operand.RemoveParentheses(), !isPositive);
             }
             else if (node is BinaryExpressionSyntax binary && binary.IsAnyKind(SyntaxKind.EqualsExpression, SyntaxKind.NotEqualsExpression))
             {
                 if (binary.IsKind(SyntaxKind.NotEqualsExpression))
                 {
-                    positive = !positive;
+                    isPositive = !isPositive;
                 }
-                return NullCheckExpression(binary) is { } expression ? new NullCheck(expression, positive) : null;
+                return NullCheckExpression(binary) is { } expression ? new NullCheck(expression, isPositive) : null;
             }
             else if (node.IsKind(SyntaxKindEx.IsPatternExpression))
             {
                 var isPattern = (IsPatternExpressionSyntaxWrapper)node;
-                return NullCheckPattern(isPattern.Expression, isPattern.Pattern.SyntaxNode, positive);
+                return NullCheckPattern(isPattern.Expression, isPattern.Pattern.SyntaxNode, isPositive);
             }
             else
             {
@@ -84,15 +85,16 @@ namespace SonarAnalyzer.Helpers
             }
         }
 
-        private static NullCheck NullCheckPattern(SyntaxNode expression, SyntaxNode pattern, bool positive)
+        /// <param name="isPositive">Flag indicating that current chain of null checks is positive 'is null'. It's flipped for each `not` operator. 'is not not null' is equal to 'is null'.</param>
+        private static NullCheck NullCheckPattern(SyntaxNode expression, SyntaxNode pattern, bool isPositive)
         {
             if (pattern.IsKind(SyntaxKindEx.ConstantPattern) && ((ConstantPatternSyntaxWrapper)pattern).Expression.IsKind(SyntaxKind.NullLiteralExpression))
             {
-                return new NullCheck(expression, positive);
+                return new NullCheck(expression, isPositive);
             }
             else if (pattern.IsKind(SyntaxKindEx.NotPattern))
             {
-                return NullCheckPattern(expression, ((UnaryPatternSyntaxWrapper)pattern).Pattern.SyntaxNode, !positive);
+                return NullCheckPattern(expression, ((UnaryPatternSyntaxWrapper)pattern).Pattern.SyntaxNode, !isPositive);
             }
             else
             {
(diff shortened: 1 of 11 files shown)


analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpNavigationHelper.cs
```diff
@@ -77,17 +77,7 @@ public static IEnumerable<SwitchSectionSyntax> GetPrecedingSections(this SwitchS
...
-            var statements = currentStatement.Parent.ChildNodes().OfType<StatementSyntax>().ToList();
-
-            var index = statements.IndexOf(currentStatement);
-
-            return index == 0 ? null : statements[index - 1];
-        }
-
-        #endregion Statement
+        public static StatementSyntax GetPrecedingStatement(this StatementSyntax currentStatement) =>
+            currentStatement.Parent.ChildNodes().OfType<StatementSyntax>().TakeWhile(x => x != currentStatement).LastOrDefault();
```
andrei-epure-sonarsource, line 81: I don't have data at hand for this change, but I know that link can sometimes be more expensive than non-LINQ queries

you could probably do some simple benchmark in LinqPad of how LINQ vs non-LINQ versions behave

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalsWithSameCondition.CSharp9.cs
```diff
@@ -62,10 +62,10 @@ public void Foo()
 
 
             if (f is not null) { }
-            if (f != null) { } // FN, same as above
+            if (f != null) { } // Compliant, even when the semantics is the same
```
andrei-epure-sonarsource: it's still an FN, even if it's "won't fix", right?

analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpEquivalenceChecker.cs
```diff
@@ -22,16 +22,74 @@
...
+        {
+            if (node is PrefixUnaryExpressionSyntax unary && unary.IsKind(SyntaxKind.LogicalNotExpression))
+            {
+                return NullCheckState(unary.Operand.RemoveParentheses(), !positive);
+            }
+            else if (node is BinaryExpressionSyntax binary && binary.IsAnyKind(SyntaxKind.EqualsExpression, SyntaxKind.NotEqualsExpression))
+            {
+                if (binary.IsKind(SyntaxKind.NotEqualsExpression))
+                {
+                    positive = !positive;
```
andrei-epure-sonarsource, line 56: `positive = !positive` is confusing logic because:
- it's a bit of a mind-twister because `!positive` is `negative` for the human mind - `isPositive` may be a better name
- the method has `bool positive = true` by default, and when seeing this I expect that "if I do not explicitly pass the parameter, it means it's positive", however it's not the case  because this is actually evaluated based on the syntax kind
 
And being a recursive function it makes it a bit harder to understand. 

Please add some documentation for the `NullCheckState` method functionality, specifying that in the first run the sign is evaluated inside. (it took me a while to understand it, which means it needs documentation to help the reader)

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/EquivalenceCheckerTest.cs
```diff
@@ -208,30 +209,34 @@ public void EqualityComparer_Node_VB()
...
 
-            var hashSet = new HashSet<SyntaxList<VisualBasic.StatementSyntax>>(new[] { vbMethods.Method1, vbMethods.Method2, vbMethods.Method3 }, comparer);
+            var hashSet = new HashSet<SyntaxList<VB.StatementSyntax>>(new[] { vbMethods.Method1, vbMethods.Method2, vbMethods.Method3 }, comparer);
             hashSet.Should().HaveCount(2);
             hashSet.Should().Contain(vbMethods.Method1);
             hashSet.Should().NotContain(vbMethods.Method4);
         }
 
+        [TestMethod]
+        public void EqualityComparer_Node_CrossLangauge() =>
```
andrei-epure-sonarsource, line 227:
```suggestion
        public void EqualityComparer_Node_CrossLanguage() =>
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalsWithSameCondition.CSharp9.cs
```diff
@@ -52,11 +52,25 @@ public void Foo()
                 f = cond ? new Apple() : new Orange();
             }
 
+            if (f is null) { }
+            if (f is null) { }  // Noncompliant
+
+            char ch = 'x';
+            if (ch is >= 'a' and <= 'z') { }
+            if (ch is >= 'a' and <= 'z') { } // Noncompliant
+            if (ch is <= 'z' and >= 'a') { } // Compliant, even when the semantics is the same
```
andrei-epure-sonarsource, line 61: could you also add some more complex patterns like 
```
c is (>= 'a' and <= 'z') or (>= 'A' and <= 'Z') or '.' or ',';
```
and
```
c is (>= 'A' and <= 'Z')  or (>= 'a' and <= 'z') or '.' or ',';
```
(I just swapped the first two)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalsWithSameCondition.CSharp9.cs
```diff
@@ -52,11 +52,25 @@ public void Foo()
                 f = cond ? new Apple() : new Orange();
             }
 
+            if (f is null) { }
```
andrei-epure-sonarsource, line 55: could you also add some noncompliant tests with type pattern? we only have compliant ones

```
if (f is Apple a) { ; }
if (f is Apple a) { ; } // Noncompliant
if (f is not Apple) { ; }
if (f is not Apple) { ; } // Noncompliant
```
Also for positional pattern you could add [the example](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/patterns#positional-pattern) from the docs - I'm not sure if the compiler is smart enough to raise that (if so, we may not want to duplicate what it does)

Also - could you please add tests with property patterns (probably some logic needs to be updated in the equivalence checker)
```
if (a is { Count: >= 5 }) {}
if (a is { Count: >= 5 }) {} // Noncompliant
```
and more complex ones
```
if (a is { Count: >= 5 }) {}
if (a is not { Count: < 5 }) {} // probably FN
```
and
```
if (segment is { Start: { X: 0, Y: 0 } } or { End: { X: 0, Y: 0 } }) {};
if (segment is { Start: { X: 0, Y: 0 } } or { End: { X: 0, Y: 0 } }) {}; // Noncompliant
```

andrei-epure-sonarsource (changes requested):
very nice refactorings

my main suggestion is to add more tests with positional and property patterns

analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpNavigationHelper.cs
pavel-mikula-sonarsource: Linq has 50% improvement over the expensive `IndexOf` call. 
I just measured it with 10000 repetitions for 10 and 100 statements, starting from middle.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalsWithSameCondition.CSharp9.cs
pavel-mikula-sonarsource: For the docs example, you can't have the same switch arm conditions. It raises error:
>Error CS8510 The pattern is unreachable. It has already been handled by a previous arm of the switch expression or it is impossible to match.

pavel-mikula-sonarsource: I'll skip the last one, it's too difficult to scaffold and we have complicated cases already.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalsWithSameCondition.CSharp9.cs
```diff
@@ -60,6 +60,11 @@ public void Foo()
             if (ch is >= 'a' and <= 'z') { } // Noncompliant
             if (ch is <= 'z' and >= 'a') { } // Compliant, even when the semantics is the same
 
+            if (ch is (>= 'a' and <= 'z') or (>= 'A' and <= 'Z') or '.' or '?') { }
+            if (ch is (>= 'a' and <= 'z') or (>= 'A' and <= 'Z') or '.' or '?') { } // Noncompliant
+            if (ch is (>= 'a' and <= 'z') or (>= 'A' and <= 'Z') or '?' or '.') { } // Compliant, even when the semantics is the same
```
andrei-epure-sonarsource, line 65: I really suggest to use `FN` instead of `Compliant` for these "even when" cases

Compare this with lines 76-78 - it _really is compliant_ - here it's just an FN. If we would raise, it would help developers

andrei-epure-sonarsource (approved):
LGTM, but please change "Compliant, even when reason" with "FN, reason" - otherwise it will confused future readers of the test cases

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalsWithSameCondition.CSharp9.cs
```diff
@@ -52,17 +52,57 @@ public void Foo()
...
+
+            if (f is not Apple) { }
+            if (f is not Apple) { } // Noncompliant
+
+            if (f is { Size: >= 5, Value: 0 }) { }
+            if (f is { Size: >= 5, Value: 0 }) { } // Noncompliant
+            if (f is { Value: 0, Size: >= 5 }) { } // FN, rule is syntax based. Only null check equivalence is currenty supported in CSharpEquivalenceChecker
+
+            if (f is { Size: >= 5 }) { }
+            if (f is not { Size: < 5 }) { } // FN, rule is syntax based. Only null check equivalence is currenty supported in CSharpEquivalenceChecker
```
andrei-epure-sonarsource, line 95: I finally created a compilable snippet with positional patterns that you could add to the tests - [see sharplab](https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzguATEBqAPgAQGYACfAJmIGFiBvAWAChjnST8AWYgWQAoBBAA4CANgFNiEAJS1GLOcSgAzYjwgKUKgEQBlAO6jRCTXGKaASqIybJk2fOb0m94gF879pSrVQNPHfsNjUwsrG3d5R2dXcLlPVXVaABVUBFEQUz0DIxNKAHthXIAndPNLTRdbJ3tI5zcq6IY6xiJSCkERcRrmFvwARgAGYmS0UQBucJ6Bqnyi8fqezgARUQBjXJg0QoBXFYQeXK2EUimADxMDo77BgE9pAF4APhUz4lviO5Vh1JyZwsk5lxAA)

```
public class C {
    public void M(Apple a) {
        if (a is ("Sweet", "Red"))
        {
        }
        if (a is ("Sweet", "Red"))
        {
        }
        if (a is {Taste: "Sweet", Color: "Red"})
        {
        }
    }
}

public class Apple {
    public string Taste;
    public string Color;
    public void Deconstruct(out string x, out string y) => (x, y) = (Taste, Color);
}
```

(sorry for the late comment)

PR #4612, opened 2021-06-29
Rule S1067: add support for C# 9 pattern matching

pavel-mikula-sonarsource:
Review per commit is suggested

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ExpressionComplexity.CSharp9.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ExpressionComplexity.CSharp9.cs
@@ -2,10 +2,10 @@
 
 object x = null;
 
-if (x is true or true or true or true or true) { } // FN
+if (x is true or true or true or true or true) { } // Noncompliant
 
-if (x is true and true and true and true and true) { } // FN
+if (x is true and true and true and true and true) { } // Noncompliant
 
-if (x is 10 or 20 or 40 or 50 or 60 or 70) { } // FN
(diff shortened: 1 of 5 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ExpressionComplexity.CSharp9.cs
```diff
@@ -2,10 +2,10 @@
 
 object x = null;
```
csaba-sagi-sonarsource, line 4: This case seems to me as a FN:
```
switch (x)
{
    case true and true and true and true and true and true:
        break;
    case false:
        break;
    default:
        break;
}
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ExpressionComplexity.CSharp9.cs
```diff
@@ -2,10 +2,10 @@
```
csaba-sagi-sonarsource, line 2: Can we add this case:
```
var y = x switch
{
    true and true and true and true and true and true => 1,
    false => 2,
};
```

csaba-sagi-sonarsource (changes requested):
Please add the suggested test cases.

csaba-sagi-sonarsource (approved):
LGTM. While doing the review, I realized that we cover also:
```
try
{
}
catch (Exception e) when (e is ArgumentNullException or ArgumentOutOfRangeException or DuplicateWaitObjectException or DivideByZeroException or NotFiniteNumberException or OverflowException) // Noncompliant
{
    throw;
}
```
So it would be nice to just add a test case for it.

PR #4614, opened 2021-06-30
Rule S2219: add support for C# 9 pattern matching

pavel-mikula-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/GetTypeWithIsAssignableFrom.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/GetTypeWithIsAssignableFrom.cs
@@ -27,6 +27,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Extensions;
 using SonarAnalyzer.Helpers;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
@@ -91,11 +92,21 @@ namespace SonarAnalyzer.Rules.CSharp
                     }
                 },
                 SyntaxKind.IsExpression);
+
+            context.RegisterSyntaxNodeActionInNonGenerated(c =>
+            {
+                var isPattern = (IsPatternExpressionSyntaxWrapper)c.Node;
+                if (ConstantPatternExpression(isPattern.Pattern) is { } constantExpression)
+                {
+                    CheckAsOperatorComparedToNull(c, isPattern.Expression, constantExpression);
+                }
+            },
+            SyntaxKindEx.IsPatternExpression);
         }
 
         private static void CheckAsOperatorComparedToNull(SyntaxNodeAnalysisContext context, ExpressionSyntax sideA, ExpressionSyntax sideB)
         {
-            if (sideA.RemoveParentheses().IsKind(SyntaxKind.NullLiteralExpression) && sideB.RemoveParentheses().IsKind(SyntaxKind.AsExpression))
+            if (sideA.RemoveParentheses().IsKind(SyntaxKind.AsExpression) && sideB.RemoveParentheses().IsKind(SyntaxKind.NullLiteralExpression))
             {
                 ReportDiagnostic(context, MessageIsOperator);
             }
@@ -137,6 +148,14 @@ namespace SonarAnalyzer.Rules.CSharp
                 }
             }
         }
+
+        private static ExpressionSyntax ConstantPatternExpression(SyntaxNode node) =>
+            node.Kind() switch
+            {
+                SyntaxKindEx.ConstantPattern => ((ConstantPatternSyntaxWrapper)node).Expression,
+                SyntaxKindEx.NotPattern => ConstantPatternExpression(((UnaryPatternSyntaxWrapper)node).Pattern),
+                _ => null
+            };
 
         private static void ReportDiagnostic(SyntaxNodeAnalysisContext context, string messageArg, bool useIsOperator = false, bool shouldRemoveGetType = false)
         {
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/GetTypeWithIsAssignableFrom.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/GetTypeWithIsAssignableFrom.cs
@@ -130,7 +130,7 @@ namespace SonarAnalyzer.Rules.CSharp
         {
             if (methodSymbol.Name == "IsInstanceOfType" && memberAccess.Expression is TypeOfExpressionSyntax)
             {
-                ReportDiagnostic(context, MessageIsOperator, true, false);
+                ReportDiagnostic(context, MessageIsOperator, true);
             }
         }
 
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/GetTypeWithIsAssignableFrom.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/GetTypeWithIsAssignableFrom.cs
@@ -115,9 +115,9 @@ namespace SonarAnalyzer.Rules.CSharp
         private static void CheckGetTypeAndTypeOfEquality(SyntaxNodeAnalysisContext context, ExpressionSyntax sideA, ExpressionSyntax sideB)
         {
             if (sideA.ToStringContains("GetType")
-                && sideB is TypeOfExpressionSyntax sideBeTypeOf
+                && sideB is TypeOfExpressionSyntax sideBTypeOf
                 && (sideA as InvocationExpressionSyntax).IsGetTypeCall(context.SemanticModel)
-                && context.SemanticModel.GetTypeInfo(sideBeTypeOf.Type).Type is { } typeSymbol // Can be null for empty identifier from 'typeof' unfinished syntax
+                && context.SemanticModel.GetTypeInfo(sideBTypeOf.Type).Type is { } typeSymbol // Can be null for empty identifier from 'typeof' unfinished syntax
                 && typeSymbol.IsSealed
                 && !typeSymbol.OriginalDefinition.Is(KnownType.System_Nullable_T))
             {
@@ -129,7 +129,7 @@ namespace SonarAnalyzer.Rules.CSharp
         {
             if (methodSymbol.Name == "IsInstanceOfType" && memberAccess.Expression is TypeOfExpressionSyntax)
             {
-                ReportDiagnostic(context, MessageIsOperator, true);
+                ReportDiagnostic(context, MessageIsOperator, useIsOperator: true);
             }
         }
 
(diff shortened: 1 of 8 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/GetTypeWithIsAssignableFrom.cs
```diff
@@ -81,9 +81,9 @@ protected override void Initialize(SonarAnalysisContext context)
             context.RegisterSyntaxNodeActionInNonGenerated(c =>
                 {
                     var isExpression = (BinaryExpressionSyntax)c.Node;
-                    if (c.SemanticModel.GetTypeInfo(isExpression.Left).Type is { } objectToCast
+                    if (c.SemanticModel.GetTypeInfo(isExpression.Left).Type is var objectToCast
```
andrei-epure-sonarsource, line 84: `public ITypeSymbol Type { get; }` : `The type of the expression represented by the syntax node. For expressions that do not have a type, null is returned.`

the below throws
```
void Main()
{
	if (X() is var objectToCast)
	{
		Console.WriteLine(objectToCast.ToString());	
	}
	
	object X() => null;
}
```

andrei-epure-sonarsource: nevermind, the nullcheck is done inside `IsClass`
(which is good for extension methods however it's un-intuitive for the reader)

analyzers/src/SonarAnalyzer.CSharp/Rules/GetTypeWithIsAssignableFrom.cs
```diff
@@ -76,105 +69,99 @@ protected override void Initialize(SonarAnalysisContext context)
...
         {
-            if (sideA.ToStringContains("GetType") &&
-                sideB is TypeOfExpressionSyntax sideBeTypeOf &&
-                sideBeTypeOf.Type is { } typeSyntax &&
-                (sideA as InvocationExpressionSyntax).IsGetTypeCall(context.SemanticModel) &&
-                context.SemanticModel.GetTypeInfo(typeSyntax).Type is { } typeSymbol &&
-                typeSymbol.IsSealed &&
-                !typeSymbol.OriginalDefinition.Is(KnownType.System_Nullable_T))
+            if (sideA.ToStringContains("GetType")
+                && sideB is TypeOfExpressionSyntax sideBeTypeOf
```
andrei-epure-sonarsource, line 118:
```suggestion
                && sideB is TypeOfExpressionSyntax sideBTypeOf
```

analyzers/src/SonarAnalyzer.CSharp/Rules/GetTypeWithIsAssignableFrom.cs
```diff
@@ -76,105 +69,99 @@ protected override void Initialize(SonarAnalysisContext context)
...
-                        .Add(UseIsOperatorKey, true.ToString())
-                        .Add(ShouldRemoveGetType, false.ToString()),
-                    MessageIsOperator));
+                if (memberAccess.Expression is TypeOfExpressionSyntax)
+                {
+                    ReportDiagnostic(context, MessageIsOperator, true, true);
+                }
+                else
+                {
+                    ReportDiagnostic(context, MessageIsInstanceOfType, false, true);
```
andrei-epure-sonarsource, line 146: note to self: I'll need to revisit this refactoring again

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/GetTypeWithIsAssignableFrom.CSharp9.cs
```diff
@@ -1,17 +1,20 @@
...
+b = f as Apple is null;         // Noncompliant
+b = f as Apple is not null;     // Noncompliant
+b = f as Apple is not not not not null; // Noncompliant
+b = f as Apple is Apple { Size: 42 };   // Compliant
+b = f as Apple is Apple;        // Noncompliant
+b = true && (apple) is Apple;   // Noncompliant
+b = !(apple is Apple);          // Noncompliant
 
-if (apple is { }) { } // Compliant
+if (apple is { }) { }           // Compliant
```
andrei-epure-sonarsource, line 17: could you also add more pattern types to make sure it doesn't break the logic or behave strange?

```
public class C {
    public void M(Apple a) {
        if (a is ("Sweet", "Red"))
        {
        }
        if (a is ("Sweet", "Red"))
        {
        }
        if (a is {Taste: "Sweet", Color: "Red"})
        {
        }
    }
}

public class Apple {
    public string Taste;
    public string Color;
    public void Deconstruct(out string x, out string y) => (x, y) = (Taste, Color);
}
```

andrei-epure-sonarsource (commented):
from what I see, it looks good

I left some minor comments

I need to have another look at the refactoring commits to clarify some things to myself before approving, but I don't expect much surprises

analyzers/src/SonarAnalyzer.CSharp/Rules/GetTypeWithIsAssignableFrom.cs
```diff
@@ -76,105 +69,99 @@ protected override void Initialize(SonarAnalysisContext context)
...
-                    MessageIsOperator)
-                : Diagnostic.Create(rule, invocation.GetLocation(),
-                    ImmutableDictionary<string, string>.Empty
-                        .Add(UseIsOperatorKey, false.ToString())
-                        .Add(ShouldRemoveGetType, true.ToString()),
-                    MessageIsInstanceOfType));
+        private static void ReportDiagnostic(SyntaxNodeAnalysisContext context, string messageArg, bool useIsOperator = false, bool shouldRemoveGetType = false)
+        {
+            var properties = ImmutableDictionary<string, string>.Empty
+            .Add(UseIsOperatorKey, useIsOperator.ToString())
```
andrei-epure-sonarsource, line 162: indent

analyzers/src/SonarAnalyzer.CSharp/Rules/GetTypeWithIsAssignableFrom.cs
andrei-epure-sonarsource: I think named arguments may help with readability here and when the arguments are passed explicitly

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/GetTypeWithIsAssignableFrom.CSharp9.Fixed.cs
```diff
@@ -0,0 +1,20 @@
...
+b = apple.GetType() == typeof(int?);                // Compliant
+b = apple.GetType().IsInstanceOfType(f.GetType());  // Compliant
+b = !(f is Apple);         // Fixed
+b = !(f is Apple);         // Fixed
+b = f is Apple;     // Fixed
+b = !(f is Apple); // Fixed
+b = f as Apple is Apple { Size: 42 };   // Compliant
+b = f is Apple;        // Fixed
+b = true && (apple != null);   // Fixed
+b = !(apple != null);          // Fixed
```
andrei-epure-sonarsource, line 15: this is not great, it should be refactored well from the first , I believe - also, it would better be `apple is not null`

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/GetTypeWithIsAssignableFrom.CSharp9.Fixed.cs
```diff
@@ -0,0 +1,20 @@
+﻿Apple apple = new();
+Fruit f = apple;
+
+var b = f is Apple;     // Compliant
+b = f is not Apple;     // Compliant
+b = apple.GetType() == typeof(int?);                // Compliant
+b = apple.GetType().IsInstanceOfType(f.GetType());  // Compliant
+b = !(f is Apple);         // Fixed
```
andrei-epure-sonarsource, line 8: this should be `f is not Apple`, same for line 11

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/GetTypeWithIsAssignableFrom.CSharp9.Fixed.cs
andrei-epure-sonarsource: to do the proper refactoring might take time, I suggest to skip doing the refactoring when having double negation

analyzers/src/SonarAnalyzer.CSharp/Rules/GetTypeWithIsAssignableFromCodeFixProvider.cs
```diff
@@ -29,143 +29,141 @@
...
+            if (isPattern.Expression is BinaryExpressionSyntax binary)
+            {
+                var negationRequired = true;
+                var current = isPattern.Pattern;
+                while (current.SyntaxNode.IsKind(SyntaxKindEx.NotPattern))
+                {
+                    negationRequired = !negationRequired;
+                    current = ((UnaryPatternSyntaxWrapper)current).Pattern;
+                }
+                var newExpression = NegatedExpression(negationRequired, isPattern.SyntaxNode.Parent, GetIsExpression(binary));
```
andrei-epure-sonarsource, line 100: this should not reuse the same `NegatedExpression()` for `IsPattern` , because `!(x is Y)` should actually be `x is not Y`

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/GetTypeWithIsAssignableFromCodeFixProvider.cs
andrei-epure-sonarsource: I opened https://github.com/SonarSource/sonar-dotnet/issues/4633 for this.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/GetTypeWithIsAssignableFrom.CSharp9.Fixed.cs
andrei-epure-sonarsource: I opened https://github.com/SonarSource/sonar-dotnet/issues/4633 for this.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/GetTypeWithIsAssignableFrom.CSharp9.Fixed.cs
andrei-epure-sonarsource: I opened https://github.com/SonarSource/sonar-dotnet/issues/4633 for this.

andrei-epure-sonarsource (approved):
LGTM.

I looked a bit and to make code fixes with `a is not Apple`, we need to improve the StyleCopAnalyzer SyntaxFactoryEx, which is out of scope for this sprint.

PR #4615, opened 2021-06-30
Rule S3981: add support for C# 9 pattern matching

csaba-sagi-sonarsource:


analyzers/src/SonarAnalyzer.Common/Rules/DoNotCheckZeroSizeCollectionBase.cs
```diff
@@ -59,32 +56,34 @@ public abstract class DoNotCheckZeroSizeCollectionBase<TLanguageKindEnum,
...
                 && GetSymbol(context, expressionValueNode) is { } symbol
                 && GetDeclaringTypeName(symbol) is { } symbolType)
             {
-                context.ReportDiagnosticWhenActive(Diagnostic.Create(SupportedDiagnostics[0], context.Node.GetLocation(), symbol.Name.ToLowerInvariant(), symbolType));
+                context.ReportDiagnosticWhenActive(Diagnostic.Create(SupportedDiagnostics[0], issueLocation.GetLocation(), symbol.Name, symbolType));
             }
         }
 
-        private string GetDeclaringTypeName(ISymbol symbol)
+        protected string GetDeclaringTypeName(ISymbol symbol)
```
dmitry-golovinov-sonarsource, line 86: Why? It could stay private, we don't use it anywhere else.

dmitry-golovinov-sonarsource (approved):
Great job, just one minor you can consider fixing.

PR #4616, opened 2021-06-30
Read SonarLint.xml with Roslyn GetText()

### LINKED ISSUE #4603: Read SonarLint.xml with Roslyn GetText() API instead of disk
We currently use the `SonarLint.xml` to:
- pass rule parameters from SonarLint for Visual Studio and from the Scanner for .NET.
- communicate when to run Security Hotspots. In batch mode, the Scanner for .NET adds all active rules to the `<Rules>` section.
- soon SonarLint for Rider will also pass rule parameters in an in-memory SonarLint.xml file - it won't be saved on disk.

Problem: we currently assume the SonarLint.xml file is on disk in the following places:
- https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs#L109
- https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/src/SonarAnalyzer.Common/Helpers/ParameterLoader.cs#L37

We should use the `GetText()` API, as the `SonarLint.xml` may be passed just in-memory (by SonarLint for Rider).

Also, if a user has a SonarLint.xml file in their project and they have edited but not saved it, VS will pass the in-memory version to the analyzers, but the analyzer will still (unexpectedly) used the saved version (this is a corner case, documented here for future reference).

rita-sonarsource:
Fixes #4603


rita-sonarsource:
@duncanp-sonar please review; it might help to review each commit separately.

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleLoaderTest.cs
```diff
@@ -33,12 +33,25 @@ public class RuleLoaderTest
...
+            sut.Invoking(x => x.GetEnabledRules("not xml")).Should().Throw<XmlException>();
         }
 
         [TestMethod]
-        public void GivenAnExistingSonarLintFile_RuleLoader_LoadsActiveRules() =>
-            CollectionAssert.AreEqual(new RuleLoader().GetEnabledRules(@"Common/Resources/SonarLint.xml").ToArray(),
-                                      new[] { "S1067" });
+        public void GivenSonarLintXml_RuleLoader_LoadsActiveRules()
+        {
+            const string content = @"<?xml version=""1.0"" encoding=""UTF-8""?>
```
rita-sonarsource, line 42: This used to be in the deleted SonarLint.xml file.
Since this class no longer reads anything from disk, I thought it would make the test clearer if the text was in plain string.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ParameterLoaderTest.cs
```diff
@@ -100,47 +101,46 @@ public void SetParameterValues_WhenGivenValidSonarLintFileAndDoesNotContainAnaly
         }
 
         [TestMethod]
-        public void SetParameterValues_CalledTwiceAfterChangeInConfigFile_UpdatesProperties()
+        public void SetParameterValues_WithNonExistingPath_UsesInMemoryText()
```
rita-sonarsource, line 104: Some comparison alignment problems here due to a mistake on my part:
This test replaces the previous `SetParameterValues_WithBadPath_DoesNotPopulateProperties` which was deleted below.
`SetParameterValues_CalledTwiceAfterChangeInConfigFile_UpdatesProperties` is modified just after this one.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/PropertiesHelperTest.cs
```diff
@@ -51,15 +53,14 @@ public class PropertiesHelperTest
         public void ShouldAnalyzeGeneratedCode_WithNotBooleanValue_ReturnsFalse() =>
             GetSetting("ResourceTests\\NotBoolean\\SonarLint.xml").Should().BeFalse();
 
-        [TestMethod]
```
rita-sonarsource, line 54: This test is no longer relevant

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/PropertiesHelperTest.cs
```diff
@@ -51,15 +53,14 @@ public class PropertiesHelperTest
...
-        public void ShouldAnalyzeGeneratedCode_WithNotExistingPath_ReturnsFalse() =>
-            GetSetting("ResourceTests\\NotExistingFolder\\SonarLint.xml").Should().BeFalse();
-
         private static bool GetSetting(string filePath)
         {
             // Arrange
+            var content = SourceText.From(File.ReadAllText(filePath));
```
rita-sonarsource, line 59: All the tests of this class are now working in-memory additional file but receiving their data from the disk prep files for the sake of minimal changes.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestHelper.cs
```diff
@@ -137,12 +137,17 @@ public static bool IsSecurityHotspot(DiagnosticDescriptor diagnostic)
                 ? NuGetMetadataReference.MSTestTestFrameworkV1  // Any reference to detect a test project
                 : Enumerable.Empty<MetadataReference>();
 
-        public static AnalyzerOptions CreateOptions(string relativePath)
+        public static AnalyzerOptions CreateOptions(string relativePath, SourceText text = null)
```
rita-sonarsource, line 140: A bit hacky. Used in the ParameterLoaderTest `SetParameterValues_CalledTwiceAfterChangeInConfigFile_UpdatesProperties`

analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
```diff
@@ -84,20 +85,20 @@ public void Initialize(AnalyzerOptions options)
 
             public void Initialize(AnalyzerOptions options)
             {
-                var currentSonarLintXmlPath = GetSonarLintXmlPath(options);
-                if (isInitialized && loadedSonarLintXmlPath == currentSonarLintXmlPath)
+                var currentSonarLintXml = GetSonarLintXml(options);
+                if (isInitialized && loadedSonarLintXmlPath == currentSonarLintXml?.Path)
```
duncanp-sonar, line 89: @rita-sonarsource, this caching is potentially an issues for the Rider scenario, isn't it?

Even if it is, we should handle it separately from this PR.

analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
```diff
@@ -84,20 +85,20 @@ public void Initialize(AnalyzerOptions options)
 
             public void Initialize(AnalyzerOptions options)
             {
-                var currentSonarLintXmlPath = GetSonarLintXmlPath(options);
-                if (isInitialized && loadedSonarLintXmlPath == currentSonarLintXmlPath)
+                var currentSonarLintXml = GetSonarLintXml(options);
```
duncanp-sonar, line 88: The old name `currentSonarLintXmlPath` made sense because it was clear it referred to a path. 
However, I found `currentSonarLintXml` a bit confusing because it suggest the variable refers to XML.

It's difficult to think of a good name that isn't too long. `currentSonarLintXmlFile` maybe, since the method is returning an additional file?

I'd rename the method to `GetSonarLintXmlFile` too.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestHelper.cs
duncanp-sonar: It might be neater to make the `text` parameter a string and to split the method into two e.g.

`CreateOptions(string relativePath)` and 
`CreateOptions(string path, string content)`.

The second method would just create and return the options based on the supplied data.

The first would contain the old "_use the content from the file if it exists_" logic.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ParameterLoaderTest.cs
duncanp-sonar: `...NonExistentPath...`

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ParameterLoaderTest.cs
```diff
@@ -100,47 +101,46 @@ public void SetParameterValues_WhenGivenValidSonarLintFileAndDoesNotContainAnaly
...
 
-            // revert change
-            File.WriteAllText(sonarLintXmlRelativePath, originalContent);
+            // Assert
+            analyzer.Maximum.Should().Be(1); // In-memory value
         }
 
         [TestMethod]
-        public void SetParameterValues_WithMalformedXml_DoesNotPopulateProperties()
+        public void SetParameterValues_CalledTwiceAfterChangeInConfigFile_UpdatesProperties()
```
duncanp-sonar, line 119: It might be neater just to inline the XML here and in `SetParameterValues_WithNonExistingPath_UsesInMemoryText`, like you did for the test in the first commit.

duncanp-sonar (commented):
Looks good, and looks like the minimum set of changes required to fix the bug.

It feels like there's an missing abstraction/set of common methods to handle reading and caching the SonarLintXml file, but that's beyond the scope of this fix.

analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
rita-sonarsource: It seems to be in use only for Hotspots. We could handle it on our side by passing a different folder name each time, as we discussed offline. I'm not sure it's worth opening a ticket now, since we're not expected to enable hotspots any time soon.

analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
```diff
@@ -84,20 +85,20 @@ public void Initialize(AnalyzerOptions options)
 
             public void Initialize(AnalyzerOptions options)
             {
-                var currentSonarLintXmlPath = GetSonarLintXmlPath(options);
-                if (isInitialized && loadedSonarLintXmlPath == currentSonarLintXmlPath)
+                var currentSonarLintXmlFile = GetSonarLintXmlFile(options);
+                if (isInitialized && loadedSonarLintXmlPath == currentSonarLintXmlFile?.Path)
```
andrei-epure-sonarsource, line 89: you do the null-check three times - maybe store the path in a local var?

andrei-epure-sonarsource (commented):
intermediate comments

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ParameterLoaderTest.cs
```diff
@@ -100,47 +101,81 @@ public void SetParameterValues_WhenGivenValidSonarLintFileAndDoesNotContainAnaly
...
+            analyzer.Maximum.Should().Be(42);
         }
 
         [TestMethod]
-        public void SetParameterValues_WithBadPath_DoesNotPopulateProperties()
+        public void SetParameterValues_WithMalformedXml_DoesNotPopulateProperties()
         {
             // Arrange
-            var options = TestHelper.CreateOptions("ResourceTests\\ThisPathDoesNotExist\\SonarLint.xml");
+            var options = TestHelper.CreateOptions("ResourceTests\\Malformed\\SonarLint.xml");
```
andrei-epure-sonarsource, line 178: where is the malformed xml? not clear from the test. I expect it's null?

To make it clear, you can make it a DataTestMethod (i.e. parameterized) and pass more "malformed" values like null, empty string and malformed XML to cover all cases

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/PropertiesHelperTest.cs
```diff
@@ -51,15 +53,14 @@ public class PropertiesHelperTest
...
             var additionalText = new Mock<AdditionalText>();
-            additionalText.Setup(x => x.Path).Returns(filePath);
+            additionalText.Setup(x => x.Path).Returns("fakePath\\SonarLint.xml"); // use in-memory additional file
```
andrei-epure-sonarsource, line 61: please add a test when the path does not contain the "SonarLint.xml" _(clean as you code methodology, as the behavior should not have changed but we should have it documented in tests)_

andrei-epure-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ParameterLoaderTest.cs
rita-sonarsource: It's from "ResourceTests\\Malformed\\SonarLint.xml" -- I didn't change it, that's the way it was.

rita-sonarsource: Added use-cases

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/PropertiesHelperTest.cs
rita-sonarsource: Done

andrei-epure-sonarsource (approved):
Lgtm

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/ParameterLoaderTest.cs
andrei-epure-sonarsource: My bad

analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
andrei-epure-sonarsource: discussed offline with Rita that anyhow this method deserves more refactoring and this PR is minimal to unblock SonarLint for Rider

PR #4618, opened 2021-07-02
Delete S3693 (deprecated)

costin-zaharia-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/PackagingTests/CsRuleTypeMapping.cs
```diff
@@ -3690,7 +3690,7 @@ internal static class CsRuleTypeMapping
             //["3690"],
```
csaba-sagi-sonarsource, line 3690: This same change in VbRuleTypeMapping is missing.

csaba-sagi-sonarsource (changes requested):
One more change needed.

analyzers/tests/SonarAnalyzer.UnitTest/PackagingTests/CsRuleTypeMapping.cs
costin-zaharia-sonarsource: Nice catch! Thanks!

csaba-sagi-sonarsource (approved):
LGTM!

PR #4621, opened 2021-07-02
AZP: Rename pipelines-yaml-templates from commonTemplates to pipelines-yaml-templates

pavel-mikula-sonarsource:
This confuses me every time I read the YML file.


costin-zaharia-sonarsource (approved):
LGTM!

PR #4625, opened 2021-07-02
Additional cases where uses of integers are clearly not magic

StingyJack:
Per request https://community.sonarsource.com/t/s109-magic-numbers-on-attributes/15784/15

@csaba-sagi-sonarsource

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
@@ -2,6 +2,7 @@
 
 namespace Tests.Diagnostics
 {
+    using System.Runtime.Caching;
     public class ValidUseCases
     {
         private const int MAGIC = 42;
@@ -27,10 +28,20 @@ namespace Tests.Diagnostics
             }
 
             const int VAL = 15;
-
+            
             Console.Write("test");
         }
 
+        /// <summary>
+        ///     A five minute sliding expiration 
+        /// </summary>
+        public static CacheItemPolicy FiveMinSlide => new CacheItemPolicy {SlidingExpiration = new TimeSpan(0, 5, 0)};
+
+        /// <summary>
+        ///     A five minute absolute expiration
+        /// </summary>
+        public static CacheItemPolicy FiveMinAbs => new CacheItemPolicy {AbsoluteExpiration = DateTime.Now.AddMinutes(5)};
+        
         public override int GetHashCode()
         {
             return MY_VALUE * 397;


StingyJack:
An additional possible case is present in https://referencesource.microsoft.com/#system.web/Security/SQLMembershipProvider.cs,1039

I had been porting some of this to .net core and because the stored procedure returns a selection of variables, the resulting field names in the SqlDataReader are all `""` and the values can only be retrieved using an index. I dont know whats worse, leaving it as it is, or creating a bunch of constants like `const int ResultSetColumnOrdinal_Password = 0`. Neither seem appealing.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
```diff
@@ -27,10 +28,20 @@ public ValidUseCases()
...
             const int VAL = 15;
-
+            
             Console.Write("test");
         }
 
+        /// <summary>
+        ///     A five minute sliding expiration 
+        /// </summary>
+        public static CacheItemPolicy FiveMinSlide => new CacheItemPolicy {SlidingExpiration = new TimeSpan(0, 5, 0)};
```
andrei-epure-sonarsource, line 38: indeed DateTime APIs could be considered as well-known cases. The problem is how to avoid having to maintain such a list, we would need to find something generic enough.

andrei-epure-sonarsource (commented):
Thanks @StingyJack 

We've discussed internally and created an issue with the cases we'd like to cover: https://github.com/SonarSource/sonar-dotnet/issues/4737

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
StingyJack: if parameter name is the name of a specific time measurement ("day" "month" "year", etc), then the usage of a literal has an obvious, specific, and easy to understand meaning and is thus not "magic".

StingyJack: The same would be true for "Order"

Corniel: May be first go back to the drawing board, and think of what to support first?

StingyJack: @Corniel  as a user of sonarcloud, I want it to report when a literal number has been used in a place where its not obvious what that number's meaning is.

Corniel: @StingyJack I got that point. This a pull request. I saw that @andrei-epure-sonarsource already made an issue to discuss the fllow up: #4737

StingyJack: Then I'm not sure what you're looking for here from me

Corniel: @StingyJack Well, I'm just a volunteer, but I would think that your input on what should not be considered magical numbers would help. As @andrei-epure-sonarsource andrei-epure-sonarsource mentioned, the challenge is to come with generic constraints, instead of some types that should be excluded, as the maintenance cost is high for those.

andrei-epure-sonarsource: I initially asked for a PR with test cases on the community forum, hence this PR.
I will close this PR in  favor of #4737.
Thanks @StingyJack for the discussion and this PR.

andrei-epure-sonarsource:
I will close this PR in favor of #4737.

PR #4626, opened 2021-07-02
Add repro for #4559

### LINKED ISSUE #4559: S2583 FP: User-defined explicit/implicit casts
### Description
S2583 ignores custom cast operators and assumes nullability remains the same between casts (i.e. nulls are always converted to nulls and not-null values are always converted to not-null values).

```csharp
public class MyClass
{
    public static implicit operator string?(MyClass? value) => null;

    public static implicit operator bool?(MyClass? value) => true;
}

public void Foo()
{
    // case 1: nullability has changed (from always not null to always null)
    var value = new MyClass();
    string? str = value;

    if (str == null) // <- FP: S2583: Change this condition so it does not always evaluate to 'false'
    {
        Console.WriteLine("null");
    }
    else
    {
        Console.WriteLine("not null");
    }

    // case 1: nullability has changed (from null to always not null)
    MyClass? value2 = null;
    bool? b = value2;

    if (b == null) // <- FP: S2583: Change this condition so it does not always evaluate to 'true'
    {
        Console.WriteLine("null");
    }
    else
    {
        Console.WriteLine("not null");
    }
}
```

### Expected behavior

The diagnostic is not reported.

### Related information

* C#/VB.NET Plugins version: 8.24.0.32949

costin-zaharia-sonarsource:
Related to #4559


andrei-epure-sonarsource (approved):
LGTM

Issue #4540, opened 2021-06-10
S108: False Positive when conditional compilation is used


rjgotten:
### Description

"S108: Nested blocks of code should not be left empty" gives a false positive when conditional compilation is used with `#if` pragmas.

### Repro steps

Create anything using conditional compilation with nothing else within the method. E.g. something trivial like

```cs
public void LogInDebug(object obj)
{
#if DEBUG
    Trace.WriteLine( "Something to log only in debug", obj);
#endif
}
```

### Expected behavior

S108 is not raised, because clearly the block isn't (always) empty.

### Actual behavior

S108 is raised.

### Known workarounds

None. (Just mark it as a false positive and live with it, I suppose.)

### Related information

* C#/VB.NET Plugins version 8.13.1
* Visual Studio version N/A
* MSBuild / dotnet version ?
* SonarScanner for .NET version (if used) ?
* Operating System ?

(This is with SonarScanner running from a TeamCity build agent inside a Docker container. I have no information wrt those versions available to me.)

costin-zaharia-sonarsource:
Hi @rjgotten, thanks for your feedback.

I've tried to add a test case for your issue here: https://github.com/SonarSource/sonar-dotnet/pull/4627 but I'm not able to reproduce the problem. It could be that I use a different version of `dotnet` and due to this a different compiler option. 

Can you please help us with the MsBuild/dotnet version? Maybe you can find it in the build logs.

rjgotten:
@costin-zaharia-sonarsource 
I can maybe add a few `RUN` lines to the container file to dump version information that way.
I'll have a look.

costin-zaharia-sonarsource:
Hi @rjgotten. I've managed to reproduce the issue this morning.
The conditional had to be in a nested block. See:

```
public void ConditionalCompilation(object obj)
{
   {
#if DEBUG
       Trace.WriteLine("message");
#endif
   }
}
```

rjgotten:
Oops. I just looked back at the issue report and yeah; I posted the wrong reduced case.
The right one was using a local function - i.e. had a nested block.

That's my bad. Apologies.

(Well... at least we have a repro now.)

costin-zaharia-sonarsource:
Yes, don't worry about that. Thanks for the feedback!

PR #4627, opened 2021-07-02
Add test case with conditional compilation

costin-zaharia-sonarsource:
Related to https://github.com/SonarSource/sonar-dotnet/issues/4540

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/EmptyNestedBlock.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/EmptyNestedBlock.cs
@@ -86,9 +86,11 @@ namespace Tests.Diagnostics
         // See: https://github.com/SonarSource/sonar-dotnet/issues/4540
         public void ConditionalCompilation(object obj)
         {
+            {
 #if DEBUG
-            Trace.WriteLine("message");
+                Trace.WriteLine("message");
 #endif
+            } // Noncompliant@-4 FP: the block is not empty
         }
     }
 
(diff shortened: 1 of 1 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/EmptyNestedBlock.cs
```diff
@@ -81,6 +82,14 @@ void F1(bool b)
         void F2()
         {
         }
+
+        // See: https://github.com/SonarSource/sonar-dotnet/issues/4540
+        public void ConditionalCompilation(object obj)
```
csaba-sagi-sonarsource, line 87: If I understood correctly this rule looks only at nested blocks. So, the FP can be reproduced with the following code example:
```
        // See: https://github.com/SonarSource/sonar-dotnet/issues/4540
        public void ConditionalCompilation(object obj)
        {
            if (obj != null)
            {
#if DEBUG
                Trace.WriteLine("message");
#endif
            }
        }
```

csaba-sagi-sonarsource (changes requested):
Please try the suggested reproducer.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/EmptyNestedBlock.cs
costin-zaharia-sonarsource: Thanks. I've took the example from the issue without reading the rule description thoroughly.

csaba-sagi-sonarsource (approved):
LGTM! I think we should tell the reporter of the issue that we managed to create a reproducer.

costin-zaharia-sonarsource:
> I think we should tell the reporter of the issue that we managed to create a reproducer.

Yes, I've done that already.

PR #4628, opened 2021-07-03
Use an Assert method to test if a specific exception was thrown also for VB.NET

Corniel:
As suggested on [the sonar forum](https://community.sonarsource.com/t/expectedexceptionattribute-is-an-anti-pattern/45415); an `Assert` method should be used to test if a specific exception was thrown, not an `[ExpectedException]` attribute.

-------------------------
# Description
Using `[ExpectedException]` is seen as bad practice for testing if a piece of code is throwing the expected exception,
as it can not distinguish if the exception was thrown by the code under test, or by any of the arranging code.

## Noncompliant Code Example
``` C#
[TestMethod]
[ExpectedException(typeof(ArgumentNullException))]
public void NullArgument_Throws()
{
    var arrange = new Instance(null);
	arrange.MethodUnderTest(null);
}
```

``` VB.NET
<TestMethod>
<ExpectedException(GetType(ArgumentNullException))>
Public Sub NullArgument_Throws()

    Dim arrange As New Instance(Nothing)
	arrange.MethodUnderTest(Nothing)

end Sub
```

## Compliant Solution
``` C#
[TestMethod]
public void NullArgument_Throws()
{
    var arrange = new Instance(null);
	Assert.ThrowsException<ArgumentNull>(() => arrange.MethodUnderTest(null));
}
```

``` VB.NET
<TestMethod>
Public Sub NullArgument_Throws()

    Dim arrange As New Instance(Nothing)
	Assert.ThrowsException(Of ArgumentNull)(Sub() arrange.MethodUnderTest(Nothing))

End Sub
```
**Labels**: bad-practice testability (not sure that is a existing label) 
**Message**: "Use an Assert method to test the thrown exception.
**Default Severity**: Major
**Impact**: Low
**Likelihood**: Medium
**Default Quality Profiles**: Sonar way (I would argue so)
**Covered Languages**: C#, VB.Net	
**Remediation Function**: ?
**Constant Cost**: 5min
**Analysis Level**: ?
**Syntactic Analysis Analysis Scope**: Test Sources


csaba-sagi-sonarsource:
@Corniel please take a look at rule [3431](https://rules.sonarsource.com/csharp/RSPEC-3431) as I think it is the same rule.

Corniel:
> 
> 
> @Corniel please take a look at rule [3431](https://rules.sonarsource.com/csharp/RSPEC-3431) as I think it is the same rule.

I guess so. Although this one also support VB.NET. I'll merge it with 3431 and extend it to VB.NET. May be, you could tell the person who asked for this feature on the Sonar forum, the rule already exists?!

Corniel:
Implementation added for VB.NET. I would like to challenge the choice of allowing the attribute for one line of code/expression-body. It might not fail in testing the exception, but having both mechanisms in a code base, is confusing.

csaba-sagi-sonarsource:
> Implementation added for VB.NET. I would like to challenge the choice of allowing the attribute for one line of code/expression-body. It might not fail in testing the exception, but having both mechanisms in a code base, is confusing.

There can be codebases where only the one liners are used, and It could be annoying for the users to be forced to change to `Assert.ThrowsException` for no obvious reason in their case.

Corniel:
> 
> 
> > Implementation added for VB.NET. I would like to challenge the choice of allowing the attribute for one line of code/expression-body. It might not fail in testing the exception, but having both mechanisms in a code base, is confusing.
> 
> There can be codebases where only the one liners are used, and It could be annoying for the users to be forced to change to `Assert.ThrowsException` for no obvious reason in their case.

To me, that feels far fetched, and up to that exceptional code base to disable it. But  if you want to support it, I'll add it.

Corniel:
Done.

analyzers/src/SonarAnalyzer.Common/Rules/ExpectedExceptionShouldNotBeUsedAttributeBase.cs
```diff
@@ -0,0 +1,57 @@
...
+                if (HasMultiLineBody(c.Node)
+                    && c.SemanticModel.GetDeclaredSymbol(c.Node) is { } methodSymbol
+                    && methodSymbol.GetAttributes(UnitTestHelper.KnownExpectedExceptionAttributes).FirstOrDefault() is { } attribute)
+                {
+                    c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, attribute.ApplicationSyntaxReference.GetSyntax().GetLocation()));
+                }
+            },
+            Language.SyntaxKind.MethodDeclarations);
+
+        protected abstract bool HasMultiLineBody(SyntaxNode syntax);
```
csaba-sagi-sonarsource, line 55: NITPICK: According to our [coding style](https://github.com/SonarSource/sonar-dotnet/blob/master/docs/coding-style.md#class-members) abstract members should go third inside classes, so please move this abstract method below the Language property.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/ExpectedExceptionAttributeShouldNotBeUsedTest.cs
```diff
@@ -19,9 +19,10 @@
  */
 
 using Microsoft.VisualStudio.TestTools.UnitTesting;
-using SonarAnalyzer.Rules.CSharp;
 using SonarAnalyzer.UnitTest.MetadataReferences;
 using SonarAnalyzer.UnitTest.TestFramework;
+using CS = SonarAnalyzer.Rules.CSharp;
+using VB = SonarAnalyzer.Rules.VisualBasic;
 
 namespace SonarAnalyzer.UnitTest.Rules
```
csaba-sagi-sonarsource, line 27: NITPICK: According to our [coding style](https://github.com/SonarSource/sonar-dotnet/blob/master/docs/coding-style.md#naming-conventions) the test methods should have _CS and _VB suffixes and not prefixes. Please rename the test methods.

csaba-sagi-sonarsource (changes requested):
Two coding style fixes are needed.

Corniel:
@csaba-sagi-sonarsource Could you generate the proper resource file?

csaba-sagi-sonarsource:
> @csaba-sagi-sonarsource Could you generate the proper resource file?

@Corniel  I created the following [branch](https://github.com/SonarSource/sonar-dotnet/tree/%C4%8Daba/S3431-VBNET), feel free to cherry pick what you need.

Corniel:
> 
> 
> > @csaba-sagi-sonarsource Could you generate the proper resource file?
> 
> @Corniel I created the following [branch](https://github.com/SonarSource/sonar-dotnet/tree/%C4%8Daba/S3431-VBNET), feel free to cherry pick what you need.

Picked my cherries.

analyzers/src/SonarAnalyzer.Utilities/Rules.Description/S3431_cs.html
```diff
@@ -1,40 +1,40 @@
...
-public void TestNullArg()
-{
-  Assert.ThrowsException&lt;ArgumentNullException&gt;(() =&gt; /*...*/);
-}
-</pre>
-<h2>Exceptions</h2>
-<p>This rule ignores one-line test methods, since it is obvious in such methods where the exception is expected to be thrown.</p>
-
+<p>It should be clear to a casual reader what code a test is testing and what results are expected. Unfortunately, that’s not usually the case with
+the <code>[ExpectedException]</code> attribute since an exception could be thrown from almost any line in the method.</p>
```
csaba-sagi-sonarsource, line 2: I just realised that I messed up with the RSPEC generation, please remove the square brackets around the ExpectedException in this line. Sorry for messing up.

csaba-sagi-sonarsource (approved):
LGTM!

Corniel:
@csaba-sagi-sonarsource 
I was able to increase the coverage for C#. However, I could not make it work for the VB.NET equivalent.

``` VB.NET
  <Test>
    <ExpectedException(GetType(DivideByZeroException))>
    Public Function NoBody() ' Error ['BC30027] 'End Function' expected.
```
Both fails on complication, and does not increase coverage.

Corniel (commented):
(I cannot cancel this?)

csaba-sagi-sonarsource:
Could you please also fix the code smell? (Remove the unnecessary using)

PR #4635, opened 2021-07-06
Rule S3440: add support for C# 9 pattern matching

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantConditionalAroundAssignment.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantConditionalAroundAssignment.cs
@@ -19,6 +19,7 @@
  */
 
 using System.Collections.Immutable;
+using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -26,6 +27,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Extensions;
 using SonarAnalyzer.Helpers;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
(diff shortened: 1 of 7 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantConditionalAroundAssignment.CSharp9.cs
```diff
@@ -3,12 +3,49 @@
```
csaba-sagi-sonarsource, line 3: Please try to cover or at least document the following FNs:
```
object a = null;

switch (a)
{
    case null:
        {
            a = null;
        }
        break;
    default:
        break;
}

if ((x,y) is (1,2))
{
    x = 1;
    y = 2;
}
```

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantConditionalAroundAssignment.cs
```diff
@@ -19,13 +19,15 @@
  */
```
csaba-sagi-sonarsource, line 19: I see that there is a codefix  provider for the rule. Lets try to update that too to work with c#9 patterns if it is a low hanging fruit.

csaba-sagi-sonarsource (changes requested):
The changes made are good, and lets try to cover some more cases.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantConditionalAroundAssignment.CSharp9.cs
dmitry-golovinov-sonarsource: Documented them as FNs for now

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantConditionalAroundAssignment.cs
```diff
@@ -36,58 +38,80 @@ public sealed class RedundantConditionalAroundAssignment : SonarDiagnosticAnalyz
...
+            {
+                c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, condition.GetLocation()));
+            }
+        }
+
+        private static void UselessConditionSwitchExpression(SyntaxNodeAnalysisContext c)
+        {
+            var switchExpression = (SwitchExpressionSyntaxWrapper)c.Node;
+
+            if (switchExpression.Arms.Count != 1
```
csaba-sagi-sonarsource, line 85: Because of this check we do not cover something like:
```
int y = 6;

y = y switch
{
    5 => 5, // Noncompliant
    6 => 6 // Noncompliant
};
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantConditionalAroundAssignment.CSharp9.Fixed.cs
```diff
@@ -0,0 +1,64 @@
+﻿int x = 5;
```
csaba-sagi-sonarsource, line 1: A case that we could also cover is the following:
```
y = y switch
{
    _ => y
};
```

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantConditionalAroundAssignmentCodeFixProvider.cs
```diff
@@ -46,11 +47,24 @@ protected override Task RegisterCodeFixesAsync(SyntaxNode root, CodeFixContext c
             var condition = root.FindNode(diagnosticSpan) as ExpressionSyntax;
             var ifStatement = condition?.FirstAncestorOrSelf<IfStatementSyntax>();
 
-            if (ifStatement == null)
+            if (ifStatement != null)
+            {
+                return HandleIfStatement(root, context, ifStatement);
+            }
+
+            var switchExpression = root.FindNode(diagnosticSpan).FirstAncestorOrSelf((SyntaxNode x) => x.IsKind(SyntaxKindEx.SwitchExpression));
```
csaba-sagi-sonarsource, line 55: This is a bit more readable for me, but feel free to leave it as is if you don`t agree with it.
```suggestion
            var switchExpression = root.FindNode(diagnosticSpan).FirstAncestorOrSelf<SyntaxNode>(x => x.IsKind(SyntaxKindEx.SwitchExpression));
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantConditionalAroundAssignment.CSharp9.cs
```diff
@@ -3,14 +3,57 @@
 
 y = y switch
 {
-    not 5 => 5 // Compliant - FN
+    not 5 => 5 // Noncompliant
 };
 
 y = y switch
```
csaba-sagi-sonarsource, line 9: I would say this is an edge case.  Because y will be set to 5 only if it was 5, otherwise an exception will be thrown.

csaba-sagi-sonarsource (changes requested):
Last round of additional cases.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantConditionalAroundAssignment.CSharp9.cs
dmitry-golovinov-sonarsource: No action

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantConditionalAroundAssignment.cs
```diff
@@ -36,58 +38,82 @@ public sealed class RedundantConditionalAroundAssignment : SonarDiagnosticAnalyz
...
+                c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, condition.GetLocation()));
+            }
+        }
+
+        private static void UselessConditionSwitchExpression(SyntaxNodeAnalysisContext c)
+        {
+            var switchExpression = (SwitchExpressionSyntaxWrapper)c.Node;
+
+            if (!(switchExpression.SyntaxNode.GetFirstNonParenthesizedParent() is AssignmentExpressionSyntax parentExpression)
+                || !CSharpEquivalenceChecker.AreEquivalent(parentExpression.Left, switchExpression.GoverningExpression))
```
csaba-sagi-sonarsource, line 86: Do we need this equivalence check? I think because of it we do not handle cases like:
```
y = x switch
{
    _ => 5       // Noncompliant
};
```

csaba-sagi-sonarsource (changes requested):
Great work, once last case left to cover.

csaba-sagi-sonarsource (approved):
LGTM!

PR #4638, opened 2021-07-07
Rule S4201 - improve support for C# 9 pattern matching

andrei-epure-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheck.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheck.cs
@@ -155,15 +155,22 @@ namespace SonarAnalyzer.Rules.CSharp
             {
                 var leftPattern = (PatternSyntaxWrapper)left;
                 var rightPattern = (PatternSyntaxWrapper)right;
-                if (leftPattern.IsNull() && rightPattern.IsNot())
+                if (leftPattern.IsNull() && IsNegativePatternMatch(rightPattern))
                 {
                     context.ReportDiagnosticWhenActive(Diagnostic.Create(RuleForPatternSyntax, left.GetLocation()));
                 }
-                else if (rightPattern.IsNull() && leftPattern.IsNot())
+                else if (rightPattern.IsNull() && IsNegativePatternMatch(leftPattern))
                 {
                     context.ReportDiagnosticWhenActive(Diagnostic.Create(RuleForPatternSyntax, right.GetLocation()));
                 }
             }
+
+            // Verifies that it's like a negative pattern except 'not null' e.g. 'not Apple', 'not (5 or 6)'.
+            // The PatternSyntax appears e.g. in switch arms and is different from IsPatternSyntax.
+            static bool IsNegativePatternMatch(PatternSyntaxWrapper node) =>
+                node.IsNot()
+                && ((UnaryPatternSyntaxWrapper)node) is var unaryPatternSyntaxWrapper
+                && !unaryPatternSyntaxWrapper.Pattern.IsNull();
         }
 
         /// <summary>
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheck.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheck.cs
@@ -65,13 +65,12 @@ namespace SonarAnalyzer.Rules.CSharp
 
         protected override SyntaxNode GetRightNode(BinaryExpressionSyntax binaryExpression) => binaryExpression.Right.RemoveParentheses();
 
-        protected override SyntaxNode GetNullCheckVariable(SyntaxNode node) => GetNullCheckVariableForKind(node, SyntaxKind.EqualsExpression);
+        protected override SyntaxNode GetNullCheckVariable(SyntaxNode node) => GetNullCheckVariableForKind(node, true);
 
-        protected override SyntaxNode GetNonNullCheckVariable(SyntaxNode node) => GetNullCheckVariableForKind(node, SyntaxKind.NotEqualsExpression);
+        protected override SyntaxNode GetNonNullCheckVariable(SyntaxNode node) => GetNullCheckVariableForKind(node, false);
 
         protected override SyntaxNode GetIsOperatorCheckVariable(SyntaxNode node)
         {
-            // FIXME use ConditionalSimplification.TryGetExpressionComparedToNull - duplicate the method and keep what makes sense
             var innerExpression = node.RemoveParentheses();
             if (innerExpression is BinaryExpressionSyntax binaryExpression && binaryExpression.IsKind(SyntaxKind.IsExpression))
             {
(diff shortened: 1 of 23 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheck.cs
```diff
@@ -86,21 +77,114 @@ protected override SyntaxNode GetIsOperatorCheckVariable(SyntaxNode node)
...
+                }
+            }
+
+            // Verifies if the given pattern syntax is like "not X" - where X can be constant pattern (except 'null'), Declaration pattern, Recursive pattern.
+            // The PatternSyntax appears e.g. in switch arms and is different from IsPatternSyntax.
+            static bool IsNegativePatternMatch(PatternSyntaxWrapper patternSyntaxWrapper) =>
+                patternSyntaxWrapper.IsNot() && !patternSyntaxWrapper.IsNull();
+        }
+
+        // expectedAffirmative
```
dmitry-golovinov-sonarsource, line 171: In my opinion such a comment does not look good without a context. But I have no idea how to make it better.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantNullCheck.CSharp9.cs
```diff
@@ -1,29 +1,163 @@
...
 {
 }
 
-if (n is not null && n is not null) // Compliant - FN
+if (!(n is null) && n is Apple) // Noncompliant
 {
 }
 
-if (n is not null && n != null) // Noncompliant
+var result = (n is string && n is not null); // Noncompliant
```
dmitry-golovinov-sonarsource, line 20: I would advice adding the following example to the test set (or even a series)
```
switch(n)
{
    case not null and "sweet":
        break;
    default:
        break;
}
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantNullCheck.CSharp9.cs
```diff
@@ -1,29 +1,163 @@
-﻿object n = 5;
+﻿object a = "";
```
dmitry-golovinov-sonarsource, line 1: General remark - we are missing location verification for new C#9 features. I would test it as well, to have an overview, whether it makes sense.

dmitry-golovinov-sonarsource (commented):
Nice work. Left just a few suggestion, where we could improve, found nothing major.

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheck.cs
```diff
@@ -155,20 +155,15 @@ private void CheckOrPattern(SyntaxNodeAnalysisContext context)
             {
                 var leftPattern = (PatternSyntaxWrapper)left;
                 var rightPattern = (PatternSyntaxWrapper)right;
-                if (leftPattern.IsNull() && IsNegativePatternMatch(rightPattern))
+                if (leftPattern.IsNull() && rightPattern.IsNot())
```
andrei-epure-sonarsource, line 158: I'm pretty sure I am missing something in the test cases, but they are green and the code coverage was incomplete, thus that previous logic was not necessary

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheck.cs
```diff
@@ -165,7 +165,7 @@ private void CheckOrPattern(SyntaxNodeAnalysisContext context)
             // Verifies if the given pattern syntax is like "not X" - where X can be constant pattern (except 'null'), Declaration pattern, Recursive pattern.
             // The PatternSyntax appears e.g. in switch arms and is different from IsPatternSyntax.
             static bool IsNegativePatternMatch(PatternSyntaxWrapper patternSyntaxWrapper) =>
-                patternSyntaxWrapper.IsNot() && !patternSyntaxWrapper.IsNull();
```
andrei-epure-sonarsource: I'm pretty sure I am missing something in the test cases, but they are green and the code coverage was incomplete, thus that previous logic was not necessary

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantNullCheck.CSharp9.cs
```diff
@@ -1,29 +1,227 @@
...
 if (n != null && n is nint) // Noncompliant
 {
 }
 
-if (n is not null && n is nint) // Compliant - FN
+if (n is not null && n is nint nintValue) // Noncompliant
 {
 }
 
-if (n is not null && n is not null) // Compliant - FN
```
andrei-epure-sonarsource, line 11: for the record: I removed this test case because this rule should not raise an issue in this case (consistency with C# 7 test cases)

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheck.cs
dmitry-golovinov-sonarsource: You are right. Actually, IsNull requires ConstantPattern and IsNot NotPattern, so even  `not null` in `null or not null` won't match it.

dmitry-golovinov-sonarsource: While investigating it, I've found a FP we should be aware of 
```
x = a switch
{
    null or not null => ""
};
```
Even the check makes no sense, removing null from it is a bad idea, since then it won't match null cases. WDYT?

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheckCodeFixProvider.cs
```diff
@@ -24,63 +24,89 @@
...
+                    RegisterBinaryExpressionCodeFix(context, root, isPatternExpression.SyntaxNode);
+                }
+            }
+            else if (PatternSyntaxWrapper.IsInstance(diagnosticNode))
             {
-                return TaskHelper.CompletedTask;
+                RegisterBinaryPatternCodeFix(context, root, ((PatternSyntaxWrapper)diagnosticNode).SyntaxNode);
             }
 
+            return TaskHelper.CompletedTask;
```
dmitry-golovinov-sonarsource, line 71: Is there a reason why don't we fix 

```
var x = a switch
{
    null or not "a" => "",
//  ^^^^
    _ => ""
};
```

Should be pretty easy to do for nulls in such patterns.

dmitry-golovinov-sonarsource (commented):
Nice work! I think we can improve coverage, probably for the line 115 this should do the trick
`if (apple == null || apple is { Taste: "Sweet", Color: "Red" }) { }`
Also, please check other comments I've left!

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheck.cs
andrei-epure-sonarsource: Thank you, this was it! I was a bit tired with all the test cases. Will fix it.

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheckCodeFixProvider.cs
andrei-epure-sonarsource: good catch!

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantNullCheck.CSharp9.Fixed.cs
```diff
@@ -23,6 +23,8 @@
 // parenthesized pattern
 result = n is ("a" or "b" or "c"); // Fixed
 
+if (n is null && apple != null && m is null && apple is ("Sweet", "Red")) { } // FN
```
andrei-epure-sonarsource, line 26: I'll open a ticket for this, as I want to keep the timebox and this is needs a non-trivial fix - basically it should model the logical expression and for each variable inside, keep track of the verifications it has. It would be a big change.

andrei-epure-sonarsource: https://github.com/SonarSource/sonar-dotnet/issues/4645

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheckCodeFixProvider.cs
```diff
@@ -24,62 +24,93 @@
...
+            }
+            else if (diagnosticNode is PrefixUnaryExpressionSyntax prefixUnary && prefixUnary.IsKind(SyntaxKind.LogicalNotExpression))
             {
-                return TaskHelper.CompletedTask;
+                RegisterBinaryExpressionCodeFix(context, root, prefixUnary);
+            }
+            else if (IsPatternExpressionSyntaxWrapper.IsInstance(diagnosticNode))
+            {
+                var isPatternExpression = (IsPatternExpressionSyntaxWrapper)diagnosticNode.RemoveParentheses();
+                if (isPatternExpression.IsNull() || isPatternExpression.IsNotNull())
```
andrei-epure-sonarsource, line 61: I think this is redundant, because anyway at this point we know a diagnostic has been raised on `diagnosticNode`, so it can only be `is null` or `is not null`

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheckCodeFixProvider.cs
```diff
@@ -24,62 +24,93 @@
...
+                {
+                    RegisterBinaryExpressionCodeFix(context, root, isPatternExpression.SyntaxNode);
+                }
+            }
+            else if (PatternSyntaxWrapper.IsInstance(diagnosticNode))
+            {
+                RegisterBinaryPatternCodeFix(context, root, ((PatternSyntaxWrapper)diagnosticNode).SyntaxNode);
+            }
+            else if (diagnosticNode.IsNullLiteral() && diagnosticNode.Parent.IsKind(SyntaxKindEx.ConstantPattern))
```
andrei-epure-sonarsource, line 70: this is defensive coding

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheckCodeFixProvider.cs
dmitry-golovinov-sonarsource: Do you want to remove it or leave as defensive?

dmitry-golovinov-sonarsource (approved):
LGTM! Please see my question. Feel free to merge it in any case.

analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantNullCheckCodeFixProvider.cs
andrei-epure-sonarsource: I removed it.

PR #4639, opened 2021-07-08
S3981 FN fix, check <0

### LINKED ISSUE #3735: Rule S3981 : should also check `<0`

csaba-sagi-sonarsource:
Fixes #3735

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCheckZeroSizeCollection.CSharp9.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCheckZeroSizeCollection.CSharp9.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 
+const int localConst_Zero = 0;
 var someEnumerable = new List<string>();
 var anotherEnumerable = new List<string>();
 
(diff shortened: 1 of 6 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCheckZeroSizeCollection.CSharp9.cs
```diff
@@ -12,14 +13,18 @@
...
+
 if (((someEnumerable.Count()), anotherEnumerable.Count()) is (>= 0, // Noncompliant
 //                                                            ^^^^
     >= 0)) // Noncompliant
 //  ^^^^
 {
 }
 
-if (someEnumerable.Count() is < 0) { } // FN
+if (someEnumerable.Count() is < 0) { } // Noncompliant
```
dmitry-golovinov-sonarsource, line 27: Let's add one example with localConst_Zero here. It's nice to show that we support it as well.

dmitry-golovinov-sonarsource (approved):
LGTM! Let's add one more test that I've stated in the comments.

PR #4640, opened 2021-07-08
Fix S1118: Rule should not raise issue on abstract classes

### LINKED ISSUE #4608: S1118: Rule being triggered for abstract classes
### Description

Rule S1118 is being triggered for abstract classes that contain only static methods.

A fix had previously been implemented for this (see #1411), however, it does not exist in the current [master](https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/src/SonarAnalyzer.CSharp/Rules/ClassWithOnlyStaticMember.cs). Was it potentially reverted (as it needs to check whether the class is both static or abstract, not just look for one of the keywords)?

### Repro steps

The following snippet will trigger the rule:-

```
    public abstract class ExampleClass
    {
        protected static bool ExampleMethod()
        {
            return true;
        }
    }
```

### Expected behavior

Rule to not be triggered for abstract classes (abstract classes are not static or non-static, that is a detail of the class which inherits from them).

### Actual behavior

Rule triggered for an abstract class containing only static methods.

### Known workarounds

None.

### Related information

SonarAnalyzer.CSharp version 8.25.0.33663

dmitry-golovinov-sonarsource:
Fixes #4608

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ClassWithOnlyStaticMember.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ClassWithOnlyStaticMember.cs
@@ -110,6 +110,7 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static bool ClassQualifiesForIssue(INamedTypeSymbol @class) =>
             !@class.IsStatic
+            && !@class.IsAbstract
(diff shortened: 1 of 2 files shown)


andrei-epure-sonarsource (approved):
LGTM!

PR #4642, opened 2021-07-09
Don't call ToUpperInvariant()

costin-zaharia-sonarsource:
This is expensive and executed in a loop it has a significant impact on performance.

image


andrei-epure-sonarsource (approved):
LGTM

PR #4644, opened 2021-07-09
S126 FP fix:  Report only for empty else blocks

### LINKED ISSUE #2284: Update S126: Rule should handle exceptions listed in the main RSPEC
Rule [RSPEC-126](https://jira.sonarsource.com/browse/RSPEC-126) does not implement in sonar-dotnet the same exceptions as the main rule description.

However this rule is not in sonarway and it is not clear why. Let's wait to have more information on the false positives rate on SonarCloud, or complaints from users before updating the rule implementation.

csaba-sagi-sonarsource:
Fixes #2284

### DIFF
--- a/analyzers/src/SonarAnalyzer.VisualBasic/Rules/IfChainWithoutElse.cs
+++ b/analyzers/src/SonarAnalyzer.VisualBasic/Rules/IfChainWithoutElse.cs
@@ -45,8 +45,8 @@ namespace SonarAnalyzer.Rules.VisualBasic
 
         private static bool IsEmptyBlock(MultiLineIfBlockSyntax multiLineIfBlock) =>
             !(multiLineIfBlock.ElseBlock.Statements.Count > 0
-            || multiLineIfBlock.ElseBlock.GetTrailingTrivia().Any(elseBlockTrivia => IsCommentOrDisabledText(elseBlockTrivia))
-            || multiLineIfBlock.EndIfStatement.GetLeadingTrivia().Any(endIfTrivia => IsCommentOrDisabledText(endIfTrivia)));
+            || multiLineIfBlock.ElseBlock.GetTrailingTrivia().Any(IsCommentOrDisabledText)
+            || multiLineIfBlock.EndIfStatement.GetLeadingTrivia().Any(IsCommentOrDisabledText));
 
         private static bool IsCommentOrDisabledText(SyntaxTrivia trivia) =>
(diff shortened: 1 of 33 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/IfChainWithoutElse.cs
```diff
@@ -34,24 +36,75 @@ public IfChainWithoutElse(bool a)
...
             else if (a) { }
-            else if (a) // Noncompliant {{Add the missing 'else' clause.}}
+            else if (a) // Noncompliant {{Add the missing 'else' clause with either the appropriate action or a suitable comment as to why no action is taken.}}
 //          ^^^^^^^
             {
             }
 
             if (a) { }
             else if (a) { }
+            else if (a) { }// Noncompliant
```
dmitry-golovinov-sonarsource, line 46:
```suggestion
            else if (a) {  }  // Noncompliant
```

analyzers/src/SonarAnalyzer.VisualBasic/Rules/IfChainWithoutElse.cs
```diff
@@ -54,5 +55,19 @@ protected override void Initialize(SonarAnalysisContext context)
...
+        private static bool IsElseIfWithoutElse(MultiLineIfBlockSyntax node)
+        {
+            return node.ElseIfBlocks.Any()
+                   && (node.ElseBlock == null || IsEmptyBlock(node));
+        }
+
+        private static bool IsEmptyBlock(MultiLineIfBlockSyntax multiLineIfBlock) =>
+            !(multiLineIfBlock.ElseBlock.Statements.Count > 0
+            || multiLineIfBlock.ElseBlock.DescendantTrivia().Any(elseBlockTrivia => IsCommentOrDisabledText(elseBlockTrivia))
+            || multiLineIfBlock.EndIfStatement.DescendantTrivia().Any(endIfTrivia => IsCommentOrDisabledText(endIfTrivia)));
```
dmitry-golovinov-sonarsource: We can make it nicer if you want
```
            || multiLineIfBlock.ElseBlock.GetTrailingTrivia().Any(IsCommentOrDisabledText)
            || multiLineIfBlock.EndIfStatement.GetLeadingTrivia().Any(IsCommentOrDisabledText));
```

dmitry-golovinov-sonarsource (approved):
LGTM! See my two minors.

PR #4646, opened 2021-07-09
S108 FP fix-when conditional compilation

### LINKED ISSUE #4540: S108: False Positive when conditional compilation is used

csaba-sagi-sonarsource:
Fixes #4540

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/EmptyNestedBlock.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/EmptyNestedBlock.cs
@@ -19,7 +19,6 @@
  */
 
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -32,50 +31,8 @@ namespace SonarAnalyzer.Rules.CSharp
 {
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     [Rule(DiagnosticId)]
-    public sealed class EmptyNestedBlock : EmptyNestedBlockBase
+    public sealed class EmptyNestedBlock : EmptyNestedBlockBase<SyntaxKind>
     {
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
-
-        protected override void Initialize(SonarAnalysisContext context)
-        {
-            context.RegisterSyntaxNodeActionInNonGenerated(
-                c =>
-                {
-                    if (IsEmpty(c.Node))
-                    {
-                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, c.Node.GetLocation()));
-                    }
-                },
-                SyntaxKind.Block,
-                SyntaxKind.SwitchStatement);
-        }
-
-        private static bool IsEmpty(SyntaxNode node)
-        {
-            var blockNode = node as BlockSyntax;
-
-            return (node is SwitchStatementSyntax switchNode && IsEmpty(switchNode)) ||
-                (blockNode != null && IsNestedAndEmpty(blockNode));
-        }
-
-        private static bool IsEmpty(SwitchStatementSyntax node)
-        {
-            return !node.Sections.Any();
-        }
-
-        private static bool IsNestedAndEmpty(BlockSyntax node)
-        {
-            return IsNested(node) && IsEmpty(node);
-        }
-
-        private static bool IsNested(BlockSyntax node)
-        {
-            return !AllowedContainerKinds.Contains(node.Parent.Kind());
-        }
-
         private static readonly ISet<SyntaxKind> AllowedContainerKinds = new HashSet<SyntaxKind>
         {
             SyntaxKind.ConstructorDeclaration,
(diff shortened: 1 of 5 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/EmptyNestedBlock.cs
```diff
@@ -32,50 +31,8 @@ namespace SonarAnalyzer.Rules.CSharp
...
-            return IsNested(node) && IsEmpty(node);
-        }
-
-        private static bool IsNested(BlockSyntax node)
-        {
-            return !AllowedContainerKinds.Contains(node.Parent.Kind());
-        }
-
         private static readonly ISet<SyntaxKind> AllowedContainerKinds = new HashSet<SyntaxKind>
         {
```
Corniel, line 37: An array of 2 items is faster then an hashset.

Corniel (approved):

analyzers/src/SonarAnalyzer.CSharp/Rules/EmptyNestedBlock.cs
```diff
@@ -86,19 +43,41 @@ private static bool IsNested(BlockSyntax node)
...
+            !node.Sections.Any();
+
+        private static bool IsEmpty(BlockSyntax node) =>
+            !node.Statements.Any() && !ContainsComment(node);
+
+        private static bool IsNestedAndEmpty(BlockSyntax node) =>
+            IsNested(node) && IsEmpty(node);
+
+        private static bool IsNested(BlockSyntax node) =>
+            !AllowedContainerKinds.Contains(node.Parent.Kind());
```
andrei-epure-sonarsource, line 75: I wonder if it's more readable to use a switch expression here ([example](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEBMBGAsAKHQGYACLEgYRIG8CT6S6HiSBTGAVzBIBlWAXfqwBONAIJwAQnApwAInACicAGIBfJvU1lSwAPZ6ANiQCSAZzEAzIcIoAKPoJElDAShIBeAHzaGLkmYA7lD8AMYAFoz4fn600TEJjjYAdGJRCRkMeqJJIsmS6ZkZ2bwCKVTeJJYQhmascL4ZAPqeXiT8whysjfRqANwEGvhAA==)) - the values would be here

downside: not sure if it's less readable or not.  (`SyntaxKind.X or SyntaxKind.Y => false, _ => true`) vs `Contains` 
upside: getting rid of a data structure, more explicit

analyzers/src/SonarAnalyzer.CSharp/Rules/EmptyNestedBlock.cs
andrei-epure-sonarsource: fair point, although it's six items and you also need to replace the `.Contains` with a loop to do the equality checks

andrei-epure-sonarsource (approved):
LGTM, I left an optional comment

analyzers/src/SonarAnalyzer.CSharp/Rules/EmptyNestedBlock.cs
```diff
@@ -86,19 +43,41 @@ private static bool IsNested(BlockSyntax node)
...
+
+        private static bool ContainsCommentOrConditionalCompilation(BlockSyntax node) =>
+            ContainsCommentOrConditionalCompilation(node.OpenBraceToken.TrailingTrivia) || ContainsCommentOrConditionalCompilation(node.CloseBraceToken.LeadingTrivia);
+
+        private static bool ContainsCommentOrConditionalCompilation(SyntaxTriviaList trivias) =>
+            trivias.Any(trivia => trivia.IsAnyKind(SyntaxKind.SingleLineCommentTrivia, SyntaxKind.MultiLineCommentTrivia, SyntaxKind.DisabledTextTrivia));
```
andrei-epure-sonarsource, line 81: SyntaxKind.DisabledTextTrivia - nice!

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/EmptyNestedBlock.cs
```diff
@@ -90,7 +90,13 @@ public void ConditionalCompilation(object obj)
 #if DEBUG
                 Trace.WriteLine("message");
 #endif
-            } // Noncompliant@-4 FP: the block is not empty
+            }
```
andrei-epure-sonarsource, line 93: coming back, you could also add a test with an enabled processor directive, like at the beginning of the file
```
#define MY_VAL

....

#if MY_VAL
   Trace.Debug();
#endif  
```

and like this you also test when the variable is asserted (or maybe it's redundant?)

analyzers/src/SonarAnalyzer.CSharp/Rules/EmptyNestedBlock.cs
csaba-sagi-sonarsource: Switch expression is unfortunately unavailable for us for now. I changed the `HashSet `to an `Array `and used `IsAnyKind()`.

analyzers/src/SonarAnalyzer.CSharp/Rules/EmptyNestedBlock.cs
Corniel: Well, the hash of an enum, will not win for 6 items in an array. It will not harm either.

analyzers/src/SonarAnalyzer.CSharp/Rules/EmptyNestedBlock.cs
csaba-sagi-sonarsource: Sorry, I wanted to say that the or pattern is unavailable, not the switch expression.

PR #4648, opened 2021-07-10
Introduction of ComparisonKind

Corniel:
There are two rules involved for checking `Count`, `Length`, and `Count()`, on collections. Those are implemented for both languages and with a Fix for one on the way (#3857) the code duplication involved to do so, felt like a burden to me.

This is a second attempt (see #4033 for the first one), to fix this. Based on @pavel-mikula-sonarsource's feedback, I futher simplified the approach.

So what is in this pull request? A enum describing the The possible outcomes of (collection) count expressions: 
``` C#
public enum CountComparisonKind
{
    None,
    SizeDepedendent, // example: list.Count() > 3
    NotAny, // list.Count() ==0
    Any, // list.Count() > 0
    AlwaysFalse, // list.Count() < 0
    AlwaysTrue, // list.Count() >= 0
}
```
An extended version of a language independent `ComparisonKind`:
``` C#
public enum ComparisonKind
{
    None = 0,
    Equals,
    NotEquals,
    LessThan,
    LessThanOrEqual,
    GreaterThan,
    GreaterThanOrEqual,
}
```

And a way to get the `CountComparisonKind` based on the `ComparisonKind` and a count constant. Used in the rewritten S1155 rule, to detect if `.Any()` (or `!.Any()`) should be used:

``` C#
if (comparision.Count(constant).AnyOrNotAny()
    && TryGetCountCall(expression, c.SemanticModel, out var location, out var typeArgument))
{
    c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, location, typeArgument));
}
```

I did not put effort in S3981 yet, to apply it there to, but if you like this contribution, I'm willing to do that.


pavel-mikula-sonarsource (changes requested):
Hi @Corniel, thanks for giving it a clean try.

As usual, I've asked for quite a few simplifications here. Basically not to modify Symbolic Execution and not extract logic that is not shared between rules yet. It can be placed in the rule for now, or you can try to update the second rule to confirm that it's extracted correctly.

analyzers/src/SonarAnalyzer.Common/Helpers/ComparisonKind.cs
```diff
@@ -18,11 +18,16 @@
...
     public enum ComparisonKind
     {
-        Less,
-        LessOrEqual
+        None = 0,
+        Equals,
+        NotEquals,
+        LessThan,
+        LessThanOrEqual,
+        GreaterThan,
```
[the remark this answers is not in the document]
Corniel: You only want to undo the changes in CSharpExplodedGraph? (and keep the existing enum too).

What would be the best namespace for this enum then?

analyzers/src/SonarAnalyzer.Common/Helpers/ComparisonKinds.cs
```diff
@@ -0,0 +1,61 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+
+namespace SonarAnalyzer.Helpers
+{
+    public static class ComparisonKinds
```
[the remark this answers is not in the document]
Corniel: To me, that makes no sense. This is PR is preparation for more to come, but for keeping things small, it is split up.

analyzers/src/SonarAnalyzer.Common/Helpers/CountComparisonKinds.cs
```diff
@@ -0,0 +1,85 @@
...
+                ComparisonKind.GreaterThanOrEqual => GreaterThanOrEqual(count),
+                ComparisonKind.GreaterThan => GreaterThan(count),
+
+                ComparisonKind.LessThan => LessThan(count),
+                ComparisonKind.LessThanOrEqual => LessThanOrEqual(count),
+
+                _ => CountComparisonKind.None,
+            };
+
+        private static CountComparisonKind Equals(int count) =>
```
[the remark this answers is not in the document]
Corniel: Given the repetition, I thought this was the lesser of the evils to choose from. But if you think another one is better, be my guest.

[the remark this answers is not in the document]
Corniel: @pavel-mikula-sonarsource I think I processed all comments and/or gave a response.

analyzers/src/SonarAnalyzer.Common/Helpers/ComparisonKind.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: The current state looks good. We'll reimplement the SE this year so we might reconsider the SE enum then.

analyzers/src/SonarAnalyzer.Common/Helpers/ComparisonKinds.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: We're doing things incrementally. While I personally believe you'll do the other PR, something unexpected can happen.

But we can do it your way. I've created `Pavel/Corniel-Feature-Comparison` branch as a target for this PR. We'll merge this and the future PR there. And then I'll merge that feature branch to master with preserved commit history.

analyzers/src/SonarAnalyzer.VisualBasic/Facade/VisualBasicSyntaxFacade.cs
```diff
@@ -33,6 +33,18 @@ internal sealed class VisualBasicSyntaxFacade : SyntaxFacade<SyntaxKind>
 
         public override SyntaxKind Kind(SyntaxNode node) => node.Kind();
 
+        public override ComparisonKind ComparisonKind(SyntaxNode node) =>
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: This PR targets a feature branch now, so we can keep it. It will be used in S3981

analyzers/src/SonarAnalyzer.Common/Helpers/CountComparisonKinds.cs
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+namespace SonarAnalyzer.Helpers
+{
+    public static class CountComparisonKinds
+    {
+        public static bool AnyOrNotAny(this CountComparisonKind comparison) =>
+            comparison == CountComparisonKind.Any || comparison == CountComparisonKind.NotAny;
+
+        public static CountComparisonKind Count(this ComparisonKind comparison, int count) =>
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: This one wasn't addressed. The `Count` is confusing on usage side.

analyzers/src/SonarAnalyzer.Common/Helpers/CountComparisonKinds.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Let's postpone the decision. I'll keep this open for now and I'll take a look at this problem later, when the rest settles down.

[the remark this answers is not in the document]
Corniel: I agree.

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
```diff
@@ -33,7 +33,7 @@
 using SonarAnalyzer.SymbolicExecution.Constraints;
 using SonarAnalyzer.SymbolicExecution.SymbolicValues;
 using StyleCop.Analyzers.Lightup;
-
+using ComparisonKind = SonarAnalyzer.SymbolicExecution.SymbolicValues.ComparisonKind;
```
pavel-mikula-sonarsource, line 36: _Nitpick:_ Missing EOL after this `using` statement.
```suggestion
using ComparisonKind = SonarAnalyzer.SymbolicExecution.SymbolicValues.ComparisonKind;

```

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/ComparisonKind.cs
```diff
@@ -23,6 +23,6 @@ namespace SonarAnalyzer.SymbolicExecution.SymbolicValues
     public enum ComparisonKind
     {
         Less,
-        LessOrEqual
+        LessOrEqual,
```
pavel-mikula-sonarsource, line 26: This `ComparisonKind` is language independent => can you move it to `SymbolicExecution/Common/SymbolicValues`? It was originally in `../SymbolicValues/Binary`

pavel-mikula-sonarsource: I would be also fine with renaming it to `SymbolicComparisonKind` to avoid the collision we have. Its values are reduced from some symbolic execution magic.

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/SymbolicValues/Binary/ComparisonSymbolicValue.cs
```diff
@@ -20,6 +20,7 @@
 
 using System.Collections.Generic;
 using System.Linq;
+using SonarAnalyzer.Helpers;
```
pavel-mikula-sonarsource, line 23: Not needed after latest changes.
```suggestion
```

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/Relationships/ComparisonRelationship.cs
```diff
@@ -21,7 +21,8 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using SonarAnalyzer.SymbolicExecution.SymbolicValues;
+using SonarAnalyzer.Helpers;
```
pavel-mikula-sonarsource, line 24: Not needed after latest changes.
```suggestion
```

pavel-mikula-sonarsource (changes requested):
Final polishing of Symbolic Execution part.

analyzers/src/SonarAnalyzer.Common/Extensions/CountComparisonResultExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarAnalyzer.Helpers
+{
+    public static class CountComparisonResultExtensions
+    {
+        public static bool EmptyOrNotEmpty(this CountComparisonResult comparison) =>
```
pavel-mikula-sonarsource, line 25:
```suggestion
        public static bool IsEmptyOrNotEmpty(this CountComparisonResult comparison) =>
```

analyzers/src/SonarAnalyzer.Common/Extensions/ComparisonExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+
+using System;
+using Microsoft.CodeAnalysis;
+using Comparison = SonarAnalyzer.Helpers.ComparisonKind;
+namespace SonarAnalyzer.Helpers
+{
+    public static class ComparisonKindExtensions
+    {
+        public static Comparison ComparisonKind(this IMethodSymbol method) =>
+            method?.Name switch
```
pavel-mikula-sonarsource, line 29: Let's check for `MethodKind == MethodKind.UserDefinedOperator` first. Some joker can declare a method like `public void op_LessThan() { }` that is not a comparison operator.

analyzers/src/SonarAnalyzer.Common/Helpers/ComparisonKinds.cs
```diff
@@ -0,0 +1,37 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+
+namespace SonarAnalyzer.Helpers
+{
+    public static class ComparisonKinds
+    {
+        public static IReadOnlyCollection<ComparisonKind> All => new[]
```
pavel-mikula-sonarsource, line 27: I don't see where this will be used in S3981. I'd remove this file. Same collection can be generated with this and stored to a static field in the rule itself.
```
System.Enum.GetValues(typeof(ComparisonKind)).Cast<ComparisonKind>().Where(x => x != ComparisonKind.None)
```

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessChecking.cs
```diff
@@ -36,115 +36,70 @@ public sealed class CollectionEmptinessChecking : SonarDiagnosticAnalyzer
         internal const string DiagnosticId = "S1155";
         private const string MessageFormat = "Use '.Any()' to test whether this 'IEnumerable<{0}>' is empty or not.";
 
-        private static readonly CSharpExpressionNumericConverter ExpressionNumericConverter = new CSharpExpressionNumericConverter();
-
         private static readonly DiagnosticDescriptor rule =
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
+        private static readonly CSharpExpressionNumericConverter ExpressionNumericConverter = new CSharpExpressionNumericConverter();
```
pavel-mikula-sonarsource, line 42: Line 39: You can rename it to `Rule` to fix the codesmell and make it a single line. Then the empty line between those two `private static readonly` fields will not be needed.

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessChecking.cs
```diff
@@ -36,115 +36,70 @@ public sealed class CollectionEmptinessChecking : SonarDiagnosticAnalyzer
...
-                SyntaxKind.LessThanOrEqualExpression);
-            context.RegisterSyntaxNodeActionInNonGenerated(
-                c =>
-                {
-                    var binary = (BinaryExpressionSyntax)c.Node;
-                    CheckCountZero(binary.Left, binary.Right, c);
-                    CheckCountZero(binary.Right, binary.Left, c);
-                },
-                SyntaxKind.EqualsExpression);
+        protected override void Initialize(SonarAnalysisContext context) =>
```
pavel-mikula-sonarsource, line 46: Line 36: `DiagnosticId` can be private.

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessChecking.cs
```diff
@@ -36,115 +36,70 @@ public sealed class CollectionEmptinessChecking : SonarDiagnosticAnalyzer
...
             var invocation = expression as InvocationExpressionSyntax;
-            if (!(invocation?.Expression is MemberAccessExpressionSyntax memberAccess))
-            {
-                return false;
-            }
 
-            if (!(semanticModel.GetSymbolInfo(memberAccess).Symbol is IMethodSymbol methodSymbol) ||
-                !IsMethodCountExtension(methodSymbol) ||
-                !methodSymbol.IsExtensionOn(KnownType.System_Collections_Generic_IEnumerable_T))
+            if ((invocation?.Expression is MemberAccessExpressionSyntax memberAccess)
```
pavel-mikula-sonarsource, line 81: Unneeded parenthesis
```suggestion
            if (invocation?.Expression is MemberAccessExpressionSyntax memberAccess
```

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessChecking.cs
```diff
@@ -36,115 +36,70 @@ public sealed class CollectionEmptinessChecking : SonarDiagnosticAnalyzer
...
+                && (semanticModel.GetSymbolInfo(memberAccess).Symbol is IMethodSymbol methodSymbol)
```
pavel-mikula-sonarsource, line 82: Let's do a syntax-based name check first for performance reasons. Querying semantic model is expensive.
```suggestion
				&& memberAccess.Name.Identifier.ValueText == nameof(Enumerable.Count)
                && (semanticModel.GetSymbolInfo(memberAccess).Symbol is IMethodSymbol methodSymbol)
```

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessChecking.cs
```diff
@@ -36,115 +36,70 @@ public sealed class CollectionEmptinessChecking : SonarDiagnosticAnalyzer
...
+                && IsMethodCountExtension(methodSymbol)
+                && methodSymbol.IsExtensionOn(KnownType.System_Collections_Generic_IEnumerable_T))
             {
-                return false;
-            }
+                if (methodSymbol.IsGenericMethod)
+                {
+                    typeArgument = methodSymbol.TypeArguments.First().ToDisplayString();
```
pavel-mikula-sonarsource, line 88: There should be just one.
```suggestion
                    typeArgument = methodSymbol.TypeArguments.Single().ToDisplayString();
```

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessChecking.cs
```diff
@@ -36,115 +36,70 @@ public sealed class CollectionEmptinessChecking : SonarDiagnosticAnalyzer
...
             context.RegisterSyntaxNodeActionInNonGenerated(
                 c =>
                 {
                     var binary = (BinaryExpressionSyntax)c.Node;
-                    CheckCountZero(binary.Left, binary.Right, c);
-                    CheckCountZero(binary.Right, binary.Left, c);
+                    var left = ExpressionNumericConverter.TryGetConstantIntValue(binary.Left, out var l_out) ? (int?)l_out : null;
+                    var right = ExpressionNumericConverter.TryGetConstantIntValue(binary.Right, out var r_out) ? (int?)r_out : null;
+
+                    if ((left ?? right) is int constant)
```
pavel-mikula-sonarsource, line 54: Can we pre-filter expected constant values to save some code execution on a quick-win condition? 
```suggestion
                    if ((left ?? right) is int constant && (constant == 0 || constant == 1))
```

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessChecking.cs
```diff
@@ -36,115 +36,70 @@ public sealed class CollectionEmptinessChecking : SonarDiagnosticAnalyzer
...
-                },
-                SyntaxKind.EqualsExpression);
+        protected override void Initialize(SonarAnalysisContext context) =>
             context.RegisterSyntaxNodeActionInNonGenerated(
                 c =>
                 {
                     var binary = (BinaryExpressionSyntax)c.Node;
-                    CheckCountZero(binary.Left, binary.Right, c);
-                    CheckCountZero(binary.Right, binary.Left, c);
+                    var left = ExpressionNumericConverter.TryGetConstantIntValue(binary.Left, out var l_out) ? (int?)l_out : null;
```
pavel-mikula-sonarsource, line 51: If `left != null` we don't need to evaluate `TryGetConstantIntValue` for `right`. Let's move the logic to a parametrized method
```
        private static bool CheckExpression(SyntaxNodeAnalysisContext c, ExpressionSyntax expression, bool mirror)
        {
            if (ExpressionNumericConverter.TryGetConstantIntValue(expression, out var value))
            {
                var comparison = CSharpFacade.Instance.Syntax.ComparisonKind(expression.Parent);
                if (mirror)
                {
                    comparison = comparison.Mirror();
                }
                if (comparison.Count(value).EmptyOrNotEmpty() && TryGetCountCall(expression, c.SemanticModel, out var location, out var typeArgument))
                {
                    c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, location, typeArgument));
                }
                return true;
            }
            else
            {
                return false;
            }
        }
```
and invoke it like this
```
                    if (!CheckExpression(c, binary.Left, true))
                    {
                        CheckExpression(c, binary.Right, false);
                    }
```
I didn't try it, there might be some minor fixes needed.

pavel-mikula-sonarsource: We'll probably need something like `countExpression` / `valueExpression` arguments to make this work because my simple refactoring just uses one of the binary nodes needed.

analyzers/src/SonarAnalyzer.Common/Extensions/CountComparisonResultExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+                ComparisonKind.Equals => Equals(count),
+                ComparisonKind.NotEquals => NotEquals(count),
+
+                ComparisonKind.GreaterThanOrEqual => GreaterThanOrEqual(count),
+                ComparisonKind.GreaterThan => GreaterThan(count),
+
+                ComparisonKind.LessThan => LessThan(count),
+                ComparisonKind.LessThanOrEqual => LessThanOrEqual(count),
+
+                _ => CountComparisonResult.None,
```
pavel-mikula-sonarsource, line 40: The handler name is always the same as the enum member = regular, easy to follow. We don't need those empty lines as the logic doesn't differ.
```suggestion
                ComparisonKind.Equals => Equals(count),
                ComparisonKind.NotEquals => NotEquals(count),
                ComparisonKind.GreaterThanOrEqual => GreaterThanOrEqual(count),
                ComparisonKind.GreaterThan => GreaterThan(count),
                ComparisonKind.LessThan => LessThan(count),
                ComparisonKind.LessThanOrEqual => LessThanOrEqual(count),
                _ => CountComparisonResult.None,
```

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/ComparisonKind.cs
Corniel: I applied the renaming.

For the movement:
- [ ] SonarAnalyzer.Common\SymbolicExecution\SymbolicValues
- [ ] SonarAnalyzer.CSharp\SymbolicExecution\Common\SymbolicValues
?

pavel-mikula-sonarsource (changes requested):
The rest of the logic

analyzers/src/SonarAnalyzer.Common/Helpers/CountComparisonKinds.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: The API seems stable after the review I just done. To deal with this, let's introduce this
```
        private static CountComparisonResult Check(int count, CountComparisonResult smallerThanLimit, CountComparisonResult sameAsLimit, CountComparisonResult higherThanLimit, int limit = 0)
        {
            if(count < limit)
            {
                return smallerThanLimit;
            }
            else if(count > limit)
            {
                return higherThanLimit;
            }
            else
            {
                return sameAsLimit;
            }
        }
```
and remove all the intermediate methods. This is basically the pattern that all the intermediates do. I'm intentionally avoiding `lessThan`, `equals` and `greaterThan` names here to avoid confusion with the enum.

analyzers/src/SonarAnalyzer.Common/Helpers/CountComparisionResult.cs
```diff
@@ -0,0 +1,32 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarAnalyzer.Helpers
+{
+    public enum CountComparisonResult
+    {
+        None,
+        SizeDepedendent,
```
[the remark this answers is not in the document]
Corniel: I think we do. `None` intended as indication for syntax that is not a count comparison, where `SizeDependent` is intended for expressions where we can not resolve the outcome of the expression runtime. I think that merging these two values will make the code harder to understand.

analyzers/src/SonarAnalyzer.Common/Extensions/ComparisonExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+           comparison switch
+           {
+               Comparison.GreaterThan => Comparison.LessThan,
+               Comparison.GreaterThanOrEqual => Comparison.LessThanOrEqual,
+               Comparison.LessThan => Comparison.GreaterThan,
+               Comparison.LessThanOrEqual => Comparison.GreaterThanOrEqual,
+               _ => comparison,
+           };
+
+        public static string CSharp(this Comparison kind) =>
```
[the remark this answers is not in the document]
Corniel: It was also used (temporary) by the (Symbolic)ComparisonKind. That's why I moved it. Can make it private again.

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessChecking.cs
Corniel: Technically we can. But I would argue that the case will not be easier to understand, and is not worth the (premature) optimization.

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/ComparisonKind.cs
pavel-mikula-sonarsource: This is currently messy because we don't have VB.NET SE (yet). Let's move the file to `SonarAnalyzer.CSharp\SymbolicExecution\Common\SymbolicValues` as there's the rest of the logic to stay focused in this PR.

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/Relationships/ComparisonRelationship.cs
pavel-mikula-sonarsource: This one wasn't addressed.

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/CSharp/CSharpExplodedGraph.cs
pavel-mikula-sonarsource: The `using` is gone, but the EOL is missing there - unintentionally removed by this PR.

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/Relationships/ComparisonRelationship.cs
```diff
@@ -180,45 +180,24 @@ private BinaryRelationship GetTransitiveRelationship(EqualsRelationship other)
...
-        public bool Equals(ComparisonRelationship other)
-        {
-            if (other == null ||
-                ComparisonKind != other.ComparisonKind)
-            {
-                return false;
-            }
+        public override bool Equals(object obj) => obj is null
+            ? false
+            : Equals(obj as ComparisonRelationship);
```
pavel-mikula-sonarsource, line 191: Can be improved
```suggestion
        public override bool Equals(object obj) => 
        	obj != null && Equals(obj as ComparisonRelationship);
```

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Common/Relationships/ComparisonRelationship.cs
```diff
@@ -180,45 +180,24 @@ private BinaryRelationship GetTransitiveRelationship(EqualsRelationship other)
...
 
-            return base.Equals(other);
-        }
+        public bool Equals(ComparisonRelationship other) => other == null ||
+                ComparisonKind != other.ComparisonKind
+                ? false
+                : base.Equals(other);
```
pavel-mikula-sonarsource, line 196: Can be simplified
```suggestion
        public bool Equals(ComparisonRelationship other) => 
            other != null && ComparisonKind == other.ComparisonKind && base.Equals(other);
```

pavel-mikula-sonarsource (changes requested):
Intermediate refactoring details

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessChecking.cs
Corniel: I did a bit different, but with the result that I do not check right, once left has a value.

analyzers/src/SonarAnalyzer.Common/Extensions/ComparisonExtensions.cs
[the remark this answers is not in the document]
Corniel: Currently not used. I removed it, and will reintroduce it when needed:

``` C#
        public static string CSharp(this ComparisonKind kind) =>
            kind switch
            {
                ComparisonKind.Equals => "==",
                ComparisonKind.NotEquals => "!=",
                ComparisonKind.LessThan => "<",
                ComparisonKind.LessThanOrEqual => "<=",
                ComparisonKind.GreaterThan => ">",
                ComparisonKind.GreaterThanOrEqual => ">=",
                _ => throw new InvalidOperationException(),
            };
```

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessChecking.cs
pavel-mikula-sonarsource: I'd still prefer the parametrized version. We have 3 times the same condition and dual processing.

Corniel: @pavel-mikula-sonarsource I'd further improved it. I think (and hope) you will like this one.

pavel-mikula-sonarsource: Yep, nice

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessChecking.cs
```diff
@@ -36,81 +36,45 @@ public sealed class CollectionEmptinessChecking : SonarDiagnosticAnalyzer
...
-            {
-                context.ReportDiagnosticWhenActive(Diagnostic.Create(rule, reportLocation, typeArgument));
-            }
-        }
-        private static void CheckCountOne(ExpressionSyntax one, ExpressionSyntax count, SyntaxNodeAnalysisContext context)
+        private static void CheckExpression(
+            SyntaxNodeAnalysisContext context,
+            ExpressionSyntax expression,
+            int constant,
+            ComparisonKind comparison)
```
pavel-mikula-sonarsource, line 72: Should be a single line, it doesn't violate S103
```suggestion
        private static void CheckExpression(SyntaxNodeAnalysisContext context, ExpressionSyntax expression, int constant, ComparisonKind comparison)
```

analyzers/src/SonarAnalyzer.Common/Helpers/CountComparisonKinds.cs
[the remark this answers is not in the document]
Corniel: I moved the limit parameter as It made the using code easier to read. I'm not really satisfied with this `Check` function, but the original was obviously not nice either, so, it's okay.

pavel-mikula-sonarsource (changes requested):
Last nitpicks.

There are also 5 unresolved comments above, most of them are hidden under the `X hidden conversations`.. GitHub button

analyzers/src/SonarAnalyzer.Common/Helpers/ComparisonKinds.cs
Corniel: Updated.

analyzers/src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
```diff
@@ -143,12 +115,17 @@ private static IEnumerable<string> GetMembersToOverride(IEnumerable<IMethodSymbo
...
 
-            foreach (var op in RequiredOperators.Except(overridenOperators))
+            foreach (var comparisonKind in ComparisonKinds.Except(overridenOperators))
             {
-                yield return OperatorNamesMap[op];
+                yield return comparisonKind.CSharp();
             }
         }
+
+        private static readonly ComparisonKind[] ComparisonKinds =
```
pavel-mikula-sonarsource, line 126: Please move this static field to the top of the class to follow our coding style - before or after the other one `ComparableInterfaces`

pavel-mikula-sonarsource (changes requested):
There's one last unresolved question above from the previous round and here's one more coding style detail.

analyzers/src/SonarAnalyzer.CSharp/Extensions/ComparisonExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+using System;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Extensions
+{
+    public static class ComparisonKindExtensions
+    {
+        public static string CSharp(this ComparisonKind kind) =>
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Hi @Corniel, can you please take a look at this one too? It's the last unresolved thread.

Corniel:
@pavel-mikula-sonarsource 

Can you fix this?

```
 "issues":  [
 {
 "id":  "S1210",
-"message":  "When implementing IComparable<T>, you should also override <, >, <=, >=, ==, !=.",
+"message":  "When implementing IComparable<T>, you should also override ==, !=, <, <=, >, >=.",
 "location":  {
 "uri":  "sources\akka.net\src\core\Akka.MultiNodeTestRunner.Shared\Reporting\MultiNodeMessage.cs",
 "region":  {
```

pavel-mikula-sonarsource:
Hi @Corniel, can you please rebase this PR on the `Pavel/Corniel-Feature-Comparison` branch to fix the failing UTs?

The jokers from FluentAssertions released version 6.0 on Friday 13th with some breaking changes and additional dependencies that broke our UTs. We'll need to rebase this PR to get the fix that was already on master.

Corniel:
> Hi @Corniel, can you please rebase this PR on the `Pavel/Corniel-Feature-Comparison` branch to fix the failing UTs?

Done.

> The jokers from FluentAssertions released version 6.0 on Friday 13th with some breaking changes (..)

That is basically asking for it. :)

pavel-mikula-sonarsource:
> Can you fix this?
> 
> ```
>  "issues":  [
>  {
>  "id":  "S1210",
> -"message":  "When implementing IComparable<T>, you should also override <, >, <=, >=, ==, !=.",
> +"message":  "When implementing IComparable<T>, you should also override ==, !=, <, <=, >, >=.",
>  "location":  {
>  "uri":  "sources\akka.net\src\core\Akka.MultiNodeTestRunner.Shared\Reporting\MultiNodeMessage.cs",
>  "region":  {
> ```

We have this UT on line 909:
```
public class NonCompliant : IComparable, IComparable<NonCompliant> // Noncompliant {{When implementing IComparable or IComparable<T>, you should also override Equals, <, >, <=, >=, ==, !=.}}
```
with the order that is removed from the IT. That looks like we have non-deterministic order of the operators in the message now. Can you `.OrderBy(x => x)` the `ComparisonKinds`?

We'll need to update the expected message in UTs and ITs, but it should give us nicer message than before (starting with ==, != where possible).

pavel-mikula-sonarsource:
Did you run the UTs after the last change? I expect at least this one to fail, because it will need to update the message:
https://github.com/SonarSource/sonar-dotnet/blob/ff59a43362200895fae0edc5e8c9f6f53aae0a14/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ComparableInterfaceImplementation.cs#L909

Corniel:
> Did you run the UTs after the last change? I expect at least this one to fail, because it will need to update the message:

Now I have. #shame

pavel-mikula-sonarsource:
@Corniel I'll prepare a commit to fix the ITs for you. 

Can you please take a look at the codesmells (except the long lines - those were there and are fine).

Coverage is also low because of some refactorings - we don't need to fix that. But can you please try to create test case in the rule to cover this line? 
https://sonarcloud.io/code?id=sonaranalyzer-dotnet&pullRequest=4648&selected=sonaranalyzer-dotnet%3Aanalyzers%2Fsrc%2FSonarAnalyzer.Common%2FExtensions%2FCountComparisonResultExtensions.cs&line=80

pavel-mikula-sonarsource:
You can cherry-pick [this commit](https://github.com/SonarSource/sonar-dotnet/commit/3bbdee6f1a3a855149233e8a5d02b1489a86e630) from https://github.com/SonarSource/sonar-dotnet/tree/Pavel/Corniel-ITs-CherryPick branch to deal with the ITs

Corniel:
> Can you please take a look at the codesmells (except the long lines - those were there and are fine).

Done. I alos did the long lines. It turned out, that where 4 calls to the private method, all requiring the cast, so I moved the cast to the method.

> Coverage is also low because of some refactorings - we don't need to fix that. But can you please try to create test case in the rule to cover this line?
Done.

pavel-mikula-sonarsource (approved):
LGTM

PR #4657, opened 2021-07-13
S3776 init-only setters support

csaba-sagi-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CognitiveComplexity.CSharp9.cs
```diff
@@ -37,14 +37,18 @@ record MyRecord
...
             if (true)
+//          ^^ Secondary {{+1}}
             {
                 field = "";
             }
         }
     }
+
+    bool Prop => field == null || field is { Length: 5 }; // Noncompliant {{Refactor this property to reduce its Cognitive Complexity from 1 to the 0 allowed.}}
+// Secondary@-1 {{+1}}
```
dmitry-golovinov-sonarsource, line 51: It would look nicer if that would be aligned with the comment above.

dmitry-golovinov-sonarsource (approved):
LGTM! Please check my comment

PR #4660, opened 2021-07-13
Support C# 9 init-only setters in Symbolic Execution

pavel-mikula-sonarsource:


csaba-sagi-sonarsource (approved):
LGTM! One single registration and a bunch of FNs are gone!

PR #4662, opened 2021-07-13
Update 3rd party licenses and include in packages

andrei-epure-sonarsource:

### DIFF
--- a/THIRD-PARTY-NOTICES.txt
+++ b/THIRD-PARTY-NOTICES.txt
@@ -5,22 +5,35 @@ In the event that we accidentally failed to list a required notice, please bring
 This project uses other open source projects, which are used under the terms of the following license(s).
 --------------------------------------------------
 
-StyleCopAnalyzers - Lightup
+License notice for StyleCopAnalyzers - Lightup
+-----------------------------------------------------
 
-    ﻿Copyright (c) Tunnel Vision Laboratories, LLC. All rights reserved.
+    MIT License
 
-    Licensed under the Apache License, Version 2.0 (the "License"); you may not use
-    these files except in compliance with the License. You may obtain a copy of the
-    License at
+    Copyright (c) Tunnel Vision Laboratories, LLC
 
-    http://www.apache.org/licenses/LICENSE-2.0
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"), to deal
+    in the Software without restriction, including without limitation the rights
+    to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+    copies of the Software, and to permit persons to whom the Software is
+    furnished to do so, subject to the following conditions:
 
-    Unless required by applicable law or agreed to in writing, software distributed
-    under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
-    CONDITIONS OF ANY KIND, either express or implied. See the License for the
-    specific language governing permissions and limitations under the License.
+    The above copyright notice and this permission notice shall be included in all
+    copies or substantial portions of the Software.
 
-.NET Compiler Platform ("Roslyn")
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+    SOFTWARE.
+
+    https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/LICENSE
+
+License notice for .NET Compiler Platform ("Roslyn")
+-----------------------------------------------------
 
     Copyright Microsoft.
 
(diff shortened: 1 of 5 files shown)


THIRD-PARTY-NOTICES.txt
```diff
@@ -5,22 +5,35 @@ In the event that we accidentally failed to list a required notice, please bring
 This project uses other open source projects, which are used under the terms of the following license(s).
 --------------------------------------------------
 
-StyleCopAnalyzers - Lightup
+License notice for StyleCopAnalyzers - Lightup
+-----------------------------------------------------
```
dmitry-golovinov-sonarsource, line 9: Feel free to disregard. What if we would align the length of those separators throughout the file (see the line number 6)

analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.CSharp.nuspec
```diff
@@ -29,5 +29,6 @@
     <file src="bin\Release\Newtonsoft.Json.dll" target="analyzers" />
     <file src="tools-cs\*.ps1" target="tools\" />
     <file src="logos\sonarsource_64.png" target="images\" />
+    <file src="..\..\..\THIRD-PARTY-NOTICES.txt" target="license\" />
```
dmitry-golovinov-sonarsource, line 32: I may be doing something very wrong here, but if I extract resulting vsix file with 7-zip it has logos directory, but there is no license folder inside.

analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.VisualBasic.nuspec
```diff
@@ -27,5 +27,6 @@
     <file src="bin\Release\Google.Protobuf.dll" target="analyzers" />
     <file src="tools-vbnet\*.ps1" target="tools\" />
     <file src="logos\sonarsource_64.png" target="images\" />
+    <file src="..\..\..\THIRD-PARTY-NOTICES.txt" target="license\" />
```
dmitry-golovinov-sonarsource, line 30: I may be doing something very wrong here, but if I extract resulting vsix file with 7-zip it has logos directory, but there is no license folder inside.

sonar-csharp-plugin/pom.xml
```diff
@@ -252,6 +252,9 @@
                         <include name="Google.Protobuf.dll"/>
                         <include name="Newtonsoft.Json.dll"/>
                       </fileset>
+                      <fileset dir="${project.build.directory}/../..">
+                        <include name="THIRD-PARTY-NOTICES.txt"/>
```
dmitry-golovinov-sonarsource, line 256: File is missing on my machine in resulting jar file.

sonar-vbnet-plugin/pom.xml
```diff
@@ -253,6 +253,9 @@
                         <include name="SonarAnalyzer.VisualBasic.dll"/>
                         <include name="Google.Protobuf.dll"/>
                       </fileset>
+                      <fileset dir="${project.build.directory}/../..">
+                        <include name="THIRD-PARTY-NOTICES.txt"/>
```
dmitry-golovinov-sonarsource, line 257: File is missing on my machine in resulting jar file.

dmitry-golovinov-sonarsource (changes requested):
Please recheck your PR. On my machine I cannot see this file in resulting artifacts.

analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.CSharp.nuspec
andrei-epure-sonarsource: The VSIX is just for debugging during development and it's not relevant.

This is used for the nuget packages we eventually publish on Nuget.org when releasing, and the nugets do have the `license` folder inside - you can download the artifact from : https://dev.azure.com/sonarsource/DotNetTeam%20Project/_build/results?buildId=34588&view=artifacts&pathAsName=false&type=publishedArtifacts

analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.VisualBasic.nuspec
andrei-epure-sonarsource: replied in https://github.com/SonarSource/sonar-dotnet/pull/4662#discussion_r669443997

sonar-csharp-plugin/pom.xml
andrei-epure-sonarsource: you can download the JAR for this PR from [burgr](https://burgr.sonarsource.com/projects/SonarSource/sonar-dotnet/pullrequests/4662) - and direct link [on repox](https://repox.jfrog.io/repox/sonarsource/org/sonarsource/dotnet/sonar-csharp-plugin/8.27.0.34588/sonar-csharp-plugin-8.27.0.34588.jar)

I am embedding the license in the ZIP in `static\SonarAnalyzer-8.27.0.34588.zip` - it's together with the dotnet assemblies 

I will investigate how the maven licenses are generated, however I believe that it's good enough to embed them like this for now.

sonar-vbnet-plugin/pom.xml
andrei-epure-sonarsource: replied in https://github.com/SonarSource/sonar-dotnet/pull/4662#discussion_r669447541

analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.CSharp.nuspec
andrei-epure-sonarsource: to generate the nuget locally you can use `nuget pack .\analyzers\src\SonarAnalyzer.Vsix\SonarAnalyzer.CSharp.nuspec` and it will get generated where you run the command from

andrei-epure-sonarsource: I know it's confusing that the nuspec files which are used in our release process are present in the VSIX folder which is just for debugging, we may want to move them to their own folder as they're important

dmitry-golovinov-sonarsource (approved):
LGTM! Thanks for the explanations you made.

PR #4663, opened 2021-07-13
CSharpExecutableLinesMetric: Init-only setter and records support

csaba-sagi-sonarsource:


andrei-epure-sonarsource (changes requested):
the changes LGTM.

Now that you touched this part of the code, could you please update the tests with other C# 8 and C# 9 syntax such as:
- based on [SQ docs](https://docs.sonarqube.org/latest/extend/executable-lines/) - see "Location" section, please add a test with the below (well, it's not C#8 or C#9 but it's in the docs so we should have it in the tests)
```
for         // +1
  (         // +0
   int i=0; // +0
   i < 10;  // +0
   i++      // +0
  )         // +0
{           // +0
}
```

- more conditionals like Null-coalescing assignment and switch expressions (should be one executable line for the full expression, like for switch in the `Conditionals` test (line 108)
-- for switch expression, you could add two cases [like here](https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzguATEBqAPgAQCYBGAWACh8BmAAiJoGEaBvCm9u2/AFhoFkAFPmIAGGigCULNh1kA3CACcaADxoBeceIDuUBAGMAFjNntW5U5dMAiCNY0A+GgkUBXAKZwTVy9eD31JwAzCAAbFE9vH1kAfUdnN3cojgBfAG5k9gVlAE8NLRRdA2MLK3Noq1sApwAxAHs6gT9rCS9Sitlm+PrGqtbM6LjAmhDwpPbTdO8U72AG0JoeoVFxKWGXDwzyFKA==)
- interpolated string on multiple lines (should be one single executable line)
- a complex pattern matching expression on multiple lines


I suggest the above to document the existing behavior. Based on what you see, we may need to make changes (a different PR).

csaba-sagi-sonarsource:
I created two issues for the CSharpExecutableLinesMetric:
https://github.com/SonarSource/sonar-dotnet/issues/4688
https://github.com/SonarSource/sonar-dotnet/issues/4689

analyzers/tests/SonarAnalyzer.UnitTest/Metrics/CSharpExecutableLinesMetricTest.cs
```diff
@@ -627,6 +627,205 @@ public void Foo(bool flag)
...
+                ",
+              5);
+
+        [TestMethod]
+        public void UsingDeclaration() =>
+            AssertLinesOfCode(
+              @"
+                        void Foo(int? i, string s)
+                        {
+                            using var file = new System.IO.StreamWriter(""WriteLines2.txt""); // FN
```
andrei-epure-sonarsource, line 711: You could rephrase  #4689 to mention all what is not covered (not only expression bodied methods, but also using statements etc).

analyzers/tests/SonarAnalyzer.UnitTest/Metrics/CSharpExecutableLinesMetricTest.cs
```diff
@@ -627,6 +627,205 @@ public void Foo(bool flag)
...
+                                ""b"" => Foo(""a""), // +1
+                                _ => false // FN
+                            };
+                        }
+                        bool Foo(string s) => true;
+                ",
+              12, 13);
+
+        [TestMethod]
+        public void MultiLineInterpolatedString() =>
```
andrei-epure-sonarsource, line 691: could you please also add a test with a multi-line interpolated string which has some complex expressions on each line e.g.

```
public class C {
    public string M(int i) {
        return @$"
        {(i == 1 ? Bar(1) : Bar(2))}
        {(i == 2 ? Bar(2) : Bar(3))}
        {Bar(5)}
            ";
    }
    
    public string Bar(int i) => "y" + i;
}
```

The interpolated string is in a single statement so it should be counted only once

> Regardless of the number of lines across which nested statements are spread, the executable line count should only be incremented by one, since typically the execution of one naturally follows from the other.

andrei-epure-sonarsource (approved):
LGTM, I've left some comments, although most are observations.

I opened https://github.com/SonarSource/sonar-dotnet/issues/4691 for line 87 (we shouldn't count empty statement line `;` )

To sum up, we have more uncovered cases:
- expression-bodied methods
- using statements
- variable declarations with initialization `var x = "foo";`

csaba-sagi-sonarsource:
I applied your suggestions and modified also the issue to cover other syntax kinds as well.

analyzers/tests/SonarAnalyzer.UnitTest/Metrics/CSharpExecutableLinesMetricTest.cs
```diff
@@ -826,6 +847,23 @@ public static bool IsLetter(char c)
...
+        [TestMethod]
+        public void MultiLineInvocation() =>
+            AssertLinesOfCode(
+                @"
+                        using System;
+                        using System.Collections.Generic;
+
+                            public static bool Foo(int a, int b)
+                            {
+                               return Foo(1,
```
andrei-epure-sonarsource, line 859: basically, this means we don't respect the existing docs.

csaba-sagi-sonarsource: Yes, I would merge this PR if you agree and maybe tell Tom tomorrow to take a look on the issue to get some feedback about priorities.

PR #4665, opened 2021-07-13
S2292: Init-only setter support

csaba-sagi-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -35,56 +35,52 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class PropertyToAutoProperty : SonarDiagnosticAnalyzer
     {
-        internal const string DiagnosticId = "S2292";
+        private const string DiagnosticId = "S2292";
         private const string MessageFormat = "Make this an auto-implemented property and remove its backing field.";
+        private const int ReadWriteAccessorCount = 2;
```
pavel-mikula-sonarsource, line 40: Can be const inside that single method that uses it. Or just accept the magic number, as  `Count` is really easy to understand.

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -94,27 +90,21 @@ private static bool HasDifferentModifiers(SyntaxList<AccessorDeclarationSyntax>
...
-                parameter.Name == "value" &&
-                parameter.IsImplicitlyDeclared)
+            if (assignment != null
+                && assignment.IsKind(SyntaxKind.SimpleAssignmentExpression)
+                && assignment.Right != null
+                && semanticModel.GetSymbolInfo(assignment.Right).Symbol is IParameterSymbol {Name: "value", IsImplicitlyDeclared: true})
             {
-                return TryGetField(assignment.Left, semanticModel.GetDeclaredSymbol(setter).ContainingType,
-                    semanticModel, out setterField);
+                return TryGetField(assignment.Left, semanticModel.GetDeclaredSymbol(setter).ContainingType, semanticModel, out setterField);
```
pavel-mikula-sonarsource, line 107: _Line 112 below:_ There's `&&` refactoring refugee. That condition could be single line.

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -139,8 +129,8 @@ private static bool TryGetFieldFromSetter(AccessorDeclarationSyntax setter, Sema
                 return field != null;
             }
 
-            if (!(expression is MemberAccessExpressionSyntax memberAccess) ||
-                !memberAccess.IsKind(SyntaxKind.SimpleMemberAccessExpression))
+            if (!(expression is MemberAccessExpressionSyntax memberAccess)
```
pavel-mikula-sonarsource, line 132: Let's make this whole method `if/else if/else if/else` for faster readability.

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -139,8 +129,8 @@ private static bool TryGetFieldFromSetter(AccessorDeclarationSyntax setter, Sema
...
+                || !memberAccess.IsKind(SyntaxKind.SimpleMemberAccessExpression))
```
pavel-mikula-sonarsource, line 133: _Line 123 above:_ Should be single line declaration.

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -185,8 +174,8 @@ private static bool TryGetFieldFromGetter(AccessorDeclarationSyntax getter, Sema
                 semanticModel, out getterField);
 
             ExpressionSyntax GetReturnExpressionFromBody(BlockSyntax body) =>
-                body?.Statements.Count == 1 &&
-                body.Statements[0] is ReturnStatementSyntax returnStatement
+                body?.Statements.Count == 1
+                && body.Statements[0] is ReturnStatementSyntax returnStatement
```
pavel-mikula-sonarsource, line 178: Could be just single line condition

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -185,8 +174,8 @@ private static bool TryGetFieldFromGetter(AccessorDeclarationSyntax getter, Sema
                 semanticModel, out getterField);
 
             ExpressionSyntax GetReturnExpressionFromBody(BlockSyntax body) =>
-                body?.Statements.Count == 1 &&
-                body.Statements[0] is ReturnStatementSyntax returnStatement
+                body?.Statements.Count == 1
```
pavel-mikula-sonarsource, line 177: _Line 182 below:_ Let's just remove the unneeded indirection. It makes the simple `?.Expression` it really hard to read.

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -35,56 +35,52 @@ namespace SonarAnalyzer.Rules.CSharp
...
                         return;
                     }
-                    if (TryGetFieldFromGetter(getter, c.SemanticModel, out var getterField) &&
-                        TryGetFieldFromSetter(setter, c.SemanticModel, out var setterField) &&
-                        getterField.Equals(setterField) &&
-                        !getterField.GetAttributes().Any() &&
-                        !getterField.IsVolatile &&
-                        getterField.IsStatic == propertySymbol.IsStatic &&
-                        getterField.Type.Equals(propertySymbol.Type))
+                    if (TryGetFieldFromGetter(getter, c.SemanticModel, out var getterField)
```
pavel-mikula-sonarsource, line 69: The previous `if return` could be reverted and embedded in to this long condition chain.

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -94,27 +90,21 @@ private static bool HasDifferentModifiers(SyntaxList<AccessorDeclarationSyntax>
             return accessors.Skip(1).Any(a => !modifiers.SetEquals(GetModifierKinds(a)));
         }
 
-        private static IEnumerable<SyntaxKind> GetModifierKinds(AccessorDeclarationSyntax accessor)
-        {
-            return accessor.Modifiers.Select(m => m.Kind());
-        }
+        private static IEnumerable<SyntaxKind> GetModifierKinds(AccessorDeclarationSyntax accessor) =>
```
pavel-mikula-sonarsource, line 93: There's way too many `Get` prefixes in this rule :(

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -158,8 +148,8 @@ private static bool TryGetFieldFromSetter(AccessorDeclarationSyntax setter, Sema
                 return false;
             }
 
-            if (!(semanticModel.GetSymbolInfo(identifier).Symbol is INamedTypeSymbol type) ||
-                !type.Equals(declaringType))
+            if (!(semanticModel.GetSymbolInfo(identifier).Symbol is INamedTypeSymbol type)
```
pavel-mikula-sonarsource, line 151: This whole method is written just wrong and with a wrong name, forcing the wrong design in the rest of the rule. Let's change the signature to 
```
private static bool FieldSymbol(ExpressionSyntax expression, INamedTypeSymbol declaringType, SemanticModel semanticModel)
```

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -173,8 +163,7 @@ private static bool TryGetFieldFromGetter(AccessorDeclarationSyntax getter, Sema
         {
             getterField = null;
 
-            var returnedExpression = GetReturnExpressionFromBody(getter.Body)
-                ?? GetReturnExpressionFromExpressionBody(getter.ExpressionBody());
+            var returnedExpression = GetReturnExpressionFromBody(getter.Body) ?? GetReturnExpressionFromExpressionBody(getter.ExpressionBody());
```
pavel-mikula-sonarsource, line 166: After addressing other comments, the flow of this method can be simplified a lot. Also by renaming it properly and changing the result type `IFieldSymbol FieldSymbolFromGetter`

pavel-mikula-sonarsource (changes requested):
You've bumped into a rule in a bad shape :)

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
pavel-mikula-sonarsource: This wasn't addressed.

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -35,158 +35,123 @@ namespace SonarAnalyzer.Rules.CSharp
...
-            return accessors.Any(a => a.AttributeLists.Any());
-        }
+        private static bool HasAttributes(SyntaxList<AccessorDeclarationSyntax> accessors) =>
+            accessors.Any(a => a.AttributeLists.Any());
 
         private static bool HasDifferentModifiers(SyntaxList<AccessorDeclarationSyntax> accessors)
         {
             var accessor1 = accessors.First();
-            var modifiers = GetModifierKinds(accessor1).ToHashSet();
+            var modifiers = ModifierKinds(accessor1).ToHashSet();
```
pavel-mikula-sonarsource, line 86: I don't see a reason to give it a name. We don't need the empty line after it as well.
```suggestion
            var modifiers = ModifierKinds(accessors.First()).ToHashSet();
```

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
pavel-mikula-sonarsource: Missing `else if` on line 123, the empty line will not be needed then.

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -35,158 +35,123 @@ namespace SonarAnalyzer.Rules.CSharp
...
+            else if (!(memberAccess.Expression is IdentifierNameSyntax identifier))
             {
-                field = null;
-                return false;
+                return null;
             }
-
-            if (!(semanticModel.GetSymbolInfo(identifier).Symbol is INamedTypeSymbol type) ||
-                !type.Equals(declaringType))
+            else if (!(semanticModel.GetSymbolInfo(identifier).Symbol is INamedTypeSymbol type)
```
pavel-mikula-sonarsource, line 136: This and the one above could be single line for regularity.

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -35,158 +35,123 @@ namespace SonarAnalyzer.Rules.CSharp
...
+                     || !type.Equals(declaringType))
             {
-                field = null;
-                return false;
+                return null;
             }
 
-            field = semanticModel.GetSymbolInfo(expression).Symbol as IFieldSymbol;
-            return field != null;
+            return semanticModel.GetSymbolInfo(expression).Symbol as IFieldSymbol;
```
pavel-mikula-sonarsource, line 142: This should be in `else` block

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -35,158 +35,123 @@ namespace SonarAnalyzer.Rules.CSharp
...
+            if (!(expression is MemberAccessExpressionSyntax memberAccess)
+                || !memberAccess.IsKind(SyntaxKind.SimpleMemberAccessExpression))
             {
-                field = null;
-                return false;
+                return null;
             }
-
-            if (memberAccess.Expression is ThisExpressionSyntax)
+            else if (memberAccess.Expression is ThisExpressionSyntax)
```
pavel-mikula-sonarsource, line 128: After this deobfuscation, it's now clear that the whole method is just wrong :D It should be like this:
```
            if (expression is IdentifierNameSyntax)
            {
                return semanticModel.GetSymbolInfo(expression).Symbol as IFieldSymbol;
            }
            else if (expression is MemberAccessExpressionSyntax memberAccess && memberAccess.IsKind(SyntaxKind.SimpleMemberAccessExpression))
            {
                return memberAccess.Expression is ThisExpressionSyntax
                    || (memberAccess.Expression is IdentifierNameSyntax identifier && semanticModel.GetSymbolInfo(identifier).Symbol is INamedTypeSymbol type && type.Equals(declaringType))
                    ? semanticModel.GetSymbolInfo(expression).Symbol as IFieldSymbol
                    : null;
            }
            else
            {
                return null;
            }
```

analyzers/src/SonarAnalyzer.CSharp/Rules/PropertyToAutoProperty.cs
```diff
@@ -35,158 +35,123 @@ namespace SonarAnalyzer.Rules.CSharp
...
-            AssignmentExpressionSyntax GetAssignmentFromBody(BlockSyntax body) =>
-                body?.Statements.Count == 1 &&
-                body.Statements[0] is ExpressionStatementSyntax statement
+            AssignmentExpressionSyntax AssignmentFromBody(BlockSyntax body) =>
+                body?.Statements.Count == 1 && body.Statements[0] is ExpressionStatementSyntax statement
                 ? statement.Expression as AssignmentExpressionSyntax
                 : null;
 
-            AssignmentExpressionSyntax GetAssignmentFromExpressionBody(ArrowExpressionClauseSyntax expressionBody) =>
+            AssignmentExpressionSyntax AssignmentFromExpressionBody(ArrowExpressionClauseSyntax expressionBody) =>
```
pavel-mikula-sonarsource, line 110: _Line 112 below_: `.ElementAt(0)` should be `.Single()`, becase we tested count=1 before.

pavel-mikula-sonarsource (changes requested):
Round 2 :)

pavel-mikula-sonarsource (approved):
LGTM, what a nice rule :)

PR #4666, opened 2021-07-13
Rule S1541: introduce support for C# 9 init-only setters

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/FunctionComplexity.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/FunctionComplexity.cs
@@ -27,6 +27,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
 using SonarAnalyzer.Metrics.CSharp;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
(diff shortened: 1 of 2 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/FunctionComplexity.cs
```diff
@@ -27,26 +27,24 @@
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
 using SonarAnalyzer.Metrics.CSharp;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     [Rule(DiagnosticId)]
     public class FunctionComplexity : ParameterLoadingDiagnosticAnalyzer
```
csaba-sagi-sonarsource, line 36: Please take a look at why it does not inherit from` FunctionComplexityBase`

csaba-sagi-sonarsource (approved):
LGTM! I approved it, and please don`t forget to report an issue for the pattern matching part and also please take a look at the comment I put regarding the inheritance.

analyzers/src/SonarAnalyzer.CSharp/Rules/FunctionComplexity.cs
csaba-sagi-sonarsource: The following issue is created to tackle this. https://github.com/SonarSource/sonar-dotnet/issues/4675

PR #4669, opened 2021-07-13
Rule S3237: introduce support for C# 9 init-only setters

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/UseValueParameter.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/UseValueParameter.cs
@@ -34,46 +34,42 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class UseValueParameter : SonarDiagnosticAnalyzer
     {
-        internal const string DiagnosticId = "S3237";
+        private const string DiagnosticId = "S3237";
         private const string MessageFormat = "Use the 'value' parameter in this {0} accessor declaration.";
 
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
 
-        protected override void Initialize(SonarAnalysisContext context)
-        {
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
+
+        protected override void Initialize(SonarAnalysisContext context) =>
             context.RegisterSyntaxNodeActionInNonGenerated(
                 c =>
                 {
                     var accessor = (AccessorDeclarationSyntax)c.Node;
 
-                    if ((accessor.Body == null && accessor.ExpressionBody() == null) ||
-                        OnlyThrows(accessor) ||
-                        accessor.DescendantNodes().OfType<IdentifierNameSyntax>().Any(x => IsAccessorValue(x, c.SemanticModel)))
+                    if ((accessor.Body == null && accessor.ExpressionBody() == null)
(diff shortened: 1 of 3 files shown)


dmitry-golovinov-sonarsource:
There is some defensive code in the end of the class that is not covered even with the help of global statements.

analyzers/src/SonarAnalyzer.CSharp/Rules/UseValueParameter.cs
```diff
@@ -82,37 +78,22 @@ private static bool IsAccessorValue(IdentifierNameSyntax identifier, SemanticMod
...
-            return semanticModel.GetSymbolInfo(identifier).Symbol is IParameterSymbol parameter &&
-                parameter.IsImplicitlyDeclared;
+            return semanticModel.GetSymbolInfo(identifier).Symbol is IParameterSymbol { IsImplicitlyDeclared: true };
         }
 
         private static string GetAccessorType(AccessorDeclarationSyntax accessorDeclaration)
         {
-            if (accessorDeclaration.IsKind(SyntaxKind.AddAccessorDeclaration) ||
-                accessorDeclaration.IsKind(SyntaxKind.RemoveAccessorDeclaration))
+            if (accessorDeclaration.IsAnyKind(SyntaxKind.AddAccessorDeclaration, SyntaxKind.RemoveAccessorDeclaration))
```
pavel-mikula-sonarsource, line 86: You can move this in the `switch` below with `EventDeclarationSyntax` to keep the code uniform

analyzers/src/SonarAnalyzer.CSharp/Rules/UseValueParameter.cs
```diff
@@ -82,37 +78,22 @@ private static bool IsAccessorValue(IdentifierNameSyntax identifier, SemanticMod
                 return false;
             }
 
-            return semanticModel.GetSymbolInfo(identifier).Symbol is IParameterSymbol parameter &&
-                parameter.IsImplicitlyDeclared;
+            return semanticModel.GetSymbolInfo(identifier).Symbol is IParameterSymbol { IsImplicitlyDeclared: true };
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: You can cover this with an ugly cheat 😃 
```
        public int ExplicitValueIdentifier
        {
            get => 42;
            set
            {
                Func<int, int> f = value => value;
            }
        }
```

analyzers/src/SonarAnalyzer.CSharp/Rules/UseValueParameter.cs
```diff
@@ -82,37 +78,22 @@ private static bool IsAccessorValue(IdentifierNameSyntax identifier, SemanticMod
...
-            {
-                return "property set";
-            }
-            else
-            {
-                return null;
-            }
+                IndexerDeclarationSyntax _ => "indexer set",
+                PropertyDeclarationSyntax _ => "property set",
+                _ => null
```
pavel-mikula-sonarsource, line 95: This is not coverable with a real code.

analyzers/src/SonarAnalyzer.CSharp/Rules/UseValueParameter.cs
```diff
@@ -82,37 +78,22 @@ private static bool IsAccessorValue(IdentifierNameSyntax identifier, SemanticMod
...
-            else if (indexerOrProperty is PropertyDeclarationSyntax)
-            {
-                return "property set";
-            }
-            else
-            {
-                return null;
-            }
+                IndexerDeclarationSyntax _ => "indexer set",
+                PropertyDeclarationSyntax _ => "property set",
```
pavel-mikula-sonarsource, line 94: Shouldn't there be a message for `property init`? Let's add `{{Error message}}` in UTs to see how it will look like.

pavel-mikula-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/UseValueParameter.cs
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: I've already done it here https://github.com/SonarSource/sonar-dotnet/pull/4669/commits/d0fa7f034088216ab78f0a22e46143cec171d9a9 , but it probably blames for the null check. Can you suggest anything to 
check for null?

pavel-mikula-sonarsource (approved):
LGTM

PR #4670, opened 2021-07-13
S1854 Support C# 9 init-only setters

pavel-mikula-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/DeadStores.cs
```diff
@@ -224,14 +190,10 @@ private void ProcessOpAssignment(SyntaxNode instruction, HashSet<SyntaxNode> ass
             {
                 var assignment = (AssignmentExpressionSyntax)instruction;
                 var left = assignment.Left.RemoveParentheses();
-                if (left.IsKind(SyntaxKind.IdentifierName))
+                if (left.IsKind(SyntaxKind.IdentifierName)
```
csaba-sagi-sonarsource, line 193: The condition of this if is repeated 3 times, it can be extracted to a method.

analyzers/src/SonarAnalyzer.CSharp/Rules/DeadStores.cs
```diff
@@ -330,22 +285,22 @@ private void ProcessPostfixExpression(SyntaxNode instruction, HashSet<ISymbol> l
...
                 {
-                    context.ReportDiagnosticWhenActive(Diagnostic.Create(rule, postfixExpression.GetLocation(), symbol.Name));
+                    context.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, postfixExpression.GetLocation(), symbol.Name));
                 }
             }
 
             private void ReportOnAssignment(AssignmentExpressionSyntax assignment, ExpressionSyntax left, ISymbol symbol, HashSet<SyntaxNode> assignmentLhs, HashSet<ISymbol> outState)
             {
-                if (CSharpLiveVariableAnalysis.IsLocalScoped(symbol, declaration)
+                if (CSharpLiveVariableAnalysis.IsLocalScoped(symbol, this.nodeSymbol)
```
csaba-sagi-sonarsource, line 298: Nitpick: this is redundant.
```suggestion
                if (CSharpLiveVariableAnalysis.IsLocalScoped(symbol, nodeSymbol)
```

csaba-sagi-sonarsource (approved):
LGTM!

I left two comments, please take a look and fix them if you agree.

analyzers/src/SonarAnalyzer.CSharp/Rules/DeadStores.cs
pavel-mikula-sonarsource: Nice, except the coverage went under 95 🤦‍♂️ 😆

PR #4672, opened 2021-07-13
Support C# 9 init-only setters in CSharpMethodDeclarationTracker

pavel-mikula-sonarsource:
There's no UT for this.
It could be reachable with `ControllingPermissions`, but it's obsolete and targets .NET Framework-only. 
It's too complicated to scaffold with artificial AnalyzerContext.


andrei-epure-sonarsource:
how do we know this works if we have no tests for it?

pavel-mikula-sonarsource:
I've found a (relatively) easy way how to scaffold it. 
Review per commit should be easier to understand the steps.
I discovered that the VB part wasn't working properly.

andrei-epure-sonarsource (approved):
LGTM!

PR #4676, opened 2021-07-14
Fix main branch SC decoration

andrei-epure-sonarsource:
see https://discuss.sonarsource.com/t/gh-integration-main-branch-commit-is-green-even-if-qg-failed/7941/9

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -196,6 +196,8 @@ stages:
             sonar.analysis.pipeline=$(Build.BuildId)
             sonar.analysis.sha1=$(Build.SourceVersion)
             sonar.analysis.repository=$(Build.Repository.ID)
+            # needed for SonarCloud main branch decoration for unbound projects - see SC-3469
+            sonar.pullrequest.github.repository=$(Build.Repository.Name)
 
       - powershell: |
           . .\scripts\utils.ps1


azure-pipelines.yml
```diff
@@ -196,6 +196,8 @@ stages:
             sonar.analysis.pipeline=$(Build.BuildId)
             sonar.analysis.sha1=$(Build.SourceVersion)
             sonar.analysis.repository=$(Build.Repository.ID)
+            # needed for SonarCloud main branch decoration for unbound projects - see SC-3469
+            sonar.pullrequest.github.repository=$(Build.Repository.Name)
```
marcin-majewski-sonarsource, line 200: Will the `$(Build.Repository.Name)` output sth like: `SonarSource/sonar-dotnet` ?

andrei-epure-sonarsource: I expect so, this is the environment variable which is used in scanner integration with AZP the https://github.com/SonarSource/sonar-scanner-vsts/blob/master/common/ts/prepare-task.ts#L12

marcin-majewski-sonarsource (approved):

andrei-epure-sonarsource:
This actually didn't work because the SC project needs to be bound to the Github repo. And this currently isn't possible via the SC UI - they plan it for Q3.

PR #4683, opened 2021-07-14
Rule S3928 - don't raise the issue when property name is used in nameof() call.

### LINKED ISSUE #4679: FP S3928: should not raise an issue in set/init, when nameof(property_name) is used
### Description

Rule should not raise an issue in set/init parts of property, when `nameof(property name)` is used.
Using `nameof(value)` here makes no sense, since it produces the exception text that is too generic.

`Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'value')`
### Repro steps
```
    public class InitOnlyProperties
    {
        private readonly string firstName;
	public string FirstName
        {
            get => firstName;
            init => firstName = (value ?? throw new ArgumentNullException(nameof(FirstName)));
        }
    }
```

### Expected behavior
Issue should not be raised.

### Actual behavior
Issue is raised.

### Related information

dmitry-golovinov-sonarsource:
Fixes #4679

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
@@ -137,9 +137,10 @@ namespace SonarAnalyzer.Rules.CSharp
                 {
                     arguments.AddRange(IdentifierNames(indexer.ParameterList));
                 }
-                if (accessor.IsAnyKind(SyntaxKind.SetAccessorDeclaration, SyntaxKindEx.InitAccessorDeclaration))
+                if (accessor.IsAnyKind(SyntaxKind.SetAccessorDeclaration, SyntaxKindEx.InitAccessorDeclaration)
+                    && accessor.Parent.Parent is PropertyDeclarationSyntax propertyDeclaration)
                 {
-                    arguments.Add("value");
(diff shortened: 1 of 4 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
```diff
@@ -137,9 +137,10 @@ static IEnumerable<string> AccessorIdentifierNames(AccessorDeclarationSyntax acc
                 {
                     arguments.AddRange(IdentifierNames(indexer.ParameterList));
                 }
-                if (accessor.IsAnyKind(SyntaxKind.SetAccessorDeclaration, SyntaxKindEx.InitAccessorDeclaration))
+                if (accessor.IsAnyKind(SyntaxKind.SetAccessorDeclaration, SyntaxKindEx.InitAccessorDeclaration)
+                    && accessor.Parent.Parent is PropertyDeclarationSyntax propertyDeclaration)
                 {
-                    arguments.Add("value");
```
csaba-sagi-sonarsource, line 142: Maybe we could add both the Property name and "value" and cover both cases?

csaba-sagi-sonarsource (approved):
I don`t have a strong opinion on "value". Maybe there are developers who prefer that one, so I would enable both.

PR #4690, opened 2021-07-15
Introduce C# 9 init only setter cases in Net5 IT project.

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/its/sources/Net5/Net5/S1185.cs
+++ b/analyzers/its/sources/Net5/Net5/S1185.cs
@@ -1,0 +1,15 @@
+﻿namespace Net5
+{
+    public class S1185
+    {
+        record Base
+        {
+            public virtual int MyProperty1 { get; init; }
+        }
+
+        record Derived : Base
+        {
+            public override int MyProperty1 { get => base.MyProperty1; init => base.MyProperty1 = value; }
+        }
+    }
+}
--- a/analyzers/its/sources/Net5/Net5/S1541.cs
+++ b/analyzers/its/sources/Net5/Net5/S1541.cs
@@ -1,0 +1,28 @@
+﻿namespace Net5
+{
+    public class S1541
+    {
+        public int Property
+        {
+            get
+            {
+                return 0;
+            }
+            init
+            {
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+            }
+        }
+    }
+}
--- a/analyzers/its/sources/Net5/Net5/S2292.cs
+++ b/analyzers/its/sources/Net5/Net5/S2292.cs
@@ -1,0 +1,15 @@
+﻿namespace Net5
+{
+    public class S2292
+    {
+        public record Record
+        {
+            private string field;
+            public string PropWithGetAndInit
+            {
+                get { return field; }
+                init { field = value; }
+            }
+        }
+    }
+}
--- a/analyzers/its/sources/Net5/Net5/S2292.cs
+++ b/analyzers/its/sources/Net5/Net5/S2292.cs
@@ -5,6 +5,7 @@
         public record Record
         {
             private string field;
+
             public string PropWithGetAndInit
             {
                 get { return field; }
--- a/analyzers/its/sources/Net5/Net5/S3440.cs
+++ b/analyzers/its/sources/Net5/Net5/S3440.cs
@@ -13,6 +13,7 @@
         }
 
         private int? f;
+
         public int Property
         {
             get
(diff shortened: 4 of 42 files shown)


dmitry-golovinov-sonarsource:
Please note, that the coverage of 2 rules is still missing, since they are still in development - S2292 S1185

analyzers/its/sources/Net5/Net5/S1185.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Net5
```
pavel-mikula-sonarsource, line 1: _Not related to this file:_ Let's add test case for S2259 to verify Symbolic Execution for `init` setters

```
    public int PropertySimple
    {
        get => 42;
        init
        {
            object o = null;
            field = o.ToString();   // Noncompliant
        }
    }
```

analyzers/its/sources/Net5/Net5/S2292.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Net5
+{
+    public class S2292
```
pavel-mikula-sonarsource, line 3: There's no S2292 in the modified JSON files

analyzers/its/sources/Net5/Net5/S1541.cs
```diff
@@ -0,0 +1,28 @@
...
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
+                if (false) { }
```
pavel-mikula-sonarsource, line 24: Why so many? Shouldn't just 2 or 3 of them do the same trick?

analyzers/its/sources/Net5/Net5/S2292.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        public record Record
+        {
+            private string field;
+            public string PropWithGetAndInit
```
pavel-mikula-sonarsource, line 8: _Nitpick_: Missing EOL
```suggestion
            private string field;
            
            public string PropWithGetAndInit
```

analyzers/its/sources/Net5/Net5/S3440.cs
```diff
@@ -11,5 +11,25 @@ public void Foo()
                 6 => 6
             };
         }
+
+        private int? f;
+        public int Property
```
pavel-mikula-sonarsource, line 16: _Nitpick:_ Missing EOL
```suggestion
        private int? f;
        
        public int Property
```

analyzers/its/sources/Net5/Net5/S3776.cs
```diff
@@ -2,6 +2,22 @@
 {
     public class S3776
     {
+        string field;
+        private int one, two;
+        string Property
```
pavel-mikula-sonarsource, line 7: _Nitpick:_ Missing EOL
```suggestion
        private int one, two;
        string Property
```

analyzers/its/sources/Net5/Net5/S3776.cs
```diff
@@ -2,6 +2,22 @@
...
+                if ((one is 1 or 2) || (two is 3 or 5) || (two is 3 or 5) || (two is 3 or 5)
+                    || (two is 3 or 5) || (two is 3 or 5) || (two is 3 or 5) || (two is 3 or 5)
+                    || (two is 3 or 5) || (two is 3 or 5) || (two is 3 or 5) || (two is 3 or 5)
+                    || (two is 3 or 5) || (two is 3 or 5) || (two is 3 or 5) || (two is 3 or 5)
+                    || (two is 3 or 5) || (two is 3 or 5) || (two is 3 or 5) || (two is 3 or 5))
+                {
+                    field = "";
+                }
+            }
+        }
```
pavel-mikula-sonarsource, line 20: _NitpicK:_ Missing EOL

pavel-mikula-sonarsource (changes requested):

analyzers/its/sources/Net5/Net5/S2292.cs
dmitry-golovinov-sonarsource: See my comment above in the PR. This rule is still not merged and it's you who reviews it :)

analyzers/its/sources/Net5/Net5/S1541.cs
dmitry-golovinov-sonarsource: Since the default minimum value is 10. So you need to launch this rule with a custom parameter to make it work here.

analyzers/its/sources/Net5/Net5/S2292.cs
pavel-mikula-sonarsource: It's merged now, so we'll need a rebase and the JSON

pavel-mikula-sonarsource (approved):
LGTM, please rebase before merging to avoid broken master

PR #4695, opened 2021-07-15
CSharpExecutableLinesMetric supports switch expressions, local functions and expression bodied methods

### LINKED ISSUE #4689: CSharpExecutableLinesMetric does not count executable line for expression bodied methods
### Description

CSharpExecutableLinesMetric does not count executable line for expression bodied methods

### Repro steps

```

 public static bool IsLetter(this char c) =>
                                c is >= 'a' and <= 'z' or >= 'A' and <= 'Z'; // FN

var x = s switch
                            {
                                    ""a"" => true, // FN
                                    ""b"" => false, // FN
                                    _ => true // FN
                            };

```

### Expected behavior

CSharpExecutableLinesMetric counts executable line for expression bodied methods and also for switch expression arms.

### Actual behavior

CSharpExecutableLinesMetric  does not count those lines.

### Known workarounds

-

### Related information

* C#/VB.NET Plugins version: 8.26.0.34506

dmitry-golovinov-sonarsource:
Fixes #4689


dmitry-golovinov-sonarsource:
PR got crazy rebased on Csaba's branch by my IDE. Please check only the last commit

analyzers/tests/SonarAnalyzer.UnitTest/Metrics/CSharpExecutableLinesMetricTest.cs
```diff
@@ -192,7 +192,7 @@ static void Main()
                         HelpLink = ""
                     };
                 }",
-              7, 11);
+                2, 7, 11);
```
andrei-epure-sonarsource, line 195: please add +1 to line 2 for a visual helper

analyzers/tests/SonarAnalyzer.UnitTest/Metrics/CSharpExecutableLinesMetricTest.cs
```diff
@@ -798,12 +813,10 @@ int M()
...
             AssertLinesOfCode(
                 @"
-                        using System;
-                        using System.Collections.Generic;
-
-                            public static bool IsLetter(this char c) =>
-                                c is >= 'a' and <= 'z' or >= 'A' and <= 'Z'; // FN
-                 ");
+                          public static bool IsLetter(this char c) =>
+                              c is >= 'a' and <= 'z' or >= 'A' and <= 'Z'; // FN
```
andrei-epure-sonarsource, line 817: remove `// FN`

analyzers/tests/SonarAnalyzer.UnitTest/Metrics/CSharpExecutableLinesMetricTest.cs
```diff
@@ -734,7 +749,7 @@ int M2()
                            static int Add(int left, int right) => left + right;
                        }
                 ",
-              5, 6, 15);
+              5, 6, 8, 15, 17);
```
andrei-epure-sonarsource, line 752: please add `// +1` comments for visual aid

andrei-epure-sonarsource (changes requested):
some comments left.

Could you also do some tests with before and after and see how this changes the metrics, and document the findings on this PR? Because based on this PMs will have to communicate these changes to users that a change in coverage numbers will come (cc @tom-howlett-sonarsource - for SC and SQ users).

You can use one of the projects in  project being analyzed https://github.com/SonarSource/sonar-dotnet/tree/master/its/projects/ to start with, and add the new syntaxes (or write your own sample project).

Also, it would be nice to improve `AssertLinesOfCode` to parse comments and not have to pass the line  numbers, but to check those lines just like the UTs do with `// Noncompliant` - maybe open a `tooling` ticket for this?

analyzers/src/SonarAnalyzer.CSharp/Metrics/CSharpExecutableLinesMetric.cs
```diff
@@ -109,6 +109,14 @@ private bool FindExecutableLines(SyntaxNode node)
                         ExecutableLineNumbers.Add(node.GetLocation().GetLineNumberToReport());
                         return true;
 
+                    case SyntaxKindEx.SwitchExpressionArm:
```
andrei-epure-sonarsource, line 112: please read this comment: https://github.com/SonarSource/sonar-dotnet/pull/4663/commits/a7d2a8624895d9acc539dc20948d5ec7bba2b5e3#r670270653

andrei-epure-sonarsource:
Please read the discussion on https://github.com/SonarSource/sonar-dotnet/pull/4663  - I am not sure we should do these updates before understanding better what we want to do.

dmitry-golovinov-sonarsource:
@andrei-epure-sonarsource I actually converted it to draft since I saw your comments after rebase and we had this discussion. I want either convert this to a background task and work on it when I have time or just pause it till the specification sprint (I am interested in it). WDYT?

tom-howlett-sonarsource:
Hi, thanks for flagging me here. I presume that this change will increase the executable lines and therefore has the potential to reduce coverage unexpectedly? It would be good to understand the magnitude of the impact on a test project as @andrei-epure-sonarsource has suggested. If I have that I can have a conversation with the SC/SQ PMs to talk about whether it's necessary to warn users and how.

andrei-epure-sonarsource:
@dmitry-golovinov-sonarsource I agree that we should first decide how to compute executable lines for switch expressions and also evaluate where we aren't consistent with existing specification, and then do one or a series of PRs to fix it. If you want as a background task you could do the evaluation of code coverage changes after this PR, it will help @tom-howlett-sonarsource understand the efforts needed for the sprint (e.g. do we need to update documentation etc).

PR #4697, opened 2021-07-16
Rule S1313: Improve performance

### LINKED ISSUE #4601: Fix S1313 performance regression: HardcodedIpAddress
As reported in this [community thread](https://community.sonarsource.com/t/hardcodedipaddress-scan-takes-long-in-sonarqube-8-9/44941):

Upgrade from SQ 7.9 LTS to 8.9 LTS changed the build time from 20m to 1h 8m for the user.

dmitry-golovinov-sonarsource:
Fixes #4601


analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/HardcodedIpAddressTest.cs
```diff
@@ -35,6 +38,15 @@ public class HardcodedIpAddressTest
         public void HardcodedIpAddress_CS() =>
             Verifier.VerifyAnalyzer(@"TestCases\Hotspots\HardcodedIpAddress.cs", new CS.HardcodedIpAddress(AnalyzerConfiguration.AlwaysEnabled));
 
+        [TestMethod]
```
csaba-sagi-sonarsource, line 41: Please Try to move this test case to the RuleTests.cs file, and run it against all the Rules we have and not just the `HardcodedIpAddress `rule. Also please rename DictionaryStressTest to something like PerformanceTestCases, because we plan to add other cases to the same file.

csaba-sagi-sonarsource (changes requested):
Looks good to me! Please see the last comment I have.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/HardcodedIpAddressTest.cs
```diff
@@ -18,6 +18,9 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
```
csaba-sagi-sonarsource, line 21: Unnecessary usings, you can remove them

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
```diff
@@ -89,6 +89,7 @@ public void RulesDoNotThrow_CS()
 
             Verifier.VerifyNoExceptionThrown(@"TestCasesForRuleFailure\InvalidSyntax.cs", analyzers, CompilationErrorBehavior.Ignore);
             Verifier.VerifyNoExceptionThrown(@"TestCasesForRuleFailure\SpecialCases.cs", analyzers, CompilationErrorBehavior.Ignore);
+            Verifier.VerifyNoExceptionThrown(@"TestCasesForRuleFailure\DictionaryStressTest.cs", analyzers, CompilationErrorBehavior.Ignore);
```
csaba-sagi-sonarsource, line 92: Maybe it would be nice to rename DictionaryStressTest to PerformanceTestCases. So the intention is clear and other syntaxes can be added in the future.

csaba-sagi-sonarsource (approved):
I approved it. Please take a look at the two suggestions I left.

PR #4699, opened 2021-07-16
AZP: Use shallow clone for non-analysis jobs

pavel-mikula-sonarsource:


pavel-mikula-sonarsource:
The improvement is not good enough, sadly

PR #4703, opened 2021-07-19
Enable concurrent analysis for .NET Analyzer rules

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/InvalidCastToInterface.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/InvalidCastToInterface.cs
@@ -35,7 +35,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private const string MessageReviewFormat = "Review this cast; in this project there's no type that {0}.";
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(InvalidCastToInterfaceRuleConstants.Rule);
-        protected override bool EnableConcurrentExecution { get; } = false;
+        protected override bool EnableConcurrentExecution => false;
 
         protected override void Initialize(SonarAnalysisContext context)
         {
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/InvalidCastToInterface.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/InvalidCastToInterface.cs
@@ -35,6 +35,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private const string MessageReviewFormat = "Review this cast; in this project there's no type that {0}.";
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(InvalidCastToInterfaceRuleConstants.Rule);
+        protected override bool EnableConcurrentExecution { get; } = false;
 
         protected override void Initialize(SonarAnalysisContext context)
         {
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/InvalidCastToInterface.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/InvalidCastToInterface.cs
@@ -35,7 +35,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private const string MessageReviewFormat = "Review this cast; in this project there's no type that {0}.";
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(InvalidCastToInterfaceRuleConstants.Rule);
-        protected override bool ParallelProcessingDisabled => true;
+        protected override bool ConcurrentProcessingDisabled => true;
 
         protected override void Initialize(SonarAnalysisContext context)
         {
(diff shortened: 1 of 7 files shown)


analyzers/src/SonarAnalyzer.Common/DiagnosticAnalyzer/SonarDiagnosticAnalyzer.cs
```diff
@@ -24,8 +24,14 @@ namespace SonarAnalyzer.Helpers
 {
     public abstract class SonarDiagnosticAnalyzer : DiagnosticAnalyzer
     {
+        protected virtual bool EnableConcurrentExecution { get; } = true;
```
csaba-sagi-sonarsource, line 27: I would do it this way, because this way it is only settable to false via override.

```suggestion
        protected virtual bool EnableConcurrentExecution => true;
```

analyzers/src/SonarAnalyzer.Common/DiagnosticAnalyzer/SonarDiagnosticAnalyzer.cs
```diff
@@ -24,8 +24,14 @@ namespace SonarAnalyzer.Helpers
...
+
         public sealed override void Initialize(AnalysisContext context)
         {
+            if (EnableConcurrentExecution)
+            {
+                context.EnableConcurrentExecution();
+            }
             Initialize(new SonarAnalysisContext(context, SupportedDiagnostics));
         }
```
csaba-sagi-sonarsource, line 37: The protected abstract initialize should go before the overriden one.

csaba-sagi-sonarsource (approved):
Two minor changes left to fix, I`ll approve anyway

analyzers/src/SonarAnalyzer.CSharp/Rules/InvalidCastToInterface.cs
```diff
@@ -35,6 +35,7 @@ public sealed class InvalidCastToInterface : SonarDiagnosticAnalyzer
         private const string MessageReviewFormat = "Review this cast; in this project there's no type that {0}.";
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(InvalidCastToInterfaceRuleConstants.Rule);
+        protected override bool ParallelProcessingDisabled => true;
```
csaba-sagi-sonarsource, line 38: I would call the Property Concurrent instead of Paralell. That is the name in the Roslyn API too.

analyzers/src/SonarAnalyzer.Common/DiagnosticAnalyzer/SonarDiagnosticAnalyzer.cs
```diff
@@ -18,17 +18,36 @@
...
         {
+            if (!ParallelProcessingDisabled)
+            {
+                context.EnableConcurrentExecution();
+            }
             Initialize(new SonarAnalysisContext(context, SupportedDiagnostics));
         }
 
-        protected abstract void Initialize(SonarAnalysisContext context);
+        private static bool TryGetBoolEnvironmentVariable(string envVar, bool defaultValue)
```
csaba-sagi-sonarsource, line 42: I would rename the method, the TryGet prefix suggest there will be an out parameter. which is not the case.

csaba-sagi-sonarsource, line 42: Also, I would say that the parameters are not needed, the environment variable name is a constant in the class so it can be used directly, and the default value can be also defined as a constant or just used as a false in the method with a comment.

csaba-sagi-sonarsource (changes requested):
Few minor changes suggested.

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
```diff
@@ -113,6 +116,51 @@ public void SonarDiagnosticAnalyzer_IsUsedInAllRules()
             }
         }
 
+        [TestMethod]
+        public void Verify_ConcurrentProcessingDisabledByDefault()
+        {
+            var retriever = new ConcurrentProcessingRetriever();
+            retriever.IsConcurrentProcessingDisabled.Should().BeNull();
+            Verifier.VerifyAnalyzer(new[] { "TestCasesForRuleFailure\\SpecialCases.cs" }, retriever);
```
csaba-sagi-sonarsource, line 124: I think that in all tests we can use VerifyNoExceptionThrown instead of VeriFyAnalyzer. The intention is to see that the rule is initialized and no exceptions are thrown during analysis.

csaba-sagi-sonarsource (approved):
One small suggestion left, and please don`t forget to create a feature branch and merge it there.

analyzers/src/SonarAnalyzer.Common/DiagnosticAnalyzer/SonarDiagnosticAnalyzer.cs
```diff
@@ -18,17 +18,36 @@
...
 
+using System;
 using Microsoft.CodeAnalysis.Diagnostics;
 
 namespace SonarAnalyzer.Helpers
 {
     public abstract class SonarDiagnosticAnalyzer : DiagnosticAnalyzer
     {
+        private const string EnableConcurrentProcessing = "SONAR_DOTNET_ENABLE_CONCURRENT_PROCESSING";
+        protected virtual bool ConcurrentProcessingDisabled => IsConcurrentProcessingDisabled();
```
pavel-mikula-sonarsource, line 29: Is there a strong reason for this negation? `true` turns something off and that's confusing brain twist.

I'd strongly prefer `ConcurrentProcessingEnabled` with default value of `true` and turning it off with `false`. That is also aligned with the environment variable name.

pavel-mikula-sonarsource: I'd actually prefer the original `EnableConcurrentExecution ` as it was specified to be aligned with Roslyn API and clear about it's action.

dmitry-golovinov-sonarsource: Done here - https://github.com/SonarSource/sonar-dotnet/pull/4708

Issue #4704, opened 2021-07-19
Fix S1172 FP: Parameter captured in a referenced generic local function


davidst:
### Description

The S1172 warning is shown for a parameter captured by a generic local function.

### Repro steps

```csharp
        private static void ConfigureAndValidateSettings(this IServiceCollection services, IConfiguration configuration)
        {
            ConfigureAndValidate<MySettings1>();
            ConfigureAndValidate<MySettings2>();
            ConfigureAndValidate<MySettings3>();

            void ConfigureAndValidate<TOptions>() where TOptions : class
            {
                var config = configuration.GetSection(typeof(TOptions).Name);
                services.AddOptions<TOptions>().Bind(config).ValidateDataAnnotations();
            }
        }
```

### Expected behavior

No S1172 warning is shown for this case.

### Actual behavior

The S1172 warning is shown even though the parameter is required.

### Known workarounds

Suppress the warning, assign the parameter to a discard, or disable the SonarLint extension.

### Related information

* C#/VB.NET Plugins version 4.36.034265
* SonarAnalyzer.CSharp version 8.26.034506
* Visual Studio version 16.10.1
* dotnet version 5.0.302
* Operating System Windows 10 Pro 21H1

PR #4706, opened 2021-07-20
FP reproducer when parameter is used in generic local function

csaba-sagi-sonarsource:
Reproduces #4704


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MethodParameterUnused.CSharp8.cs
```diff
@@ -119,4 +120,19 @@ private string UsedInTuple(string value)
             return x.value;
         }
     }
+
```
dmitry-golovinov-sonarsource, line 123: Consider removing this extra line

dmitry-golovinov-sonarsource (approved):
LGTM! Please, check my minor comment

PR #4708, opened 2021-07-22
Add Verifier for concurrent analysers

dmitry-golovinov-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -54,83 +55,149 @@ public static void VerifyNoExceptionThrown(string path, IEnumerable<DiagnosticAn
             DiagnosticVerifier.GetAllDiagnostics(compilation, diagnosticAnalyzers, checkMode);
         }
 
-        public static void VerifyCSharpAnalyzer(string snippet,
+        public static void VerifyNonConcurrentCSharpAnalyzer(string snippet,
```
pavel-mikula-sonarsource, line 58: Do we really need to rename these language-specific overloads? I'd keep the names short as they were. And maybe just add a documentation comment that it performs nonconcurrent verification.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -78,10 +79,17 @@ internal struct SolutionBuilder
                 throw new ArgumentException("Please use a collection of paths with the same extension", nameof(paths));
             }
 
-            var project = Create()
-                .AddProject(AnalyzerLanguage.FromPath(paths.First()), outputKind: outputKind)
-                .AddDocuments(paths)
-                .AddReferences(additionalReferences);
+            var project = duplicateDocumentsNeeded switch
```
pavel-mikula-sonarsource, line 82: Ternary operator instead of switch

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -78,10 +79,17 @@ internal struct SolutionBuilder
...
+            {
```
pavel-mikula-sonarsource, line 83: 2 out of 3 actions are duplicated here and we have builder => let's do the two actions in series and then decide with `if` what do do differently, like the unmodified statements below

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -78,10 +79,17 @@ internal struct SolutionBuilder
...
+                true => Create()
```
pavel-mikula-sonarsource, line 84: In general, I'm not happy with the responsibility of this class. `ProjectBuilder` and `SolutionBuilder` should not know and care about this issue at all.

Let's move the logic to a higher level (`Verifier`) that is aware of the concurrency issue and pass only the final list of files inside here.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/EnvironmentalVariableScope.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                originalValues.Add(name, Environment.GetEnvironmentVariable(name));
+            }
+            Environment.SetEnvironmentVariable(name, value, EnvironmentVariableTarget.Process);
+        }
+
+        public void SetVariableInRelease(string name, string value)
+        {
+            // do nothing in Debug mode
+#if RELEASE
```
pavel-mikula-sonarsource, line 48: This should use `TestContextHelper.IsAzureDevOpsContext`, same as `ParseOptionsHelper`

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/EnvironmentalVariableScope.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// <summary>
+    /// Defines a scope inside which new environment variables can be set.
+    /// The variables will be cleared when the scope is disposed.
+    /// Copy-pasted from S4NET codebase.
+    /// </summary>
+    public sealed class EnvironmentVariableScope : IDisposable
+    {
+        private IDictionary<string, string> originalValues = new Dictionary<string, string>();
+
+        public void SetVariable(string name, string value)
```
pavel-mikula-sonarsource, line 35: `SetVariable` vs `SetVariableInRelease` confuses me from API/usage point of view.
When using the class (long after I forget this detail), I will want to have just a single method that will do the trick for me. And that will deal with the specific context for me so I don't have to think about it everytime I use this.

I suggest to introduce constructor with optional argument that will be explicitly set only for those specific tests and that will do this trick for us by default otherwise.

pavel-mikula-sonarsource (changes requested):
Round 1 to shorten the loop. I didn't have time to see `Rules`, `TestCases` and I saw only part of `TestFramework`

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/EnvironmentalVariableScope.cs
dmitry-golovinov-sonarsource: How is it connected with Azure at all? This is just for local debugging purposes.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
dmitry-golovinov-sonarsource: Logic was moved. No longer needed.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
dmitry-golovinov-sonarsource: Logic was moved. No longer needed.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/EnvironmentalVariableScope.cs
pavel-mikula-sonarsource: Same as the IsAzureDevOpsContext is there to behave differently under CI and local debugging purposes.
On CI, we run all language versions.
On local machine (maybe except release test run from a script) we run only one language version to make it faster and easier (to avoid hitting the same breakpoint 4 times, actually)

The key point is to have only one common mechanism to deal with the same task `I want easy debugging on localhost`.

dmitry-golovinov-sonarsource: But that won't help us detecting issues with concurrency on local runs, unless you remember and set this property explicitly for that every time you modify the rule.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -261,6 +280,36 @@ private static void VerifyBuildErrors(ImmutableArray<Diagnostic> diagnostics, Co
 
         private static string IssueType(bool isPrimary) => isPrimary ? "primary" : "secondary";
 
+        internal class FileNameSourceText
+        {
+            public string FileName { get; set; }
+            public SourceText SourceText { get; set; }
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Probably should be read  only

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -261,6 +280,36 @@ private static void VerifyBuildErrors(ImmutableArray<Diagnostic> diagnostics, Co
...
+            public FileNameSourceText(string fileName, SourceText sourceText)
+            {
+                FileName = fileName;
+                SourceText = sourceText;
+            }
+
+            public FileNameIssueLocations ToIssueLocations() =>
+                new FileNameIssueLocations(FileName, IssueLocationCollector.GetExpectedIssueLocations(SourceText.Lines));
+
+            public static FileNameSourceText FromSyntaxTree(SyntaxTree syntaxTree) =>
```
pavel-mikula-sonarsource, line 296: Let's make this a constructor overload

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -261,6 +280,36 @@ private static void VerifyBuildErrors(ImmutableArray<Diagnostic> diagnostics, Co
...
+        {
+            public string FileName { get; set; }
+            public SourceText SourceText { get; set; }
+            public FileNameSourceText(string fileName, SourceText sourceText)
+            {
+                FileName = fileName;
+                SourceText = sourceText;
+            }
+
+            public FileNameIssueLocations ToIssueLocations() =>
```
pavel-mikula-sonarsource, line 293: I'd prefer to remove this method. It's intended to be used only inside this class and the other usage is a very special case.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -261,6 +280,36 @@ private static void VerifyBuildErrors(ImmutableArray<Diagnostic> diagnostics, Co
...
+
+            public static FileNameSourceText FromSyntaxTree(SyntaxTree syntaxTree) =>
+                new FileNameSourceText(syntaxTree.FilePath, syntaxTree.GetText());
+        }
+
+        internal class FileNameIssueLocations
+        {
+            public string FileName { get; set; }
+            public IList<IIssueLocation> IssueLocations { get; set; }
+            public FileNameIssueLocations(string fileName, IList<IIssueLocation> issueLocations)
```
pavel-mikula-sonarsource, line 304: Missing EOL before constructor

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -261,6 +280,36 @@ private static void VerifyBuildErrors(ImmutableArray<Diagnostic> diagnostics, Co
...
+                new FileNameIssueLocations(FileName, IssueLocationCollector.GetExpectedIssueLocations(SourceText.Lines));
+
+            public static FileNameSourceText FromSyntaxTree(SyntaxTree syntaxTree) =>
+                new FileNameSourceText(syntaxTree.FilePath, syntaxTree.GetText());
+        }
+
+        internal class FileNameIssueLocations
+        {
+            public string FileName { get; set; }
+            public IList<IIssueLocation> IssueLocations { get; set; }
```
pavel-mikula-sonarsource, line 303: Both should be readonly

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -261,6 +280,36 @@ private static void VerifyBuildErrors(ImmutableArray<Diagnostic> diagnostics, Co
...
+            public FileNameIssueLocations(string fileName, IList<IIssueLocation> issueLocations)
+            {
+                FileName = fileName;
+                IssueLocations = issueLocations;
+            }
+
+            public static FileNameIssueLocations BuildErrorsFromSyntaxTree(SyntaxTree syntaxTree) =>
```
pavel-mikula-sonarsource, line 310: Let's remove this method. It's used just once, I originally wanted to make it constructor, but that leads to naming issue => it's a sign that it shouldn't be here.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -54,83 +56,175 @@ public static void VerifyNoExceptionThrown(string path, IEnumerable<DiagnosticAn
             DiagnosticVerifier.GetAllDiagnostics(compilation, diagnosticAnalyzers, checkMode);
         }
 
+        /// <summary>
+        /// Verify analyzer from C# on a snippet. Note that analyzer is being run in non-concurrent execution mode,
+        /// that means that issues regarding concurrent execution could not be detected with this method.
```
pavel-mikula-sonarsource, line 61: For simplicity
```suggestion
        /// Verify analyzer from C# on a snippet in non-concurrent execution mode.
```

pavel-mikula-sonarsource: Same for all of them (careful about the VB one :) )

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/EnvironmentalVariableScope.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// <summary>
+    /// Defines a scope inside which new environment variables can be set.
+    /// The variables will be cleared when the scope is disposed.
+    /// Copy-pasted from S4NET codebase.
+    /// </summary>
+    public sealed class EnvironmentVariableScope : IDisposable
+    {
+        private readonly bool setOnlyInRelease;
+        private IDictionary<string, string> originalValues = new Dictionary<string, string>();
+
```
pavel-mikula-sonarsource, line 35: As an API, this class is general and therefore quite hard to use for it's main purpose.
Let's keep it's methods for advanced scenarios (like the `tRUE` case) and add writeonly property for each well-known-often-needed variable.
```
#pragma warning disable S2376 // Write-only properties should not be used

        public bool EnableConcurrentAnalysis
        {
            set
            {
                SetVariable(SonarDiagnosticAnalyzer.EnableConcurrentExecutionVariable, value.ToString());
            }
        }

#pragma warning restore S2376
```
that will enable simple usage like this on most of places
```
using var _ = new EnvironmentVariableScope { EnableConcurrentAnalysis = true };
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -191,10 +282,10 @@ static IEnumerable<TMessage> ReadProtobuf(string path)
             }
         }
 
-        public static void VerifyUtilityAnalyzerIsNotRun(IEnumerable<string> paths,
-                                                           UtilityAnalyzerBase diagnosticAnalyzer,
-                                                           string protobufPath,
-                                                           IEnumerable<ParseOptions> options = null)
+        public static void VerifyNonConcurrentUtilityAnalyzerIsNotRun(IEnumerable<string> paths,
```
pavel-mikula-sonarsource, line 285: This one doesn't need a rename, because there's no concurrent overload. Let's just add a short documentation comment.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -54,83 +56,175 @@ public static void VerifyNoExceptionThrown(string path, IEnumerable<DiagnosticAn
...
                                           IEnumerable<MetadataReference> additionalReferences,
                                           string sonarProjectConfigPath = null) =>
-            VerifyAnalyzer(new[] { path }, new[] { diagnosticAnalyzer }, options, CompilationErrorBehavior.Default, OutputKind.DynamicallyLinkedLibrary, additionalReferences, sonarProjectConfigPath);
+            VerifyAnalyzer(new[] { path },
+                                     new[] { diagnosticAnalyzer },
+                                     options,
+                                     CompilationErrorBehavior.Default,
+                                     OutputKind.DynamicallyLinkedLibrary,
+                                     additionalReferences,
+                                     sonarProjectConfigPath);
```
pavel-mikula-sonarsource, line 222: Messy indent. It's very expensive to maintain manually, I'd prefer invocations 4 on spaces.

pavel-mikula-sonarsource (changes requested):
Round 2 - all except `Rules` and `TestCases`

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -79,15 +79,44 @@ internal struct SolutionBuilder
             }
 
             var project = Create()
-                .AddProject(AnalyzerLanguage.FromPath(paths.First()), outputKind: outputKind)
-                .AddDocuments(paths)
-                .AddReferences(additionalReferences);
+                                  .AddProject(AnalyzerLanguage.FromPath(paths.First()), outputKind: outputKind)
+                                  .AddDocuments(paths)
+                                  .AddReferences(additionalReferences);
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Okay, but isn't it something we were going to discuss on one of our how we work? This kind of alignment

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/EnvironmentalVariableScope.cs
```diff
@@ -0,0 +1,88 @@
...
+        public EnvironmentVariableScope(bool setVariablesOnlyInRelease = false)
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: The new name is "setVariablesOnlyInAzureDevOpsContext" and this is what it seems to be doing, since this no longer corellates with debug and release and only with a special variable one would only find in Azure context. Do you agree?

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/EnvironmentalVariableScope.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            setOnlyInRelease = setVariablesOnlyInRelease;
+        }
+
+        public void SetVariable(string name, string value)
+        {
+            if (setOnlyInRelease)
+            {
+#if !RELEASE
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Code is outdated. It's AZP check for now

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -141,12 +156,15 @@ public static void Verify(Compilation compilation, DiagnosticAnalyzer[] diagnost
...
         }
 
-        internal static void CompareActualToExpected(string languageVersion, IEnumerable<Diagnostic> diagnostics, ICollection<IIssueLocation> expectedIssues, bool compareIdToMessage)
+        internal static void CompareActualToExpected(string languageVersion, IEnumerable<Diagnostic> diagnostics, List<FileNameIssueLocations> expectedIssuesPerFile, bool compareIdToMessage)
         {
             DumpActualDiagnostics(languageVersion, diagnostics);
 
             foreach (var diagnostic in diagnostics)
             {
+                var expectedIssues = diagnostic.Location.SourceTree != null
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Removed the comment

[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: For us, it was CBDE that was failing in the very beginning, so we've thought that this is it :) Good to know about S3904

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -261,6 +280,36 @@ private static void VerifyBuildErrors(ImmutableArray<Diagnostic> diagnostics, Co
 
         private static string IssueType(bool isPrimary) => isPrimary ? "primary" : "secondary";
 
+        internal class FileNameSourceText
+        {
+            public string FileName { get; set; }
+            public SourceText SourceText { get; set; }
+            public FileNameSourceText(string fileName, SourceText sourceText)
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Not always. There is a case when source text could not be fetched from filename

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
dmitry-golovinov-sonarsource: Yep, was talking about that one

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
dmitry-golovinov-sonarsource: It is pretty long one. Also, since I've hidden all the members - nothing is seen outside. Let's keep it as it is, since it's really a space-saver

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
dmitry-golovinov-sonarsource: I don't really like the result, but ok.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/ProjectFileAsPathAndContent.cs
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Text.RegularExpressions;
+using SonarAnalyzer.Common;
+
+namespace SonarAnalyzer.UnitTest.TestFramework
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: File got dropped

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Was reverted to master

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -79,15 +79,44 @@ internal struct SolutionBuilder
...
 
             if (isSupportForCSharp9InitNeeded)
             {
                 project = project.AddSnippet(InitSnippet);
             }
 
+            return AddReferencesToSolution(additionalReferences, project);
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Was reverted to master

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/EnvironmentalVariableScope.cs
```diff
@@ -0,0 +1,88 @@
...
+            }
+            disposed = true;
+
+            if (disposing && originalValues != null)
+            {
+                foreach (var kvp in originalValues)
+                {
+                    Environment.SetEnvironmentVariable(kvp.Key, kvp.Value);
+                }
+
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I wanted to kill current line 80, the one before `originalValues = null` 😄 

I don't mind empty lines around `#region`, they help with readability in most of the cases.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
pavel-mikula-sonarsource: Solved :)

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/EnvironmentalVariableScope.cs
pavel-mikula-sonarsource: Let's use `using var _ =` everywhere instead of `using var scope =`. We don't need that variable.

analyzers/tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
```diff
@@ -117,48 +116,49 @@ public void SonarDiagnosticAnalyzer_IsUsedInAllRules()
...
+            scope.SetVariable(SonarDiagnosticAnalyzer.EnableConcurrentExecutionVariable, value);
+            var reader = new ConcurrentExecutionReader();
+            reader.IsConcurrentExecutionEnabled.Should().BeNull();
+            Verifier.VerifyNoExceptionThrown("TestCases\\AsyncVoidMethod.cs", new[] { reader });
+            reader.IsConcurrentExecutionEnabled.Should().BeFalse();
         }
 
         [TestMethod]
-        public void Verify_ConcurrentProcessingGetsDisabledOnWrongValue()
+        public void Verify_ConcurrentExecutionIsDisabledForInvalidValue()
```
pavel-mikula-sonarsource, line 154: This can be now deleted and the test case added as `DataRow` above

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/DebugHelperTest.cs
```diff
@@ -44,8 +43,8 @@ public class DebugHelperTest
         private static void SetupRunAndReset(string value, bool shouldBeTrue)
         {
             // Arrange
-            var oldValue = Environment.GetEnvironmentVariable(DebugHelper.AnalyzerInternalDebugEnvVariable);
-            Environment.SetEnvironmentVariable(DebugHelper.AnalyzerInternalDebugEnvVariable, value);
+            using var scope = new EnvironmentVariableScope();
+            scope.SetVariable(DebugHelper.AnalyzerInternalDebugEnvVariable, value);
```
pavel-mikula-sonarsource, line 47: Step 1: This could be really useful as a get/set property on the scope.
Step 2: Step back, investigate
Step 3: We can just delete this entire file altogether with `DebutgHelper.cs` It's never used.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/EnvironmentalVariableScope.cs
```diff
@@ -0,0 +1,86 @@
...
+            set => SetVariable(SonarDiagnosticAnalyzer.EnableConcurrentExecutionVariable, value.ToString());
+        }
+
+        public bool InternalLogCBDE
+        {
+            set => SetVariable("SONAR_DOTNET_INTERNAL_LOG_CBDE", value.ToString());
+        }
+#pragma warning restore S2376
+
+        public EnvironmentVariableScope(bool setVariablesOnlyInAzureDevOpsContext = false) =>
```
pavel-mikula-sonarsource, line 48: Having `false` is only a very special testcase. Default behavior for all UTs is `true`
```suggestion
        public EnvironmentVariableScope(bool setVariablesOnlyInAzureDevOpsContext = true) =>
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -261,6 +282,39 @@ private static void VerifyBuildErrors(ImmutableArray<Diagnostic> diagnostics, Co
 
         private static string IssueType(bool isPrimary) => isPrimary ? "primary" : "secondary";
 
+        internal class File
+        {
+            private string FileName { get; }
+            private SourceText SourceText { get; }
```
pavel-mikula-sonarsource, line 288: This doesn't have to be field now, as it's hidden. I've also suggested to rename the second one.
```suggestion
            private readonly string fileName;
            private readonly SourceText content;
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -261,6 +282,39 @@ private static void VerifyBuildErrors(ImmutableArray<Diagnostic> diagnostics, Co
...
+                SourceText = SourceText.From(System.IO.File.ReadAllText(fileName));
+            }
+
+            public File(SyntaxTree syntaxTree)
+            {
+                FileName = syntaxTree.FilePath;
+                SourceText = syntaxTree.GetText();
+            }
+
+            public FileNameIssueLocations ToIssueLocations() =>
```
pavel-mikula-sonarsource, line 302: If we keep it, we should be more descriptive about the returned value
```suggestion
            public FileNameIssueLocations ToExpectedIssueLocations() =>
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -261,6 +282,39 @@ private static void VerifyBuildErrors(ImmutableArray<Diagnostic> diagnostics, Co
...
+                new (FileName, IssueLocationCollector.GetExpectedIssueLocations(SourceText.Lines));
+        }
+
+        internal class FileNameIssueLocations
```
pavel-mikula-sonarsource, line 306: For simplicity, this class is friend with the one above :) 
```suggestion
        internal class FileIssueLocations
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -54,83 +57,180 @@ public static void VerifyNoExceptionThrown(string path, IEnumerable<DiagnosticAn
...
         {
             var solution = SolutionBuilder.Create().AddProject(AnalyzerLanguage.CSharp).AddSnippet(snippet).AddReferences(additionalReferences).GetSolution();
-            VerifyAnalyzer(solution, new DiagnosticAnalyzer[] { diagnosticAnalyzer }, options, checkMode);
+            CompileAndVerifyAnalyzer(solution, new DiagnosticAnalyzer[] { diagnosticAnalyzer }, options, checkMode);
         }
 
+        /// <summary>
+        /// Verify analyzer from C# on a series of snippets in non-concurrent execution mode.
+        /// </summary>
+        public static void VerifyCSharpAnalyzer(string[] snippets,
```
pavel-mikula-sonarsource, line 93: This one exists only for testing another class `DiagnosticVerifier`. Let's remove it and do the minimal scaffolding in `DiagnosticVerifierTest`.This does way more things that we actually need to test that functionality.

As alternative, you can use another method and move those two snippets to a TestCases/somewhere

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -256,40 +355,48 @@ static IEnumerable<TMessage> ReadProtobuf(string path)
         }
 
         public static void VerifyCodeFix(string path,
-                                        string pathToExpected,
-                                        SonarDiagnosticAnalyzer diagnosticAnalyzer,
-                                        SonarCodeFixProvider codeFixProvider,
-                                        IEnumerable<ParseOptions> options = null,
-                                        OutputKind outputKind = OutputKind.DynamicallyLinkedLibrary,
-                                        IEnumerable<MetadataReference> additionalReferences = null) =>
+                                         string pathToExpected,
```
pavel-mikula-sonarsource, line 358: Line 323 above: `VerifyNoIssueReportedFromCSharp9InTest` seems to became unused and can be removed now.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -298,16 +405,60 @@ static IEnumerable<TMessage> ReadProtobuf(string path)
...
                                            IEnumerable<MetadataReference> additionalReferences,
                                            string sonarProjectConfigPath = null)
+        {
+            using var scope = new EnvironmentVariableScope(true) { EnableConcurrentAnalysis = true};
+            CreateConcurrencyTest(paths, out var pathsWithConcurrencyTests);
+            var solution = SolutionBuilder.CreateSolutionFromPaths(pathsWithConcurrencyTests, outputKind, additionalReferences);
+            CompileAndVerifyAnalyzer(solution, diagnosticAnalyzers, options, checkMode, sonarProjectConfigPath);
+        }
+
+        private static void CreateConcurrencyTest(IEnumerable<string> paths, out List<string> resultingPaths)
```
pavel-mikula-sonarsource, line 415: Please change `out` to a return type instead of `void`. Variable name can be just `ret`

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -298,16 +405,60 @@ static IEnumerable<TMessage> ReadProtobuf(string path)
...
+        {
+            resultingPaths = new List<string>(paths);
+            var language = AnalyzerLanguage.FromPath(paths.First());
+            foreach (var path in paths)
+            {
+                var fullName = new FileInfo(path).FullName;
```
pavel-mikula-sonarsource, line 421: Better API
```suggestion
                var fullName = Path.GetFullPath(path)
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -298,16 +405,60 @@ static IEnumerable<TMessage> ReadProtobuf(string path)
...
+                var content = File.ReadAllText(fullName, Encoding.UTF8);
+                if (language == AnalyzerLanguage.CSharp)
+                {
+                    var newPath = $"{fullName}.cs";
```
pavel-mikula-sonarsource, line 425: It's not very clear what happend in `Something.cs.cs` by the extra `.cs`. I liked the old name more. Please add something explanatory like `.Concurrent.cs`

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -298,16 +405,60 @@ static IEnumerable<TMessage> ReadProtobuf(string path)
...
+                    resultingPaths.Add(newPath);
+                }
+            }
+        }
+
+        private static string InsertNamespaceForVB(string content)
+        {
+            var match = Regex.Match(content, @"^\s*Imports\s+.+$", RegexOptions.Multiline | RegexOptions.RightToLeft);
+            var idx = match.Success ? match.Index + match.Length + 1 : 0;
+            return content.Insert(idx, "Namespace AppendedNamespaceForConcurrencyTest\n") + "\nEnd Namespace";
```
pavel-mikula-sonarsource, line 442: While C# version preserves line numbers, this messes them up for VB. This should do the trick and maybe re-enable some affected test cases.

```suggestion
            return content.Insert(idx, "Namespace AppendedNamespaceForConcurrencyTest : ") + " : End Namespace";
```

pavel-mikula-sonarsource (changes requested):
Round 3, only minor polishments this time

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/DebugHelperTest.cs
dmitry-golovinov-sonarsource: There is no need to take step 1 and 2 if you have the third one :) #lessismore

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
dmitry-golovinov-sonarsource: I don't think that we've disabled any VB case regarding that one. But added it for the simplicitly to detect the locations in original files

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
pavel-mikula-sonarsource: > I've also suggested to rename the second one.

😃

dmitry-golovinov-sonarsource: Done :)

analyzers/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
```diff
@@ -47,6 +47,8 @@
 
   <ItemGroup>
     <Compile Remove="TestCases\**\*" />
+    <Compile Include="TestCases\DiagnosticsVerifier\ExpectedIssuesNotRaised.cs" />
+    <Compile Include="TestCases\DiagnosticsVerifier\ExpectedIssuesNotRaised2.cs" />
```
pavel-mikula-sonarsource, line 51: This should not be here, the `<None..>` line below will take  care of them
```suggestion
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
```diff
@@ -298,16 +385,63 @@ static IEnumerable<TMessage> ReadProtobuf(string path)
...
+                    var newPath = $"{folder}{nameWithoutExtension}.Concurrent.cs";
+                    File.WriteAllText(newPath, $"namespace AppendedNamespaceForConcurrencyTest {{ {content} }}");
+                    ret.Add(newPath);
+                }
+                else
+                {
+                    var newPath = $"{folder}{nameWithoutExtension}.Concurrent.vb";
+                    File.WriteAllText(newPath, InsertNamespaceForVB(content));
+                    ret.Add(newPath);
+                }
```
pavel-mikula-sonarsource, line 416: We can simplify this 
```suggestion
                var sourcePath = Path.GetFullPath(path);
                var newPath = Path.Combine(Path.GetDirectoryName(sourcePath), Path.GetFileNameWithoutExtension(path) + ".Concurrent" + Path.GetExtension(path));
                var content = File.ReadAllText(sourcePath, Encoding.UTF8);
                File.WriteAllText(newPath, language == AnalyzerLanguage.CSharp ? $"namespace AppendedNamespaceForConcurrencyTest {{ {content} }}" : InsertNamespaceForVB(content));
                ret.Add(newPath);
```

pavel-mikula-sonarsource (changes requested):
Round 4, last one on the active code. I'll take a look at `Rules` and `TestCases` tomorrow.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Verifier.cs
dmitry-golovinov-sonarsource: 🥇 awesome, indeed :)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ReferenceEqualityCheckWhenEqualsExists2.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+
+namespace AppendedNamespaceForConcurrencyTest
+{
+    namespace Tests.Diagnostics
```
pavel-mikula-sonarsource, line 8: Same as on other places, to persist ident 
```suggestion
    namespace AppendedNamespaceForConcurrencyTestTests.Diagnostics
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ShiftDynamicNotInteger2.vb
```diff
@@ -0,0 +1,34 @@
+﻿Option Strict Off
+Namespace AppendedNamespaceForConcurrencyTest
+    Namespace Tests.Diagnostics
```
pavel-mikula-sonarsource, line 3: As on other places
```suggestion
Namespace AppendedNamespaceForConcurrencyTest.Tests.Diagnostics
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SingleStatementPerLine2.vb
```diff
@@ -0,0 +1,53 @@
+﻿Namespace Tests.Diagnostics2
```
pavel-mikula-sonarsource, line 1: Almost 😆 
```suggestion
﻿Namespace AppendedNamespaceForConcurrencyTest.Tests.Diagnostics
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/UnnecessaryUsings2.cs
```diff
@@ -0,0 +1,225 @@
+﻿using System;
```
pavel-mikula-sonarsource, line 1: This one is quite far from its sibling. Can you please uniform them more using the namespace prefix?
For example the `ReferencesInsideDocumentationTags` would be good to preserve.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SerializationConstructorsShouldBeSecured_Part1.cs
```diff
@@ -6,7 +6,7 @@
 namespace MyLibrary
 {
 
-    // Noncompliant - in Part2.cs
+
```
pavel-mikula-sonarsource, line 9: What are the empty lines for?

pavel-mikula-sonarsource (changes requested):
Round 5 - `Rules` and `TestCases`

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/AvoidExcessiveInheritance_Concurrent.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Tests.Foo.Bar.Baz
+{
+    public class Tests_5 : Tests.Diagnostics.DerivedClass_4 { }
+
+    public class Tests_6 : Tests.Diagnostics.DerivedClass_5 { } // Noncompliant
+
+    public class Tests_7 : Tests.Diagnostics.SubClass_5 { } // Noncompliant
+}
+
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Let's turn the concurrency for this rule off. It definitely needs some investigation, will create a ticket, seems like it performs incorrectly on double-nested namespaces.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotDecreaseMemberVisibility_Concurrent.cs
```diff
@@ -0,0 +1,201 @@
...
+        public int Property_10 { get; set; }
+    }
+
+    class B : A
+    {
+    }
+
+    class C : B
+    {
+        private void Method_01(int count) { } // Noncompliant
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: You cannot have here a message, since in 2 versions of file it was removed. that was the only reason of keeping this file without messages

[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Solved it another way

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotHideBaseClassMethods_Concurrent.cs
```diff
@@ -0,0 +1,76 @@
...
+        private void SomePrivateMethod(string s1, string s2) { }
+
+        public void GenericMethod<T>(T s1, string s2) { }
+
+        public void GenericMethod2<T>(IEnumerable<T> s1, string s2) { }
+    }
+
+    class Bar : Foo
+    {
+        public void SomePublicMethod(string s1, object s2) { } // Noncompliant
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: We won't be able to preserve message with autogenerated NS, since for 2 files it will be different.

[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Solved it with a duplicate test case

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/EmptyNestedBlock2.cs
```diff
@@ -0,0 +1,117 @@
+﻿#define MY_VAL
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Let's open the issue and merge this PR, so that it could be fixed in another PR.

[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: https://github.com/SonarSource/sonar-dotnet/issues/4747

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/IssueSuppression2.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("", "")] // Noncompliant
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: https://github.com/SonarSource/sonar-dotnet/issues/4747

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MarkAssemblyWithClsCompliantAttribute2.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System;
+
+[assembly: CLSCompliant(false)]  // Error [CS0579]
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Removed those cases. Confirmed that there are no other rules like that

analyzers/tests/SonarAnalyzer.UnitTest/Rules/MarkAssemblyWithAssemblyVersionAttributeTest.cs
```diff
@@ -34,40 +34,56 @@ public class MarkAssemblyWithAssemblyVersionAttributeTest
         [TestMethod]
         [TestCategory("Rule")]
         public void MarkAssemblyWithAssemblyVersionAttribute_CS() =>
-            Verifier.VerifyAnalyzer(@"TestCases\MarkAssemblyWithAssemblyVersionAttribute.cs", new CS.MarkAssemblyWithAssemblyVersionAttribute());
+            Verifier.VerifyNonConcurrentAnalyzer(@"TestCases\MarkAssemblyWithAssemblyVersionAttribute.cs", new CS.MarkAssemblyWithAssemblyVersionAttribute());
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: They could not be reverted, since otherwise the code of VerifyAnalyzer would duplicate the files. I've removed double files cases

analyzers/tests/SonarAnalyzer.UnitTest/Rules/MarkAssemblyWithClsCompliantAttributeTest.cs
```diff
@@ -33,26 +33,28 @@ public class MarkAssemblyWithClsCompliantAttributeTest
         [TestMethod]
         [TestCategory("Rule")]
         public void MarkAssemblyWithClsCompliantAttribute_CS() =>
-            Verifier.VerifyAnalyzer(@"TestCases\MarkAssemblyWithClsCompliantAttribute.cs", new CS.MarkAssemblyWithClsCompliantAttribute());
+            Verifier.VerifyAnalyzerNoDuplication(new[] { @"TestCases\MarkAssemblyWithClsCompliantAttribute.cs", @"TestCases\MarkAssemblyWithClsCompliantAttribute2.cs", },
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: They could not be reverted, since otherwise the code of VerifyAnalyzer would duplicate the files. I've removed double files cases

analyzers/tests/SonarAnalyzer.UnitTest/Rules/MarkAssemblyWithComVisibleAttributeTest.cs
```diff
@@ -33,26 +33,28 @@ public class MarkAssemblyWithComVisibleAttributeTest
         [TestMethod]
         [TestCategory("Rule")]
         public void MarkAssemblyWithComVisibleAttribute_CS() =>
-            Verifier.VerifyAnalyzer(@"TestCases\MarkAssemblyWithComVisibleAttribute.cs", new CS.MarkAssemblyWithComVisibleAttribute());
+            Verifier.VerifyAnalyzerNoDuplication(new[] { @"TestCases\MarkAssemblyWithComVisibleAttribute.cs", @"TestCases\MarkAssemblyWithComVisibleAttribute2.cs", },
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: They could not be reverted, since otherwise the code of VerifyAnalyzer would duplicate the files.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MarkAssemblyWithClsCompliantAttribute2.vb
```diff
@@ -0,0 +1,3 @@
+﻿Imports System
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: File got removed

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MarkAssemblyWithComVisibleAttribute2.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.InteropServices;
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: File got removed

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MarkAssemblyWithComVisibleAttribute2.vb
```diff
@@ -0,0 +1,3 @@
+﻿Imports System.Runtime.InteropServices
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: File got removed

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MethodsShouldUseBaseTypes2.cs
```diff
@@ -0,0 +1,896 @@
...
+        protected int protectedField_2;
+    }
+
+    public class B : A
+    {
+        public int publicField;
+
+        new protected int protectedField;
+
+        public void Test_111(B foo)
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: The file was 800+ lines long, so it was a massive renaming from Test_ to Test_1

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SerializationConstructorsShouldBeSecured_Part1.cs
dmitry-golovinov-sonarsource: That was an old trick to overcome the fact that issues were not checked in the correct files

analyzers/tests/SonarAnalyzer.UnitTest/Rules/UnnecessaryUsingsTest.cs
```diff
@@ -36,9 +36,9 @@ public class UnnecessaryUsingsTest
         [TestMethod]
         [TestCategory("Rule")]
         public void UnnecessaryUsings() =>
-            Verifier.VerifyAnalyzer(@"TestCases\UnnecessaryUsings.cs",
+            Verifier.VerifyAnalyzer(new[] { @"TestCases\UnnecessaryUsings.cs", @"TestCases\UnnecessaryUsings2.cs" },
                                     new UnnecessaryUsings(),
-                                    GetAdditionalReferences());
+                                    additionalReferences: GetAdditionalReferences());
```
pavel-mikula-sonarsource, line 41: Not needed
```suggestion
                                    GetAdditionalReferences());
```

pavel-mikula-sonarsource (approved):
LGTM! There's one last minor comment to fix before merging.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/FileShouldEndWithEmptyNewLineTest.cs
```diff
@@ -29,17 +29,13 @@ public class FileShouldEndWithEmptyNewLineTest
...
-        [TestMethod]
-        [TestCategory("Rule")]
-        public void FileShouldEndWithEmptyNewLine_NoEmptyLine() =>
-            Verifier.VerifyAnalyzer(@"TestCases\FileShouldEndWithEmptyNewLine_NoEmptyLine.cs", new FileShouldEndWithEmptyNewLine());
-
-        [TestMethod]
-        [TestCategory("Rule")]
-        public void FileShouldEndWithEmptyNewLine_EmptyFile() =>
-            Verifier.VerifyAnalyzer(@"TestCases\FileShouldEndWithEmptyNewLine_EmptyFile.cs", new FileShouldEndWithEmptyNewLine());
+        public void FileShouldEndWithEmptyNewLine_Concurrent() =>
```
pavel-mikula-sonarsource, line 32: Not needed
```suggestion
        public void FileShouldEndWithEmptyNewLine() =>
```

pavel-mikula-sonarsource (approved):

PR #4716, opened 2021-07-26
Rule S2184: add support for native integers

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/LossOfFractionInDivision.cs
```diff
@@ -32,46 +32,40 @@ namespace SonarAnalyzer.Rules.CSharp
...
-                        TryGetTypeFromReturnMappedToFloatType(division, c.SemanticModel, out assignedToType))
+                    if (TryGetTypeFromAssignmentToFloatType(division, c.SemanticModel, out var assignedToType)
+                        || TryGetTypeFromArgumentMappedToFloatType(division, c.SemanticModel, out assignedToType)
+                        || TryGetTypeFromReturnMappedToFloatType(division, c.SemanticModel, out assignedToType))
                     {
-                        c.ReportDiagnosticWhenActive(Diagnostic.Create(
-                            rule,
-                            division.GetLocation(),
-                            assignedToType.ToMinimalDisplayString(c.SemanticModel, division.SpanStart)));
+                        var diagnostic = Diagnostic.Create(Rule, division.GetLocation(), assignedToType.ToMinimalDisplayString(c.SemanticModel, division.SpanStart));
```
csaba-sagi-sonarsource, line 59: Why create a variable if the value is used only once?

csaba-sagi-sonarsource (approved):
Overall LGTM, please take a look at the comment I left.

analyzers/src/SonarAnalyzer.CSharp/Rules/LossOfFractionInDivision.cs
costin-zaharia-sonarsource: To avoid having a long line with many method calls inside method invocations.

PR #4717, opened 2021-07-26
Rule S3052: add support for native integers

costin-zaharia-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/Extensions/PropertyDeclarationSyntaxExtensionsTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        [TestMethod]
+        public void IsAutoProperty_AccessorWithExpressionBody_ReturnsFalse()
+        {
+            const string code = @"record TestCases
+{
+    public int Property
+    {
+        get => 0;
```
andrei-epure-sonarsource, line 52: could you also add a test with an implemented `set` and `init` ?

andrei-epure-sonarsource (changes requested):

andrei-epure-sonarsource:
Also, could you please add a test case in `analyzers\its\sources\Net5\Net5` - it will help with Peach validation as well (https://peach.sonarsource.com/code?id=net5-project)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MemberInitializedToDefault.CSharp9.cs
```diff
@@ -1,15 +1,15 @@
 ﻿record NativeInts
 {
     public const nint myConst = 0;  // Compliant
-    public nint field1 = 0;         // FN
+    public nint field1 = 0;         // Noncompliant
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Nice!

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/PropertyDeclarationSyntaxExtensionsTests.cs
costin-zaharia-sonarsource: As far as I can tell this test would be redundant since all accessors are treated the same; no matter their type, we check if they have a body or expression body and this case is covered by `IsAutoProperty_AccessorWithBody_ReturnsFalse`

I prefer not to add this test.

andrei-epure-sonarsource: You are testing thinking of the implementation instead of the functionality...
If someone will come and change to only test getters, should a test fail?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MemberInitializedToDefault.CSharp9.cs
```diff
@@ -1,10 +1,16 @@
...
 {
     public const nint myConst = 0;  // Compliant
     public nint field1 = 0;         // Noncompliant
     public nint field2 = 42;
     public nuint field3 = 0;        // Noncompliant
     public nuint field4 = 42;
+    public nint field5 = IntPtr.Zero;         // Noncompliant
+    public nint field6 = 0x0000000000000000;  // Noncompliant
+    public nuint field7 = UIntPtr.Zero;       // Noncompliant
+    public nint field8 = new IntPtr(0);       // Noncompliant
```
andrei-epure-sonarsource, line 13: you can add a test with `new IntPtr(someNintArgument)` to improve ccov

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/PropertyDeclarationSyntaxExtensionsTests.cs
costin-zaharia-sonarsource: As far as I can see, it is a compromise. In general checking all possible cases take more time to run (running more cases) and more time to maintain the tests so this comes with a cost we should not ignore.

From the other point of view. if the code is changed but not covered by tests we will have low coverage and we will have to do something anyway. Our threshold is 95% which is sensible enough to spot this kind of changes in most of the cases.

andrei-epure-sonarsource: In the case that I mentioned, there will be no change in code coverage: only getters are tested, so if someone changes the code to test only the getters, coverage will be 100%.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MemberInitializedToDefault.CSharp9.cs
```diff
@@ -1,15 +1,23 @@
...
+    public nuint field3 = 0;        // Noncompliant
     public nuint field4 = 42;
+    public nint field5 = IntPtr.Zero;         // Noncompliant
+    public nint field6 = 0x0000000000000000;  // Noncompliant
+    public nuint field7 = UIntPtr.Zero;       // Noncompliant
+    public nint field8 = new IntPtr(0);       // Noncompliant
+    public nint field9 = new IntPtr(staticField);
 
-    public nint Property1 { get; set; } = 0;    // FN
+    public nint Property1 { get; set; } = 0;    // Noncompliant
```
andrei-epure-sonarsource, line 17: [coverage]
to try and force [the property symbol to be null](https://sonarcloud.io/code?id=sonaranalyzer-dotnet&pullRequest=4717&selected=sonaranalyzer-dotnet%3Aanalyzers%2Fsrc%2FSonarAnalyzer.CSharp%2FRules%2FMemberInitializedToDefault.cs&line=67) you could add a test with an UnknownType property (which can happen in SonarLint when coding in the IDE) and see if the returned symbol is null

andrei-epure-sonarsource (approved):
LGTM

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/PropertyDeclarationSyntaxExtensionsTests.cs
costin-zaharia-sonarsource: As discussed also offline, if we change the code to consider only getters, most probably we will do an equality check which will create a logic branch uncovered by tests.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MemberInitializedToDefault.CSharp9.cs
costin-zaharia-sonarsource: This flow is a bit harder to reproduce since `GetDeclaredSymbol` will not return null but `NativeInts.PropertyName` in this case.

PR #4725, opened 2021-07-27
Fix S4830: CertificateValidationCheck throws NullReferenceException

### LINKED ISSUE #4710: AD0001: Analyzer SonarAnalyzer.Rules.CSharp.CertificateValidationCheck throws NullReferenceException
### Description

Using any version of SonarAnalyzer.CSharp NuGet package after 8.22.0.31243 produces this warning:
```
CSC : warning AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.CertificateValidationCheck' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'. [\Steeltoe\src\Common\src\Common.Http\Steeltoe.Common.Http.csproj]
```

### Repro steps

Simplified version of the real code, but this is what causes the exception:
```csharp
        public static void RestoreCertificateValidation(RemoteCertificateValidationCallback prevValidator)
        {
                ServicePointManager.ServerCertificateValidationCallback = prevValidator;
        }
```

### Known workarounds

Remove the code above or use 8.22.0.31243 or older

### Related information

* C#/VB.NET Plugins version
* Visual Studio version
* MSBuild / dotnet version: MS Build version 17.0.0-preview-21302-02+018bed83d or 16.10.2+857e5a733
* SonarScanner for .NET version (if used)
* Operating System: Win10 or Azure DevOps Hosted linux build agent

csaba-sagi-sonarsource:
Fixes #4710

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
@@ -123,5 +123,9 @@ namespace SonarAnalyzer.Rules.CSharp
             reference.GetSyntax();   // VB.NET has more complicated logic
 
         private protected override KnownType GenericDelegateType() => KnownType.System_Func_T1_T2_T3_T4_TResult;
+
+        private protected override bool AreSameMethods(InvocationExpressionSyntax invocation, string methodName) =>
+            invocation.Expression.GetIdentifier()?.Identifier.ValueText is { } invocationName
+            && invocationName.Equals(methodName);
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
@@ -123,9 +123,5 @@ namespace SonarAnalyzer.Rules.CSharp
             reference.GetSyntax();   // VB.NET has more complicated logic
 
         private protected override KnownType GenericDelegateType() => KnownType.System_Func_T1_T2_T3_T4_TResult;
-
-        private protected override bool AreSameMethods(InvocationExpressionSyntax invocation, string methodName) =>
-            invocation.Expression.GetIdentifier()?.Identifier.ValueText is { } invocationName
-            && invocationName.Equals(methodName);
     }
 }
--- a/analyzers/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs
@@ -308,7 +308,8 @@ namespace SonarAnalyzer.Rules
             var ret = ImmutableArray.CreateBuilder<TInvocationExpressionSyntax>();
             foreach (var invocation in root.DescendantNodesAndSelf().OfType<TInvocationExpressionSyntax>())
             {
-                if (c.SemanticModel.GetSymbolInfo(invocation).Symbol.Equals(method))
+                if (c.SemanticModel.GetSymbolInfo(invocation).Symbol is { } symbol
+                    && symbol.Equals(method))
                 {
                     ret.Add(invocation);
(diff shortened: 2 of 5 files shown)


analyzers/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs
```diff
@@ -308,7 +308,8 @@ private static ImmutableArray<TInvocationExpressionSyntax> FindInvocationList(Sy
             var ret = ImmutableArray.CreateBuilder<TInvocationExpressionSyntax>();
             foreach (var invocation in root.DescendantNodesAndSelf().OfType<TInvocationExpressionSyntax>())
             {
-                if (c.SemanticModel.GetSymbolInfo(invocation).Symbol.Equals(method))
+                if (c.SemanticModel.GetSymbolInfo(invocation).Symbol is { } symbol
```
pavel-mikula-sonarsource, line 311: For performance: Let's add a syntax-based name check first.

(I'm not reviewing, I was just curious about the NRE)

analyzers/src/SonarAnalyzer.CSharp/Rules/CertificateValidationCheck.cs
```diff
@@ -123,5 +123,9 @@ protected override ImmutableArray<Location> LambdaLocations(InspectionContext c,
             reference.GetSyntax();   // VB.NET has more complicated logic
 
         private protected override KnownType GenericDelegateType() => KnownType.System_Func_T1_T2_T3_T4_TResult;
+
+        private protected override bool AreSameMethods(InvocationExpressionSyntax invocation, string methodName) =>
```
pavel-mikula-sonarsource, line 127: This should be doable without an abstract method. Either with `node.ToStringContains(...)` or even better `Language.Syntax.InvocationIdentifier()`

analyzers/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs
```diff
@@ -308,7 +309,9 @@ private static ImmutableArray<TInvocationExpressionSyntax> FindInvocationList(Sy
             var ret = ImmutableArray.CreateBuilder<TInvocationExpressionSyntax>();
             foreach (var invocation in root.DescendantNodesAndSelf().OfType<TInvocationExpressionSyntax>())
             {
-                if (c.SemanticModel.GetSymbolInfo(invocation).Symbol.Equals(method))
+                if (AreSameMethods(invocation, method.Name)
+                    && c.SemanticModel.GetSymbolInfo(invocation).Symbol is { } symbol
```
pavel-mikula-sonarsource, line 313: With the name check in place, `null` should be recoverable with an undefined method and some `// Error` flag.

pavel-mikula-sonarsource (changes requested):
Too much code 😆

analyzers/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs
```diff
@@ -308,7 +308,10 @@ private static ImmutableArray<TInvocationExpressionSyntax> FindInvocationList(Sy
             var ret = ImmutableArray.CreateBuilder<TInvocationExpressionSyntax>();
             foreach (var invocation in root.DescendantNodesAndSelf().OfType<TInvocationExpressionSyntax>())
             {
-                if (c.SemanticModel.GetSymbolInfo(invocation).Symbol.Equals(method))
+                if (Language.Syntax.InvocationIdentifier(invocation) is { } invocationIdentifier
+                    && invocationIdentifier.ValueText.Equals(method.Name)
```
pavel-mikula-sonarsource, line 312: We'll need to use the `Language.NameComparison` here and add VB.NET test case with different casing.

pavel-mikula-sonarsource (changes requested):
One last detail

pavel-mikula-sonarsource (approved):
LGTM

PR #4726, opened 2021-07-27
Generate Protobuf structure for logging

### LINKED ISSUE #4720: Generate Protobuf structure for logging
Part of https://jira.sonarsource.com/browse/MMF-2395

Create scaffolding for logging in .NET and Java

pavel-mikula-sonarsource:
Fixes #4720


analyzers/src/SonarAnalyzer.Common/Protobuf/AnalyzerReport.proto
```diff
@@ -107,3 +107,17 @@ message FileMetadataInfo {
   bool is_generated = 2;
   string encoding = 3;
 }
+
+// Logging
+enum LogSeverity {
+  UNKNOWN_SEVERITY = 0;
+  DEBUG = 1;
+  INFO = 2;
+  WARNING = 3;
```
csaba-sagi-sonarsource, line 116: According to the MMF the severity should be INFO, ERROR or DEBUG.

pavel-mikula-sonarsource: Good catch, MMF is fixed now :D We don't want to fail the analysis (yet), so ERROR shouldn't be there.

PR #4728, opened 2021-07-27
Rule S2437: add support for native integers

costin-zaharia-sonarsource:

### DIFF
--- a/analyzers/src/SonarAnalyzer.VisualBasic/Trackers/VisualBasicConstantValueFinder.cs
+++ b/analyzers/src/SonarAnalyzer.VisualBasic/Trackers/VisualBasicConstantValueFinder.cs
@@ -21,6 +21,7 @@
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.VisualBasic;
 using Microsoft.CodeAnalysis.VisualBasic.Syntax;
+using SonarAnalyzer.Extensions;
 
 namespace SonarAnalyzer.Helpers
 {
@@ -36,5 +37,9 @@ namespace SonarAnalyzer.Helpers
 
         protected override VariableDeclaratorSyntax VariableDeclarator(SyntaxNode node) =>
             node?.Parent as VariableDeclaratorSyntax;
+
+        protected override bool IsPtrZero(SyntaxNode node) =>
+            node is MemberAccessExpressionSyntax memberAccess
+            && memberAccess.IsPtrZero(SemanticModel);
     }
 }
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LdapConnectionShouldBeSecure.CSharp9.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LdapConnectionShouldBeSecure.CSharp9.cs
@@ -8,12 +8,12 @@ var authTypeSecure = AuthenticationTypes.Secure;
 new DirectoryEntry("path", "user", "pass", authTypeSecure); // Compliant
 
 var authTypeNone = AuthenticationTypes.None;
-new DirectoryEntry("path", "user", "pass", authTypeNone);   // FN
+new DirectoryEntry("path", "user", "pass", authTypeNone);   // Noncompliant
 
 DirectoryEntry entry1 = new("path", "user", "pass", AuthenticationTypes.Secure);    // Compliant
 DirectoryEntry entry2 = new("path", "user", "pass", AuthenticationTypes.None);      // Noncompliant
 DirectoryEntry entry3 = new("path", "user", "pass", authTypeSecure);    // Compliant
-DirectoryEntry entry4 = new("path", "user", "pass", authTypeNone);      // FN
+DirectoryEntry entry4 = new("path", "user", "pass", authTypeNone);      // Noncompliant
 
 record FieldsAndProperties
 {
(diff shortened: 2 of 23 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SillyBitwiseOperation.CSharp9.cs
```diff
@@ -24,11 +27,11 @@
...
 nuint bitMaskNuint = 0x010F;
 nuint resultNuint;
-nuint nuOne = 1;
-nuint nuZero = 0;
+const nuint nuZero = 0;
 
-resultNuint = bitMaskNuint & + + +nuOne;    // FN
+resultNuint = bitMaskNuint | + + +nuZero;   // Noncompliant
 resultNuint = bitMaskNuint & nuZero;        // Compliant
+resultNuint = bitMaskNuint ^ 0;             // Noncompliant
```
andrei-epure-sonarsource, line 34: you can also add tests with `UIntPtr.Zero` and `IntPtr.Zero;`

costin-zaharia-sonarsource: I'm not sure I understand this. Operators `&`, `|` and `^` are not defined for `UIntPtr` and `IntPtr`. Do I miss something?

andrei-epure-sonarsource: This builds on my machine:
```
        nuint x = 1;
        x = x ^ UIntPtr.Zero;
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LdapConnectionShouldBeSecure.CSharp9.cs
```diff
@@ -8,12 +8,12 @@
 new DirectoryEntry("path", "user", "pass", authTypeSecure); // Compliant
 
 var authTypeNone = AuthenticationTypes.None;
-new DirectoryEntry("path", "user", "pass", authTypeNone);   // FN
+new DirectoryEntry("path", "user", "pass", authTypeNone);   // Noncompliant
```
andrei-epure-sonarsource, line 11: maybe you can remove these rules (Ldap, jwt) from the top-level support MMF?

andrei-epure-sonarsource (changes requested):
do you think it's worth supporting `x = x ^ UIntPtr.Zero;`  ?
I imagine there's a cast being done in the expression and the extra operations get supported by the framework type

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SillyBitwiseOperation.CSharp9.cs
costin-zaharia-sonarsource: Ah, my bad, I've applied `^` operator to a mix of nint and `UIntPtr`.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LdapConnectionShouldBeSecure.CSharp9.cs
costin-zaharia-sonarsource: None of these (S4433, S5659) is part of the [top-level support MMF](https://jira.sonarsource.com/browse/MMF-2222)

analyzers/src/SonarAnalyzer.VisualBasic/Trackers/VisualBasicConstantValueFinder.cs
```diff
@@ -36,5 +37,9 @@ public class VisualBasicConstantValueFinder : ConstantValueFinder<IdentifierName
 
         protected override VariableDeclaratorSyntax VariableDeclarator(SyntaxNode node) =>
             node?.Parent as VariableDeclaratorSyntax;
+
+        protected override bool IsPtrZero(SyntaxNode node) =>
```
andrei-epure-sonarsource, line 41: is there a test for this?

Please add a test, I've seen previously code for `vb.net` w/o code coverage which proved to be faulty

why is this covered w/o a test?....

andrei-epure-sonarsource (approved):
LGTM

I don't understand why the VB .NET `IsPtrZero` is fully covered - could you please double check before merging, in case there's a tooling error?

analyzers/src/SonarAnalyzer.VisualBasic/Trackers/VisualBasicConstantValueFinder.cs
costin-zaharia-sonarsource: The way I see it, there are 2 branches:
- when node is not `MemberAccessExpressionSyntax`: this is tested when analyzing a node as `-1`, see for ex: https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SillyBitwiseOperation.vb#L19
- when node is `MemberAccessExpressionSyntax`: this is tested when analyzing a node as `Types.Class`, see for ex: https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SillyBitwiseOperation.vb#L49

Both of them are covered.

The last line `&& memberAccess.IsPtrZero(SemanticModel)` is not really a branch since the result is returned directly.

PR #4730, opened 2021-07-27
Log available Roslyn version

### LINKED ISSUE #4721: Log available Roslyn version

Introduce `LogUtilityAnalyzer` to generate `log.pb`
Log Roslyn version as `INFO` message.

pavel-mikula-sonarsource:
Fixes #4721

Review per commit is suggested


analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/LogAnalyzerTest.cs
```diff
@@ -0,0 +1,97 @@
...
+
+namespace SonarAnalyzer.UnitTest.Rules
+{
+    [TestClass]
+    public class LogAnalyzerTest
+    {
+        private const string Root = @"TestCases\Utilities\LogAnalyzer\";
+
+        [TestMethod]
+        [TestCategory("Rule")]
```
csaba-sagi-sonarsource, line 40: Maybe it would make sense to have TestCategory "Utility" for the utility analyzers

analyzers/src/SonarAnalyzer.CSharp/Rules/Utilities/TokenTypeAnalyzer.cs
```diff
@@ -55,42 +47,23 @@ public TokenClassifier(SyntaxToken token, SemanticModel semanticModel, bool skip
...
+                    SyntaxKind.CharacterLiteralToken,
+                    SyntaxKind.InterpolatedStringStartToken,
+                    SyntaxKind.InterpolatedVerbatimStringStartToken,
+                    SyntaxKind.InterpolatedStringTextToken,
+                    SyntaxKind.InterpolatedStringEndToken);
 
             protected override bool IsDocComment(SyntaxTrivia trivia) =>
-                trivia.IsKind(SyntaxKind.SingleLineDocumentationCommentTrivia) ||
-                trivia.IsKind(SyntaxKind.MultiLineDocumentationCommentTrivia);
+                trivia.IsAnyKind(SyntaxKind.SingleLineDocumentationCommentTrivia, SyntaxKind.MultiLineDocumentationCommentTrivia);
```
csaba-sagi-sonarsource, line 66: The is comment and is doc checks are present in multiple rules, maybe they deserve their place in the LanguageFacade and the IsDocComment method will not be neccessary.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs
```diff
@@ -19,37 +19,31 @@
...
 
         protected abstract TokenClassifierBase GetTokenClassifier(SyntaxToken token, SemanticModel semanticModel, bool skipIdentifierTokens);
 
-        protected sealed override string FileName => TokenTypeFileName;
-
         protected sealed override TokenTypeInfo CreateMessage(SyntaxTree syntaxTree, SemanticModel semanticModel)
         {
             var tokens = syntaxTree.GetRoot().DescendantTokens();
-            var skipIdentifierTokens = tokens.Count(token => token.RawKind == identifierTokenKind) > IdentifierTokenCountThreshold;
+            var identifierTokenKind = Language.SyntaxKind.IdentifierToken;
```
csaba-sagi-sonarsource, line 45: identifierTokenKind  is only used once, so it can be directly used from the LanguageFacade.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
```diff
@@ -34,13 +35,19 @@ public abstract class UtilityAnalyzerBase : SonarDiagnosticAnalyzer
...
 
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
         protected bool IsAnalyzerEnabled { get; set; }
         protected bool IgnoreHeaderComments { get; set; }
         protected virtual bool AnalyzeGeneratedCode { get; set; }
+        protected virtual bool AnalyzeTestProjects => true;
         protected string OutPath { get; set; }
         protected bool IsTestProject { get; set; }
 
+        protected UtilityAnalyzerBase(string diagnosticId, string title) =>
```
csaba-sagi-sonarsource, line 48: The diagnosticId and title could be protected virtual properties, and a bunch of constructor calls could be removed.

csaba-sagi-sonarsource (changes requested):
Please take a look at the comments I left.

analyzers/src/SonarAnalyzer.CSharp/Rules/Utilities/TokenTypeAnalyzer.cs
pavel-mikula-sonarsource: I'll do that in another PR. It will affect more rules.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs
pavel-mikula-sonarsource: It's a small performance optimization to avoid those property invocations. There could be thousands of tokens in each syntaxtree.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/UtilityAnalyzerBase.cs
pavel-mikula-sonarsource: I prefer constructor to construct the class in this case :) It syntactically enforces everyone to do it correctly in single line manner.

pavel-mikula-sonarsource: ..and we don't need to store nor access them afterwards from the class and it's inheritors so we don't need properties.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs
csaba-sagi-sonarsource: I am not sure the optimization will happen, the compiler can also detect that the variable is not used anywhere else, and just use it from the property.

csaba-sagi-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs
pavel-mikula-sonarsource: The property access behind the interface can have side-effect so the compiler cannot undo this, because it could change the semantics of the code.

PR #4733, opened 2021-07-28
Process analyzer log messages in the SQ plugin

### LINKED ISSUE #4722: Process analyzer log messages in the SQ plugin

Read `log.pb` files in a sensor and log them according to their severity.

pavel-mikula-sonarsource:
Fixes #4722


sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/LogSensorTest.java
```diff
@@ -53,4 +66,10 @@ public void should_describe() {
     assertThat(sensorDescriptor.name()).isEqualTo(SHORT_LANG_NAME + " Analysis Log");
     assertThat(sensorDescriptor.languages()).isEmpty();     // should not filter per language
   }
+
+  @Test
+  public void executeLogsMessages() {
+    sensor.execute(context);
+    assertThat(logTester.logs()).hasSize(4);
```
csaba-sagi-sonarsource, line 73: We could also assert the content of the logs, that would clarify why the size is 4

its/src/test/java/com/sonar/it/csharp/MultipleProjectsTest.java
```diff
@@ -74,6 +74,12 @@ public void projectTypesInfoIsLogged() {
     assertThat(buildResult.getLogs()).contains("Found 4 MSBuild C# projects: 2 MAIN projects. 2 TEST projects.");
   }
 
+  @Test
+  public void roslynVersionIsLogged() {
+    // FirstProject + FirstProjectTests + 2x SecondProject (each target framework has its log) + SecondProjectTests
+    assertThat(buildResult.getLogsLines(x -> x.startsWith("INFO: Roslyn version: "))).hasSize(5);
```
csaba-sagi-sonarsource, line 80: Would it make sense to check the roslyn version too? Just to make sure it is in the correct format.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/protobuf/LogImporterTest.java
```diff
@@ -29,26 +29,40 @@
...
+  }
+
+  @Test
+  public void unknownLogReportedAsInfoWithWarning() {
+    LogImporter sut = new LogImporter();
+    sut.accept(unknownProfobuf.toPath());
+    sut.save();
+
+    assertThat(logTester.logs(LoggerLevel.DEBUG)).isEmpty();
+    assertThat(logTester.logs(LoggerLevel.INFO)).containsOnly("Unknown severify for Coverage");
```
csaba-sagi-sonarsource, line 57: TypeO: severify

sonar-dotnet-shared-library/src/test/resources/ProtobufImporterTest/README.md
```diff
@@ -41,3 +41,11 @@ foreach (var message in messages)
     message.WriteDelimitedTo(metricsStream);
 }
 ```
+
+### unknown-log.pb
+
+Same as `custom-log.pb` with this message:
+
+```
+new LogInfo {Severity = LogSeverity.UnknownSeverity, Text = "Unknown severify for Coverage" }
```
csaba-sagi-sonarsource, line 50: TypeO: severify

sonar-dotnet-shared-library/src/test/resources/ProtobufImporterTest/unknown-log.pb
```diff
@@ -0,0 +1 @@
+Unknown severify for Coverage
```
csaba-sagi-sonarsource, line 1: Here it is also a typeO, hmmmm was it intentional?

csaba-sagi-sonarsource (changes requested):
Overall it looks very nice, just a few minor changes requested.

I also don`t like the fact that the LogImporter is called importer when it does more than just import, but it is seems to be the case for all the "Importers"

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/LogSensorTest.java
pavel-mikula-sonarsource: Reading the correct values is the responsibility of `LogImporter`. `LogSensor` is just responsible for orchestrating it so in this test, I just need to be sure that something happens. I used `hasSize(4)` instead of `isNotEmpty()` to avoid some random noise in the log.

its/src/test/java/com/sonar/it/csharp/MultipleProjectsTest.java
pavel-mikula-sonarsource: No, as that's already tested in UT. This integration makes sure the message just goes end-to-end.

sonar-dotnet-shared-library/src/test/java/org/sonarsource/dotnet/shared/plugins/protobuf/LogImporterTest.java
pavel-mikula-sonarsource: oh no 🤦

pavel-mikula-sonarsource:
> I also don`t like the fact that the LogImporter is called importer when it does more than just import, but it is seems to be the case for all the "Importers"

Yep, they are all importers. They read the data and then they import them in `save()` method. Conceptual flaw is actually the fact that half of them doesn't override the `save()` method to import data to SQ and does it in some other way 😃

csaba-sagi-sonarsource (approved):
LGTM!

PR #4734, opened 2021-07-28
Rule S3877: add support for throw expressions and module initializers

andrei-epure-sonarsource:
RSPEC PR : https://github.com/SonarSource/rspec/pull/231

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
@@ -144,13 +144,27 @@ namespace SonarAnalyzer.Rules.CSharp
             };
 
         private static bool IsTrackedMethod(MethodDeclarationSyntax declaration, SemanticModel semanticModel) =>
-            semanticModel.GetDeclaredSymbol(declaration) is { } methodSymbol
+            HasNameOrAttribute(declaration)
+            && semanticModel.GetDeclaredSymbol(declaration) is { } methodSymbol
             && (methodSymbol.IsObjectEquals()
                 || methodSymbol.IsObjectGetHashCode()
                 || methodSymbol.IsObjectToString()
                 || methodSymbol.IsIDisposableDispose()
                 || methodSymbol.IsIEquatableEquals()
                 || IsModuleInitializer(methodSymbol));
+
+        private static bool HasNameOrAttribute(MethodDeclarationSyntax declaration)
+        {
+            var name = declaration.Identifier.ValueText;
+            if (name == "Equals" || name == "GetHashCode" || name == "ToString" || name == "Dispose" || name == "Equals")
+            {
+                return true;
+            }
+            else
+            {
+                return declaration.AttributeLists.SelectMany(list => list.Attributes).Any(x => x.ArgumentList == null && x.Name.ToStringContains("ModuleInitializer"));
+            }
+        }
 
         private static bool IsModuleInitializer(IMethodSymbol methodSymbol) =>
             methodSymbol.AnyAttributeDerivesFrom(KnownType.System_Runtime_CompilerServices_ModuleInitializerAttribute);
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
@@ -28,6 +28,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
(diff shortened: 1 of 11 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
```diff
@@ -36,14 +37,13 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class ExceptionShouldNotBeThrownFromUnexpectedMethods : SonarDiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S3877";
```
costin-zaharia-sonarsource, line 39: we can make this private.

analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
```diff
@@ -36,14 +37,13 @@ namespace SonarAnalyzer.Rules.CSharp
...
-        private const string MessageFormat = "Remove this 'throw' statement.";
+        private const string MessageFormat = "Remove this 'throw' {0}.";
 
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
```
costin-zaharia-sonarsource, line 42: we should reorder the properties by their visibility.  `Rule` is private, then `SupportedDiagnostics` public and then again `DefaultAllowedExceptions` is private.

analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
```diff
@@ -85,61 +83,83 @@ protected override void Initialize(SonarAnalysisContext context)
...
-            ReportOnInvalidThrowStatement(analysisContext, syntax, allowedThrowTypes);
+        private static bool HasTrackedMethodOrAttributeName(MethodDeclarationSyntax declaration)
+        {
+            var name = declaration.Identifier.ValueText;
+            return name == "Equals"
+                || name == "GetHashCode"
+                || name == "ToString"
+                || name == "Dispose"
+                || name == "Equals"
+                || declaration.AttributeLists.SelectMany(list => list.Attributes).Any(x => x.ArgumentList == null && x.Name.ToStringContains("ModuleInitializer"));
```
costin-zaharia-sonarsource, line 116: The last line is rather long. I think we should extract a method `CanBeModuleInitializer` or something similar and call it here.

analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
```diff
@@ -85,61 +83,83 @@ protected override void Initialize(SonarAnalysisContext context)
...
-        private void ReportOnInvalidThrowStatement(SyntaxNodeAnalysisContext analysisContext,
+        private static bool HasTrackedMethodOrAttributeType(IMethodSymbol methodSymbol) =>
+            methodSymbol.IsObjectEquals()
+            || methodSymbol.IsObjectGetHashCode()
+            || methodSymbol.IsObjectToString()
+            || methodSymbol.IsIDisposableDispose()
+            || methodSymbol.IsIEquatableEquals()
+            || IsModuleInitializer(methodSymbol);
+
+        private static bool IsModuleInitializer(IMethodSymbol methodSymbol) =>
```
costin-zaharia-sonarsource, line 127: `methodSymbol` can be `ISymbol`

analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
```diff
@@ -85,61 +83,83 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            methodSymbol.AnyAttributeDerivesFrom(KnownType.System_Runtime_CompilerServices_ModuleInitializerAttribute);
+
+        private static void ReportOnInvalidThrowStatement(SyntaxNodeAnalysisContext analysisContext,
```
costin-zaharia-sonarsource, line 130: The method name can be confusing. The method checks both throw statements and expressions.

analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
```diff
@@ -85,61 +83,83 @@ protected override void Initialize(SonarAnalysisContext context)
...
-                .Select(tss => new NodeAndSymbol(tss, analysisContext.SemanticModel.GetSymbolInfo(tss.Expression).Symbol))
-                .FirstOrDefault(tuple => tuple.Symbol != null &&
-                    !tuple.Symbol.ContainingType.IsAny(allowedTypes) &&
-                    !tuple.Symbol.ContainingType.DerivesFromAny(allowedTypes))
-                ?.Node;
-
-            if (throwToReportOn != null)
+            if (ExpressionBody(node) is { } expressionBody
+                && ThrowExpressionSyntaxWrapper.IsInstance(expressionBody.Expression)
+                && (ThrowExpressionSyntaxWrapper)expressionBody.Expression is { } throwExpression
```
costin-zaharia-sonarsource, line 135: Since `ThrowExpressionSyntaxWrapper` is struct it can never be null. I suggest to use `(ThrowExpressionSyntaxWrapper)expressionBody.Expression is var throwExpression` instead.

Strange thing: the coverage is 100% on this line although the condition (`is { }`) can never be false.

analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
```diff
@@ -85,61 +83,83 @@ protected override void Initialize(SonarAnalysisContext context)
...
-                methodSymbol.IsObjectToString() ||
-                methodSymbol.IsIDisposableDispose() ||
-                methodSymbol.IsIEquatableEquals();
         }
 
-        private static bool IsTrackedOperator(OperatorDeclarationSyntax declaration)
-        {
-            return TrackedOperators.Contains(declaration.OperatorToken.Kind());
-        }
+        private static ArrowExpressionClauseSyntax ExpressionBody(SyntaxNode node) =>
```
costin-zaharia-sonarsource, line 151: I think this method is generic enough and can be extracted outside this class to be reused by other rules.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ExceptionShouldNotBeThrownFromUnexpectedMethods.CSharp9.cs
```diff
@@ -1,27 +1,53 @@
...
+namespace Tests.CustomType
+{
+    public class ModuleInitializerAttribute : Attribute { }
+
+    public class Foo
+    {
+        [ModuleInitializer]
+        internal static void M1()
+        {
+            throw new Exception(); // Compliant
```
costin-zaharia-sonarsource, line 12: From my point of view these comments are redundant if we just specify that the case is compliant without any other context info.

Can you please add a comment that the attribute is user defined? It took me a while to see this 😄

analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
```diff
@@ -85,61 +83,83 @@ protected override void Initialize(SonarAnalysisContext context)
...
-                .Where(tss => tss.Expression != null)
-                .Select(tss => new NodeAndSymbol(tss, analysisContext.SemanticModel.GetSymbolInfo(tss.Expression).Symbol))
-                .FirstOrDefault(tuple => tuple.Symbol != null &&
-                    !tuple.Symbol.ContainingType.IsAny(allowedTypes) &&
-                    !tuple.Symbol.ContainingType.DerivesFromAny(allowedTypes))
-                ?.Node;
-
-            if (throwToReportOn != null)
+            if (ExpressionBody(node) is { } expressionBody
+                && ThrowExpressionSyntaxWrapper.IsInstance(expressionBody.Expression)
```
costin-zaharia-sonarsource, line 134: The expression body itself can be another expression which contains a throw. This case seems to be unsupported.

E.g.:
```
        private string name;

        public override string ToString() =>
            string.IsNullOrEmpty(name)
                ? throw new ArgumentException("...")
                : string.Empty;
```

costin-zaharia-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
andrei-epure-sonarsource: well, a `ModuleInitializer` must be a `IMethodSymbol`, right? IMO it makes more sense to keep it like this

analyzers/src/SonarAnalyzer.CSharp/Rules/ExceptionShouldNotBeThrownFromUnexpectedMethods.cs
andrei-epure-sonarsource: nice catch!

costin-zaharia-sonarsource (approved):
LGTM!

PR #4736, opened 2021-07-29
Log autogenerated files

### LINKED ISSUE #4723: Log autogenerated files

Log files detected as autogenerated with `DEBUG` severity.

pavel-mikula-sonarsource:
Fixes #4723

This is a draft, rebase will be needed after #4733 is merged. `Squash SQ` commit doesn't need a review for the same reason.


analyzers/src/SonarAnalyzer.Common/Rules/Utilities/LogAnalyzerBase.cs
```diff
@@ -31,12 +31,17 @@ public abstract class LogAnalyzerBase<TSyntaxKind> : UtilityAnalyzerBase<TSyntax
...
+        protected override bool AnalyzeGeneratedCode => true;
 
         protected LogAnalyzerBase() : base(DiagnosticId, Title) { }
 
         protected sealed override LogInfo CreateAnalysisMessage(SonarAnalysisContext context) =>
             new LogInfo { Severity = LogSeverity.Info, Text = "Roslyn version: " + typeof(SyntaxNode).Assembly.GetName().Version };
 
-        protected sealed override LogInfo CreateMessage(SyntaxTree syntaxTree, SemanticModel semanticModel) => null;
+        protected sealed override LogInfo CreateMessage(SyntaxTree syntaxTree, SemanticModel semanticModel) =>
+            Language.GeneratedCodeRecognizer.IsGenerated(syntaxTree)
```
csaba-sagi-sonarsource, line 42: If the analysis time gets worse because of the fact that the IsGenerated is called now 2x much as before, we can try to use `DiagnosticAnalyzerContextHelper.IsGenerated` isntead.

csaba-sagi-sonarsource (approved):
I left a small comment, Nothing to do now, but if we see that we have increased analysis time than we can take a look at it.

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/LogAnalyzerBase.cs
pavel-mikula-sonarsource: That's a good finding. We can just directly go for cached value as any other rule does. Caching should be always better than brute force

PR #4740, opened 2021-07-29
Rule S3353: add support for top level methods

costin-zaharia-sonarsource:
Related to https://jira.sonarsource.com/browse/MMF-2222


andrei-epure-sonarsource (approved):
LGTM

PR #4741, opened 2021-07-29
Rule S2123: add support for cast in return statement

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/ValuesUselesslyIncremented.cs
```diff
@@ -32,26 +32,21 @@ namespace SonarAnalyzer.Rules.CSharp
...
                 c =>
                 {
                     var increment = (PostfixUnaryExpressionSyntax)c.Node;
                     var symbol = c.SemanticModel.GetSymbolInfo(increment.Operand).Symbol;
-                    var localSymbol = symbol as ILocalSymbol;
-                    var parameterSymbol = symbol as IParameterSymbol;
 
-                    if (localSymbol == null &&
-                        (parameterSymbol == null || parameterSymbol.RefKind != RefKind.None))
+                    if (!(symbol is ILocalSymbol || symbol is IParameterSymbol {RefKind: RefKind.None}))
```
andrei-epure-sonarsource, line 49: nice! however, I find it easier to read it like:

```
if (symbol is not ILocalSymbol && symbol is not IParameterSymbol {RefKind: RefKind.None})
```

WDYT? The `!` is quite hidden and I initially didn't see it. Also, the above is clearer that it's an `and` condition.

Or even better, get rid of the early return (and negative logic) and add the rest of the function inside the if

```
if (symbol is ILocalSymbol || symbol is IParameterSymbol {RefKind: RefKind.None})
{
   // the logic here
}
```

Like this, it's crystal clear. Only one logic flow.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ValuesUselesslyIncremented.CSharp9.cs
```diff
@@ -2,8 +2,18 @@
...
 u = u++; // Noncompliant; u is still zero
 u = ++u; // Compliant
+u += u++;
+a = u++;
 
-return (int)i++; // Compliant FN
+int x = (int) u++;
+
+int Compute(nint v) =>
+    (int) v++; // Noncompliant
```
andrei-epure-sonarsource, line 14: could you please add the location // ^^^^ ?

andrei-epure-sonarsource (approved):
LGTM, I've left some improvements to do before merging.

PR #4744, opened 2021-07-30
Log language version

### LINKED ISSUE #4743: Log language version
Log current language version in scanner end step

pavel-mikula-sonarsource:
Fixes #4743

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Rules/Utilities/LogAnalyzerBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/Utilities/LogAnalyzerBase.cs
@@ -18,7 +18,9 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.Collections.Generic;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Helpers;
 using SonarAnalyzer.Protobuf;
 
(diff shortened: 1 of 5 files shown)


analyzers/src/SonarAnalyzer.Common/Rules/Utilities/LogAnalyzerBase.cs
```diff
@@ -30,13 +32,19 @@ public abstract class LogAnalyzerBase<TSyntaxKind> : UtilityAnalyzerBase<TSyntax
         protected const string DiagnosticId = "S9999-log";
         private const string Title = "Log generator";
 
+        protected abstract string LanguageVersion(Compilation compilation);
```
csaba-sagi-sonarsource, line 35: As there is already a GetLanguageVersion extension and it is used in multiple places, I would add a CompilationFacade to the LanguageFacade and move this retrieval there.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/LogAnalyzerTest.cs
```diff
@@ -99,6 +105,15 @@ private static void VerifyRoslynVersion(IEnumerable<LogInfo> messages)
             version.Should().BeGreaterThan(new Version(3, 0));  // Avoid 1.0.0.0
         }
 
+        private static void VerifyLanguageVersion(IEnumerable<LogInfo> messages)
+        {
+            messages.Should().NotBeEmpty();
+            var versionMessage = messages.SingleOrDefault(x => x.Text.Contains("Language version"));
+            versionMessage.Should().NotBeNull();
+            versionMessage.Severity.Should().Be(LogSeverity.Info);
+            versionMessage.Text.Should().MatchRegex(@"^Language version: (CSharp|VisualBasic)\d");
```
csaba-sagi-sonarsource, line 114:
```suggestion
            versionMessage.Text.Should().MatchRegex(@"^Language version: (CSharp|VisualBasic)\d+$");
```

csaba-sagi-sonarsource (changes requested):
Two small changes suggested.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/LogAnalyzerTest.cs
pavel-mikula-sonarsource: Enum values can get more tricky, so the general regex should be enough:
https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.languageversion?view=roslyn-dotnet-3.10.0

https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.visualbasic.languageversion?view=roslyn-dotnet-3.10.0

analyzers/src/SonarAnalyzer.Common/Rules/Utilities/LogAnalyzerBase.cs
pavel-mikula-sonarsource: This cannot be done, because they return different enums that cannot be processed in a generic way.

csaba-sagi-sonarsource: Indeed I missed the fact that on the other places the enum value is used and not the string value of it.

csaba-sagi-sonarsource (approved):
LGTM!

PR #4746, opened 2021-07-30
Rule S3052: detect default constructors

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializedToDefault.cs
```diff
@@ -151,9 +153,16 @@ private static bool IsValueTypeDefaultValueInitializer(EqualsValueClauseSyntax i
                         {
                             return true;
                         }
-                        else if (initializer.Value is ObjectCreationExpressionSyntax objectCreation)
+                        else if (initializer.Value is ObjectCreationExpressionSyntax || ImplicitObjectCreationExpressionSyntaxWrapper.IsInstance(initializer.Value))
                         {
-                            return ExpressionNumericConverter.TryGetConstantIntValue(objectCreation.ArgumentList?.Arguments.First().Expression, out var ctorParameter)
+                            var oc = ObjectCreationFactory.Create(initializer.Value);
```
andrei-epure-sonarsource, line 158: why not `objectCreation` instead of `oc`?

andrei-epure-sonarsource (approved):
LGTM

PR #4749, opened 2021-08-02
Rule S4583: Add support for C# 9 partial methods

dmitry-golovinov-sonarsource:

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/BeginInvokePairedWithEndInvoke.CSharp9.Part1.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/BeginInvokePairedWithEndInvoke.CSharp9.Part1.cs
@@ -6,7 +6,7 @@ namespace Tests.Diagnostics
 
     class Program
     {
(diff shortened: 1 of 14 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/BeginInvokePairedWithEndInvoke.CSharp9.Part1.cs
```diff
@@ -0,0 +1,109 @@
...
+namespace Tests.Diagnostics
+{
+    public delegate void AsyncMethodCaller(string name, int i);
+
+    class Program
+    {
+        private static void BeginInvokeOnDelegateWithLambdaCallback_DiscardParam()
+        {
+            var caller = new AsyncMethodCaller(AsyncMethod);
+            // here the "_" is actually an identifier, not a discard parameter
```
costin-zaharia-sonarsource, line 12: In this case, the test name is wrong.

costin-zaharia-sonarsource (approved):
LGTM! I've let just a small comment.

PR #4753, opened 2021-08-03
S3251: C#9 new partial method features support

csaba-sagi-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PartialMethodNoImplementation.CSharp9.cs
```diff
@@ -1,18 +1,11 @@
 ﻿public partial record Record
 {
     partial void Method(); //Noncompliant {{Supply an implementation for this partial method.}}
```
andrei-epure-sonarsource, line 3: please add some tests with partial methods with non-void return and also with `out` parameters

andrei-epure-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PartialMethodNoImplementation.CSharp9.cs
andrei-epure-sonarsource: and please add an IT `S3251.cs` in the .NET 5 IT project

andrei-epure-sonarsource (approved):
LGTM

PR #4758, opened 2021-08-04
Fix S4457 FP: Should not raise on `async Task Main(string[] args)'

### LINKED ISSUE #4702: Fix S4457 FP: Should not raise on `async Task Main(string[] args)'
per request at https://github.com/SonarSource/sonar-dotnet/issues/2665#issuecomment-829269767

This code triggers S4457
``` csharp
private static async Task Main(string[] args)
{
       if (args.Length == 0)
       {
           throw new ArgumentException("The base url must be provided as a command argument");
       }

       var baseUrl = new Uri(args[0].Trim());
         ...
```

I am checking the parameters, and its the entry point. Why should that be getting split for additional arg validation?

pavel-mikula-sonarsource:
Fixes #4702

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ParameterValidationInAsyncShouldBeWrapped.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ParameterValidationInAsyncShouldBeWrapped.cs
@@ -124,5 +124,14 @@ namespace Tests.Diagnostics
 
             await Task.Delay(0);
         }
+
+        // https://github.com/SonarSource/sonar-dotnet/issues/4702
+        private static async Task Main(string[] args)   // Noncompliant FP, should not raise on Main
+        {
+            if (args.Length == 0)
+            {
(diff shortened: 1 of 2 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ParameterValidationInAsyncShouldBeWrapped.cs
```diff
@@ -124,5 +124,14 @@ private async void OnButtonPressed( object sender, EventArgs e) // Compliant (th
 
             await Task.Delay(0);
         }
+
+        // https://github.com/SonarSource/sonar-dotnet/issues/4702
+        private static async Task Main(string[] args)   // Ccompliant, should not raise on Main
```
andrei-epure-sonarsource, line 129:
```suggestion
        private static async Task Main(string[] args)   // Compliant, should not raise on Main
```

andrei-epure-sonarsource (approved):
LGTM

PR #4701, opened 2021-07-16
Provide default imports for VB.NET UnitTest project

pavel-mikula-sonarsource:
Fixes #4400

PR #4765, opened 2021-08-05
Fix ConstructorArgumentValueShouldExistTest WPF support

### LINKED ISSUE #3425: Fix ConstructorArgumentValueShouldExistTest WPF support
Some tests, like `ConstructorArgumentValueShouldExistTest` reference `WPF` types and currently cannot be run while targeting .Net Core.

pavel-mikula-sonarsource:
Fixes #3425

I was doing some random improvements (leftovers from previous PRs) and I noticed that `ConstructorArgumentValueShouldExist_CS_CSharp9` test was never compiled & executed.

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Helpers/DiagnosticAnalyzerContextHelper.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Helpers/DiagnosticAnalyzerContextHelper.cs
@@ -125,17 +125,8 @@ namespace SonarAnalyzer.Helpers
             {
                 return false;
             }
-
-            // this is locking if the compilation is not present in the Cache.
-            var cache = Cache.GetOrCreateValue(compilation);
-            if (cache.TryGetValue(tree, out var result))
-            {
-                return result;
-            }
-
-            var generated = generatedCodeRecognizer.IsGenerated(tree);
-            cache.TryAdd(tree, generated);
-            return generated;
+            var cache = Cache.GetOrCreateValue(compilation);    // This is locking if the compilation is not present in the Cache.
+            return cache.GetOrAdd(tree, x => generatedCodeRecognizer.IsGenerated(x));
         }
 
         internal static bool ShouldAnalyze(SonarAnalysisContext context, GeneratedCodeRecognizer generatedCodeRecognizer, SyntaxTree syntaxTree, Compilation c, AnalyzerOptions options) =>
--- a/analyzers/tests/SonarAnalyzer.UnitTest/Rules/ConstructorArgumentValueShouldExistTest.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/Rules/ConstructorArgumentValueShouldExistTest.cs
@@ -18,10 +18,6 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-// ToDo: Add WPF support in unit tests when targeting .Net Core
-// https://github.com/SonarSource/sonar-dotnet/issues/3425
-#if NETFRAMEWORK
-
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarAnalyzer.UnitTest.MetadataReferences;
 using SonarAnalyzer.UnitTest.TestFramework;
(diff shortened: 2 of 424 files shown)


analyzers/src/SonarAnalyzer.Common/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -125,17 +125,8 @@ public static bool IsGenerated(this SyntaxTree tree, GeneratedCodeRecognizer gen
...
-            if (cache.TryGetValue(tree, out var result))
-            {
-                return result;
-            }
-
-            var generated = generatedCodeRecognizer.IsGenerated(tree);
-            cache.TryAdd(tree, generated);
-            return generated;
+            var cache = Cache.GetOrCreateValue(compilation);    // This is locking if the compilation is not present in the Cache.
+            return cache.GetOrAdd(tree, x => generatedCodeRecognizer.IsGenerated(x));
```
pavel-mikula-sonarsource, line 129: Just a better API

analyzers/tests/SonarAnalyzer.UnitTest/Rules/ConstructorArgumentValueShouldExistTest.cs
```diff
@@ -38,12 +34,12 @@ public class ConstructorArgumentValueShouldExistTest
             Verifier.VerifyAnalyzer(@"TestCases\ConstructorArgumentValueShouldExist.cs",
                 new CS.ConstructorArgumentValueShouldExist(),
                 MetadataReferenceFacade.SystemXaml);
+
```
pavel-mikula-sonarsource, line 37: This test was wrapped with lethal mix
```
#if NETFRAMEWORK
#if NET

never compile

#endif
#endif
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -125,7 +125,7 @@ private ProjectBuilder AddProject(AnalyzerLanguage language, string projectName,
             compilationOptions = languageName switch
             {
                 LanguageNames.CSharp => ((CSharpCompilationOptions)compilationOptions).WithAllowUnsafe(true),
-                LanguageNames.VisualBasic => ((VisualBasicCompilationOptions)compilationOptions).WithGlobalImports(GlobalImport.Parse(DefaultGlobalImportsVisualBasic)),
+                LanguageNames.VisualBasic => ((VisualBasicCompilationOptions)compilationOptions).WithGlobalImports(DefaultGlobalImportsVisualBasic),
```
pavel-mikula-sonarsource, line 128: Related to https://github.com/SonarSource/sonar-dotnet/pull/4701#discussion_r672121192

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConstructorArgumentValueShouldExist.CSharp9.cs
```diff
@@ -11,3 +11,20 @@ public class MyExtension3 : MarkupExtension
 
     public override object ProvideValue(IServiceProvider serviceProvider) => null;
 }
+
+// This is fake scaffolding to avoid .NET 5 reference issue.
```
costin-zaharia-sonarsource, line 15: This looks to be a fix which can be reused. What do you think about moving it to the project generation part?

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
```diff
@@ -47,7 +47,7 @@ internal struct SolutionBuilder
             "System.Linq",
             "System.Xml.Linq",
             "System.Threading.Tasks"
-        };
+        }.Select(GlobalImport.Parse).ToArray();
```
costin-zaharia-sonarsource, line 50: This creates an array to parse it and then copy the result to another array. What do you think about using the overload which takes a list of strings: https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.visualbasic.globalimport.parse?view=roslyn-dotnet-3.10.0#Microsoft_CodeAnalysis_VisualBasic_GlobalImport_Parse_System_Collections_Generic_IEnumerable_System_String__

costin-zaharia-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConstructorArgumentValueShouldExist.CSharp9.cs
pavel-mikula-sonarsource: It's just this one rule that needs these specific classes. I would keep this trick only in this rule. We have similar ones elsewhere for `System.Windows`:
https://github.com/SonarSource/sonar-dotnet/blob/ebcf626d64d8d2bbdbfd9f95db50697113e90c44/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ReferenceEqualityCheckWhenEqualsExists.cs#L6-L24

We can build something better in case we'll have a problem with it (where `problem` could be a WPF-related MMF 😃 )

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SolutionBuilder.cs
pavel-mikula-sonarsource: Waaay better :)

costin-zaharia-sonarsource (approved):
LGTM!

PR #4766, opened 2021-08-05
Fix solution path

costin-zaharia-sonarsource:
`$(solution)` was wrong since it contains `analyzers` folder too.


dmitry-golovinov-sonarsource (approved):
LGTM!

PR #4767, opened 2021-08-05
Remove TestCategory attribute

pavel-mikula-sonarsource:
That's almost -1500 LOC that we don't need to maintain


analyzers/tests/SonarAnalyzer.UnitTest/TestCasesForRuleFailure/PerformanceTestCases.cs
```diff
@@ -1,4 +1,4 @@
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
```
csaba-sagi-sonarsource, line 1: Hmm, what changed here? o.O

csaba-sagi-sonarsource (approved):
LGTM

analyzers/tests/SonarAnalyzer.UnitTest/TestCasesForRuleFailure/PerformanceTestCases.cs
pavel-mikula-sonarsource: Added BOM

PR #4768, opened 2021-08-05
Improve parsing of Roslyn Report

YulerB:
Proposal to fix: https://github.com/SonarSource/sonar-dotnet/issues/4771


pavel-mikula-sonarsource:
Hi @YulerB,

Thank you for looking into this. Unfortunately, we'll need more than just a blind change that is not covered by any test.

Can you add integration test that will produce such Sarif report and fail to reproduce the issue? And UT to support this code change?

pavel-mikula-sonarsource:
Can you please also close the other issue in SonarSource/sonar-scanner-msbuild#1042 and create similar one in this repo?

YulerB:
Closing, turns out this is an issue reading roslyn files for sonar scan, while dotnet publish is happening in parallel.

PR #4770, opened 2021-08-05
Fix S1172 FP: Parameter captured in a referenced generic local function

### LINKED ISSUE #4704: Fix S1172 FP: Parameter captured in a referenced generic local function

pavel-mikula-sonarsource:
Fixes #4704

The key idea is to handle `GenericNameSyntax` (that has `.Identifier`) the same way as `ProcessIdentifier` was updated in #3818

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/LiveVariableAnalysis/CSharp/CSharpLiveVariableAnalysis.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/LiveVariableAnalysis/CSharp/CSharpLiveVariableAnalysis.cs
@@ -79,6 +79,10 @@ namespace SonarAnalyzer.LiveVariableAnalysis.CSharp
                         ProcessIdentifier((IdentifierNameSyntax)instruction, state);
                         break;
 
+                    case SyntaxKind.GenericName:
+                        ProcessGenericName(instruction, state);
+                        break;
+
                     case SyntaxKind.SimpleAssignmentExpression:
                         ProcessSimpleAssignment((AssignmentExpressionSyntax)instruction, state);
                         break;
@@ -175,6 +179,16 @@ namespace SonarAnalyzer.LiveVariableAnalysis.CSharp
                 {
                     ProcessLocalFunction(symbol, state);
                 }
+            }
+        }
+
+        private void ProcessGenericName(SyntaxNode genericName, State state)
+        {
+            if (genericName.Parent.IsKind(SyntaxKind.InvocationExpression)
+                && semanticModel.GetSymbolInfo(genericName).Symbol is IMethodSymbol method
+                && method.MethodKind == MethodKindEx.LocalFunction)
+            {
+                ProcessLocalFunction(method, state);
             }
         }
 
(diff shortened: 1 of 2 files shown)


analyzers/src/SonarAnalyzer.CSharp/LiveVariableAnalysis/CSharp/CSharpLiveVariableAnalysis.cs
```diff
@@ -178,6 +182,16 @@ private void ProcessIdentifier(IdentifierNameSyntax identifier, State state)
             }
         }
 
+        private void ProcessGenericName(SyntaxNode genericName, State state)
+        {
+            if (genericName.Parent.IsKind(SyntaxKind.InvocationExpression)
```
pavel-mikula-sonarsource, line 187: This is the same logic as in the `ProcessIdentifier(..)` above. We just don't need to care about the `if (IsLocalScoped(..` here.

analyzers/src/SonarAnalyzer.CSharp/LiveVariableAnalysis/CSharp/CSharpLiveVariableAnalysis.cs
```diff
@@ -178,6 +182,16 @@ private void ProcessIdentifier(IdentifierNameSyntax identifier, State state)
...
+                && semanticModel.GetSymbolInfo(genericName).Symbol is IMethodSymbol method
```
costin-zaharia-sonarsource, line 188: Can be rewritten like: `&& semanticModel.GetSymbolInfo(genericName).Symbol is IMethodSymbol { MethodKind: MethodKindEx.LocalFunction } method`

costin-zaharia-sonarsource (changes requested):
I think the solution looks good so far but the implementation is not yet complete.

Can you please add the following test case?
```
private static void M(this int someNumber, string someString)
{
    Action x = PrintSomeSum<int>;

    x();

    void PrintSomeSum<TOptions>() where TOptions : struct
    {
        Console.WriteLine(someNumber + someString.Length);
    }
}
```

costin-zaharia-sonarsource (approved):
LGTM!

PR #4773, opened 2021-08-05
Fix S2365 FP: Should not raise for overriding methods and interface implementations

### LINKED ISSUE #4713: Fix S2365 FP: Should not raise for overriding methods and interface implementations
### Description

S2365 raises issue for overriding member - it cannot change it's signature.

### Repro steps
```
Friend Class Crc32
    Inherits HashAlgorithm

    '...

    Public Overrides ReadOnly Property Hash() As Byte()    ' False Positive
        Get
            Return HashFinal()
        End Get
    End Property

End Class
```

pavel-mikula-sonarsource:
Fixes #4713

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PropertyWithArrayType.vb
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PropertyWithArrayType.vb
@@ -3,7 +3,7 @@
     ' Internal state
     Dim Array = {"apple", "banana", "orange", "pineapple", "strawberry"}
 
-    Public Property NotAProblem As Byte()   'Noncompliant FP, this doens't copy anything inside
+    Public Property NotAProblem As Byte()   'Compliant, this doens't copy anything inside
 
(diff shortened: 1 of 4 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PropertyWithArrayType.vb
```diff
@@ -1,29 +1,63 @@
...
+Public Class Sample
+    Inherits Base
+    Implements IBase
+
+    Public ReadOnly Property Hash As Byte() Implements IBase.Hash   ' Compliant, this class cannot change interface signature
+        Get
+        End Get
+    End Property
+
+    Public Overrides Property Bad As Byte() ' Ccompliant, this class cannot change overriden property signature
```
csaba-sagi-sonarsource, line 56: Type-o
```suggestion
    Public Overrides Property Bad As Byte() ' Compliant, this class cannot change overriden property signature
```

csaba-sagi-sonarsource (approved):
LGTM, just one typo.

PR #4775, opened 2021-08-05
Fix S1764 FP: Don't run on Test code

### LINKED ISSUE #4659: Fix S1764 FP: Don't run on Test code
This rule should not run on test code. Equals is often subject of the test itself. 

S1764 Change one instance of 'context1' to a different value; comparing 'context1' to itself always 
```
context1.Equals(context1).Should().BeTrue(); 
```

As of now, the issue is raised in SL/NuGet context.

pavel-mikula-sonarsource:
Fixes #4659

### DIFF
--- a/analyzers/rspec/cs/S1764_c#.json
+++ b/analyzers/rspec/cs/S1764_c#.json
@@ -10,5 +10,5 @@
   "defaultSeverity": "Major",
   "ruleSpecification": "RSPEC-1764",
   "sqKey": "S1764",
-  "scope": "All"
+  "scope": "Main"
 }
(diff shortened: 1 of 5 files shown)


pavel-mikula-sonarsource:
@andrei-epure-sonarsource Can you take one more close look at the RSPEC description and our test cases? The rule handles binary operators in general (`a > a`, `a | a`, `a == a`, `a.Equals(a)`).

While it makes sense to test overridden operators, I'm not as convinced about the change as I was before.

andrei-epure-sonarsource (approved):
LGTM

PR #4776, opened 2021-08-06
Log the state of concurrent analysis.

dmitry-golovinov-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/Rules/Utilities/LogAnalyzerTest.cs
```diff
@@ -84,10 +98,17 @@ public void LogAutogenerated_VB()
...
         }
 
-        private static void VerifyCompilationMessages(IEnumerable<LogInfo> messages)
+        private static void VerifyCompilationMessagesNonConcurrentRuleExecution(IEnumerable<LogInfo> messages) =>
+            VerifyCompilationMessagesBase(messages, false);
+
+        private static void VerifyCompilationMessagesConcurrentRuleExecution(IEnumerable<LogInfo> messages) =>
+            VerifyCompilationMessagesBase(messages, true);
+
+        private static void VerifyCompilationMessagesBase(IEnumerable<LogInfo> messages, bool isEnabled)
```
pavel-mikula-sonarsource, line 107: `VerifyCompilationMessagesBase.isEnabled` - meaning of this is not clear, looks like it should enable the verification itself.

Let's change it to `string expectedConcurrencyMessage` also to avoid the ternary calculation below.

pavel-mikula-sonarsource (approved):
LGTM, there's one change in UT suggested

PR #4777, opened 2021-08-06
Update rspec

costin-zaharia-sonarsource:
Update rspec before release

### DIFF
--- a/analyzers/rspec/cs/S4784_c#.html
+++ b/analyzers/rspec/cs/S4784_c#.html
@@ -85,6 +85,4 @@ namespace N
   </li>
   <li> OWASP Regular expression Denial of Service - ReDoS </li>
 </ul>
-<h2>Deprecated</h2>
-<p>This rule is deprecated; use {rule:csharpsquid:S2631} instead.</p>
 
--- a/analyzers/rspec/cs/S4784_c#.html
+++ b/analyzers/rspec/cs/S4784_c#.html
@@ -85,4 +85,6 @@ namespace N
   </li>
   <li> OWASP Regular expression Denial of Service - ReDoS </li>
 </ul>
+<h2>Deprecated</h2>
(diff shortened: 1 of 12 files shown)


analyzers/rspec/cs/S4784_c#.html
```diff
@@ -85,6 +85,4 @@ <h2>See</h2>
   </li>
   <li> OWASP Regular expression Denial of Service - ReDoS </li>
 </ul>
-<h2>Deprecated</h2>
-<p>This rule is deprecated; use {rule:csharpsquid:S2631} instead.</p>
```
pavel-mikula-sonarsource, line 89: What's the story behind this? It was undeprecated? I see it deprecated on the RSPEC repo

pavel-mikula-sonarsource (approved):
LGTM except one question that is not clear to me

analyzers/rspec/cs/S4784_c#.html
costin-zaharia-sonarsource: It's strange indeed. The [rule is still deprecated](https://sonarsource.github.io/rspec/#/rspec/S4784/csharp) but the tool remove this anyway.

I will revert the changes.

pavel-mikula-sonarsource: Can you ping Arsenyi? There might be some (maybe already known) glitch in the rule-api

costin-zaharia-sonarsource: Yes, it is a known issue: https://jira.sonarsource.com/browse/RULEAPI-669

costin-zaharia-sonarsource: Maybe with the next release we will follow the logic and mark the rule as ready until we have the replacement

pavel-mikula-sonarsource: Can you prepare PR with rspec update for this rule already? Because we'll forget the context until the next release :/

PR #4780, opened 2021-08-06
Fix S3459 FN: Support fields in generic classes

### LINKED ISSUE #4757: S3459 FP with private inner class fields initialized in the outer class
### Description

Found [on sonarcloud](https://sonarcloud.io/project/issues?id=sonaranalyzer-dotnet&issues=AXsNH8AULrVS-fD458_u&open=AXsNH8AULrVS-fD458_u&pullRequest=4756).

It seems to repro only on generic classes. Only if `X` is used inside one of the methods - if it is not used, it is not reported.

### Repro steps

```
public sealed class OuterFoo
{
    private static class InnerFactory
    {
        public static InnerFoo<string> Instance = new InnerFoo<string> { X = "X" };
    }

    private class InnerFoo<T>
    {
        public string X; // FP S3459 - remove unassigned field `X` or set its value

        public void Foo()
        {
            Console.Write(X);
        }
    }
}

```

### Expected behavior

It is a private class initialized in another private class where the field is used - the issue should not be reported.

### Actual behavior

Reports a false positive.

### Known workarounds

None.

### Related information

It reproes on SonarCloud and in SonarLint for Visual Studio.

* Visual Studio version 16.10.3
* MSBuild / dotnet version 16.9 (on AZP)
* SonarScanner for .NET version (if used) 5.2.2

dmitry-golovinov-sonarsource:
Fixes #4757

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
@@ -194,7 +194,7 @@ namespace SonarAnalyzer.Rules.CSharp
 
                 if (PreOrPostfixOpSyntaxKinds.Contains(parentNode.Kind()))
                 {
-                    assignedMembers.Add(memberSymbol);
+                    AddAssignedMember(assignedMembers, memberSymbol);
                     continue;
                 }
 
@@ -202,7 +202,7 @@ namespace SonarAnalyzer.Rules.CSharp
                 {
                     if (assignment.Left == node)
                     {
-                        assignedMembers.Add(memberSymbol);
+                        AddAssignedMember(assignedMembers, memberSymbol);
                     }
 
                     continue;
@@ -211,11 +211,20 @@ namespace SonarAnalyzer.Rules.CSharp
                 if (parentNode is ArgumentSyntax argument
                     && (!argument.RefOrOutKeyword.IsKind(SyntaxKind.None) || TupleExpressionSyntaxWrapper.IsInstance(argument.Parent)))
                 {
-                    assignedMembers.Add(memberSymbol);
+                    AddAssignedMember(assignedMembers, memberSymbol);
                 }
             }
 
             return assignedMembers;
+        }
+
+        private static void AddAssignedMember(ISet<ISymbol> assignedMembers, ISymbol symbol)
+        {
+            assignedMembers.Add(symbol);
+            if (!symbol.Equals(symbol.OriginalDefinition))
+            {
+                assignedMembers.Add(symbol.OriginalDefinition);
+            }
         }
 
         private static bool IsParentMemberAccess(MemberAccessExpressionSyntax parent, ExpressionSyntax node) =>
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
@@ -192,39 +192,16 @@ namespace SonarAnalyzer.Rules.CSharp
 
                 var parentNode = node.Parent;
 
-                if (PreOrPostfixOpSyntaxKinds.Contains(parentNode.Kind()))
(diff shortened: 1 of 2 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -211,13 +211,22 @@ private static ISet<ISymbol> GetAssignedMemberSymbols(IList<MemberUsage> memberU
...
             return assignedMembers;
         }
 
+        private static void AddAssignedMember(ISet<ISymbol> assignedMembers, ISymbol symbol)
+        {
+            assignedMembers.Add(symbol);
+            if (!symbol.Equals(symbol.OriginalDefinition))
+            {
+                assignedMembers.Add(symbol.OriginalDefinition);
+            }
```
pavel-mikula-sonarsource, line 227: It's a `HashSet` so it will do the `Equals` check for us anyway under the hood. So we don't need to check it ourselves and we can directly just add it.
```suggestion
            assignedMembers.Add(symbol.OriginalDefinition);
```

pavel-mikula-sonarsource: Having this as just two similar lines with no logic, we can make it a local function and remove the `assignedMembers` argument on the way.

analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -194,15 +194,15 @@ private static ISet<ISymbol> GetAssignedMemberSymbols(IList<MemberUsage> memberU
 
                 if (PreOrPostfixOpSyntaxKinds.Contains(parentNode.Kind()))
                 {
-                    assignedMembers.Add(memberSymbol);
+                    AddAssignedMember(assignedMembers, memberSymbol);
```
pavel-mikula-sonarsource, line 197: Please refactor this method to use `else if` instead of those `continue` statements

analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -194,15 +194,15 @@ private static ISet<ISymbol> GetAssignedMemberSymbols(IList<MemberUsage> memberU
...
                     continue;
                 }
 
                 if (parentNode is AssignmentExpressionSyntax assignment)
                 {
                     if (assignment.Left == node)
                     {
-                        assignedMembers.Add(memberSymbol);
+                        AddAssignedMember(assignedMembers, memberSymbol);
```
pavel-mikula-sonarsource, line 205: These two nested `if` can be safely collapsed.

pavel-mikula-sonarsource: Well, the whole thing seems simplifiable into a state where we won't need the method/local function. Just some `if a || b || c then add`

pavel-mikula-sonarsource (changes requested):
Some cleaning around to make it readable

analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
dmitry-golovinov-sonarsource: That was a debugging artifact :-)

dmitry-golovinov-sonarsource: I prefer passing the arguments to the functions explicitly, rather than referencing something out of the scope :) but okay, will do it like this.

analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
dmitry-golovinov-sonarsource: Happy to eliminate that one, indeed :)

pavel-mikula-sonarsource (approved):
LGTM

PR #4781, opened 2021-08-06
Fix S3240 FP: Property initializers that share a name

### LINKED ISSUE #4607: Fix S3240 FP: Property initializers that share a name
### Description

Rule "S3240 - The simplest possible condition syntax should be used" gives a false positive when assigning via property initializer syntax and the target instance being initalized shares a common property name. Applying the suggested auto-fix leads to malformed code.

### Repro steps

Contrived example code to illustrate the problem:

```cs
public class Example
{
  public string Foo { get; set; }
  public string Bar { get; set; }
  
  public Example Fallback(Example other)
  {
    return new Example
    {
      Foo = Foo ?? other.Foo,
      Bar = Bar ?? other.Bar
    };
  }
}
```

### Expected behavior

S3240 is not raised, because the properties - though similar in name, actually are not the same properties of the same instance and this is _not_ a regular property assignment.

### Actual behavior

S3240 is raised and suggests the code is updated to:

```cs
public class Example
{
  public string Foo { get; set; }
  public string Bar { get; set; }
  
  public Example Fallback(Example other)
  {
    return new Example
    {
      Foo ??= other.Foo,
      Bar ??= other.Bar
    };
  }
}
```

This is malformed code and raises "[CS0747](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0747?f1url=%3FappId%3Droslyn%26k%3Dk(CS0747)) - Invalid member initializer syntax"

### Known workarounds

**suitable**
- Locally suppress the rule.

**unsuitable**
- Change one of the property names -- Not possible in case of external types or interfaces
- Adding an explicit `this.` -- Raises "[IDE0003](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0003-ide0009) - Name can be simplified" instead.
- Not using initializer syntax -- Raises "[IDE0017](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0017) - Object initialization can be simplified" instead.

### Related information

* C#/VB.NET Plugins version -- whatever ships with SonarLint 4.35.0.32570
* Visual Studio version -- VS 2019 v16.10.2
* MSBuild / dotnet version -- ??
* SonarScanner for .NET version (if used) N/A
* Operating System -- Windows 10

pavel-mikula-sonarsource:
Fixes #4607

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ConditionalSimplification.cs
@@ -94,6 +94,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private static void CheckCoalesceExpression(SyntaxNodeAnalysisContext context)
         {
             if (context.Node.GetFirstNonParenthesizedParent() is AssignmentExpressionSyntax assignment
+                && !assignment.Parent.IsKind(SyntaxKind.ObjectInitializerExpression)
                 && context.Compilation.IsCoalesceAssignmentSupported())
             {
                 var left = ((BinaryExpressionSyntax)context.Node).Left.RemoveParentheses();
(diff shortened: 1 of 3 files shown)


csaba-sagi-sonarsource (approved):
LGTM!

Issue #4784, opened 2021-08-06
S2259 FP: return value of ToList() is not null


csaba-sagi-sonarsource:
### Description

S2259 raises an issue telling that the return value of the ToList() is is null on at least one execution path.

### Repro steps

```
        public static long Test()
        {
            long sum = 0;
            long[] array = new[] { 1L, 2L, 3L };
            List<long> test = array.ToList();
            if (test?.Count == 0)
            {
                // Do something
            }

            foreach (long number in test)  // Reported Bug: 'test' is null on at least one execution path.
            {
                sum += number;
            }

            return sum;
        }

```

### Expected behavior

S2259 should not raise an issue.

### Actual behavior

S2259 raises an issue for this case.

### Known workarounds

-

### Related information

* C#/VB.NET Plugins version: 8.27

martin-strecker-sonarsource:
@pavel-mikula-sonarsource  I wouldn't solve this issue. It is a rabbit hole (there are literally millions of APIs in the runtime that never return null). The root cause here is the nonsensical `test?.Count == 0` check, which implies that test can be null. If the developer thinks so once, he should not change his mind midway.

PR #4785, opened 2021-08-06
S2259: ToList() FP repro

csaba-sagi-sonarsource:
Reproduces #4784

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereference.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereference.cs
@@ -951,22 +951,47 @@ namespace Repro_3395
     // https://github.com/SonarSource/sonar-dotnet/issues/4784
     class Repro_4784
     {
-        public static long Reproducer()
+        public static int Reproducer()
         {
-            long sum = 0;
-            long[] array = new[] { 1L, 2L, 3L };
-            List<long> test = array.ToList();
+            List<long> test = new[] { 1L, 2L, 3L }.ToList();
             if (test?.Count == 0)
             {
                 // Do something
             }
 
-            foreach (long number in test) // Noncompliant
+            return test.Count; // Noncompliant FP
+        }
+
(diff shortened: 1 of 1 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereference.cs
```diff
@@ -947,4 +947,26 @@ public static void MixedConstraints()
...
+        {
+            long sum = 0;
+            long[] array = new[] { 1L, 2L, 3L };
+            List<long> test = array.ToList();
+            if (test?.Count == 0)
+            {
+                // Do something
+            }
+
+            foreach (long number in test) // Noncompliant
```
costin-zaharia-sonarsource, line 964: I'm not sure if this is really a false positive (as the title says). We raise the issue due to the `?` from `if (test?.Count == 0)`. The engine reasoning: if test can be null on this line it can be null also on the `foreach` one.

What do you think about this?

As far as I can tell it will be great to raise on line 959 that `test` cannot be null, but currently by design the engine does not know that `ToList()` will never return null.

csaba-sagi-sonarsource: Please take a look at it, I added some test cases like we discussed and it seems to me that the issue is really related to ToList().

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NullPointerDereference.cs
```diff
@@ -947,4 +947,51 @@ public static void MixedConstraints()
...
+
+        public static int NoIssueReported()
+        {
+            var something = new Something();
+            if (something?.SomeProperty == 0)
+            {
+                // Do something
+            }
+
+            return something.SomeProperty;
```
costin-zaharia-sonarsource, line 973: Isn't this a false negative? `Something()` might return null.

costin-zaharia-sonarsource (approved):
LGTM!

PR #4786, opened 2021-08-09
Rule S1541: supports C#9 patterns

### LINKED ISSUE #4682: FN S1541: no support for C# 9 patterns
### Description

Rule S1541 should support new C# 9 patterns

### Repro steps
```
    public bool PatternMatchingAnd(object arg) =>
        arg is true and true and true and true;         // FN

    public bool PatternMatchingOr(object arg) =>
        arg is true or true or true or true;            // FN
```
### Expected behavior

Issue should be raised if maximal complexity set to 3

### Actual behavior

Issue is not raised.

### Related information

dmitry-golovinov-sonarsource:
Fixes #4682

### DIFF
--- a/analyzers/its/expected/Net5/Net5--net5.0-S1541.json
+++ b/analyzers/its/expected/Net5/Net5--net5.0-S1541.json
@@ -134,7 +134,7 @@
 },
 {
 "id":  "S1541",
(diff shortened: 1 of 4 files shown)


csaba-sagi-sonarsource (approved):
LGTM!

PR #4790, opened 2021-08-10
Change build tasks order

costin-zaharia-sonarsource:
Currently if the signing task fails it will let the build in an inconsistent state and any retry will lead to errors like: 
> ##[error]Artifact Binaries already exists for build 35465.

See: https://dev.azure.com/sonarsource/DotNetTeam%20Project/_build/results?buildId=35465&view=logs&jobId=f44a741f-9110-553a-520b-28a1a7416edf&j=f44a741f-9110-553a-520b-28a1a7416edf&t=af00f6af-222b-5187-006a-702a33759858

By pushing the artifacts after the signing occurs we avoid this.


csaba-sagi-sonarsource (approved):
LGTM!

PR #4793, opened 2021-08-12
Add 'Improving Performance' embedded documentation (#4783)

### LINKED ISSUE #4783: Add in-product docs regarding how to enable concurrent rule execution
Should add instructions on how to enable / disable concurrent rule execution: https://github.com/SonarSource/sonar-dotnet/blob/master/sonar-csharp-plugin/src/main/resources/documentation.md

Should show up in https://peach.sonarsource.com/documentation/analysis/languages/csharp/

tom-howlett-sonarsource:
Fixes #4783 by adding a section on Improving Performance with the new concurrent execution setting. Let me know who else  you think should review this @pavel-mikula-sonarsource and I can add them?

### DIFF
--- a/sonar-csharp-plugin/src/main/resources/documentation.md
+++ b/sonar-csharp-plugin/src/main/resources/documentation.md
@@ -36,7 +36,7 @@ dotnet_diagnostic.Sxxx.severity = none
 ## Improving Performance  
 
 You can enable parallel processing to speed up analysis and reduce overall build time by setting the following environment variable:
-```bash
+```
 SONAR_DOTNET_ENABLE_CONCURRENT_EXECUTION=true
 ```
 If you still have problems [this](https://community.sonarsource.com/t/47279) community thread may help you diagnose the issues.
--- a/sonar-csharp-plugin/src/main/resources/documentation.md
+++ b/sonar-csharp-plugin/src/main/resources/documentation.md
@@ -35,11 +35,11 @@ dotnet_diagnostic.Sxxx.severity = none
 
 ## Improving Performance  
 
-Analysis will significantly increase your build time. If this is causing a problem there are a number of things you can do. It is possible for many of the rules to use parallel processing to speed up execution. To enable this you need to set the following environment variable.
+You can enable parallel processing to speed up analysis and reduce overall build time by setting the following environment variable:
 ```bash
 SONAR_DOTNET_ENABLE_CONCURRENT_EXECUTION=true
 ```
-If you still have problems [this](https://community.sonarsource.com/t/the-sonarsource-guide-for-investigating-the-performance-of-net-analysis/47279) community thread may help you diagnose the issues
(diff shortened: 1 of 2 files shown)


sonar-csharp-plugin/src/main/resources/documentation.md
```diff
@@ -33,6 +33,14 @@ As an alternative, an `.editorconfig` file can be used to disable the analysis f
 dotnet_diagnostic.Sxxx.severity = none
 ```
 
+## Improving Performance  
+
+Analysis will significantly increase your build time. If this is causing a problem there are a number of things you can do. It is possible for many of the rules to use parallel processing to speed up execution. To enable this you need to set the following environment variable.
+```bash
```
pavel-mikula-sonarsource, line 39: Should not be needed
```suggestion
```
```

File: sonar-csharp-plugin/src/main/resources/documentation.md [hunk 2]
```diff
@@ -33,6 +33,14 @@ As an alternative, an `.editorconfig` file can be used to disable the analysis f
 dotnet_diagnostic.Sxxx.severity = none
 ```
 
+## Improving Performance  
+
+Analysis will significantly increase your build time. If this is causing a problem there are a number of things you can do. It is possible for many of the rules to use parallel processing to speed up execution. To enable this you need to set the following environment variable.
```
pavel-mikula-sonarsource, line 38: We're trying hard not to significantly increase the build time. I would not tell users that significant performance hit is what they get by using our product :) 
```suggestion
You can enable parallel processing to speed up analysis and reduce overall build time by setting the following environment variable:
```
or less personal, as it's a tech docs:
```suggestion
Parallel processing can be enabled to speed up analysis and reduce overall build time by setting the following environment variable:
```

sonar-csharp-plugin/src/main/resources/documentation.md
```diff
@@ -33,6 +33,14 @@ As an alternative, an `.editorconfig` file can be used to disable the analysis f
...
+```bash
+SONAR_DOTNET_ENABLE_CONCURRENT_EXECUTION=true
+```
+If you still have problems [this](https://community.sonarsource.com/t/the-sonarsource-guide-for-investigating-the-performance-of-net-analysis/47279) community thread may help you diagnose the issues
```
pavel-mikula-sonarsource, line 42: I'd use just the ID to make it short and title-independent.
```suggestion
If you still have problems [this](https://community.sonarsource.com/t/47279) community thread may help you diagnose the issues.
```

pavel-mikula-sonarsource (changes requested):
Just one reviewer is enough. I left some suggestions

sonar-csharp-plugin/src/main/resources/documentation.md
tom-howlett-sonarsource: Now removed, thanks

pavel-mikula-sonarsource (approved):
LGTM

PR #4802, opened 2021-08-18
S3431 fix VB implementation

csaba-sagi-sonarsource:


analyzers/src/SonarAnalyzer.Common/Rules/ExpectedExceptionShouldNotBeUsedAttributeBase.cs
```diff
@@ -25,7 +25,7 @@
 
 namespace SonarAnalyzer.Rules
 {
-    public abstract class ExpectedExceptionShouldNotBeUsedAttributeBase<TSyntaxKind> : SonarDiagnosticAnalyzer
+    public abstract class ExpectedExceptionAttributeShouldNotBeUsedBase<TSyntaxKind> : SonarDiagnosticAnalyzer
```
pavel-mikula-sonarsource, line 28: This file should be renamed accordingly

pavel-mikula-sonarsource (approved):
LGTM with one comment to address before merging

PR #4803, opened 2021-08-18
Update the nuget SDK (used in our unit test framework) to a non-deprecated version

dmitry-golovinov-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -23,21 +23,27 @@
 using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
-using System.Reflection;
+using System.Threading;
```
pavel-mikula-sonarsource, line 26: Having the `Package` class was a significant improvement in the codebase that simplified the code a lot. Can you implement the change there?
sonar-security is behind in this.

pavel-mikula-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
dmitry-golovinov-sonarsource: I was so happy to drop it :(

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -68,65 +66,37 @@ public static partial class NuGetMetadataFactory
...
-            else
-            {
-                LogMessage($"Package not found at {packageDir}");
-                PackageManager.InstallPackage("NETStandard.Library.Ref", SemanticVersion.ParseOptionalVersion("2.1.0"), ignoreDependencies: true, allowPrereleaseVersions: false);
-                if (!Directory.Exists(packageDir))
-                {
-                    throw new ApplicationException($"Test setup error: folder for downloaded package does not exist. Folder: {packageDir}");
-                }
-            }
+            var package = new Package("NETStandard.Library.Ref", "2.1.0", "netstandard2.1");
```
pavel-mikula-sonarsource, line 69: This method was here because the command-line tool could install this thing and we had to use a different approach. Now it looks that we can remove this method and use `.Create(...)` on the caller side instead.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -68,65 +66,37 @@ public static partial class NuGetMetadataFactory
...
-            }
-            var dllsPerDirectory = Directory.GetFiles(packageDirectory, "*.dll", SearchOption.AllDirectories)
-                .GroupBy(x => new FileInfo(x).Directory.Name)
-                .ToDictionary(x => x.Key.Split('+').First(), x => x.AsEnumerable());
+            var matchingDllsGroups = Directory.GetFiles(packageDirectory, "*.dll", SearchOption.AllDirectories)
+                                              .Select(path => new FileInfo(path))
+                                              .GroupBy(file => file.Directory.Name).ToArray();
+            var selectedGroup = matchingDllsGroups.Length == 1 && matchingDllsGroups[0].Key.EndsWith(".dll")
+                ? matchingDllsGroups[0]
+                : matchingDllsGroups.Join(
```
pavel-mikula-sonarsource, line 89: This painful `join` with weird grouping technique was very hard to understand and I refactored it to that more straight-forward way in this method. Can you restore that?

sonar-security is behind in this logic.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -45,149 +50,71 @@ public Package(string id, string version, string runtime)
...
-                if (filePath == null)
+                foreach (var dllFile in packageReader.GetFiles().Where(f => f.EndsWith(".dll", StringComparison.OrdinalIgnoreCase)))
                 {
-                    return;
+                    packageReader.ExtractFile(dllFile, $"{packageDirectory}\\{dllFile}", NullLogger.Instance);
                 }
-                File.WriteAllText(filePath, DateTime.Now.AddDays(VersionCheckDays).ToString("yyyy-MM-dd"));
             }
 
-            private string NextCheckFilePath()
```
pavel-mikula-sonarsource, line 184: sonar-security has a lightweight version of this because it didn't need to use the latest version. 

We need to preserve this logic, because "LATEST" on one day might not be the same version as "LATEST" one week later on the local dev build machine.

pavel-mikula-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -20,20 +20,27 @@
...
     public static partial class NuGetMetadataFactory
     {
         private class Package
         {
-            public string Id { get; }
-            public string Version { get; }
-            public string Runtime { get; }
+            private string Id { get; }
+            private string Version { get; set; }
+            private string Runtime { get; }
```
pavel-mikula-sonarsource, line 43: Since these became private, we don't need properties and we can make them fields.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -45,97 +52,71 @@ public Package(string id, string version, string runtime)
...
-                    process.StartInfo = startInfo;
-                    process.OutputDataReceived += (s, e) => LogMessage($"  nuget.exe: {e.Data}");
-                    process.ErrorDataReceived += OnErrorDataReceived;
+                var resource = await GetNuGetRepository();
 
-                    process.Start();
-                    process.BeginErrorReadLine();
-                    process.BeginOutputReadLine();
+                using var packageStream = new MemoryStream();
+                await resource.CopyNupkgToStreamAsync(
```
pavel-mikula-sonarsource, line 86: Is this too long to be a single line?
`CancellationToken.None` can be shortened to `default`

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -45,97 +52,71 @@ public Package(string id, string version, string runtime)
...
-                    {
-                        Install();
-                        WriteNextCheckTime();
-                    }
-                    else
-                    {
-                        LogMessage($"Skipping check for latest NuGet");
-                    }
+                    Version = GetLatestVersion().Result;
+                    writeTimeStamp = true;
```
pavel-mikula-sonarsource, line 65: Will something bad happen if we write it every time we install a package?

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -45,97 +52,71 @@ public Package(string id, string version, string runtime)
...
-                    LogMessage($"Path to nuget.config: {path}");
-                    return path;
+                    WriteNextCheckTime();
                 }
             }
 
+            private static async Task<FindPackageByIdResource> GetNuGetRepository()
+            {
+                var nugetOrgUrl = Settings.LoadSpecificSettings(NugetConfigFolderRelativePath,
+                                              "nuget.config")
```
pavel-mikula-sonarsource, line 111:
```suggestion
                var nugetOrgUrl = Settings.LoadSpecificSettings(NugetConfigFolderRelativePath,  "nuget.config")
```

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -160,34 +141,54 @@ private IEnumerable<string> SortedPackageFolders()
                     : Enumerable.Empty<string>();
             }
 
+            private async Task<string> GetLatestVersion()
+            {
+                var latest = SortedPackageFolders().Select(x => Path.GetFileName(x).Substring(Id.Length + 1)).LastOrDefault(x => char.IsNumber(x[0]));
```
pavel-mikula-sonarsource, line 146: This old design is weird. We're looking inside a fragile list to see when it was checked last time. We should have this information available beside the versions.
It doesn't matter on local machines, but on CI there's reused variable NUGET_PACKAGES.
So global cache is on 
NUGET_PACKAGES\packageid\version
and our cache (populated from the global one) is on
NUGET_PACKAGES\packageid.version

Let's change this to
NUGET_PACKAGES\packageid\Sonar.version
NUGET_PACKAGES\packageid\Sonar.NextCheckForUpdate.txt

Like that, we will have a fixed (version independent) location of the timestamp file.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -160,34 +141,54 @@ private IEnumerable<string> SortedPackageFolders()
...
+                if (latest == null || IsCheckForLatestPackageRequired())
+                {
+                    var resource =  await GetNuGetRepository();
+                    var versions = await resource.GetAllVersionsAsync(
+                        Id,
+                        new SourceCacheContext(),
+                        NullLogger.Instance,
+                        CancellationToken.None);
```
pavel-mikula-sonarsource, line 154: Can be single line. `CancellationToken.None` can be shortened to `default.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -160,34 +141,54 @@ private IEnumerable<string> SortedPackageFolders()
...
+                    ? timestamp
+                    : DateTime.MinValue;
                 LogMessage($"Next check for latest NuGets: {nextCheck}");
                 return nextCheck < DateTime.Now;
             }
 
             private void WriteNextCheckTime()
             {
-                const int VersionCheckDays = 5;
+                const int versionCheckDays = 5;
```
pavel-mikula-sonarsource, line 173: I'd prefer uppercase, because it doesn't change. It looks like a variable below.

pavel-mikula-sonarsource (commented):
More improvements

Corniel:
As this is about checking directories and files, I would suggest to represent files as `FileInfo`'s, and directories as `DirectoryInfo`'s.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -186,8 +166,11 @@ private string NextCheckFilePath()
                 // The file containing the next-check timestamp is stored in folder of the latest version of the package.
                 const string NextUpdateFileName = "NextCheckForUpdate.txt";
                 var directory = SortedPackageFolders().LastOrDefault();
-                return directory == null ? null : Path.Combine(directory, NextUpdateFileName);
+                return directory == null ? null : Path.Combine(directory, "..", NextUpdateFileName);
```
pavel-mikula-sonarsource, line 169: That should be same as this (the comment is outdated anyway)
```
private string NextCheckFilePath() =>
  Path.Combine(PackagesFolder, id, "NextCheckForUpdate.txt");
```

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -20,152 +20,132 @@
...
                 // : there might be multiple versions installed
                 // : there might be a package that starts with the same package id
                 //      e.g. Microsoft.AspNetCore.Core and Microsoft.AspNetCore.Core.Diagnostics
                 // Most packages have a three-part version, but some have four. We don't check
                 // the actual number of parts, as long as there is at least one.
-                var matcher = new Regex($@"{Regex.Escape(Id)}(\.\d+)+$", RegexOptions.IgnoreCase);
-
-                return Directory.Exists(PackagesFolder)
-                    ? Directory.GetDirectories(PackagesFolder, $"{Id}.*", SearchOption.TopDirectoryOnly).Where(x => matcher.IsMatch(x)).OrderBy(x => x)
+                var matcher = new Regex(@"(\.\d+)+$", RegexOptions.IgnoreCase);
```
pavel-mikula-sonarsource, line 123: Lines 119 and 120 above are now out of date and can be removed. It's not a concern anymore with the subdirectory.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -20,152 +20,132 @@
...
+                var packagePath = Path.Combine(PackagesFolder, id);
+                return Directory.Exists(packagePath)
+                    ? Directory.GetDirectories(packagePath, "*.*", SearchOption.TopDirectoryOnly).Where(x => matcher.IsMatch(x)).OrderBy(x => x)
```
pavel-mikula-sonarsource, line 126: `SearchOption.TopDirectoryOnly` is the default value so we don't need it.
```suggestion
                    ? Directory.GetDirectories(packagePath, "*.*").Where(x => matcher.IsMatch(x)).OrderBy(x => x)
```

pavel-mikula-sonarsource: We should use only our own directories, as the folder is shared with `$(NUGET_PACKAGES)` on CI and some local machines.
`$(NUGET_PACKAGES)\packageId\1.2.3` - this would be the NuGet global cache that we should not mess with.
Similar update will be needed in `PackageDirectory()`
```suggestion
                    ? Directory.GetDirectories(packagePath, "Sonar.*.*", SearchOption.TopDirectoryOnly).Where(x => matcher.IsMatch(x)).OrderBy(x => x)
```

pavel-mikula-sonarsource (changes requested):
One more design round

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
pavel-mikula-sonarsource: Why does it need the `SortedPackageFolders`? I think we can really delete most of it and directly generate the returned value.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -20,152 +20,135 @@
...
                 {
-                    var packageDir = PackageDirectory();
-                    if (Directory.Exists(packageDir))
-                    {
-                        LogMessage($"Package found at {packageDir}");
-                    }
-                    else
+                    LogMessage($"Package not found at {packageDir}, will attempt to download and install.");
+                    InstallPackageAsync(latestVersionObtained).Wait();
+                    if (!Directory.Exists(packageDir))
```
pavel-mikula-sonarsource, line 72: This condition should not be sniffing around the presumable results of `InstallPackageAsync`. Let's do this check in side `InstallPackageAsync` differently - for example
```
var dllFiles = packageReader.GetFiles...ToArray();
if(dllFiles.Any())
  foreach
else
  throw a better message about missing DLL files
```

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -20,152 +20,135 @@
...
-                    if (IsCheckForLatestPackageRequired())
-                    {
-                        Install();
-                        WriteNextCheckTime();
-                    }
-                    else
-                    {
-                        LogMessage($"Skipping check for latest NuGet");
-                    }
+                    version = GetLatestVersion().Result;
```
pavel-mikula-sonarsource, line 63: Let's do this in the constructor and make it readonly. It's a weird side effect.

pavel-mikula-sonarsource (changes requested):
Few redesigns to clean stuff up

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
dmitry-golovinov-sonarsource: This call is useless indeed. But we still need to verify that package folder exists to not write file to non-existing directory.

dmitry-golovinov-sonarsource: Okay, this is very corner case, indeed, I will drop it, otherwise we could hide the package unpacking failure

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -20,152 +20,135 @@
...
                 // Most packages have a three-part version, but some have four. We don't check
                 // the actual number of parts, as long as there is at least one.
-                var matcher = new Regex($@"{Regex.Escape(Id)}(\.\d+)+$", RegexOptions.IgnoreCase);
-
-                return Directory.Exists(PackagesFolder)
-                    ? Directory.GetDirectories(PackagesFolder, $"{Id}.*", SearchOption.TopDirectoryOnly).Where(x => matcher.IsMatch(x)).OrderBy(x => x)
+                var matcher = new Regex(@"(\d+\.?)+$", RegexOptions.IgnoreCase);
+                var packagePath = Path.Combine(PackagesFolder, id);
+                return Directory.Exists(packagePath)
+                    ? Directory.GetDirectories(packagePath, "Sonar.*.*").Where(x => matcher.IsMatch(x)).Select(x => matcher.Match(x).Groups[0].Value).OrderBy(x => x)
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Select is used to extract version number, we can drop where clause, agreed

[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: And it's looks pretty easy and straightforward with a single group.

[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: I will fix OrderBy, since NuGet version supports ordering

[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Ok, changed regex a little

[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Not even needed now

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.Package.cs
```diff
@@ -20,152 +20,135 @@
...
-                return Path.Combine(PackagesFolder, $@"{Id}.{versionFolder}\{runtimePath}lib");
+                var runtimePath = runtime == null ? string.Empty : $"runtimes\\{runtime}\\";
+                var combinedPath = Path.Combine(PackagesFolder, id, $"Sonar.{version}", runtimePath);
+                return Path.GetFullPath(combinedPath);
             }
 
             public void EnsurePackageIsInstalled()
             {
-                if (Version == Constants.NuGetLatestVersion)
+                var latestVersionObtained = version == Constants.NuGetLatestVersion;
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Nice catch. Thanks

Corniel:
@pavel-mikula-sonarsource  What about using the DirectoryInfo/FileInfo API here? Also, consider using DateTime.UtcNow instead of DateTime.Now.

pavel-mikula-sonarsource:
Hi @Corniel, in this case, FileInfo/DirectoryInfo would be just needed indirection for those simple `Path.` APIs. 
The `Now` is also better here because we're in test infrastructure and we mostly care about runs on local machines where local-machine-time is desired.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -66,67 +64,36 @@ public static partial class NuGetMetadataFactory
...
         {
-            Console.WriteLine();
-            Console.WriteLine($"Package: {package.Id}, {package.Version}");
             package.EnsurePackageIsInstalled();
-
-            var packageDirectory = package.PackageDirectory();
-            if (!Directory.Exists(packageDirectory))
+            var dllsPerDirectory = new Dictionary<string, IEnumerable<string>>();
+            var packageLibDirectory = package.PackageDirectory();
+            var foldersWithDlls = new List<string> { Path.Combine(packageLibDirectory, "lib"), Path.Combine(packageLibDirectory, "ref") };
```
pavel-mikula-sonarsource, line 73: I don't understand the `Lib` part in `packageLibDirectory`. The next line uses it to construct `lib` or `ref`. So it should be `packageDirectory` instead.

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -66,67 +64,36 @@ public static partial class NuGetMetadataFactory
...
+            foreach (var folder in foldersWithDlls.Where(Directory.Exists))
```
pavel-mikula-sonarsource, line 74: We should not not need the list, we won't add to it.
```suggestion
            foreach (var folder in new[] { "lib", "ref" }.Select(x => Path.Combine(packageLibDirectory, x)).Where(Directory.Exists))

```

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -66,67 +64,36 @@ public static partial class NuGetMetadataFactory
...
             {
-                throw new ApplicationException($"Test setup error: Package directory doesn't exist: {packageDirectory}");
+                dllsPerDirectory.AddRange(DllsPerDirectory(folder).Where(x => !dllsPerDirectory.ContainsKey(x.Key)));
```
pavel-mikula-sonarsource, line 76: What is this whole lib/ref trying to solve? That allowed directory can be twice in the list, once from `lib` and once from `ref`?
I'd prefer the old implementation and fixing this problem there (if that's the problem).

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -66,67 +64,36 @@ public static partial class NuGetMetadataFactory
...
             // if this throws because it doesn't find a DLL (because the lib contains "_._"), maybe you are referencing the wrong DLL (check the dependencies of the DLL you use)
             var directory = allowedDirectories.FirstOrDefault(x => dllsPerDirectory.ContainsKey(x))
-                ?? throw new InvalidOperationException($"No allowed directory with DLL files was found in {packageDirectory}. " +
-                                                        "Add new target framework to SortedAllowedDirectories or set targetFramework argument explicitly.");
-            var dlls = dllsPerDirectory[directory];
-            foreach (string filePath in dlls)
+                            ?? throw new InvalidOperationException($"No allowed directory with DLL files was found in {packageLibDirectory}. " +
+                                                                   "Add new target framework to SortedAllowedDirectories or set targetFramework argument explicitly.");
+            var dlls = dllsPerDirectory[directory].ToList();
+            foreach (var filePath in dlls)
```
pavel-mikula-sonarsource, line 84: We don't need the list, we'll just iterate it once. Improving `foreach` variable name can remove the `dlls`.
```suggestion
            foreach (var dllPath in dllsPerDirectory[directory])
```

pavel-mikula-sonarsource (changes requested):
Last round on the `Package`, big round on the factory :)

analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
dmitry-golovinov-sonarsource: It is trying to solve the problem, that for some libraries the target framework we need exists only in ref and not in lib. And yeah, sometimes ref contains duplicated stuff.

// some packages (see Mono.Posix.NETStandard.1.0.0) may contain target framework only in ref folder
that comment disappeared during refactoring.

PR #4805, opened 2021-08-18
Simplify S3981 implementation DoNotCheckZeroSizeCollection

Corniel:
Follow-up on #4648

### DIFF
--- a/analyzers/its/expected/Net5/Net5--net5.0-S3981.json
+++ b/analyzers/its/expected/Net5/Net5--net5.0-S3981.json
@@ -2,7 +2,7 @@
 "issues":  [
 {
 "id":  "S3981",
-"message":  "The 'Count' of 'IEnumerable<T>' is always 'True', so fix this test to get the real expected behavior.",
+"message":  "The 'Count' of 'IEnumerable<T>' always evaluates as 'True' regardless the size.",
 "location":  {
 "uri":  "sources\Net5\Net5\S3981.cs",
 "region":  {
--- a/analyzers/its/expected/Net5/Net5--net5.0-S3981.json
+++ b/analyzers/its/expected/Net5/Net5--net5.0-S3981.json
@@ -2,7 +2,7 @@
 "issues":  [
 {
 "id":  "S3981",
-"message":  "The 'Count' of 'IEnumerable<T>' is always '>=0', so fix this test to get the real expected behavior.",
+"message":  "The 'Count' of 'IEnumerable<T>' is always 'True', so fix this test to get the real expected behavior.",
 "location":  {
 "uri":  "sources\Net5\Net5\S3981.cs",
 "region":  {
--- a/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
@@ -111,6 +111,7 @@ namespace SonarAnalyzer.Helpers.Facade
             return whenNotNull;
         }
 
-        public override SyntaxNode RemoveParentheses(SyntaxNode node) => node.RemoveParentheses();
+        public override SyntaxNode RemoveParentheses(SyntaxNode node) => 
+            node.RemoveParentheses();
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
@@ -111,7 +111,7 @@ namespace SonarAnalyzer.Helpers.Facade
             return whenNotNull;
         }
 
-        public override SyntaxNode RemoveParentheses(SyntaxNode node) => 
+        public override SyntaxNode RemoveParentheses(SyntaxNode node) =>
             node.RemoveParentheses();
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
@@ -102,6 +102,16 @@ namespace SonarAnalyzer.Helpers.Facade
                 _ => string.Empty
             };
 
+        public override SyntaxNode RemoveConditionalAcesss(SyntaxNode node)
+        {
+            var whenNotNull = node.RemoveParentheses();
+            while (whenNotNull is ConditionalAccessExpressionSyntax conditionalAccess)
+            {
+                whenNotNull = conditionalAccess.WhenNotNull.RemoveParentheses();
+            }
+            return whenNotNull;
+        }
+
         public override SyntaxNode RemoveParentheses(SyntaxNode node) => node.RemoveParentheses();
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
@@ -89,8 +89,7 @@ namespace SonarAnalyzer.Helpers.Facade
                 PropertyDeclarationSyntax property => property.Identifier,
                 SimpleNameSyntax simpleName => simpleName.Identifier,
                 VariableDeclaratorSyntax variable => variable.Identifier,
-                null => null,
-                _ => throw InvalidOperation(node, nameof(NodeIdentifier)),
+                _ => null,
             };
 
         public override string NodeStringTextValue(SyntaxNode node) =>
--- a/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
@@ -72,12 +72,13 @@ namespace SonarAnalyzer.Helpers.Facade
                 InvocationExpressionSyntax invocation => invocation.Expression,
                 LockStatementSyntax @lock => @lock.Expression,
                 null => null,
-                _ => throw Unexpected(node)
+                _ => throw InvalidOperation(node, nameof(NodeExpression))
             };
 
         public override SyntaxToken? NodeIdentifier(SyntaxNode node) =>
             RemoveParentheses(node) switch
             {
+                AssignmentExpressionSyntax assignment => NodeIdentifier(assignment.Left),
                 AttributeArgumentSyntax attribute => attribute.NameColon?.Name.Identifier,
                 BaseTypeDeclarationSyntax baseType => baseType.Identifier,
                 DelegateDeclarationSyntax delegateDeclaration => delegateDeclaration.Identifier,
@@ -90,7 +91,7 @@ namespace SonarAnalyzer.Helpers.Facade
                 SimpleNameSyntax simpleName => simpleName.Identifier,
                 VariableDeclaratorSyntax variable => variable.Identifier,
                 null => null,
-                _ => throw Unexpected(node)
+                _ => throw InvalidOperation(node, nameof(NodeIdentifier)),
             };
 
         public override string NodeStringTextValue(SyntaxNode node) =>
--- a/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
@@ -78,7 +78,6 @@ namespace SonarAnalyzer.Helpers.Facade
         public override SyntaxToken? NodeIdentifier(SyntaxNode node) =>
             RemoveParentheses(node) switch
             {
-                AssignmentExpressionSyntax assignment => NodeIdentifier(assignment.Left),
                 AttributeArgumentSyntax attribute => attribute.NameColon?.Name.Identifier,
                 BaseTypeDeclarationSyntax baseType => baseType.Identifier,
                 DelegateDeclarationSyntax delegateDeclaration => delegateDeclaration.Identifier,
--- a/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
@@ -53,6 +53,12 @@ namespace SonarAnalyzer.Helpers.Facade
 
         public override bool IsNullLiteral(SyntaxNode node) => node.IsNullLiteral();
 
+        public override SyntaxNode BinaryExpressionLeft(SyntaxNode binaryExpression) =>
+            Cast<BinaryExpressionSyntax>(binaryExpression).Left;
+
+        public override SyntaxNode BinaryExpressionRight(SyntaxNode binaryExpression) =>
+            Cast<BinaryExpressionSyntax>(binaryExpression).Right;
+
         public override IEnumerable<SyntaxNode> EnumMembers(SyntaxNode @enum) =>
             @enum == null ? Enumerable.Empty<SyntaxNode>() : Cast<EnumDeclarationSyntax>(@enum).Members;
 
@@ -70,17 +76,19 @@ namespace SonarAnalyzer.Helpers.Facade
             };
 
         public override SyntaxToken? NodeIdentifier(SyntaxNode node) =>
-            node switch
+            RemoveParentheses(node) switch
             {
                 AttributeArgumentSyntax attribute => attribute.NameColon?.Name.Identifier,
                 BaseTypeDeclarationSyntax baseType => baseType.Identifier,
+                ConditionalAccessExpressionSyntax conditionalAccess => NodeIdentifier(conditionalAccess.WhenNotNull),
                 DelegateDeclarationSyntax delegateDeclaration => delegateDeclaration.Identifier,
                 EnumMemberDeclarationSyntax enumMember => enumMember.Identifier,
-                SimpleNameSyntax simpleName => simpleName.Identifier,
+                InvocationExpressionSyntax invocation => NodeIdentifier(invocation.Expression),
                 MemberAccessExpressionSyntax memberAccess => memberAccess.Name.Identifier,
                 MemberBindingExpressionSyntax memberBinding => memberBinding.Name.Identifier,
                 ParameterSyntax parameter => parameter.Identifier,
                 PropertyDeclarationSyntax property => property.Identifier,
+                SimpleNameSyntax simpleName => simpleName.Identifier,
                 VariableDeclaratorSyntax variable => variable.Identifier,
                 null => null,
                 _ => throw Unexpected(node)
@@ -94,7 +102,6 @@ namespace SonarAnalyzer.Helpers.Facade
                 _ => string.Empty
             };
 
-        public override SyntaxNode RemoveParentheses(SyntaxNode node) =>
-            node.RemoveParentheses();
+        public override SyntaxNode RemoveParentheses(SyntaxNode node) => node.RemoveParentheses();
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxKindFacade.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxKindFacade.cs
@@ -26,7 +26,11 @@ namespace SonarAnalyzer.Helpers.Facade
     internal sealed class CSharpSyntaxKindFacade : ISyntaxKindFacade<SyntaxKind>
     {
         public SyntaxKind Attribute => SyntaxKind.Attribute;
-        public SyntaxKind[] ClassAndRecordDeclaration => new[] { SyntaxKind.ClassDeclaration, SyntaxKindEx.RecordDeclaration };
+        public SyntaxKind[] ClassAndRecordDeclaration => new[]
+        {
+            SyntaxKind.ClassDeclaration,
+            SyntaxKindEx.RecordClassDeclaration,
+        };
         public SyntaxKind[] ComparisonKinds => new[]
         {
             SyntaxKind.GreaterThanExpression,
@@ -53,7 +57,8 @@ namespace SonarAnalyzer.Helpers.Facade
             SyntaxKind.StructDeclaration,
             SyntaxKind.InterfaceDeclaration,
             SyntaxKind.EnumDeclaration,
-            SyntaxKindEx.RecordDeclaration,
+            SyntaxKindEx.RecordClassDeclaration,
+            SyntaxKindEx.RecordStructDeclaration,
         };
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxKindFacade.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxKindFacade.cs
@@ -26,7 +26,16 @@ namespace SonarAnalyzer.Helpers.Facade
     internal sealed class CSharpSyntaxKindFacade : ISyntaxKindFacade<SyntaxKind>
     {
         public SyntaxKind Attribute => SyntaxKind.Attribute;
-        public SyntaxKind[] ClassAndRecordDeclaration => new[] {SyntaxKind.ClassDeclaration, SyntaxKindEx.RecordClassDeclaration};
+        public SyntaxKind[] ClassAndRecordDeclaration => new[] { SyntaxKind.ClassDeclaration, SyntaxKindEx.RecordDeclaration };
+        public SyntaxKind[] ComparisonKinds => new[]
+        {
+            SyntaxKind.GreaterThanExpression,
+            SyntaxKind.GreaterThanOrEqualExpression,
+            SyntaxKind.LessThanExpression,
+            SyntaxKind.LessThanOrEqualExpression,
+            SyntaxKind.EqualsExpression,
+            SyntaxKind.NotEqualsExpression,
+        };
         public SyntaxKind ConstructorDeclaration => SyntaxKind.ConstructorDeclaration;
         public SyntaxKind EnumDeclaration => SyntaxKind.EnumDeclaration;
         public SyntaxKind IdentifierName => SyntaxKind.IdentifierName;
@@ -34,7 +43,7 @@ namespace SonarAnalyzer.Helpers.Facade
         public SyntaxKind InvocationExpression => SyntaxKind.InvocationExpression;
         public SyntaxKind InterpolatedStringExpression => SyntaxKind.InterpolatedStringExpression;
         public SyntaxKind[] MethodDeclarations => new[] { SyntaxKind.MethodDeclaration };
-        public SyntaxKind[] ObjectCreationExpressions => new[] {SyntaxKind.ObjectCreationExpression, SyntaxKindEx.ImplicitObjectCreationExpression};
+        public SyntaxKind[] ObjectCreationExpressions => new[] { SyntaxKind.ObjectCreationExpression, SyntaxKindEx.ImplicitObjectCreationExpression };
         public SyntaxKind ReturnStatement => SyntaxKind.ReturnStatement;
         public SyntaxKind SimpleMemberAccessExpression => SyntaxKind.SimpleMemberAccessExpression;
         public SyntaxKind StringLiteralExpression => SyntaxKind.StringLiteralExpression;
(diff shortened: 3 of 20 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotCheckZeroSizeCollection.cs
```diff
@@ -33,14 +33,29 @@ namespace SonarAnalyzer.Rules.CSharp
...
+        protected override ExpressionSyntax Right(SyntaxNode binary) => ((BinaryExpressionSyntax)binary).Right;
+        protected override ExpressionSyntax Trim(ExpressionSyntax expression)
+        {
+            while (expression.RemoveParentheses() is ConditionalAccessExpressionSyntax conditionalAccess)
+            {
+                expression = conditionalAccess.WhenNotNull;
+            }
+            return (expression is InvocationExpressionSyntax invocation
+                ? invocation.Expression
+                : expression).RemoveParentheses();
```
pavel-mikula-sonarsource, line 50: let's make this single line, the parentheses are confusing to read on multiline statement

analyzers/src/SonarAnalyzer.Common/Extensions/CountComparisonResultExtensions.cs
```diff
@@ -73,7 +76,7 @@ comparison switch
                 CountComparisonResult.Empty,
                 CountComparisonResult.AlwaysFalse);
 
-        private static CountComparisonResult Check(int count, int limit, CountComparisonResult smallerThanLimit, CountComparisonResult sameAsLimit, CountComparisonResult higherThanLimit)
+        private static CountComparisonResult Check(int count, int limit, CountComparisonResult higherThanLimit, CountComparisonResult sameAsLimit, CountComparisonResult smallerThanLimit)
```
pavel-mikula-sonarsource, line 79: What happened here? How did we passed UTs if this was swapped?

I'd prefer `small, same, high` as a logical order.

pavel-mikula-sonarsource (changes requested):
Hi, thank you for taking things to the end.

I took just a first round quick look for now.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCheckZeroSizeCollection.CSharp9.cs
```diff
@@ -6,9 +6,9 @@
 var someEnumerable = new List<string>();
 var anotherEnumerable = new List<string>();
 
-var result = someEnumerable.Count() >= 0; // Noncompliant {{The 'Count' of 'IEnumerable<T>' is always '>=0', so fix this test to get the real expected behavior.}}
+var result = someEnumerable.Count() >= 0; // Noncompliant {{The 'Count' of 'IEnumerable<T>' makes no sense, so fix this test to get the real expected behavior.}}
```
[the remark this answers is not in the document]
Corniel: > While looking at the message itself without content (let's say in the error list or build log), the message is confusing. We'd need to reword it somehow to reference the comparison operation that doesn't change its result.

I agree. This is what is currently in the RSPEC: "Collection sizes and array length comparisons should make sense", is it was a first update, to fire the debate.

analyzers/src/SonarAnalyzer.Common/Extensions/CountComparisonResultExtensions.cs
Corniel: > What happened here? How did we passed UTs if this was swapped?

For Any/Empty we're only interested in equal to the limit. That's why. I was shocked to find this out.

> I'd prefer `small, same, high` as a logical order.

I first wanted to make things work, as I expected some discussion about this. :)

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotCheckZeroSizeCollection.cs
Corniel: Do we want RemoveParentheses on the LanguageFacade? I think it used a lot.

analyzers/src/SonarAnalyzer.Common/Helpers/ExpressionNumericConverterBase.cs
```diff
@@ -31,32 +31,41 @@ public abstract class ExpressionNumericConverterBase<TLiteralExpressionSyntax, T
...
-                return false;
+                value = multiplierCalculator(multiplier, value);
+                return true;
             }
-
-            if (internalExpression is TLiteralExpressionSyntax literalExpression &&
-                ConversionHelper.TryConvertWith(TokenValue(literalExpression), converter, out value))
+            else if (semanticModel is { }
+                && semanticModel.GetConstantValue(expression) is Optional<object> optional
+                && optional.HasValue)
```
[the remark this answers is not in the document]
Corniel: I was not aware that this was possible. Nice one.

analyzers/src/SonarAnalyzer.Common/Extensions/CountComparisonResultExtensions.cs
Corniel: After writing a test, I could safely refactor. I think it is better now.

Corniel:
As far as I can tell, there are no outstanding issue. Can you check what you still want to be done?
Furthere more I noticed a lot of warnings (SA*s) are popping up, can you elaborate on why that is? (like the `this.` prefix, is that intended)

Corniel:
@pavel-mikula-sonarsource can we try to finish this?

pavel-mikula-sonarsource:
Hi @Corniel , I see there are 500 changed files and commits from master that are not on the feature branch => something went wrong with the rebase :(

Shall I rebase the feature branch on the latest master-commit that is here or will you try to rebase this branch on top of the feature branch (without newer master commits)?

Corniel:
Oops, I rebased on master. Maybe you could do a rebase as well? (obviously, I should rebase afterwards again)

pavel-mikula-sonarsource:
I'll rebase the feature branch tomorrow on `master`. You can do the same in parallel.

pavel-mikula-sonarsource:
The feature branch is rebased on the latest master. There was only one easy conflict to solve because we renamed `ReportDiagnosticWhenActive` to `ReportIssue`.

pavel-mikula-sonarsource:
Oh no, I forgot to pull the remote before rebasing so we don't have #4648 on the feature branch now. I'm trying to fix it

pavel-mikula-sonarsource:
It should be fixed now. Please check that the feature branch has all commits that you expect it to have and I haven't missed any.

Corniel:
@pavel-mikula-sonarsource tests fail because it claims that the parameters of the messages have not been filled. I'm not sure, why, as I can see (debugging) that those are replaced. Might it be that the new way of reporting issues makes, that I have to change some code?

Corniel:
@pavel-mikula-sonarsource All should work now. Lets try to merge this quickly. (Obviously, re-evaluate all too)

analyzers/src/SonarAnalyzer.Common/Rules/DoNotCheckZeroSizeCollectionBase.cs
```diff
@@ -19,124 +19,134 @@
...
-                context.ReportIssue(Diagnostic.Create(SupportedDiagnostics[0], issueLocation.GetLocation(), symbol.Name, symbolType));
+                context.ReportIssue(Diagnostic.Create(rule, issue.GetLocation(), symbol.Name, symbolType, result == CountComparisonResult.AlwaysTrue));
             }
         }
 
+        private ISymbol CollectionSizeSymbol(SyntaxNode expression, SemanticModel semanticModel) =>
+            HasCandidateName(Language.Syntax.NodeIdentifier(expression)?.ValueText)
+            && semanticModel.GetSymbolInfo(expression).Symbol is ISymbol symbol
+            && IsCollecionSizeCheck(symbol)
+            ? symbol : null;
```
pavel-mikula-sonarsource, line 83: I don't like the fact that `IsCollecionSizeCheck` and `GetDeclaringTypeName` do the same thing. And `GetDeclaringTypeName` has hidden dependency on `IsCollecionSizeCheck` making the `return null` unreachable.
Let's:
* Remove `IsCollecionSizeCheck` completely
* Rename `GetDeclaringTypeName` to `CollecionSizeTypeName`
* Remove `CollectionSizeSymbol` and move it's content to `CheckExpression` like this:
```
            var result = comparison.Compare(constant);
            expression = Language.Syntax.RemoveConditionalAcesss(expression);
            if (result.IsInvalid()
                && HasCandidateName(Language.Syntax.NodeIdentifier(expression)?.ValueText)
                && context.SemanticModel.GetSymbolInfo(expression).Symbol is ISymbol symbol
                && CollecionSizeTypeName(symbol) is { } symbolType)
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCheckZeroSizeCollection.CSharp9.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I like the current message.

analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
```diff
@@ -70,17 +76,19 @@ node switch
             };
 
         public override SyntaxToken? NodeIdentifier(SyntaxNode node) =>
-            node switch
+            RemoveParentheses(node) switch
             {
                 AttributeArgumentSyntax attribute => attribute.NameColon?.Name.Identifier,
                 BaseTypeDeclarationSyntax baseType => baseType.Identifier,
+                ConditionalAccessExpressionSyntax conditionalAccess => NodeIdentifier(conditionalAccess.WhenNotNull),
```
pavel-mikula-sonarsource, line 83: Let's remove this one. It's not used and it's not clear whether it should return identifier from `Expression` or `WhenNotNull` part from the `NodeIdentifier` method name.
```suggestion
```

analyzers/src/SonarAnalyzer.VisualBasic/Facade/VisualBasicSyntaxFacade.cs
```diff
@@ -70,16 +76,18 @@ node switch
...
 
         public override SyntaxToken? NodeIdentifier(SyntaxNode node) =>
-            node switch
+            RemoveParentheses(node) switch
             {
                 EnumStatementSyntax enumStatement => enumStatement.Identifier,
                 EnumMemberDeclarationSyntax enumMember => enumMember.Identifier,
-                SimpleArgumentSyntax simpleArgument => simpleArgument.NameColonEquals?.Name.Identifier,
-                SimpleNameSyntax simpleName => simpleName.Identifier,
+                ConditionalAccessExpressionSyntax conditionalAccess => NodeIdentifier(conditionalAccess.WhenNotNull),
```
pavel-mikula-sonarsource, line 83: Same as C# - let's remove this
```suggestion
```

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotCheckZeroSizeCollection.cs
```diff
@@ -126,18 +104,34 @@ private void AnalyzeSwitchStatement(SyntaxNodeAnalysisContext c)
...
+                    expression,
+                    constant,
+                    comparison);
             }
-        }
 
-        private static bool IsOperatorOfInterest(SyntaxToken syntaxToken) =>
-            syntaxToken.ValueText.Equals(">=") || syntaxToken.ValueText.Equals("<");
+            static ComparisonKind ComparisonKind(SyntaxToken syntaxToken) =>
+                syntaxToken.ValueText switch
```
pavel-mikula-sonarsource, line 125: `==`, `!=`, `<=`, `>` and `_` cases are not covered. Can you add test cases, please?

analyzers/src/SonarAnalyzer.Common/Helpers/ExpressionNumericConverterBase.cs
```diff
@@ -31,32 +31,40 @@ public abstract class ExpressionNumericConverterBase<TLiteralExpressionSyntax, T
...
+                return true;
             }
-
-            if (internalExpression is TLiteralExpressionSyntax literalExpression &&
-                ConversionHelper.TryConvertWith(TokenValue(literalExpression), converter, out value))
+            else if (semanticModel is { }
+                && semanticModel.GetConstantValue(expression) is { HasValue: true } optional)
             {
-                value = multiplierCalculator(multiplier.Value, value);
+                value = (T)optional.Value;
```
pavel-mikula-sonarsource, line 60: This throws `InvalidCastException` for this test case. We need to check that value is `T` before casting.
```
result = result.ToString() == "Not integer";
```

pavel-mikula-sonarsource (changes requested):
Last big round

analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
Corniel: Fine with me. I'm pretty sure I added it, because I needed it, but that might be part of the fixer. We'll run into that later.

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotCheckZeroSizeCollection.cs
Corniel: Sure.

analyzers/src/SonarAnalyzer.Common/Helpers/ExpressionNumericConverterBase.cs
Corniel: Done.

analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxKindFacade.cs
```diff
@@ -44,7 +57,8 @@ internal sealed class CSharpSyntaxKindFacade : ISyntaxKindFacade<SyntaxKind>
             SyntaxKind.StructDeclaration,
             SyntaxKind.InterfaceDeclaration,
             SyntaxKind.EnumDeclaration,
-            SyntaxKindEx.RecordClassDeclaration
+            SyntaxKindEx.RecordClassDeclaration,
```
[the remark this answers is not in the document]
Corniel: I'm not sure why, but I assume we want railing comma's?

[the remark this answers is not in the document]
pavel-mikula-sonarsource: GitHub didn't show it properly here 😒 If you take a look at File changes, it says:
Before
```
SyntaxKindEx.RecordClassDeclaration
```
after
```
SyntaxKindEx.RecordClassDeclaration,
SyntaxKindEx.RecordStructDeclaration,
```

So my question is why `RecordStructDeclaration` is marked as added in this PR?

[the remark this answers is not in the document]
pavel-mikula-sonarsource: (the trailing comma is not important)

analyzers/src/SonarAnalyzer.Common/Helpers/ExpressionNumericConverterBase.cs
pavel-mikula-sonarsource: Add UT for this please. Something should fail if we remove the change.

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotCheckZeroSizeCollection.cs
pavel-mikula-sonarsource: The `==` and `!=` are actually not valid. Do we need them in the implementation then?

analyzers/src/SonarAnalyzer.Common/Helpers/ExpressionNumericConverterBase.cs
```diff
@@ -31,32 +31,41 @@ public abstract class ExpressionNumericConverterBase<TLiteralExpressionSyntax, T
...
-                return false;
+                value = multiplierCalculator(multiplier, value);
+                return true;
             }
-
-            if (internalExpression is TLiteralExpressionSyntax literalExpression &&
-                ConversionHelper.TryConvertWith(TokenValue(literalExpression), converter, out value))
+            else if (semanticModel is { }
+                && semanticModel.GetConstantValue(expression) is { HasValue: true } optional
+                && optional.Value is T)
```
pavel-mikula-sonarsource, line 59: _Nitpick:_ This should be a tiny bit safer
```suggestion
                && optional.Value is T val)
```
and then `value = val;`

analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxKindFacade.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: In general, the change makes sense so we can keep it.

analyzers/src/SonarAnalyzer.CSharp/Rules/DoNotCheckZeroSizeCollection.cs
Corniel: @pavel-mikula-sonarsource I'm not sure, Is there an other way to use these patterns, I'm not aware off?

analyzers/src/SonarAnalyzer.Common/Helpers/ExpressionNumericConverterBase.cs
pavel-mikula-sonarsource: Please add the UT that I mentioned in this thread, that should be the last thing to do..

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCheckZeroSizeCollection.cs
```diff
@@ -44,28 +45,29 @@ public void TestCountMethod()
...
             result = someEnumerable.Count() >= localConst_NonZero;
+            result = someEnumerable.Count() >= ConstField_Double_Zero; // Compliant, double is ignored
 
             result = (someEnumerable.Count()) >= (0); // Noncompliant
             result = ((((someEnumerable).Count())) >= ((0))); // Noncompliant
-//                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+            //        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
             result = 0 <= someEnumerable.Count(); // Noncompliant
-//                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+            //       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
pavel-mikula-sonarsource, line 70:
```suggestion
            //       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
            
            result = result.ToString() == "Not integer";            
```

analyzers/src/SonarAnalyzer.Common/Helpers/ExpressionNumericConverterBase.cs
pavel-mikula-sonarsource: Or not, I can do it :)

Corniel:
@pavel-mikula-sonarsource Adding the `== double` const already hit the path. Having the string constant is fine. Also raises the question if we want the constant converter be able to deal with double == int, and long == int expressions.

pavel-mikula-sonarsource:
> @pavel-mikula-sonarsource Adding the `== double` const already hit the path. Having the string constant is fine. Also raises the question if we want the constant converter be able to deal with double == int, and long == int expressions.

I've added the `string` version as we might add support for `double` in that class for some other rule. Otherwise we don't need the `long` for now.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCheckZeroSizeCollection.cs
```diff
@@ -44,28 +45,31 @@ public void TestCountMethod()
...
+            
+            result = result.ToString() == "Not integer";
```
pavel-mikula-sonarsource, line 72: IT threw this exception, please add this UT

>AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.DoNotCheckZeroSizeCollection' threw an exception of type 'System.InvalidOperationException' with message 'Unexpected node: AssignmentExpressionSyntax'.

```suggestion
            result = result.ToString() == "Not integer";            
            result = localVariable = 42 == 0;
            result = 0 == localVariable = 42;

```

pavel-mikula-sonarsource (changes requested):
There's one more issue to fix

pavel-mikula-sonarsource:
@Corniel Will you have some time to fix the exception from the last comment? This is close to being merged.

Corniel:
@pavel-mikula-sonarsource done. I extended some exception handling (gave some insides).

analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
```diff
@@ -66,24 +72,26 @@ node switch
...
                 null => null,
-                _ => throw Unexpected(node)
+                _ => throw InvalidOperation(node, nameof(NodeExpression))
             };
 
         public override SyntaxToken? NodeIdentifier(SyntaxNode node) =>
-            node switch
+            RemoveParentheses(node) switch
             {
+                AssignmentExpressionSyntax assignment => NodeIdentifier(assignment.Left),
```
pavel-mikula-sonarsource, line 81: This line is in a common code and makes sense only for this particular rule. The root cause of this problem needs to be fixed on the invocation side. Mainly because there could be other unexpected syntaxes that could lead that path to be reached.
```suggestion
```

pavel-mikula-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
Corniel: Fare enough, but I have no clue how this is triggered. Do you?

pavel-mikula-sonarsource: I thought the last code snippet will reproduce it. If I comment the `CSharpSyntaxFacade` line, I get the exception on these two:
```
result = (localVariable = 42) != -1);

if ((localVariable = 42) != -1) { }
```

Corniel: Now it does. 
``` C#
result = (localVariable = 42) == 0;
result = 0 == (localVariable = 42);
```

Do not evaluate as alwaysTrue of alwaysFalse. The question is, how to deal with this. I agree that the first attempt has downsides, which leaves us to the following options:
 1. return null for Assigenments
 2. catch the exception
 3. have an HasIdentifier
 4. Something I can not think of right now

And I dislike 1,2, and 3 for different reasons.

pavel-mikula-sonarsource: I'd go for option 4 😄 

Update `NodeIdentifier` from
```
null => null,
_ => throw InvalidOperation(node, nameof(NodeIdentifier)),
```
to
```
_ => null,
```

Background reasoning:
* `HasIdentifier` option is the best idea. And implementing it is effectively the same as `NodeIdentifier != null` except the exception
* It's hard to predict what will arrive in `NodeIdentifier`, we can accept it as a fact.
* Returning `null` from the method is already expected behavior on many paths

Corniel: I think that was option 1. I'll do that. Also reconsidered that being the best option.

pavel-mikula-sonarsource: I understood the option 1 as `AssignmentExpressionSyntax  => null`

Corniel: @pavel-mikula-sonarsource I understand that your read that into it, but that was not what I intended. Anyhow. I changed the implementation as agreed on, and now, all tests pass.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DoNotCheckZeroSizeCollection.cs
```diff
@@ -44,28 +45,34 @@ public void TestCountMethod()
...
             result = (someEnumerable.Count()) >= (0); // Noncompliant
             result = ((((someEnumerable).Count())) >= ((0))); // Noncompliant
-//                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+            //        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
             result = 0 <= someEnumerable.Count(); // Noncompliant
-//                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+            //       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+            
+            result = result.ToString() == "Not integer";            
+            result = (localVariable = 42) != -1;
```
pavel-mikula-sonarsource, line 73: Whitespace cleanup
```suggestion

            result = result.ToString() == "Not integer";
            result = (localVariable = 42) != -1;
```

pavel-mikula-sonarsource:
Burgr is probably not happy when I add last commit. Can you please push empty commit here?

analyzers/tests/SonarAnalyzer.UnitTest/Facade/SyntaxFacadeTest.cs
```diff
@@ -82,12 +82,12 @@ public class SyntaxFacadeTest
             vb.NodeIdentifier(null).Should().BeNull();
 
         [TestMethod]
-        public void NodeIdentifier_UnexpectedTypeThrows_CS() =>
-            cs.Invoking(x => x.NodeIdentifier(CS.SyntaxFactory.ThrowStatement())).Should().Throw<InvalidOperationException>();
+        public void NodeIdentifier_Unexpected_Returns_Null_CS() =>
```
pavel-mikula-sonarsource, line 85: Also `InvocationIdentifier_UnexpectedTypeThrows_CS` and `InvocationIdentifier_UnexpectedTypeThrows_VB` needs to be updated after the change of exception type

analyzers/src/SonarAnalyzer.Common/Rules/DoNotCheckZeroSizeCollectionBase.cs
```diff
@@ -19,124 +19,123 @@
...
-    public abstract class DoNotCheckZeroSizeCollectionBase<TSyntaxKind, TBinaryExpressionSyntax, TExpressionSyntax> : SonarDiagnosticAnalyzer
+    public abstract class DoNotCheckZeroSizeCollectionBase<TSyntaxKind> : SonarDiagnosticAnalyzer
         where TSyntaxKind : struct
-        where TBinaryExpressionSyntax : SyntaxNode
-        where TExpressionSyntax : SyntaxNode
     {
         protected const string DiagnosticId = "S3981";
-        private const string MessageFormat = "The '{0}' of '{1}' is always '>=0', so fix this test to get the real expected behavior.";
-
+        private const string MessageFormat = "The '{0}' of '{1}' always evaluates as '{2}' regardless the size.";
```
pavel-mikula-sonarsource, line 35: Please update `\analyzers\its\expected\Net5\Net5--net5.0-S3981.json` file with the new message to fix ITs from
```
"message":  "The 'Count' of 'IEnumerable<T>' is always '>=0', so fix this test to get the real expected 
```
to
```
"message":  "The 'Count' of 'IEnumerable<T>' always evaluates as 'True' regardless the size.",
```

analyzers/its/expected/Net5/Net5--net5.0-S3981.json
```diff
@@ -2,7 +2,7 @@
 "issues":  [
 {
 "id":  "S3981",
-"message":  "The 'Count' of 'IEnumerable<T>' is always '>=0', so fix this test to get the real expected behavior.",
+"message":  "The 'Count' of 'IEnumerable<T>' is always 'True', so fix this test to get the real expected behavior.",
```
pavel-mikula-sonarsource, line 5: Like this: (I won't commit this myself as that would make Burgr unhappy)
```suggestion
"message":  "The 'Count' of 'IEnumerable<T>' always evaluates as 'True' regardless the size.",
```

pavel-mikula-sonarsource (approved):
LGTM

PR #4809, opened 2021-08-19
Disable opencover safe mode to reduce building time

costin-zaharia-sonarsource:
From original doc: https://github.com/OpenCover/opencover/wiki/Usage

```
-safemode:[on|off|yes|no] - enable or disable safemode - default is on/yes. When safemode is on OpenCover will use a common buffer for all threads in an instrumented process and this may have performance impacts depending on your code and its tests. Turning safemode off uses individual buffers for each thread but this may lead to data loss (uncovered code reported) if the runtime shuts down the instrumented process before the profiler has had time to retrieve the data and shunt it to the host for processing.
```

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -155,7 +155,7 @@ stages:
           CreateFolder("coverage")
           CreateFolder("TestResults")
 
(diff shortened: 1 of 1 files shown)


Corniel:
What is the effect of this?

costin-zaharia-sonarsource:
> What is the effect of this?

As the title says, this is meant to reduce our build pipeline running time and it seems to be significant.

Before:
image

After:
image

dmitry-golovinov-sonarsource (approved):
LGTM!

Corniel:
> 
> 
> > What is the effect of this?
> 
> As the title says, this is meant to reduce our build pipeline running time and it seems to be significant.
> 
> Before:
> image
> 
> After:
> image

I know, I was interested in the numbers. Nice improvement!

costin-zaharia-sonarsource:
There is one more thing we have to handle. When we manually execute the code coverage on the VMs they take 6 minutes if the NuGet packages are not downloaded, and ~2.5m if they are.

When the coverage is started using the Azure agent we get the above times. There is a clear potential for improvement there.

PR #4812, opened 2021-08-20
Run tests in parallel during build to reduce the execution time

costin-zaharia-sonarsource:


costin-zaharia-sonarsource:
I'm closing this since it does not reduce the running time.

PR #4814, opened 2021-08-22
Rule S4581: Guid.Empy is preferred (both C#,  VB.NET and fix provider)

### LINKED ISSUE #5245: Fix S4581 FP: Cannot give method parameter a default value of an empty Guid.
### Description

It is not possible to provide an empty Guid as a default value for a parameter in a method.

### Repro steps

```
private void Test(Guid guid = default)
{
    Console.WriteLine(guid.ToString("N"));
}
```

### Expected behavior

We should be able to use ```default``` or  ```new Guid()``` for the default value.

### Actual behavior

As shown in the example above ```guid = default``` breaks the S4581 rule that states that Guid.NewGuid() or Guid.Empty should be used. In this case it is not possible to use Guid.Empty because it is a static readonly value.

### Known workarounds

Make the Guid parameter nullable e.g. ```Guid? guid = null``` and then check for null value within the method body and replace it with Guid.Empty.

### Related information

* C#/VB.NET Plugins version 8.33.0.40503
* Visual Studio version 17.0.4
* MSBuild / dotnet version 5
* SonarScanner for .NET version (if used) N/A
* Operating System Win-10-x64 10.0.19044

### LINKED ISSUE #5615: Rule S4581: Guid.Empy is preferred for VB.NET
Implement S4581 for VB.NET

### LINKED ISSUE #5616: Rule S4581: Add C# code fix
Add C# code fix for S4581

Before
```
var value = new Guid();
```
After
```
var value = Guid.Empty; // Fixed
```

Corniel:
Given the insights we got from #4265, I decided to start a new pull request.

As rule S4581 is not longer about discouraging `new Guid()`, but also about not using `default` (and `default(Guid)`) for assigning the `Guid.Empty` value, the rule (class name) has been renamed: `GuidEmptyIsPreferred`.

The fix provider (only for C# yet) provides as `Guid.Empty` replacement option.

Fixes #5615
Fixes #5616
Fixes #5245


Corniel:
@pavel-mikula-sonarsource Still busy?

Corniel:
@pavel-mikula-sonarsource 
I have some FN's that are - I think - worth discussing:
1. Guid.Parse("00000000-0000-0000-0000-000000000000")
2. Unassigned fields
3. `Dim unasignend As Guid` (VB.NET only)

Do we want to covert these, and if so, do we want to cover them now?

Corniel:
@pavel-mikula-sonarsource BUMP

pavel-mikula-sonarsource:
Hi @Corniel, I'm sorry for my late reply on this one. I was quite busy recently. I'll try to prioritize it now.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.cs
```diff
@@ -0,0 +1,87 @@
...
+        //         ^^^^^^^^^^
+    }
+
+    void DefaultInintiation()
+    {
+        Guid defaultValue = default; // Noncompliant
+        var defaultOfGuid = default(Guid); // Noncompliant
+        Property = default; // Noncompliant
+        var instance = new GuidClass(default); // Noncompliant
+        instance.Method(default); // Noncompliant
```
pavel-mikula-sonarsource, line 62: Can you please rename this variable? It took me a while to understand why it is noncompliant.
```suggestion
        var guidAsParameter = new GuidClass(default); // Noncompliant
        guidAsParameter.Method(default); // Noncompliant
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        int integer = default;
+        var date = default(DateTime);
+    }
+}
+
+class NonCompliant
+{
+    Guid Field; // FN
+    readonly Guid ReadonlyField; // FN
```
pavel-mikula-sonarsource, line 47: Nitpick
```suggestion
    Guid field; // FN
    readonly Guid readonlyField; // FN
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.cs
```diff
@@ -0,0 +1,87 @@
...
+        var bytes = new Guid(new byte[0]); // Compliant
+        var str = new Guid("FA97FFE7-532C-4015-8698-49D8CE4126F4"); // Compliant
+    }
+
+    void NullableDefault()
+    {
+        Guid? nullable = default; // Compliant, not equivalent to Guid.Empty.
+        var nullableOfT = default(Guid?); // Compliant
+        var instance = new NullableGuidClass(default); // Compliant
+        instance.Method(default); // Compliant
```
pavel-mikula-sonarsource, line 26: In general, it makes the comments easier to read when they are (partially) aligned with surrounding rows.
```suggestion
        Guid? nullable = default;                                    // Compliant, not equivalent to Guid.Empty.
        var nullableOfT = default(Guid?);                       // Compliant
        var instance = new NullableGuidClass(default); // Compliant
        instance.Method(default);                                   // Compliant
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.cs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    void NotInitiated(string str)
```
pavel-mikula-sonarsource, line 29:
```suggestion
    void NotInitialized(string str)
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.cs
```diff
@@ -0,0 +1,87 @@
+﻿using System;
+
+class Compliant
+{
+    Guid? NullableField; // Compliant
+    readonly Guid? ReadonlyNullableField; // Compliant
```
pavel-mikula-sonarsource, line 6: Nitpicks
```suggestion
    Guid? nullableField; // Compliant
    readonly Guid? readonlyNullableField; // Compliant
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.cs
```diff
@@ -0,0 +1,87 @@
...
+        var rnd = Guid.NewGuid();// Compliant
+        var bytes = new Guid(new byte[0]); // Compliant
+        var str = new Guid("FA97FFE7-532C-4015-8698-49D8CE4126F4"); // Compliant
+    }
+
+    void NullableDefault()
+    {
+        Guid? nullable = default; // Compliant, not equivalent to Guid.Empty.
+        var nullableOfT = default(Guid?); // Compliant
+        var instance = new NullableGuidClass(default); // Compliant
```
pavel-mikula-sonarsource, line 25: Same as another comment below. It took me a while to understand the purpose.
```suggestion
        var guidAsParameter = new NullableGuidClass(default); // Compliant
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.vb
```diff
@@ -0,0 +1,87 @@
...
+    End Sub
+End Class
+
+Class NullableGuidClass
+    Public Sub New(param As Guid?)
+    End Sub
+
+    Public Sub Method(param As Guid?)
+    End Sub
+End Class
```
pavel-mikula-sonarsource, line 87: Just some extra empty lines before/after class/struct.
```suggestion
Structure GuidAssignmentStruct

    Private Shared ReadOnly StaticField As Guid ' FP
    Private Field As Guid ' Compliant, structs Do Not allow assigned instance values
    Private ReadOnly ReadOnlyField As Guid ' Compliant, structs Do Not allow assigned instance values

End Structure

Class GuidClass

    Public Sub New(param As Guid)
    End Sub

    Public Sub Method(param As Guid)
    End Sub

End Class

Class NullableGuidClass

    Public Sub New(param As Guid?)
    End Sub

    Public Sub Method(param As Guid?)
    End Sub

End Class
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.vb
```diff
@@ -0,0 +1,87 @@
+﻿Imports System
+
+Class Compliant
```
pavel-mikula-sonarsource, line 3: In general, same comments apply here as the ones in the C# variant. Except the lower-case fields :)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.vb
```diff
@@ -0,0 +1,87 @@
+﻿Imports System
```
pavel-mikula-sonarsource, line 1: Not needed in VB codebase
```suggestion
```

analyzers/src/SonarAnalyzer.CSharp/Rules/PreferGuidEmptyCodeFix.cs
```diff
@@ -0,0 +1,65 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Formatting;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public sealed class PreferGuidEmptyCodeFix : SonarCodeFixProvider
+    {
+        internal const string Title = "Use Guid.Empty instead";
```
pavel-mikula-sonarsource, line 37: In this case, we can also register the other codefix to `Use Guid.NewGuid() instead", right?

analyzers/src/SonarAnalyzer.CSharp/Rules/PreferGuidEmptyCodeFix.cs
```diff
@@ -0,0 +1,65 @@
...
+        protected override Task RegisterCodeFixesAsync(SyntaxNode root, CodeFixContext context)
+        {
+            var toReplace = ToReplace(root, context);
+            var replacement = Replacement(root, toReplace);
+
+            context.RegisterCodeFix(
+               CodeAction.Create(
+               Title,
+               token => Task.FromResult(context.Document.WithSyntaxRoot(replacement))),
+               context.Diagnostics);
```
pavel-mikula-sonarsource, line 50: Can be a single line. And then we should not need the empty line before.
```suggestion
            context.RegisterCodeFix(
               CodeAction.Create(
               Title,
               token => Task.FromResult(context.Document.WithSyntaxRoot(replacement))),
               context.Diagnostics);
```

analyzers/src/SonarAnalyzer.Common/Rules/PreferGuidEmptyBase.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        private bool NoParameter(SyntaxNode defaultExpression) =>
+            !defaultExpression.Ancestors()
+                .Any(node => Language.Syntax.IsKind(node, Language.SyntaxKind.Parameter));
+
+        private static bool CreatesGuidEmpty(SyntaxNode[] arguments, SemanticModel semanticModel) =>
+            arguments.Length == 1
+            && semanticModel.GetConstantValue(arguments[0]) is { HasValue: true } optional
+            && optional.Value is string str
+            && Guid.TryParse(str, out var guid)
```
pavel-mikula-sonarsource, line 79: Can be simplified, as `Optional.Value` will return just `null` if `HasValue` is `false` and `TryParse` can accept `null` as input.
```suggestion
            && Guid.TryParse(semanticModel.GetConstantValue(arguments[0]).Value as string, out var guid)
```

analyzers/src/SonarAnalyzer.Common/Rules/PreferGuidEmptyBase.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            var arguments = Language.Syntax.ArgumentExpressions(ctorNode).ToArray();
+            return arguments.Length == 0 || CreatesGuidEmpty(arguments, semanticModel);
+        }
+
+        private bool NoParameter(SyntaxNode defaultExpression) =>
+            !defaultExpression.Ancestors()
+                .Any(node => Language.Syntax.IsKind(node, Language.SyntaxKind.Parameter));
+
+        private static bool CreatesGuidEmpty(SyntaxNode[] arguments, SemanticModel semanticModel) =>
```
pavel-mikula-sonarsource, line 75:
```suggestion
        private static bool CreatesEmptyGuid(SyntaxNode[] arguments, SemanticModel semanticModel) =>
```

analyzers/src/SonarAnalyzer.Common/Rules/PreferGuidEmptyBase.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        private bool NotAllowedGuidCtorArguments(SyntaxNode ctorNode, SemanticModel semanticModel)
+        {
+            var arguments = Language.Syntax.ArgumentExpressions(ctorNode).ToArray();
+            return arguments.Length == 0 || CreatesGuidEmpty(arguments, semanticModel);
+        }
+
+        private bool NoParameter(SyntaxNode defaultExpression) =>
+            !defaultExpression.Ancestors()
+                .Any(node => Language.Syntax.IsKind(node, Language.SyntaxKind.Parameter));
```
pavel-mikula-sonarsource, line 73: Just formatting (and rename)
```suggestion
            !defaultExpression.Ancestors().Any(x => Language.Syntax.IsKind(x, Language.SyntaxKind.Parameter));
```

analyzers/src/SonarAnalyzer.Common/Rules/PreferGuidEmptyBase.cs
```diff
@@ -0,0 +1,82 @@
...
+                Language.SyntaxKind.DefaultExpressions);
+        }
+
+        private bool NotAllowedGuidCtorArguments(SyntaxNode ctorNode, SemanticModel semanticModel)
+        {
+            var arguments = Language.Syntax.ArgumentExpressions(ctorNode).ToArray();
+            return arguments.Length == 0 || CreatesGuidEmpty(arguments, semanticModel);
+        }
+
+        private bool NoParameter(SyntaxNode defaultExpression) =>
```
pavel-mikula-sonarsource, line 71: The `NoParameter` took me some time to understand. Let's invert the logic to have it like this. It also better explains the implementation IMO.
```suggestion
        private bool IsInParameter(SyntaxNode defaultExpression) =>
```

pavel-mikula-sonarsource (changes requested):
Hi @Corniel, thank you for taking a look into this. I left a bunch of minor comments here.

analyzers/src/SonarAnalyzer.CSharp/Rules/PreferGuidEmpty.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class PreferGuidEmpty : PreferGuidEmptyBase<SyntaxKind>
+    {
+        protected override ILanguageFacade<SyntaxKind> Language => CSharpFacade.Instance;
+    }
```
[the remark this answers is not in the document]
Corniel: I know, and that is what I (always) aim for. :)

analyzers/src/SonarAnalyzer.Common/Rules/PreferGuidEmptyBase.cs
pavel-mikula-sonarsource: I've unresolved this one. It will also need to remove the `!` from the condition. And the renamings from the last commit will break he build, as it needs to be updated on the invocation side too

analyzers/rspec/cs/S4581_c#.html
```diff
@@ -9,11 +9,9 @@ <h2>Noncompliant Code Example</h2>
...
 public void Foo()
 {
-    var g1 = new Guid(); // Noncompliant - what's the intent?
-    Guid g2 = new(); // Noncompliant
-
-    var g3 = default(Guid); // Noncompliant
-    Guid g4 = default; // Noncompliant
+    var g = new Guid(); // Noncompliant - what's the intent?
+    var guid = default(Guid);
+    Guid other = default;
```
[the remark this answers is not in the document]
Corniel: I'm not aware. I'll roll it back.

analyzers/src/SonarAnalyzer.CSharp/Rules/PreferGuidEmptyCodeFix.cs
Corniel: What do you mean? That we offer both options? That might be a nice feature. Does the Verifier support that?

analyzers/src/SonarAnalyzer.Common/Rules/PreferGuidEmptyBase.cs
Corniel: That have been done. Tests succeed again. :)

Corniel:
I think all remarks haven been covered. Outstanding questions:
1. What to do with the FN's / extra cases?
2. Should the code fix provide also `Guid.NewGuid()` for `new Guid()`, and/or other cases?

analyzers/src/SonarAnalyzer.Common/Helpers/TypeHelper.cs
```diff
@@ -28,47 +28,27 @@ internal static class TypeHelper
     {
         #region TypeKind
 
-        public static bool IsInterface(this ITypeSymbol self)
-        {
-            return self != null && self.TypeKind == TypeKind.Interface;
-        }
+        public static bool IsInterface(this ITypeSymbol self) => self is { TypeKind: TypeKind.Interface };
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: This one is missing

analyzers/src/SonarAnalyzer.Common/Facade/ISyntaxKindFacade.cs
```diff
@@ -34,9 +34,11 @@ public interface ISyntaxKindFacade<out TSyntaxKind>
         abstract TSyntaxKind InterpolatedStringExpression { get; }
         abstract TSyntaxKind[] MethodDeclarations { get; }
         abstract TSyntaxKind[] ObjectCreationExpressions { get; }
+        abstract TSyntaxKind Parameter { get; }
         abstract TSyntaxKind ReturnStatement { get; }
         abstract TSyntaxKind SimpleMemberAccessExpression { get; }
         abstract TSyntaxKind StringLiteralExpression { get; }
         abstract TSyntaxKind[] TypeDeclaration { get; }
+        abstract TSyntaxKind[] DefaultExpressions{ get; }
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: The 2nd message about sorting was not addressed.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/PreferGuidEmptyTest.cs
```diff
@@ -0,0 +1,58 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarAnalyzer.UnitTest.TestFramework;
+using CS = SonarAnalyzer.Rules.CSharp;
+using VB = SonarAnalyzer.Rules.VisualBasic;
+
+namespace SonarAnalyzer.UnitTest.Rules
+{
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: The missing unit test was not addressed. We need to rename the test case file and add the test back.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.cs
```diff
@@ -0,0 +1,87 @@
...
+    readonly Guid ReadonlyField; // FN
+    Guid Property { get; set; }
+
+    void DefaultCtor()
+    {
+        var ctor = new Guid(); // Noncompliant {{Use 'Guid.NewGuid()' or 'Guid.Empty' or add arguments to this GUID instantiation.}}
+        //         ^^^^^^^^^^
+    }
+
+    void DefaultInintiation()
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: This one is missing

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.cs
pavel-mikula-sonarsource: This one was not addressed, it was done on other places. Please try to find all instances of the GuidClass and use it there. I saw 4 `instance` variables that could be renamed

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        Guid defaultValue = default; // Noncompliant
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: This one is missing

analyzers/src/SonarAnalyzer.CSharp/Rules/PreferGuidEmptyCodeFix.cs
pavel-mikula-sonarsource: Yes, exactly. I believe `VerifierBuilder` does support that. Calling this with a code fix title should do the trick:
https://github.com/SonarSource/sonar-dotnet/blob/9f463a5710a2a70cb12b20dbbe23fe894bd780ae/analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/VerifierBuilder.cs#L97-L98

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.Fixed.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        int integer = default;
+        var date = default(DateTime);
+    }
+}
+
+class NonCompliant
+{
+    Guid field; // FN
+    readonly Guid readonlyField; // FN
```
pavel-mikula-sonarsource, line 47: I would not mark these as FNs. There's a high chance that they are initialized somewhere.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.Fixed.cs
```diff
@@ -0,0 +1,86 @@
...
+        var defaultOfGuid = Guid.Empty; // Fixed
+        Property = Guid.Empty; // Fixed
+        var instance = new GuidClass(Guid.Empty); // Fixed
+        instance.Method(Guid.Empty); // Fixed
+    }
+
+    void EmptyString()
+    {
+        var emptyCtor = Guid.Empty; // Fixed
+        var emptyParse = Guid.Parse("00000000-0000-0000-0000-000000000000"); // FN
```
pavel-mikula-sonarsource, line 67: We can accept this as FN for now and leave it like this. There should not be many people using it this way 😄

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.Fixed.cs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+struct GuidAssignmentStruct
+{
+    static readonly Guid Static; // FN
```
pavel-mikula-sonarsource, line 73: I would not mark it as FN
```suggestion
    static readonly Guid Static; 
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.vb
```diff
@@ -0,0 +1,93 @@
...
+        Dim int As Integer = Nothing ' Compliant
+        Dim dat As Date = Nothing ' Compliant
+    End Sub
+
+End Class
+
+Class NonCompliant
+
+    Private Field As Guid ' FN
+    Private ReadOnly ReadonlyField As Guid ' FN
```
pavel-mikula-sonarsource, line 43:
```suggestion
    Private Field As Guid
    Private ReadOnly ReadonlyField As Guid
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.vb
```diff
@@ -0,0 +1,93 @@
...
+    Sub EmptyString()
+        Dim ctor As Guid = New Guid("00000000-0000-0000-0000-000000000000") ' Noncompliant
+        Dim parse As Guid = Guid.Parse("00000000-0000-0000-0000-000000000000") ' FN
+    End Sub
+
+End Class
+
+Structure GuidAssignmentStruct
+
+    Private Shared ReadOnly StaticField As Guid ' FP
```
pavel-mikula-sonarsource, line 69:
```suggestion
    Private Shared ReadOnly StaticField As Guid
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.vb
```diff
@@ -0,0 +1,93 @@
...
+    Private Field As Guid ' Compliant, structs Do Not allow assigned instance values
+    Private ReadOnly ReadOnlyField As Guid ' Compliant, structs Do Not allow assigned instance values
+
```
pavel-mikula-sonarsource, line 72: For readability (it looks weird on GitHub UI, just align the comments the same way in VS)
```suggestion
    Private Field As Guid                                 ' Compliant, structs Do Not allow assigned instance values
    Private ReadOnly ReadOnlyField As Guid ' Compliant, structs Do Not allow assigned instance values

```

pavel-mikula-sonarsource (commented):
@Corniel, I tried to resolve existing conversations. There are some unresolved. Please do not resolve conversations, add 👍 instead as a marker that you've addressed them. I'll resolve it.

The other codefix would be nice. I would actually consider doing it in another PR, as this one is big already. 
I would also be fine with removing codefixes from this PR and add both of them in another one to keep things smaller.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.Fixed.cs
Corniel: Well, that was my question (earlier): do we consider this a place where `Guid.Empty` is preferred. I would argue it does, but is a totally different approach to detect it. If we decide that we do not want to report on this, that's fine too.

pavel-mikula-sonarsource: I'm answering that question here :) I'm afraid we would have way too many FPs if we'd try to detect this scenario

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.Fixed.cs
Corniel: (In)famous last words. ;)

pavel-mikula-sonarsource:
Hi @Corniel, there are still 11 unresolved conversations above. GitHub UI is inconvinitenly hiding them under the `23 hidden conversations Load more…` button and few more are then hidden inside under `3 hidden conversations Load more…` button.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/PreferGuidEmptyTest.cs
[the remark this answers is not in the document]
Corniel: Found it. `new()` argument detection needed a Wrapper. I'm not 100% happy with the syntax, so feedback desired.

Corniel:
I propose to keep the single fix, and discuss/pickup the `Guid.NewGuid()` alternative in a separate PR. Both because it will be the first rule to have two suggestions (potentially not for the `default` assignment), and because we could add a VB Fix as well.

pavel-mikula-sonarsource:
> I propose to keep the single fix, and discuss/pickup the `Guid.NewGuid()` alternative in a separate PR. Both because it will be the first rule to have two suggestions (potentially not for the `default` assignment), and because we could add a VB Fix as well.

Sound good

analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
```diff
@@ -53,6 +54,16 @@ internal sealed class CSharpSyntaxFacade : SyntaxFacade<SyntaxKind>
...
         public override bool IsNullLiteral(SyntaxNode node) => node.IsNullLiteral();
 
+        public override IEnumerable<SyntaxNode> ArgumentExpressions(SyntaxNode node) =>
+            node switch
+            {
+                ObjectCreationExpressionSyntax objectCreation => objectCreation.ArgumentList?.Arguments.Select(x => x.Expression),
+                null => Enumerable.Empty<SyntaxNode>(),
+                _ => ImplicitObjectCreationExpressionSyntaxWrapper.IsInstance(node)
+                ? Enumerable.Empty<SyntaxNode>()
+                : throw InvalidOperation(node, nameof(ArgumentExpressions))
```
pavel-mikula-sonarsource, line 64: We can do it like this: https://github.com/SonarSource/sonar-dotnet/blob/4694178cfc3a796c7bb64e406cf3ce3a9f462fe2/analyzers/src/SonarAnalyzer.CFG/Extensions/IOperationExtensions.cs#L110-L117
```suggestion
                var _ when ImplicitObjectCreationExpressionSyntaxWrapper.IsInstance(node) => Enumerable.Empty<SyntaxNode>(),
                _ => throw InvalidOperation(node, nameof(ArgumentExpressions))
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.CSharp9.cs
```diff
@@ -16,11 +16,6 @@ public void Foo()
             string test = default;
         }
 
-        // See: https://github.com/SonarSource/sonar-dotnet/issues/5245
-        private void Test(Guid guid = default) // Noncompliant FP - Guid.Empty cannot be used in this case
-        {
-        }
-
```
pavel-mikula-sonarsource, line 23: Why was this removed? I expected it to be fixed by this PR, or left as FP (and we'll update the description that this one was not fixed)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.vb
```diff
@@ -0,0 +1,94 @@
...
+    Sub NotEmpty()
+        Dim rnd As Guid = Guid.NewGuid() ' Compliant
+        Dim bytes As Guid = New Guid({0, 3, 4}) ' Compliant
+        Dim str As Guid = New Guid("FA97FFE7-532C-4015-8698-49D8CE4126F4") ' Compliant
+    End Sub
+
+    Sub NullableDefault()
+        Dim nullable As Guid? = Nothing ' Compliant, not equivalent to Guid.Empty.
+        Dim instance As New NullableGuidClass(Nothing) ' Compliant
+        instance.Method(Nothing) ' Compliant
```
pavel-mikula-sonarsource, line 22:
```suggestion
        Dim guidAsParameter As New NullableGuidClass(Nothing) ' Compliant
        guidAsParameter.Method(Nothing) ' Compliant
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.vb
```diff
@@ -0,0 +1,94 @@
...
+    End Sub
+
+    Sub NotInitiated(str As String)
```
pavel-mikula-sonarsource, line 25:
```suggestion
    Sub NotInitialized(str As String)
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.vb
```diff
@@ -0,0 +1,94 @@
...
+    Private ReadOnly ReadonlyField As Guid
+    Property Prop As Guid
+
+    Sub DefaultCtor()
+        Dim ctor As Guid = New Guid() ' Noncompliant {{Use 'Guid.NewGuid()' or 'Guid.Empty' or add arguments to this GUID instantiation.}}
+        '                  ^^^^^^^^^^
+        Dim asignend As New Guid() ' Noncompliant
+    End Sub
+
+    Sub DefaultInintiation()
```
pavel-mikula-sonarsource, line 52:
```suggestion
    Sub DefaultInitialization()
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.vb
```diff
@@ -0,0 +1,94 @@
...
+        Dim defaultValue As Guid = Nothing ' Noncompliant
+        Dim unasignend As Guid ' FN
+        Dim instance As New GuidClass(Nothing) ' Noncompliant
+        instance.Method(Nothing) ' Noncompliant
```
pavel-mikula-sonarsource, line 56:
```suggestion
        Dim guidAsParameter As New GuidClass(Nothing) ' Noncompliant
        guidAsParameter.Method(Nothing) ' Noncompliant
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.vb
```diff
@@ -0,0 +1,94 @@
...
+        Prop = Nothing ' Noncompliant
+        Field = Nothing ' Noncompliant
+    End Sub
+
+    Sub EmptyString()
+        Dim ctor As Guid = New Guid("00000000-0000-0000-0000-000000000000") ' Noncompliant
```
pavel-mikula-sonarsource, line 62:
```suggestion
        Dim ctor1 As Guid = New Guid("00000000-0000-0000-0000-000000000000") ' Noncompliant
        Dim ctor2 As New Guid("00000000-0000-0000-0000-000000000000") ' Noncompliant
```

pavel-mikula-sonarsource (changes requested):

pavel-mikula-sonarsource:
`RulesDoNotThrow_CS` and `RulesDoNotThrow_VB` throw exception:
```
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PreferGuidEmpty.CSharp9.cs
Corniel: It is on the not specific C# 9 version (and supported now). Felt an obsolete extra test.

Corniel:
> `RulesDoNotThrow_CS` and `RulesDoNotThrow_VB` throw exception:
> 
> ```
> System.ArgumentNullException: Value cannot be null. (Parameter 'source')
>    at System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
>    at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
>    at SonarAnalyzer.Rules.PreferGuidEmptyBase`1.IsInvalidCtor(SyntaxNode ctorNode, SemanticModel semanticModel) 
> ```
How did you achieve this? Locally, these tests pass, both for .NET 6 as .NET 4.8 (classic).

Corniel:
> `RulesDoNotThrow_CS` and `RulesDoNotThrow_VB` throw exception:
> 
> ```
> System.ArgumentNullException: Value cannot be null. (Parameter 'source')
>    at System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
>    at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
>    at SonarAnalyzer.Rules.PreferGuidEmptyBase`1.IsInvalidCtor(SyntaxNode ctorNode, SemanticModel semanticModel) 
> ```
How did you achieve this? Locally, these tests pass, both for .NET 6 as .NET 4.8 (classic).

analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
```diff
@@ -53,6 +54,18 @@ internal sealed class CSharpSyntaxFacade : SyntaxFacade<SyntaxKind>
...
+            node switch
+            {
+                ObjectCreationExpressionSyntax objectCreation => ArgumentExpressions(objectCreation),
+                null => Enumerable.Empty<SyntaxNode>(),
+                var _ when ImplicitObjectCreationExpressionSyntaxWrapper.IsInstance(node) => Enumerable.Empty<SyntaxNode>(),
+                _ => throw InvalidOperation(node, nameof(ArgumentExpressions))
+            };
+
+        private static IEnumerable<SyntaxNode> ArgumentExpressions(ObjectCreationExpressionSyntax node) =>
+            node.ArgumentList?.Arguments.Select(x => x.Expression) ?? Enumerable.Empty<SyntaxNode>();
```
pavel-mikula-sonarsource, line 67: Let's embed it. It's more clear that this thing is not meant to return null.
```suggestion
                ObjectCreationExpressionSyntax objectCreation => objectCreation.ArgumentList?.Arguments.Select(x => x.Expression) ?? Enumerable.Empty<SyntaxNode>(),
                null => Enumerable.Empty<SyntaxNode>(),
                var _ when ImplicitObjectCreationExpressionSyntaxWrapper.IsInstance(node) => Enumerable.Empty<SyntaxNode>(),
                _ => throw InvalidOperation(node, nameof(ArgumentExpressions))
            };
```

pavel-mikula-sonarsource:
UT failed `RulesDoNotThrow_VB` failed. Please take a look

pavel-mikula-sonarsource:
ITs also fail for `ember-mm` with NRE. You should be able to see where exactly (source file, registered node) from detail logs. 
Rebuild project locally.
Go to `analyzers/its` directory
Run `.\regression-test.ps1 -ruleId S4581 -project Ember-MM`
Full log should be in `analyzers\its\output\Ember-MM.log`

analyzers/tests/SonarAnalyzer.UnitTest/Rules/PreferGuidEmptyTest.cs
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarAnalyzer.UnitTest.TestFramework;
+using CS = SonarAnalyzer.Rules.CSharp;
+using VB = SonarAnalyzer.Rules.VisualBasic;
```
pavel-mikula-sonarsource, line 24: Attempt to make the rebase conflict fixable
```suggestion
using CS = SonarAnalyzer.Rules.CSharp;
using VB = SonarAnalyzer.Rules.VisualBasic;
```

pavel-mikula-sonarsource (approved):
LGTM 🎉

pavel-mikula-sonarsource:
@Corniel Thank you for taking care of this!

May I ask you for a little cleanup in your fork? Can you delete all old remote branches except yours? It will make my navigation easier. Most of them are just old snapshot. Thank you

Corniel:
@pavel-mikula-sonarsource As far as I'm aware, I've no outstanding branches that should mess up with yours.

pavel-mikula-sonarsource:
Your branches are fine. It's our historical branches that can help me in navigation if you'd delete them :)

image

Corniel:
Done. I'm not even sure why they where there in the first place. Most of them I never touched. Might be left overs from forward integrations.

pavel-mikula-sonarsource:
Thank you! It was just a useless snapshot of all branches from the day you've forked the repo.

PR #4818, opened 2021-08-23
Add warning for lambda discard params

costin-zaharia-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantDeclaration.ArraySize.Fixed.cs
```diff
@@ -7,16 +7,16 @@ abstract class RedundantDeclaration
     {
         public RedundantDeclaration()
         {
-            MyEvent += new EventHandler((a, b) => { }); // Fixed
-            MyEvent += (a, b) => { };
+            MyEvent += new EventHandler(Handle); // Fixed
+            MyEvent += (a, b) => { Handle(a, b); };
```
costin-zaharia-sonarsource, line 11: I've changed the test cases to use the parameters from declarations to avoid different results when running for C# 9 language vers.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantDeclaration.CSharp9.Fixed.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+
+Record r1 = new Record(); // Compliant - FN
```
andrei-epure-sonarsource, line 3: [education] why is this an FN?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantDeclaration.ArraySize.Fixed.cs
andrei-epure-sonarsource: do we have the same test case (unused lambda params) to run for both C# 8 and C# 9? 

It feels it's valuable to have the same test case running on two different language versions, to verify for regressions and make the behavior explicit from the tests.

At least a small, minimal example which gets analyzed and fixed differently with C# 8 / C# 9.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/RedundantDeclaration.CSharp9.cs
```diff
@@ -29,8 +38,10 @@ public class RedundantDeclaration
 {
     public RedundantDeclaration()
     {
-        MyEvent += new((a, b) => { }); // Noncompliant {{Remove the explicit delegate creation; it is redundant.}}
-//                 ^^^^^^^^^^^^^^^^^^
+        MyEvent += new((a, b) => { });
```
andrei-epure-sonarsource, line 41: could you also add a test with `(unused1, used, unused2) => { Foo(used);}` to cover this mixed case?

andrei-epure-sonarsource (changes requested):
Looks good, I suggest to improve a bit the test cases

PR #4820, opened 2021-08-23
Move Old CFG to .Sonar directory and namespace

pavel-mikula-sonarsource:
There will be 
```
namespace SonarAnalyzer.CFG.Sonar
namespace SonarAnalyzer.CFG.Roslyn
```

We'll need to update namespace usages on security side once we update SonarAnalyzer.CFG to this version.


pavel-mikula-sonarsource:
I'm not going to address the coverage nor codesmells in this PR, as I touched only namespaces around them.

dmitry-golovinov-sonarsource (approved):
LGTM, but there are a number of very obvious codesmells that could be fixed, since we see them now in the QG.

PR #4821, opened 2021-08-23
Introduce Roslyn CFG wrapper

pavel-mikula-sonarsource:
This is a draft, for now, more commits with more wrappers will arrive.

### DIFF
--- a/analyzers/src/SonarAnalyzer.CFG/Roslyn/BasicBlock.cs
+++ b/analyzers/src/SonarAnalyzer.CFG/Roslyn/BasicBlock.cs
@@ -1,0 +1,43 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2021 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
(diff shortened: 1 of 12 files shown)


analyzers/src/SonarAnalyzer.CFG/Roslyn/BasicBlock.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public class BasicBlock
+    {
+        private static readonly PropertyInfo OperationsProperty;
+
+        private readonly Lazy<ImmutableArray<IOperation>> operations;
+
+        public ImmutableArray<IOperation> Operations => operations.Value;
+
+        static BasicBlock()
```
dmitry-golovinov-sonarsource, line 37: Do we really need it? Can't it be replaced with  _private static readonly PropertyInfo OperationsProperty = RoslynHelper.FlowAnalysisType("BasicBlock")?.GetProperty(nameof(Operations));_

dmitry-golovinov-sonarsource (approved):
LGTM with respect of fixing UTs later. Just got one minor comment.

analyzers/src/SonarAnalyzer.CFG/Roslyn/BasicBlock.cs
pavel-mikula-sonarsource: There will be 9 more properties coming in the next PR. For now, this class is just a dummy place holder to make UTs work.

dmitry-golovinov-sonarsource: Got it

PR #4822, opened 2021-08-23
Remove TaskHelper

costin-zaharia-sonarsource:
- `TaskHelper` is not needed since we can use the implementation from the framework `Task.CompletedTask`.


pavel-mikula-sonarsource (approved):
LGTM

PR #4826, opened 2021-08-24
S4830 - add C#8 static local function test

andrei-epure-sonarsource:
(there is already support for C# 9 static lambdas, see [line 21](https://github.com/SonarSource/sonar-dotnet/blob/8.27.0.35380/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CertificateValidationCheck.CSharp9.cs#L21)


costin-zaharia-sonarsource (approved):
LGTM!

PR #4827, opened 2021-08-24
S4000 function pointers support

csaba-sagi-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PointersShouldBePrivate.CSharp9.cs
```diff
@@ -1,29 +1,31 @@
...
+
+public unsafe class Class
+{
+    private IntPtr p1;
+    protected int* p2; // Noncompliant
     protected readonly IntPtr p3;
     public IntPtr p4; // Noncompliant
 
     private UIntPtr p5;
-    protected UIntPtr p6; // Noncompliant
```
andrei-epure-sonarsource, line 11: why replace this test with the `char*` one? 
(I'm not suggesting to change it, just trying to understand)

analyzers/src/SonarAnalyzer.CSharp/Rules/PointersShouldBePrivate.cs
```diff
@@ -53,25 +53,38 @@ protected override void Initialize(SonarAnalysisContext context)
...
+        private static IFieldSymbol SymbolIfPointerType(VariableDeclarationSyntax variableDeclaration, VariableDeclaratorSyntax variableDeclarator, SemanticModel semanticModel)
+        {
+            if (variableDeclaration.Type.IsKind(SyntaxKind.PointerType)
+                || IsUnmanagedFunctionPointer(variableDeclaration))
+            {
+                return (IFieldSymbol)semanticModel.GetDeclaredSymbol(variableDeclarator);
+            }
+            else
+            {
+                return ((IFieldSymbol)semanticModel.GetDeclaredSymbol(variableDeclarator)) is { } variableSymbol
```
andrei-epure-sonarsource, line 77: you can do an AST-based check on `variableDeclaration.Type` to be / contain `IntPtr` or `UIntPtr`, to avoid calling the semantic model for each field symbol?

andrei-epure-sonarsource (changes requested):
I suggest to do an AST-check before querying the symbol.

Looks great in the rest

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PointersShouldBePrivate.CSharp9.cs
csaba-sagi-sonarsource: IntPtr case is already covered with field p4, and just wanted to add a test case with the pointer types.

andrei-epure-sonarsource (approved):
LGTM

PR #4829, opened 2021-08-24
Rule S3927: Add support for attributes on local functions

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.Common/Rules/ImplementSerializationMethodsCorrectlyBase.cs
```diff
@@ -29,29 +29,35 @@ namespace SonarAnalyzer.Rules
     public abstract class ImplementSerializationMethodsCorrectlyBase : SonarDiagnosticAnalyzer
     {
         protected const string DiagnosticId = "S3927";
+
+        private const string AttributeOnLocalMethodMessageFormat = "Serialization attributes on local methods are not considered.";
```
andrei-epure-sonarsource, line 33:
```suggestion
        private const string AttributeOnLocalMethodMessageFormat = "Serialization attributes on local functions are not considered.";
```

andrei-epure-sonarsource: "local function" is the correct term
- https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions
- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/local-function-attributes

analyzers/src/SonarAnalyzer.Common/Rules/ImplementSerializationMethodsCorrectlyBase.cs
```diff
@@ -29,29 +29,35 @@ namespace SonarAnalyzer.Rules
...
         private const string MessageFormat = "Make this method {0}.";
         private const string ProblemParameterText = "have a single parameter of type 'StreamingContext'";
         private const string ProblemGenericParameterText = "have no type parameters";
         private const string ProblemPublicText = "non-public";
 
         private readonly DiagnosticDescriptor rule;
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+
+        protected readonly DiagnosticDescriptor AttributeOnLocalMethodRule;
```
andrei-epure-sonarsource, line 41:
```suggestion
        protected readonly DiagnosticDescriptor AttributeOnLocalFunctionRule;
```

andrei-epure-sonarsource: also - can this be private (and `attributeOnLocalFunctionRule` in that case)?

analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementSerializationMethodsCorrectly.cs
```diff
@@ -48,5 +51,29 @@ public sealed class ImplementSerializationMethodsCorrectly : ImplementSerializat
                 .FirstOrDefault()
                 ?.Identifier
                 .GetLocation();
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(c =>
+                                                           {
```
andrei-epure-sonarsource, line 58: I find this indenting exaggerating - are we expected to change this to _all_ classes? what's the advantage here? 

Please revert. This is introducing a style change and making our code inconsistent and harder to read.

andrei-epure-sonarsource (changes requested):

PR #4830, opened 2021-08-24
S2696 - update tests

andrei-epure-sonarsource:
IMO static lambdas should work as they currently work.

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/StaticFieldWrittenFromInstanceConstructor.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/StaticFieldWrittenFromInstanceConstructor.cs
@@ -32,31 +32,17 @@ namespace SonarAnalyzer.Rules.CSharp
     [Rule(DiagnosticId)]
     public sealed class StaticFieldWrittenFromInstanceConstructor : StaticFieldWrittenFrom
     {
-        internal const string DiagnosticId = "S3010";
+        private const string DiagnosticId = "S3010";
         private const string MessageFormat = "Remove this assignment of '{0}' or initialize it statically.";
 
-        private static readonly DiagnosticDescriptor rule =
+        private static readonly DiagnosticDescriptor Rule =
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
-        protected override bool IsValidCodeBlockContext(SyntaxNode node, ISymbol owningSymbol)
-        {
-            if (!(node is ConstructorDeclarationSyntax declaration))
-            {
-                return false;
-            }
+        protected override bool IsValidCodeBlockContext(SyntaxNode node, ISymbol owningSymbol) =>
+            node is ConstructorDeclarationSyntax declaration
+            && !declaration.Modifiers.Any(SyntaxKind.StaticKeyword);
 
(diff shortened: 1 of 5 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/StaticFieldWrittenFromInstanceMember.cs
```diff
@@ -31,34 +31,18 @@ namespace SonarAnalyzer.Rules.CSharp
...
-                {
-                    return false;
-                }
-            }
-
-            return true;
-        }
+        protected override bool IsValidCodeBlockContext(SyntaxNode node, ISymbol owningSymbol) =>
+            (owningSymbol != null && !owningSymbol.IsStatic)
+            && (node is MethodDeclarationSyntax || node is AccessorDeclarationSyntax);
```
andrei-epure-sonarsource, line 45: Could this be done before getting the symbol

andrei-epure-sonarsource: we get the symbol from the context, all good

andrei-epure-sonarsource:
I tried to add coverage for the defensive coding but could not.

analyzers/src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <LangVersion>8.0</LangVersion>
+    <LangVersion>9.0</LangVersion>
```
andrei-epure-sonarsource, line 4: I will change the other projects lang version (and move it to Directory.Build.props) in a future PR

analyzers/src/SonarAnalyzer.CSharp/Rules/StaticFieldWrittenFromInstanceConstructor.cs
```diff
@@ -32,31 +32,17 @@ namespace SonarAnalyzer.Rules.CSharp
...
-            }
-
-            return true;
-        }
-
-        protected override string GetDiagnosticMessageArgument(SyntaxNode node, ISymbol owningSymbol, IFieldSymbol field)
-        {
-            return field.Name;
-        }
+        protected override string GetDiagnosticMessageArgument(SyntaxNode node, ISymbol owningSymbol, IFieldSymbol field) => field.Name;
```
csaba-sagi-sonarsource, line 46: If I interpret our coding style correctly this should go to a new line.
"Expression body of method should be on the same line only for trivial cases: literal or simple identifier. Member access, indexer, invocation and other complex structures should be on the next line."

analyzers/src/SonarAnalyzer.CSharp/Rules/StaticFieldWrittenFromInstanceMember.cs
```diff
@@ -31,34 +32,18 @@ namespace SonarAnalyzer.Rules.CSharp
...
-                if (declaration == null)
-                {
-                    return false;
-                }
-            }
-
-            return true;
-        }
+        protected override bool IsValidCodeBlockContext(SyntaxNode node, ISymbol owningSymbol) =>
+            (owningSymbol != null && !owningSymbol.IsStatic)
```
csaba-sagi-sonarsource, line 45:
```suggestion
           owningSymbol is { IsStatic: false }
```

csaba-sagi-sonarsource (approved):
LGTM! I left two minor suggestions.

analyzers/src/SonarAnalyzer.CSharp/Rules/StaticFieldWrittenFromInstanceConstructor.cs
andrei-epure-sonarsource: I would consider this trivial enough...

Issue #4799, opened 2021-08-13
S4261: false diagnostic when using generic constraints


anreton:
### Description
[S4261](https://rules.sonarsource.com/csharp/RSPEC-4261)

### Repro steps

```csharp
private TResult ExecuteAsync<TResult>(Func<Client, TResult> selector) where TResult : Task
{
	using (new OperationContextScope(this.client.InnerChannel))
	{
		return selector(this.client) ?? throw new InvalidOperationException();
	}
}
```

### Expected behavior

No false diagnostic for S4261.

### Actual behavior

Get false diagnostic for S4261.

### Related information

- Microsoft Visual Studio Community 2019 Version 16.11.0
- SonarAnalyzer.CSharp 8.27.0.35380 via Nuget

dmitry-golovinov-sonarsource:
Hello @anreton! Thanks for leaving your feedback.  As it is stated in the RSPEC you've mentioned, this rule is deprecated, and will eventually be removed. We are not going to provide any support regarding deprecated rules. If you have activated it manually using ruleset or .editorconfig file, I recommend you to disable it.

dmitry-golovinov-sonarsource:
Reopened the issue due to this thread https://github.com/SonarSource/sonar-dotnet/issues/4059

PR #4833, opened 2021-08-25
Rule S4261: Undeprecate and add FP example

### LINKED ISSUE #4059: Make S4261 active again (revert deprecation) and keep it out of the SonarWay default profile
I see that [rule S4261 has recently been deprecated](https://github.com/SonarSource/sonar-dotnet/pull/4028) and I've had a quick read around the context for this:

* https://community.sonarsource.com/t/s4261-has-incorrect-examples/36324/6
* https://github.com/SonarSource/sonar-dotnet/issues/3928

Whilst acknowledging the weaknesses in that old rule, I do still find it useful.  I'd very much like there to be a new rule added, replacing that deprecated one (albeit using the correct rule logic) which I can use instead.

dmitry-golovinov-sonarsource:
Fixes #4059 
Documents #4799


csaba-sagi-sonarsource (approved):
LGTM!

PR #4834, opened 2021-08-25
S2699 - raise on local functions for xUnit 

andrei-epure-sonarsource:
RE: MMF-2464 and S2699

Only xUnit allows for  local functions to be test methods.

For NUnit and VSTest, a test method **must be public**, thus local functions don't work (there's S3433 to raise for that scenario - local functions with test attributes - #4838).

Tested this behavior with NUnit and VSTest - see `andrei/test-sample` branch with UT - https://github.com/SonarSource/sonar-dotnet/tree/andrei/test-sample/analyzers/its/sources/Net5/Net5Test
- for VSTest, the tests don't show up at all
- for NUnit , there is an exception in the test explorer
image


andrei-epure-sonarsource:
Offline talk with @csaba-sagi-sonarsource - we said it's ok to have them `using csharp::SonarAnalyzer.Wrappers;` grouped together with `using SonarAnalyzer.UnitTest.TestFramework;` because it's the same Sonar* namespace

PR #4835, opened 2021-08-25
S3603 Attributes on local functions support

csaba-sagi-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/PureAttributeOnVoidMethod.cs
```diff
@@ -31,5 +32,19 @@ namespace SonarAnalyzer.Rules.CSharp
...
     {
         protected override ILanguageFacade<SyntaxKind> Language => CSharpFacade.Instance;
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            base.Initialize(context);
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
+                {
+                    if (InvalidPureDataAttributeUsage(c.SemanticModel.GetDeclaredSymbol(c.Node)) is { } pureAttribute)
```
andrei-epure-sonarsource, line 42: could you first check the Attribute name at AST-level? To avoid querying symbols for all local functions (even those with no attributes or with attributes which are not `[Pure]` or `[PureAttribute]`)

analyzers/src/SonarAnalyzer.Common/Rules/PureAttributeOnVoidMethodBase.cs
```diff
@@ -31,26 +31,31 @@ public abstract class PureAttributeOnVoidMethodBase<TSyntaxKind> : SonarDiagnost
...
-                    if (c.Symbol is IMethodSymbol { ReturnsVoid: true } methodSymbol
-                        && !methodSymbol.Parameters.Any(p => p.RefKind != RefKind.None)
-                        && methodSymbol.GetAttributes().FirstOrDefault(a => a.AttributeClass.Is(KnownType.System_Diagnostics_Contracts_PureAttribute)) is { } pureAttribute)
+                    if (InvalidPureDataAttributeUsage(c.Symbol) is { } pureAttribute)
                     {
-                        c.ReportDiagnosticWhenActive(Diagnostic.Create(rule, pureAttribute.ApplicationSyntaxReference.GetSyntax().GetLocation()));
+                        c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, pureAttribute.ApplicationSyntaxReference.GetSyntax().GetLocation()));
                     }
                 },
                 SymbolKind.Method);
```
andrei-epure-sonarsource, line 52: it's weird that this doesn't work for local functions.. because it's still an `IMethodSymbol` - did you check to see for what [SymbolKind](https://github.com/dotnet/roslyn/blob/main/src/Compilers/Core/Portable/Symbols/SymbolKind.cs) the local function symbol is returned?

andrei-epure-sonarsource (changes requested):

andrei-epure-sonarsource (commented):
Also please consider adding an IT to the `Net5` ITs project

analyzers/src/SonarAnalyzer.Common/Rules/PureAttributeOnVoidMethodBase.cs
andrei-epure-sonarsource: nevermind - it's an open issue https://github.com/dotnet/roslyn/issues/35770

PR #4840, opened 2021-08-26
Update coding style (2021-06-24,  2021-08-26 decisions)

andrei-epure-sonarsource:


docs/coding-style.md
```diff
@@ -39,8 +40,26 @@ Variable name `sut` (System Under Test) is recommended in unit tests that really
 ## Multi-line statements
 
 * Operators (&&, ||, +, :, ?, ?? and others) are placed at the beginning of a line.
-* Dots with invocation `.Method()` are placed at the beginning of a line.
-* Comma separating arguments is placed at the end of a line.
+* Dot inside an invocation `.Method()` is placed at the beginning of a line.
```
pavel-mikula-sonarsource, line 43: Even when it's syntactically `inside` of the `SimpleMemberAccessExpression`
```suggestion
* Dot before an invocation `.Method()` is placed at the beginning of a line.
```

docs/coding-style.md
```diff
@@ -39,8 +40,26 @@ Variable name `sut` (System Under Test) is recommended in unit tests that really
...
+* The comma separating arguments is placed at the end of a line.
+* Method declaration parameters should be on the same line. If S103 is violated, parameters should be placed each on a separate line; the first parameter should be on the same line with the declaration; the other parameters should be aligned with the first parameter.
+```
+public void MethodWithManyParameters(int firstParameter,
+                                     string secondParameter,
+                                     Function<int, string, string> complexParameter);
+```
```
pavel-mikula-sonarsource, line 50: I intentionally didn't document this last time, because we said we need tooling first that we don't have. But with `should` it's probably OK-ish.

docs/coding-style.md
```diff
@@ -39,8 +40,26 @@ Variable name `sut` (System Under Test) is recommended in unit tests that really
...
+* Method invocation arguments should be placed on the same line only when they are few and simple. Otherwise, they should be placed on separate lines. The first argument should be on a separate line, aligned with a left-most single indendation.
+```
+object.MethodName(
+  firstArgument,
+  x => x.Bar(),
+  thirdArgument.Property);
```
pavel-mikula-sonarsource, line 56: Default tab is 4 spaces.
```suggestion
    firstArgument,
    x => x.Bar(),
    thirdArgument.Property);
```

pavel-mikula-sonarsource (approved):
LGTM with some suggestions

docs/coding-style.md
andrei-epure-sonarsource: I added it because from the last many PRs, I see this is the way we go.

Indeed if we'd have tooling in place, it shouldn't need to be documented.

PR #4847, opened 2021-08-30
Fix S4830 FN: Extended partial methods feature support is missing

### LINKED ISSUE #4415: Fix S4830 FN: Extended partial methods feature support is missing
### Description

S4830 rule does not support classes with extended partial methods

### Repro steps

```
    public partial class PartialClass
    {
        public static partial RemoteCertificateValidationCallback FindInvalid()
        {
            return (sender, certificate, chain, SslPolicyErrors) => true;  // Secondary FN
        }
    }

    public partial class PartialClass
    {
        public static partial RemoteCertificateValidationCallback FindInvalid();

        HttpWebRequest CreateRQ()
        {
            return (HttpWebRequest)System.Net.HttpWebRequest.Create("http://localhost");
        }

        public void Init()
        {
            CreateRQ().ServerCertificateValidationCallback += FindInvalid();  // Non-compliant FN
        }

        static void Execute()
        {
            new PartialClass().Init();
        }
    }

```
### Expected behavior

Issue should be raised inside Init method with a secondary location in FindInvalid method.

### Actual behavior

Issue is not being raised.

### Related information

costin-zaharia-sonarsource:
Fixes #4415

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/CertificateValidationCheckBase.cs
@@ -201,19 +201,7 @@ namespace SonarAnalyzer.Rules
                 case TLambdaSyntax lambda:
                     return LambdaLocations(c, lambda);
                 case TInvocationExpressionSyntax invocation:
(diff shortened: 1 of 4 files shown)


csaba-sagi-sonarsource (approved):
LGTM!

PR #4848, opened 2021-08-30
Introduce IOperation wrapper

pavel-mikula-sonarsource:
There already exist `IOperationWrapper` interface in the ShimLayer as a base type for all wrapped interfaces.

This class has the same prefix to be easily found.


dmitry-golovinov-sonarsource (approved):
LGTM!

PR #4850, opened 2021-08-30
Fix S3881 FP: Extended partial methods feature support is missing

### LINKED ISSUE #4402: Fix S3881 FP: Extended partial methods feature support is missing
### Description

S3881 rule does not support classes with extended partial methods

### Repro steps

```
    public partial record PartialCompliant : IDisposable // Noncompliant FP, should be Compliant
                                                                                      // Secondary@-1 FP
    {
        public partial void Dispose();
    }

    public partial record PartialCompliant // Noncompliant FP, should be Compliant
                                                                 // Secondary@-1 FP
    {
        public partial void Dispose() { }
    }
```

### Expected behavior

PartialCompliant type declaration should be considered as compliant for this rule.

### Actual behavior

Issues being raised for both partial record declarations.

### Related information

costin-zaharia-sonarsource:
Fixes #4402

### DIFF
--- a/analyzers/its/expected/Net5/Net5--net5.0-S3881.json
+++ b/analyzers/its/expected/Net5/Net5--net5.0-S3881.json
@@ -40,6 +40,30 @@
 {
 "uri":  "sources\Net5\Net5\S2952.Part1.cs",
 "region":  {
+"startLine":  6,
+"startColumn":  26,
+"endLine":  6,
+"endColumn":  40
+}
+},
+{
+"uri":  "sources\Net5\Net5\S2952.Part1.cs",
+"region":  {
+"startLine":  6,
+"startColumn":  26,
+"endLine":  6,
+"endColumn":  40
+}
+}
+]
+},
+{
+"id":  "S3881",
+"message":  "Fix this implementation of 'IDisposable' to conform to the dispose pattern.",
+"location":  [
+{
+"uri":  "sources\Net5\Net5\S2952.Part1.cs",
+"region":  {
 "startLine":  19,
 "startColumn":  26,
 "endLine":  19,
@@ -53,6 +77,15 @@
 "startColumn":  26,
 "endLine":  19,
 "endColumn":  40
+}
+},
+{
+"uri":  "sources\Net5\Net5\S2952.Part1.cs",
+"region":  {
+"startLine":  24,
+"startColumn":  29,
+"endLine":  24,
+"endColumn":  36
 }
 }
 ]
@@ -78,29 +111,14 @@
 "endLine":  28,
 "endColumn":  41
 }
-}
-]
-},
-{
-"id":  "S3881",
-"message":  "Fix this implementation of 'IDisposable' to conform to the dispose pattern.",
-"location":  [
-{
-"uri":  "sources\Net5\Net5\S2952.Part1.cs",
-"region":  {
-"startLine":  6,
-"startColumn":  26,
-"endLine":  6,
-"endColumn":  40
-}
 },
 {
 "uri":  "sources\Net5\Net5\S2952.Part1.cs",
 "region":  {
-"startLine":  6,
-"startColumn":  26,
-"endLine":  6,
-"endColumn":  40
+"startLine":  33,
+"startColumn":  29,
+"endLine":  33,
+"endColumn":  36
 }
 }
 ]
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
@@ -114,7 +114,7 @@ namespace SonarAnalyzer.Rules.CSharp
 
                     if (HasVirtualDisposeBool(typeSymbol.BaseType))
                     {
-                        VerifyDisposeOverrideCallsBase(FindMethodImplementation(typeSymbol, IsDisposeBool, typeDeclarationSyntax)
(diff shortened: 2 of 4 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
```diff
@@ -269,15 +268,20 @@ private void VerifyDispose(MethodDeclarationSyntax disposeMethod, bool isSealedC
...
-                      .OfType<IMethodSymbol>()
-                      .Where(IsDisposeBool)
-                      .Any(methodSym => !methodSym.IsAbstract);
+                          .SelectMany(symbol => symbol.PartialImplementationPart?.DeclaringSyntaxReferences ?? symbol.DeclaringSyntaxReferences)
+                          .Select(reference => reference.GetSyntax());
+
+            private static IEnumerable<SyntaxNode> FindMethodImplementation(INamedTypeSymbol typeSymbol, Func<IMethodSymbol, bool> predicate, TypeDeclarationSyntax typeDeclarationSyntax) =>
+                FindMethodDeclarations(typeSymbol, predicate)
+                    .OfType<BaseMethodDeclarationSyntax>()
+                    .Where(syntax => typeDeclarationSyntax.Contains(syntax) && (syntax.HasBodyOrExpressionBody() || syntax.Modifiers.AnyOfKind(SyntaxKind.AbstractKeyword)));
```
andrei-epure-sonarsource, line 277:
```suggestion
                    .Where(x => typeDeclarationSyntax.Contains(x) && (x.HasBodyOrExpressionBody() || x.Modifiers.AnyOfKind(SyntaxKind.AbstractKeyword)));
```

(it's clear that it's a BaseMethodDeclarationSyntax here)

andrei-epure-sonarsource: why do you check for `AbstractKeyword` here? I couldn't find it in the tests

analyzers/its/expected/Net5/Net5--net5.0-S3881.json
```diff
@@ -40,18 +40,18 @@
 {
 "uri":  "sources\Net5\Net5\S2952.Part1.cs",
```
andrei-epure-sonarsource, line 41: was there any change in this IT? I see that some lines moved (line 6 was at the end of the file)

analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
```diff
@@ -269,15 +268,20 @@ private void VerifyDispose(MethodDeclarationSyntax disposeMethod, bool isSealedC
...
-            private static bool HasVirtualDisposeBool(INamedTypeSymbol symbol) =>
-                symbol.GetSelfAndBaseTypes()
-                      .SelectMany(t => t.GetMembers())
-                      .OfType<IMethodSymbol>()
-                      .Where(IsDisposeBool)
-                      .Any(methodSym => !methodSym.IsAbstract);
+                          .SelectMany(symbol => symbol.PartialImplementationPart?.DeclaringSyntaxReferences ?? symbol.DeclaringSyntaxReferences)
+                          .Select(reference => reference.GetSyntax());
+
+            private static IEnumerable<SyntaxNode> FindMethodImplementation(INamedTypeSymbol typeSymbol, Func<IMethodSymbol, bool> predicate, TypeDeclarationSyntax typeDeclarationSyntax) =>
```
andrei-epure-sonarsource, line 274: this method returns both implementations (partial or not) and abstract methods (so not implemented). should improve the name. e.g. `FindMethodImplementationOrAbstractDeclaration` (or better)

Would be useful to document why we look also for abstract methods.

andrei-epure-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.CSharp/Rules/ImplementIDisposableCorrectly.cs
costin-zaharia-sonarsource: We check for `AbstractKeyword` since I want to preserve the existing functionality for abstract methods and raise secondary issues for them. 

You can find the test cases with abstract methods here: https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ImplementIDisposableCorrectly.AbstractClass.cs#L14

analyzers/its/expected/Net5/Net5--net5.0-S3881.json
costin-zaharia-sonarsource: Although there were no changes in the ITs, we had an existing case with partial methods for which the behavior changed. 

You can find it here: https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/its/sources/Net5/Net5/S2952.Part1.cs

Raised by the old version:
Issue 1: Primary (line 6) + Secondary (line 6)
Issue 2: Primary (line 19) + Secondary (line 19)
Issue 3: Primary (line 28) + Secondary (line 28)

Raised by the new version:
Issue 1: Primary (line 6) + Secondary (line 6)
Issue 2: Primary (line 19) + Secondary (line 19) + Secondary (line 24)
Issue 3: Primary (line 28) + Secondary (line 28) + Secondary (line 33)

The difference is the additional secondary location with the message:
> 'ResourceHolder2.Dispose()' should call 'Dispose(true)' and 'GC.SuppressFinalize(this)'.

This was previously raised for "normal" methods but not for partial ones.

PR #4854, opened 2021-08-31
Introduce the possibility to create CFG from IOperation

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.CFG/Roslyn/ControlFlowGraph.cs
```diff
@@ -60,6 +61,12 @@ static ControlFlowGraph()
                 OriginalOperationProperty = type.GetProperty(nameof(OriginalOperation));
                 RootProperty = type.GetProperty(nameof(Root));
                 CreateMethod = type.GetMethod(nameof(Create), new[] { typeof(SyntaxNode), typeof(SemanticModel), typeof(CancellationToken) });
+                CreateCoreMethod = type.GetMethod(
```
pavel-mikula-sonarsource, line 64: We should not reflect on non-public members to keep the solution clean. They obviously don't want users to be able to create the CFG from any IOperation. Isn't there any better way?

pavel-mikula-sonarsource (changes requested):

PR #4855, opened 2021-08-31
Rule S2190: Supports Roslyn CFG

### LINKED ISSUE #2342: Update S2190: FN with try/catch block
### Description

Sorry for my english, the example of the rule S2190 indicates the suggestion of a conditional to stop the method, ¿a try / catch block is thought for this even though it does not do something?

### Repro steps

Rule example:

```
int Pow(int num, int exponent)   // Noncompliant; no condition under which pow isn't re-called
{
  num = num * Pow(num, exponent-1);
  return num;  // this is never reached
}
```

Try/Catch example:
```
int Pow(int num, int exponent)   
        {
            try
            {
                num = num * Pow(num, exponent - 1);
                return num;  // this is never reached
            }
            catch (Exception ex) // Compliant;
            {
                registerLog(string.Concat(ex.Message, "\n ", ex.StackTrace), GetType(), Enumeration.Error);
                throw;
            }                      
        }
```

### Related information

* SonarC# Version: 7.3 (build 5690)instalado
* Visual Studio Version: Microsoft Visual Studio Professional 2017 - 15.9.8

### LINKED ISSUE #3624: Fix S2190 FN: Consider property setter as recursion
### Description

S2190 triggers infinite recursion warning on the getter. Same warning should be triggered for the setter as well, because the setter is also recursive call.

```C#
public string Name
{
    get // Noncompliant
    {
        return Name;
    }
    set // False Negative
    {
        Name = value;
    }
}
```

Original source: [community thread](https://community.sonarsource.com/t/getters-and-setters-without-private-variable/30719)

dmitry-golovinov-sonarsource:
Fixes #2342
Fixes #3624

### DIFF
--- a/analyzers/src/SonarAnalyzer.CFG/Extensions/IOperationExtensions.cs
+++ b/analyzers/src/SonarAnalyzer.CFG/Extensions/IOperationExtensions.cs
@@ -49,6 +49,9 @@ namespace SonarAnalyzer.Extensions
         public static OperationExecutionOrder ToExecutionOrder(this IEnumerable<IOperation> operations) =>
             new OperationExecutionOrder(operations, false);
 
+        public static OperationExecutionOrder ToReversedExecutionOrder(this IOperation operation) =>
+            new[] { operation }.ToReversedExecutionOrder();
+
         public static OperationExecutionOrder ToReversedExecutionOrder(this IEnumerable<IOperation> operations) =>
             new OperationExecutionOrder(operations, true);
 
--- a/analyzers/src/SonarAnalyzer.CFG/Extensions/IOperationExtensions.cs
+++ b/analyzers/src/SonarAnalyzer.CFG/Extensions/IOperationExtensions.cs
@@ -36,6 +36,16 @@ namespace SonarAnalyzer.Extensions
         public static bool IsAnyKind(this IOperation operation, params OperationKind[] kinds) =>
             kinds.Contains(operation.Kind);
 
+        public static IOperation RootOperation(this IOperation operation)
+        {
+            var wrapper = new IOperationWrapperSonar(operation);
+            while (wrapper.Parent != null)
+            {
+                wrapper = new IOperationWrapperSonar(wrapper.Parent);
+            }
+            return wrapper.Instance;
+        }
+
         public static OperationExecutionOrder ToExecutionOrder(this IEnumerable<IOperation> operations) =>
             new OperationExecutionOrder(operations, false);
 
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
@@ -53,7 +53,7 @@ namespace SonarAnalyzer.Rules.CSharp
                         var cfg = ControlFlowGraph.Create(accessor, c.SemanticModel);
                         var context = new RecursionContext<ControlFlowGraph>(cfg, propertySymbol, accessor.Keyword.GetLocation(), c, "property accessor's recursion");
                         var walker = new CommonRecursionSearcher(context, !accessor.Keyword.IsKind(SyntaxKind.SetKeyword));
-                        if (!walker.CheckPaths() && !cfg.ExitBlock.IsReachable)
+                        if (!walker.CheckPaths() && CfgHasUnescapableLoop(cfg))
                         {
                             context.ReportIssue(accessor.Keyword.GetLocation());
                         }
@@ -83,17 +83,28 @@ namespace SonarAnalyzer.Rules.CSharp
                             var operationWrapper = cfg.FlowAnonymousFunctionOperations().Single(x => x.WrappedOperation.Syntax == enclosingFunction);
                             cfg = cfg.GetAnonymousFunctionControlFlowGraph(operationWrapper);
                         }
+                        if (cfg == null)
+                        {
+                            return;
+                        }
                     }
 
                     cfg = cfg.GetLocalFunctionControlFlowGraph(symbol as IMethodSymbol);
+                    if (cfg == null)
+                    {
+                        return;
+                    }
                 }
 
                 var walker = new CommonRecursionSearcher(new RecursionContext<ControlFlowGraph>(cfg, symbol, identifier.GetLocation(), c, "method's recursion"));
-                if (!walker.CheckPaths() && !cfg.ExitBlock.IsReachable)
+                if (!walker.CheckPaths() && CfgHasUnescapableLoop(cfg))
                 {
                     c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, identifier.GetLocation(), "method's recursion"));
                 }
             }
+
+            private static bool CfgHasUnescapableLoop(ControlFlowGraph cfg) =>
+                !cfg.ExitBlock.IsReachable && !cfg.Blocks.Any(x => x.FallThroughSuccessor?.Semantics == ControlFlowBranchSemantics.Throw && x.IsReachable);
 
             private class CommonRecursionSearcher : CfgAllPathValidator
             {
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
@@ -18,7 +18,6 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.CodeAnalysis;
@@ -42,7 +41,7 @@ namespace SonarAnalyzer.Rules.CSharp
                 if (property.ExpressionBody?.Expression != null)
                 {
                     var cfg = ControlFlowGraph.Create(property.ExpressionBody, c.SemanticModel);
-                    var walker = new CommonRecursionSearcher(new RecursionContext<ControlFlowGraph>(cfg, propertySymbol, property.Identifier.GetLocation(), c, "property's recursion"));
+                    var walker = new RecursionSearcher(new RecursionContext<ControlFlowGraph>(cfg, propertySymbol, property.Identifier.GetLocation(), c, "property's recursion"));
                     walker.CheckPaths();
                     // cannot meet goto s here, check for unreachable exit node is not needed here
                 }
@@ -52,7 +51,7 @@ namespace SonarAnalyzer.Rules.CSharp
                     {
                         var cfg = ControlFlowGraph.Create(accessor, c.SemanticModel);
                         var context = new RecursionContext<ControlFlowGraph>(cfg, propertySymbol, accessor.Keyword.GetLocation(), c, "property accessor's recursion");
-                        var walker = new CommonRecursionSearcher(context, !accessor.Keyword.IsKind(SyntaxKind.SetKeyword));
+                        var walker = new RecursionSearcher(context, !accessor.Keyword.IsKind(SyntaxKind.SetKeyword));
                         if (!walker.CheckPaths() && CfgHasUnescapableLoop(cfg))
                         {
                             context.ReportIssue(accessor.Keyword.GetLocation());
@@ -96,35 +95,34 @@ namespace SonarAnalyzer.Rules.CSharp
                     }
                 }
 
-                var walker = new CommonRecursionSearcher(new RecursionContext<ControlFlowGraph>(cfg, symbol, identifier.GetLocation(), c, "method's recursion"));
+                var context = new RecursionContext<ControlFlowGraph>(cfg, symbol, identifier.GetLocation(), c, "method's recursion");
+                var walker = new RecursionSearcher(context);
                 if (!walker.CheckPaths() && CfgHasUnescapableLoop(cfg))
                 {
-                    c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, identifier.GetLocation(), "method's recursion"));
+                    context.ReportIssue();
                 }
             }
 
             private static bool CfgHasUnescapableLoop(ControlFlowGraph cfg) =>
                 !cfg.ExitBlock.IsReachable && !cfg.Blocks.Any(x => x.FallThroughSuccessor?.Semantics == ControlFlowBranchSemantics.Throw && x.IsReachable);
 
-            private class CommonRecursionSearcher : CfgAllPathValidator
+            private class RecursionSearcher : CfgAllPathValidator
             {
-                private readonly ISymbol memberToCheck;
-                private readonly Action reportIssue;
+                private readonly RecursionContext<ControlFlowGraph> context;
                 private readonly bool isGetAccesor;
 
-                public CommonRecursionSearcher(RecursionContext<ControlFlowGraph> context, bool isGet = true)
+                public RecursionSearcher(RecursionContext<ControlFlowGraph> context, bool isGetAccesor = true)
                     : base(context.ControlFlowGraph)
                 {
-                    memberToCheck = context.AnalyzedSymbol;
-                    reportIssue = context.ReportIssue;
-                    isGetAccesor = isGet;
+                    this.context = context;
+                    this.isGetAccesor = isGetAccesor;
                 }
 
                 public bool CheckPaths()
                 {
                     if (CheckAllPaths())
                     {
-                        reportIssue();
+                        context.ReportIssue();
                         return true;
                     }
 
@@ -135,9 +133,9 @@ namespace SonarAnalyzer.Rules.CSharp
                 {
                     foreach (var operation in block.OperationsAndBranchValue)
                     {
-                        foreach (var child in operation.DescendantsAndSelf().ToReversedExecutionOrder())
+                        foreach (var child in operation.ToReversedExecutionOrder())
                         {
-                            if (memberToCheck.Equals(MemberSymbol(child.Instance)))
+                            if (context.AnalyzedSymbol.Equals(MemberSymbol(child.Instance)))
                             {
                                 var isWrite = child.Parent is { Kind: OperationKindEx.SimpleAssignment } parent && ISimpleAssignmentOperationWrapper.FromOperation(parent).Target == child.Instance;
                                 return isGetAccesor ^ isWrite;
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
@@ -78,17 +78,9 @@ namespace SonarAnalyzer.Rules.CSharp
                             var operationWrapper = cfg.FlowAnonymousFunctionOperations().Single(x => x.WrappedOperation.Syntax == enclosingFunction);
                             cfg = cfg.GetAnonymousFunctionControlFlowGraph(operationWrapper);
                         }
-                        if (cfg == null)
-                        {
-                            return;
-                        }
                     }
 
                     cfg = cfg.GetLocalFunctionControlFlowGraph(symbol as IMethodSymbol);
-                    if (cfg == null)
-                    {
-                        return;
-                    }
                 }
 
                 var context = new RecursionContext<ControlFlowGraph>(cfg, symbol, identifier.GetLocation(), c, "method's recursion");
@@ -110,7 +102,7 @@ namespace SonarAnalyzer.Rules.CSharp
 
                 public void CheckPaths()
                 {
-                    if (CheckAllPaths() || CfgHasUnescapableLoop())
+                    if (!CfgCanExit() || CheckAllPaths())
                     {
                         context.ReportIssue();
                     }
@@ -118,15 +110,13 @@ namespace SonarAnalyzer.Rules.CSharp
 
                 protected override bool IsValid(BasicBlock block)
                 {
-                    foreach (var operation in block.OperationsAndBranchValue)
+                    foreach (var operation in block.OperationsAndBranchValue.ToReversedExecutionOrder())
                     {
-                        foreach (var child in operation.ToReversedExecutionOrder())
+                        if (context.AnalyzedSymbol.Equals(MemberSymbol(operation.Instance)))
                         {
-                            if (context.AnalyzedSymbol.Equals(MemberSymbol(child.Instance)))
-                            {
-                                var isWrite = child.Parent is { Kind: OperationKindEx.SimpleAssignment } parent && ISimpleAssignmentOperationWrapper.FromOperation(parent).Target == child.Instance;
-                                return isGetAccesor ^ isWrite;
-                            }
+                            var isWrite = operation.Parent is { Kind: OperationKindEx.SimpleAssignment } parent
+                                          && ISimpleAssignmentOperationWrapper.FromOperation(parent).Target == operation.Instance;
+                            return isGetAccesor ^ isWrite;
                         }
                     }
                     return false;
@@ -142,9 +132,9 @@ namespace SonarAnalyzer.Rules.CSharp
 
                 protected override bool IsInvalid(BasicBlock block) => false;
 
-                private bool CfgHasUnescapableLoop() =>
-                    !context.ControlFlowGraph.ExitBlock.IsReachable
-                    && !context.ControlFlowGraph.Blocks.Any(x => x.FallThroughSuccessor?.Semantics == ControlFlowBranchSemantics.Throw && x.IsReachable);
+                private bool CfgCanExit() =>
+                    context.ControlFlowGraph.ExitBlock.IsReachable
+                    || context.ControlFlowGraph.Blocks.Any(x => x.FallThroughSuccessor?.Semantics == ControlFlowBranchSemantics.Throw && x.IsReachable);
             }
         }
     }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
@@ -110,7 +110,6 @@ namespace SonarAnalyzer.Rules.CSharp
             {
                 private readonly ISymbol memberToCheck;
                 private readonly Action reportIssue;
-                private readonly ControlFlowGraph cfg;
                 private readonly bool isGetAccesor;
 
                 public CommonRecursionSearcher(RecursionContext<ControlFlowGraph> context, bool isGet = true)
@@ -118,7 +117,6 @@ namespace SonarAnalyzer.Rules.CSharp
                 {
                     memberToCheck = context.AnalyzedSymbol;
                     reportIssue = context.ReportIssue;
-                    cfg = context.ControlFlowGraph;
                     isGetAccesor = isGet;
                 }
 
@@ -133,48 +131,20 @@ namespace SonarAnalyzer.Rules.CSharp
                     return false;
                 }
 
-                protected override bool IsValid(BasicBlock block) =>
-                    FindReferenceToSelf(block);
-
-                protected override bool IsInvalid(BasicBlock block) => false;
-
-                private bool FindReferenceToSelf(BasicBlock block)
+                protected override bool IsValid(BasicBlock block)
                 {
                     foreach (var operation in block.OperationsAndBranchValue)
                     {
-                        if (ProcessOperation(operation, cfg, out var result))
+                        foreach (var child in operation.DescendantsAndSelf().ToReversedExecutionOrder())
                         {
-                            return result;
+                            if (memberToCheck.Equals(MemberSymbol(child.Instance)))
+                            {
+                                var isWrite = child.Parent is { Kind: OperationKindEx.SimpleAssignment } parent && ISimpleAssignmentOperationWrapper.FromOperation(parent).Target == child.Instance;
+                                return isGetAccesor ^ isWrite;
+                            }
                         }
                     }
                     return false;
-
-                    bool ProcessOperation(IOperation operation, ControlFlowGraph controlFlowGraph, out bool result)
-                    {
-                        foreach (var child in operation.DescendantsAndSelf().ToReversedExecutionOrder())
-                        {
-                            if (child.Instance.Kind == OperationKindEx.FlowAnonymousFunction)
-                            {
-                                var anonymousFunctionCfg = controlFlowGraph.GetAnonymousFunctionControlFlowGraph(IFlowAnonymousFunctionOperationWrapper.FromOperation(child.Instance));
-                                foreach (var subOperation in anonymousFunctionCfg.Blocks.SelectMany(x => x.OperationsAndBranchValue).SelectMany(x => x.DescendantsAndSelf()))
-                                {
-                                    if (ProcessOperation(subOperation, anonymousFunctionCfg, out result))
-                                    {
-                                        return true;
-                                    }
-                                }
-                            }
-                            else if (memberToCheck.Equals(MemberSymbol(child.Instance)))
-                            {
-                                var isWrite = child.Parent is { Kind: OperationKindEx.SimpleAssignment } parent && ISimpleAssignmentOperationWrapper.FromOperation(parent).Target == child.Instance;
-                                result = isGetAccesor ^ isWrite;
-                                return true;
-                            }
-                        }
-
-                        result = false;
-                        return false;
-                    }
 
                     static ISymbol MemberSymbol(IOperation operation) =>
                         operation.Kind switch
@@ -184,6 +154,8 @@ namespace SonarAnalyzer.Rules.CSharp
                             _ => null
                         };
                 }
+
+                protected override bool IsInvalid(BasicBlock block) => false;
             }
         }
     }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
@@ -36,35 +36,29 @@ namespace SonarAnalyzer.Rules.CSharp
 {
     public partial class InfiniteRecursion
     {
-        private class InfiniteRecursion_RoslynCfg : IInfiniteRecursion
+        private class RoslynChecker : IChecker
         {
             public void CheckForNoExitProperty(SyntaxNodeAnalysisContext c, PropertyDeclarationSyntax property, IPropertySymbol propertySymbol)
             {
                 if (property.ExpressionBody?.Expression != null)
                 {
                     var cfg = ControlFlowGraph.Create(property.ExpressionBody, c.SemanticModel);
-                    var walker = new CfgWalkerForProperty(
-                        new RecursionAnalysisContext<ControlFlowGraph>(cfg, propertySymbol, property.Identifier.GetLocation(), c),
-                        "property's recursion",
+                    var walker = new RecursionSearcherForProperty(
+                        new RecursionContext<ControlFlowGraph>(cfg, propertySymbol, property.Identifier.GetLocation(), c, "property's recursion"),
                         isSetAccessor: false);
                     walker.CheckPaths();
                     // cannot meet goto s here, check for unreachable exit node is not needed here
-                    return;
                 }
-
-                var accessors = property.AccessorList?.Accessors.Where(a => a.HasBodyOrExpressionBody());
-                if (accessors != null)
+                else if (property.AccessorList != null)
                 {
-                    foreach (var accessor in accessors)
+                    foreach (var accessor in property.AccessorList.Accessors.Where(a => a.HasBodyOrExpressionBody()))
                     {
                         var cfg = ControlFlowGraph.Create(accessor, c.SemanticModel);
-                        var walker = new CfgWalkerForProperty(
-                            new RecursionAnalysisContext<ControlFlowGraph>(cfg, propertySymbol, accessor.Keyword.GetLocation(), c),
-                            "property accessor's recursion",
-                            isSetAccessor: accessor.Keyword.IsKind(SyntaxKind.SetKeyword));
+                        var context = new RecursionContext<ControlFlowGraph>(cfg, propertySymbol, accessor.Keyword.GetLocation(), c, "property accessor's recursion");
+                        var walker = new RecursionSearcherForProperty(context, isSetAccessor: accessor.Keyword.IsKind(SyntaxKind.SetKeyword));
                         if (!walker.CheckPaths() && !cfg.ExitBlock.IsReachable)
                         {
-                            c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, accessor.Keyword.GetLocation(), "property accessor's recursion"));
+                            context.ReportIssue(accessor.Keyword.GetLocation());
                         }
                     }
                 }
@@ -73,13 +67,7 @@ namespace SonarAnalyzer.Rules.CSharp
             public void CheckForNoExitMethod(SyntaxNodeAnalysisContext c, CSharpSyntaxNode body, SyntaxToken identifier, ISymbol symbol)
             {
                 var operation = c.SemanticModel.GetOperation(body.Parent);
-                var wrapper = new IOperationWrapperSonar(operation);
-                while (wrapper.Parent != null)
-                {
-                    wrapper = new IOperationWrapperSonar(wrapper.Parent);
-                }
-
-                var cfg = ControlFlowGraph.Create(wrapper.Instance.Syntax, c.SemanticModel);
+                var cfg = ControlFlowGraph.Create(operation.RootOperation().Syntax, c.SemanticModel);
                 var enclosingKinds = new HashSet<SyntaxKind>
                 {
                     SyntaxKindEx.LocalFunctionStatement, SyntaxKind.SimpleLambdaExpression, SyntaxKind.AnonymousMethodExpression, SyntaxKind.ParenthesizedLambdaExpression
@@ -89,7 +77,7 @@ namespace SonarAnalyzer.Rules.CSharp
                     // we need to go up and track all possible enclosing local function statements
                     foreach (var enclosingFunction in c.Node.Ancestors().Where(x => enclosingKinds.Contains(x.Kind())).Reverse())
                     {
-                        if (enclosingFunction.IsAnyKind(SyntaxKindEx.LocalFunctionStatement))
+                        if (enclosingFunction.IsKind(SyntaxKindEx.LocalFunctionStatement))
                         {
                             cfg = cfg.GetLocalFunctionControlFlowGraph(c.SemanticModel.GetDeclaredSymbol(enclosingFunction) as IMethodSymbol);
                         }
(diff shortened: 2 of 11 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.cs
```diff
@@ -18,32 +18,43 @@
...
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        public static DiagnosticDescriptor Rule => DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
+
+        public InfiniteRecursion() : this(false) { }
+
+        public InfiniteRecursion(bool runSonarAnalyzer)
+        {
+            if (!CFG.Roslyn.ControlFlowGraph.IsAvailable || runSonarAnalyzer)
```
AraHaan, line 49: Why not shorten this down to:

```suggestion
            if (!CFG.Roslyn.ControlFlowGraph.IsAvailable || runSonarAnalyzer)
            
```

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.cs
```diff
@@ -18,32 +18,43 @@
...
+            {
+                analyzer = new InfiniteRecursionSonarCfg();
+            }
+            else
+            {
+                analyzer = new InfiniteRecursionRoslynCfg();
+            }
```
AraHaan, line 56: Why not shorten this down to:

```suggestion
            analyzer = !CFG.Roslyn.ControlFlowGraph.IsAvailable || runSonarAnalyzer ? new InfiniteRecursionSonarCfg() : new InfiniteRecursionRoslynCfg();
```
?

dmitry-golovinov-sonarsource: Good point. Please note that this is a draft, that means that I am just making it work and do no not clean the code yet, so code contains some duplicated/uncleaned code, since we want to verify what works and what does not.

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursionSonarCfg.cs
```diff
@@ -0,0 +1,245 @@
...
+namespace SonarAnalyzer.Rules.CSharp
+{
+    public class InfiniteRecursionSonarCfg : InfiniteRecursion.IInfiniteRecursion
+    {
+        public void CheckForNoExitProperty(SyntaxNodeAnalysisContext c, PropertyDeclarationSyntax property, IPropertySymbol propertySymbol)
+        {
+            IControlFlowGraph cfg;
+            if (property.ExpressionBody?.Expression != null)
+            {
+                if (CSharpControlFlowGraph.TryGet(property.ExpressionBody.Expression, c.SemanticModel, out cfg))
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Pls see above

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.cs
```diff
@@ -77,106 +88,27 @@ private static void CheckForNoExitProperty(SyntaxNodeAnalysisContext c)
...
-                var walker = new CfgWalkerForMethod(new RecursionAnalysisContext(cfg, symbol, identifier.GetLocation(), c));
-                walker.CheckPaths();
-                CheckInfiniteJumpLoop(body, cfg, "method", c);
+                analyzer.CheckForNoExitMethod(c, body, identifier, symbol);
             }
         }
 
-        private static void CheckInfiniteJumpLoop(SyntaxNode body, IControlFlowGraph cfg, string declarationType,
-            SyntaxNodeAnalysisContext analysisContext)
+        public class RecursionAnalysisContext<TControlFlowGraph>
```
pavel-mikula-sonarsource, line 103: Should not need to be public

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.cs
```diff
@@ -187,128 +119,10 @@ private class RecursionAnalysisContext
             }
         }
 
-        private class CfgWalkerForMethod : CfgRecursionSearcher
+        public interface IInfiniteRecursion
```
pavel-mikula-sonarsource, line 122: Should not need to be public

pavel-mikula-sonarsource (changes requested):
First round, big-picture only.
I didn't inspected the rule implementation yet

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.Roslyn.cs
```diff
@@ -0,0 +1,357 @@
...
+        {
+            var i = 10;
+            switch (i)
+            {
+                case 1:
+                    goto default;
+                case 2:
+                    break;
+                default:
+                    goto case 1; // Compliant FN
```
pavel-mikula-sonarsource, line 60: In general, this doesn't comply with the RSPEC so it's not compliant. It's just a FN
```suggestion
                    goto case 1; // FN
```

analyzers/src/SonarAnalyzer.CFG/Helpers/OperationWalker.cs
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+using System;
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+using StyleCop.Analyzers.Lightup;
+
+namespace SonarAnalyzer.CFG.Helpers
+{
+    public class OperationWalker
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Walker removed

analyzers/src/SonarAnalyzer.CFG/Helpers/OperationWalker.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        private readonly Func<IOperation, bool> anyVisitor;
+
+        public OperationWalker(Func<IOperation, bool> anyVisitor)
+        {
+            this.anyVisitor = anyVisitor;
+        }
+
+        public bool Visit(IOperation operation)
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Walker removed

analyzers/src/SonarAnalyzer.CFG/Helpers/OperationWalker.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            if (operation == null)
+            {
+                return false;
+            }
+            var queue = new Queue<IOperation>();
+            queue.Enqueue(operation);
+            while (queue.Count != 0 && queue.Dequeue() is var elem)
+            {
+                if (anyVisitor(elem))
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Walker removed

analyzers/src/SonarAnalyzer.CFG/Roslyn/BasicBlock.cs
```diff
@@ -63,6 +66,30 @@ public class BasicBlock
         public int Ordinal => ordinal.Value;
         public ImmutableArray<ControlFlowBranch> Predecessors => predecessors.Value;
 
+        public ImmutableArray<BasicBlock> SuccessorBlocks
+        {
+            get
+            {
+                if (successorBlocks == null)
+                {
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Code moved in another PR

analyzers/src/SonarAnalyzer.CFG/Roslyn/ControlFlowGraph.cs
```diff
@@ -49,6 +50,8 @@ public class ControlFlowGraph
         public ImmutableArray<IMethodSymbol> LocalFunctions => localFunctions.Value;
         public IOperation OriginalOperation => originalOperation.Value;
         public ControlFlowRegion Root => root.Value;
+        public BasicBlock EntryBlock => Blocks.First(x => x.Kind == BasicBlockKind.Entry);
+        public BasicBlock ExitBlock => Blocks.First(x => x.Kind == BasicBlockKind.Exit);
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Code moved in another PR

analyzers/tests/SonarAnalyzer.UnitTest/Rules/InfiniteRecursionTest.cs
```diff
@@ -29,9 +29,21 @@ namespace SonarAnalyzer.UnitTest.Rules
...
-        public void InfiniteRecursion() =>
+        public void InfiniteRecursionSonarCfg() =>
             Verifier.VerifyAnalyzer(
                 @"TestCases\InfiniteRecursion.cs",
+                new InfiniteRecursion(true),
+#if NETFRAMEWORK
+                ParseOptionsHelper.FromCSharp8,
+                NuGetMetadataReference.NETStandardV2_1_0);
+#else
+                ParseOptionsHelper.FromCSharp8);
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: That was not easy, since I had to change other test cases as well (like expression like returns, nullable stuff and so on). The reason - is that when you run this test locally it starts with C# 5

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.Roslyn.cs
```diff
@@ -0,0 +1,357 @@
...
+
+        void Test2()
+        {
+            var i = 10;
+            switch (i)
+            {
+                case 1:
+                    goto default;
+                default:
+                    goto case 1; // Compliant FN
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Let's deliver this in baby steps. I have no objection doing this when we have time in the end of the sprint.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.Roslyn.cs
```diff
@@ -0,0 +1,357 @@
...
+        }
+
+        void Foo()          // Compliant FN
+        {
+            Bar();
+            void Bar()      // Compliant FN
+            {
+                Baz();
+
+                void Baz()  // Noncompliant
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: The case you've stated is valid, though the case I've supplied in test, tests the ability to retrieve and "not fail" on nested functions.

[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Applied some renamings

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.cs
```diff
@@ -18,32 +18,43 @@
...
+        private readonly IInfiniteRecursion analyzer;
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        public static DiagnosticDescriptor Rule => DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
+
+        public InfiniteRecursion() : this(false) { }
+
+        public InfiniteRecursion(bool runSonarAnalyzer)
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Put what? I don't really understand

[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Ok, we agreed on the implementation details, done

analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
```diff
@@ -76,6 +86,8 @@ public void Initialize(AnalyzerOptions options)
 
             public HotspotConfiguration(IRuleLoader ruleLoader) => this.ruleLoader = ruleLoader;
 
+            public bool ForceSonarCfg => false;
```
dmitry-golovinov-sonarsource, line 89: Need to move up a bit

analyzers/tests/SonarAnalyzer.UnitTest/Rules/InfiniteRecursionTest.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: You could have temporarily updated your local test framework to use the latest version instead of the oldest for this purpose.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.Roslyn.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: This still isn't clear why it is noncompliant. Is this detected, because `NestedLocalFunction()`:
* Invoking `MethodWithNestedLocalFunctions();` causes the recursion?
* Invoking `LocalFunction();` causes the recursion?
* Invoking `NestedLocalFunction();` causes the recursion?
* Or all of them?

And if we break implementation for one of them, this test case will still pass?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.Roslyn.cs
```diff
@@ -0,0 +1,357 @@
...
+            void Foo()  // Noncompliant
+            {
+                Foo();
+            }
+        };
+
+        private void LamdaInIf(List<string> input)
+        {
+            if (input.Where(x =>
+            {
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Line 306-308 is the block from the `if` statement, they should be indented to the `if` statement. That illustrates the original confusion nicely :D

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.Roslyn.cs
pavel-mikula-sonarsource: These were fixed in Sonar flavor, please fix them in Roslyn one as well.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.RoslynCfg.cs
```diff
@@ -0,0 +1,398 @@
...
+        int Prop5 => Prop5; // Noncompliant
+
+        int Prop6
+        {
+            get => 42;
+        }
+
+        int Prop7
+        {
+            get => Prop6; // Compliant - FP
```
pavel-mikula-sonarsource, line 126: Wait, what? 😆 

Same in Sonar flavor.
```suggestion
            get => Prop6;
```

analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
```diff
@@ -41,10 +41,20 @@ public class AnalyzerConfiguration
             new HotspotConfiguration(new RuleLoader());
 
         public static IAnalyzerConfiguration AlwaysEnabled { get; } =
-            new AlwaysEnabledConfiguration();
+            new AlwaysEnabledConfiguration(false);
+
```
pavel-mikula-sonarsource, line 45: There were few improvements suggested on the other PR. I'm not sure which PR will make it first to be merged.
This is just a placeholder to resolve in the future.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/InfiniteRecursionTest.cs
```diff
@@ -29,9 +30,21 @@ namespace SonarAnalyzer.UnitTest.Rules
...
     {
         [TestMethod]
-        public void InfiniteRecursion() =>
+        public void InfiniteRecursion_SonarCfg() =>
             Verifier.VerifyAnalyzer(
-                @"TestCases\InfiniteRecursion.cs",
+                @"TestCases\InfiniteRecursion.SonarCfg.cs",
+                new InfiniteRecursion(AnalyzerConfiguration.AlwaysEnabledWithSonarCfg),
+#if NETFRAMEWORK
+                ParseOptionsHelper.OnlyCSharp7,
```
pavel-mikula-sonarsource, line 38: Why not `BeforeCSharp8` instead? We're trying to test all language versions because we can have surprises. It already happened that one token had one SyntaxKind in one version and another in newer version.

pavel-mikula-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.Roslyn.cs
dmitry-golovinov-sonarsource: What do you mean by that? I've applied the suggestion in Roslyn. In Sonar they were not fixed, but supported by the previous implementation since 2017.

analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
dmitry-golovinov-sonarsource: Fixed it in that PR. Leaving it untouched here to not produce conflicts for now

analyzers/tests/SonarAnalyzer.UnitTest/Rules/InfiniteRecursionTest.cs
dmitry-golovinov-sonarsource: Because it will fail, since a number of the cases will fail the compilation, among them there are issues to be raised.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.Roslyn.cs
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: I've delivered a number of easy cases, but still some gotos remain unsupported

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.Roslyn.cs
pavel-mikula-sonarsource: Roslyn test, lines 268 and 271 - I want to change `Compliant FN` comment only to `FN`, because the case doesn't comply with the RSPEC

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
```diff
@@ -0,0 +1,224 @@
...
+                {
+                    existReferenceToDeclaringSymbolMatcher = new BooleanBlockMatcher(operation =>
+                    {
+                        if (!(operation.Instance.Syntax is InvocationExpressionSyntax invocation))
+                        {
+                            return false;
+                        }
+
+                        return IsInstructionOnThisAndMatchesDeclaringSymbol(invocation.Expression, declaringSymbol, semanticModel);
+                    });
```
pavel-mikula-sonarsource, line 132: Can be simplified
```suggestion
                    existReferenceToDeclaringSymbolMatcher = new BooleanBlockMatcher(x =>
                        x.Instance.Syntax is InvocationExpressionSyntax invocation
                        && IsInstructionOnThisAndMatchesDeclaringSymbol(invocation.Expression, declaringSymbol, semanticModel));
```

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
```diff
@@ -0,0 +1,224 @@
...
+                {
+                    var expr = (ExpressionSyntax)node;
+                    var propertyAccess = expr.GetSelfOrTopParenthesizedExpression();
+                    var isNodeASet = propertyAccess.Parent is AssignmentExpressionSyntax assignment && assignment.Left == propertyAccess;
+                    return isNodeASet == isSet;
+                }
+            }
+
+            private abstract class CfgRecursionSearcher : CfgAllPathValidator
+            {
```
pavel-mikula-sonarsource, line 174: Both inheritors declare `existReferenceToDeclaringSymbolMatcher` and override `BlockHasReferenceToDeclaringSymbol` in exactly the same way. Let's make `Func<..> referenceToDeclaringSymbolMatcher` a constructor argument that will build a private field `referenceToDeclaringSymbolFinder` here.

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
```diff
@@ -0,0 +1,224 @@
...
+                {
+                    isSet = isSetAccessor;
+                    existReferenceToDeclaringSymbolMatcher = new BooleanBlockMatcher(operation =>
+                        operation.Instance.Syntax is var syntax
+                        && TypesForReference.Contains(syntax.GetType())
+                        && MatchesAccessor(syntax)
+                        && IsInstructionOnThisAndMatchesDeclaringSymbol(syntax, declaringSymbol, semanticModel));
+                }
+
+                private static readonly ISet<Type> TypesForReference = new HashSet<Type> { typeof(IdentifierNameSyntax), typeof(MemberAccessExpressionSyntax) };
```
pavel-mikula-sonarsource, line 159: Static field should be before constructor.

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
```diff
@@ -0,0 +1,224 @@
...
+                    : base(
+                        context.ControlFlowGraph,
+                        context.AnalyzedSymbol,
+                        context.SemanticModel,
+                        () => context.AnalysisContext.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.IssueLocation, reportOn)))
+                {
+                    isSet = isSetAccessor;
+                    existReferenceToDeclaringSymbolMatcher = new BooleanBlockMatcher(operation =>
+                        operation.Instance.Syntax is var syntax
+                        && TypesForReference.Contains(syntax.GetType())
```
pavel-mikula-sonarsource, line 154: This looks heavy. Can we use `.IsAnyKind()` instead?

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
```diff
@@ -0,0 +1,224 @@
...
+                public CfgWalkerForProperty(RecursionAnalysisContext<ControlFlowGraph> context, string reportOn, bool isSetAccessor)
+                    : base(
+                        context.ControlFlowGraph,
+                        context.AnalyzedSymbol,
+                        context.SemanticModel,
+                        () => context.AnalysisContext.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, context.IssueLocation, reportOn)))
+                {
+                    isSet = isSetAccessor;
+                    existReferenceToDeclaringSymbolMatcher = new BooleanBlockMatcher(operation =>
+                        operation.Instance.Syntax is var syntax
```
pavel-mikula-sonarsource, line 153: It makes it easier to identify that it's a lambda arg
```suggestion
                    existReferenceToDeclaringSymbolMatcher = new BooleanBlockMatcher(x =>
                        x.Instance.Syntax is var syntax
```

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
```diff
@@ -0,0 +1,224 @@
...
+
+                private static readonly ISet<Type> TypesForReference = new HashSet<Type> { typeof(IdentifierNameSyntax), typeof(MemberAccessExpressionSyntax) };
+
+                protected override bool BlockHasReferenceToDeclaringSymbol(BasicBlock block) =>
+                    existReferenceToDeclaringSymbolMatcher.Matches(block);
+
+                private bool MatchesAccessor(SyntaxNode node)
+                {
+                    var expr = (ExpressionSyntax)node;
+                    var propertyAccess = expr.GetSelfOrTopParenthesizedExpression();
```
pavel-mikula-sonarsource, line 167: I'd collapse them, we don't need to store `expr`.
```suggestion
                    var propertyAccess = ((ExpressionSyntax)node).GetSelfOrTopParenthesizedExpression();
```

pavel-mikula-sonarsource (changes requested):
Almost full review. One more will follow in a minute (start with that one, as it might make many comments here out of date)

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
```diff
@@ -0,0 +1,224 @@
...
+                {
+                    c.ReportDiagnosticWhenActive(Diagnostic.Create(Rule, identifier.GetLocation(), "method's recursion"));
+                }
+            }
+
+            private class CfgWalkerForMethod : CfgRecursionSearcher
+            {
+                private readonly BooleanBlockMatcher existReferenceToDeclaringSymbolMatcher;
+
+                public CfgWalkerForMethod(RecursionAnalysisContext<ControlFlowGraph> context)
```
pavel-mikula-sonarsource, line 117: These two walkers are very syntax-based. But what they do as a core functionality is `BlockHasReferenceToDeclaringSymbol`. That one should be easy to rewrite using operation-based approach that could simplify a lot of stuff around here.

pavel-mikula-sonarsource (changes requested):
In the previous review, I went through the code and asked for some simplifications. 

After that, I saw this question below that might simplify stuff a lot.

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.cs
```diff
@@ -18,32 +18,36 @@
...
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        private static DiagnosticDescriptor Rule => DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
+
+        public InfiniteRecursion() : this(AnalyzerConfiguration.AlwaysEnabled) { }
+
+        internal /* for testing */ InfiniteRecursion(IAnalyzerConfiguration configuration) =>
+            analyzer = (CFG.Roslyn.ControlFlowGraph.IsAvailable && !configuration.ForceSonarCfg)
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: I would rather do this in another PR. Will do it right after processing of all the changes here

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_SonarCfg.cs
```diff
@@ -0,0 +1,208 @@
...
+                    block.Instructions.Any(i =>
+                        TypesForReference.Contains(i.GetType())
+                        && MatchesAccessor(i)
+                        && IsInstructionOnThisAndMatchesDeclaringSymbol(i, declaringSymbol, semanticModel));
+
+                private bool MatchesAccessor(SyntaxNode node)
+                {
+                    var expr = (ExpressionSyntax)node;
+                    var propertyAccess = expr.GetSelfOrTopParenthesizedExpression();
+                    var isNodeASet = propertyAccess.Parent is AssignmentExpressionSyntax assignment && assignment.Left == propertyAccess;
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Yes, it was intentionally removed, since this code was not callable. I believe that this is an artifact after last refactoring.

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
```diff
@@ -0,0 +1,224 @@
...
+                    SyntaxKindEx.LocalFunctionStatement, SyntaxKind.SimpleLambdaExpression, SyntaxKind.AnonymousMethodExpression, SyntaxKind.ParenthesizedLambdaExpression
+                };
+                if (c.Node.IsKind(SyntaxKindEx.LocalFunctionStatement))
+                {
+                    // we need to go up and track all possible enclosing local function statements
+                    foreach (var enclosingFunction in c.Node.Ancestors().Where(x => enclosingKinds.Contains(x.Kind())).Reverse())
+                    {
+                        if (enclosingFunction.IsAnyKind(SyntaxKindEx.LocalFunctionStatement))
+                        {
+                            cfg = cfg.GetLocalFunctionControlFlowGraph(c.SemanticModel.GetDeclaredSymbol(enclosingFunction) as IMethodSymbol);
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: I was not able to reproduce that. We can add a code to return if CFG is null

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
dmitry-golovinov-sonarsource: Nope, since MemberAccessExpressionSyntax has no matching syntax kind

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
```diff
@@ -0,0 +1,224 @@
...
+            {
+                var operation = c.SemanticModel.GetOperation(body.Parent);
+                var wrapper = new IOperationWrapperSonar(operation);
+                while (wrapper.Parent != null)
+                {
+                    wrapper = new IOperationWrapperSonar(wrapper.Parent);
+                }
+
+                var cfg = ControlFlowGraph.Create(wrapper.Instance.Syntax, c.SemanticModel);
+                var enclosingKinds = new HashSet<SyntaxKind>
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: This method covers only the case when node is local function, all the rest are not covered at all, since they are not point of this code. I am okay doing this when this will be really usable and testable. Or as a bonus in this sprint when we have time.

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.cs
```diff
@@ -77,106 +81,44 @@ private static void CheckForNoExitProperty(SyntaxNodeAnalysisContext c)
...
             }
-        }
 
-        #region CFG walkers for call recursion
+            var assignedSymbol = semanticModel.GetSymbolInfo(name).Symbol;
+            return declaringSymbol.Equals(assignedSymbol);
+        }
 
-        private class RecursionAnalysisContext
+        private class RecursionAnalysisContext<TControlFlowGraph>
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Please make it a `private readonly` field instead of private property. There's no practical reason for it to be a property.

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
pavel-mikula-sonarsource: This was done just halfway. I wanted `RecursionSearcher` to have constructor like this. That should remove some duplications.
```C#
protected RecursionSearcher(RecursionContext<ControlFlowGraph> context, Func<IOperationWrapperSonar, bool> referenceToDeclaringSymbolMatcher)
{
   // ...
   referenceToDeclaringSymbolFinder = new BooleanBlockFinder(referenceToDeclaringSymbolMatcher);
}
```

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
pavel-mikula-sonarsource: It has. `SyntaxKind.SimpleMemberAccessExpression` and `SyntaxKind.PointerMemberAccessExpression` as seen here
https://github.com/dotnet/roslyn/blob/0700dcde657135078bf8c869129a64a266e110cf/src/Compilers/CSharp/Portable/Syntax/Syntax.xml#L556-L559

pavel-mikula-sonarsource: But I'm resolving this, as it went away anyway.

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Added 2 protective ifs

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
dmitry-golovinov-sonarsource: ??? There is no BooleanBlockFinder now. This was removed while switching to operations

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
```diff
@@ -0,0 +1,179 @@
...
+                        if (ProcessOperation(operation, cfg, out var result))
+                        {
+                            return result;
+                        }
+                    }
+                    return false;
+
+                    bool ProcessOperation(IOperation operation, ControlFlowGraph controlFlowGraph, out bool result)
+                    {
+                        foreach (var child in operation.DescendantsAndSelf().ToReversedExecutionOrder())
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Or maybe something like this one level higher, but that might change the semantics.
```
foreach (var operation in block.OperationsAndBranchValue.ToReversedExecutionOrder())
```
or
```
foreach (var operation in block.OperationsAndBranchValue.Reverse().ToReversedExecutionOrder())
```

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
```diff
@@ -0,0 +1,179 @@
...
+                            if (child.Instance.Kind == OperationKindEx.FlowAnonymousFunction)
+                            {
+                                var anonymousFunctionCfg = controlFlowGraph.GetAnonymousFunctionControlFlowGraph(IFlowAnonymousFunctionOperationWrapper.FromOperation(child.Instance));
+                                foreach (var subOperation in anonymousFunctionCfg.Blocks.SelectMany(x => x.OperationsAndBranchValue).SelectMany(x => x.DescendantsAndSelf()))
+                                {
+                                    if (ProcessOperation(subOperation, anonymousFunctionCfg, out result))
+                                    {
+                                        return true;
+                                    }
+                                }
```
pavel-mikula-sonarsource, line 165: _Thinking out loud:_ This part is very similar to `anonymousFunctionCfg.Blocks.Any(FindReferenceToSelf)`, it just returns a little different result. 

Can we unify them somehow? It's complicated so I'd like to have it just once there.

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
```diff
@@ -0,0 +1,179 @@
...
+
+                    return false;
+                }
+
+                protected override bool IsValid(BasicBlock block) =>
+                    FindReferenceToSelf(block);
+
+                protected override bool IsInvalid(BasicBlock block) => false;
+
+                private bool FindReferenceToSelf(BasicBlock block)
```
pavel-mikula-sonarsource, line 141: While looking at the xor logic and `bool` result, this name doesn't help too much with the understanding. Can it be improved somehow?

Mechanically, let's start the discussion with this proposal. But I feel that we should try to explain what kind of reference. Probably not read, not write. Maybe something like useful reference (don't like it, I can't find the word)
```suggestion
                private bool ContainsReferenceToSelf(BasicBlock block)
```

pavel-mikula-sonarsource (changes requested):
Another round. 
There are also few open conversations above to address (some are waiting for rebasing, but some needs change)

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
```diff
@@ -0,0 +1,179 @@
...
+                    foreach (var operation in block.OperationsAndBranchValue)
+                    {
+                        if (ProcessOperation(operation, cfg, out var result))
+                        {
+                            return result;
+                        }
+                    }
+                    return false;
+
+                    bool ProcessOperation(IOperation operation, ControlFlowGraph controlFlowGraph, out bool result)
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: This code got refactored, phew

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
dmitry-golovinov-sonarsource: Part was not needed and got removed

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
dmitry-golovinov-sonarsource: We no longer need it, since we've embedded the logic to IsValid method

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion_RoslynCfg.cs
pavel-mikula-sonarsource: I was reviewing per commit :)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.Roslyn.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: This one wasn't addressed

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.RoslynCfg.cs
```diff
@@ -0,0 +1,504 @@
...
+            {
+                throw new ArgumentException();
+            }
+            A:
+            goto A;
+
+            return 42;
+        }
+
+        int ThrowsInTry(int a)
```
pavel-mikula-sonarsource, line 442: Let's add something like this, just to document the case because finally blocks are also special.
```suggestion
        int ThrowsInFinally(int a)
        {
        	try
        	{
        	}
        	finally
        	{
        		throw new ArgumentException();
        	}
        }
```

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
```diff
@@ -0,0 +1,151 @@
...
+                        }
+                        else
+                        {
+                            var operationWrapper = cfg.FlowAnonymousFunctionOperations().Single(x => x.WrappedOperation.Syntax == enclosingFunction);
+                            cfg = cfg.GetAnonymousFunctionControlFlowGraph(operationWrapper);
+                        }
+                        if (cfg == null)
+                        {
+                            return;
+                        }
```
pavel-mikula-sonarsource, line 84: The null should not happen indeed, we can remove it. See
https://github.com/dotnet/roslyn/blob/6a9d04542e4df19817e2526d1f68ec7687f93ce9/src/Compilers/Core/Portable/Operations/ControlFlowGraph.cs#L285-L288

https://github.com/dotnet/roslyn/blob/6a9d04542e4df19817e2526d1f68ec7687f93ce9/src/Compilers/Core/Portable/Operations/ControlFlowGraph.cs#L236-L239

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
```diff
@@ -0,0 +1,151 @@
...
+                    }
+
+                    cfg = cfg.GetLocalFunctionControlFlowGraph(symbol as IMethodSymbol);
+                    if (cfg == null)
+                    {
+                        return;
+                    }
```
pavel-mikula-sonarsource, line 91: Could be removed indeed

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
```diff
@@ -0,0 +1,151 @@
...
+                public RecursionSearcher(RecursionContext<ControlFlowGraph> context, bool isGetAccesor = true)
+                    : base(context.ControlFlowGraph)
+                {
+                    this.context = context;
+                    this.isGetAccesor = isGetAccesor;
+                }
+
+                public void CheckPaths()
+                {
+                    if (CheckAllPaths() || CfgHasUnescapableLoop())
```
pavel-mikula-sonarsource, line 113: Nice simplification!

_Minor performance:_ The loop part seems cheaper than diving deep into operations. But it will not help in too many cases.
```suggestion
                    if (CfgHasUnescapableLoop() || CheckAllPaths())
```

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
```diff
@@ -0,0 +1,151 @@
...
+                    {
+                        context.ReportIssue();
+                    }
+                }
+
+                protected override bool IsValid(BasicBlock block)
+                {
+                    foreach (var operation in block.OperationsAndBranchValue)
+                    {
+                        foreach (var child in operation.ToReversedExecutionOrder())
```
pavel-mikula-sonarsource, line 123: Isn't this the same as single loop like this? I'm not 100% sure
```suggestion
                    foreach (var operation in block.OperationsAndBranchValue.Reversed().ToReversedExecutionOrder())
```

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
```diff
@@ -0,0 +1,151 @@
...
+                            OperationKindEx.Invocation => IInvocationOperationWrapper.FromOperation(operation).TargetMethod,
+                            _ => null
+                        };
+                }
+
+                protected override bool IsInvalid(BasicBlock block) => false;
+
+                private bool CfgHasUnescapableLoop() =>
+                    !context.ControlFlowGraph.ExitBlock.IsReachable
+                    && !context.ControlFlowGraph.Blocks.Any(x => x.FallThroughSuccessor?.Semantics == ControlFlowBranchSemantics.Throw && x.IsReachable);
```
pavel-mikula-sonarsource, line 147: `Un*` + `!` + `!` = too many negations. Let's make it positive. Something like `CfgCanExit`

pavel-mikula-sonarsource (changes requested):
Final polishing. What's the merging plan with the other PR? This is still a draft.

dmitry-golovinov-sonarsource:
Take action on https://github.com/SonarSource/sonar-dotnet/pull/4882#discussion_r710172411 before merging that one

dmitry-golovinov-sonarsource:
The plan is to merge another branch, rebase this one and then merge it with all suggested changes

pavel-mikula-sonarsource (changes requested):
This one wasn't addressed yet https://github.com/SonarSource/sonar-dotnet/pull/4855#discussion_r700927884

It's the last change that should be resolved before rebase.

I have no new comments 🎉

dmitry-golovinov-sonarsource:
Take action on https://github.com/SonarSource/sonar-dotnet/pull/4882#discussion_r712184812 before merging

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.Roslyn.cs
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: It is non-compliant since  NestedLocalFunction calls itself recursively.

Tried to make FN more understandable in a separate example.

analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
dmitry-golovinov-sonarsource: Rebased

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.cs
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Rebased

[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Fixed the code

analyzers/src/SonarAnalyzer.CFG/Extensions/IOperationExtensions.cs
```diff
@@ -36,9 +36,22 @@ public static class IOperationExtensions
...
+                wrapper = new IOperationWrapperSonar(wrapper.Parent);
+            }
+            return wrapper.Instance;
+        }
+
         public static OperationExecutionOrder ToExecutionOrder(this IEnumerable<IOperation> operations) =>
             new OperationExecutionOrder(operations, false);
 
+        public static OperationExecutionOrder ToReversedExecutionOrder(this IOperation operation) =>
+            new[] { operation }.ToReversedExecutionOrder();
```
pavel-mikula-sonarsource, line 53: _Coverage:_  Unused, can be removed

pavel-mikula-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.CFG/Extensions/IOperationExtensions.cs
pavel-mikula-sonarsource: I'll remove it in my PR

Issue #1353, opened 2018-04-24
Update S2699: to support NFluent


agigleux:
Once implemented, https://jira.sonarsource.com/browse/RSPEC-3413 should be amended to list NFluent, Shoudly and FluentAssertions 5.x

- [x]  NFluent (http://www.n-fluent.net/) Done in #4862
- [x] Shouldly (https://github.com/shouldly/shouldly) Support was already there, tested in #4863
- [x] FluentAssertions 5.x => Done in #2715
- [x] [NSubstitute](https://nsubstitute.github.io/) (from [community](https://community.sonarsource.com/t/how-to-make-sonarqube-recognize-nsubstitute-assertions/13432)) => Done in  #2709

costin-zaharia-sonarsource:
Added support for NSubstitute (4.2.1) in #2709

costin-zaharia-sonarsource:
Updated the codebase to support FluentAssertions 5 in #2715

andrei-epure-sonarsource:
Shouldly is already supported implicitly (we consider any method containing "Should" to be an assertion).

agigleux:
Can you update the description of the rule to make it clear which UT Framework is supported?

https://rules.sonarsource.com/csharp/RSPEC-2699 is not listing NFluent, Shoudly for example.

csaba-sagi-sonarsource:
> @csaba-sagi-sonarsource Can you update the description of the rule to make it clear which UT Framework is supported?
> 
> https://rules.sonarsource.com/csharp/RSPEC-2699 is not listing NFluent, Shoudly for example.

Thank you @agigleux for pointing it out.
The description got updated in #4867, so I am closing the issue once more :)

PR #4862, opened 2021-09-01
S2699 NFluent support

csaba-sagi-sonarsource:
Fixes NFluent part of #1353

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
@@ -53,13 +53,16 @@ namespace SonarAnalyzer.Rules.CSharp
         private static readonly ImmutableArray<KnownType> KnownAssertionTypes = ImmutableArray.Create(
                 KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_Assert,
                 KnownType.NUnit_Framework_Assert,
-                KnownType.Xunit_Assert);
+                KnownType.Xunit_Assert,
+                KnownType.NFluent_Check);
 
         private static readonly ImmutableArray<KnownType> KnownAsertionExceptionTypes = ImmutableArray.Create(
             KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_AssertFailedException,
             KnownType.NUnit_Framework_AssertionException,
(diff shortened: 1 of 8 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldContainAssertion.cs
```diff
@@ -53,13 +53,16 @@ public sealed class TestMethodShouldContainAssertion : SonarDiagnosticAnalyzer
         private static readonly ImmutableArray<KnownType> KnownAssertionTypes = ImmutableArray.Create(
                 KnownType.Microsoft_VisualStudio_TestTools_UnitTesting_Assert,
                 KnownType.NUnit_Framework_Assert,
-                KnownType.Xunit_Assert);
+                KnownType.Xunit_Assert,
+                KnownType.NFluent_Check);
```
andrei-epure-sonarsource, line 57: nitpick: alphabetical order

analyzers/tests/SonarAnalyzer.UnitTest/Rules/TestMethodShouldContainAssertionTest.cs
```diff
@@ -163,6 +169,29 @@ public void ShouldNotThrowTest()
...
+using NFluent;
+using NUnit.Framework;
+
+[TestFixture]
+public class Foo
+{
+    [Test]
+    public void Test1()
+    {
+        var x = 42;
```
andrei-epure-sonarsource, line 185: you could even remove lines 185-187 and 189 and just keep the minimum necessary code for the test

(applies to all)

analyzers/tests/SonarAnalyzer.UnitTest/Rules/TestMethodShouldContainAssertionTest.cs
```diff
@@ -38,6 +38,12 @@ private static class FluentAssertionVersions
             public const string Ver5 = "5.9.0";
         }
 
+        private static class NFluentVersions
```
andrei-epure-sonarsource, line 41: where are you using these? you're only using `Ver1` for legacy, and in the rest you are using the default `NuGetLatestVersion`, right?

Do you need this static class at all, then?

andrei-epure-sonarsource (changes requested):

PR #4863, opened 2021-09-01
S2699: add tests with Shouldly

andrei-epure-sonarsource:
Shouldly is already supported by the UnitTestHelper KnownAssertionMethodParts - https://github.com/SonarSource/sonar-dotnet/blob/8.27.0.35380/analyzers/src/SonarAnalyzer.Common/Helpers/UnitTestHelper.cs#L60 - all methods contain "Should" .

I don't want to add explicit support because it would be additional logic with no functional gain.

Related to #1353

Might have conflicts with #4862 (rebase will be needed)

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.NUnit.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.NUnit.cs
@@ -2,11 +2,13 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Threading;
     using System.Threading.Tasks;
     using FluentAssertions;
     using NFluent;
     using NSubstitute;
     using NUnit.Framework;
+    using Shouldly;
 
     using static NUnit.Framework.Assert;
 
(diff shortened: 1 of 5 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.NUnit.cs
```diff
@@ -519,6 +521,53 @@ public void ReceivedInOrder()
...
+            empty.ShouldBeEmpty();
+
+            var dict = new Dictionary<string, string> { { "x", "x" } };
+            dict.ShouldContainKey("y");
+
+            Should.Throw<IndexOutOfRangeException>(() => Bar());
+        }
+
+        [TestCase]
+        public void X() => "Foo".ShouldMatchApproved();
```
andrei-epure-sonarsource, line 549: I am testing each because they are extension methods defined in separate static classes, so we need good coverage given we rely on the "Should" string and the logic may change in the future

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.NUnit.cs
```diff
@@ -519,6 +521,53 @@ public void ReceivedInOrder()
...
+        public void Y() => i.ShouldBeInRange(30000, 40000);
+
+        [TestCase]
+        public void Z() => Should.NotThrow(() => Bar());
+
+        [TestCase]
+        public void Dynamic()
+        {
+            dynamic theFuture = Foo();
+            DynamicShould.HaveProperty(theFuture, "X");
```
andrei-epure-sonarsource, line 561: this method doesn't start with "should" but the logic actually splits the full expression in words and verifies if any is "should"

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/TestMethodShouldContainAssertion.NUnit.cs
```diff
@@ -519,6 +521,53 @@ public void ReceivedInOrder()
...
     }
 
+    // All assert methods start with "Should*". We do not test all APIs, just some.
+    public class ShouldlyTests
+    {
+        int i = 1;
+        public dynamic Foo() => null;
+        public void Bar() { }
+
+        [TestCase]
```
csaba-sagi-sonarsource, line 531: We don`t need this NoAssert case, as it is already tested and has nothing specific to Shoudly.

csaba-sagi-sonarsource (approved):
Looks good, one minor suggestion to handle and also a rebase to be done.

PR #4866, opened 2021-09-02
Add LVA UTs

pavel-mikula-sonarsource:
We need to cover the existing LVA with isolated UTs.

This is a draft for now to simplify the review, more commits will arrive for:
- [x] ProcessSimpleAssignment*
- [x] ProcessVariableDeclarator*
- [x] ProcessVariableInForeach*
- [x] More branched scenario


analyzers/tests/SonarAnalyzer.UnitTest/LiveVariableAnalysis/SonarLiveVariableAnalysisTest.cs
```diff
@@ -0,0 +1,585 @@
...
+        }
+
+        [TestMethod]
+        public void StaticLocalFunction_NotLiveIn()
+        {
+            var code = @"
+outParameter = LocalFunction(0);
+static int LocalFunction(int a)
+{
+    return 42
```
dmitry-golovinov-sonarsource, line 483: Does this even compile?

dmitry-golovinov-sonarsource (approved):
LGTM! See my minors

PR #4869, opened 2021-09-03
OperationWalker, BasicBlockWalker, CfgAllPathValidator,supplementary properties for Roslyn CFG

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.CFG/Helpers/OperationVisitor.cs
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Microsoft.CodeAnalysis;
+
+namespace SonarAnalyzer.CFG.Helpers
+{
+    public abstract class OperationVisitor<TResult>
```
pavel-mikula-sonarsource, line 25: This is currently only an interface and if I delete it, nothing really happens => we don't need it. What I meant by `OperationVisitor` is currently implemented in `OperationWalker`

analyzers/src/SonarAnalyzer.CFG/Helpers/OperationWalker.cs
```diff
@@ -0,0 +1,66 @@
...
+        protected abstract void VoidVisitOperation(IOperation operation);
+
+        protected override object VisitOperation(IOperation operation)
+        {
+            VoidVisitOperation(operation);
+            return null;
+        }
+    }
+
+    public abstract class OperationWalker<TResult> : OperationVisitor<TResult>
```
pavel-mikula-sonarsource, line 38: I tried to improve the code and then I realized that this is actually `OperationFinder`, because it finds first and returns that. So let's write these in a different way:

First general walker:
```
    public abstract class OperationWalker
    {
        protected abstract void DefaultVisit(IOperationWrapperSonar operation);

        protected virtual bool VisitChildren(IOperationWrapperSonar operation) => true;

        public void Visit(IOperationWrapperSonar operation) =>
            Visit(operation.Instance);

        public void Visit(IOperation operation)
        {
            if (operation == null)
            {
                return;
            }
            var queue = new Queue<IOperation>();
            queue.Enqueue(operation);
            while (queue.Any())
            {
                var current = new IOperationWrapperSonar(queue.Dequeue());
                DefaultVisit(current);
                if (VisitChildren(current))
                {
                    foreach (var child in current.Children)
                    {
                        queue.Enqueue(child);
                    }
                }
            }
        }
    }
```
Then finder on top of it:
```
    public abstract class OperationFinder<TResult> : OperationWalker
        where TResult : class
    {
        private TResult result;

        protected abstract TResult ReturnValue(IOperationWrapperSonar operation);

        public TResult FindFirst(IOperation operation)
        {
            result = null;
            Visit(operation);
            return result;
        }

        protected override bool VisitChildren(IOperationWrapperSonar operation) =>
            result == null;

        protected override void DefaultVisit(IOperationWrapperSonar operation) =>
            result ??= ReturnValue(operation);
    }
```
This is not thread-safe, but we can do this if we need to:
```
    public abstract class OperationFinder<TResult> : OperationWalker
        where TResult : class
    {
        private readonly IOperationWrapperSonar root;
        private TResult result;

        protected abstract TResult ReturnValue(IOperationWrapperSonar operation);

        protected OperationFinder(IOperationWrapperSonar root) =>
            this.root = root;

        public TResult FindFirst()
        {
            Visit(root);
            return result;
        }

        protected override bool VisitChildren(IOperationWrapperSonar operation) =>
            result == null;

        protected override void DefaultVisit(IOperationWrapperSonar operation) =>
            result ??= ReturnValue(operation);
    }
```

pavel-mikula-sonarsource (changes requested):
We'll need some UTs :)

analyzers/src/SonarAnalyzer.CFG/Roslyn/CfgAllPathValidator.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            if (block.SuccessorBlocks.Contains(this.cfg.ExitBlock) ||
+                !block.SuccessorBlocks.Except(this.alreadyVisitedBlocks).Any())
+            {
+                return false;
+            }
+
+            return block.SuccessorBlocks
+                        .Except(this.alreadyVisitedBlocks)
+                        .All(b => IsBlockValidWithSuccessors(b));
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: You cannot just drop the condition "&& block.SuccessorBlocks.Except(visited).Any()",
imagine that block.SuccessorBlocks == visited and does not contain ExitBlock, then original condition would compute
!false && false -> true && false -> false
and your would compute
!false && !false -> true && true -> true  (block.SuccessorBlocks.Except(visited) empty and Any would return false)

Am I wrong?

analyzers/src/SonarAnalyzer.CFG/Roslyn/ControlFlowGraph.cs
```diff
@@ -72,6 +76,8 @@ private ControlFlowGraph(object instance)
             localFunctions = LocalFunctionsProperty.ReadImmutableArray<IMethodSymbol>(instance);
             originalOperation = OriginalOperationProperty.ReadValue<IOperation>(instance);
             root = RootProperty.ReadValue(instance, ControlFlowRegion.Wrap);
+            Debug.Assert(Blocks.Single(x => x.Kind == BasicBlockKind.Entry) == EntryBlock);
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Indeed

analyzers/src/SonarAnalyzer.CFG/Helpers/OperationWalker.cs
dmitry-golovinov-sonarsource: OperationWalker you've suggested does not stop on the first found operation, instead it will continue to process what's left in the queue. After a long brainstorming session with @csaba-sagi-sonarsource we've found another interesting approach.

dmitry-golovinov-sonarsource: But thanks about wrapper inspirations!

analyzers/src/SonarAnalyzer.CFG/Roslyn/CfgAllPathValidator.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: damn

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/ControlFlowBranchTest.cs
```diff
@@ -37,10 +41,17 @@ public class Sample
...
-            var exit = cfg.Blocks[1];
+            var entry = cfg.EntryBlock;
+            var exit = cfg.ExitBlock;
             entry.Kind.Should().Be(BasicBlockKind.Entry);
+            entry.Operations.Should().BeEmpty();
+            entry.BranchValue.Should().BeNull();
+            entry.OperationsAndBranchValue.Should().BeEmpty();
             exit.Kind.Should().Be(BasicBlockKind.Exit);
+            exit.Operations.Should().BeEmpty();
+            exit.Operations.Should().BeEmpty();
```
pavel-mikula-sonarsource, line 52: Duplicate lines
```suggestion
            exit.Operations.Should().BeEmpty();
```

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/ControlFlowBranchTest.cs
```diff
@@ -37,10 +41,17 @@ public class Sample
...
 }";
             var cfg = TestHelper.CompileCfg(code);
-            var entry = cfg.Blocks[0];
-            var exit = cfg.Blocks[1];
+            var entry = cfg.EntryBlock;
+            var exit = cfg.ExitBlock;
             entry.Kind.Should().Be(BasicBlockKind.Entry);
+            entry.Operations.Should().BeEmpty();
+            entry.BranchValue.Should().BeNull();
+            entry.OperationsAndBranchValue.Should().BeEmpty();
```
pavel-mikula-sonarsource, line 49: This new lines shouldn't be here. It should be in `BasicBlockTest` file.

These assertions are here just to make sure we've picked up correct `Entry` and `Exit` blocks. But with the new properties `EntryBlock` and `ExitBlock` even these assertions can be deleted.

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/ControlFlowBranchTest.cs
```diff
@@ -96,10 +107,12 @@ public void Method(bool condition)
              *            v
              *         Exit 4
              */
+            var entryBlock = cfg.EntryBlock;
             var initBlock = cfg.Blocks[1];
             var tryBlock = cfg.Blocks[2];
             var finallyBlock = cfg.Blocks[3];
-            var exitBlock = cfg.Blocks[4];
+            var exitBlock = cfg.ExitBlock;
+            entryBlock.Kind.Should().Be(BasicBlockKind.Entry);
```
pavel-mikula-sonarsource, line 115: This should not be needed. `entryBlock` isn't used to assert branches below and we can remove it.
```suggestion
```

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/ControlFlowBranchTest.cs
```diff
@@ -128,5 +141,51 @@ public void Method(bool condition)
             entering.LeavingRegions.Should().BeEmpty();
             entering.FinallyRegions.Should().BeEmpty();
         }
+
+        [TestMethod]
+        public void ValidateOperations()
```
pavel-mikula-sonarsource, line 146: This test should be in `BasicBlockTests` file.

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/ControlFlowBranchTest.cs
```diff
@@ -128,5 +141,51 @@ public void Method(bool condition)
...
+            var cfg = TestHelper.CompileCfg(code);
+            var entry = cfg.EntryBlock;
+            var exit = cfg.ExitBlock;
+            var body = cfg.Blocks[1];
+            entry.Kind.Should().Be(BasicBlockKind.Entry);
+            entry.Operations.Should().BeEmpty();
+            exit.Kind.Should().Be(BasicBlockKind.Exit);
+            exit.Operations.Should().BeEmpty();
+            body.Kind.Should().Be(BasicBlockKind.Block);
+            body.Operations.Length.Should().Be(1);
```
pavel-mikula-sonarsource, line 166:
```suggestion
            body.Operations.Should().HaveCount(1);
```

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/ControlFlowBranchTest.cs
```diff
@@ -128,5 +141,51 @@ public void Method(bool condition)
...
+            body.Operations[0].Syntax.Kind().Should().Be(SyntaxKind.ExpressionStatement);
+            body.BranchValue.Syntax.Kind().Should().Be(SyntaxKind.NumericLiteralExpression);
```
pavel-mikula-sonarsource, line 168: We don't really care about the syntax, we just need correct operations.
```suggestion
            body.Operations[0].Should().BeAssignableTo<I...Something...Operation>();
            body.BranchValue.Should().BeAssignableTo<I...Something...Operation>();
```

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/ControlFlowBranchTest.cs
```diff
@@ -128,5 +141,51 @@ public void Method(bool condition)
...
+            body.OperationsAndBranchValue.Should().OnlyContainInOrder(body.Operations[0], body.BranchValue);
+        }
+
+        [TestMethod]
+        public void ValidateAnonymousFunctionFinder()
+        {
+            const string code = @"
+using System;
+public class Sample {
+    private Action<int> Simple()
```
pavel-mikula-sonarsource, line 178: _Optional:_ In this simple & single usage cases, I prefer full name to save line that is not important for the test case.
```suggestion
public class Sample {
    private System.Action<int> Simple()
```

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/ControlFlowBranchTest.cs
```diff
@@ -128,5 +141,51 @@ public void Method(bool condition)
...
+            exit.Operations.Should().BeEmpty();
+            body.Kind.Should().Be(BasicBlockKind.Block);
+            body.Operations.Length.Should().Be(1);
+            body.Operations[0].Syntax.Kind().Should().Be(SyntaxKind.ExpressionStatement);
+            body.BranchValue.Syntax.Kind().Should().Be(SyntaxKind.NumericLiteralExpression);
+            body.OperationsAndBranchValue.Should().OnlyContainInOrder(body.Operations[0], body.BranchValue);
+        }
+
+        [TestMethod]
+        public void ValidateAnonymousFunctionFinder()
```
pavel-mikula-sonarsource, line 173: This test should be in `RoslynControlFlowGraphTests`
Let's rename it to `FlowAnonymousFunctionOperations_FindsAll` and add two functions inside.

analyzers/src/SonarAnalyzer.CFG/Helpers/OperationWalker.cs
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using StyleCop.Analyzers.Lightup;
+
+namespace SonarAnalyzer.CFG.Helpers
+{
+    public abstract class OperationWalker : OperationWalker<object>
```
pavel-mikula-sonarsource, line 28: What is this class for and how do I use it without reimplementing the generic one? It doesn't make sense to me.

Let's just remove it. We can have `protected virtual void DefaultVisit()` in the generic one below.

analyzers/src/SonarAnalyzer.CFG/Helpers/OperationWalker.cs
```diff
@@ -0,0 +1,69 @@
...
+            foreach (var operation in operations)
+            {
+                queue.Enqueue(operation);
+                while (queue.Any())
+                {
+                    var wrapper = new IOperationWrapperSonar(queue.Dequeue());
+                    if (!VisitOperation(wrapper))
+                    {
+                        return Result;
+                    }
```
pavel-mikula-sonarsource, line 57: This is now based on the same side effect as the previous one. 
This is protected & `Result` is protected & the result is a side effect => it doesn't make sense to return the result here. Inheritor has to set `Result` and needs to return the result by himself.

We can have this as `void` without the `Result` part. And have second `Finder` class to be explicit about its purpose.

analyzers/src/SonarAnalyzer.CFG/Helpers/BasicBlockWalker.cs
```diff
@@ -0,0 +1,37 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarAnalyzer.CFG.Roslyn;
+
+namespace SonarAnalyzer.CFG.Helpers
+{
+    public abstract class BasicBlockWalker : OperationWalker
```
pavel-mikula-sonarsource, line 25: Will you need this one? I will not. So we might remove it.

pavel-mikula-sonarsource (changes requested):
Round 2

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/ControlFlowBranchTest.cs
dmitry-golovinov-sonarsource: Moving it there introduces some difficulties testing it as an extension method, since I have to call it by FQN to avoid conflicts with _DescendantsAndSelf_ extension method on line 85, but it still works.

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/BasicBlockTest.cs
```diff
@@ -118,6 +124,88 @@ public void Method(bool condition)
...
+    int Pow(int num, int exponent)
+    {
+        num = num * Pow(num, exponent - 1);
+        return 42;
+    }
+}";
+            var cfg = TestHelper.CompileCfg(code);
+            var entry = cfg.EntryBlock;
+            var exit = cfg.ExitBlock;
+            var body = cfg.Blocks[1];
```
pavel-mikula-sonarsource, line 149: _Nitpick:_ I'd swap the lines to have a logical order.
Same for the assertions below.
```suggestion
            var body = cfg.Blocks[1];
            var exit = cfg.ExitBlock;

```

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/BasicBlockTest.cs
```diff
@@ -118,6 +124,88 @@ public void Method(bool condition)
...
+             *         Block 1
+             *        true => 42
+             * Else  /          \    WhenFalse
+             *      /            \
+             *     /              \
+             *  Block 2          Block 3
+             *   => 42           _ => 43
+             *    |                |            WhenFalse from Block 3 - Block 5 should not be reachable
+             *    |                |
+             *    |              Block 4        Block 5
```
pavel-mikula-sonarsource, line 185: The branch is missing, so it's hard to understand
```suggestion
             *    |                |       \      WhenFalse from Block 3 - Block 5 should not be reachable
             *    |                |        \
             *    |              Block 4        Block 5
```

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/BasicBlockTest.cs
```diff
@@ -118,6 +124,88 @@ public void Method(bool condition)
...
+             *    |               => 43         no match => throw exception
+             *    \                /
```
pavel-mikula-sonarsource, line 187: _Beauty:_ I would start it one space further. And have it two spaces narrower. To align it with line 180 style and indent.
```suggestion
             *     \              /
```

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/BasicBlockTest.cs
```diff
@@ -118,6 +124,88 @@ public void Method(bool condition)
...
+            var block2 = cfg.Blocks[2];
+            var block3 = cfg.Blocks[3];
+            var block4 = cfg.Blocks[4];
+            var block5 = cfg.Blocks[5];
+            var block6 = cfg.Blocks[6];
+            block1.FallThroughSuccessor.Destination.Should().Be(block2);
+            block1.ConditionalSuccessor.Destination.Should().Be(block3);
+            block1.SuccessorBlocks.Should().ContainInOrder(block2, block3);
+            block2.FallThroughSuccessor.Destination.Should().Be(block6);
+            block2.SuccessorBlocks.Single().Should().Be(block6);
```
pavel-mikula-sonarsource, line 204: Let's assert `ConditionalSuccessor` to be `null` for regular structure

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/BasicBlockTest.cs
pavel-mikula-sonarsource: I expect this to show me the original CFG shape, not our modifications.

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/BasicBlockTest.cs
```diff
@@ -118,6 +124,88 @@ public void Method(bool condition)
...
+            block3.FallThroughSuccessor.Destination.Should().Be(block4);
+            block3.ConditionalSuccessor.Destination.Should().Be(block5);
+            block3.SuccessorBlocks.Single().Should().Be(block4);
```
pavel-mikula-sonarsource, line 207: Let's add a comment for the irregularity
```suggestion
            block3.SuccessorBlocks.Single().Should().Be(block4);	// We don't add the unreachable ConditionalSuccessor in this case
```

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/ControlFlowBranchTest.cs
```diff
@@ -96,10 +96,11 @@ public void Method(bool condition)
              *            v
              *         Exit 4
              */
+            var entryBlock = cfg.EntryBlock;
```
pavel-mikula-sonarsource, line 99: Unused
```suggestion
```

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/ControlFlowBranchTest.cs
pavel-mikula-sonarsource: And this should be deleted, because it's moved already :)

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/RoslynControlFlowGraphTest.cs
```diff
@@ -85,5 +85,27 @@ public void Method()
...
+            const string code = @"
+public class Sample {
+    private System.Action<int> Simple(int a)
+    {
+        var x = 42;
+        if (a == 42)
+        {
+            return (x) => {  };
+        }
+        return (x) => {  };
```
pavel-mikula-sonarsource, line 101: Let's have one without parenthesized. Just in case.
```suggestion
        return x => {  };
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/BasicBlockOperationFinder.cs
```diff
@@ -0,0 +1,69 @@
...
+            result.Should().Be(42);
+        }
+
+        private class FirstNumericLiteralFinder : BasicBlockOperationFinder<int>
+        {
+            protected override bool TryFindOperation(IOperationWrapperSonar operation, out int result)
+            {
+                if (operation.Instance is ILiteralOperation)
+                {
+                    result = int.Parse(operation.Instance.Syntax.ToString());
```
pavel-mikula-sonarsource, line 61: The value should be already there
```suggestion
                if (operation.Instance is ILiteralOperation)
                {
                    result = (int)operation.Instance.ConstantValue.Value;
```

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/CfgAllPathValidatorTest.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    public int Method2(bool condition)
+    {
+        if (condition)
+            return 42;
+        else
+            return 1;
+    }
+}";
+            var cfg = TestHelper.CompileCfg(code);
```
pavel-mikula-sonarsource, line 45: Can you draw the blocks in a very simple way? Just to see the shape (not the operations, they are not important here).

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/CfgAllPathValidatorTest.cs
```diff
@@ -0,0 +1,75 @@
...
+            var validator = new CfgValidate(cfg, new List<int> { 3, 4 });
```
pavel-mikula-sonarsource, line 46: Let's add a testcase where entry block is the only valid block to check that path.

pavel-mikula-sonarsource (changes requested):
Final polishing

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/BasicBlockTest.cs
dmitry-golovinov-sonarsource: It is the original shape now, having edge from block 3 to 5

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/BasicBlockOperationFinder.cs
dmitry-golovinov-sonarsource: Wow, that's handy

analyzers/src/SonarAnalyzer.CFG/Helpers/BasicBlockOperationFinder.cs
```diff
@@ -0,0 +1,30 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarAnalyzer.CFG.Roslyn;
+
+namespace SonarAnalyzer.CFG.Helpers
+{
+    public abstract class BasicBlockOperationFinder<TResult> : OperationFinder<TResult>
+    {
+        public bool TryFind(BasicBlock block, out TResult var) =>
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Waaay better 😄

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/BasicBlockTest.cs
pavel-mikula-sonarsource: >Same for the assertions below.

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/CfgAllPathValidatorTest.cs
```diff
@@ -0,0 +1,108 @@
...
+            protected override bool IsValid(BasicBlock block) => block.Ordinal > 0;
+
+            protected override bool IsInvalid(BasicBlock block) => false;
+        }
+
+        private class CfgValidate : CfgAllPathValidator
+        {
+            private readonly List<int> invalidBlocks;
+
+            public CfgValidate(ControlFlowGraph cfg, List<int> invalidBlocks) : base(cfg)
```
pavel-mikula-sonarsource, line 96: We don't really need a list, we won't be adding items to it. This will simplify the invocation. the field could be array too.
```suggestion
            public CfgValidate(ControlFlowGraph cfg, params int[] invalidBlocks) : base(cfg)
```

analyzers/src/SonarAnalyzer.CFG/Helpers/OperationFinder.cs
```diff
@@ -0,0 +1,61 @@
...
+using StyleCop.Analyzers.Lightup;
+
+namespace SonarAnalyzer.CFG.Helpers
+{
+    public abstract class OperationFinder<TResult>
+    {
+        protected abstract bool TryFindOperation(IOperationWrapperSonar operation, out TResult result);
+
+        public bool TryFind(BasicBlock block, out TResult var) =>
+            TryFind(block.OperationsAndBranchValue, out var);
```
pavel-mikula-sonarsource, line 34:
```suggestion
        public bool TryFind(BasicBlock block, out TResult result) =>
            TryFind(block.OperationsAndBranchValue, out result);
```

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/CfgAllPathValidatorTest.cs
```diff
@@ -0,0 +1,108 @@
...
+             *        \        /
+             *         \      /
+             *          \    /
+             *           Exit 4
+             */
+            var validator = new CfgValidate(cfg, new List<int> { 3, 4 });
+            validator.CheckAllPaths().Should().BeTrue();
+
+            // only entry block is valid
+            validator = new CfgValidate(cfg, new List<int> { 1, 2, 3, 4 });
```
pavel-mikula-sonarsource, line 64: I would prefer to list valid blocks here. To have positive logic.

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/CfgAllPathValidatorTest.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        private class NonEntryBlockValid : CfgAllPathValidator
+        {
+            public NonEntryBlockValid(ControlFlowGraph cfg) : base(cfg) { }
+
+            protected override bool IsValid(BasicBlock block) => block.Ordinal > 0;
+
+            protected override bool IsInvalid(BasicBlock block) => false;
+        }
```
pavel-mikula-sonarsource, line 90: I don't think we need the two. The last one is powerful enough to do the job.
```suggestion
```

analyzers/tests/SonarAnalyzer.UnitTest/CFG/Roslyn/CfgAllPathValidatorTest.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        private class CfgValidate : CfgAllPathValidator
```
pavel-mikula-sonarsource, line 92: To make it clear that it's UT-only class
```suggestion
        private class TestCfgAllPathValidator : CfgAllPathValidator
```
or simplier
```suggestion
        private class TestCfgValidator : CfgAllPathValidator
```

pavel-mikula-sonarsource (changes requested):
Last round.

pavel-mikula-sonarsource (approved):
LGTM. It would be also nice to add last case
```
// only exit block is valid
validator = new TestCfgValidator(cfg, 4);
validator.CheckAllPaths().Should().BeTrue();
```

PR #4874, opened 2021-09-06
Rule S3267: Loops should be simplified with "LINQ" expressions

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
```diff
@@ -43,15 +48,96 @@ public sealed class LoopsAndLinq : SonarDiagnosticAnalyzer
...
+        private static bool CanIfStatementBeMoved(IfStatementSyntax ifStatementSyntax) =>
+            ifStatementSyntax.Else == null
+            && !ContainsExitPoints(ifStatementSyntax)
+            && ifStatementSyntax.Condition is InvocationExpressionSyntax invocationExpressionSyntax
+            && !invocationExpressionSyntax.DescendantNodes().OfType<ArgumentSyntax>().Any(argument => argument.RefOrOutKeyword.IsKind(SyntaxKind.OutKeyword));
+
+        private static bool ContainsExitPoints(SyntaxNode syntaxNode) =>
+            syntaxNode.DescendantNodes()
+                      .Any(node => node.IsAnyKind(SyntaxKind.ReturnStatement,
+                                                  SyntaxKind.YieldReturnStatement,
```
andrei-epure-sonarsource, line 82: ThrowStatementSyntax, ContinueStatementSyntax

what about containing an `await` inside the `if` block?

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
```diff
@@ -43,15 +48,96 @@ public sealed class LoopsAndLinq : SonarDiagnosticAnalyzer
...
+                IfStatementSyntax ifStatementSyntax => ifStatementSyntax,
+                BlockSyntax blockSyntax when blockSyntax.ChildNodes().Count() == 1 => GetIfStatement(blockSyntax.ChildNodes().Single()),
+                _ => null
+            };
+
+        private static bool CanIfStatementBeMoved(IfStatementSyntax ifStatementSyntax) =>
+            ifStatementSyntax.Else == null
+            && !ContainsExitPoints(ifStatementSyntax)
+            && ifStatementSyntax.Condition is InvocationExpressionSyntax invocationExpressionSyntax
+            && !invocationExpressionSyntax.DescendantNodes().OfType<ArgumentSyntax>().Any(argument => argument.RefOrOutKeyword.IsKind(SyntaxKind.OutKeyword));
```
andrei-epure-sonarsource, line 77: what if the method is invoked with a `ref` argument? IMO it might be lost (harder to read) doing this inside a LINQ statement (although feel free to disagree, I don't have a strong opinion here)

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
```diff
@@ -43,15 +48,96 @@ public sealed class LoopsAndLinq : SonarDiagnosticAnalyzer
...
+
+        private static bool ContainsExitPoints(SyntaxNode syntaxNode) =>
+            syntaxNode.DescendantNodes()
+                      .Any(node => node.IsAnyKind(SyntaxKind.ReturnStatement,
+                                                  SyntaxKind.YieldReturnStatement,
+                                                  SyntaxKind.BreakStatement,
+                                                  SyntaxKind.YieldBreakStatement,
```
andrei-epure-sonarsource, line 84: Please add tests with `Yield break`, `Yield return` etc

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LoopsAndLinq.cs
```diff
@@ -1,30 +1,122 @@
...
+            p2 = string.Empty;
+            return true;
+        }
+
+        public void ForEachWithBlockWithIfSuggestWhere(ICollection<string> collection, Predicate<string> condition)
         {
             var result = new List<string>();
 
             foreach (var element in collection) // Noncompliant {{Loops should be simplified with "LINQ" expressions}}
+            //                      ^^^^^^^^^^
```
andrei-epure-sonarsource, line 20: could you also add a test with

```
foreach (var element in collection)
{
  Foo(1);
  if (condition(element))
  {
    result.Add(element);
   }
}

```
and
could you also add a test with

```
foreach (var element in collection)
{
  result.Add(element);
  if (condition(element))
  {
  }
}
 
```

(you have with member access usage, but not with method call before/after the if)

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
costin-zaharia-sonarsource: I've added `ThrowStatementSyntax` and `ContinueStatementSyntax`. I think `await` is not really changing the flow and we can skip it for now.

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
costin-zaharia-sonarsource: I've excluded `out` since it could lead to un-compilable code. With `ref` the code will compile so it's usually up for the developer to decide; a matter of preferences.

Now I think that it's better to exclude it to avoid annoying messages for the users who don't want to pass `ref` parameters to the invocations inside LINQ functions.

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
costin-zaharia-sonarsource: And, in the end, I've decided to skip this early exit point check. It didn't bring any value.

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
costin-zaharia-sonarsource: Thanks for raising the point. It made me rethink this and remove the check.

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
```diff
@@ -0,0 +1,137 @@
...
+                    && !(identifierSyntax.Parent.Parent is AssignmentExpressionSyntax assignment && assignment.Left == identifierSyntax.Parent)
+                    && c.SemanticModel.GetSymbolInfo(identifierSyntax).Symbol.Equals(declaredSymbol))
+                {
+                    var symbol = c.SemanticModel.GetSymbolInfo(memberAccessExpressionSyntax.Name).Symbol;
+                    var usageStats = accessedProperties.GetOrAdd(symbol, _ => new UsageStats());
+
+                    if (identifierSyntax.Parent.Parent is EqualsValueClauseSyntax { Parent: VariableDeclaratorSyntax })
+                    {
+                        usageStats.IsInVarDeclarator = true;
+                    }
```
andrei-epure-sonarsource, line 103:
```suggestion

                    usageStats.IsInVarDeclarator = memberAccessExpressionSyntax.Parent is EqualsValueClauseSyntax { Parent: VariableDeclaratorSyntax };
```

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
```diff
@@ -0,0 +1,137 @@
...
+            // There are multiple scenarios where the code can be simplified using LINQ.
+            // For simplicity, in the first version of the rule, we consider that Select() can be used
+            // only when a single property from the foreach variable is used.
+            // The issue is raised if:
+            //  - the property is used more than once
+            //  - the property is the right side of a variable declaration
+            var declaredSymbol = c.SemanticModel.GetDeclaredSymbol(forEachStatementSyntax);
+            var accessedProperties = new Dictionary<ISymbol, UsageStats>();
+
+            foreach (var identifierSyntax in GetIdentifiers(forEachStatementSyntax))
```
andrei-epure-sonarsource, line 91:
```suggestion
            foreach (var identifierInBody in GetIdentifiersInBody(forEachStatementSyntax))
```

- `identifierSyntax` doesn't convey a lot of meaning, it's quite generic

- `GetIdentifiers` is looking in the `Statement` which is actually the body of the foreach (BlockStatement or ExpressionStatement if no block) - to make it clear it's not looking into the `foreach (...)` part of the syntax

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
```diff
@@ -0,0 +1,137 @@
...
+            {
+                if (identifierSyntax.Parent is MemberAccessExpressionSyntax { Parent: not InvocationExpressionSyntax } memberAccessExpressionSyntax
```
andrei-epure-sonarsource, line 93: why `Parent: not InvocationExpressionSyntax` ? maybe add a short comment?

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
```diff
@@ -0,0 +1,137 @@
...
+                    && !(identifierSyntax.Parent.Parent is AssignmentExpressionSyntax assignment && assignment.Left == identifierSyntax.Parent)
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: Looking at it, moving it into a local function `IsNotLeftSideOfAssignment(memberAccessExpressionSyntax)` may be clearer

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LoopsAndLinq.cs
```diff
@@ -0,0 +1,228 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Tests.Diagnostics
+{
+    public class S3267
```
andrei-epure-sonarsource, line 7: what about the scenarios with LINQ statements inside

```
foreach (var element in collection.Select(...).Where(...))
{
  if (condition(element))
  {

  }
}

```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LoopsAndLinq.cs
```diff
@@ -0,0 +1,228 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
```
andrei-epure-sonarsource, line 3: could you also the examples inside the RSPEC? To have the in the UTs

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
```diff
@@ -0,0 +1,137 @@
+﻿/*
```
andrei-epure-sonarsource, line 1: we should have a ticket to create a code fixer for this one

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
```diff
@@ -0,0 +1,137 @@
...
+                }
+            }
+
+            if (accessedProperties.Count == 1
+                && accessedProperties.First().Value is var stats
+                && (stats.IsInVarDeclarator || stats.Count > 1))
+            {
+                var diagnostic = Diagnostic.Create(Rule,
+                                                   forEachStatementSyntax.Expression.GetLocation(),
+                                                   string.Format(SelectMessageFormat, forEachStatementSyntax.Identifier.ValueText, accessedProperties.Single().Key.Name));
```
andrei-epure-sonarsource, line 119: you don't need string.Format here, you just pass the message followed by the arguments and the `Create` method does the formatting, right?

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
costin-zaharia-sonarsource: In this case, the main message is "{0}", which gets replaced by the generic one or the specific "select". Then, the select message is "Loop should be simplified by calling Select({0} => {0}.{1}))" for which we need the string format to apply the arguments.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LoopsAndLinq.cs
costin-zaharia-sonarsource: We will still raise. I will add a test case.

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
costin-zaharia-sonarsource: I think that moving method invocations can be debatable since depending on the method being called it can make the code harder to read.

I've added a test case for this since I realized it was missing.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LoopsAndLinq.cs
costin-zaharia-sonarsource: As far as I can tell they are already there.

The first case:
https://github.com/SonarSource/sonar-dotnet/pull/4874/files#diff-120f54707b417f7640b7722b69ca748f7d88d8a53b8fe03450b82d22a5990870R21

The second case: 
https://github.com/SonarSource/sonar-dotnet/pull/4874/files#diff-120f54707b417f7640b7722b69ca748f7d88d8a53b8fe03450b82d22a5990870R162

I've added one more with a null check to be extra safe 😄

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
costin-zaharia-sonarsource: Added here: https://github.com/SonarSource/sonar-dotnet/issues/4894

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LoopsAndLinq.cs
andrei-epure-sonarsource: what I see is the Noncompliant cases from the RSPEC.

I meant the _compliant_ cases where LINQ expressions are used.

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
```diff
@@ -120,11 +118,14 @@ private static void CheckIfCanBeSimplifiedUsingSelect(ForEachStatementSyntax for
...
 
-            static IEnumerable<IdentifierNameSyntax> GetIdentifiers(ForEachStatementSyntax forEachStatementSyntax) =>
+            static IEnumerable<IdentifierNameSyntax> GetStatementIdentifiers(ForEachStatementSyntax forEachStatementSyntax) =>
                 forEachStatementSyntax.Statement
                                       .DescendantNodes()
                                       .OfType<IdentifierNameSyntax>()
                                       .Where(identifierNameSyntax => identifierNameSyntax.Identifier.ValueText == forEachStatementSyntax.Identifier.ValueText);
+
+            static bool IsNotLeftSideOfAssignment(IdentifierNameSyntax identifierSyntax) =>
+                !(identifierSyntax.Parent.Parent is AssignmentExpressionSyntax assignment && assignment.Left == identifierSyntax.Parent);
```
andrei-epure-sonarsource, line 128: why not pass `memberAccessExpressionSyntax` here? 

`identifierSyntax.Parent.Parent` is quite hard to understand - when reading this function you need to go back to realize that `identifierSyntax.Parent` is a `MemberAccessExpressionSyntax` ...

if you pass `memberAccessExpressionSyntax` instead of `identifierSyntax` you'll only need

`!(memberAccessExpressionSyntax.Parent is AssignmentExpressionSyntax assignment && assignment.Left == memberAccessExpressionSyntax);`

which is easier to understand

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LoopsAndLinq.cs
andrei-epure-sonarsource: what I see is the Noncompliant cases from the RSPEC.

I meant the _compliant_ cases where LINQ expressions are used - otherwise how can the reader see that the "compliant" cases from the RSPEC are truly compliant?

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
andrei-epure-sonarsource: could you add a comment "we don't want to move method invocations" in the code? 

as a reader who didn't write the rule, it's quite difficult to find the test case which covers this without a helping comment

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LoopsAndLinq.cs
```diff
@@ -169,6 +177,15 @@ public void ForEachWithPropertyGet_SuggestsSelect(ICollection<string> collection
                 }
             }
 
+            foreach (var element in collection) // Noncompliant
```
andrei-epure-sonarsource, line 180: looking at this better - wouldn't a secondary location be helpful, pointing at the `if` ? to have a visual helper that "hey, I am raising on the `foreach` and here's an `if` which can be moved in the foreach"

IMO this will help the developer see how to fix the issue without opening the rule description

andrei-epure-sonarsource (changes requested):
Looks good. I've left some more comments and a question about secondary locations

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
costin-zaharia-sonarsource: What do you think about this one:
https://github.com/SonarSource/sonar-dotnet/pull/4874/files#diff-7d38d8331402fc4becf0c5d9dd102d11e36b953c2455d84f5ba20b001f39d2cbR86
We have this comment describing the big picture and it is already part of it.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LoopsAndLinq.cs
costin-zaharia-sonarsource: Yes, my bad. I've misread your comment. I've added the cases and I will add a new PR on RSPEC to update them a bit since the current version does not compile :)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LoopsAndLinq.cs
costin-zaharia-sonarsource: Indeed.

costin-zaharia-sonarsource: I've added secondary locations for Where suggestions. For Select(), I think this is not needed since the property can be used in multiple places and we already include the property name in the error message.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/LoopsAndLinq.cs
costin-zaharia-sonarsource: See here: https://github.com/SonarSource/rspec/pull/337

Corniel:
I like the rule. I would suggest to consider also provide a fix for this one. That might be a challenging one, and is (probably) better done in a separate pull request, but I think it adds a lot of value for the users.

analyzers/src/SonarAnalyzer.CSharp/Rules/LoopsAndLinq.cs
```diff
@@ -0,0 +1,146 @@
...
+                                          .Any(argument => argument.RefOrOutKeyword.IsAnyKind(SyntaxKind.OutKeyword, SyntaxKind.RefKeyword));
+
+        private static void CheckIfCanBeSimplifiedUsingSelect(ForEachStatementSyntax forEachStatementSyntax, SyntaxNodeAnalysisContext c)
+        {
+            // There are multiple scenarios where the code can be simplified using LINQ.
+            // For simplicity, we consider that Select() can be used
+            // only when a single property from the foreach variable is used.
+            // We skip checking method invocations since depending on the method being called, moving it can make the code harder to read.
+            // The issue is raised if:
+            //  - the property is used more than once
```
Corniel, line 96: Use ` /// <remarks>` on the method instead.

costin-zaharia-sonarsource:
> I like the rule. I would suggest to consider also provide a fix for this one. That might be a challenging one, and is (probably) better done in a separate pull request, but I think it adds a lot of value for the users.

Indeed. We have an issue for this: https://github.com/SonarSource/rspec/pull/337

andrei-epure-sonarsource (approved):
LGTM

mary-georgiou-sonarsource (approved):
LGTM

PR #4875, opened 2021-09-06
S2933 FP fix: Tuple deconstruction 

### LINKED ISSUE #4858: False Positive: S2933
### Description

Tuple deconstruction breaks S2933 (make field readonly)

### Repro steps

```c#
class Program
{
    private int x; // S2933: make 'x' 'readonly'
    private int y; // S2933: make 'y' 'readonly'

    public Program()
    {
        x = 0;
        y = 0;
    }

    public void ModifyXAndY()
    {
        (x, y) = NewValues();
    }

    private (int, int) NewValues() => (1, 2);
}
```

### Expected behavior

I don't expect any warning.

### Actual behavior

I get S2933 in `x` and `y` definition.
If I define `x` and `y` as `readonly` fields, I get compilation error: 	"A readonly field cannot be assigned to (except in a constructor or init-only setter of the type in which the field is defined or a variable initializer)"

### Known workarounds

No.

### Related information

**C#**
Version: 7.3
.NetFramework 4.72

**Visual Studio version**
16.10.4

**MSBuild / dotnet version**
```
msbuild -ver
Microsoft (R) Build Engine version 16.10.2+857e5a733 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

16.10.2.30804
```

**Sonar Nuget Package**
version: 8.27.0.35380

**Operating System**
OS Name: Microsoft Windows 10 Pro
Version: 10.0.19043 Build 19043

csaba-sagi-sonarsource:
Fixes #4858

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/FieldShouldBeReadonly.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/FieldShouldBeReadonly.cs
@@ -238,7 +238,11 @@ namespace SonarAnalyzer.Rules.CSharp
 
                     foreach (var assignment in assignments)
                     {
-                        ProcessExpression(assignment.Left);
+                        var assignmentExpressions = Assignments(assignment.Left);
+                        foreach (var theAssignment in assignmentExpressions)
+                        {
+                            ProcessExpression(theAssignment);
+                        }
                     }
                 }
 
@@ -292,11 +296,8 @@ namespace SonarAnalyzer.Rules.CSharp
 
                 private void ProcessAssignedExpression(ExpressionSyntax expression)
                 {
-                    var symbolAndExpressionPairs = FieldSymbolFromAssignment(expression);
-                    foreach (var symbolAndExpressionPair in symbolAndExpressionPairs)
-                    {
-                        ProcessExpressionOnField(symbolAndExpressionPair.Item2, symbolAndExpressionPair.Item1);
-                    }
+                    var fieldSymbol = this.partialTypeDeclaration.SemanticModel.GetSymbolInfo(expression).Symbol as IFieldSymbol;
+                    ProcessExpressionOnField(expression, fieldSymbol);
                 }
 
                 private void ProcessExpressionOnField(ExpressionSyntax expression, IFieldSymbol fieldSymbol)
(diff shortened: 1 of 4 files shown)


csaba-sagi-sonarsource:
I created the following issue for the code fix provider: https://github.com/SonarSource/sonar-dotnet/issues/4879

analyzers/src/SonarAnalyzer.CSharp/Rules/FieldShouldBeReadonly.cs
```diff
@@ -309,37 +306,31 @@ private void ProcessExpressionOnField(ExpressionSyntax expression, IFieldSymbol
...
-                private static bool IsFieldRelevant(IFieldSymbol fieldSymbol)
-                {
-                    return fieldSymbol != null &&
-                           !fieldSymbol.IsStatic &&
-                           !fieldSymbol.IsConst &&
-                           !fieldSymbol.IsReadOnly &&
-                           fieldSymbol.DeclaredAccessibility == Accessibility.Private;
-                }
+                private static bool IsFieldRelevant(IFieldSymbol fieldSymbol) =>
+                    fieldSymbol is { IsStatic: false, IsConst: false, IsReadOnly: false, DeclaredAccessibility: Accessibility.Private };
```
andrei-epure-sonarsource, line 331: nice :)

analyzers/src/SonarAnalyzer.CSharp/Rules/FieldShouldBeReadonly.cs
```diff
@@ -231,13 +229,17 @@ private void CollectFieldsFromPrefixUnaryExpressions()
...
-                    var assignments = this.partialTypeDeclaration.Node.DescendantNodes()
+                    var assignments = partialTypeDeclaration.Node.DescendantNodes()
                         .OfType<AssignmentExpressionSyntax>()
-                        .Where(a => assignmentKinds.Contains(a.Kind()));
+                        .Where(a => AssignmentKinds.Contains(a.Kind()));
 
                     foreach (var assignment in assignments)
                     {
-                        ProcessExpression(assignment.Left);
+                        var identifierNames = IdentifierNames(assignment.Left);
```
andrei-epure-sonarsource, line 238: I suggest to use `leftSideExpressions` to make it clear from the name what we're looking at

and then the call below would be `ProcessExpression(leftSideExpression);`

analyzers/src/SonarAnalyzer.CSharp/Rules/FieldShouldBeReadonly.cs
```diff
@@ -304,34 +306,45 @@ private void ProcessExpressionOnField(ExpressionSyntax expression, IFieldSymbol
...
-                        this.readonlyFieldCollector.excludedFields.Add(fieldSymbol);
+                        readonlyFieldCollector.excludedFields.Add(fieldSymbol);
                     }
                 }
 
-                private static bool IsFieldRelevant(IFieldSymbol fieldSymbol)
+                private static bool IsFieldRelevant(IFieldSymbol fieldSymbol) =>
+                    fieldSymbol is { IsStatic: false, IsConst: false, IsReadOnly: false, DeclaredAccessibility: Accessibility.Private };
+
+                private static IEnumerable<ExpressionSyntax> IdentifierNames(ExpressionSyntax expression)
```
andrei-epure-sonarsource, line 333: this is not really `IdentifierNames`, right? it's returning expressions, not IdentifierNameSyntax ...

could you please find a better name? what does this method do? It either gets the expressions inside a tuple, or it returns the expression

`TupleExpressionsOrSelf` sounds more like what it does, right?

andrei-epure-sonarsource (approved):
LGTM, please improve the names before merging

PR #4878, opened 2021-09-07
Repack newtonsoft inside the analyzer DLL to avoid conflicts in NET Core and NET 5+

### LINKED ISSUE #4808: Could not load file or assembly 'System.Security.Permissions..' when building projects with `dotnet`
Source: https://community.sonarsource.com/t/azure-devops-pipeline-build-fails-with-sonarcloud-and-source-generator/45725

Context:
- `sonar-dotnet` embeds the `Newtonsoft.Json.dll` library compiled for net45.
- Our user is using a code generator project which targets netstandard2.0 and is compiled for the net5 runtime.

Stack Trace:
```
Message: Could not load file or assembly 'System.Security.Permissions, Version=0.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.
   at Newtonsoft.Json.Serialization.JsonTypeReflector.get_DynamicCodeGeneration()
   at Newtonsoft.Json.Serialization.JsonTypeReflector.get_ReflectionDelegateFactory()
   at Newtonsoft.Json.Serialization.DefaultContractResolver.GetDefaultCreator(Type createdType)
   at Newtonsoft.Json.Serialization.DefaultContractResolver.InitializeContract(JsonContract contract)
   at Newtonsoft.Json.Serialization.DefaultContractResolver.CreateArrayContract(Type objectType)
   at Newtonsoft.Json.Serialization.DefaultContractResolver.CreateContract(Type objectType)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Newtonsoft.Json.Utilities.ThreadSafeStore`2.Get(TKey key)
   at Newtonsoft.Json.Serialization.DefaultContractResolver.ResolveContract(Type type)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.GetContract(Type type)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.GetContractSafe(Type type)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
   at Newtonsoft.Json.JsonSerializer.Deserialize(JsonReader reader, Type objectType)
   at Newtonsoft.Json.JsonConvert.DeserializeObject(String value, Type type, JsonSerializerSettings settings)
   at Newtonsoft.Json.JsonConvert.DeserializeObject[T](String value, JsonSerializerSettings settings)
   at Newtonsoft.Json.JsonConvert.DeserializeObject[T](String value)
   at SourceGen.Generator.Execute(GeneratorExecutionContext context)
```

The problem appears when JsonTypeReflector.get_DynamicCodeGeneration is called since .Net Core tries to resolve `SecurityPermission` from `System.Security.Permissions.dll` which cannot be found. Source code: https://github.com/JamesNK/Newtonsoft.Json/blob/11.0.2/Src/Newtonsoft.Json/Serialization/JsonTypeReflector.cs#L449

The type was made obsolete starting with .Net 5: https://docs.microsoft.com/en-us/dotnet/api/system.security.permissions.securitypermission?view=net-5.0

There seem to be differences when building with MSBuild vs dotnet since MSBuild will not even try to load the `System.Security.Permissions.dll` assembly.

I’ve done a short test and, if I embed the netstandard2.0 version of `Newtonsoft.Json.dll`, I can build the repro project without any problems. The limitation with this approach is that we will have compatibility issues with older versions of Visual Studio.

As far as I can tell the compiler will always load the version of Newtonsoft.Json we provide. Seems to be related to this issue: https://github.com/dotnet/roslyn/issues/41421.

One possible solution is to remove the dependency on `Json.Net`.

andrei-epure-sonarsource:
Fixes #4808


andrei-epure-sonarsource:
There is a problem with the RuleDescriptorGenerator which attempts to load all the types in the SonarAnalyzer.* assemblies.

The problem is that, when the Newtonsoft.Json assembly is merged into the SonarAnalyzer.CSharp assembly, it attempts to load all Newtonsoft types and it fails because it doesn't find all the dependencies on the machine.

At this stage, after ea031e7, I managed to fix locally the problem with `System.Xml.ReaderWriter` but on the CI it's failing for probably another type. Which type it's failing for cannot be easily found because the error message is:

```
Unhandled Exception: System.TypeInitializationException: The type initializer for 'SonarAnalyzer.Utilities.RuleFinder' threw an exception. ---> System.IO.FileLoadException: Could not load file or assembly 'SonarAnalyzer.CSharp, Version=8.29.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. A strongly-named assembly is required. (Exception from HRESULT: 0x80131044)
     at SonarAnalyzer.Utilities.RuleFinder..cctor()
     --- End of inner exception stack trace ---
     at SonarAnalyzer.Utilities.RuleFinder..ctor() in C:\ProgramData\vsts-agent\1\s\analyzers\src\SonarAnalyzer.Utilities\RuleFinder.cs:line 45
     at SonarAnalyzer.Utilities.RuleDetailBuilder.GetAllRuleDetails(AnalyzerLanguage language) in C:\ProgramData\vsts-agent\1\s\analyzers\src\SonarAnalyzer.Utilities\RuleDetailBuilder.cs:line 50
     at SonarAnalyzer.RuleDescriptorGenerator.Program.WriteXmlDescriptorFiles(String rulePath, String profilePath, String lang) in C:\ProgramData\vsts-agent\1\s\analyzers\src\SonarAnalyzer.RuleDescriptorGenerator\Program.cs:line 57
     at SonarAnalyzer.RuleDescriptorGenerator.Program.Main(String[] args) in C:\ProgramData\vsts-agent\1\s\analyzers\src\SonarAnalyzer.RuleDescriptorGenerator\Program.cs:line 48
```

To find which type it was failing for locally, I had to debug the `RuleDescriptorGenerator` and inspect the `TypeInitializationException` , the `LoaderExceptions` property and inspect the `FusionLog` inside. See below.
image

This is why I explicitly added the `System.Xml.ReaderWriter` dependency. However it's not feasible to do that for every Newtonsoft dependency which is not present in the CI environment for any reason.

So what I plan to do is to run the ILRepack target _after_ the RuleDescriptorGenerator gets built...

andrei-epure-sonarsource:
It is still failing because of the cumbersome way we pack things.

The VSIX project is responsible for:
- creating the nuget package
- deploying the analyzer DLLs to the .NET IT binaries folder
- the binaries from the VSIX project are the ones zipped by the SQ Plugin pom.xml and included in the plugin

Currently the VSIX was getting built _before_ the ILRepack project, so the repacked binaries where not reaching the :
- .NET ITs
- SQ Plugin (implicitly the ITs)
- nuget

andrei-epure-sonarsource:
Not sure why the NET FW 4.8 UTs are failing as locally they were green.

andrei-epure-sonarsource:
This is now failing on the CI, not sure but I expect it may have something to do with signing (because locally it works so it must be something which is done only on the CI).

I will stop here because https://github.com/SonarSource/sonar-dotnet/pull/4888 is already out and we will be able (in a few weeks) to completely replace the newtonsoft library with our own parsing logic. Thus there's really no need to make our lives more difficult by modifying the order of the build.

PR #4882, opened 2021-09-07
Rule S3604: Supports Roslyn CFG

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant_SonarCfg.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                foreach (var instruction in block.Instructions)
+                {
+                    switch (instruction.Kind())
+                    {
+                        case SyntaxKind.IdentifierName:
+                        case SyntaxKind.SimpleMemberAccessExpression:
+                            {
+                                var memberAccess = RedundancyChecker.GetPossibleMemberAccessParent(instruction);
+                                if (memberAccess != null && redundancyChecker.TryGetReadWriteFromMemberAccess(memberAccess, out var isRead))
```
pavel-mikula-sonarsource, line 51: Can use `is { } memberAccess` instead, but I'm not sure if it's worth fixing this class.
```suggestion
                                if (RedundancyChecker.GetPossibleMemberAccessParent(instruction) is { } memberAccess && redundancyChecker.TryGetReadWriteFromMemberAccess(memberAccess, out var isRead))
```

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant_RoslynCfg.cs
```diff
@@ -0,0 +1,161 @@
...
+ */
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.CFG.Helpers;
+using SonarAnalyzer.CFG.Roslyn;
+using SonarAnalyzer.Extensions;
+using StyleCop.Analyzers.Lightup;
+using SymbolWithInitializer = System.Collections.Generic.KeyValuePair<Microsoft.CodeAnalysis.ISymbol, Microsoft.CodeAnalysis.CSharp.Syntax.EqualsValueClauseSyntax>;
```
pavel-mikula-sonarsource, line 28: Seems unused
```suggestion
```

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant_RoslynCfg.cs
```diff
@@ -0,0 +1,161 @@
...
+        private class MemberInitializerRedundancyCheckerRoslyn : CfgAllPathValidator
+        {
+            private readonly IsValidMatcher isValidMatcher;
+            private readonly IsInvalidMatcher isInvalidMatcher;
+
+            public MemberInitializerRedundancyCheckerRoslyn(ControlFlowGraph cfg, ISymbol memberToCheck, SemanticModel semanticModel) : base(cfg)
+            {
+                var redundancyChecker = new RedundancyChecker(memberToCheck, semanticModel);
+                this.isValidMatcher = new IsValidMatcher(redundancyChecker);
+                this.isInvalidMatcher = new IsInvalidMatcher(redundancyChecker);
```
pavel-mikula-sonarsource, line 43:
```suggestion
                isValidMatcher = new IsValidMatcher(redundancyChecker);
                isInvalidMatcher = new IsInvalidMatcher(redundancyChecker);
```

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant_RoslynCfg.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            // Returns true if the block contains assignment before access
+            protected override bool IsValid(BasicBlock block) =>
+                isValidMatcher.Matches(block);
+
+            // Returns true if the block contains access before assignment
+            protected override bool IsInvalid(BasicBlock block) =>
+                isInvalidMatcher.Matches(block);
+
+            private abstract class RedundancyMatcherBase : OperationFinder<bool>
```
pavel-mikula-sonarsource, line 54: Remove `Redundancy` because the inherited classes don't have it in the name.
`Finder` because it's a finder.
Only `FinderBase`, because it's private class and it doesn't need a long name.

```suggestion
            private abstract class FinderBase : OperationFinder<bool>
```

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant_RoslynCfg.cs
```diff
@@ -0,0 +1,161 @@
...
+                    TryFind(block, out var result);
+                    return result;
+                }
+            }
+
+            private  class IsValidMatcher : RedundancyMatcherBase
+            {
+                public IsValidMatcher(RedundancyChecker redundancyChecker) : base(redundancyChecker) { }
+
+                protected override bool TryFindOperation(IOperationWrapperSonar operation, out bool result)
```
pavel-mikula-sonarsource, line 72: This whole thing should be way easier and shorter using operations. We should be looking for few operations, like `IAssignmentOperation`. That will directly give us the `Target` symbol and we should not need to work with Syntax here at all. Something like I did in the LVA.

Imagine this part to be common for C# and VB.

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant_RoslynCfg.cs
```diff
@@ -0,0 +1,161 @@
...
+                    result = default;
+                    return false;
+                }
+            }
+
+            private  class IsInvalidMatcher : RedundancyMatcherBase
+            {
+                public IsInvalidMatcher(RedundancyChecker redundancyChecker) : base(redundancyChecker) { }
+
+                protected override bool TryFindOperation(IOperationWrapperSonar operation, out bool result)
```
pavel-mikula-sonarsource, line 112: Same here with Operations.

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.cs
```diff
@@ -220,7 +351,6 @@ public void Report(SyntaxNodeAnalysisContext c, TypeDeclarationSyntax declaratio
                         new DeclarationTuple<IPropertySymbol>
                         {
                             Initializer = p.Initializer,
-                            SemanticModel = semanticModel,
```
pavel-mikula-sonarsource, line 223: Please add a simple constructor to the `DeclarationTuple` to remove these painful constructions. They make the code very hard to read (lot of useless lines)

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.cs
```diff
@@ -63,16 +70,131 @@ public sealed class MemberInitializerRedundant : SonarDiagnosticAnalyzer
...
+            private readonly ISymbol memberToCheck;
+            private readonly SemanticModel semanticModel;
+
+            public RedundancyChecker(ISymbol memberToCheck, SemanticModel semanticModel)
+            {
+                this.memberToCheck = memberToCheck;
+                this.semanticModel = semanticModel;
+            }
+
+            public bool IsMemberUsedInsideLambda(SyntaxNode instruction) =>
```
pavel-mikula-sonarsource, line 94: The method order here is not easy to follow. Can you put the `Is` methods together and those two `GetPossibleMemberAccessParent` together?

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.cs
```diff
@@ -63,16 +70,131 @@ public sealed class MemberInitializerRedundant : SonarDiagnosticAnalyzer
...
+            {
+                if (node is MemberAccessExpressionSyntax memberAccess)
+                {
+                    return memberAccess;
+                }
+                else
+                {
+                    return GetPossibleMemberAccessParent(node as IdentifierNameSyntax);
+                }
+            }
```
pavel-mikula-sonarsource, line 110: Could be arrow method with ternary operator

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.cs
```diff
@@ -63,16 +70,131 @@ public sealed class MemberInitializerRedundant : SonarDiagnosticAnalyzer
...
+                this.semanticModel = semanticModel;
+            }
+
+            public bool IsMemberUsedInsideLambda(SyntaxNode instruction) =>
+                instruction.DescendantNodes()
+                    .OfType<IdentifierNameSyntax>()
+                    .Select(GetPossibleMemberAccessParent)
+                    .Any(IsMatchingMember);
+
+            public static ExpressionSyntax GetPossibleMemberAccessParent(SyntaxNode node)
```
pavel-mikula-sonarsource, line 100: Same for the other one
```suggestion
            public static ExpressionSyntax PossibleMemberAccessParent(SyntaxNode node)
```

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.cs
```diff
@@ -63,16 +70,131 @@ public sealed class MemberInitializerRedundant : SonarDiagnosticAnalyzer
...
-                    InitializationVerifierFactory.StaticMemberInitializationVerifier.Report(c, declaration, members);
+                    InitializationVerifierFactory.StaticMemberInitializationVerifier.Report(c, declaration, members, forceSonarCfg);
                 },
                 // For record support, see details in https://github.com/SonarSource/sonar-dotnet/pull/4756
                 // it is difficult to work with instance record constructors w/o raising FPs
                 SyntaxKind.ClassDeclaration,
                 SyntaxKind.StructDeclaration,
                 SyntaxKind.InterfaceDeclaration);
 
+        private class RedundancyChecker
```
pavel-mikula-sonarsource, line 83: There's a lot of syntax stuff here that might not be needed at all in Operation-based implementation. Then it could be moved into the SonarCfg-specific part.

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.cs
```diff
@@ -63,16 +70,131 @@ public sealed class MemberInitializerRedundant : SonarDiagnosticAnalyzer
...
+                    && memberAccess.Expression.IsKind(SyntaxKind.ThisExpression))
+                {
+                    identifier = memberAccess.Name as IdentifierNameSyntax;
+                }
+
+                if (expression is ConditionalAccessExpressionSyntax conditionalAccess
+                    && conditionalAccess.Expression.IsKind(SyntaxKind.ThisExpression))
+                {
+                    identifier = (conditionalAccess.WhenNotNull as MemberBindingExpressionSyntax)?.Name as IdentifierNameSyntax;
+                }
```
pavel-mikula-sonarsource, line 131: Let's move this identifier-searching part to a local method `ExtractIdentifier` with `if return else if return else if return else return null` structure.
After that, this whole thing could be somethign like
```
return ExtractIdentifier() is identifer
    && semanticModel.GetSymbolInfo(identifier).Symbol is { } symbol
    && memberToCheck.Equals(assignedSymbol);
```
This improvement probably only makes sense if we'll use this in the Operation-base version.

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.cs
```diff
@@ -63,16 +70,131 @@ public sealed class MemberInitializerRedundant : SonarDiagnosticAnalyzer
...
+
+            private static bool IsOutArgument(ExpressionSyntax parenthesized) =>
+                parenthesized.Parent is ArgumentSyntax argument
+                && argument.RefOrOutKeyword.IsKind(SyntaxKind.OutKeyword);
+
+            private static bool IsReadAccess(ExpressionSyntax parenthesized, SemanticModel semanticModel) =>
+                !IsBeingAssigned(parenthesized)
+                && !parenthesized.IsInNameOfArgument(semanticModel);
+        }
+
```
pavel-mikula-sonarsource, line 197: This `InitializationVerifierFactory` is a triple indirection - really impossible to read and understand.
And all we get is unified 20 lines of code inside `Report` method, but we need 50 complicated lines to orchestrate it.

Let's just remove `InitializationVerifierFactory` and `InitializationVerifier` completely. It will make the code flat and straightforward. There will be two different implemenations for the current `Report` method, something like `CheckStaticMembers` and `CheckInstanceMembers`

pavel-mikula-sonarsource (changes requested):
I left some major redesign requests so I didn't go deep in the review as stuff will likely change a lot.

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.cs
```diff
@@ -208,258 +201,45 @@ public void Report(SyntaxNodeAnalysisContext c, TypeDeclarationSyntax declaratio
...
-                        && !MemberInitializedToDefault.IsDefaultValueInitializer(t.Initializer, typeSelector(t.Symbol)));
-        }
+        private static IEnumerable<DeclarationTuple<IPropertySymbol>> GetInitializedPropertyDeclarations(TypeDeclarationSyntax declaration,
+                                                                                                         Func<PropertyDeclarationSyntax, bool> declarationFilter,
+                                                                                                         SemanticModel semanticModel) =>
+            declaration.Members
+                .OfType<PropertyDeclarationSyntax>()
+                .Where(p => declarationFilter(p)
+                            && p.Initializer != null
+                            && p.IsAutoProperty())
```
pavel-mikula-sonarsource, line 211: Can be single line

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.cs
```diff
@@ -208,258 +201,45 @@ public void Report(SyntaxNodeAnalysisContext c, TypeDeclarationSyntax declaratio
...
+                .Select(p => new DeclarationTuple<IPropertySymbol>(p.Initializer, semanticModel.GetDeclaredSymbol(p)))
+                .Where(t =>
+                    t.Symbol != null
+                    && !MemberInitializedToDefault.IsDefaultValueInitializer(t.Initializer, t.Symbol.Type));
```
pavel-mikula-sonarsource, line 215: Can be single line

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant_RoslynCfg.cs
```diff
@@ -0,0 +1,117 @@
...
+using Microsoft.CodeAnalysis;
+using SonarAnalyzer.CFG.Roslyn;
+using SonarAnalyzer.Extensions;
+using StyleCop.Analyzers.Lightup;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    public sealed partial class MemberInitializerRedundant
+    {
+        private class MemberInitializerRedundancyCheckerRoslyn : CfgAllPathValidator
```
pavel-mikula-sonarsource, line 31: For simplicity, it's private nested
```suggestion
        private class RoslynChecker : CfgAllPathValidator
```

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant_RoslynCfg.cs
```diff
@@ -0,0 +1,117 @@
...
+                private readonly ISymbol memberToCheck;
+                private readonly ControlFlowGraph cfg;
+
+                public Matcher(ControlFlowGraph cfg, ISymbol memberToCheck)
+                {
+                    this.cfg = cfg;
+                    this.memberToCheck = memberToCheck;
+                }
+
+                public bool Match(BasicBlock block, bool isRead)
```
pavel-mikula-sonarsource, line 57: Isn't there (almost) the same thing in the other PR? Will you extract it after rebase or what is the plan?

pavel-mikula-sonarsource (changes requested):
Another round, getting close

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant_RoslynCfg.cs
dmitry-golovinov-sonarsource: I will extract it as rebase. Will create a comment in that PR

dmitry-golovinov-sonarsource: Comment added

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.cs
```diff
@@ -63,143 +70,129 @@ public sealed class MemberInitializerRedundant : SonarDiagnosticAnalyzer
...
+            var typeInitializers = typeMembers.OfType<IMethodSymbol>().Where(x => x is { MethodKind: MethodKind.Constructor, IsStatic: false, IsImplicitlyDeclared: false }).ToList();
+            if (typeInitializers.Count == 0)
+            {
+                return;
+            }
 
-            private static bool IsStatic(SyntaxToken token) =>
-                token.IsKind(SyntaxKind.StaticKeyword);
+            // only retrieve the member symbols (an expensive call) if there are explicit class initializers
+            var initializedMembers = GetInitializedMembers(c.SemanticModel, declaration, field => !field.Modifiers.Any(IsStaticOrConst), property => !property.Modifiers.Any(IsStaticOrConst));
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: Actually, this won't work, since on line 92 we check for negative condition and on line 129 - for positive. Was tricky to spot at the first sight

analyzers/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs
```diff
@@ -37,14 +37,19 @@ public class AnalyzerConfiguration
...
 
-        public static IAnalyzerConfiguration AlwaysEnabled { get; } =
-            new AlwaysEnabledConfiguration();
+        public static IAnalyzerConfiguration AlwaysEnabled { get; } = new AlwaysEnabledConfiguration(false);
+
+        public static IAnalyzerConfiguration AlwaysEnabledWithSonarCfg { get; } = new AlwaysEnabledConfiguration(true);
 
         private class AlwaysEnabledConfiguration : IAnalyzerConfiguration
         {
+            public bool ForceSonarCfg { get; }
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: You will not like me, but with the recent changes, I think the answer is `No, FoceSonarCfg is fine as it's exactly what it does now` 😆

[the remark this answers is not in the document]
pavel-mikula-sonarsource: Let's revert it to the `ForceSonarCfg` here

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Ah, ok. 
Round 2: Can we simplify it like this then?
```suggestion
            var initializedMembers = GetInitializedMembers(c.SemanticModel, declaration, IsStaticOrConst);
```

analyzers/src/SonarAnalyzer.CSharp/Extensions/IAnalyzerConfigurationExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+using SonarAnalyzer.Common;
+
+namespace SonarAnalyzer.Extensions
+{
+    public static class IAnalyzerConfigurationExtensions
+    {
+        public static bool FallBackToSonarCfg(this IAnalyzerConfiguration configuration) =>
+            !CFG.Roslyn.ControlFlowGraph.IsAvailable || configuration.UseSonarCfg;
```
pavel-mikula-sonarsource: I slightly prefer `a || !b` then `!b || a` when possible. The `ForceSonarCfg` should be slightly (unnoticably) cheaper as well.
```suggestion
            configuration.ForceSonarCfg || !CFG.Roslyn.ControlFlowGraph.IsAvailable;
```

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.RoslynCfg.cs
```diff
@@ -0,0 +1,111 @@
...
+                        return result;
+                    }
+                }
+
+                return false;
+            }
+
+            private bool ProcessOperation(IOperation operation, ControlFlowGraph controlFlowGraph, bool isRead, out bool result)
+            {
+                foreach (var child in operation.DescendantsAndSelf().ToReversedExecutionOrder())
```
pavel-mikula-sonarsource, line 65: I see here the same problem as on the other PR (where this code got deleted) - https://github.com/SonarSource/sonar-dotnet/pull/4855#discussion_r709968524

>The `ToReversedExectuionOrder()` will descent into children as well so we would end up iterating all of them twice.
>
>The intended usage is like this. I'd be also fine with creating a handy overload of the extension method so we avoid this pitfall in the future as well.
```suggestion
                        foreach (var child in new[] {operation}.ToReversedExecutionOrder())
```
>Or maybe something like this one level higher, but that might change the semantics.
```
foreach (var operation in block.OperationsAndBranchValue.ToReversedExecutionOrder())
```
>or
```
foreach (var operation in block.OperationsAndBranchValue.Reverse().ToReversedExecutionOrder())
```

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.RoslynCfg.cs
```diff
@@ -0,0 +1,111 @@
...
+                    if (isRead && child.Instance.Kind == OperationKindEx.FlowAnonymousFunction)
+                    {
+                        var anonymousFunctionCfg = controlFlowGraph.GetAnonymousFunctionControlFlowGraph(IFlowAnonymousFunctionOperationWrapper.FromOperation(child.Instance));
+                        foreach (var subOperation in anonymousFunctionCfg.Blocks.SelectMany(x => x.OperationsAndBranchValue).SelectMany(x => x.DescendantsAndSelf()))
+                        {
+                            if (ProcessOperation(subOperation, anonymousFunctionCfg, isRead, out result))
+                            {
+                                return true;
+                            }
+                        }
```
pavel-mikula-sonarsource, line 76: This is also similar to the comment on the other PR - https://github.com/SonarSource/sonar-dotnet/pull/4855#discussion_r709975109

>_Thinking out loud:_ This part is very similar to `anonymousFunctionCfg.Blocks.Any(FindReferenceToSelf)`, it just returns a little different result. 
>
>Can we unify them somehow? It's complicated so I'd like to have it just once there.

pavel-mikula-sonarsource: The key idea is, that 
`anonymousFunctionCfg.Blocks.SelectMany(x => x.OperationsAndBranchValue).SelectMany(x => x.DescendantsAndSelf())` 
is basically the same as 
`foreach (var operation in block.OperationsAndBranchValue)` 
plus
 `foreach (var child in operation.DescendantsAndSelf().ToReversedExecutionOrder())` 
if you'd merge them based on the previous comment.

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.RoslynCfg.cs
```diff
@@ -0,0 +1,111 @@
...
+                    if (ProcessOperation(operation, cfg, checkReadBeforeWrite, out var result))
+                    {
+                        return result;
+                    }
+                }
+
+                return false;
+            }
+
+            private bool ProcessOperation(IOperation operation, ControlFlowGraph controlFlowGraph, bool isRead, out bool result)
```
pavel-mikula-sonarsource, line 63: Should be propagated all the way down
```suggestion
            private bool ProcessOperation(IOperation operation, ControlFlowGraph controlFlowGraph, bool checkReadBeforeWrite, out bool result)
```

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.RoslynCfg.cs
```diff
@@ -0,0 +1,111 @@
...
+                        result = IsReadOrWrite(child, isRead);
+                        return true;
+                    }
+                }
+
+                result = false;
+                return false;
+            }
+
+            private bool IsReadOrWrite(IOperationWrapperSonar child, bool isRead)
```
pavel-mikula-sonarsource, line 89: To propagate the rename
```suggestion
            private bool IsReadOrWrite(IOperationWrapperSonar child, bool checkReadBeforeWrite)
```

pavel-mikula-sonarsource (changes requested):
Déjà vu :)

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.RoslynCfg.cs
dmitry-golovinov-sonarsource: duly noted on the other PR to take action on it before merging it

dmitry-golovinov-sonarsource: For now, converted to array of one element

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.RoslynCfg.cs
dmitry-golovinov-sonarsource: You still need to fetch inner cfg to get its own operations, that's why this block is needed.  But indeed, I've simplified
anonymousFunctionCfg.Blocks.SelectMany(x => x.OperationsAndBranchValue).SelectMany(x => x.DescendantsAndSelf())
to
anonymousFunctionCfg.Blocks.SelectMany(x => x.OperationsAndBranchValue)
and the outer foreach does the trick we need

pavel-mikula-sonarsource: I think you still don't see what I see in these two comments. After 3 or 4 rounds of decremental deduplications, this is what I see:

Delete line 42-88 (`IsValid` - updated, `IsInvalid` - updated, `Match` & `ProcessOperation` merged) and replace it with this:
```C#

            // Returns true if the block contains assignment before access
            protected override bool IsValid(BasicBlock block) =>
                ProcessBlock(block, cfg, false);

            // Returns true if the block contains access before assignment
            protected override bool IsInvalid(BasicBlock block) =>
                ProcessBlock(block, cfg, true);

            private bool ProcessBlock(BasicBlock block, ControlFlowGraph controlFlowGraph, bool checkReadBeforeWrite)
            {
                foreach (var operation in block.OperationsAndBranchValue.Reverse().ToReversedExecutionOrder())
                {
                    if (checkReadBeforeWrite && operation.Instance.Kind == OperationKindEx.FlowAnonymousFunction)
                    {
                        var anonymousFunctionCfg = controlFlowGraph.GetAnonymousFunctionControlFlowGraph(IFlowAnonymousFunctionOperationWrapper.FromOperation(operation.Instance));
                        if (anonymousFunctionCfg.Blocks.Any(x => ProcessBlock(x, anonymousFunctionCfg, checkReadBeforeWrite)))
                        {
                            return true;
                        }
                    }
                    else if (memberToCheck.Equals(MemberSymbol(operation.Instance)))
                    {
                        return IsReadOrWrite(operation, checkReadBeforeWrite);
                    }
                }
                return false;
            }
```

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.RoslynCfg.cs
pavel-mikula-sonarsource: Resolving - this follows in the other comment below https://github.com/SonarSource/sonar-dotnet/pull/4882#discussion_r713020235

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.RoslynCfg.cs
```diff
@@ -0,0 +1,111 @@
...
+                {
+                    // it is out argument - that means that this is write
+                    return !checkReadBeforeWrite;
+                }
+
+                var isWrite = child.Parent is { Kind: OperationKindEx.SimpleAssignment } parent && ISimpleAssignmentOperationWrapper.FromOperation(parent).Target == child.Instance;
+                return checkReadBeforeWrite ^ isWrite;
+            }
+
+            private static ISymbol MemberSymbol(IOperation operation) =>
```
pavel-mikula-sonarsource, line 101: We will also need to check if the fields is from `this` or something else. We have a FP in this case:
```
    class Person16
    {
        int a = 42;

        public Person16(Person16 other)
        {
            other.a = 0;
        }
    }
```

pavel-mikula-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant.RoslynCfg.cs
dmitry-golovinov-sonarsource: Suggested FP was addressed for three cases

pavel-mikula-sonarsource (approved):
LGTM 🎉

analyzers/src/SonarAnalyzer.CSharp/Rules/MemberInitializerRedundant_RoslynCfg.cs
dmitry-golovinov-sonarsource: (news from another pr) - code was extracted during one of many refactories

PR #4884, opened 2021-09-08
WinForms support in unit tests when targeting .Net Core

### LINKED ISSUE #3426: Add WinForms support in unit tests when targeting .Net Core

csaba-sagi-sonarsource:
Fixes #3426

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/Helpers/RemovableDeclarationCollectorTest.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/Helpers/RemovableDeclarationCollectorTest.cs
@@ -20,7 +20,6 @@
 
 using System.Linq;
 using FluentAssertions;
-using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.VisualBasic;
 using Microsoft.CodeAnalysis.VisualBasic.Syntax;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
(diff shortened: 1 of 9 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/Helpers/RemovableDeclarationCollectorTest.cs
```diff
@@ -20,7 +20,6 @@
 
 using System.Linq;
 using FluentAssertions;
-using Microsoft.CodeAnalysis;
```
andrei-epure-sonarsource, line 23: why this change?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MarkWindowsFormsMainWithStaThread.vb
```diff
@@ -14,8 +14,6 @@ Namespace Tests.TestCases
   Public Class Program_01
     Shared Sub Main() ' Noncompliant {{Add the 'STAThread' attribute to this entry point.}}
 '              ^^^^
-      Dim winForm As Form = New Form
```
andrei-epure-sonarsource, line 17: why remove this?

andrei-epure-sonarsource (changes requested):
I don't understand some changes, could you please explain?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MarkWindowsFormsMainWithStaThread.vb
andrei-epure-sonarsource: or is it just looking at the namespace?

csaba-sagi-sonarsource: After some offline talk with Andrei, we decided there is no action to be taken.

andrei-epure-sonarsource (approved):
LGTM

PR #4888, opened 2021-09-10
Introduce Json parsing

pavel-mikula-sonarsource:
Lightweight Json parser based on https://www.json.org/ grammar

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Json/JsonException.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Json/JsonException.cs
@@ -1,0 +1,35 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2021 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.Serialization;
+
+namespace SonarAnalyzer.Json
+{
+    [Serializable]
+    public sealed class JsonException : Exception
+    {
+        public JsonException(string message) : base(message) { }
+
+        [ExcludeFromCodeCoverage]
+        private JsonException(SerializationInfo info, StreamingContext context) : base(info, context) { }
+    }
+}
--- a/analyzers/src/SonarAnalyzer.Common/Json/JsonException.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Json/JsonException.cs
@@ -21,13 +21,14 @@
 using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.Serialization;
+using Microsoft.CodeAnalysis.Text;
 
 namespace SonarAnalyzer.Json
 {
     [Serializable]
     public sealed class JsonException : Exception
     {
-        public JsonException(string message) : base(message) { }
+        public JsonException(string message, LinePosition position) : base(message + $" at line {position.Line + 1} position {position.Character + 1}") { }
 
         [ExcludeFromCodeCoverage]
         private JsonException(SerializationInfo info, StreamingContext context) : base(info, context) { }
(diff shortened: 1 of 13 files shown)


analyzers/src/SonarAnalyzer.Common/Json/JsonNode.cs
```diff
@@ -0,0 +1,110 @@
...
+                case Kind.Object:
+                    map = new Dictionary<string, JsonNode>();
+                    break;
+                default:
+                    throw InvalidKind();
+            }
+        }
+
+        public static JsonNode FromString(string json) =>
+            new SyntaxAnalyzer(json).Parse();
```
pavel-mikula-sonarsource, line 76: This should be the main entry point. I want the `Parsing` namespace to be hidden from consumers.

analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
```diff
@@ -0,0 +1,313 @@
...
+        {
+            Value = null;
+            NextPosition(false);
+            SkipWhiteSpace();               // FIXME: This should be SkipWhitespaceAndComments(), current implementation will throw on comments
+            if (!CanReadCurrentChar)
+            {
+                return Symbol.EOI;
+            }
+            LastStart = CurrentPosition(0);
+            LastEnd = CurrentPosition(1);   // FIXME: Verify that Roslyn needs to be one character behind the end, otherwise, we don't need LastEnd and we don't need int increment in CurrentPosition
```
pavel-mikula-sonarsource, line 76: This should be verified and updated in the PR that will use this in the rules

analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        public LinePosition CurrentPosition(int increment) =>
+            new LinePosition(line, position + increment);
+
+        public Symbol NextSymbol()
+        {
+            Value = null;
+            NextPosition(false);
+            SkipWhiteSpace();               // FIXME: This should be SkipWhitespaceAndComments(), current implementation will throw on comments
```
pavel-mikula-sonarsource, line 70: This should be implemented in 2nd PR on the feaure branch

analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
```diff
@@ -0,0 +1,313 @@
...
+                {
+                    lines.Add(sb.ToString());
+                    sb.Clear();
+                }
+                else
+                {
+                    sb.Append(c);
+                }
+            }
+            lines.Add(sb.ToString());
```
pavel-mikula-sonarsource, line 60: Using `Span<string>` instead should save some memory, but I'm not sure if we can use it in our codebase

analyzers/src/SonarAnalyzer.Common/Json/Parsing/Enums.cs
```diff
@@ -0,0 +1,43 @@
...
+        EOI,
+        // Special
+        OpenCurlyBracket,
+        CloseCurlyBracket,
+        OpenSquareBracket,
+        CloseSquareBracket,
+        Comma,
+        Colon,
+        // Terms
+        Value   // String, Number (Integer, Decimal, Double), Null, True, False
```
Corniel, line 41: Is this intended as documentation? (If so, why not use `/// <summary>` or as TODO? (in that case, mark it as such).

analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
Corniel: What are the objections? (Just curious).

pavel-mikula-sonarsource: There are no objections, as it is a clear winner. But we target `net46` as a technical limitation.

analyzers/src/SonarAnalyzer.Common/Json/Parsing/Enums.cs
pavel-mikula-sonarsource: This is just an explanatory comment for me

analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
Corniel: Oops, fair point.

andrei-epure-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.Common/Json/JsonException.cs
```diff
@@ -21,13 +21,14 @@
...
 using System.Runtime.Serialization;
+using Microsoft.CodeAnalysis.Text;
 
 namespace SonarAnalyzer.Json
 {
     [Serializable]
     public sealed class JsonException : Exception
     {
-        public JsonException(string message) : base(message) { }
+        public JsonException(string message, LinePosition position) : base(message + $" at line {position.Line + 1} position {position.Character + 1}") { }
```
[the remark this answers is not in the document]
Corniel: I would argue that it is not a JsonException, but a ParseError, or ParseException. but that is also nitpicking.

andrei-epure-sonarsource:
This has been merged to the feature branch so should not be in Validate Peach column because it won't be deployed.

PR #4890, opened 2021-09-10
Ensure we don't throw on AndPattern, OrPattern and NotPattern

csaba-sagi-sonarsource:
Partial fix for #4880


dmitry-golovinov-sonarsource (approved):
LGTM!

PR #4893, opened 2021-09-13
Fix S2302: do not fail on ObjectInitializerExpressions.

dmitry-golovinov-sonarsource:


csaba-sagi-sonarsource (changes requested):
This seems like a good fix, and please apply it to the VB part too, as the VB part can also have null for ArgumentList.

csaba-sagi-sonarsource (approved):
LGTM

PR #4901, opened 2021-09-16
Rule S3059: Types should not have members with visibility set higher than the type's visibility

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/TypeMemberVisibility.cs
```diff
@@ -0,0 +1,94 @@
...
+            var parentType = GetParentType(type);
+            if (parentType == null && type.Modifiers.AnyOfKind(SyntaxKind.InternalKeyword))
+            {
+                return type.DescendantNodes()
+                           .Where(node => node.IsAnyKind(MemberDeclarationKinds))
+                           .OfType<MemberDeclarationSyntax>()
+                           // We skip overridden methods since they need to keep the public visibility (if present)
+                           .Where(declaration => declaration.Modifiers().AnyOfKind(SyntaxKind.PublicKeyword) && !declaration.Modifiers().AnyOfKind(SyntaxKind.OverrideKeyword))
+                           .Select(declaration => declaration.Modifiers().Single(modifier => modifier.IsKind(SyntaxKind.PublicKeyword)).GetLocation())
+                           .ToList();
```
mary-georgiou-sonarsource, line 85: Not a huge performance hit but maybe `ToList` conversion is not needed here.
It should work with return type `IEnumerable`.

costin-zaharia-sonarsource: I've returned a list here to ensure that we have a single iteration. Otherwise both the call to `Any()` and `c.ReportDiagnosticWhenActive` will evaluate the Enumerable.

mary-georgiou-sonarsource:
LGTM

mary-georgiou-sonarsource (approved):

analyzers/tests/SonarAnalyzer.UnitTest/Rules/TypeMemberVisibilityTest.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace SonarAnalyzer.UnitTest.Rules
+{
+    [TestClass]
+    public class TypeMemberVisibilityTest
+    {
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void TypeMemberVisibility_CS() =>
+#if NET
+            Verifier.VerifyAnalyzer(@"TestCases\TypeMemberVisibility.cs", new TypeMemberVisibility(), new[] { CoreMetadataReference.SystemRuntimeSerializationFormatters });
```
csaba-sagi-sonarsource, line 38: We could make these tests more simple if we would not use `SystemRuntimeSerializationFormatters` . We can simply use  override GetHashCode or ToString for the example where this was needed.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/TypeMemberVisibility.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        public struct MyStruct { }
+
+        public enum MyEnum { }
+    }
+
+    internal sealed class UnsafeBinder : SerializationBinder
+    {
+        public override Type BindToType(string assemblyName, string typeName) => null; // Compliant - the overriden method is public
+    }
```
csaba-sagi-sonarsource, line 181: Could you please add the following case?
```
    public interface ISomeInterface
    {
        void DoSomething();
    }

    internal class SomeClass: ISomeInterface // FP?
    {
        public void DoSomething() // FP
        {
        }
    }
```
I would say that the case is a FP, do you agree with it?

csaba-sagi-sonarsource (changes requested):
Very nice, just one minor change suggested, and a potential FP to be fixed or documented.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/TypeMemberVisibilityTest.cs
costin-zaharia-sonarsource: Nice!

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/TypeMemberVisibility.cs
costin-zaharia-sonarsource: Nice find!

csaba-sagi-sonarsource (approved):
LGTM!

PR #4904, opened 2021-09-17
Use the correct variable name for build reason

costin-zaharia-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestContextHelper.cs
```diff
@@ -25,9 +25,9 @@ namespace SonarAnalyzer.UnitTest.Helpers
     internal static class TestContextHelper
     {
         public static bool IsAzureDevOpsContext =>
-            Environment.GetEnvironmentVariable("Build.Reason") != null;
+            Environment.GetEnvironmentVariable("BUILD_REASON") != null;
```
dmitry-golovinov-sonarsource, line 28: Lets use this variable in both places here

dmitry-golovinov-sonarsource (approved):
LGTM!

analyzers/tests/SonarAnalyzer.UnitTest/TestSuiteInitialization.cs
```diff
@@ -14,6 +15,8 @@ public class TestSuiteInitialization
         public static void AssemblyInit(TestContext context)
         {
             Console.WriteLine(@"Running tests initialization...");
+            Console.WriteLine(@$"Build reason: {Environment.GetEnvironmentVariable(TestContextHelper.BuildReason) ?? "Local build"}");
```
pavel-mikula-sonarsource, line 18: Maybe like this? 
```suggestion
            Console.WriteLine(@$"Build reason: {Environment.GetEnvironmentVariable(TestContextHelper.BuildReason) ?? "Not set / Local build"}");
```

Issue #2081, opened 2018-11-06
Improve analyzer performances using Roslyn Clr Heap Allocation Analyzer


Evangelink:
### Description

Installing the [Roslyn Clr Heap Allocation Analyzer](https://github.com/Microsoft/RoslynClrHeapAllocationAnalyzer) has flagged a lot of issues on our codebase. Some of them seem to behave like the hotspot rules (i.e. we have to review them) but others are actually raising interesting improvements we could apply to reduce boxing and un-needed allocations.

I think we should install the tool, review and fix issues.

Evangelink:
@costin-zaharia-sonarsource Just FYI, the idea was more to use this analyzer as some kind of "audit" tool from time to time to track some unnecessary instantiations or allocations that could have a perf impact. But globally the main slow down we used to have were the CFG, Symbolic Execution and the fact that we weren't using roslyn operations but instead combination of syntax analyzers and access to symbols.

PR #4911, opened 2021-09-20
Show block instruction in Roslyn LVA UTs

pavel-mikula-sonarsource:
This changes Roslyn LVA assertion error messages from

>  Expected Lva.LiveOut(block).Select(x => x.Name) to be a collection with 3 item(s) because Block #2081 , but {"boolParameter", "intParameter"}"
   
to

>  Expected Lva.LiveOut(block).Select(x => x.Name) to be a collection with 3 item(s) because Block #2 Method(0);, but {"boolParameter", "intParameter"}"

to make it easier to find affected block


dmitry-golovinov-sonarsource (approved):
LGTM! Please consider fixing code smells if that's possible

pavel-mikula-sonarsource:
I have `params` argument. That would lead to `Enumerable.Empty<Expected>().ToArray()` 😒

PR #4912, opened 2021-09-20
Use UniqueQueue in LVA

pavel-mikula-sonarsource:
Enqueuing predecessors after each LiveIn change tends to enqueue a single block multiple times when it is already in the queue. `Analyze()` then has to do some collection and HashSet manipulations to figure out that nothing has changed => no operation.

This PR should prevent that and improve the performance a little.

### DIFF
--- a/analyzers/src/SonarAnalyzer.CFG/Helpers/UniqueQueue.cs
+++ b/analyzers/src/SonarAnalyzer.CFG/Helpers/UniqueQueue.cs
@@ -1,0 +1,53 @@
+﻿/*
+ * SonarAnalyzer for .NET
+ * Copyright (C) 2015-2021 SonarSource SA
+ * mailto: contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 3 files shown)


analyzers/src/SonarAnalyzer.CFG/Helpers/UniqueQueue.cs
```diff
@@ -0,0 +1,53 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections;
+using System.Collections.Generic;
+
+namespace SonarAnalyzer.CFG.Helpers
+{
+    public class UniqueQueue<T> : IEnumerable<T>
```
dmitry-golovinov-sonarsource, line 26: Let's make it internal to not introduce interesting backlinks to this part of the code from the rules that have nothing to do with CFGs at all.

dmitry-golovinov-sonarsource (approved):
LGTM! See my minor suggestion

PR #4915, opened 2021-09-21
Fix S1172 FPs by using Roslyn CFG

### LINKED ISSUE #4096: Fix S1172 FP: Parameter used in catch block

### LINKED ISSUE #4199: False positive on S1172 when using async lambdas an local functions
### Description

Function parameter is marked as unused while it's used through a lambda and a local function.

### Repro steps

Here is a simple case:

```csharp
private Task DoSomethingAsync(string text) // 'text' is reported as unused 
{
    return Task.Run(async () => await UseAsync());

    async Task UseAsync()
    {
        await Task.Delay(100);
        Console.WriteLine(text); // while it's used there
    }
}
```

### Known workarounds

We can pass the parameter explicitly to the local function:

```csharp
private Task DoSomethingAsync(string text)
{
    return Task.Run(async () => await UseAsync(text));

    async Task UseAsync(string text1)
    {
        await Task.Delay(100);

        Console.WriteLine(text1);
    }
}
```

The problem also disappears if we remove the lambda:

```csharp
private async Task DoSomethingAsync(string text)
{
    await UseAsync();

    async Task UseAsync()
    {
        await Task.Delay(100);
        Console.WriteLine(text);
    }
}
```

### Related information

* C#/VB.NET Plugins version 8.20.0.28934
* Visual Studio version 16.9.2
* MSBuild / dotnet version 5.0
* Operating System Windows 10

### LINKED ISSUE #4377: S1172 FP when parameter is invoked inside expression inside local function
### Description

Found on [sonarcloud](https://sonarcloud.io/project/issues?id=sonaranalyzer-dotnet&issues=AXk7yuoYe9K6Y2sWNErF&open=AXk7yuoYe9K6Y2sWNErF). Similar to #4199.

### Repro steps

```
        private static bool Foo(IEnumerable<int> a, int b) // Noncompliant FP
        {
            bool InsideFoo(int x) => x.Equals(b);
            bool CallInsideFoo(IEnumerable<int> numbers) => numbers.Any(x => false | InsideFoo(x));

            return CallInsideFoo(a);
        }
```

### Expected behavior

It should detect the parameter is used inside the expression in the local function.

### Related information

pavel-mikula-sonarsource:
Fixes #4096
Fixes #4377
Fixes #4199


dmitry-golovinov-sonarsource (commented):
Please fix the tests

pavel-mikula-sonarsource:
@dmitry-golovinov-sonarsource This is a draft now. The root cause of the IT diff is fixed in #4922. The last Squash commit doesn't need a review for the same reason.

dmitry-golovinov-sonarsource (approved):
LGTM!

PR #4916, opened 2021-09-21
Skip JSON comments

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
```diff
@@ -29,6 +29,7 @@ namespace SonarAnalyzer.Json.Parsing
 {
     internal class LexicalAnalyzer
     {
+        private const int MultiLineCommentTokenLength = 2;
```
pavel-mikula-sonarsource, line 32: This is needed only in one method => could be defined in that method.

analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
```diff
@@ -129,25 +130,78 @@ private void NextPosition(bool checkEOI = true)
...
+                else if (IsSingleLineComment())
+                {
+                    // We just need to read starting from the next line
+                    SeekToNextLine(false);
+                }
+                else if (IsCloseMultiLineComment())
+                {
+                    isInMultiLineComment = false;
+                    // need to skip */ part of the comment
+                    NthPosition(MultiLineCommentTokenLength, false);
```
pavel-mikula-sonarsource, line 177: Is this legal?
```
*
/
```
If yes, let's add UT.
If no, we can do just `position += MultiLineCommentTokenLength` instead and remove `NthPosition` as that one does too many things.

We could also do this instead, for simplicity (remove the constant and `NthPosition`) - I find it more direct and easier to understand:
```
NextPosition(false); // Skip *
NextPosition(false); // Skip /
```

analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
```diff
@@ -129,25 +130,78 @@ private void NextPosition(bool checkEOI = true)
...
+                    isInMultiLineComment = true;
+                    NextPosition(); // we still want to fail on non-closed comments
+                }
+                else
+                {
+                    return;
+                }
             }
+
+            char? NextCharSameLine() =>
```
pavel-mikula-sonarsource, line 195: Let's put EOL between local functions

analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
```diff
@@ -129,25 +130,78 @@ private void NextPosition(bool checkEOI = true)
...
+                    // We just need to read starting from the next line
+                    SeekToNextLine(false);
+                }
+                else if (IsCloseMultiLineComment())
+                {
+                    isInMultiLineComment = false;
+                    // need to skip */ part of the comment
+                    NthPosition(MultiLineCommentTokenLength, false);
+                }
+                else if (IsInMultiLineComment())
```
pavel-mikula-sonarsource, line 179: For simplicity, because it's actually `IsStartOrInMultiLineComment`
```suggestion
                else if (IsMultiLineComment())
```

analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
```diff
@@ -129,25 +130,78 @@ private void NextPosition(bool checkEOI = true)
...
+                if (IsWhitespace())
+                {
+                    NextPosition(false);
+                }
+                else if (IsSingleLineComment())
+                {
+                    // We just need to read starting from the next line
+                    SeekToNextLine(false);
+                }
+                else if (IsCloseMultiLineComment())
```
pavel-mikula-sonarsource, line 173: `Close` resonatest as `Near` here a little
```suggestion
                else if (IsEndOfMultiLineComment())
```

analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
```diff
@@ -129,25 +130,78 @@ private void NextPosition(bool checkEOI = true)
...
+                    // need to skip */ part of the comment
+                    NthPosition(MultiLineCommentTokenLength, false);
+                }
+                else if (IsInMultiLineComment())
+                {
+                    if (!isInMultiLineComment)
+                    {
+                        // need to skip /* part of the comment
+                        NthPosition(MultiLineCommentTokenLength, false);
+                    }
```
pavel-mikula-sonarsource, line 185: Let's add case with `/*/` to `InvalidInput_ThrowsJsonException` to prove that this if is really needed

analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
```diff
@@ -129,25 +130,78 @@ private void NextPosition(bool checkEOI = true)
...
+                else
+                {
+                    return;
+                }
             }
+
+            char? NextCharSameLine() =>
+                (position + 1 < lines[line].Length) ? lines[line][position + 1] : (char?)null;
+            bool IsWhitespace() =>
+                char.IsWhiteSpace(CurrentChar);
```
pavel-mikula-sonarsource, line 198: I don't think we need this `IsWhitespace`, the body itself is clear enough

pavel-mikula-sonarsource (changes requested):
Some minor stuff

analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
dmitry-golovinov-sonarsource: No this is not legal. Actually, comments are not allowed in all JSON, we use it to support highlighting of our issues within JSON files, that makes it easy.

dmitry-golovinov-sonarsource: Will do the suggested NextPosition way. It was like this before, but it looked smelly. But since you find it explanatory, I will leave it there

analyzers/src/SonarAnalyzer.Common/Json/Parsing/LexicalAnalyzer.cs
dmitry-golovinov-sonarsource: Good point, thanks

pavel-mikula-sonarsource (approved):
LGTM

pavel-mikula-sonarsource:
@dmitry-golovinov-sonarsource I've converted this to draft to prevent merging before #4888

PR #4917, opened 2021-09-22
Configure CS8032 as Error in ITs

pavel-mikula-sonarsource:
.NET ITs silently produced this

>Warning CS8032: An instance of analyzer SonarAnalyzer.Rules.CSharp.TokenTypeAnalyzer cannot be created from C:\Projects\sonar-dotnet\analyzers\its\binaries\SonarAnalyzer.CSharp.dll : Could not load file or assembly 'Google.Protobuf, Version=3.6.1.0, Culture=neutral, PublicKeyToken=a7d26565bac4d604' or one of its dependencies. The system cannot find the file specified.

because `Google.Protobuf.dll` was not part of `Analyzers` property in targets file.

### DIFF
--- a/analyzers/its/AllSonarAnalyzerRules.ruleset
+++ b/analyzers/its/AllSonarAnalyzerRules.ruleset
@@ -5,6 +5,11 @@
     <Rule Id="AD0001" Action="Error" />
   </Rules>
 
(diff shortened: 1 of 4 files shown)


andrei-epure-sonarsource (approved):
LGTM

PR #4921, opened 2021-09-23
Limit the version of Roslyn to the one that creates fully supported CFGs

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.CFG/Roslyn/ControlFlowGraph.cs
```diff
@@ -60,7 +60,7 @@ public class ControlFlowGraph
 
         static ControlFlowGraph()
         {
-            if (RoslynHelper.FlowAnalysisType("ControlFlowGraph") is { } type)
+            if (RoslynHelper.RoslynVersionIsSupported() && RoslynHelper.FlowAnalysisType("ControlFlowGraph") is { } type)
             {
                 IsAvailable = true;
```
pavel-mikula-sonarsource, line 65: We actually don't need the 2nd part as a condition. The new version check is all that matters.
```suggestion
            if (RoslynHelper.RoslynVersionIsSupported() && RoslynHelper.FlowAnalysisType("ControlFlowGraph") is { } type)
            {
                IsAvailable = true;
                var type = RoslynHelper.FlowAnalysisType("ControlFlowGraph");
```

analyzers/src/SonarAnalyzer.CFG/Helpers/RoslynHelper.cs
```diff
@@ -29,6 +29,11 @@ namespace SonarAnalyzer.CFG.Helpers
 {
     internal static class RoslynHelper
     {
+        private const int MinimalSupportedMajorVersion = 3;
+
+        public static bool RoslynVersionIsSupported(int minimalVersion = MinimalSupportedMajorVersion) =>
```
pavel-mikula-sonarsource, line 34: Let's add `Cfg` into the name, because everything is `Roslyn` around here.
```suggestion
        public static bool IsRoslynCfgSupported(int minimalVersion = MinimalSupportedMajorVersion) =>
```

pavel-mikula-sonarsource (changes requested):
Polishing

pavel-mikula-sonarsource (approved):
LGTM

PR #4925, opened 2021-09-27
Fix S3254 FP: Don't raise if the parameter isn't last

### LINKED ISSUE #4620: Fix S3254 FP: Don't raise if the parameter isn't last.
Don't raise when argument is followed by other non-default arguments

>Remove this default value assigned to parameter 'a'.
```
        public void Go()
        {
            WithOptional(false, false); // 2x True Positive
            WithOptional(false, true); // FP on the 1st argument
        }

        public void WithOptional(bool a = false, bool b = false) { }
```

dmitry-golovinov-sonarsource:
Fixes #4620

### DIFF
--- a/analyzers/its/expected/akka.net/Akka--netstandard2.0-S3254.json
+++ b/analyzers/its/expected/akka.net/Akka--netstandard2.0-S3254.json
@@ -45,19 +45,6 @@
 "location":  {
 "uri":  "sources\akka.net\src\core\Akka\Actor\CoordinatedShutdown.cs",
 "region":  {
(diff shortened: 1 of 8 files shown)


csaba-sagi-sonarsource (changes requested):
Looks good, however it would be nice to report for cases like M3(1, y: 4); and M3(x: 1, y: 4); from the UTs as by stating explicitly y:4 there is no need to provide the default value of 1 for x.

csaba-sagi-sonarsource (approved):
LGTM!

PR #4926, opened 2021-09-27
Fix S3433: FP on xUnit async void methods

### LINKED ISSUE #4861: S3433 FP on xUnit `async void` methods
### Description

https://sonarsource.github.io/rspec/#/rspec/S3433

The rule raises issues on MSTest, NUnit and xUnit test methods which are `async void`. For MSTest and NUnit it is correct - they don't work (NUnit will raise an error, MSTest won't run the method at all).

However, xUnit `async void` methods run as expected and are supported by the framework (tested with xUnit 2.4.1).

### Repro steps

Checkout https://github.com/andrei-epure-sonarsource/dotnet-test-playground commit fde6fc1dfab5f2ddcf93dc0e3a1508f5cb5f03a6.

Open `UnitTest1` in VS with SonarLint (tested with SLVS 4.37).

```
[Fact]
public async void xUnit_Async_Void() =>
    await new TestAsync().Foo(o => Assert.Equal(32, o));
```

### Expected behavior

`xUnit_Async_Void` should not raise an issue.

### Actual behavior

`Make this method non-async or return "Task"` is raised.

### Known workarounds

None.

### Related information

* C#/VB.NET Plugins version - 8.27
* Visual Studio version - 16.10.3

dmitry-golovinov-sonarsource:
Fixes #4861

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldHaveCorrectSignature.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/TestMethodShouldHaveCorrectSignature.cs
@@ -88,15 +88,15 @@ namespace SonarAnalyzer.Rules.CSharp
             // XUnit - note that local functions can be test methods, thus we skip checking the syntax node
             {
                 KnownType.Xunit_FactAttribute,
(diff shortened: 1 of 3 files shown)


csaba-sagi-sonarsource (approved):
LGTM!

PR #4927, opened 2021-09-27
Fix S1172 FP: Support reassignment in Roslyn LVA

pavel-mikula-sonarsource:


dmitry-golovinov-sonarsource (approved):
Looks good! Thanks for fixing that

PR #4931, opened 2021-09-28
Return CaptureId from ShimLayer operations

pavel-mikula-sonarsource:
We need to return shimmed `CaptureId` instead of `InvalidOpertationException` when invoking `IFlowCaptureOperationWrapper.Id` and `IFlowCaptureReferenceOperationWrapper.Id`


dmitry-golovinov-sonarsource (approved):
LGTM!

PR #4933, opened 2021-09-28
Fix S2190: FP when calling virtual method/virtual property

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.cs
```diff
@@ -104,8 +104,8 @@ private class RecursionContext<TControlFlowGraph>
 
             public TControlFlowGraph ControlFlowGraph { get; }
             public ISymbol AnalyzedSymbol { get; }
-            public Location IssueLocation { get; }
             public SemanticModel SemanticModel => analysisContext.SemanticModel;
+            private Location IssueLocation { get; }
```
pavel-mikula-sonarsource, line 108: Let's make it a field and move up then.

analyzers/src/SonarAnalyzer.CSharp/Rules/InfiniteRecursion.RoslynCfg.cs
```diff
@@ -100,10 +100,17 @@ protected override bool IsValid(BasicBlock block)
                     static ISymbol MemberSymbol(IOperation operation) =>
                         operation.Kind switch
                         {
-                            OperationKindEx.PropertyReference => IPropertyReferenceOperationWrapper.FromOperation(operation).Property,
-                            OperationKindEx.Invocation => IInvocationOperationWrapper.FromOperation(operation).TargetMethod,
+                            OperationKindEx.PropertyReference when IPropertyReferenceOperationWrapper.FromOperation(operation) is var propertyReference
+                                                                   && InstanceReferencesThis(propertyReference.Instance) =>
+                                propertyReference.Property,
```
pavel-mikula-sonarsource, line 105: Can we format it like this?
```
                            OperationKindEx.PropertyReference
                                when IPropertyReferenceOperationWrapper.FromOperation(operation) is var propertyReference && InstanceReferencesThis(propertyReference.Instance) =>
                                propertyReference.Property,
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/InfiniteRecursion.RoslynCfg.cs
```diff
@@ -497,6 +497,50 @@ public interface IWithDefaultImplementation
...
+            get
+            {
+                return Parent.Prop;
+            }
+        }
+
+        public virtual int Method()
+        {
+            return Parent.Method();
+        }
```
pavel-mikula-sonarsource, line 516: Let's also add a case like this, because it produces the FlowReference stuff
```
        public virtual int? NullableMethod()
        {
            return Parent?.NullableMethod();
        }
```

pavel-mikula-sonarsource (approved):
LGTM, few suggestions before merging

PR #4935, opened 2021-09-29
Fix S3604: FN after ?? or ?: due to the problem in CFG visitor logic

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.CFG/Roslyn/CfgAllPathValidator.cs
```diff
@@ -37,16 +44,32 @@ public abstract class CfgAllPathValidator
         public bool CheckAllPaths() =>
             IsBlockOrAllSuccessorsValid(this.cfg.EntryBlock);
 
-        private bool IsBlockOrAllSuccessorsValid(BasicBlock block) =>
-            !IsInvalid(block) && (IsValid(block) || AreAllSuccessorsValid(block));
+        private bool IsBlockOrAllSuccessorsValid(BasicBlock block)
```
pavel-mikula-sonarsource, line 47: What about some UT that would fail for the old logic and will succeed for the new one in `CfgAllPathValidatorTest.cs`?

analyzers/src/SonarAnalyzer.CFG/Roslyn/CfgAllPathValidator.cs
```diff
@@ -37,16 +44,32 @@ public abstract class CfgAllPathValidator
...
 
         private bool AreAllSuccessorsValid(BasicBlock block)
         {
-            this.visited.Add(block);
-
-            return !block.SuccessorBlocks.Contains(cfg.ExitBlock)
-                   && block.SuccessorBlocks.Except(visited).Any()
-                   && block.SuccessorBlocks.Except(visited).All(IsBlockOrAllSuccessorsValid);
+            lattice[block] = BlockResult.INCONCLUSIVE;
+            if (!block.SuccessorBlocks.Contains(cfg.ExitBlock) && block.SuccessorBlocks.Any())
```
pavel-mikula-sonarsource, line 57: Let's swap this. I find it easier to understand.
```suggestion
            if (block.SuccessorBlocks.Any() && !block.SuccessorBlocks.Contains(cfg.ExitBlock))
```

analyzers/src/SonarAnalyzer.CFG/Roslyn/CfgAllPathValidator.cs
```diff
@@ -37,16 +44,32 @@ public abstract class CfgAllPathValidator
         public bool CheckAllPaths() =>
             IsBlockOrAllSuccessorsValid(this.cfg.EntryBlock);
 
-        private bool IsBlockOrAllSuccessorsValid(BasicBlock block) =>
-            !IsInvalid(block) && (IsValid(block) || AreAllSuccessorsValid(block));
+        private bool IsBlockOrAllSuccessorsValid(BasicBlock block)
+        {
```
pavel-mikula-sonarsource, line 48: Line 45 above: `this.` can be removed

analyzers/src/SonarAnalyzer.CFG/Roslyn/CfgAllPathValidator.cs
```diff
@@ -37,16 +44,32 @@ public abstract class CfgAllPathValidator
...
-            return !block.SuccessorBlocks.Contains(cfg.ExitBlock)
-                   && block.SuccessorBlocks.Except(visited).Any()
-                   && block.SuccessorBlocks.Except(visited).All(IsBlockOrAllSuccessorsValid);
+            lattice[block] = BlockResult.INCONCLUSIVE;
+            if (!block.SuccessorBlocks.Contains(cfg.ExitBlock) && block.SuccessorBlocks.Any())
+            {
+                foreach (var successorBlock in block.SuccessorBlocks)
+                {
+                    if ((lattice.ContainsKey(successorBlock) && lattice[successorBlock] == BlockResult.FALSE)
+                        || (!lattice.ContainsKey(successorBlock) && !IsBlockOrAllSuccessorsValid(successorBlock)))
```
pavel-mikula-sonarsource, line 62: Isn't this going to cause infinite recursion when there's a loop to self for a block that has `IsInvalid=false` and `IsValid=false`?

analyzers/src/SonarAnalyzer.CFG/Roslyn/CfgAllPathValidator.cs
```diff
@@ -37,16 +44,32 @@ public abstract class CfgAllPathValidator
...
-
-            return !block.SuccessorBlocks.Contains(cfg.ExitBlock)
-                   && block.SuccessorBlocks.Except(visited).Any()
-                   && block.SuccessorBlocks.Except(visited).All(IsBlockOrAllSuccessorsValid);
+            lattice[block] = BlockResult.INCONCLUSIVE;
+            if (!block.SuccessorBlocks.Contains(cfg.ExitBlock) && block.SuccessorBlocks.Any())
+            {
+                foreach (var successorBlock in block.SuccessorBlocks)
+                {
+                    if ((lattice.ContainsKey(successorBlock) && lattice[successorBlock] == BlockResult.FALSE)
```
pavel-mikula-sonarsource, line 61: I don't see `TRUE` to be ever read and used. So we might not need that state.

_Thinking out loud:_ What does it actually mean to have `TRUE` there? I see it the same as `INCONCLUSIVE`. So when does the algorithm stop on loops?

pavel-mikula-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CFG/Roslyn/CfgAllPathValidator.cs
```diff
@@ -25,8 +25,15 @@ namespace SonarAnalyzer.CFG.Roslyn
 {
     public abstract class CfgAllPathValidator
     {
+        private enum BlockResult
+        {
+            INCONCLUSIVE,
+            TRUE,
+            FALSE,
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: I can, but we've agreed that this is more explicit.

analyzers/src/SonarAnalyzer.CFG/Roslyn/CfgAllPathValidator.cs
dmitry-golovinov-sonarsource: Discussed online

analyzers/src/SonarAnalyzer.CFG/Roslyn/CfgAllPathValidator.cs
```diff
@@ -35,18 +35,34 @@ public abstract class CfgAllPathValidator
...
+            {
+                foreach (var successorBlock in block.SuccessorBlocks)
+                {
+                    if ((visitedStatus.ContainsKey(successorBlock) && visitedStatus[successorBlock])
+                        || (!visitedStatus.ContainsKey(successorBlock) && !IsBlockOrAllSuccessorsValid(successorBlock)))
+                    {
+                        return false;
+                    }
+                }
+                return true;
```
pavel-mikula-sonarsource, line 60: Check this out :) 
1) The `if` condition can be simplified as `visitedStatus.ContainsKey(successorBlock) ? visitedStatus[successorBlock] : !IsBlockOrAllSuccessorsValid(successorBlock)`
2) The `foreach` can be replaced with 
```
return !block.SuccessorBlocks.Any(x => visitedStatus.ContainsKey(x) ? visitedStatus[x] : !IsBlockOrAllSuccessorsValid(x));
```
3) We can merge the `if`s
4) We can swap `!Any` to `All` WHEN we align the state with the `isValid` like this `visitedStatus[block] = isValid;` (removing negation) 
```
return block.SuccessorBlocks.Any()
    && !block.SuccessorBlocks.Contains(cfg.ExitBlock)
    && block.SuccessorBlocks.All(x => x != cfg.ExitBlock && (visitedStatus.ContainsKey(x) ? visitedStatus[x] : IsBlockOrAllSuccessorsValid(x)));
```
5) _Optional:_ We can merge `Contains` and `All` actually
```
visitedStatus[block] = true; // protects from loops, don't fail the computation if hits itself
return block.SuccessorBlocks.Any()
    && block.SuccessorBlocks.All(x => x != cfg.ExitBlock && (visitedStatus.ContainsKey(x) ? visitedStatus[x] : IsBlockOrAllSuccessorsValid(x)));
```

analyzers/src/SonarAnalyzer.CFG/Roslyn/CfgAllPathValidator.cs
```diff
@@ -35,18 +35,34 @@ public abstract class CfgAllPathValidator
...
+        }
 
         private bool AreAllSuccessorsValid(BasicBlock block)
         {
-            this.visited.Add(block);
-
-            return !block.SuccessorBlocks.Contains(cfg.ExitBlock)
-                   && block.SuccessorBlocks.Except(visited).Any()
-                   && block.SuccessorBlocks.Except(visited).All(IsBlockOrAllSuccessorsValid);
+            visitedStatus[block] = false; // protects from loops, don't fail the computation if hits itself
```
pavel-mikula-sonarsource, line 49: Both UTs in CfgAllPathValidatorTest` are green even if I change this to `true` 😒 Can you craft one that would fail?

pavel-mikula-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CFG/Roslyn/CfgAllPathValidator.cs
```diff
@@ -35,18 +35,21 @@ public abstract class CfgAllPathValidator
...
         private bool AreAllSuccessorsValid(BasicBlock block)
         {
-            this.visited.Add(block);
-
-            return !block.SuccessorBlocks.Contains(cfg.ExitBlock)
-                   && block.SuccessorBlocks.Except(visited).Any()
-                   && block.SuccessorBlocks.Except(visited).All(IsBlockOrAllSuccessorsValid);
+            visitedStatus[block] = true; // protects from loops, don't fail the computation if hits itself
+            return block.SuccessorBlocks.Any()
+                   && !block.SuccessorBlocks.Contains(cfg.ExitBlock)
```
pavel-mikula-sonarsource, line 51: This should go away, the condition is embedded in the next line
```suggestion
```

PR #4939, opened 2021-09-30
Add RemoveParentheses to SyntaxFacade

Corniel:
As discussed in #4805

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
@@ -24,7 +24,6 @@ using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using SonarAnalyzer.Extensions;
-using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Helpers.Facade
 {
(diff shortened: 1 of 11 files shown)


analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
```diff
@@ -81,5 +82,17 @@ node switch
                 LiteralExpressionSyntax literalExpression => literalExpression.Token.ValueText,
                 _ => string.Empty
             };
+
+        public override SyntaxNode RemoveParentheses(SyntaxNode expression)
```
pavel-mikula-sonarsource, line 86: This should just call the `expression.RemoveParentheses()` extension method to avoid duplicate implementation of the same thing.

analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
```diff
@@ -81,5 +82,17 @@ node switch
...
+        {
+            var current = expression;
+            while (current?.IsAnyKind(SyntaxKind.ParenthesizedExpression, SyntaxKindEx.ParenthesizedPattern) ?? false)
```
pavel-mikula-sonarsource, line 89: The old implementation `? .. ?? false` is somehow overengineered. Can you please update it to `current != null && current.IsAnyKind()) to simplify it?

pavel-mikula-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
Corniel: That is fine.

analyzers/src/SonarAnalyzer.CSharp/Facade/CSharpSyntaxFacade.cs
Corniel: Done both.

PR #4947, opened 2021-10-04
Update RSPEC

pavel-mikula-sonarsource:
There are quite a few changes after rule-api 2.1.0 release.

There's an open question on appsec side if VB part of S4784 should be undeprecated. For now, I'm not changing the state.


dmitry-golovinov-sonarsource (approved):
LGTM!

Issue #4949, opened 2021-10-04
Fix S1854 FPs: Improve catch block links


pavel-mikula-sonarsource:
Similar to #4948 

Roslyn LVA needs to be updated for better predecessors/successors computation.

`catch` should link to all blocks in `try` to propagate LiveIn through the loop.
```C#
public void TryReturns_Loop()
{
    var counter = 0;
    while (counter < 5)
    {
        counter++;  // Noncompliant FP related to LVA
        try
        {
            SomethingThatCanThrow();
            return;
        }
        catch (TimeoutException)
        {
            // Continue loop to the next try
        }
    }
}
```

Issue #4948, opened 2021-10-04
Fix S1854 FP: Value used in finally should LiveIn after throw


pavel-mikula-sonarsource:
Should not raise in complex scenarios related to Roslyn LVA:

Finally should link predecessors/sucessors to all blocks of try.
```C#
public void UsedInFinally_Throw()
{
    var value = 42; // Noncompliant FP related to LVA
    try
    {
        throw new Exception();
    }
    finally
    {
        Use(value);
    }
}
```

PR #4950, opened 2021-10-04
S1854: Add FP repro for #4948 and #4949

pavel-mikula-sonarsource:
Reproduces #4948
Reproduces #4949

This is a draft to prevent merging before release


dmitry-golovinov-sonarsource (approved):
LGTM!

PR #4953, opened 2021-10-05
Rule S2437: false positive with local fields and properties

### LINKED ISSUE #4774: Rule S2437: FP with local fields
### Description

Rule S2437 is raising false positives when one of the operators is a local field initialized with an expected value (0 or -1).

### Repro steps

```
class LocalFields
{
    private int start = 0;

    public void UpdateStart(int val) => start = val;

    public override int GetHashCode()
    {
        return Method() ^ start; // Noncompliant - FP: start is a local field which can be updated to have a value != 0

        int Method() => 42;
    }
}
```

### Expected behavior

The issue should not be raised.

### Actual behavior

Issue is raised on the line with `Noncompliant` comment.

### Known workarounds

None.

### Related information

dmitry-golovinov-sonarsource:
Fixes #4774


dmitry-golovinov-sonarsource:
DO NOT MERGE!

analyzers/src/SonarAnalyzer.CSharp/Rules/SillyBitwiseOperation.cs
```diff
@@ -76,5 +76,10 @@ private void CheckBinary(SyntaxNodeAnalysisContext context, int constValueToLook
             var binary = (BinaryExpressionSyntax)context.Node;
             CheckBinary(context, binary.Left, binary.OperatorToken, binary.Right, constValueToLookFor);
         }
+
+        private static bool IsFieldOrProperty(SemanticModel semanticModel, SyntaxNode node) =>
+            semanticModel.GetSymbolInfo(node).Symbol is { } symbol
+            && symbol.ContainingNamespace.Name != "System"  // Excluding built-in constants
```
costin-zaharia-sonarsource, line 82: I don't understand this case. Can you give an example?

dmitry-golovinov-sonarsource: Sure
`
resultNint = bitMaskNint | IntPtr.Zero;   // Noncompliant`

dmitry-golovinov-sonarsource: Where  IntPtr belongs to System namespace, but Zero is a field there

```
[Intrinsic]
public static readonly IntPtr Zero;
```

analyzers/src/SonarAnalyzer.CSharp/Rules/SillyBitwiseOperation.cs
```diff
@@ -78,8 +78,10 @@ private void CheckBinary(SyntaxNodeAnalysisContext context, int constValueToLook
         }
 
         private static bool IsFieldOrProperty(SemanticModel semanticModel, SyntaxNode node) =>
-            semanticModel.GetSymbolInfo(node).Symbol is { } symbol
-            && symbol.ContainingNamespace.Name != "System"  // Excluding built-in constants
-            && symbol.Kind is SymbolKind.Field or SymbolKind.Property;
+            semanticModel.GetSymbolInfo(node).Symbol is { Kind: SymbolKind.Field or SymbolKind.Property } symbol
+            && !FieldOrPropertyBelongToSystemNamespace(symbol);
```
costin-zaharia-sonarsource, line 82: I think we should move the `FieldOrPropertyBelongToSystemNamespace` call to `FindConstant` since it's a bit misleading. Even if it's in the system namespace it is still a field or property.

dmitry-golovinov-sonarsource: Done

analyzers/src/SonarAnalyzer.CSharp/Rules/SillyBitwiseOperation.cs
```diff
@@ -76,5 +78,14 @@ private void CheckBinary(SyntaxNodeAnalysisContext context, int constValueToLook
...
             CheckBinary(context, binary.Left, binary.OperatorToken, binary.Right, constValueToLookFor);
         }
+
+        private static bool IsFieldOrProperty(SemanticModel semanticModel, SyntaxNode node, out ISymbol symbol)
+        {
+            symbol = semanticModel.GetSymbolInfo(node).Symbol;
+            return symbol is {Kind: SymbolKind.Field or SymbolKind.Property};
+        }
+
+        private static bool FieldOrPropertyBelongToSystemNamespace(ISymbol symbol) =>
```
costin-zaharia-sonarsource, line 88: `IsInSystemNamespace`, WDYT?

costin-zaharia-sonarsource (approved):
LGTM!

PR #4958, opened 2021-10-07
Rule S3260: FP static classes can not be marked as sealed

Corniel:
Simple fix for a false positive: (private) static classes can not be marked as sealed.


costin-zaharia-sonarsource:
It looks like the integration tests fail since the expected results are now different. I'm not sure if you can run them and then the update script. If not, I can do this.

Corniel:
> It looks like the integration tests fail since the expected results are now different. I'm not sure if you can run them and then the update script. If not, I can do this.

Not in my powers.

PR #4971, opened 2021-10-13
Introduction of a VerifyBuilder

Corniel:
The SonarAnalyzer TestFramework is awesome. It contains a lot of options to pick when verifying an analyzer.

However, given the enormous number of overloads makes it a bit clumsy. By providing a builder usage could improve:

``` C#
[TestMethod]
public void BreakOutsideSwitch() =>
    new BreakOutsideSwitch().AddSource(@"TestCases\BreakOutsideSwitch.cs").Verify();
```

And a more complex example:

``` C#
[TestMethod]
public void DeclareTypesInNamespaces_CS_Before8() =>
    new CS.DeclareTypesInNamespaces()
    .AddSource(@"TestCases\DeclareTypesInNamespaces.BeforeCSharp8.cs")
    .WithOptions(ParseOptionsHelper.BeforeCSharp8)
    .Verify();
```

### DIFF
--- a/analyzers/rspec/vbnet/S4784_vb.net.html
+++ b/analyzers/rspec/vbnet/S4784_vb.net.html
@@ -82,6 +82,4 @@ End Namespace
   </li>
   <li> OWASP Regular expression Denial of Service - ReDoS </li>
 </ul>
-<h2>Deprecated</h2>
-<p>This rule is deprecated, and will eventually be removed.</p>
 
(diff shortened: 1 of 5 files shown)


Corniel:
@costin-zaharia-sonarsource I'm not aware that I closed this one. (although Github states so, so it is probably true) Can you re-open it?

costin-zaharia-sonarsource:
Sure

Corniel:
I see, strange. Anyway, it is about the approach.

costin-zaharia-sonarsource:
> Anyway, it is about the approach.

I like the idea to have a fluent interface for our verifier (like the one suggested by you). We plan to have an internal discussion about this when we have some time and then see how to proceed further.

PR #4986, opened 2021-10-18
Refactor S4226 and S4225 to improve perf 

andrei-epure-sonarsource:
Minor refactoring to do AST checks before querying the symbol.
The rules are not in SonarWay.


andrei-epure-sonarsource:
I validated using peach times and the improvement works.

PR #5019, opened 2021-11-01
Add test cases for Utility analyzers

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Metrics/CSharpMetrics.cs
```diff
@@ -161,8 +162,13 @@ protected override bool IsStatement(SyntaxNode node)
                 case SyntaxKind.YieldReturnStatement:
                     return true;
 
-                default:
+                case SyntaxKind.Block:
                     return false;
+
+                default:
```
csaba-sagi-sonarsource, line 168: This will help us catch if we don`t support some syntax, on the other side it may create additional issues for the users.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Utilities/MetricsAnalyzer/AllMetrics.cs
```diff
@@ -15,7 +15,15 @@ public int Go(bool condition)
...
                 return 42;
             }
+
+            var y = string.Empty;
+            (y, var z) = ("a", 'x');
+
             throw new NotImplementedException();
         }
     }
+
```
csaba-sagi-sonarsource, line 25: Could you please add a simple struct, I see that the class count in the tests is now 3 which is a bit confusing. As the record struct is technically not a class.

csaba-sagi-sonarsource (approved):
LGTM!

analyzers/src/SonarAnalyzer.CSharp/Metrics/CSharpMetrics.cs
costin-zaharia-sonarsource: Indeed. The purpose is to fail as fast as possible to know when we don't compute the metrics correctly. This is also a way to check if our safety net is good enough or if we have to improve it.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Utilities/MetricsAnalyzer/AllMetrics.cs
costin-zaharia-sonarsource: sure.

PR #5031, opened 2021-11-08
Rule S3260: FP when nested private class inherits from outer private class.

### LINKED ISSUE #4983: S3260 does not detect if the class is inherited by a sub-class
### Description

S3260 does not detect if the class is inherited by a sub-class.

### Repro steps

Just paste this example:

```csharp
internal class Test
{
    private class PropertyFetch // S3260 Private classes or records which are not derived in the current assembly should be marked as 'sealed'.
    {
        public virtual bool Fetch() => false;

        private sealed class CustomPropertyFetch : PropertyFetch
        {
            public override bool Fetch() => true;
        }
    }
}
```
And observe that S3260 is displayed while it shouldn't. 
Anyway the class can't be sealed as it has a virtual method.

### Known workarounds

Set the class as abstract (might not be applicable).

### Related information

```xml
<PackageReference Include="SonarAnalyzer.CSharp" Version="8.30.0.37606" />
```

mary-georgiou-sonarsource:
Fixes #4983
Related PR for virtual members FP: #5027

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/NonDerivedPrivateClassesShouldBeSealed.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/NonDerivedPrivateClassesShouldBeSealed.cs
@@ -26,7 +26,6 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
 using StyleCop.Analyzers.Lightup;
-using System.Collections.Generic;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/NonDerivedPrivateClassesShouldBeSealed.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/NonDerivedPrivateClassesShouldBeSealed.cs
@@ -26,6 +26,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
 using StyleCop.Analyzers.Lightup;
+using System.Collections.Generic;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/NonDerivedPrivateClassesShouldBeSealed.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/NonDerivedPrivateClassesShouldBeSealed.cs
@@ -47,7 +47,6 @@ namespace SonarAnalyzer.Rules.CSharp
                 if (IsPrivateButNotSealedType(typeDeclarationSyntax) && !HasVirtualMembers(typeDeclarationSyntax))
                 {
                     var nestedPrivateTypeInfo = (INamedTypeSymbol)c.SemanticModel.GetDeclaredSymbol(c.Node);
-
                     if (!IsPrivateTypeInherited(nestedPrivateTypeInfo))
                     {
                         c.ReportIssue(Diagnostic.Create(Rule, typeDeclarationSyntax.Identifier.GetLocation()));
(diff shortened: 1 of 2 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataFactory.cs
```diff
@@ -41,6 +41,7 @@ public static partial class NuGetMetadataFactory
             "net",
             "netstandard2.1",
             "netstandard2.0",
+            "net6.0",
```
andrei-epure-sonarsource, line 44: wasn't this done by Pavel's PR?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NonDerivedPrivateClassesShouldBeSealed.cs
```diff
@@ -50,16 +50,34 @@ private class PrivateDerivedClassSecondExtension : PrivateDerivedClass { }
 
         private class TheThirdExtension : PrivateDerivedClassSecondExtension { } // Noncompliant
 
-        private class PrivateClass // Noncompliant
```
andrei-epure-sonarsource, line 53: was this previously an FP?

analyzers/tests/SonarAnalyzer.UnitTest/Rules/UnusedPrivateMemberTest.cs
```diff
@@ -249,7 +249,7 @@ public void UnusedPrivateMember_Performance()
         {
             Action verifyAnalyzer = () => Verifier.VerifyAnalyzer(new[] {@"TestCases\UnusedPrivateMember.Performance.cs"},
                                                                   new UnusedPrivateMember(),
-                                                                  GetEntityFrameworkCoreReferences(Constants.NuGetLatestVersion));
+                                                                  GetEntityFrameworkCoreReferences("5.0.12")); // Latest before 6.0.0 for .NET 6 that has Linq versioning collision issue
```
andrei-epure-sonarsource, line 252: this seems unrelated

andrei-epure-sonarsource (changes requested):
could you please clean the commit history as it seems there are unrelated changes?

otherwise it looks good

analyzers/src/SonarAnalyzer.CSharp/Rules/NonDerivedPrivateClassesShouldBeSealed.cs
```diff
@@ -26,6 +26,7 @@
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
 using StyleCop.Analyzers.Lightup;
+using System.Collections.Generic;
```
andrei-epure-sonarsource: I don't understand this commit...

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NonDerivedPrivateClassesShouldBeSealed.cs
```diff
@@ -50,16 +50,34 @@ private class PrivateDerivedClassSecondExtension : PrivateDerivedClass { }
...
+        private class PrivateExternalA { } // Compliant, it's extended in 'ExtendsExternalPrivateClass'
+        private class PrivateExternalB { } // Compliant, it's extended in 'ExtendsExternalPrivateClass'
```
andrei-epure-sonarsource, line 54:
```suggestion
        private class PrivateExternalB { } // Compliant, it's extended in 'SeriouslyNestedClass'
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NonDerivedPrivateClassesShouldBeSealed.cs
mary-georgiou-sonarsource: This particular line no, because this `PrivateClass` was not inherited in any case.

mary-georgiou-sonarsource:
Rebase issue has been fixed.

analyzers/src/SonarAnalyzer.CSharp/Rules/NonDerivedPrivateClassesShouldBeSealed.cs
mary-georgiou-sonarsource: Rebase issue has been fixed.

andrei-epure-sonarsource (approved):
LGTM

PR #5037, opened 2021-11-10
S4581: Issue is raised for implicit constructor and default expressions.

### LINKED ISSUE #4459: S4581: FN with default and default(Guid)
### Description

No issue is raised if default or default(Guid) is used instead of new Guid().

### Repro steps

```
public void Foo()
{
    var guid = default(Guid); // Noncompliant
    Guid other = default; // Noncompliant
}
```

### Expected behavior

Issue should be raised for both default and default(Guid).

### Actual behavior

No issue is raised.

### Related information

* C#/VB.NET Plugins version : 8.22.0.31243

mary-georgiou-sonarsource:
Fixes #4459

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/NewGuidShouldNotBeUsed.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/NewGuidShouldNotBeUsed.cs
@@ -25,6 +25,8 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.Wrappers;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
@@ -35,25 +37,23 @@ namespace SonarAnalyzer.Rules.CSharp
         internal const string DiagnosticId = "S4581";
         private const string MessageFormat = "Use 'Guid.NewGuid()' or 'Guid.Empty' or add arguments to this Guid instantiation.";
 
-        private static readonly DiagnosticDescriptor rule =
-            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
-        protected override void Initialize(SonarAnalysisContext context)
-        {
-            context.RegisterSyntaxNodeActionInNonGenerated(
-                c =>
+        protected override void Initialize(SonarAnalysisContext context) =>
+            context.RegisterSyntaxNodeActionInNonGenerated(c =>
+            {
+                var objectCreationSyntax = ObjectCreationFactory.Create(c.Node);
+                if (objectCreationSyntax != null)
                 {
-                    var objectCreationSyntax = (ObjectCreationExpressionSyntax)c.Node;
-
-                    if (objectCreationSyntax.ArgumentList?.Arguments.Count == 0 &&
-                        c.SemanticModel.GetSymbolInfo(objectCreationSyntax).Symbol is IMethodSymbol methodSymbol &&
-                        methodSymbol.ContainingType.Is(KnownType.System_Guid))
+                    if (objectCreationSyntax.ArgumentList?.Arguments.Count == 0
+                        && objectCreationSyntax.MethodSymbol(c.SemanticModel).ContainingType.Is(KnownType.System_Guid))
                     {
-                        c.ReportIssue(Diagnostic.Create(rule, objectCreationSyntax.GetLocation()));
+                        c.ReportIssue(Diagnostic.Create(Rule, objectCreationSyntax.Expression.GetLocation()));
                     }
-                },
-                SyntaxKind.ObjectCreationExpression);
-        }
+                }
+            },
+            SyntaxKind.ObjectCreationExpression,
+            SyntaxKindEx.ImplicitObjectCreationExpression);
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/NewGuidShouldNotBeUsed.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/NewGuidShouldNotBeUsed.cs
@@ -40,7 +40,8 @@ namespace SonarAnalyzer.Rules.CSharp
         private static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
-        protected override void Initialize(SonarAnalysisContext context) =>
+        protected override void Initialize(SonarAnalysisContext context)
+        {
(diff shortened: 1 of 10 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/NewGuidShouldNotBeUsed.cs
```diff
@@ -33,27 +35,56 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class NewGuidShouldNotBeUsed : SonarDiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S4581";
-        private const string MessageFormat = "Use 'Guid.NewGuid()' or 'Guid.Empty' or add arguments to this Guid instantiation.";
+        private const string MessageFormat = "Use 'Guid.NewGuid()' or 'Guid.Empty' or use Guid instantiation with arguments.";
```
csaba-sagi-sonarsource, line 38: I am not the reviewer so, I just popped up out of nowhere 😃. The message in the RSPEC repo should be also updated for the rule if we changed the message format here.

mary-georgiou-sonarsource: I am planning to mention this today.  😄 
I want to propose changing the RPSEC description a bit and adding more examples.

costin-zaharia-sonarsource: Sounds good! If there is nothing unclear, I think you can add a PR directly and put it in review.

costin-zaharia-sonarsource: Regarding the message, what do you think about the following: `Use 'Guid.NewGuid()', 'Guid.Empty' or the constructor with arguments.`

mary-georgiou-sonarsource: Voila: https://github.com/SonarSource/rspec/pull/590

analyzers/tests/SonarAnalyzer.UnitTest/Rules/NewGuidShouldNotBeUsedTest.cs
```diff
@@ -30,5 +30,9 @@ public class NewGuidShouldNotBeUsedTest
         [TestMethod]
         public void NewGuidShouldNotBeUsed() =>
             Verifier.VerifyAnalyzer(@"TestCases\NewGuidShouldNotBeUsed.cs", new NewGuidShouldNotBeUsed());
+
+        [TestMethod]
+        public void NewGuidShouldNotBeUsed_CSharp9() =>
+            Verifier.VerifyAnalyzerFromCSharp9Library(@"TestCases\NewGuidShouldNotBeUsed.CSharp9.cs", new NewGuidShouldNotBeUsed());
     }
```
costin-zaharia-sonarsource, line 37: We prefer to run the C# 9 test cases only on Net context (by using conditional compilation `#if NET`)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NewGuidShouldNotBeUsed.cs
```diff
@@ -1,17 +1,33 @@
...
+using System.Text;
 
 namespace Tests.Diagnostics
 {
     public class Program
     {
         public void Foo()
         {
-            var result = new Guid(); // Noncompliant {{Use 'Guid.NewGuid()' or 'Guid.Empty' or add arguments to this Guid instantiation.}}
+            var result = new Guid(); // Noncompliant {{Use 'Guid.NewGuid()' or 'Guid.Empty' or use Guid instantiation with arguments.}}
```
costin-zaharia-sonarsource, line 11: The message needs to be updated also in tests

costin-zaharia-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/NewGuidShouldNotBeUsed.CSharp9.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public void Foo()
+        {
+            Guid result = new(); // Noncompliant
+
+            result = default; // Noncompliant
+//                   ^^^^^^^
+
+            StringBuilder st = new(); // Checking that the rules raises issue only for the Guid class.
+            string test = default;
```
costin-zaharia-sonarsource, line 16: We can add one more test with a method:
`public Guid Get() => default; // Noncompliant`

analyzers/src/SonarAnalyzer.CSharp/Rules/NewGuidShouldNotBeUsed.cs
```diff
@@ -33,27 +35,56 @@ namespace SonarAnalyzer.Rules.CSharp
...
-                        c.ReportIssue(Diagnostic.Create(rule, objectCreationSyntax.GetLocation()));
-                    }
-                },
-                SyntaxKind.ObjectCreationExpression);
+        private static bool DefaultExpressionIdentifierIsGuid(DefaultExpressionSyntax defaultExpression)
+        {
+            var expressionIdentifier = defaultExpression.Type.GetIdentifier();
+            return expressionIdentifier != null
+                ? expressionIdentifier.Identifier.Text.Equals("Guid")
+                : defaultExpression.Type.ToString().Equals("System.Guid");
```
costin-zaharia-sonarsource, line 87: I think we can simplify this a bit:
```suggestion
            var typeName = defaultExpression.Type.ToString();
            return typeName == "Guid" || typeName == "System.Guid";
```
It will not work with aliases but I think that's good enough for now.

mary-georgiou-sonarsource: Yes and it's a false negative, so it won't create too much noise.

analyzers/src/SonarAnalyzer.CSharp/Rules/NewGuidShouldNotBeUsed.cs
```diff
@@ -25,35 +25,63 @@
...
+            DetectIssueInConstructors(context);
+            DetectIssueInDefaultExpressions(context);
+        }
+
+        private static void DetectIssueInConstructors(SonarAnalysisContext context) =>
+            context.RegisterSyntaxNodeActionInNonGenerated(c =>
+            {
+                var objectCreationSyntax = ObjectCreationFactory.Create(c.Node);
+                if (objectCreationSyntax.ArgumentList?.Arguments.Count == 0
+                    && objectCreationSyntax.MethodSymbol(c.SemanticModel) is IMethodSymbol methodSymbol
```
costin-zaharia-sonarsource, line 54: Currently `is IMethodSymbol` is only doing a null check since `MethodSymbol` returns already a `IMethodSymbol`. This is equivalent to `is { }`.
```suggestion
                    && objectCreationSyntax.MethodSymbol(c.SemanticModel) is { } methodSymbol
```

analyzers/src/SonarAnalyzer.CSharp/Rules/NewGuidShouldNotBeUsed.cs
```diff
@@ -25,35 +25,63 @@
...
-                    {
-                        c.ReportIssue(Diagnostic.Create(rule, objectCreationSyntax.GetLocation()));
-                    }
-                },
-                SyntaxKind.ObjectCreationExpression);
+        private static bool DefaultExpressionIdentifierIsGuid(DefaultExpressionSyntax defaultExpression)
+        {
+            var typeName = defaultExpression.Type.ToString();
+            return typeName == "Guid" || typeName == "System.Guid";
         }
```
costin-zaharia-sonarsource, line 85: I've realized now that it can be reduced to:
```
        private static bool DefaultExpressionIdentifierIsGuid(DefaultExpressionSyntax defaultExpression) =>
            defaultExpression.Type.ToString() is var typeName
            && (typeName == "Guid" || typeName == "System.Guid");
```

Up to you if you want to change it.

costin-zaharia-sonarsource (approved):
Nice! Looks good to me. I've added though 2 small comments, it's up to you if you want to address them.

mary-georgiou-sonarsource:
Fixes #5037

Issue #5034, opened 2021-11-10
End step is failing if a rule raises an issue without a message


costin-zaharia-sonarsource:
### Description

End step is failing if a rule raises an issue without a message

### Repro steps

Try to import a Sarif report with an issue without a message.

### Known workarounds

Please provide a description of any known workarounds.

### Related information

* C#/VB.NET Plugins version 8.30.0.37606

PR #5042, opened 2021-11-11
Do not push issues without a message

costin-zaharia-sonarsource:
- SQ/SC will reject them

Related to #5034


csaba-sagi-sonarsource (approved):
LGTM!

PR #5045, opened 2021-11-12
Introduce a rule that detects unicode vulnerabilities

Corniel:
See: https://community.sonarsource.com/t/warn-on-suspicious-unicode-characters/52992
I gave it a try, and think it works just as it should. Obviously, some (a lot) of administration has to be done.


pavel-mikula-sonarsource:
Hi @Corniel, thank you for bringing this topic up. 

There's a new rule [S6389](https://sonarsource.github.io/rspec/#/rspec/S6389) for this that is implemented as a language-independent rule for SonarCloud/SonarQube. So we won't have it in our repository.

Corniel:
From a commercial point of view, I understand this decision.

PR #5047, opened 2021-11-12
Get rid of Newtonsoft.JSON dependency

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/DatabasePasswordsShouldBeSecure.cs
```diff
@@ -130,21 +131,25 @@ private void ReportEmptyPassword(XDocument doc, string webConfigPath, Compilatio
...
                 }
             }
         }
 
-        private static Location CreateLocation(IJsonLineInfo lineInfo, int length, string path)
+        private static Location CreateLocation(JsonNode node, string path)
         {
-            var start = new LinePosition(lineInfo.LineNumber - 1, lineInfo.LinePosition - 1 - length);
-            var end = new LinePosition(lineInfo.LineNumber - 1, lineInfo.LinePosition - 1);
+            var length = node.Value.ToString().Length;
```
Corniel, line 150: Is this correct for things like:

``` JSON
{
   value: 123.000,
}
```
Or is the value still a JsonNode here? (and not a floating-point or decimal)

dmitry-golovinov-sonarsource: This method is used on the strings that passed the check inside _IsVulnerable_, so they are already supposed to be strings

analyzers/src/SonarAnalyzer.CSharp/Rules/DatabasePasswordsShouldBeSecure.cs
```diff
@@ -28,10 +29,10 @@
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Text;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.Json;
+using SonarAnalyzer.Json.Parsing;
```
pavel-mikula-sonarsource, line 35: There was a method to be called to avoid `.Parsing` namespace to be visible from usage side.. `JsonNode.FromString(...)`

analyzers/src/SonarAnalyzer.CSharp/Rules/DatabasePasswordsShouldBeSecure.cs
```diff
@@ -103,13 +104,13 @@ private void CheckAppSettingJson(CompilationAnalysisContext c, string fullPath)
             var appSettings = File.ReadAllText(fullPath);
             if (appSettings.Contains("\"ConnectionStrings\""))
             {
-                using var jsonReader = new JsonTextReader(new StringReader(appSettings));
                 try
                 {
-                    var json = JObject.Load(jsonReader, new JsonLoadSettings {LineInfoHandling = LineInfoHandling.Load, CommentHandling = CommentHandling.Ignore});
+                    var analyzer = new SyntaxAnalyzer(appSettings);
+                    var json = analyzer.Parse();
```
pavel-mikula-sonarsource, line 110: We can remove these two lines if we embedthe `JsonNode.FromString(appSettings)` in the `ReportEmptyPassword`.

pavel-mikula-sonarsource (approved):
LGTM. There one comment (split into two places) to simplify before merging

PR #5049, opened 2021-11-12
Rm `Newtonsoft.Json` dependency in favour of own parser

dmitry-golovinov-sonarsource:


andrei-epure-sonarsource:
@dmitry-golovinov-sonarsource it would be worth to look into the missing ccov , I had the impression that this got solved in one of the PRs

image

dmitry-golovinov-sonarsource:
@andrei-epure-sonarsource we've already discussed this switch with Pavel, seems like it's more a problem with the coverage computation, since if you will make it more explicit, you will see that all switch cases are covered
image
There is only 1 case missed (with 9), that was somehow lost in the rebases. I will open a PR with it..

pavel-mikula-sonarsource:
The coverage numbers comes from the way how IL is generated for `switch` statement and how some optimizations are there related to character grouping. 

We can cover this, but we need to find what characters to add to the test to please the tool. To do this, we'd need to read the IL and do a bit of MSIL math there.

PR #5051, opened 2021-11-15
Add missing test for the JSON parser

dmitry-golovinov-sonarsource:
Adds missing test from https://github.com/SonarSource/sonar-dotnet/pull/5049

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/Json/LexicalAnalyzerTest.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/Json/LexicalAnalyzerTest.cs
@@ -90,6 +90,7 @@ namespace SonarAnalyzer.UnitTest.Common
         [DataRow("6")]
         [DataRow("7")]
         [DataRow("8")]
+        [DataRow("9")]
         [DataRow("42")]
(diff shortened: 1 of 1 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/Json/LexicalAnalyzerTest.cs
```diff
@@ -90,6 +90,7 @@ public void ReadSpecialCharacters()
         [DataRow("6")]
         [DataRow("7")]
         [DataRow("8")]
+        [DataRow("9")]
```
pavel-mikula-sonarsource, line 93: I did the MSIL math for `ReadSymbol()` method. Please add UTs that will start (and throw) with these characters:
```
#
$
%
&
'
(
)
*
+
.
/
```
That should be enough to get 96.26%

dmitry-golovinov-sonarsource: Thanks for the math! I've pushed the corresponding commit

pavel-mikula-sonarsource (approved):
LGTM

PR #5052, opened 2021-11-15
AZP opencover: add excludebyattribute

andrei-epure-sonarsource:
(tested in #5053)


pavel-mikula-sonarsource (approved):
LGTM. I'm not merging it because of the 🥶 master

PR #5054, opened 2021-11-15
Update RSPEC

csaba-sagi-sonarsource:


csaba-sagi-sonarsource:
The rule is indeed deprecated, but it does not have a replacement rule. From what I understood it should be in ready state until the replacement is implemented for it.

costin-zaharia-sonarsource (approved):
LGTM!

PR #5055, opened 2021-11-15
Remove comments related to Json feature branch

pavel-mikula-sonarsource:
`FIXME` comments should not be on `master`

These should have been removed in #5047 before the feature branch was merged to master as #5047 ensures the correct issue reporting.

This is a draft to prevent merging before release.


andrei-epure-sonarsource (approved):
lgtm

PR #5066, opened 2021-11-17
Rule S4462: FP when in top level statement.

mary-georgiou-sonarsource:


analyzers/src/SonarAnalyzer.Common/Helpers/KnownMethods.cs
```diff
@@ -35,7 +35,8 @@ public static bool IsMainMethod(this IMethodSymbol methodSymbol)
             // Adding support for new async main: https://blogs.msdn.microsoft.com/mazhou/2017/05/30/c-7-series-part-2-async-main
             return methodSymbol != null
                 && methodSymbol.IsStatic
-                && methodSymbol.Name.Equals("Main", StringComparison.OrdinalIgnoreCase) // VB.NET is case insensitive
+                && (methodSymbol.Name.Equals("Main", StringComparison.OrdinalIgnoreCase) // VB.NET is case insensitive
+                    || methodSymbol.Name.Equals("<Main>$", StringComparison.OrdinalIgnoreCase))
```
costin-zaharia-sonarsource, line 39: Shouldn't we use here `StringComparison.Ordinal`? I think that top level method name is always `<Main>$` for C#.

analyzers/src/SonarAnalyzer.CSharp/Rules/CallToAsyncMethodShouldNotBeBlocking.cs
```diff
@@ -185,12 +185,17 @@ expression switch
...
 
         private static bool IsNamedSymbolOfExpectedType(MemberAccessExpressionSyntax memberAccess, SemanticModel semanticModel, Dictionary<string, KnownType> expectedTypes) =>
             expectedTypes.Keys.Any(memberAccess.NameIs)
             && semanticModel.GetSymbolInfo(memberAccess).Symbol?.ContainingType?.ConstructedFrom is { } memberAccessSymbol
             && memberAccessSymbol.Is(expectedTypes[memberAccess.Name.Identifier.ValueText]);
+
+        private static bool IsInTopLevelStatement(SyntaxNodeAnalysisContext context, MemberAccessExpressionSyntax memberAccess) =>
+            context.ContainingSymbol is IMethodSymbol containingMethodSymbol
+            && containingMethodSymbol.IsMainMethod()
+            && !memberAccess.Ancestors().Any(x => x.IsKind(SyntaxKindEx.LocalFunctionStatement));
```
costin-zaharia-sonarsource, line 199: According to [docs](https://rules.sonarsource.com/csharp/RSPEC-4462): 
> Main methods of Console Applications are not subject to this deadlock issue and so are ignored by this rule.

I think we should skip also the local functions in this case since they can be used only from the `Main` method

costin-zaharia-sonarsource (changes requested):

mary-georgiou-sonarsource:
The `SnippetCompiler` used in the `NavigationHelperTest` does not seem to work for C#9 + syntax.
I think I'll open an issue for this since it might be a bit out of focus for now.

costin-zaharia-sonarsource (approved):
LGTM!

PR #5069, opened 2021-11-18
Rule S2760: Detects when same condition is used in top level statements

mary-georgiou-sonarsource:


costin-zaharia-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpNavigationHelper.cs
```diff
@@ -77,7 +77,26 @@ public static IEnumerable<SwitchSectionSyntax> GetPrecedingSections(this SwitchS
...
-        public static StatementSyntax GetPrecedingStatement(this StatementSyntax currentStatement) =>
-            currentStatement.Parent.ChildNodes().OfType<StatementSyntax>().TakeWhile(x => x != currentStatement).LastOrDefault();
+        public static StatementSyntax GetPrecedingStatement(this StatementSyntax currentStatement)
+        {
+            var previousStatement = currentStatement.Parent
+                                                    .ChildNodes()
+                                                    .OfType<StatementSyntax>()
+                                                    .TakeWhile(x => x != currentStatement)
+                                                    .LastOrDefault();
+            if (previousStatement == null) // this means that we might be in a top-level-statement
```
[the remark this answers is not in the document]
mary-georgiou-sonarsource: Yes, that's right. The function detects the previous statement that is at the same level.
I think it might be better if I move it into the only rule that it's actually using it and rename it since it's not really generic in its usage.
The use of `StatementSyntax` in the input is only because this function is used for `IfStatement` and `SwitchStament` . WDYT?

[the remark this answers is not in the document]
costin-zaharia-sonarsource: In the in the other rule you changed, you had a different way to check if the `node` is on the top level or not. Can't we use that also here? I think that having to maintain multiple ways to check this will prove hard and confusing.

analyzers/src/SonarAnalyzer.Shared/Extensions/StatementSyntaxExtensions.cs
```diff
@@ -43,5 +45,40 @@ public static IEnumerable<StatementSyntax> GetPreviousStatements(this StatementS
                 ? previousStatements.Union(GetPreviousStatements(parentStatement))
                 : previousStatements;
         }
+
+#if CS
```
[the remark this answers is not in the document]
mary-georgiou-sonarsource: It is already like so, but the `GetPrecedingStatement` does not compile for VisualBasic (gives me an error for the `SyntaxEx` namespace).

costin-zaharia-sonarsource (changes requested):
I'm rejecting the PR since the build is failing.

mary-georgiou-sonarsource:
> 

I rebased and now it seems to be fine locally.

costin-zaharia-sonarsource:
> I rebased and now it seems to be fine locally.

Same problem:
2021-11-22T16:00:47.1122163Z   Failed GetPrecedingIfsInConditionChain [25 ms]
2021-11-22T16:00:47.1123271Z   Error Message:
2021-11-22T16:00:47.1123754Z    Expected preceding to contain a single item, but the collection is empty.
2021-11-22T16:00:47.1124174Z   Stack Trace:
2021-11-22T16:00:47.1124833Z      at FluentAssertions.Execution.LateBoundTestFramework.Throw(String message)
2021-11-22T16:00:47.1125440Z    at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message)
2021-11-22T16:00:47.1126138Z    at FluentAssertions.Execution.DefaultAssertionStrategy.HandleFailure(String message)
2021-11-22T16:00:47.1126924Z    at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
2021-11-22T16:00:47.1127553Z    at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
2021-11-22T16:00:47.1128197Z    at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args)
2021-11-22T16:00:47.1129009Z    at FluentAssertions.Collections.SelfReferencingCollectionAssertions`2.ContainSingle(String because, Object[] becauseArgs)
2021-11-22T16:00:47.1130015Z    at SonarAnalyzer.UnitTest.Helpers.NavigationHelperTest.GetPrecedingIfsInConditionChain() in C:\ProgramData\vsts-agent\1\s\analyzers\tests\SonarAnalyzer.UnitTest\Helpers\NavigationHelperTest.cs:line 97

See build logs:
https://dev.azure.com/sonarsource/399fb241-ecc7-4802-8697-dcdd01fbb832/_apis/build/builds/39896/logs/160

analyzers/src/SonarAnalyzer.Shared/Extensions/StatementSyntaxExtensions.cs
```diff
@@ -43,5 +47,34 @@ public static IEnumerable<StatementSyntax> GetPreviousStatements(this StatementS
...
+                                                                                    .FirstOrDefault())
+                                                                      .Where(x => x != null);
+                return currentStatement.GetPrecedingStatement(childrenOfParentNode);
+            }
+            else
+            {
+                var childrenOfParentNode = currentStatement.Parent.ChildNodes();
+                return currentStatement.GetPrecedingStatement(childrenOfParentNode);
+            }
+        }
```
costin-zaharia-sonarsource, line 72: As far as I can tell, it's enough to check if the parent is a `GlobalStatementSyntax` node or not. In case it is, we can check for the siblings by skipping the `GlobalStatementSyntax` nodes and using their statement.

I've tried to simplify the code a bit using the following snippet and the tests are green: 
```suggestion
        public static StatementSyntax GetPrecedingStatement(this StatementSyntax statement)
        {
            var siblings = statement.Parent is GlobalStatementSyntax
                               ? statement.SyntaxTree
                                                 .GetCompilationUnitRoot()
                                                 .ChildNodes()
                                                 .OfType<GlobalStatementSyntax>()
                                                 .Select(x => x.Statement)
                               : statement.Parent.ChildNodes();

            return statement.GetPrecedingStatement(siblings);
        }
```

costin-zaharia-sonarsource (changes requested):

mary-georgiou-sonarsource:
I deleted the `IsTopLevelMain` for now because it's not used and it drops the coverage.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/NavigationHelperTest.cs
```diff
@@ -36,14 +38,17 @@ namespace Test
     class TestClass
     {
         public void DoSomething(){}
-        public void IfMethod()
+        public void IfMethod(int j)
```
costin-zaharia-sonarsource, line 41: If the change is not needed anymore I think we should revert it.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/NavigationHelperTest.cs
```diff
@@ -65,16 +70,39 @@ public void SwitchMethod()
...
         private MethodDeclarationSyntax ifMethod;
         private MethodDeclarationSyntax switchMethod;
 
+        private SyntaxTree syntaxTreeTopLevelStatement;
+
         [TestInitialize]
         public void TestSetup()
         {
             var syntaxTree = CSharpSyntaxTree.ParseText(Source);
+            syntaxTreeTopLevelStatement = CSharpSyntaxTree.ParseText(SourceTopLevelStatement);
```
costin-zaharia-sonarsource, line 95: TestInitialize is run before each test and this field is used in only one of them. We should move this to `GetPrecedingStatementTopLevelStatements`.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/NavigationHelperTest.cs
```diff
@@ -145,5 +173,18 @@ public void GetPrecedingStatement()
...
+        {
+            var globalStatements = syntaxTreeTopLevelStatement.GetRoot()
+                                                              .ChildNodes()
+                                                              .Select(x => x.ChildNodes().FirstOrDefault())
+                                                              .OfType<StatementSyntax>()
+                                                              .ToArray();
+
+            globalStatements[1].GetPrecedingStatement().Should().BeEquivalentTo(globalStatements[0]);
+            globalStatements[0].GetPrecedingStatement().Should().Be(null);
+        }
```
costin-zaharia-sonarsource, line 188: I would rewrite the test to avoid running a similar alg in the test with the one in production code. We can look for the statements and check them directly. 

WDYT about the following snippet?
```suggestion
        [TestMethod]
        public void GetPrecedingStatementTopLevelStatements()
        {
            var variableDeclarators = syntaxTreeTopLevelStatement.GetRoot().DescendantNodes().OfType<LocalDeclarationStatementSyntax>().ToArray();
            var aDeclaration = variableDeclarators[0];
            var bDeclaration = variableDeclarators[0];

            aDeclaration.GetPrecedingStatement().Should().Be(null);
            bDeclaration.GetPrecedingStatement().Should().BeEquivalentTo(aDeclaration);
        }
```

costin-zaharia-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CFG/ShimLayer/StatementSyntaxExtensions.cs
```diff
@@ -1,12 +1,12 @@
 ﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
```
costin-zaharia-sonarsource, line 7: The code from ShimLayer is imported from Stylecop and we don't want to change it in order to be able to easily update to the latest change. Could you please revert this change?

Probably I forgot to mention this, sorry for that.

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/StatementSyntaxExtensionsTest.cs
```diff
@@ -0,0 +1,89 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarAnalyzer.Extensions;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.UnitTest.Extensions
+{
+    [TestClass]
+    public class StatementSyntaxExtensionsTest
+    {
+        private const string Source = @"
```
costin-zaharia-sonarsource, line 34: What do you think about moving the snippets in the tests? It looks that each of these is used in only one place and we could avoid navigating the code to understand the tests.

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/StatementSyntaxExtensionsTest.cs
```diff
@@ -0,0 +1,89 @@
...
+if (a == b)
+{
+    DoSomething();
+}
+void DoSomething() { }";
+
+        [TestMethod]
+        public void GetPrecedingStatement()
+        {
+            var syntaxTree = CSharpSyntaxTree.ParseText(Source);
```
costin-zaharia-sonarsource, line 63: `syntaxTree` is used only in a single place and I think it can be inlined there.

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/StatementSyntaxExtensionsTest.cs
```diff
@@ -0,0 +1,89 @@
...
+            var ifMethod = syntaxTree.GetRoot()
+                                     .DescendantNodes()
+                                     .OfType<MethodDeclarationSyntax>()
+                                     .First(m => m.Identifier.ValueText == "IfMethod");
+
+            var statements = ifMethod.Body.Statements.ToList();
```
costin-zaharia-sonarsource, line 69: Since `ifMethod` is used in one single place I wouldn't define it and use `statements` instead, by chaining `.Body.Statements.ToList();` after the `.First` call.

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/StatementSyntaxExtensionsTest.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            statements[1].GetPrecedingStatement().Should().BeEquivalentTo(statements[0]);
```
costin-zaharia-sonarsource, line 71: In my opinion, giving names to `statements[0]` and `statements[1]` would make the code easier to understand.

costin-zaharia-sonarsource (changes requested):
Just a few minor comments regarding the unit tests.

analyzers/tests/SonarAnalyzer.UnitTest/Extensions/StatementSyntaxExtensionsTest.cs
```diff
@@ -0,0 +1,89 @@
...
+        public void IfMethod()
+        {
+            if (a>1)
+                DoSomething();
+            if (a<0)
+                DoSomething();
+        }
+    }
+}";
+            var syntaxTree = CSharpSyntaxTree.ParseText(source);
```
costin-zaharia-sonarsource, line 53: I think you forgot to remove this one.

costin-zaharia-sonarsource (approved):
LGTM!

PR #5070, opened 2021-11-18
Update Vsix NuGet references

pavel-mikula-sonarsource:
The old version of `Microsoft.VSSDK.BuildTools` could not be used in VS 2022 to compile the project.

>The target "CreateClientEnabledPkgFiles" does not exist

### DIFF
--- a/analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.Vsix.csproj
+++ b/analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.Vsix.csproj
@@ -58,7 +58,6 @@
     <None Include="app.config" />
   </ItemGroup>
   <ItemGroup>
-    <None Include="packages.lock.json" />
     <None Include="SonarAnalyzer.CSharp.nuspec">
       <SubType>Designer</SubType>
     </None>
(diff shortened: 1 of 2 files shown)


pavel-mikula-sonarsource:
@costin-zaharia-sonarsource Please check that this stunt works in Rider as well :)

costin-zaharia-sonarsource:
We really need to get rid of this project. I'll try to find some time next week to do this.

analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.Vsix.csproj
```diff
@@ -58,6 +58,7 @@
     <None Include="app.config" />
   </ItemGroup>
   <ItemGroup>
+    <None Include="packages.lock.json" />
```
costin-zaharia-sonarsource, line 61: I think this should not be part of the csproj.

costin-zaharia-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Vsix/SonarAnalyzer.Vsix.csproj
pavel-mikula-sonarsource: I'd also rather delete the whole file 😆 It was autogenerated by VS, because it's old framework-based noisy Csproj.

costin-zaharia-sonarsource (approved):
LGTM!

PR #5073, opened 2021-11-18
Update vulnerable NuGet packages

costin-zaharia-sonarsource:
Highlighted by WhiteSource:
- system.text.regularexpressions.4.1.0.nupkg
- system.net.http.4.3.0.nupkg
- system.text.regularexpressions.4.3.0.nupkg

The others were transient dependencies that failed the build since were resolved with different versions (we build for both net46 and netcore3.1).


analyzers/src/SonarAnalyzer.CFG/SonarAnalyzer.CFG.csproj
```diff
@@ -13,10 +13,14 @@
...
       <NoWarn>NU1701</NoWarn>
     </PackageReference>
+    <PackageReference Include="System.Collections" Version="4.3.0" />
     <PackageReference Include="System.Collections.Immutable" Version="1.1.37">
       <!-- Downgrade System.Collections.Immutable to support VS2015 Update 3 -->
       <NoWarn>NU1605, NU1701</NoWarn>
     </PackageReference>
+    <PackageReference Include="System.IO.FileSystem" Version="4.3.0" />
+    <PackageReference Include="System.IO.FileSystem.Primitives" Version="4.3.0" />
+    <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
```
pavel-mikula-sonarsource, line 23: Can we add a comment that this is an explicit version bump for transitive dependency? To explain why it's here and why we shouldn't remove it in the future?

analyzers/src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
```diff
@@ -20,6 +20,7 @@
       <!-- Downgrade System.Collections.Immutable to support VS2015 Update 3 -->
       <NoWarn>NU1605, NU1701</NoWarn>
     </PackageReference>
+    <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
```
pavel-mikula-sonarsource, line 23: Comment? (Depends on the question in CFG)

analyzers/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
```diff
@@ -17,10 +17,14 @@
       <NoWarn>NU1701</NoWarn>
     </PackageReference>
     <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="1.3.2" />
+    <PackageReference Include="System.Collections" Version="4.3.0" />
```
pavel-mikula-sonarsource, line 20: Same questions as for CFG

analyzers/src/SonarAnalyzer.VisualBasic/SonarAnalyzer.VisualBasic.csproj
```diff
@@ -14,6 +14,7 @@
       <!-- Downgrade System.Collections.Immutable to support VS2015 Update 3 -->
       <NoWarn>NU1605, NU1701</NoWarn>
     </PackageReference>
+    <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
```
pavel-mikula-sonarsource, line 17: Comment? (Depends on the CFG question)

analyzers/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
```diff
@@ -45,6 +45,8 @@
       <NoWarn>NU1701</NoWarn>
     </PackageReference>
     <PackageReference Include="NuGet.Protocol" Version="5.6.0" />
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
+    <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
```
pavel-mikula-sonarsource, line 49: Comment? (Depends on the CFG question)

analyzers/src/SonarAnalyzer.CFG/SonarAnalyzer.CFG.csproj
```diff
@@ -13,10 +13,14 @@
       <!-- This package is a dependency of Microsoft.CodeAnalysis.CSharp.Workspaces. It is safe to use since it's compatible with .Net Portable runtime -->
       <NoWarn>NU1701</NoWarn>
     </PackageReference>
+    <PackageReference Include="System.Collections" Version="4.3.0" />
     <PackageReference Include="System.Collections.Immutable" Version="1.1.37">
       <!-- Downgrade System.Collections.Immutable to support VS2015 Update 3 -->
       <NoWarn>NU1605, NU1701</NoWarn>
     </PackageReference>
+    <PackageReference Include="System.IO.FileSystem" Version="4.3.0" />
+    <PackageReference Include="System.IO.FileSystem.Primitives" Version="4.3.0" />
```
pavel-mikula-sonarsource, line 22: How do these break the build? On CI only? I can still build it locally if I comment them.

analyzers/src/SonarAnalyzer.CFG/SonarAnalyzer.CFG.csproj
```diff
@@ -13,10 +13,14 @@
       <!-- This package is a dependency of Microsoft.CodeAnalysis.CSharp.Workspaces. It is safe to use since it's compatible with .Net Portable runtime -->
       <NoWarn>NU1701</NoWarn>
     </PackageReference>
+    <PackageReference Include="System.Collections" Version="4.3.0" />
```
pavel-mikula-sonarsource, line 16: This changes transitive 4.3.0 to explicit 4.3.0 if I compare `dotnet list package --include-transitive`. Does it really have to be here?

pavel-mikula-sonarsource (changes requested):
There are 3 questions in CFG csproj, the rest just repeats itself.

analyzers/src/SonarAnalyzer.CFG/SonarAnalyzer.CFG.csproj
costin-zaharia-sonarsource: It was a misunderstanding, there were a few problems with other libraries since a downgrade was detected at build time, and I thought that this library was also part of the problem:
image

analyzers/src/SonarAnalyzer.CFG/SonarAnalyzer.CFG.csproj
costin-zaharia-sonarsource: I've noticed now that only `System.IO.FileSystem` is needed and only on the `SonarAnalyzer.Common`. 
If missing from there they fail the local build. This is the message I get :
```
SonarAnalyzer.RuleDescriptorGenerator.csproj: [NU1605] Detected package downgrade: System.Text.Encoding from 4.3.0 to 4.0.11. Reference the package directly from the project to select a different version. 
 SonarAnalyzer.RuleDescriptorGenerator -> SonarAnalyzer -> Microsoft.CodeAnalysis.Workspaces.Common 1.3.2 -> Microsoft.CodeAnalysis.Common 1.3.2 -> System.Console 4.0.0 -> runtime.win.System.Console 4.3.1 -> System.Text.Encoding (>= 4.3.0) 
 SonarAnalyzer.RuleDescriptorGenerator -> SonarAnalyzer -> Microsoft.CodeAnalysis.Workspaces.Common 1.3.2 -> Microsoft.CodeAnalysis.Common 1.3.2 -> System.Text.Encoding (>= 4.0.11)
```

analyzers/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
```diff
@@ -21,6 +21,10 @@
       <!-- Downgrade System.Collections.Immutable to support VS2015 Update 3 -->
       <NoWarn>NU1605, NU1701</NoWarn>
     </PackageReference>
+    <!--  Manually upgrade since the version transitively linked (4.0.1) is marked as vulnerable by WhiteSource. -->
+    <PackageReference Include="System.IO.FileSystem" Version="4.3.0" />
```
pavel-mikula-sonarsource, line 25: This `FileSystem` doesn't seem intentional to me.
```suggestion
```

pavel-mikula-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
pavel-mikula-sonarsource: Can we move this `System.IO.FileSystem` thing to Rule descriptor, as it fails only there? I gave it a try and the build worked.

We'd avoid messing with production project to please a build tooling.
In the long run, I want to get rid of the rule descriptor one day so the workaround would vanish with it.

pavel-mikula-sonarsource (approved):
LGTM

PR #5075, opened 2021-11-19
Small consistency and readability changes

costin-zaharia-sonarsource:


costin-zaharia-sonarsource:
We discussed offline executing the WhiteSource scan in a parallel job:
- we're not able to do this without restoring the dependencies and rebuilding the project (the scan is more efficient if we do this)
- the time will decrease significantly once we update the VM image to contain the WhiteSource unified agent and remove the download during the build (the PR is now in review)

pavel-mikula-sonarsource (approved):
LGTM

PR #5085, opened 2021-11-23
Rule S1186: check also the local functions

costin-zaharia-sonarsource:


mary-georgiou-sonarsource (approved):
LGTM

costin-zaharia-sonarsource:
> I'd add one test for this to increase coverage.

Do you see a specific case that is not covered?

PR #5083, opened 2021-11-23
Cleanup SE namespaces

pavel-mikula-sonarsource:

PR #5086, opened 2021-11-23
Move SE TestCases to separate directory

pavel-mikula-sonarsource:
This is a draft, rebase will be needed once #5083 is merged. First `Squash Cleanup NSs` commit doesn't need a review.

I'm not going to fix the QG issue. There will be a lot more moving files soon.


andrei-epure-sonarsource (approved):
LGTM

PR #5087, opened 2021-11-23
Move all SE files to Sonar directory

pavel-mikula-sonarsource:
`Squash` named commits don't need a review. 
Namespace renaming will follow in a separate PR to preserve clean change history


andrei-epure-sonarsource (approved):
LGTM

PR #5088, opened 2021-11-23
Rename CSharpExplodedGraph to SonarExplodedGraph

pavel-mikula-sonarsource:
Only `Rename CSharpExplodedGraph to SonarExplodedGraph` commit needs a review.


andrei-epure-sonarsource (approved):
LGTM

PR #5089, opened 2021-11-23
Simplify NETStandard21 reference in UTs

pavel-mikula-sonarsource:
Only these commits need review:
* Prepare
* Remove pain
* Improve


analyzers/tests/SonarAnalyzer.UnitTest/MetadataReferences/NuGetMetadataReference.cs
```diff
@@ -30,9 +30,11 @@ internal static class NuGetMetadataReference
...
         public static References MSTestTestFrameworkV1 => Create("MSTest.TestFramework", "1.1.11");
-        public static References XunitFrameworkV1 =>
-            Create("xunit", "1.9.1")
-            .Concat(Create("xunit.extensions", "1.9.1"));
+        public static References XunitFrameworkV1 => Create("xunit", "1.9.1").Concat(Create("xunit.extensions", "1.9.1"));
+
+        /// <summary>
+        /// Use MetadataReferenceFacade.NETStandard21 instead to avoid #if directives.
+        /// </summary>
         public static References NETStandard21 => Create("NETStandard.Library.Ref", "2.1.0", "netstandard2.1");
```
andrei-epure-sonarsource, line 38: IMO this should be made private and moved to `MetadataReferenceFacade`

we shouldn't rely on comments to avoid usage of public APIs

andrei-epure-sonarsource (changes requested):
Very nice refactoring!

Minor change requested before merging.

andrei-epure-sonarsource:
very weird, the Github Workflow didn't trigger for this PR after requesting the review @pavel-mikula-sonarsource ... 

I checked on https://github.com/SonarSource/sonar-dotnet/actions/workflows/RequestReview.yml and it didn't even start. 

Github Status is green at the moment https://www.githubstatus.com/

andrei-epure-sonarsource (approved):
LGTM!

PR #5096, opened 2021-11-24
 Rule S3776: Introduce support for top level statements 

dmitry-golovinov-sonarsource:


dmitry-golovinov-sonarsource:
PR is supposed to fail until rebased on master that has fixed sarif reports logic

dmitry-golovinov-sonarsource:
@costin-zaharia-sonarsource why have you unassigned it? I've asked for a review, since there are no changes expected to be delivered.

costin-zaharia-sonarsource:
> why have you unassigned it? 

The build is failing and the PR is in draft mode. I was waiting for the problem to be solved.

dmitry-golovinov-sonarsource:
@costin-zaharia-sonarsource AFAIK we are doing draft reviews that shorten the feedback loop. Could you please do it ? 

FYI the problem with build was explained in my first comment (that this PR is expected to fail until it gets rebased)

analyzers/src/SonarAnalyzer.CSharp/Rules/CognitiveComplexity.cs
```diff
@@ -37,6 +38,21 @@ public sealed class CognitiveComplexity : CognitiveComplexityBase<SyntaxKind>
 
         protected override void Initialize(ParameterLoadingAnalysisContext context)
         {
+            context.RegisterSyntaxNodeActionInNonGenerated(c =>
+                {
+                    if (c.ContainingSymbol is IMethodSymbol)
```
costin-zaharia-sonarsource, line 43: Interesting take, I had to solve the same issue and didn't know how to determine if a `compilationUnit` is containing top level statements. Thanks 😄 

Could you please extract this to a method outside of this class so it can be reused?

dmitry-golovinov-sonarsource: :) Thanks! Any ideas where should it belong to?

costin-zaharia-sonarsource: Extension method (`IsTopLevelStatementsFile`) on the `SyntaxNodeAnalysisContext`?

analyzers/src/SonarAnalyzer.CSharp/Rules/CognitiveComplexity.cs
```diff
@@ -37,6 +38,21 @@ public sealed class CognitiveComplexity : CognitiveComplexityBase<SyntaxKind>
...
+                    {
+                        CheckComplexity<CompilationUnitSyntax>(
+                            c,
+                            cu => cu,
```
costin-zaharia-sonarsource, line 47: I had to look for the type to understa what `cu` actually is and understand the code. We try to avoid abreviations. Can please rename this to `compilationUnit`?

analyzers/tests/SonarAnalyzer.UnitTest/Common/MetricsTest.cs
```diff
@@ -495,8 +495,8 @@ public void Complexity()
         [TestMethod]
         public void CognitiveComplexity()
         {
-            var csharpText = System.IO.File.ReadAllText(@"TestCases\CognitiveComplexity.cs");
-            CognitiveComplexity(AnalyzerLanguage.CSharp, csharpText).Should().Be(109);
+            var csharpText = System.IO.File.ReadAllText(@"TestCases\CognitiveComplexity.CSharp9.cs");
```
costin-zaharia-sonarsource, line 498: I think we need to add a separate case for C# 9; otherwise `TestCases\CognitiveComplexity.cs` will never be verified.

costin-zaharia-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/Common/MetricsTest.cs
dmitry-golovinov-sonarsource: TestCases\CognitiveComplexity.cs is tested within CognitiveComplexity tests. But okay, will add a separate line with the check

analyzers/src/SonarAnalyzer.CSharp/Extensions/SyntaxNodeAnalysisContextExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using StyleCop.Analyzers.Lightup;
+
+namespace SonarAnalyzer.Extensions
+{
+    public static class SyntaxNodeAnalysisContextExtensions
+    {
+        public static bool IsTopLevelCompilationUnit(this SyntaxNodeAnalysisContext context) =>
+           context.ContainingSymbol is IMethodSymbol { Name: INamedTypeSymbolExtensions.MainMethodImplicitName };
```
mary-georgiou-sonarsource, line 30: I think it would be better if there was a new declaration of a `private const MainMethodImplicitName` rather than turning into `public` the one from `INamedTypeSymbolExtensions`.
In case we decide to change something in the other class or remove the function from there for example, we'll need to do a change also here.

dmitry-golovinov-sonarsource: There should be as less duplicates as possible. @costin-zaharia-sonarsource agreed to find a better place for this constant in the near future.

analyzers/src/SonarAnalyzer.CSharp/Metrics/CSharpCognitiveComplexityMetric.cs
```diff
@@ -85,6 +94,19 @@ public override void Visit(SyntaxNode node)
...
             }
 
+            public override void VisitCompilationUnit(CompilationUnitSyntax node)
+            {
+                foreach (var globalStatement in node.Members.Where(x => x.IsKind(SyntaxKind.GlobalStatement)))
+                {
+                    base.Visit(globalStatement);
+                }
+
+                if (!onlyGlobalStatements)
```
mary-georgiou-sonarsource, line 104: Instead of adding this extra bool `onlyGlobalStatements`, I would check inside the `VisitCompilationUnit` if the first symbolic member is in a top-level program with this [method](https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/src/SonarAnalyzer.CFG/ShimLayer/INamedTypeSymbolExtensions.cs#L34).
Unless there's some other use case for this bool that I am missing.

mary-georgiou-sonarsource: As discussed with @dmitry-golovinov-sonarsource this bool is needed in order to have compatibility also with other components.

analyzers/src/SonarAnalyzer.CSharp/Extensions/SyntaxNodeAnalysisContextExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using StyleCop.Analyzers.Lightup;
+
+namespace SonarAnalyzer.Extensions
+{
+    public static class SyntaxNodeAnalysisContextExtensions
+    {
+        public static bool IsTopLevelCompilationUnit(this SyntaxNodeAnalysisContext context) =>
```
mary-georgiou-sonarsource, line 29: Seeing this again:
I would add this function to the `ISymbolExtentions` so that it takes as an input an `ISymbol symbol` which checks if `symbol.Name.Equals(MainMethodImplicitName )`.

This is because I've understood that `context.ContainingSymbol` checks the symbol for the registered node.
If we decide to check for example the node's parent, this should not properly work. While with the change I proposed above we would give the parent node's containing symbol and it should be fine.

analyzers/src/SonarAnalyzer.CSharp/Metrics/CSharpCognitiveComplexityMetric.cs
```diff
@@ -31,19 +32,27 @@ namespace SonarAnalyzer.Metrics.CSharp
...
 
         private class CognitiveWalker : CSharpSyntaxWalker
         {
+            private readonly bool onlyGlobalStatements;
+
             public CognitiveComplexityWalkerState<MethodDeclarationSyntax> State { get; }
                 = new CognitiveComplexityWalkerState<MethodDeclarationSyntax>();
 
+            public CognitiveWalker(bool onlyGlobalStatements) =>
+                this.onlyGlobalStatements = onlyGlobalStatements;
```
mary-georgiou-sonarsource, line 54: Coding Style: This could go in one line to increase readability.

mary-georgiou-sonarsource (changes requested):
A couple of more comments.

analyzers/src/SonarAnalyzer.CSharp/Extensions/SyntaxNodeAnalysisContextExtensions.cs
dmitry-golovinov-sonarsource: The idea was to control it on the context level, since you won't need to remember which kind of property to use for the check within a few months from now. :) Unless @costin-zaharia-sonarsource  is ready to change his mind on that, I propose to leave it as it is now.

analyzers/src/SonarAnalyzer.CSharp/Metrics/CSharpCognitiveComplexityMetric.cs
dmitry-golovinov-sonarsource: AFAIK, this is considered to be "complex enough" to be on the new line. 

So we are having one liners only in case of => false or  => 42, which are indeed easy ones

analyzers/src/SonarAnalyzer.CSharp/Extensions/SyntaxNodeAnalysisContextExtensions.cs
costin-zaharia-sonarsource: At a second look at this, I think that Mary's suggestion makes sense: this method does not need the full context to assert that the symbol is the one for top-level functions. Do to this we can move it to `ISymbolEextensions`. With this we could also change the name to something similar with `IsTopLevelMain` since it will only look for that.

costin-zaharia-sonarsource (approved):
LGTM!

mary-georgiou-sonarsource (approved):
LGTM

PR #5097, opened 2021-11-24
Simplify SymbolicExecutionRunner

pavel-mikula-sonarsource:
Only these commits needs a review:
* Remove unneeded indirection
* Simplify SymbolicExecutionRunner


analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
```diff
@@ -100,8 +100,10 @@ protected override void Initialize(SonarAnalysisContext context)
 
         private void Analyze(SyntaxNodeAnalysisContext context, CSharpSyntaxNode body, ISymbol symbol)
         {
+            var enabledAnalyzers = symbolicExecutionAnalyzerFactory.GetEnabledAnalyzers(context);
```
andrei-epure-sonarsource, line 103: this change would benefit of an issue + separate PR, as
- it's quite a nice catch
- it should be tracked in our issues 
- and the PR itself should be separated from the refactoring-only changes

analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
```diff
@@ -112,24 +80,29 @@ protected override void Initialize(SonarAnalysisContext context)
                 c =>
                 {
                     var declaration = (AnonymousFunctionExpressionSyntax)c.Node;
-                    var symbol = c.SemanticModel.GetSymbolInfo(declaration).Symbol;
-                    if (symbol == null)
+                    if (c.SemanticModel.GetSymbolInfo(declaration).Symbol is { } symbol)
```
andrei-epure-sonarsource, line 83: I think it's possible to replace this with 
```
c => Analyze<AnonymousFunctionExpressionSyntax>(c, x => x.Body),
```

This will make all callbacks uniform. I have to admit I spent a bit wondering why this was special.

andrei-epure-sonarsource (approved):
Great refactoring, LGTM!

Please move commit 14843c45b652bc1b37d729ae314780edceb6c2d2 from this branch to a different one and consider my other comment.

andrei-epure-sonarsource:
On this one I think StartProgress (https://github.com/SonarSource/sonar-dotnet/actions/runs/1515704441) was triggered instead of SubmitReview (https://github.com/SonarSource/sonar-dotnet/actions/workflows/SubmitReview.yml)

andrei-epure-sonarsource (changes requested):
requesting changes to make it obvious there's some additional work before merge

analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
pavel-mikula-sonarsource: Done in #5120.

When you've pointed it out, it should actually be a quite impactful change, as we don't have SE rules on a test code. I originally considered it more like a corner case.

analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
pavel-mikula-sonarsource: No, we can't. `AnonymousFunctionExpressionSyntax` needs `GetSymbolInfo` while the rest needs `GetDeclaredSymbol`.

That's the reason why this one is a little different and `Analyze` is split in two overloads.

andrei-epure-sonarsource (approved):
LGTM

PR #5100, opened 2021-11-25
Scaffold RoslynSymbolicExecution

pavel-mikula-sonarsource:
Only commits after `Squash SimplifyRunner` needs a review

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
+++ b/analyzers/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
@@ -9,12 +9,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Compile Remove="NewFolder\**" />
-    <EmbeddedResource Remove="NewFolder\**" />
-    <None Remove="NewFolder\**" />
-  </ItemGroup>
-
-  <ItemGroup>
     <PackageReference Include="Google.Protobuf" Version="3.6.1" />
     <!-- When changing this reference update ProtocExe property too! -->
     <PackageReference Include="Google.Protobuf.Tools" Version="3.6.1" />
--- a/analyzers/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
+++ b/analyzers/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
@@ -9,6 +9,12 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Remove="NewFolder\**" />
+    <EmbeddedResource Remove="NewFolder\**" />
+    <None Remove="NewFolder\**" />
+  </ItemGroup>
+
+  <ItemGroup>
     <PackageReference Include="Google.Protobuf" Version="3.6.1" />
     <!-- When changing this reference update ProtocExe property too! -->
     <PackageReference Include="Google.Protobuf.Tools" Version="3.6.1" />
@@ -31,6 +37,10 @@
       <SonarQubeExclude>true</SonarQubeExclude>
     </Compile>
     <Compile Include="..\Signing.cs" Link="Properties\Signing.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\SonarAnalyzer.CFG\SonarAnalyzer.CFG.csproj" />
   </ItemGroup>
 
   <!-- Generate protobuf classes, supports windows and linux execution -->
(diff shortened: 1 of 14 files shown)


analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
```diff
@@ -32,7 +33,14 @@ public class RoslynSymbolicExecution
 
         public void Execute()
         {
-            // ToDo: Something is missing here
+            foreach (var block in cfg.Blocks)    // ToDo: This is a temporary simplification until we support proper branching
```
andrei-epure-sonarsource: great!

I love simplifications 😄

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
```diff
@@ -27,20 +28,47 @@ namespace SonarAnalyzer.SymbolicExecution.Roslyn
...
         public void Execute()
         {
-            foreach (var block in cfg.Blocks)    // ToDo: This is a temporary simplification until we support proper branching
+            foreach (var block in cfg.Blocks)   // ToDo: This is a temporary simplification until we support proper branching
             {
-                foreach(var operation in block.OperationsAndBranchValue.ToExecutionOrder()) // ToDo: This is a temporary oversimplification to scaffold the engine
+                var state = new ProgramState(); // ToDo: This is a temporary oversimplification to scaffold the enginge
+                foreach (var operation in block.OperationsAndBranchValue.ToExecutionOrder()) // ToDo: This is a temporary oversimplification to scaffold the engine
                 {
+                    state = InvokeChecks(state, (check, ps) => check.PreProcess(ps, operation));
```
andrei-epure-sonarsource, line 50: silly question: shouldn't this return a list of states?

pavel-mikula-sonarsource: _In theory:_ yes.
_In practice:_ no, it's not needed by our rules.
_In the future:_ we can always upgrade if we need to.

analyzers/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
```diff
@@ -8,6 +8,12 @@
     <ProjectGuid>{8a8a663e-1318-4361-bc97-2e63666feadb}</ProjectGuid>
   </PropertyGroup>
 
+  <ItemGroup>
```
andrei-epure-sonarsource, line 11: what is this about?

should it be removed?

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/SymbolicExecutionCheck.cs
```diff
@@ -0,0 +1,32 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using StyleCop.Analyzers.Lightup;
+
+namespace SonarAnalyzer.SymbolicExecution.Roslyn
+{
+    public class SymbolicExecutionCheck
+    {
+        protected SymbolicExecutionCheck() { } // Avoid abstract class
```
andrei-epure-sonarsource, line 27: >Avoid abstract class

what do you mean here?

andrei-epure-sonarsource (approved):
LGTM

(please check `SonarAnalyzer.Common.csproj` before merge)

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
pavel-mikula-sonarsource: 😅

analyzers/src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
pavel-mikula-sonarsource: Autogenerated noise

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/SymbolicExecutionCheck.cs
pavel-mikula-sonarsource: It fixes S1694, I'll add a comment

PR #5103, opened 2021-11-26
Rule S1199: Introduce support for top level statements

dmitry-golovinov-sonarsource:


mary-georgiou-sonarsource (approved):
LGTM
Please check if the master is unfrozen before merging!

PR #5105, opened 2021-11-26
Rule S138: check size of top level method body

costin-zaharia-sonarsource:


dmitry-golovinov-sonarsource (approved):
LGTM! Please, don't forget to rebase before merging!

analyzers/src/SonarAnalyzer.CSharp/Extensions/CompilationUnitSyntaxExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Extensions
+{
+    public static class CompilationUnitSyntaxExtensions
+    {
+        public static IEnumerable<SyntaxNode> GetTopLevelMainBody(this CompilationUnitSyntax compilationUnit) =>
+            compilationUnit.ChildNodes()
```
mary-georgiou-sonarsource, line 33: - I would change this name to `GlobalStatements`. I think it describes better the purpose as my understating is that it gets all the `GlobaStatement` nodes under the `CompilationUnit`.
- I think also `compilationUnit.ChildNodes().Where(x => x.IsKind(SyntaxKind.GlobalStatement))` will also work (but maybe the query in lines 34-35 is actually providing some optimization?). 
- Would it make sense to retrieve the first child of each `GlobalStatement` instead and return that, since my understanding, for example from this [AST](https://sharplab.io/#v2:C4LghgzsA0AmIGoA+ABATARgLACgCWAdsAAR7EC8xGA3KQgtbnvYzrigCzEAyA9gMZgANigCswPLwIAKAJTFcAb1zFVdBrgC+uXACcApv167YxAMpgAtgAch+pTpza2OAlf0RrYfvuIAVDxJlHHRiAEEAqGJFZ00gA==) is that the node of interest is the first child of the `GlobalStatement`.

analyzers/src/SonarAnalyzer.CSharp/Rules/MethodsShouldNotHaveTooManyLines.cs
```diff
@@ -86,5 +110,11 @@ protected override string GetMethodKindAndName(SyntaxToken identifierToken)
 
             return "method";
         }
+
+        private static long CountLines(IEnumerable<SyntaxNode> nodes) =>
```
mary-georgiou-sonarsource, line 114: Would it be fairer to maybe distinguish between local functions and expressions in top-level statements?
Not for now of course (baby steps). But would it be worth opening an issue?

mary-georgiou-sonarsource (commented):
Please check the comments.

analyzers/src/SonarAnalyzer.CSharp/Rules/MethodsShouldNotHaveTooManyLines.cs
costin-zaharia-sonarsource: For the moment I want to be consistent with the behavior for normal methods. I will open a discussion to see if we want to update the rule in this regard.

analyzers/src/SonarAnalyzer.CSharp/Extensions/CompilationUnitSyntaxExtensions.cs
costin-zaharia-sonarsource: > I would change this name to GlobalStatements. I think it describes better the purpose as my understating is that it gets all the GlobaStatement nodes under the `CompilationUnit`.

I prefer to use `GetTopLevelMainBody` since the intent is to get the Main method body, if later we find some exceptions, or if the syntax tree changes (new features added to the language) we will want to update the method in that regard and we will not focus on `GlobalStatements`.

> I think also compilationUnit.ChildNodes().Where(x => x.IsKind(SyntaxKind.GlobalStatement)) will also work (but maybe the query in lines 34-35 is actually providing some optimization?).

I prefer `TakeWhile(` since the iteration will stop once we finish reading `GlobalStatements`. Currently you cannot have `GlobalStatements` after type declarations start.

> Would it make sense to retrieve the first child of each GlobalStatement instead and return that, since my understanding, for example from this AST is that the node of interest is the first child of the GlobalStatement.

I prefer to let the nodes unchanged for now just to keep things simple. We can see later if they are not needed and if removing them will help in some regard.

mary-georgiou-sonarsource (approved):
LGTM

PR #5107, opened 2021-11-29
Rule S4144: support for top-level statements

mary-georgiou-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Extensions/TypeDeclarationSyntaxExtensions.cs
```diff
@@ -33,14 +33,19 @@ public static class TypeDeclarationSyntaxExtensions
...
+                           .OfType<MethodDeclarationSyntax>()
+                           .SelectMany(method => GetLocalFunctions(method).Union(new List<IMethodDeclaration> { MethodDeclarationFactory.Create(method) }));
 
-        private static IEnumerable<IMethodDeclaration> GetLocalFunctions(MethodDeclarationSyntax methodDeclaration)
-            => methodDeclaration.DescendantNodes()
-                .Where(member => member.IsKind(SyntaxKindEx.LocalFunctionStatement))
-                .Select(member => MethodDeclarationFactory.Create(member));
+        public static IEnumerable<IMethodDeclaration> GetMethodDeclarations(this CompilationUnitSyntax compilationUnitSyntax) =>
+            compilationUnitSyntax.Members.OfType<GlobalStatementSyntax>()
+                                         .Select(x => x.ChildNodes().FirstOrDefault(y => y.IsKind(SyntaxKindEx.LocalFunctionStatement)))
```
dmitry-golovinov-sonarsource, line 42: Let's check that it's immediate ancestor, then we will be sure that we won't be able to capture something else

analyzers/src/SonarAnalyzer.CSharp/Extensions/TypeDeclarationSyntaxExtensions.cs
```diff
@@ -33,14 +33,19 @@ public static class TypeDeclarationSyntaxExtensions
...
+                                         .Where(x => x != null)
+                                         .Select(x => MethodDeclarationFactory.Create(x));
+
+        private static IEnumerable<IMethodDeclaration> GetLocalFunctions(MethodDeclarationSyntax methodDeclaration) =>
+            methodDeclaration.DescendantNodes()
```
dmitry-golovinov-sonarsource, line 47: This could also capture local functions inside local functions. Is it a desirable behavior?

analyzers/src/SonarAnalyzer.CSharp/Extensions/TypeDeclarationSyntaxExtensions.cs
```diff
@@ -33,14 +33,19 @@ public static class TypeDeclarationSyntaxExtensions
...
-                .SelectMany(method => GetLocalFunctions(method).Union(new List<IMethodDeclaration> { MethodDeclarationFactory.Create(method) }));
+            typeDeclaration.Members
+                           .OfType<MethodDeclarationSyntax>()
+                           .SelectMany(method => GetLocalFunctions(method).Union(new List<IMethodDeclaration> { MethodDeclarationFactory.Create(method) }));
 
-        private static IEnumerable<IMethodDeclaration> GetLocalFunctions(MethodDeclarationSyntax methodDeclaration)
-            => methodDeclaration.DescendantNodes()
-                .Where(member => member.IsKind(SyntaxKindEx.LocalFunctionStatement))
-                .Select(member => MethodDeclarationFactory.Create(member));
+        public static IEnumerable<IMethodDeclaration> GetMethodDeclarations(this CompilationUnitSyntax compilationUnitSyntax) =>
```
dmitry-golovinov-sonarsource, line 40: This extension method should be placed in the corresponding file

analyzers/src/SonarAnalyzer.CSharp/Rules/MethodsShouldNotHaveIdenticalImplementations.cs
```diff
@@ -47,5 +58,11 @@ public sealed class MethodsShouldNotHaveIdenticalImplementations : MethodsShould
             && firstMethod.Body.IsEquivalentTo(secondMethod.Body, false);
 
         protected override SyntaxToken GetMethodIdentifier(IMethodDeclaration method) => method.Identifier;
+        protected override bool IsExcludedFromBeingExamined(ISymbol nodeContainingSymbol) =>
+            nodeContainingSymbol.Kind != SymbolKind.NamedType && !(nodeContainingSymbol is INamespaceSymbol namespaceSymbol
+                                                                   && namespaceSymbol.IsGlobalNamespace
```
dmitry-golovinov-sonarsource, line 63: Can't we just provide the node AND containing symbol? And then if the node is CompilationUnit - just check that this is toplevelprogram? I think that this would make this check more explainable.

analyzers/src/SonarAnalyzer.Common/Rules/MethodsShouldNotHaveIdenticalImplementationsBase.cs
```diff
@@ -26,20 +26,19 @@
...
+    public abstract class MethodsShouldNotHaveIdenticalImplementationsBase<TMethodDeclarationSyntax, TLanguageKindEnum> : SonarDiagnosticAnalyzer
         where TLanguageKindEnum : struct
     {
         protected const string DiagnosticId = "S4144";
         private const string MessageFormat = "Update this method so that its implementation is not identical to '{0}'.";
 
         private readonly DiagnosticDescriptor rule;
-
         protected abstract ILanguageFacade<TLanguageKindEnum> Language { get; }
+        protected abstract TLanguageKindEnum[] RegisteredSyntax { get; }
```
dmitry-golovinov-sonarsource, line 37: I think that we name it SyntaxKinds. Could you please double check?

analyzers/src/SonarAnalyzer.Common/Rules/MethodsShouldNotHaveIdenticalImplementationsBase.cs
```diff
@@ -26,20 +26,19 @@
...
         protected abstract IEnumerable<TMethodDeclarationSyntax> GetMethodDeclarations(SyntaxNode node);
         protected abstract SyntaxToken GetMethodIdentifier(TMethodDeclarationSyntax method);
         protected abstract bool AreDuplicates(TMethodDeclarationSyntax firstMethod, TMethodDeclarationSyntax secondMethod);
-
+        protected abstract bool IsExcludedFromBeingExamined(ISymbol nodeContainingSymbol);
```
dmitry-golovinov-sonarsource, line 41: EOL.

analyzers/src/SonarAnalyzer.VisualBasic/Rules/MethodsShouldNotHaveIdenticalImplementations.cs
```diff
@@ -49,5 +51,7 @@ protected override IEnumerable<MethodBlockSyntax> GetMethodDeclarations(SyntaxNo
 
         protected override SyntaxToken GetMethodIdentifier(MethodBlockSyntax method) =>
             method.SubOrFunctionStatement.Identifier;
+
+        protected override bool IsExcludedFromBeingExamined(ISymbol nodeContainingSymbol) => nodeContainingSymbol.Kind != SymbolKind.NamedType;
```
dmitry-golovinov-sonarsource, line 55: You can provide the default implementation in the base class and override it only where it is needed (C#). WDYT?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MethodsShouldNotHaveIdenticalImplementations.CSharp9.cs
```diff
@@ -1,18 +1,19 @@
 ﻿using System;
 
-void Method1()
+void Method1()// Secondary
+              // Secondary@-1
```
dmitry-golovinov-sonarsource, line 4: Let's introduce [] notation here to differentiate between to issues at line 10 and 16?

dmitry-golovinov-sonarsource (changes requested):
Please check my comments

analyzers/src/SonarAnalyzer.CSharp/Extensions/TypeDeclarationSyntaxExtensions.cs
mary-georgiou-sonarsource: Indeed.
I'll wait for this [PR](https://github.com/SonarSource/sonar-dotnet/pull/5105/files) to be merged and I'll apply this change.

analyzers/src/SonarAnalyzer.CSharp/Extensions/TypeDeclarationSyntaxExtensions.cs
mary-georgiou-sonarsource: If I understand correctly you are referring to `x.ChildNodes()`? This should return only the immediate ancestors and in the case of the `GlobalStatement` it should be only one.
Is there some case I miss?

dmitry-golovinov-sonarsource: Indeed!

analyzers/src/SonarAnalyzer.CSharp/Extensions/TypeDeclarationSyntaxExtensions.cs
mary-georgiou-sonarsource: This is not a change for this PR.
 I made the mistake to add the `GetMethodDeclarations(this CompilationUnitSyntax compilationUnitSyntax) ` above and it moved this one lower now it marks it as a change.
Apologies for that!

analyzers/src/SonarAnalyzer.Common/Rules/MethodsShouldNotHaveIdenticalImplementationsBase.cs
mary-georgiou-sonarsource: the `RegisteredSyntax` name you mean?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MethodsShouldNotHaveIdenticalImplementations.CSharp9.cs
mary-georgiou-sonarsource: I'm not sure I understand.
Could you provide me with an example?

analyzers/src/SonarAnalyzer.CSharp/Rules/MethodsShouldNotHaveIdenticalImplementations.cs
mary-georgiou-sonarsource: I tried to apply this but didn't manage to write something that looks good.
Any hints?
Meanwhile, I changed this function to evaluate the opposite (since I think it makes it more readable), and broke it into two pieces for the CSharp version for the same reason.
WDYT?

analyzers/src/SonarAnalyzer.Common/Rules/MethodsShouldNotHaveIdenticalImplementationsBase.cs
dmitry-golovinov-sonarsource: Yes.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MethodsShouldNotHaveIdenticalImplementations.CSharp9.cs
dmitry-golovinov-sonarsource: You can take a look at issues markup in the file analyzers\tests\SonarAnalyzer.UnitTest\TestCases\CastShouldNotBeDuplicated.CSharp10.cs

analyzers/src/SonarAnalyzer.Common/Rules/MethodsShouldNotHaveIdenticalImplementationsBase.cs
```diff
@@ -38,7 +38,7 @@ public abstract class MethodsShouldNotHaveIdenticalImplementationsBase<TMethodDe
         protected abstract IEnumerable<TMethodDeclarationSyntax> GetMethodDeclarations(SyntaxNode node);
         protected abstract SyntaxToken GetMethodIdentifier(TMethodDeclarationSyntax method);
         protected abstract bool AreDuplicates(TMethodDeclarationSyntax firstMethod, TMethodDeclarationSyntax secondMethod);
-        protected abstract bool IsExcludedFromBeingExamined(ISymbol nodeContainingSymbol);
+        protected virtual bool ShouldBeAnalyzed(ISymbol nodeContainingSymbol) => nodeContainingSymbol.Kind == SymbolKind.NamedType;
```
dmitry-golovinov-sonarsource, line 41: Implementation should be on the new line

analyzers/src/SonarAnalyzer.CSharp/Rules/MethodsShouldNotHaveIdenticalImplementations.cs
```diff
@@ -58,11 +58,12 @@ public sealed class MethodsShouldNotHaveIdenticalImplementations : MethodsShould
...
         protected override SyntaxToken GetMethodIdentifier(IMethodDeclaration method) => method.Identifier;
-        protected override bool IsExcludedFromBeingExamined(ISymbol nodeContainingSymbol) =>
-            nodeContainingSymbol.Kind != SymbolKind.NamedType && !(nodeContainingSymbol is INamespaceSymbol namespaceSymbol
-                                                                   && namespaceSymbol.IsGlobalNamespace
-                                                                   && namespaceSymbol.GetMembers()
-                                                                                     .OfType<INamedTypeSymbol>()
-                                                                                     .Any(x => x.IsTopLevelProgram()));
+        protected override bool ShouldBeAnalyzed(ISymbol nodeContainingSymbol) =>
+            nodeContainingSymbol.Kind == SymbolKind.NamedType
+            && nodeContainingSymbol is INamespaceSymbol namespaceSymbol
```
dmitry-golovinov-sonarsource: Still not understandable.  Why in this case this should be INamespaceSymbol and not IMethod (that we used before)? Can't we follow the same pattern and allow only in case it is IMethodSymbol (only to be able to process top level file case)? 

Anyways, I think that this part of the code deserves some comments.

analyzers/src/SonarAnalyzer.CSharp/Rules/MethodsShouldNotHaveIdenticalImplementations.cs
dmitry-golovinov-sonarsource: See below. Moreover, I think that our code style is more friendly to positive conditions, rather negative. So I think that the way it was before - matched this style better.

PR #5108, opened 2021-11-29
Introduce ExplodedNode in SE

pavel-mikula-sonarsource:


pavel-mikula-sonarsource:

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
```diff
@@ -42,20 +43,46 @@ public RoslynSymbolicExecution(ControlFlowGraph cfg, SymbolicExecutionCheck[] ch
...
             {
-                var state = new ProgramState(); // ToDo: This is a temporary oversimplification to scaffold the enginge
-                foreach (var operation in block.OperationsAndBranchValue.ToExecutionOrder()) // ToDo: This is a temporary oversimplification to scaffold the engine
+                queue.Enqueue(new ExplodedNode(block, ProgramState.Empty));
+                while (queue.Any())
                 {
-                    state = InvokeChecks(state, (check, ps) => check.PreProcess(ps, operation));
-                    // ToDo: Check state == null
+                    var current = queue.Dequeue();
+                    var successors = current.Operation == null ? ProcessBranching(current) : ProcessOperation(current);
```
andrei-epure-sonarsource, line 53: [edu] so for branching, the Operation is null? will it be based on block type?

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ExplodedNode.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    public sealed class ExplodedNode
+    {
+        private readonly IOperationWrapperSonar[] operations;   // Same memory footprint as ImmutableArray, regular array is marginally faster
+        private readonly int index;
+
+        public ProgramState State { get; }
+        public IOperationWrapperSonar Operation => index < operations.Length ? operations[index] : null;
+
+        public ExplodedNode(BasicBlock block, ProgramState state)
```
andrei-epure-sonarsource, line 37: could the block / ExplodedNode be null? it is a public API, so it should deal with null arguments?

or maybe this should be internal?

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/ExplodedNodeTest.cs
```diff
@@ -0,0 +1,100 @@
...
+            // Visualize operations to understand the rest of this UT
+            block.OperationsAndBranchValue.ToExecutionOrder().Select(TestHelper.Serialize).Should().OnlyContainInOrder(
+                 "LocalReference: value = 42 (Implicit)",
+                 "Literal: 42",
+                 "SimpleAssignment: value = 42 (Implicit)");
+
+            var sut = new ExplodedNode(block, ProgramState.Empty);
+            TestHelper.Serialize(sut.Operation).Should().Be("LocalReference: value = 42 (Implicit)");
+
+            sut = new ExplodedNode(sut, ProgramState.Empty);
```
andrei-epure-sonarsource, line 67: I think 
```
sut = new ExplodedNode(sut, _);
```
is not very friendly

I don't have a much better idea than `sut1`, `sut2`, `sut3`, `sut4`

OR

Have the list 
```
"LocalReference: value = 42 (Implicit)",
"Literal: 42",
"SimpleAssignment: value = 42 (Implicit)"
```
stored in an array and then loop through the array and verify
```
for (var op in ops)
{
  sut = new ExplodedNode(sut, ProgramState.Empty);
  TestHelper.Serialize(sut.Operation).Should().Be(op );
}
sut = new ExplodedNode(sut, ProgramState.Empty);
sut.Operation.Should().BeNull();
```

IMO having this assignment inside a loop makes it more readable

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.cs
```diff
@@ -63,16 +67,55 @@ public void SequentialInput_VB()
...
                 "ExpressionStatement: B = Not B",
                 "LocalReference: A",
                 "LocalReference: B",
                 "Parenthesized: (B)",
+                "SimpleAssignment: A = (B) (Implicit)",
                 "ExpressionStatement: A = (B)");
         }
+
+        [TestMethod]
+        public void PreProcess_Null_StopsExecution()
```
andrei-epure-sonarsource, line 89: open, possibly silly question: is there a way to design the system without relying on null?

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/FluentTestHelper.cs
```diff
@@ -41,7 +41,7 @@ internal static class FluentTestHelper
              where TAssertions : SelfReferencingCollectionAssertions<T, TAssertions>
         {
             _ = expected ?? throw new ArgumentNullException(nameof(expected));
-            self.Subject.Should().ContainInOrder(expected).And.HaveSameCount(expected);
+            self.Subject.Should().ContainInOrder(expected).And.BeEquivalentTo(expected).And.HaveSameCount(expected);    // BeEquivalentTo to have better collection message
```
andrei-epure-sonarsource, line 44: is `HaveSameCount` needed anymore? won't `BeEquivalentTo` fail if the numbers differ?

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SymbolicExecution/TestChecks.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Linq;
+using FluentAssertions;
+using SonarAnalyzer.SymbolicExecution.Roslyn;
+using SonarAnalyzer.UnitTest.Helpers;
+using StyleCop.Analyzers.Lightup;
+using ProcessFunc = System.Func<SonarAnalyzer.SymbolicExecution.Roslyn.ProgramState, StyleCop.Analyzers.Lightup.IOperationWrapperSonar, SonarAnalyzer.SymbolicExecution.Roslyn.ProgramState>;
+
+namespace SonarAnalyzer.UnitTest.TestFramework.SymbolicExecution
+{
+    internal class CollectorTestCheck : SymbolicExecutionCheck
```
andrei-epure-sonarsource, line 31: why not keep each Check in its own file under namespace `SonarAnalyzer.UnitTest.TestFramework.SymbolicExecution.TestChecks`

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestHelper.cs
```diff
@@ -59,6 +61,17 @@ public static (SyntaxTree, SemanticModel) Compile(string snippet, bool ignoreErr
...
+        public static ControlFlowGraph CompileCfgBodyCS(string body = null) =>
+            CompileCfg($"public class Sample {{ public void Main() {{ {body} }} }}");
+
+        public static ControlFlowGraph CompileCfgBodyVB(string body = null) =>
+            CompileCfg(
+$@"Public Class Sample
+    Public Sub Main()
+        {body}
+    End Sub
+End Class", false);
```
andrei-epure-sonarsource, line 73: nitpick (totally optional): using `isCSharp: false` named comment would increase readability (even if the file is small, I prefer verbosity in test code)

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
pavel-mikula-sonarsource: Yes, it is the same principle as in the old SE. Take a look at the `ExplodedNode.index` lifecycle. It goes `+1` in each step to process all instructions. Until it exceeds `operations.Length`, resulting in `null` being returned as Operation. Meaning that block branching should be processed.

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ExplodedNode.cs
pavel-mikula-sonarsource: Can be internal, at least for now.

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/ExplodedNodeTest.cs
pavel-mikula-sonarsource: The `for` approach doesn't work, because the 1st overload needs a `block` to start and the others needs `sut`.

I've renamed it to `current` to highlight that it changes over time.

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.cs
pavel-mikula-sonarsource: It is, but `null` is the best and simplest option for now. Other options:
* exception driven (baaad)
* array-based (not needed, see https://github.com/SonarSource/sonar-dotnet/pull/5100#discussion_r758575062)
* flag-based (but `flag` would be redundant to `state==null` anyway => confusing redundancy)

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SymbolicExecution/TestChecks.cs
pavel-mikula-sonarsource: Files: It's less than 1 page of code. I might move `CollectorTestCheck` out if it grows.
Namespace: Additional namespace would add only pain to maintain, the `TestFramework.SymbolicExecution` should remain reasonably small.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestHelper.cs
pavel-mikula-sonarsource: I actually don't like the `bool isCSharp` for being `bool` in general. It should be `AnalyzerLanguage language` instead. I'll get rid of it in another PR.

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.cs
andrei-epure-sonarsource: I was suggesting more an alternative State which no Empty but maybe `None` . 

Because `null` can lead to confusion too, as it can either be in  the program at some point on purpose or as a non-initialized variable.

Also, a program state None can also contain some additional meta-information which can be used to debug (e.g. what was the context in which it was initially returned)

pavel-mikula-sonarsource: I prefer the `null` variant. We don't need to deal with the question `what is the difference between null and None?`

andrei-epure-sonarsource (approved):
LGTM 

(and we can continue the discussion around null in the comment , not related strictly to this PR)

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestHelper.cs
pavel-mikula-sonarsource: See #5132

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.cs
andrei-epure-sonarsource: well, if we design a system in which null should not exist:
- `None` is intended
- `null` is a bug

PR #5113, opened 2021-11-29
Rule S1192: Introduce support for top level statements

dmitry-golovinov-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/StringLiteralShouldNotBeDuplicated.cs
```diff
@@ -60,5 +71,8 @@ public sealed class StringLiteralShouldNotBeDuplicated : StringLiteralShouldNotB
 
         protected override string GetLiteralValue(LiteralExpressionSyntax literal) =>
             literal.Token.Text;
+
+        protected override bool IsNotNamedTypeOrTopLevelMain(SyntaxNodeAnalysisContext context) =>
```
costin-zaharia-sonarsource, line 75: I find the negation in the method name very hard to read. It's also a bit confusing since I don't know from start if the negation applies for top-level main too.

I suggest to re-write the code to avoid the negation in the method name.
Additionally, we can split this in 2 different methods: `IsTopLevelMain` and `IsNamedType`

costin-zaharia-sonarsource (changes requested):

costin-zaharia-sonarsource (approved):
LGTM!

PR #5122, opened 2021-11-30
Improve analyzer performance when no SE rule is active

### LINKED ISSUE #5120: Improve analyzer performance when no SE rule is active
We don't need to run SE engine when there's no SE rule active. For example on a Test Code

pavel-mikula-sonarsource:
Fixes #5120


andrei-epure-sonarsource (approved):
LGTM

PR #5123, opened 2021-11-30
Fix S1905: false positive with OfType

### LINKED ISSUE #5003: S1905 reports false positive with `OfType` and nullable types
### Description

It looks like #3273 is back in version 8.30.0.37606:
image

### Repro steps

```
public static IEnumerable<T> WhereNotNull<T>(this IEnumerable<T?> result) =>
            result.OfType<T>();
```

### Expected behavior

S1905 does not trigger.

### Actual behavior

S1905 triggers.
### Related information

* C# analyzer 8.30.0.37606
* Visual Studio version 16.11.5
* MSBuild / dotnet version 5.0.402

dmitry-golovinov-sonarsource:
Documents/Fixes #5003

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantCast.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantCast.cs
@@ -111,7 +111,9 @@ namespace SonarAnalyzer.Rules.CSharp
                     }
 
                     var elementType = GetElementType(invocation, methodSymbol, context.SemanticModel);
-                    if (elementType != null && elementType.Equals(castType))
+                    // Generic types {T} and {T?} are equal and there is no way to access NullableAnnotation field right now
+                    // See https://github.com/SonarSource/sonar-dotnet/issues/3273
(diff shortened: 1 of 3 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/RedundantCast.cs
```diff
@@ -111,7 +111,9 @@ private static void CheckExtensionMethodInvocation(SyntaxNodeAnalysisContext con
                     }
 
                     var elementType = GetElementType(invocation, methodSymbol, context.SemanticModel);
-                    if (elementType != null && elementType.Equals(castType))
+                    // Generic types {T} and {T?} are equal and there is no way to access NullableAnnotation field right now
+                    // See https://github.com/SonarSource/sonar-dotnet/issues/3273
+                    if (elementType != null && elementType.Equals(castType) && elementType.ToString() == castType.ToString())
```
csaba-sagi-sonarsource, line 116: Use string.equals with StringComparison.Ordinal

csaba-sagi-sonarsource (approved):
I left one comment, please address it, otherwise good to go.

PR #5124, opened 2021-11-30
Make test verbosity minimal

andrei-epure-sonarsource:


andrei-epure-sonarsource:
Some thoughts from slack:
- The detailed verbosity was introduced recently in #4903 by @Costin to deal with language version issues.
- Indeed, if I remember correctly I've introduced this to help us identify issues with failed tests when depend on a specific language version.
- I've found that there is also an option to use TestContext.Out for the same result and I need to check if it works in minimal mode. Maybe we can simplify it if we really have a problem

andrei-epure-sonarsource:
- with verbose logs: 7-8 minutes https://dev.azure.com/sonarsource/DotNetTeam%20Project/_build/results?buildId=40306&view=results
- with minimal: 3-4 minutes https://dev.azure.com/sonarsource/DotNetTeam%20Project/_build/results?buildId=40279&view=results

andrei-epure-sonarsource:
My bad:
- the time on master is actually because it runs all tests, not because the logging.
- on PRs, logging doesn't impact the UT execution time

PR #5125, opened 2021-11-30
Rule S2930: Remove duplicate issues when in top-level statements.

mary-georgiou-sonarsource:


costin-zaharia-sonarsource (changes requested):
It looks good, I've added just a few small comments.

PR #5126, opened 2021-12-01
Enable preview features for .net6 ITs

csaba-sagi-sonarsource:


analyzers/its/sources/Net6/Net6/Net6.csproj
```diff
@@ -5,7 +5,7 @@
     <TargetFramework>net6.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>disable</Nullable>
-    <LangVersion>preview</LangVersion>
```
costin-zaharia-sonarsource, line 8: It was intended to remove this?

csaba-sagi-sonarsource: Yes. In VS 2022-preview with that line the preview features were enabled at the build was successful, however it changed and the build was failing. Now there is this new way to enable the preview features.

costin-zaharia-sonarsource: As far as I can tell these are 2 very different properties, `LangVersion` is related to the language version, and `EnablePreviewFeatures` is related to APIs. See: https://github.com/dotnet/designs/blob/main/accepted/2021/preview-features/preview-features.md#enabling-a-preview-feature

costin-zaharia-sonarsource: In any case, we don't really need `<LangVersion>preview</LangVersion>` now so I can merge it like it is. I was though curious if there was a reason behind the change.

costin-zaharia-sonarsource (approved):
LGTM!

PR #5130, opened 2021-12-01
Adding ITs for top level statements

dmitry-golovinov-sonarsource:


analyzers/its/sources/Net5/Net5/Main.cs
```diff
@@ -15,4 +17,204 @@
...
+// S3928
+if (args == null)
+{
+    throw new ArgumentException("args", "message");
+}
+
+// S2760
+int a = 0, b = 1;
+if (a == b) { LocalMethodInMain(); }
+if (a == b) { AnotherLocalMethodInMain(); }
```
mary-georgiou-sonarsource, line 29: This rule is covering also the `switch` statement.
Could you please add also such a case here [like so](https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ConditionalsWithSameCondition.CSharp9.TopLevelStatements.cs#L26), so that we are sure that this is covered as well?

mary-georgiou-sonarsource (approved):
LGTM - Just a small comment.

PR #5131, opened 2021-12-01
Introduce SymbolicValue in SE

pavel-mikula-sonarsource:
`Squash ExplodedNode` commit doesn't need a review

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
+++ b/analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
@@ -18,11 +18,25 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.Collections.Immutable;
+using StyleCop.Analyzers.Lightup;
+
 namespace SonarAnalyzer.SymbolicExecution.Roslyn
 {
     public sealed class ProgramState
     {
-        public static readonly ProgramState Empty = new ProgramState();
-        // ToDo: Some stuff is missing here
+        public static readonly ProgramState Empty = new ProgramState(ImmutableDictionary<IOperationWrapperSonar, SymbolicValue>.Empty);
+
+        private readonly ImmutableDictionary<IOperationWrapperSonar, SymbolicValue> operationValue;
+
+        public SymbolicValue this[IOperationWrapperSonar operation] => operationValue.TryGetValue(operation, out var value) ? value : null;
+
+        private ProgramState(ImmutableDictionary<IOperationWrapperSonar, SymbolicValue> operationValue)
+        {
+            this.operationValue = operationValue;
+        }
+
+        public ProgramState AddOperationValue(IOperationWrapperSonar operation, SymbolicValue value) =>
+            new ProgramState(operationValue.Add(operation, value));
     }
 }
--- a/analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
+++ b/analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
@@ -27,7 +27,7 @@ namespace SonarAnalyzer.SymbolicExecution.Roslyn
     {
         public static readonly ProgramState Empty = new ProgramState(ImmutableDictionary<IOperationWrapperSonar, SymbolicValue>.Empty);
 
-        private readonly ImmutableDictionary<IOperationWrapperSonar, SymbolicValue> operationValue;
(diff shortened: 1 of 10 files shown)


pavel-mikula-sonarsource:

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
```diff
@@ -18,11 +18,25 @@
...
 {
     public sealed class ProgramState
     {
-        public static readonly ProgramState Empty = new ProgramState();
-        // ToDo: Some stuff is missing here
+        public static readonly ProgramState Empty = new ProgramState(ImmutableDictionary<IOperationWrapperSonar, SymbolicValue>.Empty);
+
+        private readonly ImmutableDictionary<IOperationWrapperSonar, SymbolicValue> operationValue;
+
+        public SymbolicValue this[IOperationWrapperSonar operation] => operationValue.TryGetValue(operation, out var value) ? value : null;
```
andrei-epure-sonarsource, line 32: nitpick: public properties/accessor before private fields?

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
```diff
@@ -18,11 +18,25 @@
...
+
 namespace SonarAnalyzer.SymbolicExecution.Roslyn
 {
     public sealed class ProgramState
     {
-        public static readonly ProgramState Empty = new ProgramState();
-        // ToDo: Some stuff is missing here
+        public static readonly ProgramState Empty = new ProgramState(ImmutableDictionary<IOperationWrapperSonar, SymbolicValue>.Empty);
+
+        private readonly ImmutableDictionary<IOperationWrapperSonar, SymbolicValue> operationValue;
```
andrei-epure-sonarsource, line 30: I think a comment explaining what this is will help - is the dictionary value the result of the operation?

or maybe rename to `operationResult` (and from the signature we can see the result is a SymbolicValue)?  I'd find this more intuitive

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
```diff
@@ -18,10 +18,25 @@
...
+
+        public SymbolicValue this[IOperationWrapperSonar operation] => operationValue.TryGetValue(operation, out var value) ? value : null;
+
+        private ProgramState(ImmutableDictionary<IOperationWrapperSonar, SymbolicValue> operationValue)
+        {
+            this.operationValue = operationValue;
+        }
+
+        public ProgramState AddOperationValue(IOperationWrapperSonar operation, SymbolicValue value) =>
+            new ProgramState(operationValue.Add(operation, value));
```
andrei-epure-sonarsource, line 40: double-check : if the `key` already exists, it is an illegal operation and it's fine to let the dictionary throw?

please add a test for this scenario?

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/SymbolicValue.cs
```diff
@@ -0,0 +1,33 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarAnalyzer.SymbolicExecution.Roslyn
+{
+    public class SymbolicValue
+    {
+        private readonly int identifier;
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: This used to be `static int` int the old SE. Shared for the lifetime of the Roslyn instance. So we should be fine now 😆

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
pavel-mikula-sonarsource: This is correct. We should have all fields (public, protected, private) followed by all properties (public, protected, private).

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
pavel-mikula-sonarsource: Good catch

andrei-epure-sonarsource (approved):
LGTM!

PR #5132, opened 2021-12-02
Refactor TestHelper isCSharp to AnalyzerLanguage

pavel-mikula-sonarsource:
`Squash ExplodedNode` commit doesn't need a review

Follow up for https://github.com/SonarSource/sonar-dotnet/pull/5108#discussion_r760438989

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestHelper.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestHelper.cs
@@ -49,20 +49,23 @@ namespace SonarAnalyzer.UnitTest
     <OutPath>{2}</OutPath>
 </SonarProjectConfig>";
 
-        public static (SyntaxTree, SemanticModel) CompileIgnoreErrors(string snippet, bool isCSharp = true, params MetadataReference[] additionalReferences) =>
-            Compile(snippet, true, isCSharp, additionalReferences);
+        public static (SyntaxTree, SemanticModel) CompileIgnoreErrorsCS(string snippet, params MetadataReference[] additionalReferences) =>
+            Compile(snippet, true, AnalyzerLanguage.CSharp, additionalReferences);
 
-        public static (SyntaxTree, SemanticModel) Compile(string snippet, bool isCSharp = true, params MetadataReference[] additionalReferences) =>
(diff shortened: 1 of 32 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestHelper.cs
```diff
@@ -49,37 +49,43 @@ internal static class TestHelper
...
-        public static (SyntaxTree, SemanticModel) Compile(string snippet, bool isCSharp = true, params MetadataReference[] additionalReferences) =>
-            Compile(snippet, false, isCSharp, additionalReferences);
+        public static (SyntaxTree Tree, SemanticModel Model) CompileCS(string snippet, params MetadataReference[] additionalReferences) =>
+            Compile(snippet, false, AnalyzerLanguage.CSharp, additionalReferences);
 
-        public static (SyntaxTree, SemanticModel) Compile(string snippet, bool ignoreErrors, bool isCSharp, params MetadataReference[] additionalReferences)
+        public static (SyntaxTree Tree, SemanticModel Model) CompileVB(string snippet, params MetadataReference[] additionalReferences) =>
+            Compile(snippet, false, AnalyzerLanguage.VisualBasic, additionalReferences);
+
+        public static (SyntaxTree Tree, SemanticModel Model) Compile(string snippet, bool ignoreErrors, AnalyzerLanguage language, params MetadataReference[] additionalReferences)
```
csaba-sagi-sonarsource, line 61: `Compile` and `CompileCfg` methods should be private as there are other ` AnalyzerLanguages` that are not handled in the method. Where those methods are used now the CS and VB variants could be used to replace it.

csaba-sagi-sonarsource (changes requested):
One single comment left.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestHelper.cs
pavel-mikula-sonarsource: I've replaced the single usage that could be simplified.

csaba-sagi-sonarsource (approved):
Please try to improve the coverage, otherwise good to go!

PR #5133, opened 2021-12-02
Remove ITs/Expected from Solution Explorer

pavel-mikula-sonarsource:
It obstructs search results when searching for `reference`, `serial`, `engine`, ... in solution explorer file filter

### DIFF
--- a/analyzers/SonarAnalyzer.sln
+++ b/analyzers/SonarAnalyzer.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29519.87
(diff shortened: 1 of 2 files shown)


analyzers/SonarAnalyzer.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29519.87
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31912.275
```
pavel-mikula-sonarsource, line 3: Autogenerated noise

andrei-epure-sonarsource (approved):
lgtm

PR #5138, opened 2021-12-02
Use C# 9

pavel-mikula-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
```diff
@@ -1,11 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <LangVersion>9.0</LangVersion>
```
pavel-mikula-sonarsource, line 4: We already used 9 in this single project from #4830. But it was not used on all other places.

analyzers/Analyzers.targets
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
```
pavel-mikula-sonarsource, line 1: src+test props are merged in this single common file.

It's all renamed from `*.props` to `*.targets` to have a single file extension to deal with. And because the `props` files do on only set properties, but do other things as well.

analyzers/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
```diff
@@ -1,11 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <LangVersion>preview</LangVersion>
```
pavel-mikula-sonarsource, line 4: We should not use preview features. Switched to C# 9 instead.

analyzers/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
```diff
@@ -1,11 +1,9 @@
...
     <IsPackable>false</IsPackable>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
     <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
-    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
```
pavel-mikula-sonarsource, line 8: Moved to common `Analyzers.targets`

analyzers/tests/SonarAnalyzer.UnitTest/Wrappers/IOperationWrapperSonarTest.cs
```diff
@@ -96,9 +96,10 @@ public void Method()
         var value = 42;
     }
 }";
-            (var tree, semanticModel) = TestHelper.Compile(code);
+            var (tree, model) = TestHelper.Compile(code);
```
pavel-mikula-sonarsource, line 99: Removal of `preview` language feature

analyzers/Analyzers.targets
andrei-epure-sonarsource: IMO keeping the `Directory.Build.props` for properties and .targets for targets means respecting the standard - https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022#directorybuildprops-and-directorybuildtargets

IMO we should keep things simple and follow standards when we can.

andrei-epure-sonarsource: >And because the props files do on only set properties, but do other things as well.

what do you mean? `analyzers/Directory.Build.props` only sets properties, except `<SourceRoot Include="$(MSBuildThisFileDirectory)/"/>` which respects the [guideline](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022#choose-between-adding-properties-to-a-props-or-targets-file):

> Include items in .props files (conditioned on a property). All properties are considered before any item, so user-project property customizations get picked up, and this gives the user's project the opportunity to Remove or Update any item brought in by the import.

I'd rather have the conversation based on what we need:
-     .props files are imported early in the import order.
-     .targets files are imported late in the build order.

And the [guideline](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022#choose-between-adding-properties-to-a-props-or-targets-file).

If in the end we'll use `Build.props.targets` I'm fine with that, as long as we make a conscious decision knowing the difference in evaluation and what could possibly in the build.

analyzers/src/Directory.Build.targets
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <Import Project="$(MSBuildThisFileDirectory)\..\Analyzers.targets" />
```
andrei-epure-sonarsource, line 3: this makes sense to be a `Directory.Build.targets` because it's importing the other targets

analyzers/src/Directory.Build.targets
```diff
@@ -0,0 +1,8 @@
...
+
+  <PropertyGroup>
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)\..\.sonarlint\sonaranalyzer-dotnetCSharp.ruleset</CodeAnalysisRuleSet>
+  </PropertyGroup>
+</Project>
```
andrei-epure-sonarsource, line 8: nitpick EOL

andrei-epure-sonarsource (changes requested):
let's discuss Directory.Build.props

analyzers/Analyzers.targets
pavel-mikula-sonarsource: There's [Choose between adding properties to .props or .targets file](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022#choose-between-adding-properties-to-a-props-or-targets-file) section mentioning that placing properties is not a standard but a convention.

The functional difference is only the import order. We don't need to prefer any order functionally. I prefer to have it in `targets` to prevent overriding on a project level. If we need to override something, it should not be defined in `targets`. Or it should be handled explicitly/conditionally there. The purpose of this is to avoid inconsistency (see the example of having `LangVersion`, 8, 9 and preview at the same time)

In my understanding `Props` file should only define properties, being a data-only file. Where `Targets` files can define anything, including additional logic. I don't like the fact that the property file used to define project items, imported targets files and that targets file imported nuget package. 

To look at it from the other side, I'd use `targets` (for all logic) AND `props` (for properties-only) file in case we'd need to import properties before the project or we would have a lot of such properties (like [Roslyn has](https://github.com/dotnet/roslyn/blob/main/eng/Versions.props))

pavel-mikula-sonarsource:
Converting to draft to prevent me from merging it before the release

andrei-epure-sonarsource (approved):
LGTM

analyzers/Analyzers.targets
andrei-epure-sonarsource: Ok, maybe we should make this explicit somewhere, because I think it's an important convention for devs ramping up the project structure.

PR #5139, opened 2021-12-02
BUILD-1158: Use staging pool with PR-27 demand

tomverin:
Fixes #


tomverin:
Pipeline is green, validated the process, no need to merge as no functionnal changes comes from here

pavel-mikula-sonarsource:
Works

PR #5143, opened 2021-12-06
Delete VSIX project

costin-zaharia-sonarsource:


analyzers/src/packaging/SonarAnalyzer.CSharp.nuspec
```diff
@@ -22,10 +22,10 @@
     </frameworkAssemblies>
   </metadata>
   <files>
-    <file src="bin\Release\SonarAnalyzer.CFG.dll" target="analyzers" />
-    <file src="bin\Release\SonarAnalyzer.dll" target="analyzers" />
-    <file src="bin\Release\SonarAnalyzer.CSharp.dll" target="analyzers" />
-    <file src="bin\Release\Google.Protobuf.dll" target="analyzers" />
+    <file src="..\SonarAnalyzer.CSharp\bin\Release\net46\SonarAnalyzer.CFG.dll" target="analyzers" />
```
pavel-mikula-sonarsource, line 25: Can this whole directory be in `analyzers\packaging` instead of `analyzers\src\packaging`?

analyzers/its/.gitignore
```diff
@@ -1,6 +1,3 @@
-# SonarLint binaries
```
pavel-mikula-sonarsource, line 1: I'd be also fine with deleting this whole file and moving its content to the root. To manage these settings on a single place.

.gitignore
```diff
@@ -63,3 +63,6 @@ packages/
 *.restext
 /analyzers/src/SonarAnalyzer.CSharp/CBDE/windows
 .generated/
+
+# Analizer binaries
```
pavel-mikula-sonarsource, line 67:
```suggestion
# Analyzer binaries
```

analyzers/src/SonarAnalyzer.VisualBasic/SonarAnalyzer.VisualBasic.csproj
```diff
@@ -30,6 +30,18 @@
     <DefineConstants>TRACE;VB</DefineConstants>
   </PropertyGroup>
 
+  <Target Name="SignDlls" Condition="'$(SignAssembly)' == 'true' And '$(TargetFramework)' == 'NET46' " BeforeTargets="CopyBinaries">
```
pavel-mikula-sonarsource, line 33: I think the `BeforeTargets` from the VSIX worked only by accident and can break at any point. IMO we should use this instead.
```
<Target Name="CopyBinaries" AfterTargets="Build" DependsOnTargets="SignDlls" Condition=" '$(TargetFramework)' == 'NET46' ">
```
Because in theory, msbuild can process these 3 targets in this order:
* SignDlls (will fail)
* Build
* CopyBinaries

analyzers/its/regression-test.ps1
```diff
@@ -457,8 +457,8 @@ try {
     $msBuildImportBeforeCurrent = Get-MSBuildImportBeforePath "Current"
 
     Push-Location $PSScriptRoot
-    Test-FileExists ".\binaries\SonarAnalyzer.dll"
-    Test-FileExists ".\binaries\SonarAnalyzer.CFG.dll"
+    Test-FileExists ".\..\binaries\SonarAnalyzer.dll"
+    Test-FileExists ".\..\binaries\SonarAnalyzer.CFG.dll"
```
pavel-mikula-sonarsource, line 461:
```suggestion
    Test-FileExists "..\binaries\SonarAnalyzer.dll"
    Test-FileExists "..\binaries\SonarAnalyzer.CFG.dll"
```

analyzers/its/SonarAnalyzer.Testing.ImportBefore.targets
```diff
@@ -3,7 +3,8 @@
           AfterTargets="ResolveCodeAnalysisRuleSet"
           BeforeTargets="CoreCompile">
     <PropertyGroup>
-      <SonarAnalyzerRegressionTestRunning Condition="Exists('$(MSBuildStartupDirectory)\binaries\SonarAnalyzer.dll')">true</SonarAnalyzerRegressionTestRunning>
+      <BinariesFolder>$(MSBuildStartupDirectory)\..\binaries\</BinariesFolder>
```
pavel-mikula-sonarsource, line 6: Because the next line has it's own slash `$(BinariesFolder)\SonarAnalyzer.dll`
```suggestion
      <BinariesFolder>$(MSBuildStartupDirectory)\..\binaries</BinariesFolder>
```

analyzers/src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
```diff
@@ -46,10 +46,26 @@
   <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
     <DefineConstants>TRACE;CS</DefineConstants>
   </PropertyGroup>
+
   <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
     <DefineConstants>TRACE;CS</DefineConstants>
   </PropertyGroup>
 
+  <Target Name="SignDlls" Condition="'$(SignAssembly)' == 'true' And '$(TargetFramework)' == 'NET46' " BeforeTargets="CopyBinaries">
```
pavel-mikula-sonarsource, line 54: Same concern as on C#

analyzers/src/SonarAnalyzer.VisualBasic/SonarAnalyzer.VisualBasic.csproj
```diff
@@ -30,6 +30,18 @@
...
+    <Message Importance="high" Text="Signing analyzer assemblies from $(OutputPath)" />
+    <Exec WorkingDirectory="$(OutputPath)" Command="&quot;$(SIGNTOOL_PATH)&quot; sign /fd SHA256 /f $(PFX_PATH) /p $(PFX_PASSWORD) /tr http://sha256timestamp.ws.symantec.com/sha256/timestamp SonarAnalyzer*.dll" />
+  </Target>
+
+  <Target Name="CopyBinaries" AfterTargets="Build" Condition=" '$(TargetFramework)' == 'NET46' ">
+    <ItemGroup>
+      <BinariesToCopy Include="$(OutputPath)\SonarAnalyzer.VisualBasic.dll" />
```
pavel-mikula-sonarsource, line 40: Let's add these as well.
```
      <BinariesToCopy Include="$(OutputPath)\SonarAnalyzer.dll" />
      <BinariesToCopy Include="$(OutputPath)\Google.Protobuf.dll" />
```
Without that, cleaning CSharp project and rebuilding VB one will not play nicely with running ITs.

pavel-mikula-sonarsource (changes requested):

analyzers/src/packaging/SonarAnalyzer.CSharp.nuspec
pavel-mikula-sonarsource: Like that, we could have `analyzers\packaging\binaries` to keep the directory count in `analyzers` low

analyzers/its/SonarAnalyzer.Testing.ImportBefore.targets
costin-zaharia-sonarsource: Not really a problem but sure, I can remove it.

analyzers/src/Directory.Build.props
```diff
@@ -18,4 +20,8 @@
     </Compile>
   </ItemGroup>
 
-</Project>
\ No newline at end of file
+  <Target Name="CleanBinaries" AfterTargets="Clean">
```
pavel-mikula-sonarsource, line 23: _Rebase hint:_ This file was renamed to `.targets`
Most of the content was moved to `..\Analyzers.targets` and is shared between `src` and `tests`.
Your changes makes sense to remain in this `Directory.Build.targets`.

analyzers/packaging/SonarAnalyzer.CSharp.nuspec
```diff
@@ -22,12 +22,12 @@
     </frameworkAssemblies>
   </metadata>
   <files>
-    <file src="bin\Release\SonarAnalyzer.CFG.dll" target="analyzers" />
-    <file src="bin\Release\SonarAnalyzer.dll" target="analyzers" />
-    <file src="bin\Release\SonarAnalyzer.CSharp.dll" target="analyzers" />
-    <file src="bin\Release\Google.Protobuf.dll" target="analyzers" />
+    <file src="..\src\SonarAnalyzer.CSharp\bin\Release\net46\SonarAnalyzer.CFG.dll" target="analyzers" />
```
pavel-mikula-sonarsource, line 25: An idea for future refactoring (not this PR): We can unify `install.ps1` and `uninstall.ps1` for C# and VB.NET

analyzers/src/SonarAnalyzer.VisualBasic/SonarAnalyzer.VisualBasic.csproj
pavel-mikula-sonarsource: The ` BeforeTargets="CopyBinaries"` should be removed now, the `DependsOnTargets` takes care of it

analyzers/its/SonarAnalyzer.Testing.ImportBefore.targets
```diff
@@ -3,7 +3,8 @@
           AfterTargets="ResolveCodeAnalysisRuleSet"
           BeforeTargets="CoreCompile">
     <PropertyGroup>
-      <SonarAnalyzerRegressionTestRunning Condition="Exists('$(MSBuildStartupDirectory)\binaries\SonarAnalyzer.dll')">true</SonarAnalyzerRegressionTestRunning>
+      <BinariesFolder>$(MSBuildStartupDirectory)\..\packaging\binaries</BinariesFolder>
+      <SonarAnalyzerRegressionTestRunning Condition="Exists('$(BinariesFolder)\SonarAnalyzer.dll')">true</SonarAnalyzerRegressionTestRunning>
```
pavel-mikula-sonarsource, line 7: _Not in this PR:_ Let's remove `SonarAnalyzerRegressionTestRunning` completely. 
We can update line 40 like this:
```
<Error Text="Could not find '$(BinariesFolder)\SonarAnalyzer.dll'" Condition="Exists('$(BinariesFolder)\SonarAnalyzer.dll')" />
```
and move that line right after this `<PropertyGroup>`

azure-pipelines.yml
```diff
@@ -90,7 +90,7 @@ stages:
         displayName: "Build NuGet packages"
         inputs:
           command: pack
-          packagesToPack: 'analyzers/src/**/*.nuspec'
+          packagesToPack: 'analyzers/src/**/*.nuspec;analyzers/packaging/**/*.nuspec'
```
pavel-mikula-sonarsource, line 93: AFAICT there are no `nuspec` files in `src` now
```suggestion
          packagesToPack: 'analyzers/packaging/**/*.nuspec'
```

costin-zaharia-sonarsource: This will skip the CFG package generation.

pavel-mikula-sonarsource: Or like this
```suggestion
          packagesToPack: 'analyzers/src/**/*.nuspec;analyzers/packaging/*.nuspec'
```

pavel-mikula-sonarsource (approved):
Awesome!

analyzers/its/SonarAnalyzer.Testing.ImportBefore.targets
pavel-mikula-sonarsource: Done in #5154

azure-pipelines.yml
```diff
@@ -90,7 +90,7 @@ stages:
         displayName: "Build NuGet packages"
         inputs:
           command: pack
-          packagesToPack: 'analyzers/src/**/*.nuspec'
+          packagesToPack: 'analyzers/src/**/*.nuspec;analyzers/packaging/*.nuspec'
```
andrei-epure-sonarsource, line 93: @costin-zaharia-sonarsource  are both needed? do we have nuspecs in other folders than packaging?

pavel-mikula-sonarsource: I asked the same question 😄  https://github.com/SonarSource/sonar-dotnet/pull/5143#discussion_r763871347

It's needed for CFG

pavel-mikula-sonarsource: And in theory, we could move the CFG out as well

PR #5147, opened 2021-12-06
Track symbols in ProgramState

pavel-mikula-sonarsource:
Squash* commits do not need a review


pavel-mikula-sonarsource:

csaba-sagi-sonarsource (approved):
LGTM!

PR #5152, opened 2021-12-07
Sign .NET assemblies only on master

costin-zaharia-sonarsource:


azure-pipelines.yml
```diff
@@ -18,6 +18,8 @@ variables:
   # To make sure we don`t use MsBuild 17.
   - name: MsBuildPath
     value: 'C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin\msbuild.exe'
+  - name: isMaster
```
pavel-mikula-sonarsource, line 21: Early comment on WIP PR: `master` OR `branch-*`

So maybe something like
```suggestion
  - name: isReleaseBranch
```

azure-pipelines.yml
```diff
@@ -18,6 +18,8 @@ variables:
   # To make sure we don`t use MsBuild 17.
   - name: MsBuildPath
     value: 'C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin\msbuild.exe'
+  - name: isReleaseBranch
```
pavel-mikula-sonarsource, line 21: Nice!

azure-pipelines.yml
```diff
@@ -83,7 +85,7 @@ stages:
           solution: '$(solution)'
           platform: '$(buildPlatform)'
           configuration: '$(buildConfiguration)'
-          msbuildArgs: '/p:DeployExtension=false /p:SignAssembly=true /p:AssemblyOriginatorKeyFile="$(snk.secureFilePath)" /p:Sha1=$(Build.SourceVersion) /p:BuildNumber=$(Build.BuildId) /p:WarningLevel=0'
+          msbuildArgs: '/p:DeployExtension=false /p:SignAssembly=$(isReleaseBranch) /p:AssemblyOriginatorKeyFile="$(snk.secureFilePath)" /p:Sha1=$(Build.SourceVersion) /p:BuildNumber=$(Build.BuildId) /p:WarningLevel=0'
```
pavel-mikula-sonarsource, line 88: Speaking of signing, where are the JAR files actually signed?

Running this on our latest release gives `jar is unsigned.` 😒 
>jarsigner -verify sonar-csharp-plugin-8.33.0.40503.jar

pavel-mikula-sonarsource (commented):
Some questions

azure-pipelines.yml
costin-zaharia-sonarsource: The signing should happen here: https://github.com/SonarSource/sonar-dotnet/blob/master/azure-pipelines.yml#L385 but it looks like it does not work. I've added a card in Trello for this.

azure-pipelines.yml
```diff
@@ -98,6 +100,7 @@ stages:
 
       - task: NuGetCommand@2
         displayName: "Sign NuGet packages"
+        condition: eq('$(isReleaseBranch)', 'true')
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Not for conditions. You can find here the example from the docs: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml#run-if-a-variable-is-set-to-true

If you try your snippet, the build fails with: `##[error]Unrecognized value: '$'. Located at position 1 within expression: $(isReleaseBranch).` See https://dev.azure.com/sonarsource/DotNetTeam%20Project/_build/results?buildId=40647&view=logs&j=645a1db5-d210-5b98-34db-87d2a8fd8bb0&t=87dc8e68-39f9-4481-9ac2-035b8bab1738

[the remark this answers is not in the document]
costin-zaharia-sonarsource: Not for conditions. You can find here the example from the docs: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml#run-if-a-variable-is-set-to-true

If you try your snippet, the build fails with: `##[error]Unrecognized value: '$'. Located at position 1 within expression: $(isReleaseBranch).` See https://dev.azure.com/sonarsource/DotNetTeam%20Project/_build/results?buildId=40647&view=logs&j=645a1db5-d210-5b98-34db-87d2a8fd8bb0&t=87dc8e68-39f9-4481-9ac2-035b8bab1738

azure-pipelines.yml
costin-zaharia-sonarsource: Getting back to this. We don't use the jar signer tool for our signing. Instead, the signature is detached and we use PGP signatures file which can be used to authenticate the artifacts. In this case, we have [sonar-csharp-plugin-8.33.0.40503-sources.jar.asc](https://repox.jfrog.io/artifactory/sonarsource-public-releases/org/sonarsource/dotnet/sonar-csharp-plugin/8.33.0.40503/sonar-csharp-plugin-8.33.0.40503.jar.asc).

costin-zaharia-sonarsource: To sum it up, we sign now all the java artifacts. We always did. I will tackle java signing for PR branches separately.

pavel-mikula-sonarsource (approved):
LGTM

PR #5153, opened 2021-12-07
Ordinal block processing in SE

pavel-mikula-sonarsource:
`Squash ImproveChecks` commit doesn't need a review


analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SymbolicExecution/CollectorTestCheck.cs
```diff
@@ -30,20 +31,42 @@ namespace SonarAnalyzer.UnitTest.TestFramework.SymbolicExecution
     internal class CollectorTestCheck : SymbolicCheck
```
csaba-sagi-sonarsource, line 31: Very nice idea, it simplifies the tests a lot.

csaba-sagi-sonarsource (approved):
Please improve the coverage. 100% should be easy to achieve.

pavel-mikula-sonarsource:
> Please improve the coverage. 100% should be easy to achieve.

I had to remove the null check, because `null` can not pass through the public constructors due to NREs there.

pavel-mikula-sonarsource:

PR #5156, opened 2021-12-07
Add UTs w/ Mutex for S2222

andrei-epure-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.NetFramework.cs
```diff
@@ -0,0 +1,97 @@
+﻿using System;
+using System.Security.AccessControl;
+using System.Threading;
+
+namespace Net6
+{
+    public class MutexTest
+    {
+        public static void AcquiredNotReleased(MutexSecurity mutexSecurity)
```
andrei-epure-sonarsource, line 9: .NET Framework has some methods with `MutexSecurity`, hence this test

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            // Note that dispose does not release the resource
+
+            using (var m = new Mutex(true, "foo")) // FN, 'true' means it owns the mutex if no exception gets thrown
+            {
+                // do stuff
+            }
+
+            var m0 = new Mutex(initiallyOwned: true, "bar", out var m0WasCreated); // FN
+            m0.Dispose();
```
pavel-mikula-sonarsource, line 19: All these cases (except `m5`) are not FNs, because the lock is not released on at least one execution path.

RSPEC says:
>If a lock is acquired and released within a method, **then** it must be released along all execution paths of that method.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                // not released
+            }
+        }
+
+        public static void UnsupportedWaitAny(Mutex m1, Mutex m2, Mutex m3)
+        {
+            // it is too complex to support this scenario
+            WaitHandle[] handles = new[] { m1, m2, m3 };
+            var index = WaitHandle.WaitAny(handles);
```
pavel-mikula-sonarsource, line 59: Let's release him on at least one execution path to make it a FN (that will not be supported)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.cs
```diff
@@ -0,0 +1,140 @@
...
+            // it is too complex to support this scenario
+            WaitHandle[] handles = new[] { m1, m2, m3 };
+            var allHaveBeenAcquired = WaitHandle.WaitAll(handles);
+            if (allHaveBeenAcquired)
+            {
+                // all indexes should be released
+            }
+        }
+
+        public static void CompliantNotAcquired(Mutex paramMutex)
```
pavel-mikula-sonarsource, line 74: Let's add a case where we first release and then acquire.  Should lead to FP eventually

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        public static void UnsupportedWaitAll(Mutex m1, Mutex m2, Mutex m3)
+        {
+            // it is too complex to support this scenario
+            WaitHandle[] handles = new[] { m1, m2, m3 };
+            var allHaveBeenAcquired = WaitHandle.WaitAll(handles);
+            if (allHaveBeenAcquired)
+            {
+                // all indexes should be released
```
pavel-mikula-sonarsource, line 70: Let's conditionally release one of them to make it a FN

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.cs
```diff
@@ -0,0 +1,140 @@
...
+            if (m2.WaitOne(500))
+            {
+                m2.ReleaseMutex();
+            }
+
+            var isAcquired = paramMutex.WaitOne(400, false);
+            if (isAcquired)
+            {
+                paramMutex.ReleaseMutex();
+            }
```
pavel-mikula-sonarsource, line 105: Let's also add
```
            if (paramMutex.WaitOne(400, false))
            {
                paramMutex.ReleaseMutex();
            }
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                // do stuff but don't acquire
+            }
+        }
+
+        public static void CompliantAcquiredAndReleased(Mutex paramMutex)
+        {
+            var m1 = new Mutex(false);
+            m1.WaitOne();
+            m1.ReleaseMutex();
```
pavel-mikula-sonarsource, line 93: Let's do this with a static field and non static field

pavel-mikula-sonarsource: And then add `, NetMutexTest other)` parameter and use `other.instanceField`

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.cs
```diff
@@ -0,0 +1,140 @@
...
+                    acquired = true;
+                }
+            }
+            catch (UnauthorizedAccessException)
+            {
+                return;
+            }
+            finally
+            {
+                // there can be other exceptions when opening
```
pavel-mikula-sonarsource, line 127: I'd remove the comment. Even the `UnauthorizedAccessException` will make it to this `finally`.
```suggestion
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.cs
```diff
@@ -0,0 +1,140 @@
...
+                    if (acquired)
+                    {
+                        m.ReleaseMutex();
+                    }
+                    m.Dispose();
+                }
+            }
+        }
+    }
+}
```
pavel-mikula-sonarsource, line 140: Did you consider [MutexAcl.Create](https://docs.microsoft.com/en-us/dotnet/api/system.threading.mutexacl.create?view=dotnet-plat-ext-6.0)? It also has the `initiallyOwned` argument. I'm not sure if it's worth the trouble.

pavel-mikula-sonarsource (changes requested):

pavel-mikula-sonarsource:
Does this need to be a draft? It targets Čaba's branch.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/SymbolicExecution/LocksShouldBeReleasedTest.cs
```diff
@@ -51,6 +51,18 @@ public class LocksShouldBeReleasedTest
         public void LocksShouldBeReleased_ReaderWriterLockSlim_CS() =>
             Verifier.VerifyAnalyzer(@"TestCases\SymbolicExecution\Sonar\LocksShouldBeReleased.ReaderWriterLockSlim.cs", new LocksShouldBeReleased());
 
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void LocksShouldBeReleased_Mutex_CS() =>
+            Verifier.VerifyAnalyzer(@"TestCases\SymbolicExecution\Sonar\LocksShouldBeReleased.Mutex.cs", new LocksShouldBeReleased(), ParseOptionsHelper.FromCSharp8);
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: commodity to use named arguments before non-named arguments, for the visibility of the test

the APIs are the same, and we are testing various language features with multiple C# versions - in this case there's nothing particular in older C# versions to test

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.cs
andrei-epure-sonarsource: good catch...

analyzers/its/sources/Net6/Net6/MutexAclTest.cs
```diff
@@ -0,0 +1,41 @@
+﻿namespace Net6
```
andrei-epure-sonarsource, line 1: this will probably fail the PR, I'll send an update after the 2nd review (it's just a formality)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.NetFx.cs
```diff
@@ -0,0 +1,111 @@
+﻿using System;
+using System.Security.AccessControl;
+using System.Threading;
+
+// .NET Framework 4.8 APIs https://docs.microsoft.com/en-us/dotnet/api/system.threading.mutex.-ctor?view=netframework-4.8
+
+// In .NET 6, the constructors/methods with MutexSecurity and MutexRights are in the MutexAcl class inside the Microsoft.Windows.Compatibility NuGet.
```
andrei-epure-sonarsource, line 7: here is why I added an IT

csaba-sagi-sonarsource:
To make the review of the other PR as easy as possible, I suggest not to merge this PR until the other one is approved too.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.NetFx.cs
csaba-sagi-sonarsource: It would be nice to have the reason in the IT too. To make sure somebody does not remove it.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/SymbolicExecution/LocksShouldBeReleasedTest.cs
```diff
@@ -51,6 +51,18 @@ public class LocksShouldBeReleasedTest
         public void LocksShouldBeReleased_ReaderWriterLockSlim_CS() =>
             Verifier.VerifyAnalyzer(@"TestCases\SymbolicExecution\Sonar\LocksShouldBeReleased.ReaderWriterLockSlim.cs", new LocksShouldBeReleased());
 
+        [TestMethod]
+        [TestCategory("Rule")]
+        public void LocksShouldBeReleased_Mutex_CS() =>
```
csaba-sagi-sonarsource, line 56: The rule has been renamed from `LocksShouldBeReleased` to `LockReleasedAllPaths`. Please apply the changes.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/SymbolicExecution/LocksShouldBeReleasedTest.cs
```diff
@@ -51,6 +51,18 @@ public class LocksShouldBeReleasedTest
...
+            Verifier.VerifyAnalyzer(@"TestCases\SymbolicExecution\Sonar\LocksShouldBeReleased.Mutex.cs", new LocksShouldBeReleased());
```
pavel-mikula-sonarsource, line 57: This rule will only be implemented for Roslyn-based CFG. All files should be in the `Roslyn` subdirectory. For old rules (after few more MMFs) the `Sonar` directory will hold all test cases for the old engine (no C# 8 for example). `Roslyn` will hold test cases for the new engine (with different FPs/FNs, C# 8+ syntax)

cc @csaba-sagi-sonarsource as the issue comes from the base branch.
```suggestion
            Verifier.VerifyAnalyzer(@"TestCases\SymbolicExecution\Roslyn\LocksShouldBeReleased.Mutex.cs", new LocksShouldBeReleased());
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.cs
pavel-mikula-sonarsource: Yeah! That's properly ugly now 😅

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.cs
pavel-mikula-sonarsource: My intention with the 2nd comment here was to try to confuse the engine like this:
```
public void DifferentInstancesOnThis(Foo foo)
{
    foo.instanceMutex.WaitOne(); // Compliant
    instanceMutex.WaitOne();       // FN
    if (cond)
    {
        instanceMutex.ReleaseMutex();
    }
}

public void DifferentInstancesOnParameter(Foo foo)
{
    foo.instanceMutex.WaitOne(); // FN
    instanceMutex.WaitOne();       // Compliant
    if (cond)
    {
        foo.instanceMutex.ReleaseMutex();
    }
}
```

analyzers/tests/SonarAnalyzer.UnitTest/Rules/SymbolicExecution/LocksShouldBeReleasedTest.cs
csaba-sagi-sonarsource: I suggest not ti fix it in this PR. After it gets merged to the "feature" branch, I`ll fix it there.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.cs
```diff
@@ -0,0 +1,261 @@
...
+using System.Threading;
+
+// https://docs.microsoft.com/en-us/dotnet/standard/threading/mutexes
+internal class Foo
+{
+    bool cond;
+    public Mutex instanceMutex;
+    public static Mutex staticMutex;
+
+    // Note that dispose does not release the resource
```
pavel-mikula-sonarsource, line 11: It releases resource, but other one (`waitHandle`).
I'd also move this comment around line 38  where the dispose is happening
```suggestion
    // Note that dispose does not release the mutex
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Sonar/LocksShouldBeReleased.Mutex.cs
```diff
@@ -0,0 +1,261 @@
...
+            Foo.staticMutex.ReleaseMutex();
+        }
+        m0.Dispose();
+        m1.Dispose();
+        m2.Dispose();
+        m3.Dispose();
+
+        using (var mutexInUsing = new Mutex(true, "foo")) // FN
+        {
+            // 'true' means it owns the mutex if no exception gets thrown
```
pavel-mikula-sonarsource, line 45: I don't understand this comment as it looks like related to `cond` variable. Should it be on line 43, referring to the first argument?

pavel-mikula-sonarsource (changes requested):
Nice. Just few nitpicks left and well need to move the files (see comment above the review)

analyzers/tests/SonarAnalyzer.UnitTest/Rules/SymbolicExecution/LocksShouldBeReleasedTest.cs
pavel-mikula-sonarsource: OK

analyzers/tests/SonarAnalyzer.UnitTest/Rules/SymbolicExecution/LocksShouldBeReleasedTest.cs
pavel-mikula-sonarsource: This can be also fixed after the merge? Like the file locations?

csaba-sagi-sonarsource: I`ll take care of it after the merge

analyzers/tests/SonarAnalyzer.UnitTest/Rules/SymbolicExecution/LocksShouldBeReleasedTest.cs
andrei-epure-sonarsource: Thanks @csaba-sagi-sonarsource . Indeed, it will be easier to do that after merging on Csaba's branch, to avoid more rebase conflicts.

pavel-mikula-sonarsource (approved):
LGTM. Moving and renaming files will be fixed on the target branch later.

@andrei-epure-sonarsource will you update the failing ITs? The new comment shifted line numbers.

I'm also fine with merging it as-is and letting @csaba-sagi-sonarsource update it. As this review has already served its purpose. Target branch is not protected.

csaba-sagi-sonarsource (approved):
Please just fix the ITs and it is good to go. I`ll fix the code smell and move the files.

andrei-epure-sonarsource:
sorry @csaba-sagi-sonarsource @pavel-mikula-sonarsource for the delay

PR #5158, opened 2021-12-08
Bump StyleCop.Analyzers to 1.2.0-beta.354

pavel-mikula-sonarsource:
Using beta version to fix SA1000 FP on C# 9 target-typed new keyword `new()`

### DIFF
--- a/analyzers/StyleCopAnalyzers.targets
+++ b/analyzers/StyleCopAnalyzers.targets
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
   <ItemGroup>
-    <PackageReference Include="StyleCop.Analyzers" Version="1.1.118">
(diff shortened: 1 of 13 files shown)


costin-zaharia-sonarsource:
> Using beta version to fix SA1000 FP on C# 9 target-typed new keyword new()

I was thinking if we should wait for the proper release but then I've noticed that the latest non-beta release was https://github.com/DotNetAnalyzers/StyleCopAnalyzers/releases/tag/1.1.118 which was released on Apr 29, 2019.

costin-zaharia-sonarsource (approved):
LGTM!

pavel-mikula-sonarsource:
> but then I've noticed that the latest non-beta release was ... (long time ago)

Exactly :) I'm watching releases of that repo and will do an update if there's a proper release. I expect it any day now, just because I  created this PR 😆

PR #5159, opened 2021-12-08
Integrate SE with SymbolicExecutionRunner

pavel-mikula-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/RspecStrings.resx
```diff
@@ -1,17 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!--
-    Microsoft ResX Schema
-
+  <!-- 
+    Microsoft ResX Schema 
+
```
csaba-sagi-sonarsource, line 5: Where does all this whitespace come from? I guess it is unintentional.

analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
```diff
@@ -98,62 +111,83 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            if (checks.Any())
             {
-                var explodedGraph = new SonarExplodedGraph(cfg, symbol, context.SemanticModel, lva);
-                var analyzerContexts = enabledAnalyzers.Select(x => x.CreateContext(explodedGraph, context)).ToList();
                 try
                 {
-                    explodedGraph.ExplorationEnded += ExplorationEndedHandler;
-                    explodedGraph.Walk();
+                    var cfg = body.CreateCfg(nodeContext.SemanticModel);
+                    var engine = new RoslynSymbolicExecution(cfg, checks);
```
csaba-sagi-sonarsource, line 141: From a performance view it is understandable that we run all the checks from all the rules in a single instance of the Engine. On the other hand if we mess up something in one of the rules it will affect all the others. We might revisit this decision in the future if we see that  some rules have negative impact on others.

analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.RuleFactory.cs
```diff
@@ -0,0 +1,58 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarAnalyzer.Helpers;
+using SonarAnalyzer.SymbolicExecution.Roslyn;
+
+namespace SonarAnalyzer.Rules.SymbolicExecution
```
csaba-sagi-sonarsource, line 26: Maybe it should go straight to SonarAnalyzer.Common to be reused by the VB part?

csaba-sagi-sonarsource (changes requested):
Looks great, please take a look at the comments i left.

analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
pavel-mikula-sonarsource: Previously, each rule run its own engine instance and there was a major performance penalty for that. By running all rules on a single pass was a major improvement.

analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.RuleFactory.cs
pavel-mikula-sonarsource: I'll do that once we get to the VB rule. There should be more stuff extracted to the base class.

csaba-sagi-sonarsource (approved):
LGTM!

PR #5165, opened 2021-12-09
Fix S3603 Bug and FN: NRE for local method with qualified attributes

### LINKED ISSUE #5117: Fix S3603 Bug and FN: NRE for local method with qualified attributes
### Description

When having a local function, with an attribute, PureAttributeOnVoidMethod  has a NullReferenceException.
`CSC : warning AD0001: Analyzer 'SonarAnalyzer.Rules.CSharp.PureAttributeOnVoidMethod' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.`

### Repro steps

``` csharp
namespace ClassLibrary1
{
    public static class Class1
    {
        public static void Function1()
        {
            DoSomething();

            [System.Diagnostics.CodeAnalysis.SuppressMessage("category", "checkId")]
            static void DoSomething()
            {
                // nothing to see here
            }
        }
    }
}
```

### Known workarounds

Use `pragma warning disable` if this is a `SuppressMessage`.

### Related information

* SonarAnalyzer.CSharp: 8.32.0.39516
* Visual Studio 2022
* dotnet sdk: 6.0.100
* Windows 11

pavel-mikula-sonarsource:
Fixes #5117

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/PureAttributeOnVoidMethod.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/PureAttributeOnVoidMethod.cs
@@ -21,6 +21,7 @@
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Helpers;
@@ -41,7 +42,7 @@ namespace SonarAnalyzer.Rules.CSharp
                 c =>
                 {
                     if ((LocalFunctionStatementSyntaxWrapper)c.Node is var localFunction
-                        && localFunction.AttributeLists.SelectMany(x => x.Attributes).Any(attribute => IsPureAttribute(attribute.Name.GetIdentifier().Identifier.ValueText))
+                        && localFunction.AttributeLists.SelectMany(x => x.Attributes).Any(IsPureAttribute)
                         && InvalidPureDataAttributeUsage((IMethodSymbol)c.SemanticModel.GetDeclaredSymbol(c.Node)) is { } pureAttribute)
                     {
                         c.ReportIssue(Diagnostic.Create(Rule, pureAttribute.ApplicationSyntaxReference.GetSyntax().GetLocation()));
@@ -50,7 +51,8 @@ namespace SonarAnalyzer.Rules.CSharp
                 SyntaxKindEx.LocalFunctionStatement);
         }
 
-        private static bool IsPureAttribute(string attributeName) =>
-            attributeName.Equals("Pure") || attributeName.Equals("PureAttribute");
+        private static bool IsPureAttribute(AttributeSyntax attribute) =>
+            attribute.Name.GetIdentifier() is { } name
+            && (name.Identifier.ValueText == "Pure" || name.Identifier.ValueText == "PureAttribute");
     }
 }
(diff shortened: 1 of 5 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/PureAttributeOnVoidMethod.cs
```diff
@@ -50,7 +51,8 @@ protected override void Initialize(SonarAnalysisContext context)
                 SyntaxKindEx.LocalFunctionStatement);
         }
 
-        private static bool IsPureAttribute(string attributeName) =>
-            attributeName.Equals("Pure") || attributeName.Equals("PureAttribute");
+        private static bool IsPureAttribute(AttributeSyntax attribute) =>
+            attribute.Name.GetIdentifier() is { } name
+            && (name.Identifier.ValueText == "Pure" || name.Identifier.ValueText == "PureAttribute");
```
costin-zaharia-sonarsource, line 56: Don't we want to check here also the actual symbol? (Otherwise we will have false positives with user-defined attributes named `Pure`)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/ExpandingArchives.cs
```diff
@@ -35,6 +35,7 @@ public void ExtractEntry(ZipArchiveEntry entry)
             Stream stream;
 
             entry.ExtractToFile(""); // Noncompliant
+            entry?.ExtractToFile(""); // Noncompliant
```
costin-zaharia-sonarsource, line 38: Is this change intended?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/JwtSigned.cs
```diff
@@ -22,6 +22,7 @@ void DecodingWithDecoder(JwtDecoder decoder)
             string decoded;
             decoded = decoder.Decode(invalidToken, secret, true);
             decoded = decoder.Decode(invalidToken, secret, false); // Noncompliant {{Use only strong cipher algorithms when verifying the signature of this JWT.}}
+            decoded = decoder?.Decode(invalidToken, secret, false); // Noncompliant
```
costin-zaharia-sonarsource, line 25: Same comment also here.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/ExpandingArchives.cs
pavel-mikula-sonarsource: Yes, it's the [Bump Coverage](https://github.com/SonarSource/sonar-dotnet/pull/5165/commits/319b0e759ccedb5d3240b3366af5270ed3a242b4) commit. `MemberBindingExpressionSyntax` case was not covered in the handler.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/JwtSigned.cs
pavel-mikula-sonarsource: Same as above (just to have the same thing covered in two rules in case we delete one of them in the far future)

analyzers/src/SonarAnalyzer.CSharp/Rules/PureAttributeOnVoidMethod.cs
pavel-mikula-sonarsource: `InvalidPureDataAttributeUsage` does that inside (see `GetPureAttribute` in the base class)

pavel-mikula-sonarsource: I've added UT

costin-zaharia-sonarsource (approved):
LGTM!

PR #5166, opened 2021-12-09
Filter rule scope in SymbolicExecutionFactory

pavel-mikula-sonarsource:
It does not change the behavior for now, because all rules run only on the main code. 

But having at least one rule running on Test&Main or Test causes `SymbolicExecutionRunner` to contain at least one diagnostic descriptor that should execute so all rules are executed, causing FPs in terms of their scope.

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
@@ -57,7 +57,7 @@ namespace SonarAnalyzer.Rules.SymbolicExecution
         protected override void Initialize(SonarAnalysisContext context)
         {
             context.RegisterSyntaxNodeActionInNonGenerated(
-                c => Analyze<BaseMethodDeclarationSyntax>(c, x => (CSharpSyntaxNode)x.Body ?? x.ExpressionBody()),
+                c => Analyze<BaseMethodDeclarationSyntax>(context, c, x => (CSharpSyntaxNode)x.Body ?? x.ExpressionBody()),
                 SyntaxKind.ConstructorDeclaration,
                 SyntaxKind.DestructorDeclaration,
                 SyntaxKind.ConversionOperatorDeclaration,
@@ -65,11 +65,11 @@ namespace SonarAnalyzer.Rules.SymbolicExecution
                 SyntaxKind.MethodDeclaration);
 
             context.RegisterSyntaxNodeActionInNonGenerated(
-                c => Analyze<PropertyDeclarationSyntax>(c, x => x.ExpressionBody),
+                c => Analyze<PropertyDeclarationSyntax>(context, c, x => x.ExpressionBody),
                 SyntaxKind.PropertyDeclaration);
 
             context.RegisterSyntaxNodeActionInNonGenerated(
-                c => Analyze<AccessorDeclarationSyntax>(c, x => (CSharpSyntaxNode)x.Body ?? x.ExpressionBody()),
+                c => Analyze<AccessorDeclarationSyntax>(context, c, x => (CSharpSyntaxNode)x.Body ?? x.ExpressionBody()),
                 SyntaxKind.GetAccessorDeclaration,
                 SyntaxKind.SetAccessorDeclaration,
                 SyntaxKindEx.InitAccessorDeclaration,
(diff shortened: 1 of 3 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/SymbolicExecution/SymbolicExecutionRunner.cs
```diff
@@ -82,25 +82,27 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                Analyze(analysisContext, context, body, symbol);
             }
         }
 
-        private void Analyze(SyntaxNodeAnalysisContext context, CSharpSyntaxNode body, ISymbol symbol)
+        private void Analyze(SonarAnalysisContext analysisContext, SyntaxNodeAnalysisContext context, CSharpSyntaxNode body, ISymbol symbol)
         {
-            var enabledAnalyzers = symbolicExecutionAnalyzerFactory.GetEnabledAnalyzers(context);
+            var isTestProject = analysisContext.IsTestProject(context.Compilation, context.Options);
+            var isScannerRun = analysisContext.IsScannerRun(context.Options);
```
pavel-mikula-sonarsource, line 104: I need these two here for the future PR. That's why I don't pass `context` itself.

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Sonar/SymbolicExecutionAnalyzerFactoryTest.cs
```diff
@@ -88,13 +85,22 @@ public void GetEnabledAnalyzers_ReturnsEmptyList_WhenDiagnosticsAreDisabled()
...
-                    .SelectMany(analyzer => analyzer.SupportedDiagnostics.Select(descriptor => descriptor.Id))
-                    .ToList();
+            var analyzers = sut.GetEnabledAnalyzers(context, false, false).ToList();
+            var enabledAnalyzers = analyzers.SelectMany(analyzer => analyzer.SupportedDiagnostics.Select(descriptor => descriptor.Id)).ToList();
+
+            enabledAnalyzers.Should().BeEmpty();
+        }
+
+        [TestMethod]
+        public void GetEnabledAnalyzers_ReturnsEmptyList_OnTestProjectScannerRun()  // Old SE rules do not run on test code
```
csaba-sagi-sonarsource, line 95: The comment is a bit misleading, as there are rules that are run in standalone nuget package case.

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Sonar/SymbolicExecutionAnalyzerFactoryTest.cs
```diff
@@ -88,13 +85,22 @@ public void GetEnabledAnalyzers_ReturnsEmptyList_WhenDiagnosticsAreDisabled()
...
+        {
+            var sut = new SymbolicExecutionAnalyzerFactory();
+            var diagnostics = symbolicExecutionRuleIds.ToImmutableDictionary(v => v, _ => ReportDiagnostic.Warn);
+            var context = CreateSyntaxNodeAnalysisContext(diagnostics);
+            var analyzers = sut.GetEnabledAnalyzers(context, true, true).ToList();
+            var enabledAnalyzers = analyzers.SelectMany(analyzer => analyzer.SupportedDiagnostics.Select(descriptor => descriptor.Id)).ToList();
 
-            CollectionAssert.AreEquivalent(new List<string>(), enabledAnalyzers);
+            enabledAnalyzers.Should().BeEmpty();
         }
```
csaba-sagi-sonarsource, line 104: Please add another test like the one above with the value of false for the isScannerRun parameter. That will check that the "All" scoped rules are run in the Nuget case.

csaba-sagi-sonarsource (changes requested):
Please check comments.

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Sonar/SymbolicExecutionAnalyzerFactoryTest.cs
pavel-mikula-sonarsource: The entire `SymbolicExecutionAnalyzerFactory` and `SymbolicExecutionAnalyzerFactoryTest` will be deleted quite soon. All these various scenarios will be tested in #5159

csaba-sagi-sonarsource (approved):
LGTM!

PR #5170, opened 2021-12-10
Simplify DiagnosticVerifier

pavel-mikula-sonarsource:
This is a prerequisite for #5159

`DiagnosticVerifier.GetAllDiagnostics` does also some verification that should not be its responsibility. Making the API hard to user (and long to write)

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
@@ -69,7 +69,7 @@ namespace SonarAnalyzer.UnitTest.TestFramework
             SuppressionHandler.HookSuppression();
             try
             {
-                var diagnostics = GetDiagnostics(compilation, diagnosticAnalyzers, checkMode, sonarProjectConfigPath: sonarProjectConfigPath);
+                var diagnostics = GetAnalyzerDiagnostics(compilation, diagnosticAnalyzers, checkMode, sonarProjectConfigPath: sonarProjectConfigPath);
                 var expectedIssues = sources.Select(x => x.ToExpectedIssueLocations()).ToArray();
                 VerifyNoExceptionThrown(diagnostics);
                 CompareActualToExpected(compilation.LanguageVersionString(), diagnostics, expectedIssues, false);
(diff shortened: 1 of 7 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -87,70 +86,55 @@ internal static class DiagnosticVerifier
...
-        {
-            var ids = diagnosticAnalyzers
-                .SelectMany(analyzer => analyzer.SupportedDiagnostics)
-                .Select(diagnostic => diagnostic.Id)
-                .Distinct()
-                .ToHashSet();
-
-            return GetAllDiagnostics(compilation, diagnosticAnalyzers, checkMode, verifyNoExceptionIsThrown, sonarProjectConfigPath: sonarProjectConfigPath)
-                .Where(d => ids.Contains(d.Id));
+            return diagnostics.Where(x => ids.Contains(x.Id)).ToImmutableArray();
```
pavel-mikula-sonarsource, line 134: _Converted to draft:_ I need to validate if this is correct in one more scenario.

pavel-mikula-sonarsource: It's good. No changes needed.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/DiagnosticVerifier.cs
```diff
@@ -87,70 +86,55 @@ internal static class DiagnosticVerifier
...
-            bool verifyNoException = true,
-            CancellationToken? cancellationToken = null,
+        public static void VerifyNoIssueReported(
+                Compilation compilation,
+                DiagnosticAnalyzer diagnosticAnalyzer,
+                CompilationErrorBehavior checkMode = CompilationErrorBehavior.Default,
+                string sonarProjectConfigPath = null) =>
+            GetDiagnosticsNoExceptions(compilation, diagnosticAnalyzer, checkMode, sonarProjectConfigPath: sonarProjectConfigPath).Should().BeEmpty();
+
+        public static IEnumerable<Diagnostic> GetDiagnosticsNoExceptions(
```
costin-zaharia-sonarsource, line 96: I think it's nice that you split the old code into these 2 new methods. 

In my opinion, we should get the exception assertion outside of this class but as this code will be changed anyway it can also stay here for now. Up to you.

costin-zaharia-sonarsource (changes requested):

costin-zaharia-sonarsource (approved):
LGTM!

PR #5171, opened 2021-12-10
RSPEC files for #4207 (S6354)

andrei-epure-sonarsource:
cc @Corniel 

To generate this I had to update the rspec.ps1 script - see #5172

### DIFF
--- a/analyzers/rspec/cs/S6354_c#.html
+++ b/analyzers/rspec/cs/S6354_c#.html
@@ -1,5 +1,5 @@
 <p>One of the principles of a unit test is that it must have full control of the system under test. This is problematic when production code includes
-calls to static methods, and date/time functions are usually provided by system libraries as static methods.</p>
+calls to static methods, which cannot be changed or controlled. Date/time functions are usually provided by system libraries as static methods.</p>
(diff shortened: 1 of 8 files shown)


analyzers/rspec/cs/S6354_c#.html
```diff
@@ -0,0 +1,72 @@
+<p>One of the principles of a unit test is that it must have full control of the system under test. This is problematic when production code includes
+calls to static methods, and date/time functions are usually provided by system libraries as static methods.</p>
```
Corniel, line 2: which behavior can not be changed/controlled

Corniel (changes requested):
It has to be made explicit why statics functions might be an issue here.

pavel-mikula-sonarsource:
Is this PR needed? Changed files on a branch should do the trick.

Corniel:
> Is this PR needed? Changed files on a branch should do the trick.

It's both fine by me.

andrei-epure-sonarsource:
I've updated https://github.com/SonarSource/rspec/pull/285 and pushed the new version.
Indeed, this PR is not needed, the branch is fine.

andrei-epure-sonarsource:
(don't delete the branch please)

PR #5177, opened 2021-12-14
Implement ExitReached callback

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.CFG/Roslyn/BasicBlock.cs
```diff
@@ -127,6 +127,9 @@ private BasicBlock(object instance)
             });
         }
 
+        public bool ContainsThrow() =>
+            Successors.Any(successor => successor.Semantics == ControlFlowBranchSemantics.Throw);
```
pavel-mikula-sonarsource, line 131: _Nitpick:_ 
```suggestion
            Successors.Any(x => x.Semantics == ControlFlowBranchSemantics.Throw);
```

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/SymbolicCheck.cs
```diff
@@ -25,6 +25,9 @@ public class SymbolicCheck
         protected SymbolicCheck() { } // Avoid abstract class, fixes S1694
 
         public virtual ProgramState PreProcess(SymbolicContext context) => context.State;
+
```
pavel-mikula-sonarsource, line 28: I think we don't need these empty lines in this case. There will be no logic here so it's just a list of methodnames.

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
```diff
@@ -60,9 +60,20 @@ public void Execute()
...
             // ToDo: This is a temporary simplification until we support proper branching. This only continues to the next ordinal block
-            if (node.Block.Kind != BasicBlockKind.Exit)
+            if (node.Block.Kind == BasicBlockKind.Exit)
             {
-                yield return new ExplodedNode(cfg.Blocks[node.Block.Ordinal + 1], node.State);
+                NotifyExitReached(node.State);
+            }
+            else
+            {
+                if (node.Block.ContainsThrow())
```
pavel-mikula-sonarsource, line 69: Can be simple (at least for now)
```suggestion
            else if (node.Block.ContainsThrow())
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SymbolicExecution/SETestContext.cs
```diff
@@ -54,6 +54,13 @@ public void Main(bool boolParameter{additionalParameters})
             return new SETestContext(code, AnalyzerLanguage.CSharp, additionalChecks);
         }
 
+        public static SETestContext CreateCSMethod(string method, params SymbolicCheck[] additionalChecks) =>
+            new($@"
```
pavel-mikula-sonarsource, line 58: _Cleanup:_ The target typed-new can be used in more places in this file

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.cs
```diff
@@ -172,5 +172,83 @@ public void Execute_PersistSymbols_InsideBlock()
             static ISymbol LocalReferenceOperationSymbol(IOperationWrapperSonar operation) =>
                 ((ILocalReferenceOperation)operation.Instance).Local;
         }
+
```
pavel-mikula-sonarsource, line 175: Please move all these UTs to `RoslynSymbolicExecutionTest.Branching.cs` file

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.cs
```diff
@@ -172,5 +172,83 @@ public void Execute_PersistSymbols_InsideBlock()
...
+        [TestMethod]
+        public void ExitReached_SimpleFlow()
+        {
+            var exitCount = 0;
+            SETestContext.CreateCS("var a = true;", new ExitReachedTestCheck(state => CountExits(state, ref exitCount)));
```
pavel-mikula-sonarsource, line 180: This thing is used 5 times `new ExitReachedTestCheck(state => CountExits(state, ref exitCount)));` exactly the same way and it's complicated (double indirection, class, method, lambda, lot of copy/paste keystrokes).

Let's add this functionality to `CollectorTestCheck` 
```
        public override ProgramState ExitReached(ProgramState state) =>
            exitReachedCount++;
```

and assert it like this:
```
public void ExitReached_SimpleFlow() =>
    SETestContext.CreateCS("var a = true;").Collector.ValidateExitReachCount(1);
```

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.cs
```diff
@@ -172,5 +172,83 @@ public void Execute_PersistSymbols_InsideBlock()
...
+            Assert.AreEqual(1, exitCount);
```
pavel-mikula-sonarsource, line 181: Should shoudly.. for purpose of the `ValidateExitReachCount` internals
```suggestion
            exitCount.Should().Be(1);
```

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.cs
```diff
@@ -172,5 +172,83 @@ public void Execute_PersistSymbols_InsideBlock()
...
+        public void ExitReached_Throw()
+        {
+            const string method = @"
+public int Get(bool a)
+{
+    if (a)
+        throw new System.NullReferenceException();
+    else
+        return 2;
+}";
```
pavel-mikula-sonarsource, line 210: I'd like to document the fact that `return 2` block is currently not reached at all. The easiest way is probably to add this:
```
var returnNotReached = new PreProcessTestCheck(x =>
{
    x.Operation.Instance.Kind.Should().NotBe(OperationKind.Literal, "we don't support  multiple branches yet");
    return x.State;
});
```

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.cs
```diff
@@ -172,5 +172,83 @@ public void Execute_PersistSymbols_InsideBlock()
...
+            var exitCount = 0;
+            SETestContext.CreateCS("var a = true;", new ExitReachedTestCheck(state => CountExits(state, ref exitCount)));
+            Assert.AreEqual(1, exitCount);
+        }
+
+        [TestMethod]
+        public void ExitReached_MultipleBranches()
+        {
+            const string method = @"
+public int Get(bool a)
```
pavel-mikula-sonarsource, line 188: Same below. The `Get` is just too close to `get` from property. Same below
```suggestion
public int Main(bool a)
```
or
```suggestion
public int Method(bool a)
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SymbolicExecution/ProcessTestChecks.cs
```diff
@@ -42,4 +43,14 @@ internal class PostProcessTestCheck : SymbolicCheck
 
         public override ProgramState PostProcess(SymbolicContext context) => postProcess(context);
     }
+
+    internal class ExitReachedTestCheck : SymbolicCheck
```
pavel-mikula-sonarsource, line 47: Can be removed (see the test itself)

pavel-mikula-sonarsource (changes requested):
UTs can be simplified

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/SymbolicCheck.cs
```diff
@@ -25,6 +25,9 @@ public class SymbolicCheck
...
         public virtual ProgramState PostProcess(SymbolicContext context) => context.State;
+
+        public virtual ProgramState ExitReached(ProgramState state) => state;
```
pavel-mikula-sonarsource, line 31: We will need more than just the context.
We don't need to return a new state, because we don't expect any cross-check modifications there. And we don't use it.
```suggestion
        public virtual void ExitReached(SymbolicContext context) { }
```

pavel-mikula-sonarsource: If you can think of a scenario where `ProgramState` is needed, we can use this
```suggestion
        public virtual ProgramState ExitReached(SymbolicContext context) => context.State;
```
and then invoke it like this from the engine:
```
InvokeChecks(context, x => x.ExitReached)
```

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/SymbolicCheck.cs
costin-zaharia-sonarsource: These are still methods and I don't like exceptions :)

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.cs
costin-zaharia-sonarsource: I think it's a nice idea which will simplify things. What do you think about changing the collector to just gather data and letting the test to do the assertion?

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/SymbolicCheck.cs
costin-zaharia-sonarsource: Currently, the `SymbolicContext` has only the State and the Operation. When this part is reached the operation is always null. I prefer to keep this as simple as possible and extend it only when needed.

pavel-mikula-sonarsource: The context will pretty soon contain more information, especially about the flows. Please give it the `SymbolicContext context` to keep the pattern the same for all.

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.cs
pavel-mikula-sonarsource: I'd keep the pattern of having assertions inside the collector. We will soon need to collect and assert more stuff there (collect states, assert on them) and it will be easier to modify the internals of this single class then updating every single usage.
It also makes the code in UTs a bit shorter.

costin-zaharia-sonarsource: We will have a mix of responsibilities there, collection + validation. In my opinion, we should separate those.

costin-zaharia-sonarsource: As discussed offline, I've extracted the validate method and in the next PR I will rename the collector to validator.

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/SymbolicContext.cs
```diff
@@ -33,8 +33,9 @@ public class SymbolicContext
         public SymbolicContext(SymbolicValueCounter symbolicValueCounter, IOperationWrapperSonar operation, ProgramState state)
         {
             this.symbolicValueCounter = symbolicValueCounter ?? throw new ArgumentNullException(nameof(symbolicValueCounter));
-            Operation = operation ?? throw new ArgumentNullException(nameof(operation));
             State = state ?? throw new ArgumentNullException(nameof(state));
+
+            Operation = operation; // Operation can be null for the branch nodes.
```
pavel-mikula-sonarsource, line 38: Let's restore the order. It's just plain copying values so removing the `??` without the additional empty line was enough. The comment could be used instead of the `??` part or on the field itself.

To illustrate the ordering, the old version can be read as
```
a
b
c
a = a
b = b
c = c
```
while the new one is irregular - harder to read
```
a
b
c
a = a
c = c

b = b
```

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.Branching.cs
```diff
@@ -117,5 +115,70 @@ public void Branching_PersistSymbols_BetweenBlocks()
...
+    if (a)
+        throw new System.NullReferenceException();
+    else
+        return 2;
+}";
+            var returnNotReached = new PreProcessTestCheck(x =>
+                                                           {
+                                                               x.Operation.Instance.Kind.Should().NotBe(OperationKind.Literal, "we don't support multiple branches yet");
+                                                               return x.State;
+                                                           });
```
pavel-mikula-sonarsource, line 152: The `{...}` lambda block should be aligned on 12 spaces.  Same as `, c=> ` in rule registrations. The convention of far-aligning on lambdas is for cases like
```
xxxxxxxxxxxxxxx ( x=> longconditioooooooooooooooooooooooooooooon
                                   && longconditioooooooooooooooooooon)
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/SymbolicExecution/CollectorTestCheck.cs
```diff
@@ -33,6 +33,8 @@ internal class CollectorTestCheck : SymbolicCheck
         public readonly List<SymbolicContext> PostProcessed = new();
         private readonly List<(string Name, SymbolicContext Context)> tags = new();
 
+        private int exitReachedCount;
```
pavel-mikula-sonarsource, line 36: _Nitpick:_ the preceding empty line doesn't need to be there. It's a series of fields.

pavel-mikula-sonarsource (approved):
LGTM, there are some formatting nitpicks left.

PR #5181, opened 2021-12-15
Migrate null-checking SE rules to the SE Verifier overloads

pavel-mikula-sonarsource:
`Squash RunnerTesting` commit doesn't need a review.


csaba-sagi-sonarsource (approved):
LGTM!

PR #5182, opened 2021-12-15
Swap UT target frameworks

pavel-mikula-sonarsource:
Make net5.0 UTs first in the list for VS to run these UTs first.


costin-zaharia-sonarsource (approved):
LGTM!

PR #5183, opened 2021-12-15
Migrate security SE rules to the SE Verifier overloads

pavel-mikula-sonarsource:
Squash* commits doesn't need a review


analyzers/tests/SonarAnalyzer.UnitTest/Rules/SymbolicExecution/InitializationVectorShouldBeRandomTest.cs
```diff
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using Microsoft.CodeAnalysis;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarAnalyzer.Helpers;
```
csaba-sagi-sonarsource, line 23: Nitpick: Unnecessary using

csaba-sagi-sonarsource (approved):
LGTM! One small action to take before merging.

PR #5184, opened 2021-12-15
Migrate object state SE rules to the SE Verifier overloads

pavel-mikula-sonarsource:
Squash* commits don't need a review


csaba-sagi-sonarsource (approved):
LGTM!

PR #5189, opened 2021-12-16
Improve SE ProgramState operation storage

pavel-mikula-sonarsource:
Use the underlaying operation as a key, instead of the wrapper

Squash commit doesn't need a review


csaba-sagi-sonarsource (approved):
LGTM!

PR #5190, opened 2021-12-16
Fix a couple of code smells

csaba-sagi-sonarsource:
This PR was initially created to test if the external pipeline works, however after checking that the PR can be merged as it contains "useful" stuff.


costin-zaharia-sonarsource (approved):
LGTM!

PR #5192, opened 2021-12-16
Move BoolConstraint to common SE

pavel-mikula-sonarsource:
Squash commit doesn't need a review

This prepares namespaces for the future move of other constraints. It solves the `true ~ first` relation issue in UTs


csaba-sagi-sonarsource (approved):
LGTM!

PR #5202, opened 2021-12-17
Do not create default SymbolicValue for operations

pavel-mikula-sonarsource:

PR #5203, opened 2021-12-17
SE ConversionOperation support

csaba-sagi-sonarsource:


analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/OperationProcessors/Conversion.cs
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+using StyleCop.Analyzers.Lightup;
+
+namespace SonarAnalyzer.SymbolicExecution.Roslyn.OperationProcessors
+{
+    internal static class Conversion
+    {
+        public static ProgramState Process(SymbolicContext context, IConversionOperationWrapper conversion) =>
+            context.State.SetOperationValue(context.Operation, context.State[conversion.Operand]);
```
pavel-mikula-sonarsource, line 28: Let's check `context.State[conversion.Operand]` for null first and return `context.State` when it is.. With #5202 operations do not have it by default. And from memory consumption point of view, we don't need to create that null memory slot in the dictionary

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.Operations.cs
```diff
@@ -117,5 +117,19 @@ public void SimpleAssignment_ToUnsupported_FromLiteral(string snipet)
             collector.Validate("Literal: 42", x => x.State[x.Operation].HasConstraint(DummyConstraint.Dummy).Should().BeTrue("it's scaffolded"));
             collector.ValidateTag("Target", x => (x?.HasConstraint(DummyConstraint.Dummy) ?? false).Should().BeFalse());
         }
+
+        [TestMethod]
+        public void Conversion_ToLocalVariable_FromTrackedSymbol_ExplicitCast()
+        {
+            var collector = SETestContext.CreateCS(@"int a = 42; byte b = (byte)a; Tag(""b"", b);", new LiteralDummyTestCheck()).Validator;
+            collector.ValidateTag("b", x => x.HasConstraint(DummyConstraint.Dummy).Should().BeTrue());
```
pavel-mikula-sonarsource, line 125: Let's assert that there exists a conversion operator in the CFG. For example by `collector.ValidateOrder`.

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.Operations.cs
```diff
@@ -117,5 +117,19 @@ public void SimpleAssignment_ToUnsupported_FromLiteral(string snipet)
...
+        }
+
+        [TestMethod]
+        public void Conversion_ToLocalVariable_FromLiteral_ImplicitCast()
+        {
+            var collector = SETestContext.CreateCS(@"byte b = 42; Tag(""b"", b);", new LiteralDummyTestCheck()).Validator;
```
pavel-mikula-sonarsource, line 131: All `collector`s should be renamed to `validator`s

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.Operations.cs
```diff
@@ -117,5 +117,19 @@ public void SimpleAssignment_ToUnsupported_FromLiteral(string snipet)
...
+            collector.ValidateTag("b", x => x.HasConstraint(DummyConstraint.Dummy).Should().BeTrue());
```
pavel-mikula-sonarsource, line 132: Let's add the same assertion here as well

pavel-mikula-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/OperationProcessors/Conversion.cs
```diff
@@ -25,6 +25,8 @@ namespace SonarAnalyzer.SymbolicExecution.Roslyn.OperationProcessors
     internal static class Conversion
     {
         public static ProgramState Process(SymbolicContext context, IConversionOperationWrapper conversion) =>
-            context.State.SetOperationValue(context.Operation, context.State[conversion.Operand]);
+            context.State[conversion.Operand] is { } newState
```
pavel-mikula-sonarsource, line 28: It's not a `newState` but `value`

pavel-mikula-sonarsource (changes requested):
We need to rename the variable as it's confusing in this case

pavel-mikula-sonarsource (approved):
LGTM

PR #5206, opened 2021-12-20
Update VB.NET expected plugin size

pavel-mikula-sonarsource:
This also aligns the size with C# range to avoid VB.NET maintenance too often

https://github.com/SonarSource/sonar-dotnet/blob/26115eaece6f30560c2fa1362eddd0c46ac8f43d/sonar-csharp-plugin/pom.xml#L209-L212


costin-zaharia-sonarsource (approved):
LGTM!

PR #5211, opened 2021-12-21
Introduce DiagnosticsComparer

costin-zaharia-sonarsource:
Part of #5212


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/TestAnalysisComparer_2.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
```
pavel-mikula-sonarsource, line 1: Why is this actually `_2`? To prove that it works on multiple files? 
I see the `flows` here as a common topic. Can this be `TestAnalysisComparer.Flows.cs` instead? Or `TestAnalysisComparer.Secondary.cs` to play with all secondary locations?

The first one could be `TestAnalysisComparer.Primary.cs` then.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/TestAnalysisComparer_2.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Tests.Diagnostics
+{
+    public class SecondClass
+    {
+        async void WrongLength() { } // Noncompliant {{Return 'Task' instead.}}
+        //    ^^^^^
+
+        void Default()// Secondary [flow-0]
```
pavel-mikula-sonarsource, line 10: The first test case file explains each problem with a method name. Here I'm struggling with some. Can you rename them in a similar way as they are named in the first file?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/TestAnalysisComparer_2.cs
```diff
@@ -0,0 +1,29 @@
...
+                      // Secondary@-1 [flow-1] {{Wrong message}}
```
pavel-mikula-sonarsource, line 11: _More cases:_ Secondary can also have `^^^^^` markings on the secondary location (that can be correct or wrong)

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/TestAnalysisComparerTest.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace SonarAnalyzer.UnitTest.Helpers
+{
+    [TestClass]
+    public class TestAnalysisComparerTest
+    {
+        [TestMethod]
+        public void CompareShouldListAllDifferences()
+        {
+            const string expected = @"Comparing actual issues with expected:
+Language version: C# 10.
```
pavel-mikula-sonarsource, line 40: I'd format it like this to keep it optically more similar to reality:
```suggestion
            const string expected = 
@"Comparing actual issues with expected:
Language version: C# 10.
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/TestAnalysisComparerTest.cs
```diff
@@ -0,0 +1,74 @@
...
+
+TestAnalysisComparer_1.cs
+Line 8: unexpected primary issue S3168 with message 'Return 'Task' instead.'.
```
pavel-mikula-sonarsource, line 43: I'd expect upper case letters at the beginning of the sentence. English is also fine with `Primary` in the middle as it's more a name (to deal with the others in a simple manner).
```suggestion
Line 8: Unexpected Primary issue S3168 with message 'Return 'Task' instead.'.
```

pavel-mikula-sonarsource, line 43: Can be removed for simplicity
```suggestion
Line 8: unexpected primary issue S3168 'Return 'Task' instead.'.
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/TestAnalysisComparerTest.cs
```diff
@@ -0,0 +1,74 @@
...
+Line 10: primary issue S3168 message 'Wrong message' does not match the actual message 'Return 'Task' instead.'.
+Line 12: primary issue S3168 should start on column 19 but got column 14.
+Line 15: primary issue S3168 should have a length of 5 but got a length of 4.
+Line 3: expected primary issue (no id) was not raised!
```
pavel-mikula-sonarsource, line 47: We could do something like this in all messages. If the flow-id is specified.
```suggestion
Line 3: expected primary issue was not raised! ID: flow-42
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/TestAnalysisComparer_1.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        async void CorrectNoncompliant() { } // Noncompliant {{Return 'Task' instead.}}
+        //    ^^^^
+
+        async void Unexpected() { }
+
+        async void WrongMessage() { } // Noncompliant {{Wrong message}}
+
+        async void WrongPlace() { } // Noncompliant {{Return 'Task' instead.}}
+        //         ^^^^^^^^^^
```
pavel-mikula-sonarsource, line 13: What if it has a wrong position AND wrong length? Let's document it (whatever it does)

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/TestAnalysisComparerTest.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            var summary = TestAnalysisComparer.Compare(diagnostics, expectedIssues, "C# 10");
+
+            summary.ToString().Should().Be(expected);
+        }
+
+        private static Diagnostic[] GetDiagnostics(Compilation compilation) =>
+            DiagnosticVerifier.GetAnalyzerDiagnostics(compilation, new DiagnosticAnalyzer[] { new AsyncVoidMethod(), new MethodsShouldNotHaveIdenticalImplementations() }, CompilationErrorBehavior.Ignore).ToArray();
+
+        private static Compilation GetCompilation(IEnumerable<string> filePaths) =>
```
pavel-mikula-sonarsource, line 68: _Nitpick:_ Because it does quite complex creation of it.
```suggestion
        private static Compilation CreateCompilation(IEnumerable<string> filePaths) =>
```
or
```suggestion
        private static Compilation Compile(IEnumerable<string> filePaths) =>
```

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/TestAnalysisComparerTest.cs
```diff
@@ -0,0 +1,74 @@
...
+using SonarAnalyzer.Rules.CSharp;
+using SonarAnalyzer.UnitTest.TestFramework;
+
+namespace SonarAnalyzer.UnitTest.Helpers
+{
+    [TestClass]
+    public class TestAnalysisComparerTest
+    {
+        [TestMethod]
+        public void CompareShouldListAllDifferences()
```
pavel-mikula-sonarsource, line 37: Let's add (way smaller) VB version of it. Only with the basic stuff (single file, correct case, unexpected, wrong start, wrong length)

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/TestAnalysisComparerTest.cs
```diff
@@ -0,0 +1,74 @@
...
+            var diagnostics = GetDiagnostics(compilation);
+            var expectedIssues = GetExpectedIssues(compilation);
+
+            var summary = TestAnalysisComparer.Compare(diagnostics, expectedIssues, "C# 10");
+
+            summary.ToString().Should().Be(expected);
+        }
+
+        private static Diagnostic[] GetDiagnostics(Compilation compilation) =>
+            DiagnosticVerifier.GetAnalyzerDiagnostics(compilation, new DiagnosticAnalyzer[] { new AsyncVoidMethod(), new MethodsShouldNotHaveIdenticalImplementations() }, CompilationErrorBehavior.Ignore).ToArray();
```
pavel-mikula-sonarsource, line 66: Why `CompilationErrorBehavior.Ignore`? I'd prefer an error-free example. Or is it for the future when errors will be added?
```suggestion
            DiagnosticVerifier.GetAnalyzerDiagnostics(compilation, new DiagnosticAnalyzer[] { new AsyncVoidMethod(), new MethodsShouldNotHaveIdenticalImplementations() }).ToArray();
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+                Message = secondaryLocation.Message,
+                IssueId = null,
+                Start = secondaryLocation.Location.GetLineSpan().StartLinePosition.Character,
+                Length = secondaryLocation.Location.SourceSpan.Length
+            };
+
+        private static Dictionary<string, IList<IIssueLocation>> ToIssueLocations(this IEnumerable<Diagnostic> diagnostics)
+        {
+            var actualIssues = new Dictionary<string, IList<IIssueLocation>>();
+
```
pavel-mikula-sonarsource, line 109: This method can be without empty lines. There is no major unexpected stuff going on

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+            foreach (var diagnostic in diagnostics)
+            {
+                var path = diagnostic.Location.SourceTree?.FilePath ?? string.Empty;
+                if (!actualIssues.ContainsKey(path))
+                {
+                    actualIssues.Add(path, new List<IIssueLocation>());
+                }
+
```
pavel-mikula-sonarsource, line 117: This empy line is actually harmful because the preceding `if` and the next `[path] =` are trying to achieve a common goal.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+                actualIssues[path].Add(diagnostic.ToIssueLocation());
+
+                foreach (var secondaryLocation in diagnostic.AdditionalLocations.Select((_, i) => diagnostic.GetSecondaryLocation(i)))
+                {
+                    actualIssues[path].Add(secondaryLocation.ToIssueLocation());
+                }
```
pavel-mikula-sonarsource, line 123: The `.Select(_, i)` is actually doing just this in an obfuscated way. Let's simplify the old code to more streight-forward variant.
```suggestion
                for (var i = 0; i < diagnostic.AdditionalLocations.Count; i++)
                {
                    actualIssues[path].Add(diagnostic.GetSecondaryLocation(i).ToIssueLocation());
                }
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.UnitTest.TestFramework
+{
+    public static class TestAnalysisComparer
+    {
+        public static StringBuilder Compare(Diagnostic[] diagnostics, Dictionary<string, IList<IIssueLocation>> expectedIssuesPerFile, string languageVersion)
+        {
+            var summary = new StringBuilder($"Comparing actual issues with expected:\nLanguage version: {languageVersion}.\n");
+            var actualIssues = diagnostics.ToIssueLocations();
```
pavel-mikula-sonarsource, line 35: Just to highlight the dictionary content there. And align with the `expectedIssuesPerFile` parameter
```suggestion
            var actualIssuesPerFile = diagnostics.ToLocationsPerFile();
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            foreach (var fileName in actualIssues.Keys.OrderBy(x => x))
+            {
+                summary.Append('\n').Append(fileName).Append('\n');
```
pavel-mikula-sonarsource, line 39: Simplification
```suggestion
                summary.AppendLine().AppendLine(fileName);
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+
```
pavel-mikula-sonarsource, line 40: I'd be fine without empy lines in this method as well. Having `;\n\n` as a separator makes the language readability itself questionable (compared to `\n` from VB for example :trollface: )

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+                foreach (var actualIssue in actualIssues[fileName].OrderBy(x => x.LineNumber))
+                {
+                    var issue = expectedIssuesPerFile[fileName].FirstOrDefault(x => x.IsPrimary == actualIssue.IsPrimary && x.LineNumber == actualIssue.LineNumber);
```
pavel-mikula-sonarsource, line 43: We'll need more selective (priority-based) filtering here and add UTs for those. There could be a single line with:
* Two issues of same rule (no problem)
* Two issues of different rule IDs
* Two issues on different locations (and it should not matter which one we mark first in UTs)*
* Two issues, one on wrong location, one on correct location
* Two issues, both on wrong location


*This should work exactly the same way
```
  Something(); // Noncompliant {{Some message}}
                        // Noncompliant {{Another mesage}}
```
and
```
  Something(); // Noncompliant {{Another mesage}}
                        // Noncompliant {{Some message}}
```

pavel-mikula-sonarsource: Having something like this
```
Method(wrongLocation, alsoWrongLocation)
        ^^^^^^^^^^         ^^^^^^^^^^^^
```
should try to do it's best, somehow like this:
```
.OrderBy(x => Math.Abs(x.Start - actualIssue.Start))
.ThenBy(x => Math.Abs(x.Length - actualIssue.Length))
.FirstOrDefault()
```
to get the best possible approximate in case of issues (and will get the correct one in case of multiple correct issues)

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+                    summary.Append(Compare(actualIssue, issue));
```
pavel-mikula-sonarsource, line 44: To avoid the `new StringBuilder() inside. Most of the issues will not append anything
```suggestion
                    AppendComparison(summary, actualIssue, issue));
```
or an extension method (let me guess what you'll prefer 😄 )
```suggestion
                    summary.AppendComparison(actualIssue, issue));
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+                    expectedIssuesPerFile[fileName].Remove(issue);
```
pavel-mikula-sonarsource, line 45: Just to be explicit about the `expected` vs. `actual` everywhere around here.
```suggestion
                    expectedIssuesPerFile[fileName].Remove(expectedIssue);
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            var summary = new StringBuilder($"Comparing actual issues with expected:\nLanguage version: {languageVersion}.\n");
+            var actualIssues = diagnostics.ToIssueLocations();
+
+            foreach (var fileName in actualIssues.Keys.OrderBy(x => x))
+            {
+                summary.Append('\n').Append(fileName).Append('\n');
+
+                foreach (var actualIssue in actualIssues[fileName].OrderBy(x => x.LineNumber))
+                {
```
pavel-mikula-sonarsource, line 42: Line 42 (below): What if the key does not exist in the collection? Can that happen?

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+                    var issue = expectedIssuesPerFile[fileName].FirstOrDefault(x => x.IsPrimary == actualIssue.IsPrimary && x.LineNumber == actualIssue.LineNumber);
+                    summary.Append(Compare(actualIssue, issue));
+                    expectedIssuesPerFile[fileName].Remove(issue);
+                }
+
+                foreach (var expectedIssue in expectedIssuesPerFile[fileName].OrderBy(x => x.LineNumber))
```
pavel-mikula-sonarsource, line 48: `expectedIssuesPerFile[fileName]` repeats 3 times, it deserves a variable.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+                {
+                    summary.Append($"Line {expectedIssue.LineNumber}: expected {IssueType(expectedIssue.IsPrimary)} issue {expectedIssue.IssueId ?? "(no id)"} was not raised!\n");
```
pavel-mikula-sonarsource, line 50:
```suggestion
                    summary.AppendLine($"Line {expectedIssue.LineNumber}: expected {IssueType(expectedIssue.IsPrimary)} issue {expectedIssue.IssueId ?? "(no id)"} was not raised!");
```

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+                }
+            }
+
+            return summary;
+        }
+
+        private static StringBuilder Compare(IIssueLocation actual, IIssueLocation expected)
```
pavel-mikula-sonarsource, line 57: `AppendLine` all over the place to remove `\n` :)

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+                {
+                    actualIssues[path].Add(secondaryLocation.ToIssueLocation());
+                }
+            }
+
+            return actualIssues;
+        }
+
+        private static string IssueType(bool isPrimary) =>
+            isPrimary ? "primary" : "secondary";
```
pavel-mikula-sonarsource, line 130: When we'll add the `Error`, we might consider an enum for these. Maybe, if that works.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            if (expected == null)
+            {
+                summary.Append($"{prefix}: unexpected {issueType} issue {issueId} with message '{actual.Message}'.\n");
+            }
+            else if (expected.Message != null && actual.Message != expected.Message)
+            {
+                summary.Append($"{prefix}: {issueType} issue {issueId} message '{expected.Message}' does not match the actual message '{actual.Message}'.\n");
+            }
+            else if (expected.Start.HasValue && actual.Start != expected.Start)
```
pavel-mikula-sonarsource, line 72: The question is what to do if `actual.Start` doesn't have value.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+            {
+                summary.Append($"{prefix}: {issueType} issue {issueId} should start on column {expected.Start} but got column {actual.Start}.\n");
+            }
+            else if (expected.Length.HasValue && actual.Length != expected.Length)
```
pavel-mikula-sonarsource, line 76: What to do when `actual.Lenght` doesn't have a value?

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+            {
+                summary.Append($"{prefix}: {issueType} issue {issueId} should have a length of {expected.Length} but got a length of {actual.Length}.\n");
+            }
+
+            return summary;
+        }
+
+        private static IIssueLocation ToIssueLocation(this Diagnostic diagnostic) =>
+            new IssueLocationCollector.IssueLocation
```
pavel-mikula-sonarsource, line 85: Let's give `IssueCollection` two handy constructors that will simplify this syntax. 
```
new IssueLocation(diagnostics) : this(diagnostics.GetMessage, diagnostics.Location){
IsPrimary=true
IssueId = diagnostics.Id
}
```
that will invoke the second one that also suits the overload below `(message, location)`

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
```diff
@@ -0,0 +1,132 @@
...
+            {
+                IsPrimary = false,
+                LineNumber = secondaryLocation.Location.GetLineNumberToReport(),
+                Message = secondaryLocation.Message,
+                IssueId = null,
+                Start = secondaryLocation.Location.GetLineSpan().StartLinePosition.Character,
+                Length = secondaryLocation.Location.SourceSpan.Length
+            };
+
+        private static Dictionary<string, IList<IIssueLocation>> ToIssueLocations(this IEnumerable<Diagnostic> diagnostics)
```
pavel-mikula-sonarsource, line 106: I'd prefer this one to be before `ToIssueLocation` because it calls both of them as helpers

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
pavel-mikula-sonarsource: Actually, once you do this, we will not even need these noisy workarounds (methods) at all.

pavel-mikula-sonarsource (changes requested):
Mostly improvements of the old code and more test cases

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/TestAnalysisComparer_1.cs
```diff
@@ -0,0 +1,18 @@
...
+
+        async void WrongLength() { } // Noncompliant {{Return 'Task' instead.}}
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: It should not matter, but sure, I can add a test

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/TestAnalysisComparer_2.cs
costin-zaharia-sonarsource: There is a problem with how secondary locations ids are parsed in some of the cases so I will add this case in a following PR

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/TestAnalysisComparerTest.cs
costin-zaharia-sonarsource: Yes, I want this code to handle the errors too but for that, there is more refactoring needed.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
costin-zaharia-sonarsource: For me it's much more easy to read the code if I have some spaces around the `foreach` and the `if` branches.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
costin-zaharia-sonarsource: I don't agree. I think that this space nicely separates some logical blocks.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
costin-zaharia-sonarsource: I would like to have all the line endings set to `\n` in order to avoid future assertion problems when running tests on different operating systems. In this way, we can hardcode the result in the test as a single string.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
costin-zaharia-sonarsource: For me, the code is much easier to read if I have these empty lines. It really makes a big difference.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
pavel-mikula-sonarsource: It introduces a break in the middle of a single logical action `store issue in actualIssues[path]`

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
costin-zaharia-sonarsource: I try to do baby steps and not everything at once. This was not present in the previous implementation and I think we should add it in a different PR, when needed.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
costin-zaharia-sonarsource: I like more the current version since:
- the number of parameters is small
- it self contained and it does not alter external state
- performance should not be a concern here

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
costin-zaharia-sonarsource: No, it should not happen. If this happen, it should throw.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
costin-zaharia-sonarsource: Same discussion with end of line on different OSs.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
costin-zaharia-sonarsource: Same

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
costin-zaharia-sonarsource: Indeed, I'll have this in mind

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
costin-zaharia-sonarsource: I think the behavior is ok now: we will report a problem since we expect something else (the value is different)

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
costin-zaharia-sonarsource: Raise an issue since the value is different than what we expect

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
costin-zaharia-sonarsource: No longer needed.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
pavel-mikula-sonarsource: If we'll get to that point, we'll need some global solution/extension/conversion method for all cases. This AppendLine is way easier to read.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
pavel-mikula-sonarsource: It's +1 parameter from 2 to 3. Not really an issue.
Why not the extension method then? That one makes most sense.
We can do better in an easy way. There are a lot of diagnostics in our UTs.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/TestAnalysisComparer.cs
pavel-mikula-sonarsource: This is soooo much better 😆

pavel-mikula-sonarsource (approved):
LGTM

Issue #5186, opened 2021-12-16
Refactor Verifier to builder approach


pavel-mikula-sonarsource:
There are too many overloads in `Verifier` with too many arguments and duplications. We should refactor it to a builder approach instead. 

The exact approach and API shape will be discussed.

### Original proposal by @Corniel in #4971

The SonarAnalyzer TestFramework is awesome. It contains a lot of options to pick when verifying an analyzer.

However, given the enormous number of overloads makes it a bit clumsy. By providing a builder usage could improve:

``` C#
[TestMethod]
public void BreakOutsideSwitch() =>
    new BreakOutsideSwitch().AddSource(@"TestCases\BreakOutsideSwitch.cs").Verify();
```

And a more complex example:

``` C#
[TestMethod]
public void DeclareTypesInNamespaces_CS_Before8() =>
    new CS.DeclareTypesInNamespaces()
    .AddSource(@"TestCases\DeclareTypesInNamespaces.BeforeCSharp8.cs")
    .WithOptions(ParseOptionsHelper.BeforeCSharp8)
    .Verify();
```

Corniel:
I made a version that I myself for testing my own rules. That has no Sonar dependency, but therefor lacks some of the features you (probably still) need. Also, I handle secondary and primary issues a bit different I think.

What I also did, I always log all issues to the console, both the expected, missing, as extra. Made my live easier.

Anyway, feel inspired:
https://github.com/Corniel/roslyn-test-tools

PR #5213, opened 2021-12-21
Migrate more options to VerifierBuilder

pavel-mikula-sonarsource:
Contributes to #5186 

Add more options:
*  `.WithErrorBehavior`
*  `.WithOnlyDiagnostics`
*  `.WithOutputKind`
*  `.WithTopLevelSupport`
*  `.WithSonarProjectConfig`

This is a draft, rebase will be needed after #5208 is merged. The 1st squash commit doesn't need a review for the same reason.


Corniel:
Don't forget to add `.AddSnippet()`.

Also, C# and VB.NET have some features that do not apply to other languages (like allow unsafe code). It might be worth to have language specific builders (that derive from the same base), as I did [here](https://github.com/Corniel/roslyn-test-tool). I like the approach, including being immutable.

pavel-mikula-sonarsource:
@Corniel The current `OldVerifier` is not fully replaced yet, more PRs will come.

C# vs VB is currently solved in SolutionBuilder and doesn't pass through Verifier.

Corniel:
> C# vs VB is currently solved in SolutionBuilder and doesn't pass through Verifier.

THe difference should (indeed) not be in the verifier, but in the builder, that creates the verifier.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/OldVerifier.cs
```diff
@@ -20,6 +20,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
 using System.Text;
```
costin-zaharia-sonarsource, line 26: `using SonarAnalyzer.UnitTest.Helpers;` is no longer needed

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/OldVerifier.cs
```diff
@@ -301,68 +302,104 @@ public static void VerifyNoExceptionThrown(string path, DiagnosticAnalyzer[] dia
...
         public static void VerifyAnalyzer(string path,
                                           DiagnosticAnalyzer[] diagnosticAnalyzers,
-                                          IEnumerable<ParseOptions> options = null,
+                                          ImmutableArray<ParseOptions> options = default,
                                           CompilationErrorBehavior checkMode = CompilationErrorBehavior.Default,
                                           OutputKind outputKind = OutputKind.DynamicallyLinkedLibrary,
                                           IEnumerable<MetadataReference> additionalReferences = null,
                                           DiagnosticDescriptor[] onlyDiagnostics = null) =>
-            VerifyAnalyzer(new[] { path }, diagnosticAnalyzers, options, checkMode, outputKind, additionalReferences, null, onlyDiagnostics);
+            diagnosticAnalyzers.Aggregate(new VerifierBuilder(), (builder, analyzer) => builder.AddAnalyzer(() => analyzer))
```
costin-zaharia-sonarsource, line 313: That's a lot of builders created without a real need. I think we should either change `AddAnalyzer` to be `AddAnalyzers` and accept more analyzers, or add an overload.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/OldVerifier.cs
```diff
@@ -301,68 +302,104 @@ public static void VerifyNoExceptionThrown(string path, DiagnosticAnalyzer[] dia
...
+                .AddReferences(additionalReferences)
+                .AddPaths(RemoveTestCasesPrefix(path))
+                .WithOptions(options.IsDefault ? ImmutableArray<ParseOptions>.Empty : options)
+                .WithErrorBehavior(checkMode)
+                .WithOutputKind(outputKind)
+                .WithOnlyDiagnostics(onlyDiagnostics ?? Array.Empty<DiagnosticDescriptor>())
+                .Verify();
```
costin-zaharia-sonarsource, line 320: pedantic: The alignment is a bit off

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/OldVerifier.cs
```diff
@@ -301,68 +302,104 @@ public static void VerifyNoExceptionThrown(string path, DiagnosticAnalyzer[] dia
...
-                           null, CompilationErrorBehavior.Default,
-                           OutputKind.DynamicallyLinkedLibrary,
-                           additionalReferences,
-                           sonarProjectConfigPath);
+            new VerifierBuilder()
+                .AddAnalyzer(() => diagnosticAnalyzer)
+                .AddReferences(additionalReferences)
+                .AddPaths(RemoveTestCasesPrefix(path))
+                .WithSonarProjectConfigPath(sonarProjectConfigPath)
+                .Verify();
```
costin-zaharia-sonarsource, line 332: pedantic: We usually make the first method invocation on the same line and then align the rest of the calls with the first one.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Tests/IssueLocationCollectorTest.GetPreciseIssueLocations.cs
```diff
@@ -19,18 +19,14 @@
...
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
-namespace SonarAnalyzer.UnitTest.TestFramework.IssueLocationCollectorTests
+namespace SonarAnalyzer.UnitTest.TestFramework.Tests
 {
-    [TestClass]
-    public class IssueLocationCollector_GetPreciseIssueLocations
+    public partial class IssueLocationCollectorTest
     {
-        [TestMethod]
```
costin-zaharia-sonarsource, line 33: Did you remove this by mistake? If I put it back, the test is passing.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Tests/VerifierTest.cs
```diff
@@ -144,10 +147,57 @@ public void Verify_ParseOptions()
                 .WithMessage("CSharp5: Unexpected build error [CS8026]: Feature 'target-typed object creation' is not available in C# 5. Please use language version 9.0 or greater. on line 3");
         }
 
+        [TestMethod]
+        public void Verify_ErrorBehavior()
+        {
+            var builder = WithSnippet("undefined");
+            builder.Invoking(x => x.Verify()).Should().Throw<UnexpectedDiagnosticException>()
+                .WithMessage("CSharp7: Unexpected build error [CS0116]: A namespace cannot directly contain members such as fields, methods or statements on line 1");
```
costin-zaharia-sonarsource, line 155: pedantic: alignment if off

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/VerifierBuilder.cs
```diff
@@ -58,11 +68,34 @@ private VerifierBuilder(VerifierBuilder original)
             new(this) { Paths = Paths.Concat(paths).ToImmutableArray() };
 
         public VerifierBuilder AddReferences(IEnumerable<MetadataReference> references) =>
-            new(this) { References = References.Concat(references).ToImmutableArray() };
+            references != null && references.Any()
```
costin-zaharia-sonarsource, line 71: we evaluate `references` enumerable multiple times.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/VerifierBuilder.cs
```diff
@@ -58,11 +68,34 @@ private VerifierBuilder(VerifierBuilder original)
...
         public VerifierBuilder WithOptions(ImmutableArray<ParseOptions> parseOptions) =>
             new(this) { ParseOptions = parseOptions };
 
+        public VerifierBuilder WithOutputKind(OutputKind outputKind) =>
+            new(this) { OutputKind = outputKind };
+
+        public VerifierBuilder WithSonarProjectConfigPath(string sonarProjectConfigPath) =>
+            new(this) { SonarProjectConfigPath = sonarProjectConfigPath };
+
+        public VerifierBuilder WithTopLevelStatements() =>
```
costin-zaharia-sonarsource, line 96: I would remove this method:
- it's misleading (it should be ensure that the language versions contains C# 9 and no other previous version)
- can be called also for VB
- we can use directly `WithOutputKind(OutputKind.ConsoleApplication)` which is more clear about the behavior

analyzers/tests/SonarAnalyzer.UnitTest/ResourceTests/GeneratedResourcesTest.cs
```diff
@@ -1,124 +0,0 @@
...
-using Microsoft.VisualStudio.TestTools.UnitTesting;
-using SonarAnalyzer.Common;
-using SonarAnalyzer.Helpers;
-using SonarAnalyzer.Rules;
-using SonarAnalyzer.Utilities;
-
-namespace SonarAnalyzer.UnitTest.ResourceTests
-{
-    [TestClass]
-    public class GeneratedResourcesTest
```
costin-zaharia-sonarsource, line 37: Not sure I understand why this was removed. It does not seem related to this PR.

costin-zaharia-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj
```diff
@@ -67,25 +67,6 @@
     <None Remove="CBDEMocks\CBDEWaitAndSucceeds.exe" />
   </ItemGroup>
 
-  <ItemGroup>
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: _Short:_ It's not needed. 

_Middle:_ We don't need to use those resources in our UT project. Those autogenerated files are only meant to be TestCases.

_TL;DR;_ 
Check the [last commit](https://github.com/SonarSource/sonar-dotnet/pull/5213/commits/c3ba8ce0fd8e08e924f40da7fa2761a7de5dd396) in this PR above to see the isolated change.

`MarkAssemblyWithNeutralResourcesLanguageAttributeTest` used resx designer files as part of the UT. And while it only needed `just another cs file` that should be in `TestCases` directory, it was using `resx` file and its autogenerated designer file in `Resources` directory.

And as all test cases should be in `TestCases` directory, this irregularity poped up and revealed itself.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/OldVerifier.cs
pavel-mikula-sonarsource: This `Aggregate` hack should only exist in the `OldVerifier` so I would not add a special overload for it. And I plan to remove these in the 1st round of cleanup - there's a bunch of overloads used only a few times that I'd like to remove in another PR.
This specific one is used only once with 2 analyzers. So it's not very expensive.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/OldVerifier.cs
pavel-mikula-sonarsource: I'm probably not looking at the correct place and I don't see it. What is off exactly?

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/OldVerifier.cs
pavel-mikula-sonarsource: While I usually prefer less lines, I find this alignment easier to understand in this case. All modifications start on the same indent here.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Tests/IssueLocationCollectorTest.GetPreciseIssueLocations.cs
pavel-mikula-sonarsource: oops :)

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Tests/VerifierTest.cs
pavel-mikula-sonarsource: It is meant to be like this according to the `How we work` from 2021-12-22. It's not documented yet, I have an action item to document it.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/OldVerifier.cs
pavel-mikula-sonarsource: I now better see what you meant based on this comment:
https://github.com/SonarSource/sonar-dotnet/pull/5213#discussion_r774479132

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/OldVerifier.cs
pavel-mikula-sonarsource: I now better see what you meant based on this comment:
https://github.com/SonarSource/sonar-dotnet/pull/5213#discussion_r774479132

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/VerifierBuilder.cs
pavel-mikula-sonarsource: I gave it a try, but it's not very nice. Should I just remove the `Any` check instead?

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/VerifierBuilder.cs
pavel-mikula-sonarsource: The next PR #5214 adds `.WithConcurrentAnalysis(false);` that needs to be always called with this.
I'll improve the C# 9 part.
I'm fine with ignoring the VB design flaw :)

analyzers/tests/SonarAnalyzer.UnitTest/ResourceTests/GeneratedResourcesTest.cs
pavel-mikula-sonarsource: Doesn't seem intentional at all.

pavel-mikula-sonarsource: Ok, so I probably deleted it by accident when moving files out of `ResourcesTest` directory. 
The `ResourcesTest` directory is meant to be a data-only directory, see this
https://github.com/SonarSource/sonar-dotnet/blob/0cd6231a8756eaee401ba9dfcc5b7f4d37165e9d/analyzers/tests/SonarAnalyzer.UnitTest/SonarAnalyzer.UnitTest.csproj#L102-L104

So we should rename it to `TestResources` instead (to have it near `TestCases`) in another PR.

I've moved the file to another directory to decofuse its location.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Tests/VerifierTest.cs
costin-zaharia-sonarsource: My understanding was that we discussed "method declaration arguments" and not about this. I find aligned indentations a lot easier to read.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/VerifierBuilder.cs
costin-zaharia-sonarsource: Yes. Does it bring any value :) (pun intended)

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/VerifierBuilder.cs
costin-zaharia-sonarsource: I'm still not sure about the value of this method. We can easily live without it

costin-zaharia-sonarsource (commented):
Waiting for the removal of `Any` call and let's discuss directly the alignment and the `WithTopLevelStatements` method.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Tests/VerifierTest.cs
pavel-mikula-sonarsource: We've discussed both, see notes.
_Method declaration:_ We've decided not to change anything. Mary will investigate VS tools.
_Identation on chained invocation calls:_ This is the one that applies here and that I have the action item for to document it.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/VerifierBuilder.cs
pavel-mikula-sonarsource: We could live without it, and it will help us many times to save complicated copy/paste scaffolding that will always go together (considering the #5214 modification)

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/VerifierBuilder.cs
pavel-mikula-sonarsource: We can actually simplify it even more, as the null could come only from the `OldVerifier` anyway and should not happen under normal usage.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/OldVerifier.cs
```diff
@@ -485,7 +485,7 @@ static IEnumerable<TMessage> ReadProtobuf(string path)
         {
             var builder = SolutionBuilder.Create()
                 .AddProject(AnalyzerLanguage.FromPath(path), outputKind: outputKind)
-                .AddReferences(additionalReferences)
+                .AddReferences(additionalReferences ?? Enumerable.Empty<MetadataReference>())
```
costin-zaharia-sonarsource, line 488: I would add this directly in the method to avoid duplication in the calls.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/Tests/VerifierBuilderTest.cs
```diff
@@ -74,14 +74,6 @@ public void AddReferences_Concatenates_IsImmutable()
             two.References.Should().BeEquivalentTo(MetadataReferenceFacade.MsCorLib.Concat(MetadataReferenceFacade.SystemData)).And.HaveCount(2);
         }
 
-        [TestMethod]
```
costin-zaharia-sonarsource, line 72: And then we could leave these.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/VerifierBuilder.cs
```diff
@@ -68,9 +68,7 @@ private VerifierBuilder(VerifierBuilder original)
             new(this) { Paths = Paths.Concat(paths).ToImmutableArray() };
 
         public VerifierBuilder AddReferences(IEnumerable<MetadataReference> references) =>
-            references?.ToArray() is { Length: not 0 } referenceArray
-                ? new(this) { References = References.Concat(referenceArray).ToImmutableArray() }
-                : this;
+            new(this) { References = References.Concat(references).ToImmutableArray() };
```
costin-zaharia-sonarsource, line 71: nice, much shorter.

costin-zaharia-sonarsource (approved):
LGTM! Would be great to reduce the duplication (see my comments) before merging.

analyzers/tests/SonarAnalyzer.UnitTest/TestFramework/OldVerifier.cs
pavel-mikula-sonarsource: It's duplicate and `OldVerifier` is supposed to go away. I'd like to keep `OldVerifier`-specific pain only there to keep the `VerifierBuilder` clean.

PR #5218, opened 2021-12-24
Fix Symbolic Execution FPs when `ValidatedNotNullAttribute` is used in extension method

### LINKED ISSUE #2639: Fix S3900 FP: when object is validated with extension method
### Description

When we use method extension to check collection are not null, S3900 raise :( 

### Repro steps

Have an method extension like that
`public static bool IsNotNullOrEmpty<T>([ValidatedNotNull]this IEnumerable<T> src)
`
My code 
```
public static void Test(List<string> lst)
{
    if (lst.IsNotNullOrEmpty())
    {
        foreach (string item in lst) <-- S3900 raise on "lst"
        {
            // Do something
        }
    }
}

```
### Expected behavior

Don't raise rules

### Actual behavior

Rules raise

### Related information

* SonarLint 4.12.0.0
* Visual Studio 2017

### LINKED ISSUE #3850: S2259 False Positive using extension method
### Description

Appears to be related to #2639 

When creating a custom null validation method with a `[ValidatedNotNullAttribute]` marking the parameter in question, there appears to be a discrepancy based on how the method is called. If invoked directly, Sonar correctly recognizes that `value` cannot be null. However, a false positive is returned when the null-checking method is invoked as an extension method:

### Repro steps

```cs
namespace SonarS2259FalsePositive
{
    public static class ListExtensions
    {
        public static bool HasItems<T>([ValidatedNotNull] this List<T> list)
        {
            return !(list is null) && list.Count > 0;
        }
    }

    public static class ObjectExtensions
    {
        public static void AssertNotNull([ValidatedNotNull] this object obj)
        {
            if (obj is null)
            {
                throw new ArgumentNullException(nameof(obj));
            }
        }
    }

    public class SonarFalsePositive
    {
        public IEnumerable<int> HasItemsExtension(Wrapper wrapper)
        {
            List<int> items = wrapper.List?.ListOfInts;

            if (items.HasItems())
            {
                foreach (var t in items) // <-- this gives a false positive
                {
                    yield return t;
                }
            }
        }

        public IEnumerable<int> HasItemsDirect(Wrapper wrapper)
        {
            List<int> items = wrapper.List?.ListOfInts;

            if (ListExtensions.HasItems(items))
            {
                foreach (var t in items) // <-- this works as expected
                {
                    yield return t;
                }
            }
        }

        public IEnumerable<int> NotNullExtension(Wrapper wrapper)
        {
            List<int> items = wrapper.List?.ListOfInts;

            items.AssertNotNull();

            foreach (var t in items) // <-- this gives a false positive
            {
                yield return t;
            }
        }

        public IEnumerable<int> NotNullDirect(Wrapper wrapper)
        {
            List<int> items = wrapper.List?.ListOfInts;

            ObjectExtensions.AssertNotNull(items);

            foreach (var t in items) // <-- this works as expected
            {
                yield return t;
            }
        }
    }

    public class SomethingWithAListOfInts
    {
        public List<int> ListOfInts { get; set; }
    }

    public class Wrapper
    {
        public SomethingWithAListOfInts List { get; set; }
    }
}
```

### Expected behavior

Rule should not be violated when null-checking method is invoked as an extension method.

### Actual behavior

Rule is violated when null-checking method is invoked as an extension method.

### Known workarounds

Invoke null-checking method directly

### Related information

* SonarQube version
  * 7.9.4 (build 35981)
* C# Plugin Version
  * 8.6.1

andrei-epure-sonarsource:
Fixes #3850  
Fixes #2639

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Sonar/SonarExplodedGraph.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Sonar/SonarExplodedGraph.cs
@@ -1004,7 +1004,7 @@ namespace SonarAnalyzer.SymbolicExecution.Sonar
                         newProgramState = newProgramState.StoreSymbolicValue(symbol, sv);
                     }
                 }
-                else if (symbol is IMethodSymbol { IsExtensionMethod: true, Parameters: { IsEmpty: true } } methodSymbol
+                else if (symbol is IMethodSymbol { IsExtensionMethod: true } methodSymbol
                     && methodSymbol.GetFirstSyntaxRef() is MethodDeclarationSyntax syntax
                     && syntax.ParameterList.Parameters.FirstOrDefault().AttributeLists.SelectMany(list => list.Attributes).Any(x => x.ToStringContains("ValidatedNotNull")))
                 {
--- a/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Sonar/SonarExplodedGraph.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Sonar/SonarExplodedGraph.cs
@@ -1004,18 +1004,11 @@ namespace SonarAnalyzer.SymbolicExecution.Sonar
                         newProgramState = newProgramState.StoreSymbolicValue(symbol, sv);
                     }
                 }
-                else if (symbol is IMethodSymbol { IsExtensionMethod: true, Parameters: { IsEmpty: true } } methodSymbol)
+                else if (symbol is IMethodSymbol { IsExtensionMethod: true, Parameters: { IsEmpty: true } } methodSymbol
+                    && methodSymbol.GetFirstSyntaxRef() is MethodDeclarationSyntax syntax
+                    && syntax.ParameterList.Parameters.FirstOrDefault().AttributeLists.SelectMany(list => list.Attributes).Any(x => x.ToStringContains("ValidatedNotNull")))
                 {
-                    var declaration = methodSymbol.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax();
-                    if (declaration is MethodDeclarationSyntax syntax)
-                    {
-                        var x = syntax.ParameterList.Parameters.First();
-                        if (x.AttributeLists.SelectMany(list => list.Attributes).Any(x => x.ToStringContains("ValidatedNotNull")))
-                        {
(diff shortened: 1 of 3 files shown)


analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Sonar/SonarExplodedGraph.cs
```diff
@@ -1004,6 +1004,12 @@ private ProgramState VisitMemberAccess(MemberAccessExpressionSyntax memberAccess
                         newProgramState = newProgramState.StoreSymbolicValue(symbol, sv);
                     }
                 }
+                else if (symbol is IMethodSymbol { IsExtensionMethod: true } methodSymbol
+                    && methodSymbol.GetFirstSyntaxRef() is MethodDeclarationSyntax syntax
+                    && syntax.ParameterList.Parameters.FirstOrDefault().AttributeLists.SelectMany(list => list.Attributes).Any(x => x.ToStringContains("ValidatedNotNull")))
```
andrei-epure-sonarsource, line 1009: I'm not sure I have a better option than the `ToString`, unless I query the symbol for the attributes to verify the type, like here: https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Sonar/InvocationVisitor.cs#L243

I guess in practice there won't be that many extension methods with lots of attributes on the first parameter so it may be cleaner

csaba-sagi-sonarsource: I agree with this approach, we only look at attributes on extension methods, and there should not be that many in practice, just like you said. And this way we don`t have to obtain the symbol for the parameter.

csaba-sagi-sonarsource (approved):
LGTM!

analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Sonar/SonarExplodedGraph.cs
andrei-epure-sonarsource: I've just realized that `FirstOrDefault` is used there without a null check. 
In any case an extension method should always have the first `this param`, however then it should be `First()` instead of `FirstOrDefault()`

I've opened https://github.com/SonarSource/sonar-dotnet/pull/5233 for a defensive check

Issue #4465, opened 2021-05-25
Rule S1125: FP when using ternary operator and throw expressions


costin-zaharia-sonarsource:
### Description

Rule [S1125](https://rules.sonarsource.com/csharp/RSPEC-1125) is raising FPs when analyzing ternary operators.

### Repro steps

```
    public bool Method(bool condition)
    {
        return condition == false
            ? true
            : throw new Exception();
    }

    public bool Method2(bool condition)
    {
        return condition
            ? throw new Exception()
            : true;
    }
```

### Expected behavior

Either the message needs to be improved to suggest a refactoring (remove the == false and flip the condition) or don't raise the message at all.

For the second method the issue should not be raised at all.

### Actual behavior

Issues are raised for both `false` and `true` literals.

Additionally, the generated code by the code fix provider is not correct and it will not compile:
```
    public bool Method3(bool condition)
    {
        return !condition || throw new Exception();
    }

    public bool Method4(bool condition)
    {
        return !condition
               || throw new Exception()
            ; // Fixed
    }
```

StingyJack:
As there is nothing incorrect about the first one, it should not be flagging it either. 

Code isn't always automatically more readable in truthy conditional paths (2nd example). A lot of times it's going to be clearer using a falsy (1st example). Sonar shouldn't be trying to dictate one way over the other, or be forcing someone's particular code style preference when there isn't a specific problem it's trying to prevent.

While you are in there, please fix the other bug with S1125 where it is flagging a nonredundant false as redundant so we can actually activate this rule in our QP's.

pavel-mikula-sonarsource:
The `== false` is independent discussion happening on other places that should not be sneaked into this specific use case of ternary operator unrelated to `== false` part.

StingyJack:
@pavel-mikula-sonarsource   This reported defect wont be fixed unless it doesnt raise an issue for either true or false. You (sonar) cant determine what the more correct condition to be in the `if` or `else` path, so you shouldn't be coercing it.

I dont see how this can be fixed for both true and false without fixing `== false` as well. There arent any logical differences between these. The pattern is the same.

``` csharp
return  condition == false
            ? DoSomething()
            : throw new Exception();
 //    or 
if (condition == false) 
{
     return DoSomething();
}
throw new Exception();
```

costin-zaharia-sonarsource:
A possible solution is to suggest the removal of `== false` and flip the ternary branches. This will be easier to read and maintain.

[...]

PR #5220, opened 2021-12-28
S1125 FP with nullable booleans - Fix

### LINKED ISSUE #5219: S1125 FP with nullable booleans in ternary expressions C#9
### Description

C#9 allows to have target typed ternary expressions if both branches can be converted to the target type. In this case the rule raises a false positive.

### Repro steps

```
bool B { get; set; }

void M()
{
    bool? v = B ? true : null; // S1125 on this line
}
```

### Expected behavior

No issue is raised.

### Actual behavior

An issue is raised.

### Related information

* C#/VB.NET Plugins version: 8.33.0.40503

csaba-sagi-sonarsource:
Fixes #5219
Also solves the second case from #4465

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/BooleanLiteralUnnecessaryCodeFixProvider.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/BooleanLiteralUnnecessaryCodeFixProvider.cs
@@ -36,23 +36,17 @@ namespace SonarAnalyzer.Rules.CSharp
     public sealed class BooleanLiteralUnnecessaryCodeFixProvider : SonarCodeFixProvider
     {
         internal const string Title = "Remove the unnecessary Boolean literal(s)";
-        public override ImmutableArray<string> FixableDiagnosticIds
-        {
-            get
-            {
-                return ImmutableArray.Create(BooleanLiteralUnnecessary.DiagnosticId);
-            }
-        }
-        public override FixAllProvider GetFixAllProvider()
-        {
-            return WellKnownFixAllProviders.BatchFixer;
-        }
+        public override ImmutableArray<string> FixableDiagnosticIds =>
+            ImmutableArray.Create(BooleanLiteralUnnecessary.DiagnosticId);
+
+        public override FixAllProvider GetFixAllProvider() =>
+            WellKnownFixAllProviders.BatchFixer;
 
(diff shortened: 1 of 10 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/BooleanLiteralUnnecessaryCodeFixProvider.cs
```diff
@@ -203,58 +187,47 @@ private static SyntaxNode FindNodeToKeep(BinaryExpressionSyntax binary)
...
-        }
-
-        private static Document RemovePrefixUnary(Document document, SyntaxNode root,
-            SyntaxNode literal)
+        private static bool TwoSidesAreDifferentBooleans(BinaryExpressionSyntax binary) =>
+            (CSharpEquivalenceChecker.AreEquivalent(binary.Left, CSharpSyntaxHelper.TrueLiteralExpression)
+             && CSharpEquivalenceChecker.AreEquivalent(binary.Right, CSharpSyntaxHelper.FalseLiteralExpression))
+            || (CSharpEquivalenceChecker.AreEquivalent(binary.Left, CSharpSyntaxHelper.FalseLiteralExpression)
+               && CSharpEquivalenceChecker.AreEquivalent(binary.Right, CSharpSyntaxHelper.TrueLiteralExpression));
+        private static bool TwoSidesAreSameBooleans(BinaryExpressionSyntax binary) =>
```
andrei-epure-sonarsource, line 224: add a line above

analyzers/src/SonarAnalyzer.VisualBasic/Rules/BooleanLiteralUnnecessary.cs
```diff
@@ -74,7 +74,10 @@ private void CheckConditional(SyntaxNodeAnalysisContext context)
             var whenFalse = conditional.WhenFalse;
             var typeLeft = context.SemanticModel.GetTypeInfo(whenTrue).Type;
             var typeRight = context.SemanticModel.GetTypeInfo(whenFalse).Type;
-            if (typeLeft.IsNullableBoolean() || typeRight.IsNullableBoolean())
+            if (typeLeft.IsNullableBoolean()
```
andrei-epure-sonarsource, line 77: great!

andrei-epure-sonarsource (approved):
LGTM

PR #5223, opened 2021-12-29
S1542 FP when member is an override or interface implementation - Fix

### LINKED ISSUE #4945: S1542 FP: Don't raise for overridden members and interface implementations
### Description

S1542 Rename function 'xxx' to match the regular expression: 'xxx' should not raise for overriden members and interface implementations

### Repro steps

```VB.NET
Public MustInherit Class Base

    Public MustOverride Function rqSomething() As Object ' Noncompliant

End Class

Public Interface ISomething

    Function rqFirst() As Object      ' Noncompliant
    Function rqSecond() As Object     ' Noncompliant

End Interface

Public Class Sample
    Inherits Base
    Implements ISomething

    Public Overrides Function rqSomething() As Object   ' False Positive, can't change
        Throw New NotImplementedException()
    End Function

    Public Function rqFirst() As Object Implements ISomething.rqFirst   ' False Positive, should not raise
        Throw New NotImplementedException()
    End Function

    Public Function rqRenamed() As Object Implements ISomething.rqSecond ' Noncompliant, because existing name doesn't match the interface member
        Throw New NotImplementedException()
    End Function
End Class
```

csaba-sagi-sonarsource:
Fixes #4945


mary-georgiou-sonarsource (approved):
LGTM

Issue #5005, opened 2021-10-26
S927 False positive on discard


shuebner:
### Description

S927 will report a method argument named `_` as not conforming to the base declaration's method argument name:
![image](https://user-images.githubusercontent.com/1770684/138819047-5b270376-59d8-46a7-929f-6bccd1d39b93.png)

In case of discard names like `_`, we should assume a deliberate deviation and not report an inconsistency.

### Repro steps

```
interface IS927Repro
{
    int Calculate(int value1, int value2);
}

class S927Repro : IS927Repro
{
    public int Calculate(int _, int value2) => value2;
}
```

### Expected behavior

S927 does not trigger for arguments whose names consist of only underscores, like `_`, `__` etc..

### Actual behavior

S927 triggers for discard arguments just like for any other arguments.

### Known workarounds

Conform to the base declaration's names and make it less obvious that method arguments are not actually used by the particular implementation.

### Related information

* C# analyzer 8.30.0.37606
* Visual Studio version 16.11.5
* MSBuild / dotnet version 5.0.402
* Operating System: Windows 10

pavel-mikula-sonarsource:
I don't agree with this to be FP, because `_` is not a discard, but a parameter named `_` that can be used in the code with that name. 

Mainly because this cannot be compiled:
```
interface IS927Repro
{
    int Calculate(int value1, int value2, int value3);
}

class S927Repro : IS927Repro
{
    public int Calculate(int _, int _, int value3) => value3; // Error	CS0100 The parameter name '_' is a duplicate
}
```

csaba-sagi-sonarsource:
@pavel-mikula-sonarsource  I`ll try to take a look at it in the upcoming few days.

shuebner:
> I don't agree with this to be FP, because `_` is not a discard, but a parameter named `_` that can be used in the code with that name.

This is technically correct, of course. However, the name `_` was conventionally used even before C# had an explicit discard. Also, .NET's own style analyzers, such as S927's sister rule [IDE0060](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0060) consider names like `_`, `_1` etc. (not `__` though) special cases with attached semantics, namely: I do not care about this parameter.

shuebner:
No response? Do I really need to disable S927 now manually?

pavel-mikula-sonarsource:
Hi @shuebner ,

It will cause this confusing situation on your API and that is not a good practice:
[...]

PR #5224, opened 2021-12-29
Remove reproducer for S927 FP on discard - as we don't want to handle it

csaba-sagi-sonarsource:
Related to #5005

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ParameterNameMatchesOriginal.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ParameterNameMatchesOriginal.cs
@@ -205,15 +205,4 @@ namespace Tests.Diagnostics
     {
         public void Apply(SomeInterface<int> intValueParam) { }   //Noncompliant FN
     }
-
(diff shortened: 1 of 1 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/ParameterNameMatchesOriginal.cs
```diff
@@ -205,15 +205,4 @@ public class BasicImplementation : BaseInterface<int>
...
-
-    // The following two classes reproduce: https://github.com/SonarSource/sonar-dotnet/issues/5005
-    public abstract class BaseRepro
-    {
-        public abstract int SomeMethod(int SomeParam1, int SomeParam2);
-    }
-
-    public class Reproducer : BaseRepro
-    {
-        public override int SomeMethod(int _, int SomeParam2) => SomeParam2; // Noncompliant FP
```
pavel-mikula-sonarsource, line 217: What about keeping the test case and marking this as `// Noncompliant, it's not a discard`  to document the decision and prevent us from bumping it into it again in the future?

pavel-mikula-sonarsource (changes requested):

pavel-mikula-sonarsource (approved):
LGTM

PR #5225, opened 2021-12-29
S3442 suggests `private protected` instead of `protected` for internal ctors

### LINKED ISSUE #5006: S3442 should suggest `private protected` instead of `protected`
### Description

S3442 suggests to make an `internal` constructor of an abstract class `protected`:

image

When the abstract class only has internal constructors, it cannot be inherited from by other assemblies.
Following the suggestion changes semantics, since now other assemblies may derive from the class.
This may not be wanted.

I propose that S3442 suggest using `private protected` instead.
This way, the `internal` aspect is not lost.

### Repro steps

```
public abstract class S3442Repro
{
    internal S3442Repro()
    {
    }
}
```

### Expected behavior

S3442 suggests using `private protected` instead of `internal`.

### Actual behavior

S3442 suggests using `protected` instead of `internal`.

### Related information

* C# analyzer 8.30.0.37606
* Visual Studio version 16.11.5
* MSBuild / dotnet version 5.0.402

csaba-sagi-sonarsource:
Fixes #5006

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/AbstractTypesShouldNotHaveConstructors.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/AbstractTypesShouldNotHaveConstructors.cs
@@ -51,7 +51,7 @@ namespace SonarAnalyzer.Rules.CSharp
                         var invalidAccessModifier = GetModifiers(c.Node).FirstOrDefault(IsPublicOrInternal);
                         if (invalidAccessModifier != default)
                         {
-                            c.ReportIssue(Diagnostic.Create(Rule, invalidAccessModifier.GetLocation(), SuggestedModifier(invalidAccessModifier)));
+                            c.ReportIssue(Diagnostic.Create(Rule, invalidAccessModifier.GetLocation(), SuggestModifier(invalidAccessModifier)));
                         }
                     }
                 },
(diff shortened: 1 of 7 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/AbstractTypesShouldNotHaveConstructors.cs
```diff
@@ -51,21 +51,25 @@ public sealed class AbstractTypesShouldNotHaveConstructors : SonarDiagnosticAnal
...
+                SyntaxKind.ConstructorDeclaration);
 
         private static SyntaxTokenList GetModifiers(SyntaxNode node) =>
             node switch
             {
                 ClassDeclarationSyntax classDeclaration => classDeclaration.Modifiers,
                 ConstructorDeclarationSyntax ctorDeclaration => ctorDeclaration.Modifiers,
                 {} syntaxNode when RecordDeclarationSyntaxWrapper.IsInstance(syntaxNode) => ((RecordDeclarationSyntaxWrapper)syntaxNode).Modifiers,
-                _ => new SyntaxTokenList()
+                _ => default
```
andrei-epure-sonarsource, line 66: review note: this change is fine because `SyntaxTokenList` is a `struct`

analyzers/src/SonarAnalyzer.CSharp/Rules/AbstractTypesShouldNotHaveConstructors.cs
```diff
@@ -51,21 +51,25 @@ public sealed class AbstractTypesShouldNotHaveConstructors : SonarDiagnosticAnal
...
             };
 
         private static bool IsPublicOrInternal(SyntaxToken token) =>
             token.IsKind(SyntaxKind.PublicKeyword) || token.IsKind(SyntaxKind.InternalKeyword);
+
+        private static string SuggestedModifier(SyntaxToken token) =>
```
andrei-epure-sonarsource, line 72: should be a verb

```suggestion
        private static string SuggestModifier(SyntaxToken token) =>
```

andrei-epure-sonarsource (changes requested):
Nice!

please update the RSPEC as well

you can do it with 
```
.\scripts\rspec\rspec.ps1 -language cs -ruleKey S123 -rspecBranch "rule/xyz"
```

PR #5228, opened 2021-12-30
Fix S4261 FP: When using generic constraints

### LINKED ISSUE #4799: S4261: false diagnostic when using generic constraints

csaba-sagi-sonarsource:
Fixes #4799

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNamedAccordingToSynchronicity.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNamedAccordingToSynchronicity.cs
@@ -19,6 +19,7 @@
  */
 
 using System.Collections.Immutable;
+using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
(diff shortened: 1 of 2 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNamedAccordingToSynchronicity.cs
```diff
@@ -87,19 +88,17 @@ public sealed class MethodShouldBeNamedAccordingToSynchronicity : SonarDiagnosti
...
-        private static bool HasAsyncReturnType(IMethodSymbol methodSymbol)
-        {
-            var returnSymbol = (methodSymbol.ReturnType as INamedTypeSymbol)?.ConstructedFrom;
-            if (returnSymbol == null || returnSymbol.Is(KnownType.Void))
-            {
-                return false;
-            }
+        private static bool HasAsyncReturnType(IMethodSymbol methodSymbol) =>
+            methodSymbol.ReturnType is ITypeParameterSymbol typeParameter
+                ? typeParameter.ConstraintTypes.Any(x => IsAsyncType(x))
```
Corniel, line 93: What about `type` or `symbol` indead fo `x`?

Corniel (approved):

analyzers/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNamedAccordingToSynchronicity.cs
csaba-sagi-sonarsource: According to our [coding style](https://github.com/SonarSource/sonar-dotnet/blob/master/docs/coding-style.md#naming-conventions), we should use `x`.

Corniel: This is probably not the place to challenge the coding style guidelines. ;) But a short descriptive name for a variable could not harm, and could improve repeatability. That being said, the rest looks good to me.

mary-georgiou-sonarsource (approved):
LGTM

analyzers/src/SonarAnalyzer.CSharp/Rules/MethodShouldBeNamedAccordingToSynchronicity.cs
pavel-mikula-sonarsource: The `x` is an intentional convention based on lambda-x calculus. In simple cases, it helps to easily distinguish between the iterated variable vs. local fields/variables.

In this specific case, it could have been `Any(IsAsyncType)` instead.

PR #5229, opened 2021-12-30
Fix S3260 FP: Class inherited  by generic classes

### LINKED ISSUE #5160: Fix S3260 FP: Should not raise for inherited generic classes
### Description

S3260 should not raise in this case:

### Repro steps

```csharp
    public class Outer
    {
        private class Inner { } // Noncompliant FP

        private sealed class Inner<T> : Inner { }
    }
```

csaba-sagi-sonarsource:
Fixes #5160


mary-georgiou-sonarsource (approved):
LGTM

PR #5230, opened 2021-12-30
S3906 FP fix - do not raise when override or interface implementation

### LINKED ISSUE #3453: S3906 false positive when implementing interface from 3rd party
### Description
SonarLint reports [S3906](https://rules.sonarsource.com/csharp/RSPEC-3906) violation (Event Handlers should have the correct signature) when I implement an interface from an external library. The defined delegates even look ok (according to the rule). Even if they were non-compliant, I couldn't change them because they are defined in the referenced 3rd party library.

### Repro steps
Create a new project and install [WPFLocalizeExtension](https://github.com/XAMLMarkupExtensions/WPFLocalizationExtension) via nuget.

Create a new class that implements the [`ILocalizationProvider.cs`](https://github.com/XAMLMarkupExtensions/WPFLocalizationExtension/blob/master/src/Providers/ILocalizationProvider.cs) interface. Three eventhandlers are defined via delegates in [`ProviderEventArgs.cs`](https://github.com/XAMLMarkupExtensions/WPFLocalizationExtension/blob/master/src/Providers/ProviderEventArgs.cs) and SonarLint will warn about S3906 rule violations. I have tried the following code:

```cs
    public class LocalizationProvider : ILocalizationProvider
    {
        public event ProviderChangedEventHandler ProviderChanged;

        public event ProviderErrorEventHandler ProviderError;
        
        public event ValueChangedEventHandler ValueChanged
        {
            add
            {
                // not implemented
            }
            remove
            {
                // not implemented
            }
        }

        // other interface members not shown here
    }
```

### Expected behavior

No rule violation should appear.

### Actual behavior

SonarLint complains about S3906, even if the code looks correct to me (and I couldn't change it anyways).

### Known workarounds

I couldn't find any workaround (except for ignoring another rule).

### Related information

* C#/VB.NET Plugins Version: SonarLint 4.23.0.19399 (July 7th 2020)
* Visual Studio Version: 2019 v16.6.4

csaba-sagi-sonarsource:
Fixes #3453

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/DeclareEventHandlersCorrectly.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/DeclareEventHandlersCorrectly.cs
@@ -35,10 +35,11 @@ namespace SonarAnalyzer.Rules.CSharp
     {
         internal const string DiagnosticId = "S3906";
         private const string MessageFormat = "Change the signature of that event handler to match the specified signature.";
+        private const int EventArgumentPosition = 2;
 
-        private static readonly DiagnosticDescriptor rule =
+        private static readonly DiagnosticDescriptor Rule =
             DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
 
         protected override void Initialize(SonarAnalysisContext context)
         {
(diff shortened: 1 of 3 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/DeclareEventHandlersCorrectly.cs
```diff
@@ -35,51 +35,49 @@ public sealed class DeclareEventHandlersCorrectly : SonarDiagnosticAnalyzer
...
 
-            if (!IsCorrectEventHandlerSignature(methodSymbol))
+            var eventHandlerType = analysisContext.SemanticModel.GetSymbolInfo(typeSyntax).Symbol as INamedTypeSymbol;
+            if (eventHandlerType?.DelegateInvokeMethod is { } methodSymbol
+                && !IsCorrectEventHandlerSignature(methodSymbol))
             {
-                analysisContext.ReportIssue(Diagnostic.Create(rule, typeSyntax.GetLocation()));
+                analysisContext.ReportIssue(Diagnostic.Create(Rule, typeSyntax.GetLocation()));
             }
         }
```
Corniel, line 68: Please group:

``` C#
if (!eventSymbol.IsOverride 
    && eventSymbol.GetInterfaceMember() is null
    && analysisContext.SemanticModel.GetSymbolInfo(typeSyntax).Symbol is INamedTypeSymbol methodSymbol
	&& !IsCorrectEventHandlerSignature(methodSymbol))
{
	analysisContext.ReportIssue(Diagnostic.Create(rule, typeSyntax.GetLocation()));
}
```

analyzers/src/SonarAnalyzer.CSharp/Rules/DeclareEventHandlersCorrectly.cs
```diff
@@ -35,51 +35,49 @@ public sealed class DeclareEventHandlersCorrectly : SonarDiagnosticAnalyzer
...
-            return methodSymbol.ReturnsVoid &&
-                methodSymbol.Parameters.Length == 2 &&
-                methodSymbol.Parameters[0].Name == "sender" &&
-                methodSymbol.Parameters[0].Type.Is(KnownType.System_Object) &&
-                methodSymbol.Parameters[1].Name == "e" &&
-                IsDerivedFromEventArgs(methodSymbol.Parameters[1].Type);
-        }
+        private static bool IsCorrectEventHandlerSignature(IMethodSymbol methodSymbol) =>
+            methodSymbol.ReturnsVoid
+            && methodSymbol.Parameters.Length == EventArgumentPosition
```
Corniel, line 72: `0` and `1` are magic numbers here too. Replacing `2` only with a constant, do not make it easier to read.

analyzers/src/SonarAnalyzer.CSharp/Rules/DeclareEventHandlersCorrectly.cs
```diff
@@ -35,51 +35,48 @@ public sealed class DeclareEventHandlersCorrectly : SonarDiagnosticAnalyzer
...
                SyntaxKind.EventFieldDeclaration);
 
             context.RegisterSyntaxNodeActionInNonGenerated(
-               c => AnalyzeEventType(c, ((EventDeclarationSyntax)c.Node).Type),
+               c => AnalyzeEventType(c, ((EventDeclarationSyntax)c.Node).Type, c.ContainingSymbol),
                SyntaxKind.EventDeclaration);
         }
 
-        private void AnalyzeEventType(SyntaxNodeAnalysisContext analysisContext, TypeSyntax typeSyntax)
+        private static void AnalyzeEventType(SyntaxNodeAnalysisContext analysisContext, TypeSyntax typeSyntax, ISymbol eventSymbol)
```
andrei-epure-sonarsource, line 57: this is not the event symbol, but the enclosing type symbol
```suggestion
        private static void AnalyzeEventType(SyntaxNodeAnalysisContext analysisContext, TypeSyntax typeSyntax, ISymbol containingSymbol)
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/DeclareEventHandlersCorrectly.cs
```diff
@@ -53,11 +60,30 @@ public class ProviderChangedEventArgs : EventArgs {}
...
 
-    public class Repro3453 : ILocalizationProvider
+    public abstract class BaseClass
     {
-        public event ProviderChangedEventHandler ProviderChanged; // Noncompliant (See https://github.com/SonarSource/sonar-dotnet/issues/3453)
+        public abstract event ProviderChangedEventHandler ProviderChanged3; // Noncompliant
+        public abstract event ProviderChangedEventHandler ProviderChanged4; // Noncompliant
+    }
+
+    public class Repro3453 : BaseClass, ILocalizationProvider
```
andrei-epure-sonarsource, line 73: could you please add a similar test, but with `IDisposable` instead of `ILocalizationProvider` ? 

This means that we'll ignore the issue whenever the containing class implements an interface? Isn't this too lax?

Shouldn't we check if the interface contains some event signatures?

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/DeclareEventHandlersCorrectly.cs
```diff
@@ -35,51 +35,48 @@ public sealed class DeclareEventHandlersCorrectly : SonarDiagnosticAnalyzer
...
                SyntaxKind.EventFieldDeclaration);
 
             context.RegisterSyntaxNodeActionInNonGenerated(
-               c => AnalyzeEventType(c, ((EventDeclarationSyntax)c.Node).Type),
+               c => AnalyzeEventType(c, ((EventDeclarationSyntax)c.Node).Type, c.ContainingSymbol),
                SyntaxKind.EventDeclaration);
         }
 
-        private void AnalyzeEventType(SyntaxNodeAnalysisContext analysisContext, TypeSyntax typeSyntax)
+        private static void AnalyzeEventType(SyntaxNodeAnalysisContext analysisContext, TypeSyntax typeSyntax, ISymbol containingSymbol)
```
andrei-epure-sonarsource, line 57: ok, I was wrong, it seem that this actually _is_ the `eventSymbol`

the containing symbol is extracted inside `public static T GetInterfaceMember<T>(this T symbol)`

```suggestion
        private static void AnalyzeEventType(SyntaxNodeAnalysisContext analysisContext, TypeSyntax typeSyntax, ISymbol eventSymbol)
```

andrei-epure-sonarsource (approved):
please revert the name

LGTM

analyzers/src/SonarAnalyzer.CSharp/Rules/DeclareEventHandlersCorrectly.cs
csaba-sagi-sonarsource: It is the containingSymbol of the syntaxNode.

PR #5233, opened 2022-01-04
SE: Be defensive w/ extension method check

andrei-epure-sonarsource:
follow-up of #5218 - see https://github.com/SonarSource/sonar-dotnet/pull/5218#discussion_r777904674


csaba-sagi-sonarsource (approved):
LGTM!

PR #5234, opened 2022-01-04
S3459 FP fix: Property with expression body

### LINKED ISSUE #5232: S3459 - "Unassigned auto-property" is not actually an auto property (FP)
### Description

We've got a property in our code. It's not an auto-property. It does not have a setter and we do specify the getter.

But SonarQube thinks, it's an auto-property and an issue.

### Repro steps

```c#
private IMyInterfaceName MyInterfaceProp {
    get => TimeStampProvider.GetCurrentTimeStamp();
}
```

### Expected behavior

No warning at all or (alternatively) a better explanation why what we're doing is not a good idea.

We can't write the code like below because this does not call the GetCurrentTimeStamp on each call:

```c#
private IMyInterfaceName MyInterfaceProp { get; } = TimeStampProvider.GetCurrentTimeStamp();
```

We could maybe replace it with a method that returns `TimeStampProvider.GetCurrentTimeStamp();`, but is that really necessary? And if we should do that, this case should probably be explained in the description on SonarQube.

### Actual behavior

Warning S3459 with 2 titles:
- Removed unsassigned **auto-property** MyInterfaceProp, or set its value.
- Unassigned members should be removed.

### Known workarounds

Replacing the property with a method.

csaba-sagi-sonarsource:
Fixes #5232

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/NotAssignedPrivateMember.cs
@@ -133,7 +133,7 @@ namespace SonarAnalyzer.Rules.CSharp
         private static bool IsAutoPropertyWithNoInitializer(PropertyDeclarationSyntax declaration) =>
             declaration.Initializer == null
             && declaration.AccessorList != null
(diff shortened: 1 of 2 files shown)


mary-georgiou-sonarsource (approved):
LGTM

PR #5235, opened 2022-01-05
S3928 FP fix: Handle named parameters

### LINKED ISSUE #5226: S3928 "ArgumentException constructor arguments have been inverted" triggered with named parameters
### Description

Please provide a succinct description of your issue. Also please provide the rule ID (e.g. S1234)
Concerning: S3928
When using ArgumentException with named parameters in different order the S3928 still triggers

### Repro steps
Sample of case that triggers it:
```
using System;

namespace Sample;
public class Program
{
    public static void Main(string?[] args)
    {
        // The following line triggers S3928
        throw new ArgumentException(paramName: nameof(args), message: "Sample message");
    }
}
```

### Expected behavior

Because named parameters are used the rule S3928 is not expected to be triggered as the developer is aware and uses it to clarify.

### Actual behavior

Rule S3928 is triggered by SonarQube and SonarLint
The rule cannot be disabled as ArgumentExceptions that were made without named parameters still need to be detected

### Known workarounds

Switching the named parameters in the actual place of the method fixes it, but for hardcoded messages where no culture or other ways are needed or deep in depth resources references it because hard to read while only wanted to quickly see the parameter name.

### Related information

Related information
Sonarqube: Version 8.9.6 (build 50800) and higher
Visual Studio Version: VS 2022
SonarLint: SonarLint for Visualstudio 2022 v5.3.0.41207

csaba-sagi-sonarsource:
Fixes #5226

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
@@ -71,8 +71,8 @@ namespace SonarAnalyzer.Rules.CSharp
                 return;
             }
 
-            var parameterNameValue = new Optional<object>();
-            var messageValue = new Optional<object>();
+            var parameterNameValue = default(Optional<object>);
+            var messageValue = default(Optional<object>);
             for (var i = 0; i < methodSymbol.Parameters.Length; i++)
             {
                 var argumentExpression = objectCreation.ArgumentList.Arguments[i].Expression;
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
```diff
@@ -71,23 +72,20 @@ private static void CheckForIssue(SyntaxNodeAnalysisContext analysisContext)
...
-                else if (methodSymbol.Parameters[i].MetadataName == "message")
-                {
-                    messageValue = analysisContext.SemanticModel.GetConstantValue(argumentExpression);
-                }
-                else
-                {
-                    // do nothing
-                }
+                RetrieveParameterAndMessageArgumentValue(
+                    argument.NameColon != null
```
andrei-epure-sonarsource, line 82: IMO it would be more readable to extract the result of this expression into a nicely named variable

analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
```diff
@@ -71,23 +72,20 @@ private static void CheckForIssue(SyntaxNodeAnalysisContext analysisContext)
...
+                        ? argument.NameColon.Name.Identifier.ValueText
+                        : methodSymbol.Parameters[i].MetadataName,
+                    argumentExpression,
+                    analysisContext.SemanticModel,
+                    ref parameterNameValue,
+                    ref messageValue);
```
andrei-epure-sonarsource, line 88: the logic of these lines 75-88 seems cumbersome - at the end of the day, after this loop `parameterNameValue` and `messageValue` will have a single value 

wouldn't it be simpler to extract the logic into a method that would return a tuple with the results?

you could move the logic to `RetrieveParameterAndMessageArgumentValue`  and like this you can get rid of the `ref` arguments which are a code smell to me

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
```diff
@@ -72,52 +72,47 @@ private static void CheckForIssue(SyntaxNodeAnalysisContext analysisContext)
...
                 // can't check non-constant strings OR argument is not set
                 return;
             }
 
             var methodArgumentNames = GetMethodArgumentNames(objectCreation.Expression).ToHashSet();
-            if (!methodArgumentNames.Contains(TakeOnlyBeforeDot(parameterNameValue)))
+            if (!methodArgumentNames.Contains(TakeOnlyBeforeDot(parameterAndMessage.Item1)))
             {
-                var message = messageValue.HasValue && messageValue.Value != null && methodArgumentNames.Contains(TakeOnlyBeforeDot(messageValue))
+                var message = parameterAndMessage.Item2.HasValue && parameterAndMessage.Item2.Value != null && methodArgumentNames.Contains(TakeOnlyBeforeDot(parameterAndMessage.Item2))
```
andrei-epure-sonarsource, line 86: I don't understand the idea of an `Optional` that has a null value...

maybe you can remove the `Optional` boiler plate in this case? it's a private contract anyway

analyzers/src/SonarAnalyzer.CSharp/Rules/CheckArgumentException.cs
```diff
@@ -72,52 +72,47 @@ private static void CheckForIssue(SyntaxNodeAnalysisContext analysisContext)
...
                     ? ConstructorParametersInverted
-                    : string.Format(InvalidParameterName, parameterNameValue.Value);
+                    : string.Format(InvalidParameterName, parameterAndMessage.Item1.Value);
```
andrei-epure-sonarsource, line 88: I am confused here because from what I see, both `parameter` and `message` are used in constructing the message, right?

Also, `methodArgumentNames.Contains(TakeOnlyBeforeDot(parameterAndMessage.Item2))`  - we look for the `messageValue` inside the `methodArgumentsName` 

Actually, in the end the `messageArgument` (previously `messageValue` ) isn't used at all in the message


would it be possible to take a moment and find better names for this logic?

andrei-epure-sonarsource (changes requested):

andrei-epure-sonarsource (approved):
LGTM

PR #5237, opened 2022-01-05
Fix S4275 FP: Related to logical operator assignment and improved field selection.

### LINKED ISSUE #4339: Fix S4275 false positives
### Description

There are multiple FP raised by [S4275](https://rules.sonarsource.com/csharp/RSPEC-4275).
sonar-dotnet version: 8.22

### Repro steps

```
public class TestCases
    {
        bool pause;
        public bool Pause
        {
            get => pause;
            set => pause |= value; // Noncompliant - FP
        }

        private int textBufferUndoHistory;
        public int TextBufferUndoHistory
        {
            get // Noncompliant - FP
            {
                return textBufferUndoHistory = GetValue();
            }
        }
        private static int GetValue() => 1;

        private int[] attributes;
        public int[] Attributes
        {
            set { value.CopyTo(attributes, 0); } // Noncompliant - FP
        }

        private const string PREFIX = "pre";
        private string m_prefix;
        public string Prefix
        {
            get { return m_prefix; } // Noncompliant - FP
            set { m_prefix = value; } // Noncompliant - FP
        }
    }

    public class ContainsConstraint
    {
        private bool _ignoreCase;

        public ContainsConstraint IgnoreCase
        {
            // Refactor this getter so that it actually refers to the field '_ignoreCase'.
            // _ignoreCase has a different type than the property return type
            get // Noncompliant - FP
            {
                _ignoreCase = true;
                return this;
            }
        }
    }

    public class Base
    {
        protected long writeLockTimeout;
    }

    public class SubClass : Base
    {
        public static long WRITE_LOCK_TIMEOUT = 1000;
        public long WriteLockTimeout
        {
            get
            {
                return writeLockTimeout; // Noncompliant - FP
            }
            set
            {
                this.writeLockTimeout = value; // Noncompliant - FP
            }
        }
    }
```

mary-georgiou-sonarsource:
Fixes #4339


analyzers/src/SonarAnalyzer.CSharp/Helpers/CSharpSyntaxHelper.cs
```diff
@@ -333,9 +333,9 @@ expression switch
         public static bool IsLeftSideOfAssignment(this ExpressionSyntax expression)
         {
             var topParenthesizedExpression = expression.GetSelfOrTopParenthesizedExpression();
-            return topParenthesizedExpression.Parent.IsKind(SyntaxKind.SimpleAssignmentExpression) &&
-                topParenthesizedExpression.Parent is AssignmentExpressionSyntax assignment &&
-                assignment.Left == topParenthesizedExpression;
+            return topParenthesizedExpression.Parent.IsKind(SyntaxKind.SimpleAssignmentExpression)
+                    && topParenthesizedExpression.Parent is AssignmentExpressionSyntax assignment
```
csaba-sagi-sonarsource, line 337: NITPICK: the indentation should be aligned with the start of the topParenthesizedExpression in the line above. (Currently there is 1 space more before both && operators)

csaba-sagi-sonarsource (commented):
So far so good.  However, as this rule has both C# and VB implementation, some of the FP cases from C# could be also applicable for VBNet. I suggest to at least document them while you have the context (and maybe some of them will have the same solution for both languages so could be fixed easily). In the end if some of the VBNet cases are too difficult to solve, lets open an issue for them and have them documented.

mary-georgiou-sonarsource:
As discussed with @csaba-sagi-sonarsource, in this PR, we'll fix only the FPs for C# version of S4275. Afterward, we'll open a PR  adding the equivalent FP test cases in VB and if there are FP issues also raised there, we'll open an issue to fix them.

csaba-sagi-sonarsource (commented):
LGTM!

analyzers/src/SonarAnalyzer.Common/Rules/PropertiesAccessCorrectFieldBase.cs
```diff
@@ -251,6 +260,13 @@ public IFieldSymbol GetSingleMatchingFieldOrNull(IPropertySymbol propertySymbol)
 
             private static bool AreCanonicalNamesEqual(string name1, string name2) =>
                 name1.Equals(name2, StringComparison.OrdinalIgnoreCase);
+
+            private bool FieldMatchesTheProperty(IFieldSymbol field, IPropertySymbol property) =>
+                // We're not caching the property name as only expect to be called once per property
+                AreCanonicalNamesEqual(fieldToStandardNameMap[field], GetCanonicalFieldName(property.Name))
```
csaba-sagi-sonarsource, line 266: Nitpick: I would rename the GetCanonicalFieldName to GetCanonicalName as it does the same for property names too.

analyzers/src/SonarAnalyzer.Common/Rules/PropertiesAccessCorrectFieldBase.cs
```diff
@@ -251,6 +260,13 @@ public IFieldSymbol GetSingleMatchingFieldOrNull(IPropertySymbol propertySymbol)
...
+                && field.Type.Equals(property.Type)
+                && ((property.IsStatic && field.IsStatic) || (!property.IsStatic && !field.IsStatic))
+                && !field.IsConst;
```
csaba-sagi-sonarsource, line 269: It might be worth rearranging the order of the checks. For example it should be faster to check if the field is not a constant than to do the string comparisions and "Canonification" of the name.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/PropertiesAccessCorrectField.cs
```diff
@@ -768,34 +786,43 @@ public class ContainsConstraint
...
+        public static long TestStaticCase
+        {
+            get
+            {
+                return ALong; // Noncompliant
+            }
+            set
+            {
+                ALong = value; // Noncompliant
             }
```
csaba-sagi-sonarsource, line 824: I think you can improve the coverage with the case when the property is static but the field is not.

csaba-sagi-sonarsource (changes requested):
Looks good! If you want you can also align the coding style and clean up some low hanging issues in the base class.

analyzers/src/SonarAnalyzer.Common/Rules/PropertiesAccessCorrectFieldBase.cs
```diff
@@ -169,8 +181,11 @@ private IList<PropertyData> CollectPropertyData(IEnumerable<IPropertySymbol> pro
 
         private readonly struct PropertyData
         {
-            public PropertyData(IPropertySymbol propertySymbol, IEnumerable<FieldData> read, IEnumerable<FieldData> updated,
-                bool ignoreGetter, bool ignoreSetter)
+            public PropertyData(IPropertySymbol propertySymbol,
```
csaba-sagi-sonarsource, line 184: NITPICK: This could have been a one liner as it does not go over the limit of 200 characters

csaba-sagi-sonarsource (approved):
LGTM!

Issue #2752, opened 2019-10-25
S4487 False Positive


nnpcYvIVl:
SonarAnalyzer.CSharp version 8.0.0.9566 throws S4487 on `public uint cbSize;` in the following code. Is this a false positive?

```
using System.Runtime.InteropServices;

namespace S4487_False_Positive
{
    public static class Class
    {
        private struct LASTINPUTINFO
        {
            public uint cbSize;
            public uint dwTime;
        }

        public static ulong M()
        {
            LASTINPUTINFO lastinputinfo = new LASTINPUTINFO();
            lastinputinfo.cbSize = (uint)Marshal.SizeOf(lastinputinfo);
            GetLastInputInfo(ref lastinputinfo);

            return lastinputinfo.dwTime;
        }

        [DllImport("user32.dll")]
        private static extern bool GetLastInputInfo(ref LASTINPUTINFO plii);
    }
}
```

andrei-epure-sonarsource:
Hi @nnpcYvIVl , thanks for the repro.
From what I can see, you set `cbSize` inside `M()`, but don't read it. And the struct `LASTINPUTINFO` is `private`, so if you're not reading it in `Class`, it won't be read in a different place.

Am I missing something? Can `cbSize` be read inside `GetLastInputInfo` ?

Corniel:
Confirmed. The report is not as clear as you might hope for. The point is (it took me some time to understand it), is that because it passed to an external call, it becomes - de facto - a public contract. So you can not longer assume to know all use cases.

``` C#
public class S4487
{
    public ulong Reference(uint part1)
    {
        var reference = new S4487Ref();
        reference.part1 = part1;
        ExternalMethod(ref reference);
        return reference.part2;
    }

    private struct S4487Ref
    {
        public uint part1; // Compliant. Type is communicated an external call.
        public ulong part2;
    }

    [DllImport("user32.dll")]
    private static extern bool ExternalMethod(ref S4487Ref reference);
}
```

andrei-epure-sonarsource:
thanks for the confirmation, @Corniel

andrei-epure-sonarsource:
We should collect the types passed to a `DllImport` methods and before reporting an issue, exclude private members that get passed to external calls. This is quite an edge case and it's not straight forward to fix it without impacting the performance of the whole rule.

I'd vote to close this as won't fix.

[...]

PR #5239, opened 2022-01-05
add repro for S4487 FP

andrei-epure-sonarsource:
add repro for #2752

I've motivated in https://github.com/SonarSource/sonar-dotnet/issues/2752#issuecomment-1005668009 why we should close as won't fix, but if after https://jira.sonarsource.com/browse/MMF-2417 the rule will be re-written, we could fix the FP if very easy and not impacting perf for such an edge case


csaba-sagi-sonarsource (approved):
LGTM!

PR #5240, opened 2022-01-05
S4507: relax `env.IsDevelopment()` validator check

### LINKED ISSUE #5032: Reduce S4507 FP when `env.IsDevelopment();` is used
S4507 

Currently the following code will raise an issue:

```
   var isDevelopmentEnvironment = env.IsDevelopment();
   if (isDevelopmentEnvironment)
   {
        app.UseDeveloperExceptionPage(); // FP
    }
```

This is a minor security hotspot, so it's not worth investing in a complex dataflow capability.

We should reduce the level of FPs by checking if `env.IsDevelopment()` is invoked in the same method/class and not raise issues in that case.

Conversation: https://community.sonarsource.com/t/make-sure-this-debug-feature-is-deactivated-before-delivering-the-code-in-production/42336/4 (see internal as well)

andrei-epure-sonarsource:
S4507 

Fixes #5032 - only for C#

(opening draft to discuss first with AppSec if this really makes sense)


pierre-loup-tristant-sonarsource (approved):
I had a look at the UT and it looks good to me.
As expected the heuristic bring FNs in the same time that it fix FPs and I thinks its ok.

andrei-epure-sonarsource:
The uncovered condition is defensive coding.

csaba-sagi-sonarsource (approved):
LGTM!

PR #5241, opened 2022-01-05
S927 FP fix: Fix cases when generic type arguments are changed

### LINKED ISSUE #4370: S927 FP

csaba-sagi-sonarsource:
Fixes #4370

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Rules/ParameterNameMatchesOriginalBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/ParameterNameMatchesOriginalBase.cs
@@ -80,8 +80,11 @@ namespace SonarAnalyzer.Rules
             }
         }
 
-        private void VerifyGenericParameters(SyntaxNodeAnalysisContext context, TMethodDeclarationSyntax methodSyntax, IList<IParameterSymbol> actualParameters,
-            IList<IParameterSymbol> expectedParameters, string expectedLocation)
+        private void VerifyGenericParameters(SyntaxNodeAnalysisContext context,
+                                             TMethodDeclarationSyntax methodSyntax,
+                                             IList<IParameterSymbol> actualParameters,
+                                             IList<IParameterSymbol> expectedParameters,
+                                             string expectedLocation)
         {
             var parameters = ParameterIdentifiers(methodSyntax).ToList();
             for (var i = 0; i < parameters.Count; i++)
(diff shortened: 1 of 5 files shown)


analyzers/src/SonarAnalyzer.Common/Rules/ParameterNameMatchesOriginalBase.cs
```diff
@@ -80,20 +80,42 @@ private void VerifyParameters(SyntaxNodeAnalysisContext context, TMethodDeclarat
...
+                    && !AreGenericWithDifferentTypes(actualParameters[i].Type, expectedParameter.Type)
+                    && (expectedParameter.Type.Kind != SymbolKind.TypeParameter
+                        || actualParameters[i].Type.Kind == SymbolKind.TypeParameter))
                 {
                     context.ReportIssue(Diagnostic.Create(rule, parameter.GetLocation(), parameter.ValueText, expectedParameter.Name, expectedLocation));
                 }
             }
         }
+
+        private static bool AreGenericWithDifferentTypes(ITypeSymbol actualType, ITypeSymbol expectedType)
```
mary-georgiou-sonarsource, line 104: I would change this name to something like `AreGenericTypeParametersWithDifferentTypes`. To be a bit more clear what the method does.
WDYT?

analyzers/src/SonarAnalyzer.Common/Rules/ParameterNameMatchesOriginalBase.cs
```diff
@@ -80,20 +80,42 @@ private void VerifyParameters(SyntaxNodeAnalysisContext context, TMethodDeclarat
...
+        {
+            if (actualType is INamedTypeSymbol actualNamedType
+                && expectedType is INamedTypeSymbol expectedNamedType)
+            {
+                for (var i = 0; i < actualNamedType.TypeArguments.Count(); i++)
```
mary-georgiou-sonarsource, line 109: For readability maybe extract lines 109-114 in a different method?

mary-georgiou-sonarsource (changes requested):
Seems ok, a couple of comments.

mary-georgiou-sonarsource (approved):
LGTM

PR #5242, opened 2022-01-05
Fix S4035 FP

### LINKED ISSUE #5101: S4035 causes confusion and it looks like it gives us false positives.
### Description

This is about rule S4035.

We are using version 8.9.1 of SonarQube, but it does not look like you changed something related to this problem in the newer versions.

Rule S4035 is "Seal class 'AbstractBase' or implement 'IEqualityComparer<T>' instead".
If I click "Why is this an issue", it tells me "Classes implementing 'IEquatable<T> should be sealed".

But in our case, it does not make sense for multiple reasons.

### Repro steps

```c#
public abstract class AbstractBase{
    public override bool Equals(object obj){
        //calls the private Equals method in the end.
    }

    private bool Equals(AbstractBase other){
        // return true oder false
    }

    public override int GetHasCode(){
        // returns hash code
    }
}
```

```c#
public class Derived : AbstractBase{
    public override bool Equals(object obj){
        //calls the private Equals method in the end.
    }

    private bool Equals(Derived other){
        // return base.Equals(other) && .....
    }

    public override int GetHasCode(){
        // returns hash code
    }
}
```

### Expected behavior

No issues.

### Actual behavior

For both my classes, I get a S4035 warning.
Either it does not make sense or the descriptions of the issue does not explain it good enough.

1) Both classes do not implement the IEquatable<T> interface. Regardless, the S4035 title says "Classes implementing 'IEquatable<T> should be sealed". I do not implement it. So there should be no reason for this warning.

2) My first class is abstract. The warning: "Seal class 'AbstractBase' or implement 'IEqualityComparer<T>' instead". It is not even possible to seal an abstract class because it does not make sense.

3) I've seen [this issue](https://github.com/SonarSource/sonar-dotnet/issues/4000) that is similar to mine, but the argument of pavel does not seem to be true as my own Equals method is not protected. It's private.

### Related information

* MSBuild version: 16.11.0
* Operating System: Windows
* SonarScanner for MsBuild version: 5.3.2

andrei-epure-sonarsource:
Fixes #5101

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/EquatableClassShouldBeSealed.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/EquatableClassShouldBeSealed.cs
@@ -48,12 +48,12 @@ namespace SonarAnalyzer.Rules.CSharp
                     var classDeclaration = (ClassDeclarationSyntax)c.Node;
                     var classSymbol = c.SemanticModel.GetDeclaredSymbol(classDeclaration);
 
-                    if (classSymbol != null &&
-                        !classSymbol.IsSealed &&
-                        !classSymbol.IsStatic &&
-                        classSymbol.IsPubliclyAccessible() &&
-                        !classDeclaration.Identifier.IsMissing &&
-                        HasAnyInvalidIEquatableEqualsMethod(classSymbol))
+                    if (classSymbol != null
+                        && !classSymbol.IsSealed
+                        && !classSymbol.IsStatic
+                        && classSymbol.IsPubliclyAccessible()
+                        && !classDeclaration.Identifier.IsMissing
+                        && HasAnyInvalidIEquatableEqualsMethod(classSymbol))
                     {
                         c.ReportIssue(Diagnostic.Create(Rule, classDeclaration.Identifier.GetLocation(), classDeclaration.Identifier));
                     }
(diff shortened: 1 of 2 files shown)


andrei-epure-sonarsource:
uncovered conditions, from what I see, are about defensive coding

analyzers/src/SonarAnalyzer.CSharp/Rules/EquatableClassShouldBeSealed.cs
```diff
@@ -86,19 +84,17 @@ private static bool HasAnyInvalidIEquatableEqualsMethod(INamedTypeSymbol classSy
...
-        private static bool IsIEquatableEqualsMethodCandidate(IMethodSymbol methodSymbol)
-        {
-            return methodSymbol.MethodKind == MethodKind.Ordinary &&
-                methodSymbol.Name == EqualsMethodName &&
-                !methodSymbol.IsOverride &&
-                methodSymbol.ReturnType.Is(KnownType.System_Boolean) &&
-                methodSymbol.Parameters.Length == 1;
-        }
+        private static bool IsCompilableIEquatableTSymbol(INamedTypeSymbol namedTypeSymbol) =>
+            namedTypeSymbol.ConstructedFrom.Is(KnownType.System_IEquatable_T) &&
```
mary-georgiou-sonarsource, line 89: nitpick: According to our coding style the operators should be at the beginning of the line (same for the method below).

mary-georgiou-sonarsource (approved):
LGTM -  just a small comment on coding style.

PR #5243, opened 2022-01-05
Fix S1125 FP: ternary operator and throw expressions

### LINKED ISSUE #4465: Rule S1125: FP when using ternary operator and throw expressions

andrei-epure-sonarsource:
Fixes  #4465

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/BooleanLiteralUnnecessary.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/BooleanLiteralUnnecessary.cs
@@ -25,6 +25,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarAnalyzer.Common;
 using SonarAnalyzer.Extensions;
 using SonarAnalyzer.Helpers;
+using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.Rules.CSharp
 {
@@ -45,6 +46,10 @@ namespace SonarAnalyzer.Rules.CSharp
         protected override bool IsTrueLiteralKind(SyntaxNode syntaxNode) => syntaxNode.IsKind(SyntaxKind.TrueLiteralExpression);
 
         protected override bool IsFalseLiteralKind(SyntaxNode syntaxNode) => syntaxNode.IsKind(SyntaxKind.FalseLiteralExpression);
+
+        protected override bool IsInsideTernaryWithThrowExpression(BinaryExpressionSyntax syntaxNode) =>
+            syntaxNode.Parent is ConditionalExpressionSyntax conditionalExpression
+            && (IsThrowExpression(conditionalExpression.WhenTrue) || IsThrowExpression(conditionalExpression.WhenFalse));
 
         protected override void Initialize(SonarAnalysisContext context)
         {
@@ -86,6 +91,10 @@ namespace SonarAnalyzer.Rules.CSharp
             var conditional = (ConditionalExpressionSyntax)context.Node;
             var whenTrue = conditional.WhenTrue;
             var whenFalse = conditional.WhenFalse;
+            if (IsThrowExpression(whenTrue) || IsThrowExpression(whenFalse))
+            {
+                return;
+            }
             var typeLeft = context.SemanticModel.GetTypeInfo(whenTrue).Type;
             var typeRight = context.SemanticModel.GetTypeInfo(whenFalse).Type;
             if (typeLeft.IsNullableBoolean()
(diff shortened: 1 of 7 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/BooleanLiteralUnnecessary.cs
```diff
@@ -97,5 +106,8 @@ private void CheckConditional(SyntaxNodeAnalysisContext context)
             }
             CheckTernaryExpressionBranches(context, conditional.SyntaxTree, whenTrue, whenFalse);
         }
+
+        private static bool IsThrowExpression(ExpressionSyntax expressionSyntax) =>
+            ThrowExpressionSyntaxWrapper.IsInstance(expressionSyntax.RemoveParentheses());
```
andrei-epure-sonarsource, line 111: in practice I think `RemoveParentheses()` is overkill and wasting perf , so I'd remove it

analyzers/src/SonarAnalyzer.Common/Rules/BooleanLiteralUnnecessaryBase.cs
```diff
@@ -67,6 +68,10 @@ protected enum ErrorLocation
         protected void CheckAndExpression(SyntaxNodeAnalysisContext context)
         {
             var binaryExpression = (TBinaryExpression)context.Node;
+            if (IsInsideTernaryWithThrowExpression(binaryExpression))
```
csaba-sagi-sonarsource, line 71: This if can be merged with the one below.

analyzers/src/SonarAnalyzer.CSharp/Rules/BooleanLiteralUnnecessary.cs
csaba-sagi-sonarsource: In this case I agree to remove it, just document the case in the UTs to know that we have FNs.
In general I would say that  `RemoveParentheses()` is not always an overkill, specially in rules which handle expressions with logical operators.

analyzers/src/SonarAnalyzer.Common/Rules/BooleanLiteralUnnecessaryBase.cs
```diff
@@ -55,6 +55,7 @@ protected enum ErrorLocation
         protected abstract SyntaxToken GetOperatorToken(TBinaryExpression binaryExpression);
         protected abstract bool IsTrueLiteralKind(SyntaxNode syntaxNode);
         protected abstract bool IsFalseLiteralKind(SyntaxNode syntaxNode);
+        protected abstract bool IsInsideTernaryWithThrowExpression(TBinaryExpression syntaxNode);
```
csaba-sagi-sonarsource, line 58: For this I don't have a strong opinion so please feel free to ignore it if you don't agree with it: It might be a bit nicer if this would be a virtual method returning false by default so the VB part should not be forced to override a method that is specific to C#.

csaba-sagi-sonarsource (changes requested):
Overall looks nice, I just left a couple of minor comments.

analyzers/src/SonarAnalyzer.CSharp/Rules/BooleanLiteralUnnecessary.cs
andrei-epure-sonarsource: totally agree with you for the general usecase

analyzers/src/SonarAnalyzer.Common/Rules/BooleanLiteralUnnecessaryBase.cs
andrei-epure-sonarsource: Agree. It might break coverage because more conditions will need to be supported. But we don't need to please the tool.

csaba-sagi-sonarsource (approved):
LGTM!

PR #5247, opened 2022-01-07
S109 - reduce false positives for named arguments, attribute usage, "From...()" methods and single digit comparisons for well-known properties

### LINKED ISSUE #4737: S109 - reduce false positives
As an output of this community post: https://community.sonarsource.com/t/s109-magic-numbers-on-attributes/15784/16

And verifying the implementation of the same rule in our Java analyzer: https://github.com/SonarSource/sonar-java/blob/7.1.0.26670/java-checks/src/main/java/org/sonar/java/checks/MagicNumberCheck.java

We want to ignore magic numbers when used in:
- attributes (still to discuss whether we should consider just well-known values - like `Order`)
- in equality-checks or assignments to `Count` , `Size`, `Length`,  `Order` for single-digit integer values 
- enum values

Also consider (maybe have another internal talk) to tolerate:
- numbers passed to constructors 
- numbers passed to methods with named parameters

andrei-epure-sonarsource:
Fixes #4737

For validation - doc with [Peach numbers](https://docs.google.com/document/d/1RH6GUbNLxjFk3A-S9x1wpHwiUY68-NeuqyoU3wf4OcY/edit?usp=sharing) before this PR

Please review commit by commit. Most of the changed lines are in updated ITs - 8df471fdaa330cd20bed416efcd7f6e9b57dcca8


analyzers/src/SonarAnalyzer.CSharp/Rules/MagicNumberShouldNotBeUsed.cs
```diff
@@ -68,7 +72,28 @@ public sealed class MagicNumberShouldNotBeUsed : SonarDiagnosticAnalyzer
             // It's ok to use magic numbers in pragma directives
             || literalExpression.FirstAncestorOrSelf<PragmaWarningDirectiveTriviaSyntax>() != null
             // It's ok to use magic numbers in property declaration
```
andrei-epure-sonarsource, line 74: remove these comments, they don't have any added value

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
```diff
@@ -14,18 +34,31 @@ public static double FooProp
...
 
-        public ValidUseCases()
+        public ValidUseCases(string s, FooBar foo)
         {
             int i1 = 0;
             int i2 = -1;
             int i3 = 1;
             int i4 = 42;
+            var list = new List<string>(42); // Compliant, class constructor
+            var bigInteger = new IntPtr(1612342); // Compliant, class constructor
```
andrei-epure-sonarsource, line 44: add test with `Foo(new IntPtr(1612342))`

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
```diff
@@ -14,18 +34,31 @@ public static double FooProp
...
+            var g = new Guid(0xA, 0xB, 0xC, new Byte[] { 0, 1, 2, 3, 4, 5, 6, 7 }); // Compliant, class constructor
+            long[] values1 = { 30L, 40L }; // Compliant, array initialisation
+            int[] values2 = new int[] { 100, 200 }; // Compliant, array initialisation
```
andrei-epure-sonarsource, line 47: add test with `Foo(new int[] { 100, 200 })`

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
```diff
@@ -53,26 +116,53 @@ public class WrongUseCases
...
             {
 
             }
 
             var array = new string[10];
-            array[5] = "test"; // Noncompliant
+            array[5] = "test"; // Noncompliant {{Assign this magic number '5' to a well-named variable/constant, and use that instead.}}
+            Foo(new int[] { 100 }); // Noncompliant
+
+            new WrongUseCases(100, Foo(200, 300)); // Noncompliant {{Assign this object creation with magic number '100' to a well-named variable/constant, and use that instead.}}
```
andrei-epure-sonarsource, line 138: should tell users they can use named arguments to get rid of the warning

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
```diff
@@ -53,26 +116,53 @@ public class WrongUseCases
...
+            var result = list.Count == 99;
+            result = list.Count < 400; // Noncompliant
+            result = s.Length == 121; // Noncompliant
+            result = foo.Size == 472; // Noncompliant
+            result = foo.baz == 4; // Noncompliant
+            var x = 1;
+            result = x == 2; // Noncompliant
         }
 
+        [Foo(42, 43)] // Noncompliant
```
andrei-epure-sonarsource, line 156: should tell users they can use named arguments to get rid of the warnings

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
```diff
@@ -1,11 +1,32 @@
...
+        public int Size { get; set; }
+        public int COUNT { get; set; }
+        public int length { get; set; }
+        public int baz { get; set; }
+    }
+
     public class ValidUseCases
     {
         private const int MAGIC = 42;
+        private IntPtr Pointer = new IntPtr(19922); // Compliant, class constructor
```
andrei-epure-sonarsource, line 26:
```suggestion
        private IntPtr Pointer = new IntPtr(19922); // Compliant, stored in field
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
```diff
@@ -53,26 +116,53 @@ public class WrongUseCases
...
+            Foo(new IntPtr(123456)); // Noncompliant {{Assign this object creation with magic number '123456' to a well-named variable/constant, and use that instead.}}
+//              ^^^^^^^^^^^^^^^^^^
+            Foo(new int[] { 100, 200 }); // Noncompliant {{Assign this magic number '100' to a well-named variable/constant, and use that instead.}}
+            // Noncompliant@-1 {{Assign this magic number '200' to a well-named variable/constant, and use that instead.}}
+
+            var result = list.Count == 99;
+            result = list.Count < 400; // Noncompliant
+            result = s.Length == 121; // Noncompliant
+            result = foo.Size == 472; // Noncompliant
+            result = foo.baz == 4; // Noncompliant
```
andrei-epure-sonarsource, line 151: maybe for single digits should just ignore the rule when doing comparisons 

because there could be also cases like `HowMany() == 2` in code... not sure on this one

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
```diff
@@ -36,10 +86,23 @@ public override int GetHashCode()
             return MY_VALUE * 397;
```
andrei-epure-sonarsource, line 86: should also add `TimeSpan.FromMinutes` and generally ignore if it's with `TimeSpan.From...` because it's a common usecase

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
andrei-epure-sonarsource: https://github.com/SonarSource/sonar-dotnet/issues/5251

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
andrei-epure-sonarsource: https://github.com/SonarSource/sonar-dotnet/issues/5251

andrei-epure-sonarsource:
To avoid a rabbit-hole, I am splitting the discussion&implementation around:
- re-designing the rule
- fixing false negatives
- improving the messaging
in https://github.com/SonarSource/sonar-dotnet/issues/5251 .

This PR will only focus on reducing the FPs.

analyzers/src/SonarAnalyzer.CSharp/Rules/MagicNumberShouldNotBeUsed.cs
```diff
@@ -61,52 +61,49 @@ public sealed class MagicNumberShouldNotBeUsed : SonarDiagnosticAnalyzer
 
         private static bool IsExceptionToTheRule(LiteralExpressionSyntax literalExpression) =>
             NotConsideredAsMagicNumbers.Contains(literalExpression.Token.ValueText)
-            // It's ok to use magic numbers as part of a variable declaration
```
andrei-epure-sonarsource, line 60: the comments were useless, removing them

analyzers/src/SonarAnalyzer.CSharp/Rules/MagicNumberShouldNotBeUsed.cs
```diff
@@ -61,52 +61,49 @@ public sealed class MagicNumberShouldNotBeUsed : SonarDiagnosticAnalyzer
...
         private static bool IsSingleOrNamedAttributeArgument(LiteralExpressionSyntax literalExpression) =>
             literalExpression.Parent is AttributeArgumentSyntax arg
             && (arg.NameColon is not null
                 || arg.NameEquals is not null
                 || (arg.Parent is AttributeArgumentListSyntax argList && argList.Arguments.Count == 1));
 
-        // Inside property we consider magic numbers as exceptions in the following cases:
-        //   - A {get; set;} = MAGIC_NUMBER
-        //   - A { get { return MAGIC_NUMBER; } }
-        private static bool IsInsideProperty(SyntaxNode node)
```
andrei-epure-sonarsource, line 76: this was just reordered above, no functional change

analyzers/src/SonarAnalyzer.CSharp/Rules/MagicNumberShouldNotBeUsed.cs
```diff
@@ -61,52 +61,49 @@ public sealed class MagicNumberShouldNotBeUsed : SonarDiagnosticAnalyzer
...
+            {
+                return false;
+            }
+            var parent = node.Parent;
+            return parent is ReturnStatementSyntax || parent is EqualsValueClauseSyntax;
+        }
 
         private static bool IsSingleDigitInToleratedComparisons(LiteralExpressionSyntax literalExpression) =>
             literalExpression.Parent is BinaryExpressionSyntax binaryExpression
-            && binaryExpression.IsAnyKind(SyntaxKind.EqualsExpression, SyntaxKind.LessThanExpression, SyntaxKind.LessThanOrEqualExpression)
```
andrei-epure-sonarsource: this was fixed in bc1eebc

analyzers/src/SonarAnalyzer.CSharp/Rules/MagicNumberShouldNotBeUsed.cs
```diff
@@ -61,52 +61,49 @@ public sealed class MagicNumberShouldNotBeUsed : SonarDiagnosticAnalyzer
...
-            || IsNamedArgument(literalExpression)
-            || IsSingleOrNamedAttributeArgument(literalExpression)
-            || IsSingleDigitInToleratedComparisons(literalExpression);
+            || IsSingleDigitInToleratedComparisons(literalExpression)
+            || IsToleratedArgument(literalExpression);
+
+        // Inside property we consider magic numbers as exceptions in the following cases:
+        //   - A {get; set;} = MAGIC_NUMBER
+        //   - A { get { return MAGIC_NUMBER; } }
+        private static bool IsInsideProperty(SyntaxNode node)
```
andrei-epure-sonarsource, line 86: this was just reordered from below, no functional change (see in overall changes this part didn't change)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/MagicNumberShouldNotBeUsed.cs
mary-georgiou-sonarsource: I agree. Collection initialization should be Noncompliant unless in UTs.

analyzers/src/SonarAnalyzer.CSharp/Rules/MagicNumberShouldNotBeUsed.cs
```diff
@@ -96,5 +107,11 @@ private static bool IsInsideProperty(SyntaxNode node)
             var parent = node.Parent;
             return parent is ReturnStatementSyntax || parent is EqualsValueClauseSyntax;
         }
+
+        private static bool ToStringContainsAnyAcceptedNames(SyntaxNode syntaxNode)
+        {
+            var toString = syntaxNode.ToString().ToLower();
```
mary-georgiou-sonarsource, line 126: My understanding is that we accept the methods `Count`, `Size`, `Length`, `Order` coming from particular namespaces (e.g., `IEnumerable`). 
In this case, the full name needs to be checked otherwise if a developer creates their own method `count` which might do something different, it's going to be a false negative.
Also, I think the character case needs to be taken into account (no need to turn the method names to lower), as `COUNT()` and `Count()` are two different methods even if they are in the same class.

mary-georgiou-sonarsource: In case the intention is to avoid using the semantic model, at least a strict comparison with equals will create fewer FNs IMO.

mary-georgiou-sonarsource (changes requested):
One comment regarding the `AcceptedNamesForSingleDigitComparison`.

analyzers/src/SonarAnalyzer.CSharp/Rules/MagicNumberShouldNotBeUsed.cs
andrei-epure-sonarsource: well, I'd say that in general if we want to compare anything that is Count to a single digit, it should be fine

`if (Foo.BarCount < 3) {}`

why would the rule ask the dev to put 3 in a magic constant?

andrei-epure-sonarsource: I think the risk would be something like `if (CountOfMonteChristo == 9)`, or something like this, but I'd say it's fairly low

And I dropped the idea of `Order` - should I actually add it?

Will people do something like `if (Foo.Order < 3)` - I guess I should?

andrei-epure-sonarsource:
Updated RSPEC (to be merged after this gets merged) https://github.com/SonarSource/rspec/pull/694

analyzers/rspec/cs/S109_c#.html
```diff
@@ -27,5 +27,12 @@ <h2>Compliant Solution</h2>
...
 </pre>
 <h2>Exceptions</h2>
-<p>This rule doesn’t raise an issue when the magic number is used as part of the <code>GetHashCode</code> method or a variable/field declaration.</p>
+<p>This rule doesn’t raise an issue when the magic number is used as part of:</p>
+<ul>
+  <li> the <code>GetHashCode</code> method </li>
+  <li> a variable/field declaration </li>
+  <li> the single argument of an attribute </li>
+  <li> a named argument for a method or attribute </li>
+  <li> a constructor call </li>
```
andrei-epure-sonarsource, line 36: I haven't included the single-digit comparison because it's hard to explain and isn't worth the effort TBH

analyzers/src/SonarAnalyzer.CSharp/Rules/MagicNumberShouldNotBeUsed.cs
andrei-epure-sonarsource: (offline discussion - the initial idea was to avoid noise when someone is comparing the size of a collection with single-digit number to access the first elements - so I reduced it to this scenario

we will still have FPs because this is a simple AST-based rule so we won't have dataflow and we don't plan to 
we will still have FPs , in general, with this rule - this is why it's not (and probably never will be) part of SonarWay)

analyzers/src/SonarAnalyzer.CSharp/Rules/MagicNumberShouldNotBeUsed.cs
```diff
@@ -82,5 +92,50 @@ private static bool IsInsideProperty(SyntaxNode node)
...
+            && (arg.NameColon is not null
+                || arg.NameEquals is not null
+                || (arg.Parent is AttributeArgumentListSyntax argList && argList.Arguments.Count == 1));
+
+        private static bool IsSingleDigit(string text) => byte.TryParse(text, out var result) && result <= 9;
+
+        // We allow single-digit comparisons when checking the size of a collection, which is usually done to access the first elements.
+        private static bool IsComparingCollectionSize(BinaryExpressionSyntax binaryComparisonToLiteral)
+        {
+            var comparedToLiteral = binaryComparisonToLiteral.Left is LiteralExpressionSyntax ? binaryComparisonToLiteral.Right : binaryComparisonToLiteral.Left;
```
andrei-epure-sonarsource, line 127: add a test with number on left

mary-georgiou-sonarsource (approved):
LGTM

andrei-epure-sonarsource:
Validation: for some projects it didn't reduce the noise, for others it significantly did. For example:

- dnSpy went from 5,3K to 3,9K S109 issues (26% less)
- Microsoft Automatic Graph Layout from 3K to 2.1K (30% less)
- cloudscribe from 1,4K to 770 (~50% less)

Validation done.

PR #5258, opened 2022-01-11
Rule S3928: update case to be a false positive

costin-zaharia-sonarsource:
See https://community.sonarsource.com/t/s3928-with-local-function-false-positive-or-expected-behavior/56003

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CheckArgumentException.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CheckArgumentException.cs
@@ -254,7 +254,7 @@ namespace Tests.Diagnostics
             {
                 if (methodArg == null)
                 {
-                    throw new ArgumentNullException(nameof(methodArg));   // Noncompliant
+                    throw new ArgumentNullException(nameof(methodArg)); // Noncompliant FP, we should add this specific pattern as an exception (see: https://community.sonarsource.com/t/s3928-with-local-function-false-positive-or-expected-behavior/56003)
                 }
             }
         }
--- a/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CheckArgumentException.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CheckArgumentException.cs
@@ -245,6 +245,19 @@ namespace Tests.Diagnostics
             }
         }
 
+        public void ValidationDoneByLocalMethod(string methodArg)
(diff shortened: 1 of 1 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CheckArgumentException.cs
```diff
@@ -254,7 +254,7 @@ void ValidateLocal()
             {
                 if (methodArg == null)
                 {
-                    throw new ArgumentNullException(nameof(methodArg));   // Noncompliant
+                    throw new ArgumentNullException(nameof(methodArg));   // Noncompliant FP, we should add this specific pattern as an exception (see: https://community.sonarsource.com/t/s3928-with-local-function-false-positive-or-expected-behavior/56003)
```
pavel-mikula-sonarsource, line 257: This TP/FP decision gets tricky in the `SendItToSomewhere` case below.  See the comment there, as the other `methodArg` is never null and it still throws `methodArg null exception`

costin-zaharia-sonarsource: The change I did was strictly for this case. And I think that the fact that `methodArg` is still in the scope matters.

Depending on their needs, some devs might want to move the parameter validation to a local method and I think this rule should allow this.

pavel-mikula-sonarsource: Let's say we would allow local functions to do the validation - that makes sense when it's really the case. In that case, we need a bigger change in the code example due to line 251.
`LocalMethodValidatingMethodArgument` needs to be duplicated 2 more times:
* once as is, noncompliant, renamed to `LocalMethodValidatingUsedAsDelegate`
* 2nd the new case, as FP, where `SendItToSomewhere` is not present. The `ValidateLocal` is only invoked directly.
* 3rd the new case, where `ValidateLocal` is not used at all - I'm fine with marking it as FP too.

pavel-mikula-sonarsource (changes requested):
Restarting to move it from "Todo" column back in progress

andrei-epure-sonarsource:
can we close this one, if it's not important?

costin-zaharia-sonarsource:
I've clarified this with Pavel and we plan to merge it. It's a corner case that we still want to document.

pavel-mikula-sonarsource (approved):
LGTM

PR #5106, opened 2021-11-26
Implement S1155 for VB.NET: Collections emptiness checking

Corniel:
Fixes #5253

Given all the clean-up done lately, a walk in the park. Also introduced a base class to make live easier for multi language rules:

``` C#
public abstract class SonarDiagnosticAnalyzer<TSyntaxKind> : SonarDiagnosticAnalyzer
        where TSyntaxKind : struct
{
	protected abstract string MessageFormat { get; }
	protected abstract ILanguageFacade<TSyntaxKind> Language { get; }
	protected DiagnosticDescriptor Rule { get; }

	protected SonarDiagnosticAnalyzer(string diagnosticId) =>
	   Rule = DiagnosticDescriptorBuilder.GetDescriptor(diagnosticId, MessageFormat, Language.RspecResources);

	public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
}
```

Two issues that are worth a consideration of their own: Hexadecimals are not picked up as literals in VB.NET, and you can not retrieve the Generic argument of a method as `methodSymbol.IsGenericMethod` returns false for the `.Count()` extension method. Also, for VB.NET only.

PR #5268, opened 2022-01-12
Fix S2197 FP: False Positive when collection size methods or properties are used

### LINKED ISSUE #4748: Rule S2197: false positive when `Array.Length` or `Enumerable.Count` are used
### Description

Rule S2197 raises false positives when `Array.Length` or `Enumerable.Count` are used.

### Repro steps

See: https://github.com/SonarSource/sonar-dotnet/blob/master/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/EqualityOnModulus.cs#L23-L27

### Expected behavior

No issue raised for this occurrences.

### Actual behavior

Issue is raised

### Related information

* C#/VB.NET Plugins version: 8.26

mary-georgiou-sonarsource:
Fixes #4748

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
@@ -98,13 +98,19 @@ namespace SonarAnalyzer.Rules.CSharp
             && methodSymbol.ReceiverType != null
             && methodSymbol.IsExtensionOn(KnownType.System_Collections_Generic_IEnumerable_T);
 
-        private static bool IsCollectionSizeProperty(IPropertySymbol propertySymbol) =>
-            ((LengthName.Equals(propertySymbol.Name) || LongLengthName.Equals(propertySymbol.Name))
-                && propertySymbol.ContainingType.IsAny(KnownType.System_Array, KnownType.System_String))
-            || (CountName.Equals(propertySymbol.Name)
-                && (propertySymbol.ContainingType.Implements(KnownType.System_Collections_Generic_ICollection_T) || propertySymbol.ContainingType.Implements(KnownType.System_Collections_Generic_IReadOnlyCollection_T)))
-            || (ListCapacityName.Equals(propertySymbol.Name)
-                && propertySymbol.ContainingType.Implements(KnownType.System_Collections_Generic_IList_T));
-    }
+        private static bool IsCollectionSizeProperty(IPropertySymbol propertySymbol)
+        {
+            var containingType = propertySymbol.ContainingType;
+            return (EqualsLengthName(propertySymbol.Name) && containingType.IsAny(KnownType.System_Array, KnownType.System_String))
+            || (CountName.Equals(propertySymbol.Name) && ImplementsICollection(containingType))
+            || (ListCapacityName.Equals(propertySymbol.Name) && containingType.Implements(KnownType.System_Collections_Generic_IList_T));
+        }
 
+        private static bool ImplementsICollection(INamedTypeSymbol typeSymbol) =>
+            typeSymbol.ContainingType.Implements(KnownType.System_Collections_Generic_ICollection_T)
+            || typeSymbol.ContainingType.Implements(KnownType.System_Collections_Generic_IReadOnlyCollection_T);
+
+        private static bool EqualsLengthName(string name) =>
+            LengthName.Equals(name) || LongLengthName.Equals(name);
+    }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
@@ -65,27 +65,25 @@ namespace SonarAnalyzer.Rules.CSharp
             }
         }
 
-        private static bool CheckExpression(SyntaxNode node, ExpressionSyntax modulus, SemanticModel semanticModel, out int constantValue) =>
+        private static bool CheckExpression(SyntaxNode node, ExpressionSyntax expression, SemanticModel semanticModel, out int constantValue) =>
             ExpressionNumericConverter.TryGetConstantIntValue(node, out constantValue)
             && constantValue != 0
-            && IsModulus(modulus)
-            && !ExpressionIsAlwaysPositive(modulus, semanticModel);
+            && expression.RemoveParentheses() is BinaryExpressionSyntax binary
+            && binary.IsKind(SyntaxKind.ModuloExpression)
+            && !ExpressionIsAlwaysPositive(binary, semanticModel);
 
-        private static bool IsModulus(ExpressionSyntax expression) =>
-            expression.RemoveParentheses() is BinaryExpressionSyntax binary
-            && binary.IsKind(SyntaxKind.ModuloExpression);
-
-        private static bool ExpressionIsAlwaysPositive(ExpressionSyntax expression, SemanticModel semantic)
+        private static bool ExpressionIsAlwaysPositive(BinaryExpressionSyntax binaryExpression, SemanticModel semantic)
         {
-            var type = semantic.GetTypeInfo(expression).Type;
+            var type = semantic.GetTypeInfo(binaryExpression).Type;
             if (type.IsAny(KnownType.UnsignedIntegers) || type.Is(KnownType.System_UIntPtr))
             {
                 return true;
             }
 
-            var leftExpression = ((BinaryExpressionSyntax)expression).Left;
-            return CollectionSizePropertyOrMethodNames.Any(x => leftExpression.ToString().Contains(x))
-                   && semantic.GetSymbolInfo(((BinaryExpressionSyntax)expression).Left).Symbol is { } symbol
+            var leftExpression = binaryExpression.Left;
+            var leftExpressionStringForm = leftExpression.ToString();
+            return CollectionSizePropertyOrMethodNames.Any(x => leftExpressionStringForm.Contains(x))
+                   && semantic.GetSymbolInfo(leftExpression).Symbol is { } symbol
                    && IsCollectionSize(symbol);
         }
 
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
@@ -19,6 +19,7 @@
  */
 
 using System.Collections.Immutable;
+using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -33,6 +34,14 @@ namespace SonarAnalyzer.Rules.CSharp
     {
         private const string DiagnosticId = "S2197";
         private const string MessageFormat = "The result of this modulus operation may not be {0}.";
+
+        private static readonly string[] MethodsOrPropertiesReturningAlwaysPositiveInt =
+        {
+            "System.Array.Length",
+            "System.Collections.Generic.IEnumerable<int>.Count<int>()",
+            "System.Collections.Generic.IEnumerable<int>.LongCount<int>()",
+            "System.Collections.Generic.List<int>.Count"
+        };
 
         private static readonly CSharpExpressionNumericConverter ExpressionNumericConverter = new CSharpExpressionNumericConverter();
 
@@ -50,9 +59,7 @@ namespace SonarAnalyzer.Rules.CSharp
             if (CheckExpression(equalsExpression.Left, equalsExpression.Right, c.SemanticModel, out var constantValue)
                 || CheckExpression(equalsExpression.Right, equalsExpression.Left, c.SemanticModel, out constantValue))
             {
-                c.ReportIssue(Diagnostic.Create(Rule,
-                                                               equalsExpression.GetLocation(),
-                                                               constantValue < 0 ? "negative" : "positive"));
+                c.ReportIssue(Diagnostic.Create(Rule, equalsExpression.GetLocation(), constantValue < 0 ? "negative" : "positive"));
             }
         }
 
@@ -60,7 +67,8 @@ namespace SonarAnalyzer.Rules.CSharp
             ExpressionNumericConverter.TryGetConstantIntValue(node, out constantValue)
             && constantValue != 0
             && IsModulus(modulus)
-            && !IsUnsigned(modulus, semanticModel);
+            && !IsUnsigned(modulus, semanticModel)
+            && !IsMethodOrPropertyThatAlwaysReturnsPositiveInt(((BinaryExpressionSyntax)modulus).Left, semanticModel);
 
         private static bool IsModulus(ExpressionSyntax expression) =>
             expression.RemoveParentheses() is BinaryExpressionSyntax binary
@@ -71,6 +79,13 @@ namespace SonarAnalyzer.Rules.CSharp
             var type = semantic.GetTypeInfo(expression).Type;
             return type.IsAny(KnownType.UnsignedIntegers)
                    || type.Is(KnownType.System_UIntPtr);
+        }
+
+        private static bool IsMethodOrPropertyThatAlwaysReturnsPositiveInt(ExpressionSyntax expression, SemanticModel semantic)
+        {
+            var symbol = semantic.GetSymbolInfo(expression).Symbol;
+            return (symbol is IPropertySymbol || symbol is IMethodSymbol)
+                    && MethodsOrPropertiesReturningAlwaysPositiveInt.Contains(symbol.ToDisplayString());
         }
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
@@ -94,8 +94,8 @@ namespace SonarAnalyzer.Rules.CSharp
             IsEnumerableCountMethod(symbol)
             || IsLengthProperty(symbol)
             || IsCollectionCountProperty(symbol)
-            || IsReadonlyCollectionCountProperty(symbol)
             || IsListCapacityProperty(symbol);
+
         private static bool IsEnumerableCountMethod(ISymbol symbol) =>
             (CountName.Equals(symbol.Name) || LongCountName.Equals(symbol.Name))
             && symbol is IMethodSymbol methodSymbol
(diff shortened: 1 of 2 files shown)


analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -50,17 +59,16 @@ private static void VisitEquality(SyntaxNodeAnalysisContext c)
...
+                c.ReportIssue(Diagnostic.Create(Rule, equalsExpression.GetLocation(), constantValue < 0 ? "negative" : "positive"));
             }
         }
 
         private static bool CheckExpression(SyntaxNode node, ExpressionSyntax modulus, SemanticModel semanticModel, out int constantValue) =>
             ExpressionNumericConverter.TryGetConstantIntValue(node, out constantValue)
             && constantValue != 0
             && IsModulus(modulus)
-            && !IsUnsigned(modulus, semanticModel);
+            && !IsUnsigned(modulus, semanticModel)
```
andrei-epure-sonarsource, line 70: pro tip: now that you do this check, you can filter early based on the AST if it contains `Length` or `Count` before querying the semantic model

also, you seem to do 2 semantic model checks for the same `expression` - is it possible to do the job with just one?

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -34,6 +35,14 @@ public sealed class EqualityOnModulus : SonarDiagnosticAnalyzer
         private const string DiagnosticId = "S2197";
         private const string MessageFormat = "The result of this modulus operation may not be {0}.";
 
+        private static readonly string[] MethodsOrPropertiesReturningAlwaysPositiveInt =
+        {
+            "System.Array.Length",
+            "System.Collections.Generic.IEnumerable<int>.Count<int>()",
```
csaba-sagi-sonarsource, line 41: It should not raise for any IEnumerable, not just the one of integers.

csaba-sagi-sonarsource: Rules S1155 and S3981 check similar things and might contain some things that could help you..

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -34,6 +35,14 @@ public sealed class EqualityOnModulus : SonarDiagnosticAnalyzer
...
+            "System.Collections.Generic.IEnumerable<int>.LongCount<int>()",
```
csaba-sagi-sonarsource, line 42: Ignorable: I am thinking if we should cover `String.Length` and `List.Capacity` or not. It might be an overkill and in the same time it is an FP

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
mary-georgiou-sonarsource: I wanted to keep them separate since they are logically checking 2 different things but I see the potential perf improvement.

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -72,5 +80,12 @@ private static bool IsUnsigned(ExpressionSyntax expression, SemanticModel semant
             return type.IsAny(KnownType.UnsignedIntegers)
```
csaba-sagi-sonarsource, line 80: Lets create a field containing both the UnsignedIntegers and the UintPtr and using that with .IsAny().

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -50,27 +61,62 @@ private static void VisitEquality(SyntaxNodeAnalysisContext c)
...
             && binary.IsKind(SyntaxKind.ModuloExpression);
 
-        private static bool IsUnsigned(ExpressionSyntax expression, SemanticModel semantic)
+        private static bool ExpressionIsAlwaysPositive(ExpressionSyntax expression, SemanticModel semantic)
         {
             var type = semantic.GetTypeInfo(expression).Type;
-            return type.IsAny(KnownType.UnsignedIntegers)
-                   || type.Is(KnownType.System_UIntPtr);
+            var isUint = type.IsAny(KnownType.UnsignedIntegers)
+                          || type.Is(KnownType.System_UIntPtr);
```
andrei-epure-sonarsource, line 82: I think it's clearer if you do an early return based on isUint

```suggestion
            if (type.IsAny(KnownType.UnsignedIntegers, KnownType.System_UIntPtr))
            {
              return true;
             }
```

and then below you won't need `!isUint` and `return isUint`

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -50,27 +61,62 @@ private static void VisitEquality(SyntaxNodeAnalysisContext c)
...
+
+            var leftExpression = ((BinaryExpressionSyntax)expression).Left;
+            if (!isUint && CollectionSizePropertyOrMethodNames.Any(x => leftExpression.ToString().Contains(x)))
+            {
+                var symbol = semantic.GetSymbolInfo(((BinaryExpressionSyntax)expression).Left).Symbol;
+                return IsCollectionSizeMethodOrProperty(symbol);
+            }
+            return isUint;
```
andrei-epure-sonarsource, line 90: this can then be
```suggestion
            return CollectionSizePropertyOrMethodNames.Any(x => leftExpression.ToString().Contains(x)))
              && semantic.GetSymbolInfo(((BinaryExpressionSyntax)expression).Left).Symbol is { } symbol
              && IsCollectionSizeMethodOrProperty(symbol);

```

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -50,27 +61,62 @@ private static void VisitEquality(SyntaxNodeAnalysisContext c)
...
         }
+
+        private static bool IsCollectionSizeMethodOrProperty(ISymbol symbol) =>
```
andrei-epure-sonarsource, line 93: IMO `IsCollectionSize` is a better name

method or property is an implementation detail that can be seen easily

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -50,27 +61,62 @@ private static void VisitEquality(SyntaxNodeAnalysisContext c)
...
+            (CountName.Equals(symbol.Name) || LongCountName.Equals(symbol.Name))
+            && symbol is IMethodSymbol methodSymbol
+            && methodSymbol.IsExtensionMethod
+            && methodSymbol.ReceiverType != null
+            && methodSymbol.IsExtensionOn(KnownType.System_Collections_Generic_IEnumerable_T);
+
+        private static bool IsLengthProperty(ISymbol symbol) =>
+            (LengthName.Equals(symbol.Name) || LongLengthName.Equals(symbol.Name))
+            && symbol is IPropertySymbol propertySymbol
+            && (propertySymbol.ContainingType.Is(KnownType.System_Array) || propertySymbol.ContainingType.Is(KnownType.System_String));
```
andrei-epure-sonarsource, line 109: can use the `IsAny()` extension method

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -50,27 +61,62 @@ private static void VisitEquality(SyntaxNodeAnalysisContext c)
...
+
+        private static bool IsCollectionCountProperty(ISymbol symbol) =>
+            CountName.Equals(symbol.Name)
+            && symbol is IPropertySymbol propertySymbol
+            && (propertySymbol.ContainingType.DerivesOrImplements(KnownType.System_Collections_Generic_ICollection_T)
```
andrei-epure-sonarsource, line 114: will `Implements()` suffice, given that you're only checking interfaces? what's the use case for derives? I cannot see it in the UTs

andrei-epure-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
mary-georgiou-sonarsource: I kept also `Derives()` for safety. I think I will keep only `Implements()` to keep it simple for now.
WDYT?

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -50,27 +61,61 @@ private static void VisitEquality(SyntaxNodeAnalysisContext c)
...
             var type = semantic.GetTypeInfo(expression).Type;
-            return type.IsAny(KnownType.UnsignedIntegers)
-                   || type.Is(KnownType.System_UIntPtr);
+            if (type.IsAny(KnownType.UnsignedIntegers) || type.Is(KnownType.System_UIntPtr))
+            {
+                return true;
+            }
+
+            var leftExpression = ((BinaryExpressionSyntax)expression).Left;
+            return CollectionSizePropertyOrMethodNames.Any(x => leftExpression.ToString().Contains(x))
```
andrei-epure-sonarsource, line 87: I realize that `leftExpression.ToString()` is executed multiple times although it's not necessary - you can extract that

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -50,27 +61,61 @@ private static void VisitEquality(SyntaxNodeAnalysisContext c)
...
+                c.ReportIssue(Diagnostic.Create(Rule, equalsExpression.GetLocation(), constantValue < 0 ? "negative" : "positive"));
             }
         }
 
         private static bool CheckExpression(SyntaxNode node, ExpressionSyntax modulus, SemanticModel semanticModel, out int constantValue) =>
             ExpressionNumericConverter.TryGetConstantIntValue(node, out constantValue)
             && constantValue != 0
             && IsModulus(modulus)
-            && !IsUnsigned(modulus, semanticModel);
+            && !ExpressionIsAlwaysPositive(modulus, semanticModel);
```
andrei-epure-sonarsource, line 72: the check `expression.RemoveParentheses() is BinaryExpressionSyntax binary` is done in `IsModulus()`

and  `var leftExpression = ((BinaryExpressionSyntax)expression).Left;` is done inside `ExpressionIsAlwaysPositive`

this makes the code hard to read

you could just inline `IsModulus` and then change the signature of `ExpressionIsAlwaysPositive` to receive `BinaryExpressionSyntax`

like this you would also avoid having the surprise that `expression` is actually a parenthesized expression (you don't have a test but I would expect it would throw when doing the cast in `ExpressionIsAlwaysPositive()`

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
mary-georgiou-sonarsource: Update: After doing a quick test `Derives` is not really needed at all.

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -50,27 +61,61 @@ private static void VisitEquality(SyntaxNodeAnalysisContext c)
...
             expression.RemoveParentheses() is BinaryExpressionSyntax binary
             && binary.IsKind(SyntaxKind.ModuloExpression);
 
-        private static bool IsUnsigned(ExpressionSyntax expression, SemanticModel semantic)
+        private static bool ExpressionIsAlwaysPositive(ExpressionSyntax expression, SemanticModel semantic)
         {
             var type = semantic.GetTypeInfo(expression).Type;
-            return type.IsAny(KnownType.UnsignedIntegers)
-                   || type.Is(KnownType.System_UIntPtr);
+            if (type.IsAny(KnownType.UnsignedIntegers) || type.Is(KnownType.System_UIntPtr))
```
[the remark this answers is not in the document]
mary-georgiou-sonarsource: `KnownType.UnsignedIntegers` is an `ImmutableArray`.
I should first put all the elements from `KnownType.UnsignedIntegers` + the single element `KnownType.System_UIntPtr`  in one `IEnumerable<KnownType>` and then use `IsAny`. 
I don't think it really worth it here. Except if I'm missing something of course.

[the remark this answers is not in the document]
andrei-epure-sonarsource: ah ok sorry

downside of not doing the CR in the IDE

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
Corniel: Note: Rule S1155 is currently in the process of being refactored (to support VB.NET as well) #5106

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -34,7 +37,15 @@ public sealed class EqualityOnModulus : SonarDiagnosticAnalyzer
         private const string DiagnosticId = "S2197";
         private const string MessageFormat = "The result of this modulus operation may not be {0}.";
 
-        private static readonly CSharpExpressionNumericConverter ExpressionNumericConverter = new CSharpExpressionNumericConverter();
+        private const string CountName = nameof(Enumerable.Count);
+        private const string LongCountName = nameof(Enumerable.LongCount);
+        private const string LengthName = "Length";
```
Corniel, line 42: nameof(String.Length)

mary-georgiou-sonarsource: I intentionally set it as `"Length"` because it represents both the `Array.Length` and the `String.Length`.
Although `nameof(String.Length)` would give the same result, it would probably be misleading.

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
andrei-epure-sonarsource: your update matches what I had in mind

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/EqualityOnModulus.cs
```diff
@@ -20,11 +20,26 @@ public void isOdd(int x)
...
+
+            var list = new List<int>();
+            y = list.Count % 2 == 1; // Compliant - Count property is > 0;
+            y = list.Capacity % 2 == 1; // Compliant - Capacityproperty is > 0;
+
+            var someString = "HelloImAWhale";
+            y = someString.Length % 2 == 1; // Compliant - Length property is > 0;
+
+            y = FakeCount() % 2 == 1; // Noncompliant
         }
```
andrei-epure-sonarsource, line 41: have a test w/ parens?
`(enumerableLong.Count() % 2) == 1`

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
andrei-epure-sonarsource: you can add that test to the UT to document the expected behavior

andrei-epure-sonarsource (approved):
I left 2 comments to add more tests, LGTM

leaving @csaba-sagi-sonarsource to finish the review

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
csaba-sagi-sonarsource, line 42: Why didn't you use nameof(Array.Length); if yo used the same approach for all the other names?

csaba-sagi-sonarsource: Lets just put a comment there please that it stands for both.

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
```diff
@@ -50,27 +61,59 @@ private static void VisitEquality(SyntaxNodeAnalysisContext c)
...
+            || IsListCapacityProperty(symbol);
+
+        private static bool IsEnumerableCountMethod(ISymbol symbol) =>
+            (CountName.Equals(symbol.Name) || LongCountName.Equals(symbol.Name))
+            && symbol is IMethodSymbol methodSymbol
+            && methodSymbol.IsExtensionMethod
+            && methodSymbol.ReceiverType != null
+            && methodSymbol.IsExtensionOn(KnownType.System_Collections_Generic_IEnumerable_T);
+
+        private static bool IsLengthProperty(ISymbol symbol) =>
```
csaba-sagi-sonarsource, line 103: I would  group these three checks IsLengthProperty(symbol) IsCollectionCountProperty(symbol) and IsListCapacityProperty(symbol) into a different method and do the `is IPropertySymbol` check once and use the `PropertySymbol` as an argument for them.

csaba-sagi-sonarsource (changes requested):
Last round of changes. :)

Corniel:
How this rule behave on types that define the % operator themselves? May be just ignore them.

analyzers/src/SonarAnalyzer.CSharp/Rules/EqualityOnModulus.cs
csaba-sagi-sonarsource: NITPICK: I think you could have left the three methods to divide the different types with different properties like: 

```
        private static bool IsLengthProperty(IPropertySymbol propertySymbol) =>
            (LengthName.Equals(propertySymbol.Name) || LongLengthName.Equals(propertySymbol.Name))
            && propertySymbol.ContainingType.IsAny(KnownType.System_Array, KnownType.System_String);

        private static bool IsCollectionCountProperty(IPropertySymbol propertySymbol) =>
            CountName.Equals(propertySymbol.Name)
            && (propertySymbol.ContainingType.Implements(KnownType.System_Collections_Generic_ICollection_T)
                || propertySymbol.ContainingType.Implements(KnownType.System_Collections_Generic_IReadOnlyCollection_T));

        private static bool IsListCapacityProperty(IPropertySymbol propertySymbol) =>
            ListCapacityName.Equals(propertySymbol.Name)
            && propertySymbol.ContainingType.Implements(KnownType.System_Collections_Generic_IList_T);
```

csaba-sagi-sonarsource (approved):
Please take a look at my last comment, however I approved the PR, so if you think it does not make sense you can skip applying it.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/EqualityOnModulus.cs
```diff
@@ -37,8 +37,9 @@ public void isOdd(int x)
 
             y = FakeCount() % 2 == 1; // Noncompliant
 
-            y = (enumerableLong.Count() % 2) == 1; // Compliant
-            y = 1 == (x % 2); // Noncompliant {{The result of this modulus operation may not be positive.}}
```
andrei-epure-sonarsource: why did you remove this test with the literal on the left?

PR #5270, opened 2022-01-13
Add VerifyCodeFix to VerifierBuilder

pavel-mikula-sonarsource:
Contributes to #5186 

*Squash* commits doesn't need a review

PR #5277, opened 2022-01-14
UT Verifier: Read codefix paths from TestCases directory

pavel-mikula-sonarsource:
Leftover from #5270 where I forgot to make the fixed path relative to the `TestCases` directory.


costin-zaharia-sonarsource (approved):
LGTM! with the note that I would rename `TestCasePath` to `ComputeTestCasePath` or `GetTestCasePath`

pavel-mikula-sonarsource:
I would not :)

PR #5279, opened 2022-01-17
Remove run_rules_cov from release.yml

mary-georgiou-sonarsource:
According to the release engineering team, we should not run `rules-cov` anymore as it's handled differently now.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,6 @@ jobs:
           publish_to_binaries: true
           distribute: true
           attach_artifacts_to_github_release: true
-          run_rules_cov: true
           slack_channel: team-lang-dotnet
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -19,7 +19,6 @@ jobs:
           publish_to_binaries: true
           distribute: true
           attach_artifacts_to_github_release: true
-          run_rules_cov: true
```
pavel-mikula-sonarsource, line 22: @tomverin Should we update all actions to @v4 as well? There was some random message on #team-lang about it.

pavel-mikula-sonarsource: @mary-georgiou-sonarsource Let's update it everywhere from `@v3` to `@v4`, it worked on the VB6 one.

pavel-mikula-sonarsource (changes requested):

pavel-mikula-sonarsource (approved):
LGTM

PR #5282, opened 2022-01-18
S1155: Implement a fix provider for C#

Corniel:
As specified in #2334 .

Note to @pavel-mikula-sonarsource the `VerifyBuilder` (still) requires the 'TestCases' root directory to be specified, is that intentionally?

### DIFF
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
@@ -66,9 +66,9 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static Task<Document> Replacement(SyntaxNode root, ExpressionSyntax expression, InvocationExpressionSyntax count, CountComparisonResult comparison, CodeFixContext context)
         {
-            var memberAccess = (MemberAccessExpressionSyntax)count.Expression;
-            var name = memberAccess.WithName(SyntaxFactory.IdentifierName(nameof(Enumerable.Any)));
-            var any = SyntaxFactory.InvocationExpression(name, count.ArgumentList);
+            var any = NotAsExtenion(count)
+                ? AnyAsExtension(count)
+                : AnyFromExtension(count);
 
             SyntaxNode replacement = comparison == CountComparisonResult.Empty
                 ? SyntaxFactory.PrefixUnaryExpression(SyntaxKind.LogicalNotExpression, any)
@@ -76,5 +76,25 @@ namespace SonarAnalyzer.Rules.CSharp
 
             return Task.FromResult(context.Document.WithSyntaxRoot(root.ReplaceNode(expression, replacement).WithAdditionalAnnotations(Formatter.Annotation)));
         }
+
+        private static InvocationExpressionSyntax AnyFromExtension(InvocationExpressionSyntax count)
+        {
+            var memberAccess = (MemberAccessExpressionSyntax)count.Expression;
+            var name = memberAccess.WithName(SyntaxFactory.IdentifierName(nameof(Enumerable.Any)));
+            return SyntaxFactory.InvocationExpression(name, count.ArgumentList);
+        }
+
+        private static InvocationExpressionSyntax AnyAsExtension(InvocationExpressionSyntax count)
+        {
+            var identifier = (IdentifierNameSyntax)count.ArgumentList.Arguments[0].Expression;
+            var name = SyntaxFactory.MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression, identifier, SyntaxFactory.IdentifierName(nameof(Enumerable.Any)));
+            var arguments = SyntaxFactory.ArgumentList(count.ArgumentList.Arguments.RemoveAt(0));
+            return SyntaxFactory.InvocationExpression(name, arguments);
+        }
+
+        private static bool NotAsExtenion(InvocationExpressionSyntax count)
+            => count.ArgumentList.Arguments.Any()
+            && count.ArgumentList.Arguments[0].Expression is IdentifierNameSyntax
+            && ((MemberAccessExpressionSyntax)count.Expression).Expression.NameIs(nameof(Enumerable));
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
@@ -32,7 +32,7 @@ using SonarAnalyzer.Helpers;
 namespace SonarAnalyzer.Rules.CSharp
 {
     [ExportCodeFixProvider(LanguageNames.CSharp)]
-    public sealed class CollectionEmptinessCheckingFixProvider : SonarCodeFixProvider
+    public sealed class CollectionEmptinessCheckingCodeFix : SonarCodeFixProvider
     {
         internal const string Title = "Use Any() instead";
         public override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(CollectionEmptinessChecking.DiagnosticId);
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
@@ -66,9 +66,9 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static Task<Document> Replacement(SyntaxNode root, ExpressionSyntax expression, InvocationExpressionSyntax count, CountComparisonResult comparison, CodeFixContext context)
         {
-            var any = NotAsExtenion(count)
-                ? AnyAsExtension(count)
-                : AnyFromExtension(count);
+            var any = IsExtension(count)
+                ? AnyFromExtension(count)
+                : AnyAsExtension(count);
 
             SyntaxNode replacement = comparison == CountComparisonResult.Empty
                 ? SyntaxFactory.PrefixUnaryExpression(SyntaxKind.LogicalNotExpression, any)
@@ -86,15 +86,14 @@ namespace SonarAnalyzer.Rules.CSharp
 
         private static InvocationExpressionSyntax AnyAsExtension(InvocationExpressionSyntax count)
         {
-            var identifier = (IdentifierNameSyntax)count.ArgumentList.Arguments[0].Expression;
-            var name = SyntaxFactory.MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression, identifier, SyntaxFactory.IdentifierName(nameof(Enumerable.Any)));
+            var expression = count.ArgumentList.Arguments[0].Expression;
+            var name = SyntaxFactory.MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression, expression, SyntaxFactory.IdentifierName(nameof(Enumerable.Any)));
             var arguments = SyntaxFactory.ArgumentList(count.ArgumentList.Arguments.RemoveAt(0));
             return SyntaxFactory.InvocationExpression(name, arguments);
         }
 
-        private static bool NotAsExtenion(InvocationExpressionSyntax count)
-            => count.ArgumentList.Arguments.Any()
-            && count.ArgumentList.Arguments[0].Expression is IdentifierNameSyntax
-            && ((MemberAccessExpressionSyntax)count.Expression).Expression.NameIs(nameof(Enumerable));
+        private static bool IsExtension(InvocationExpressionSyntax count) =>
+            !count.ArgumentList.Arguments.Any()
+            || !((MemberAccessExpressionSyntax)count.Expression).Expression.NameIs(nameof(Enumerable));
     }
 }
--- a/analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
+++ b/analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
@@ -32,9 +32,10 @@ using SonarAnalyzer.Helpers;
 namespace SonarAnalyzer.Rules.CSharp
 {
     [ExportCodeFixProvider(LanguageNames.CSharp)]
-    public sealed class CollectionEmptinessCheckingCodeFix : SonarCodeFixProvider
+    public sealed class CollectionEmptinessCheckingCodeFix : SonarCodeFix
     {
-        internal const string Title = "Use Any() instead";
+        private const string Title = "Use Any() instead";
+
         public override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(CollectionEmptinessChecking.DiagnosticId);
 
         private readonly CSharpFacade language = CSharpFacade.Instance;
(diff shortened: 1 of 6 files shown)


Corniel:
@pavel-mikula-sonarsource : any ETA for this feature?

pavel-mikula-sonarsource:
Hi @Corniel, the explicit `TestCases` in codefixes was not intentional and it was fixed since then. I've bumped into it when doing some sample refactorings.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/CollectionEmptinessCheckingTest.cs
```diff
@@ -30,10 +30,25 @@ public class CollectionEmptinessCheckingTest
...
+            new VerifierBuilder()
+                .AddAnalyzer(() => new CS.CollectionEmptinessChecking())
+                .AddPaths("CollectionEmptinessChecking.cs")
+                .Verify();
 
         [TestMethod]
         public void CollectionEmptinessChecking_VB() =>
-            OldVerifier.VerifyAnalyzer(@"TestCases\CollectionEmptinessChecking.vb", new VB.CollectionEmptinessChecking());
+            new VerifierBuilder()
+                .AddAnalyzer(() => new VB.CollectionEmptinessChecking())
```
pavel-mikula-sonarsource, line 41: Can be simplified. And then, the `AddPaths` and `Verify` can be on a single line.
```suggestion
            new VerifierBuilder<VB.CollectionEmptinessChecking>()
```

analyzers/tests/SonarAnalyzer.UnitTest/Rules/CollectionEmptinessCheckingTest.cs
```diff
@@ -30,10 +30,25 @@ public class CollectionEmptinessCheckingTest
     {
         [TestMethod]
         public void CollectionEmptinessChecking_CS() =>
-            OldVerifier.VerifyAnalyzer(@"TestCases\CollectionEmptinessChecking.cs", new CS.CollectionEmptinessChecking());
+            new VerifierBuilder()
+                .AddAnalyzer(() => new CS.CollectionEmptinessChecking())
+                .AddPaths("CollectionEmptinessChecking.cs")
```
pavel-mikula-sonarsource, line 35: You can actually do this, because rule and the codefix share this. And we don't have any other complicated scenario here.
```
private readonly VerifierBuilder builderCS = new VerifierBuilder<CS.CollectionEmptinessChecking>().AddPaths("CollectionEmptinessChecking.cs");

[TestMethod]
        public void CollectionEmptinessChecking_CS() =>
            builderCS.Verify();
```

pavel-mikula-sonarsource: If you don't like the idea of sharing `AddPaths`, the `builderCS` defining the shared rule should be used.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/CollectionEmptinessCheckingTest.cs
```diff
@@ -30,10 +30,25 @@ public class CollectionEmptinessCheckingTest
...
+                .AddPaths("CollectionEmptinessChecking.vb")
+                .Verify();
+
+        [TestMethod]
+        public void CollectionEmptinessChecking_CodeFix_CS() =>
+            new VerifierBuilder()
+                .AddAnalyzer(() => new CS.CollectionEmptinessChecking())
+                .AddPaths("CollectionEmptinessChecking.cs")
+                .WithCodeFix<CS.CollectionEmptinessCheckingCodeFix>()
+                .WithCodeFixedPath(@"TestCases\CollectionEmptinessChecking.Fixed.cs")
```
pavel-mikula-sonarsource, line 51: This problem was already fixed on our master
```suggestion
                .WithCodeFixedPath(@"CollectionEmptinessChecking.Fixed.cs")
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CollectionEmptinessChecking.Fixed.cs
```diff
@@ -0,0 +1,96 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace Tests.Diagnostics
+{
+    public class CollectionEmptinessChecking
+    {
+        private static bool HasContent1(IEnumerable<string> l)
+        {
+            return l.Any; // Fixed
```
pavel-mikula-sonarsource, line 10: This would be fine for VB, but C# forces devs to be more compiler-friendly 😄 
```suggestion
            return l.Any(); // Fixed
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CollectionEmptinessChecking.Fixed.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+        private static bool HasContent1b(IEnumerable<string> l)
+        {
+            return l.Any; // Fixed
+        }
+        private static bool HasContent2(List<string> l)
+        {
+            return Enumerable.Any; // Fixed
```
pavel-mikula-sonarsource, line 18: This is not a correct fix for the original line
```suggestion
            return l.Any(); // Fixed
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CollectionEmptinessChecking.Fixed.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+        private static bool HasContent2b(List<string> l)
+        {
+            return Enumerable.Any; // Fixed
```
pavel-mikula-sonarsource, line 22: Same as above, we need to preserve the `l` part.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CollectionEmptinessChecking.Fixed.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            return !l.Any; // Fixed
+        }
+        private static bool IsEmpty4b(List<string> l)
+        {
+            return !l.Any; // Fixed
+        }
+        private static bool HasContentWithCondition(List<int> numbers)
+        {
+            return numbers.Any; // Fixed
```
pavel-mikula-sonarsource, line 54: This isn't the correct fix. We need to preserve the linq condition to  achieve the same result.
```suggestion
            return numbers.Any(n => n % 2 == 0); // Fixed
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CollectionEmptinessChecking.Fixed.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+        private static bool IsEmptyWithCondition(List<int> numbers)
+        {
+            return !numbers.Any; // Fixed
```
pavel-mikula-sonarsource, line 58: Same as above, we need to preserve the linq condition

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
```diff
@@ -0,0 +1,92 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Formatting;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public sealed class CollectionEmptinessCheckingCodeFix : SonarCodeFixProvider
+    {
+        internal const string Title = "Use Any() instead";
```
pavel-mikula-sonarsource, line 37: Can be private. Should be followed by an empty line.
```suggestion
        private const string Title = "Use Any() instead";
```

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+            return Task.CompletedTask;
+        }
+
+        public static void Simplify(
+            SyntaxNode root,
+            ExpressionSyntax expression,
+            ExpressionSyntax countExpression,
+            CountComparisonResult comparisonResult,
+            CodeFixContext context) =>
```
pavel-mikula-sonarsource, line 66: This is short enough to fit on a single line. Otherwise it would need to be formatted differenly.

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            context.RegisterCodeFix(
+                CodeAction.Create(
+                    Title,
+                    c => Replacement(root, expression, countExpression, comparisonResult, context)),
```
pavel-mikula-sonarsource, line 71: This can be a single line. 
We haven't cleaned up code fixes in a while so others are not a very nice source of inspiration, unfortunately.
```suggestion
                CodeAction.Create(Title, c => Replacement(root, expression, countExpression, comparisonResult, context)),
```

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
```diff
@@ -0,0 +1,92 @@
...
+                context.Diagnostics);
+
+        private static Task<Document> Replacement(
```
pavel-mikula-sonarsource, line 74: Same formatting as above

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
```diff
@@ -0,0 +1,92 @@
...
+            SyntaxNode root,
+            ExpressionSyntax expression,
+            ExpressionSyntax countExpression,
+            CountComparisonResult comparisonResult,
+            CodeFixContext context)
+        {
```
pavel-mikula-sonarsource, line 80: This logic will need to be more complicated, see comments on the TestCases

pavel-mikula-sonarsource (changes requested):
Hi @Corniel, thank you for trying to implement this codefix. I'll try to prioritize this PR now.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/CollectionEmptinessCheckingTest.cs
Corniel: I would argue that you want to have these tests isolated; and therefore not share setup. But I've you prefer this, I can change it.

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
Corniel: @pavel-mikula-sonarsource Well, it was wrong, but not that complicated (after all). Should be covered now. I also did a forward integration.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CollectionEmptinessChecking.Fixed.cs
pavel-mikula-sonarsource: This still doesn't produce `l.Any`. I would not be happy with `Enumerable.Any(l)` syntax in my code :/

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
pavel-mikula-sonarsource: Still needs to be a bit more complicated to cover the last unresolved case above :)

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CollectionEmptinessChecking.Fixed.cs
```diff
@@ -0,0 +1,96 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace Tests.Diagnostics
+{
+    public class CollectionEmptinessChecking
+    {
```
pavel-mikula-sonarsource, line 7: Let's add UT like this to see how the codefix behaves:
```
return a.Count() > 0 && b.Count() > 0 && c.Count() > 0
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CollectionEmptinessChecking.Fixed.cs
Corniel: Sorry for the misunderstanding. If I typed Enumerable.Count(l), i would argue that was intended. But I'm happy to convert that expression too.

pavel-mikula-sonarsource: Let's try to convert it. 
The current implementation is rather short and clean. If it explodes into something too complicated, we can stop and keep it as it is.

Corniel:
I added some complex (nested, and composed) tests. Furthermore I extended the logic to convert `Emumerable.Count()` into an extenion  method call. I'm 100% happy with some of the sub methods names, so I'm open to suggestions.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CollectionEmptinessChecking.cs
```diff
@@ -59,6 +58,52 @@ private static bool IsEmptyWithCondition(List<int> numbers)
...
+        {
+            return Enumerable.Count(n, x => x != 2) == 0; // Noncompliant
+        }
+    }
+
+    public class Complex
+    {
+        bool Nested(string[] words)
+        {
+            return words.Count(w => w.Count(ch => ch == 'Q') == 0) > 0;
```
pavel-mikula-sonarsource, line 96: That's a nasty test case 😄 💪

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
```diff
@@ -0,0 +1,100 @@
...
+        public static void Simplify(SyntaxNode root, ExpressionSyntax expression, ExpressionSyntax countExpression, CountComparisonResult comparisonResult, CodeFixContext context) =>
+            context.RegisterCodeFix(
+                CodeAction.Create(Title, c => Replacement(root, expression, (InvocationExpressionSyntax)countExpression, comparisonResult, context)),
+                context.Diagnostics);
+
+        private static Task<Document> Replacement(SyntaxNode root, ExpressionSyntax expression, InvocationExpressionSyntax count, CountComparisonResult comparison, CodeFixContext context)
+        {
+            var any = NotAsExtenion(count)
+                ? AnyAsExtension(count)
+                : AnyFromExtension(count);
```
pavel-mikula-sonarsource, line 71: Proposal: To keep the `From` in both cases to describe the difference.
```suggestion
                ? AnyFromStaticMethod(count)
                : AnyFromExtension(count);
```

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        public static void Simplify(SyntaxNode root, ExpressionSyntax expression, ExpressionSyntax countExpression, CountComparisonResult comparisonResult, CodeFixContext context) =>
+            context.RegisterCodeFix(
+                CodeAction.Create(Title, c => Replacement(root, expression, (InvocationExpressionSyntax)countExpression, comparisonResult, context)),
+                context.Diagnostics);
+
+        private static Task<Document> Replacement(SyntaxNode root, ExpressionSyntax expression, InvocationExpressionSyntax count, CountComparisonResult comparison, CodeFixContext context)
+        {
+            var any = NotAsExtenion(count)
```
pavel-mikula-sonarsource, line 69: I'd name this `IsExtension` and then swap the true/false part. It will make the logic positive and easier to understand.
It would read as `if IsExtension then AnyFromExtension else AnyFrom<the other>`

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CollectionEmptinessChecking.cs
Corniel: And it works. :)

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        private static InvocationExpressionSyntax AnyAsExtension(InvocationExpressionSyntax count)
+        {
+            var identifier = (IdentifierNameSyntax)count.ArgumentList.Arguments[0].Expression;
+            var name = SyntaxFactory.MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression, identifier, SyntaxFactory.IdentifierName(nameof(Enumerable.Any)));
+            var arguments = SyntaxFactory.ArgumentList(count.ArgumentList.Arguments.RemoveAt(0));
+            return SyntaxFactory.InvocationExpression(name, arguments);
+        }
+
+        private static bool NotAsExtenion(InvocationExpressionSyntax count)
```
pavel-mikula-sonarsource, line 95: _Nitpick:_ Formatting
```suggestion
        private static bool NotAsExtenion(InvocationExpressionSyntax count) =>
```

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/CollectionEmptinessChecking.Fixed.cs
Corniel: See implementation: it works, but is less elegant.

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
```diff
@@ -0,0 +1,100 @@
...
+            => count.ArgumentList.Arguments.Any()
+            && count.ArgumentList.Arguments[0].Expression is IdentifierNameSyntax
```
pavel-mikula-sonarsource, line 97: This shouldn't be limited to `IdentifierNameSyntax`.
```suggestion
```
Please add this test case:
```
return Enumerable.Count(n.Concat(n)) > 0; // Noncompliant
```

pavel-mikula-sonarsource (changes requested):
We're almost there

analyzers/src/SonarAnalyzer.CSharp/Rules/CollectionEmptinessCheckingCodeFix.cs
Corniel: Applied your suggestions @pavel-mikula-sonarsource

pavel-mikula-sonarsource (approved):
LGTM! Thank you @Corniel for taking a look into this.

Coverage: I'll merge it as is. Coverage is low due to defensive coding that is not recoverable by UTs. This is typical for CodeFixes as the rule introduces some constraints about where the code fix can be triggered.

PR #5283, opened 2022-01-18
Use SonarDiagnosticAnalyzer<TSyntaxKind> base type

Corniel:
Refactored all SonarDiagnosticAnalyzer base classes contained

``` C#
protected abstract ILanguageFacade<TSyntaxKind> Language { get; }
```

To inhertit from `SonarDiagnosticAnalyzer <TSyntaxKind>` instead (introduced in PR #5106 )

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Rules/ClassNotInstantiatableBase.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Rules/ClassNotInstantiatableBase.cs
@@ -19,7 +19,6 @@
  */
 
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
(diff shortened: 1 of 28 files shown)


analyzers/src/SonarAnalyzer.Common/Rules/ClassNotInstantiatableBase.cs
```diff
@@ -28,23 +27,17 @@
...
-        protected abstract ILanguageFacade<TSyntaxKind> Language { get; }
 
         protected abstract IEnumerable<ConstructorContext> CollectRemovableDeclarations(INamedTypeSymbol namedType, Compilation compilation, string messageArg);
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(rule);
+        protected override string MessageFormat => "This {0} can't be instantiated; make {1} 'public'.";
 
-        protected ClassNotInstantiatableBase() =>
-            rule = DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, Language.RspecResources);
+        protected ClassNotInstantiatableBase() : base(DiagnosticId) { }
```
pavel-mikula-sonarsource, line 40: Both inheritors of theese class needs the `rule` -> `Rule` rename

pavel-mikula-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Common/Rules/ClassNotInstantiatableBase.cs
pavel-mikula-sonarsource: Fixed

pavel-mikula-sonarsource:
3 UTs throws `NullReferenceException`. Can you please take a look at that?

Corniel:
> 3 UTs throws `NullReferenceException`. Can you please take a look at that?

Done. Small issue, what should have been spotted If I would have run the tests locally, what I forgot (it was late, lame excuse).

pavel-mikula-sonarsource (approved):
LGTM, thank you @Corniel for taking the further step after introducing the generic base class.

PR #5288, opened 2022-01-19
Fix UTs after FluentAssertions change on net6.0

costin-zaharia-sonarsource:
FluentAssertions behaves inconsistently when showing the error message. Sometimes it will find the actual variable name as `ActualCodeWithReplacedComments().ToUnixLineEndings()` and sometimes as `string`.

In order to update the library to the latest effort additional effort is needed since there are breaking changes.


pavel-mikula-sonarsource (approved):
LGTM

PR #5289, opened 2022-01-19
UTs: Remove OldVerifier.VerifyVisualBasicAnalyzer and one VerifyCSharpAnalyzer overload

pavel-mikula-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -70,7 +70,7 @@ public TestSetup(string testCase, SonarDiagnosticAnalyzer analyzer, IEnumerable<
             new TestSetup("AnonymousDelegateEventUnsubscribe.cs", new AnonymousDelegateEventUnsubscribe()),
             // S2699 - TEST only
             new TestSetup("TestMethodShouldContainAssertion.NUnit.cs", new TestMethodShouldContainAssertion(),
-                TestMethodShouldContainAssertionTest.AdditionalTestReferences(NuGetMetadataReference.NUnit(Constants.NuGetLatestVersion))),
+                TestMethodShouldContainAssertionTest.WithTestReferences(NuGetMetadataReference.NUnit(Constants.NuGetLatestVersion)).References),    // ToDo: Reuse the entire builder in TestSetup
```
csaba-sagi-sonarsource, line 73: NITPICK: Is the ToDO comment really needed?

analyzers/tests/SonarAnalyzer.UnitTest/Rules/ParametersCorrectOrderTest.cs
```diff
@@ -28,15 +28,16 @@ namespace SonarAnalyzer.UnitTest.Rules
     [TestClass]
     public class ParametersCorrectOrderTest
     {
+        private readonly VerifierBuilder builderCS = new VerifierBuilder<CS.ParametersCorrectOrder>();
```
csaba-sagi-sonarsource, line 31: I see this line repeated in a lot of tests. WDYT about creating a generic base class that would contain this logic and the specific tests would just provide the analyzer type to them?

analyzers/tests/SonarAnalyzer.UnitTest/Rules/SqlKeywordsDelimitedBySpaceTest.cs
```diff
@@ -31,54 +31,42 @@ namespace SonarAnalyzer.UnitTest.Rules
     [TestClass]
```
csaba-sagi-sonarsource, line 31: There are 3 unused usings in the file.

csaba-sagi-sonarsource (changes requested):
Please take a look at the comments.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/ParametersCorrectOrderTest.cs
Corniel: I like the idea. It is a little challenge to get two rules (VB and C# variant) both in, but that should also be possible.

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
pavel-mikula-sonarsource: Yes, as I will not remember this intentino in the future when we bump into redesigning `SonarAnalysisCOntextTest`.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/ParametersCorrectOrderTest.cs
pavel-mikula-sonarsource: I had the same idea (of course, with having two base classes as @Corniel pointed out), but @costin-zaharia-sonarsource was not happy with that design.
I will not address this in this PR. We can have a separate discussion about it.

costin-zaharia-sonarsource: I think that having a base class for all the tests can be a major trap. It will be too easy to put things there and soon it will get out of control.

csaba-sagi-sonarsource: We should agree on not putting there anything else :)

csaba-sagi-sonarsource (approved):
LGTM!

analyzers/tests/SonarAnalyzer.UnitTest/Rules/ParametersCorrectOrderTest.cs
costin-zaharia-sonarsource: > We should agree on not putting there anything else :)

It's to say this now and later "forget" the rule just because it's again very convenient in the short term. I think that adding this line with the constructor call is not that bad.

PR #5290, opened 2022-01-20
Small cleanup around cookie rules

costin-zaharia-sonarsource:


analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/CookieShouldBeHttpOnly.cs
```diff
@@ -38,8 +38,8 @@ public sealed class CookieShouldBeHttpOnly : ObjectShouldBeInitializedCorrectlyB
                 KnownType.System_Web_HttpCookie,
                 KnownType.Microsoft_AspNetCore_Http_CookieOptions);
 
-        protected override CSharpObjectInitializationTracker ObjectInitializationTracker { get; } = new CSharpObjectInitializationTracker(
-            isAllowedConstantValue: constantValue => constantValue is bool value && value,
+        protected override CSharpObjectInitializationTracker ObjectInitializationTracker { get; } = new(
+            isAllowedConstantValue: constantValue => constantValue is true,
```
mary-georgiou-sonarsource, line 42: The `constantValue` is an object. Is it always a bool (thus there won't be an exception)?

mary-georgiou-sonarsource (changes requested):
Just one comment regarding a type check.

analyzers/src/SonarAnalyzer.CSharp/Rules/Hotspots/CookieShouldBeHttpOnly.cs
costin-zaharia-sonarsource: Is, in this case, will check both the type to be bool and the value to be true. It should behave just the same as the previous version, just that it's a bit more concise.

mary-georgiou-sonarsource (approved):
LGTM

PR #5293, opened 2022-01-20
S4792: Add ASP.NET Core 6.0 test case

mary-georgiou-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/ConfiguringLoggers_AspNetCore6.cs
```diff
@@ -0,0 +1,100 @@
...
+            EventLogSettings eventLogSettings = null;
+
+            using (var loggerFactory = LoggerFactory.Create(builder => builder.AddAzureWebAppDiagnostics() // Noncompliant [1,2,3,4,5,6,7]
+                                                                              .AddAzureWebAppDiagnostics(azureSettings => { azureSettings.IncludeScopes = true; }) // Every invocation in this chain is noncompliant
+                                                                              .AddConsole(options => { options.FormatterName = "simple"; })
+                                                                              .AddDebug()
+                                                                              .AddEventLog()
+                                                                              .AddEventLog(eventLogSettings)
+                                                                              .AddEventSourceLogger()
+                                                                              .AddFilter(filter))) // Filter cannot be given to AddConsole as input. Is this an FN?
```
mary-georgiou-sonarsource, line 64: Note: I'm not sure an issue should be raised also for `AddFilter`.
In previous versions of apsnetcore the `filter` input could be given to `AddConsole` but now it seems that this is possible only through `AddFilter`.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/ConfiguringLoggersTest.cs
```diff
@@ -33,60 +33,75 @@ namespace SonarAnalyzer.UnitTest.Rules
...
+        public void ConfiguringLoggers_AspNetCore6_CS() =>
+            builderCS.AddPaths(@"Hotspots\ConfiguringLoggers_AspNetCore6.cs")
+                .AddReferences(AspNetCoreLoggingReferences(Constants.DotNetCore600Version))
+                .Verify();
+
+        [TestMethod]
+        public void ConfiguringLoggers_AspNetCoreLatest_CS() =>
+            builderCS.AddPaths(@"Hotspots\ConfiguringLoggers_AspNetCore6.cs")
+                .AddReferences(AspNetCoreLoggingReferences(Constants.NuGetLatestVersion))
+                .Verify();
```
costin-zaharia-sonarsource, line 93: Is there a difference between ver 6 and the Latest? I think we should have a single test here. For ConfiguringLoggers_AspNetCoreLatest_CS

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/ConfiguringLoggersTest.cs
```diff
@@ -95,16 +110,35 @@ public class ConfiguringLoggersTest
...
+            {
+                    AspNetCoreMetadataReference.MicrosoftAspNetCoreDiagnostics,
+                    AspNetCoreMetadataReference.MicrosoftAspNetCoreHostingAbstractions,
+                    AspNetCoreMetadataReference.MicrosoftAspNetCoreHostingWebHostBuilderExtensions,
+                    AspNetCoreMetadataReference.MicrosoftAspNetCoreHttpAbstractions,
+                    AspNetCoreMetadataReference.MicrosoftExtensionsHostingAbstractions,
+                    AspNetCoreMetadataReference.MicrosoftAspNetCoreWebHost,
+                    AspNetCoreMetadataReference.MicrosoftAspNetCoreEventSourceLoggerFactoryExtensions
+            });
+        }
```
costin-zaharia-sonarsource, line 141:
```suggestion
private static IEnumerable<MetadataReference> AspNetCoreLoggingReferences(string version) =>
            new[]
                {
                    AspNetCoreMetadataReference.MicrosoftAspNetCoreDiagnostics,
                    AspNetCoreMetadataReference.MicrosoftAspNetCoreHostingAbstractions,
                    AspNetCoreMetadataReference.MicrosoftAspNetCoreHostingWebHostBuilderExtensions,
                    AspNetCoreMetadataReference.MicrosoftAspNetCoreHttpAbstractions,
                    AspNetCoreMetadataReference.MicrosoftExtensionsHostingAbstractions,
                    AspNetCoreMetadataReference.MicrosoftAspNetCoreWebHost,
                    AspNetCoreMetadataReference.MicrosoftAspNetCoreEventSourceLoggerFactoryExtensions
                }
                .Concat(NuGetMetadataReference.MicrosoftExtensionsConfigurationAbstractions(version))
                .Concat(NuGetMetadataReference.MicrosoftExtensionsOptions(version))
                .Concat(NuGetMetadataReference.MicrosoftExtensionsLoggingPackages(version))
                .Concat(NuGetMetadataReference.MicrosoftExtensionsDependencyInjectionAbstractions(version));
```

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/ConfiguringLoggersTest.cs
```diff
@@ -95,16 +110,35 @@ public class ConfiguringLoggersTest
             NuGetMetadataReference.SerilogPackages(Constants.NuGetLatestVersion);
 
 #if NET
-        private static IEnumerable<MetadataReference> AspNetCoreLoggingReferences =>
+        private static IEnumerable<MetadataReference> AspNetCore2LoggingReferences =>
             NetStandardMetadataReference.Netstandard
```
costin-zaharia-sonarsource, line 114: `NetStandardMetadataReference.Netstandard` is no longer needed. I will do a separate PR to remove it from most of the test cases.

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/ConfiguringLoggers_AspNetCore6.cs
costin-zaharia-sonarsource: If I understand correctly the concern in this rule is with adding loggers. All the previous methods will create a logger (e.g. [AddDebug](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.debugloggerfactoryextensions.adddebug?view=dotnet-plat-ext-6.0#microsoft-extensions-logging-debugloggerfactoryextensions-adddebug(microsoft-extensions-logging-iloggingbuilder)) ). 

Since AddFilter is only concerned with filtering and no logger is created, it should not be marked as vulnerable.

I think it would be better to document its behavior on a separate line.

costin-zaharia-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/ConfiguringLoggersTest.cs
mary-georgiou-sonarsource: My understanding was that we'd have one test with a stable version (in this case I build the example on v6) and then one more test getting the latest version to see if we have breaking changes with newer versions.

Otherwise, you propose to have a test only for the latest version?

analyzers/tests/SonarAnalyzer.UnitTest/TestCases/Hotspots/ConfiguringLoggers_AspNetCore6.cs
mary-georgiou-sonarsource: I'm not sure what you mean regarding documenting the behavior of `AddFilter`.

I added one separate line in the code though to make clear that it's compliant.

costin-zaharia-sonarsource: This was it :)

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/ConfiguringLoggersTest.cs
costin-zaharia-sonarsource: The idea is to have tests only where the behavior changes: e.g. if the APIs would have changed in .Net 5 we would have added one test for Asp.Net Core 5 and one with Latest (Asp .Net Core 6).

In this case, both tests, .Net 6 and latest check the same result so we can have only the Latest test.

costin-zaharia-sonarsource (changes requested):
Please see the comment with duplicate tests

costin-zaharia-sonarsource (approved):
LGTM!

PR #5294, opened 2022-01-21
UTs: Remove OldVerifier overloads for C# 9 and 10

pavel-mikula-sonarsource:
Squash commits doesn't need a review

### DIFF
--- a/analyzers/tests/SonarAnalyzer.UnitTest/Rules/CertificateValidationCheckTest.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/Rules/CertificateValidationCheckTest.cs
@@ -19,10 +19,7 @@
  */
 
 using System;
-using System.Collections.Generic;
-using System.Linq;
 using FluentAssertions;
-using Microsoft.CodeAnalysis;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarAnalyzer.UnitTest.MetadataReferences;
 using SonarAnalyzer.UnitTest.TestFramework;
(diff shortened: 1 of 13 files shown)


analyzers/tests/SonarAnalyzer.UnitTest/Rules/CertificateValidationCheckTest.cs
```diff
@@ -34,37 +31,38 @@ namespace SonarAnalyzer.UnitTest.Rules
     [TestClass]
     public class CertificateValidationCheckTest
     {
+        private readonly VerifierBuilder builderCS = new VerifierBuilder<CS.CertificateValidationCheck>()
```
andrei-epure-sonarsource, line 34: couldn't you use here GetAdditionalReferences?

it's not a huge duplication but less is more...

andrei-epure-sonarsource (approved):
LGTM

analyzers/tests/SonarAnalyzer.UnitTest/Rules/CertificateValidationCheckTest.cs
pavel-mikula-sonarsource: A clean approach would be like this, but it produces more lines than it saves in this specific case.
```
private readonly VerifierBuilder builderCS = WithReferences<CS.CertificateValidationCheck>();
private readonly VerifierBuilder builderVB = WithReferences<VB.CertificateValidationCheck>();

        private static VerifierBuilder WithReferences<TAnalyzer>() where TAnalyzer : DiagnosticAnalyzer, new() =>
            new VerifierBuilder<TAnalyzer>()
                .AddReferences(MetadataReferenceFacade.SystemNetHttp)
                .AddReferences(MetadataReferenceFacade.SystemSecurityCryptography)
                .AddReferences(NetStandardMetadataReference.Netstandard);
```
I gave it a try with a static instance to duplicate it. The original `GetAdditionalReferences` approach should be preferably removed on the way.

PR #5295, opened 2022-01-21
Update coding style: Chained invocations

pavel-mikula-sonarsource:


docs/coding-style.md
```diff
@@ -45,23 +45,38 @@ Variable name `sut` (System Under Test) is recommended in unit tests that really
...
+        : object;
+    ```
 * Method invocation arguments should be placed on the same line only when they are few and simple. Otherwise, they should be placed on separate lines. The first argument should be on a separate line, aligned with a left-most single indendation.
-```
-object.MethodName(
-    firstArgument,
-    x => x.Bar(),
-    thirdArgument.Property);
-```
+    ```
```
pavel-mikula-sonarsource, line 69: This +4 space fencing is needed for github to properly align the code block under the bullet point and to fix the nested bullet point for `Exception from this rule:` below.

As seen [here](https://github.com/SonarSource/sonar-dotnet/blob/master/docs/coding-style.md#multi-line-statements), the nested bullet point is not understood by GitHub in the previous version

docs/coding-style.md
```diff
@@ -45,23 +45,38 @@ Variable name `sut` (System Under Test) is recommended in unit tests that really
...
-                                     Function<int, string, string> complexParameter);
-```
+    ```
+    public void MethodWithManyParameters(int firstParameter,
+                                         string secondParameter,
+                                         Function<int, string, string> complexParameter);
+    ```
+* Chained invocations and member accesses violating S103 can have a chain on the first line until the first lambda, every other `.Invocation()` or `.Member` should be on a separate line, aligned with a left-most single indendation.
+    ```
+    object.Property.Method().Select(x => x.Something)
```
costin-zaharia-sonarsource, line 55: In my opinion, all method calls should be on separate lines. I prefer not to have this kind of exception ("until the first lambda").
Our conventions should be intuitive and easy to follow.

docs/coding-style.md
```diff
@@ -45,23 +45,38 @@ Variable name `sut` (System Under Test) is recommended in unit tests that really
...
+        thirdArgument.Property);
+    ```
   * Exception from this rule: chained LINQ queries where the alignment of parameter expressions should be right-most.
-```
-someEnumerable.Select(x => x.Foo).Where(x => x.Condition1
-                                             && x.Condition2);
-```
+        ```
+        someEnumerable.Select(x => x.Foo).Where(x => x.Condition1
+                                                     && x.Condition2);
```
costin-zaharia-sonarsource, line 78: This example does not respect one of the previous rules. It should have the Where invocation on a separate line.

pavel-mikula-sonarsource: LOL 🤦‍♂️

docs/coding-style.md
pavel-mikula-sonarsource: That's why there's `can`, I wanted to define lambda as the starting point for the mandatory split. Do you have another idea how to allow boring introductions like 
```
c.GetSyntaxTree().GetRoot().DescendantNodes().OfType<xxx>()
    .Where(xxx)
```

Corniel: I think is an enormous waste of (screen)space. It also is likely to be violated (unintentionally) when renaming something.

I would argue that this is a preferable way to allign this.

``` C#
public void MethodWithManyParameters(
  int firstParameter,
  string secondParameter,
  Function<int, string, string> complexParameter);
```
It is unlikely not to fit in the screen, it is not dependent on the length of any name, and easy to read. (Just my 2cts)

costin-zaharia-sonarsource: Hi @Corniel, we were discussing here chain invocations. For method declaration parameters we already defined a rule very close to your example.

@pavel-mikula-sonarsource : I think we should be consistent and have all method calls on separate lines. 

@andrei-epure-sonarsource , @csaba-sagi-sonarsource, @mary-georgiou-sonarsource: WDYT?

csaba-sagi-sonarsource: I would say this is something that should be discussed during the How we work meeting, so everybody is aligned and hopefully happy with the decision we take. I don't see why the urgency to make the change in the coding style, we will have the meeting anyway in 3 days time.

andrei-epure-sonarsource: I agree with having chain of property access on the same line and method calls on separate lines.

The example @pavel-mikula-sonarsource gives is getters which are like properties (you can easily see them as `c.SyntaxTree.Root.DescendantNodes`), probably the MS Team made them methods because of their perf? And from this point of view, having them on separate lines will outline there is a higher cost associated with them (even if they're boring).

For example, `DescendantNodes()` without a filter can lead to major perf hit (e.g. https://github.com/SonarSource/sonar-security/pull/1783/files).

In short, I favor methods on separate lines because they do an action which has a cost.

And we can discuss this in 3 days.

mary-georgiou-sonarsource: One more point to discuss on the following "How we work" meeting:
        
 ```csharp
private readonly VerifierBuilder builderCS = new VerifierBuilder()
        .AddAnalyzer(() => new())
        .WithBasePath(@"Hotspots")
        .AddReferences(GetAdditionalReferences());
```
or

 ```csharp
private readonly VerifierBuilder builderCS = new VerifierBuilder()
                                                                      .AddAnalyzer(() => new())
                                                                      .WithBasePath(@"Hotspots")
                                                                      .AddReferences(GetAdditionalReferences());
```
so that the invocations are closer to the invocation target.

docs/coding-style.md
```diff
@@ -27,6 +27,8 @@ Static fields and properties should be placed before instance ones.
 
 Static methods are preferred to be after instance methods.
 
+Individual members must be separated by empty line, except sequence of constants, fields and abstract members. These should not be separated by empty lines.
```
pavel-mikula-sonarsource, line 30: @costin-zaharia-sonarsource This is new. I realized we didn't have it documented.

docs/coding-style.md
```diff
@@ -72,6 +98,7 @@ someEnumerable.Select(x => x.Foo).Where(x => x.Condition1
   * when it helps to understand overall structure of the method,
   * especially when each branch ends with a `return` statement.
 * Explicit `else` is not used after input validation.
+* Each compiler directive (namely `#if`/`#endif`) should be preceded and followed by an empty line.
```
pavel-mikula-sonarsource, line 101: @costin-zaharia-sonarsource This is new, from the previous how we work.

docs/coding-style.md
```diff
@@ -45,23 +47,47 @@ Variable name `sut` (System Under Test) is recommended in unit tests that really
...
-public void MethodWithManyParameters(int firstParameter,
-                                     string secondParameter,
-                                     Function<int, string, string> complexParameter);
-```
+    ```
+    public void MethodWithManyParameters(int firstParameter,
+                                         string secondParameter,
+                                         Function<int, string, string> complexParameter);
+    ```
+* Chained invocations and member accesses violating S103 can have a chain of properties and property-like methods (e.g. `GetRoot()`) on the first line. Every other `.Invocation()` or `.Member` should be on a separate line, aligned with a left-most single indendation.
```
andrei-epure-sonarsource, line 55: property-like methods are traps, see my [comment](https://github.com/SonarSource/sonar-dotnet/pull/5295#discussion_r790546438) about `DescendantNodes()`

> For example, DescendantNodes() without a filter can lead to major perf hit (e.g. SONARSEC-1603 and associated https://github.com/SonarSource/sonar-security/pull/1783).

Treating them like property access can hide their possible impact...

docs/coding-style.md
andrei-epure-sonarsource: @mary-georgiou-sonarsource I'd go for the first as well, for simplicity and having a single rule without too many exceptions

docs/coding-style.md
pavel-mikula-sonarsource: There's no universal easy way out of it. So the `should` word should provide a way to solve anything out of ordinary

andrei-epure-sonarsource: I was more worried about the 

> can have a chain of properties and property-like methods (e.g. `GetRoot()`) on the first line

andrei-epure-sonarsource:
```suggestion
* Chained invocations and member accesses violating S103 can have a chain of properties and property-like methods (e.g. `GetRoot()`) on the first line. Every other `.Invocation()` or `.Member` should be on a separate line, aligned with a left-most single indendation.
  * When property-like methods are computationally intensive (e.g. `DescendantNodes()`), they should be on a separate line.
```

pavel-mikula-sonarsource: I'd leave it implicit to keep the rule simple. Putting it on another line of course doesn't violate the rule and there might be other aspects except for performance (like long names).

costin-zaharia-sonarsource: I would also skip the mention of property-like methods and consider them as invocations. In my opinion, when formatting code we should not care much about how the method is implemented. 

We should also consider that method implementation can change in time. If this happens should we re-format the code?

docs/coding-style.md
```diff
@@ -60,6 +60,21 @@ Variable name `sut` (System Under Test) is recommended in unit tests that really
         ? object.Parent.Value
         : object;
     ```
+* Chained invocations and member accesses violating S103 can have a chain of properties on the first line. Every other `.Invocation()` or `.Member` should be on a separate line, aligned with a left-most single indendation.
+    ```
+    object.Property.GetRoot().Children
```
costin-zaharia-sonarsource, line 65: We need to update also the example to have `.GetRoot()` on the next line.

costin-zaharia-sonarsource (approved):
LGTM!

PR #5297, opened 2022-01-21
Cleanup around MarkAssemblyWithAssemblyVersionAttribute

costin-zaharia-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/Rules/MarkAssemblyWithAssemblyVersionAttributeTest.cs
```diff
@@ -40,30 +42,35 @@ public class MarkAssemblyWithAssemblyVersionAttributeTest
...
+            builderCS
+                .AddPaths("MarkAssemblyWithAssemblyVersionAttributeRazor.cs")
+                .WithConcurrentAnalysis(false)
+                .AddReferences(GetAspNetCoreRazorReferences())
+                .Verify();
 
         [TestMethod]
         public void MarkAssemblyWithAssemblyVersionAttribute_CS_Concurrent() =>
             builderCS
                 .AddPaths("MarkAssemblyWithAssemblyVersionAttribute.cs", "MarkAssemblyWithAssemblyVersionAttributeRazor.cs")
```
mary-georgiou-sonarsource, line 54: Nitpick:  I notice that sometimes when there are chained invocations, we follow [this ](https://github.com/SonarSource/sonar-dotnet/blob/aa929d4e19ba0651f62e4b7b982f3ecc246b2f79/analyzers/tests/SonarAnalyzer.UnitTest/Rules/CommentedOutCodeTest.cs#L45)style.
In this case, the first invocation goes immediately in the first line.

Which one to follow? I noticed this is not explicit in our [coding style doc](https://github.com/SonarSource/sonar-dotnet/blob/master/docs/coding-style.md).

mary-georgiou-sonarsource (approved):
LGTM - Just a coding-style comment/question.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/MarkAssemblyWithAssemblyVersionAttributeTest.cs
costin-zaharia-sonarsource: Yes, we don't have a convention for this. I prefer to have all the method invocations one under the other.

PR #5301, opened 2022-01-24
Fix reference issue and SonarAnalysisContextTest cleanup

mary-georgiou-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -46,50 +46,62 @@ public class SonarAnalysisContextTest
...
+                                                           .Concat(MetadataReferenceFacade.SystemData)
+                                                           .Concat(MetadataReferenceFacade.SystemNetHttp);
             }
         }
 
         // Various classes that invoke all the `ReportIssue` methods in AnalysisContextExtensions
         // We mention in comments the type of Context that is used to invoke (directly or indirectly) the `ReportIssue` method
-        private readonly List<TestSetup> testCases = new (new[]
-        {
+        private readonly Dictionary<TestSetup, VerifierBuilder> testCases = new()
```
costin-zaharia-sonarsource, line 67: Is there any reason not to have the builder in the `TestSetup`?

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -46,15 +46,17 @@ public class SonarAnalysisContextTest
         private class TestSetup
         {
             public string Path { get; }
-            public SonarDiagnosticAnalyzer Analyzer { get; }
+            public DiagnosticAnalyzer Analyzer { get; }
+            public VerifierBuilder Verifier { get; set; }
```
costin-zaharia-sonarsource, line 50: the `set` is not needed in this case

analyzers/tests/SonarAnalyzer.UnitTest/Helpers/SonarAnalysisContextTest.cs
```diff
@@ -63,48 +65,54 @@ public TestSetup(string testCase, SonarDiagnosticAnalyzer analyzer, IEnumerable<
...
             {
                 foreach (var testCase in testCases)
                 {
                     // ToDo: We should find a way to ack the fact the action was not run
-                    OldVerifier.VerifyNoIssueReported(testCase.Path,
-                                                   testCase.Analyzer,
-                                                   ParseOptionsHelper.FromCSharp8,
-                                                   testCase.AdditionalReferences);
+                    testCase.Verifier
+                        .AddPaths(testCase.Path)
```
costin-zaharia-sonarsource, line 112: I think that we can simplify this further by adding the `Path` and the references in the `TestSetup` constructor. We already have these when the builder is created.

Then we will be able to remove the `AdditionalReferences` since it will no longer be needed.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/TestMethodShouldContainAssertionTest.cs
```diff
@@ -55,28 +56,30 @@ private static class XUnitVersions
...
+        private readonly VerifierBuilder builder = new VerifierBuilder<TestMethodShouldContainAssertion>();
+
         [DataTestMethod]
         [DataRow(MsTestVersions.Ver1)]
         [DataRow(Constants.NuGetLatestVersion)]
         public void TestMethodShouldContainAssertion_MSTest(string testFwkVersion) =>
-            OldVerifier.VerifyAnalyzer(new[] { @"TestCases\TestMethodShouldContainAssertion.MsTest.cs", @"TestCases\TestMethodShouldContainAssertion.MsTest.AnotherFile.cs" },
-                new TestMethodShouldContainAssertion(),
-                AdditionalTestReferences(NuGetMetadataReference.MSTestTestFramework(testFwkVersion)));
+            builder.AddPaths(new[] { "TestMethodShouldContainAssertion.MsTest.cs", "TestMethodShouldContainAssertion.MsTest.AnotherFile.cs" })
```
costin-zaharia-sonarsource, line 65: It's not necessary to create a new array here since the parameter is defined with `params`.

costin-zaharia-sonarsource (changes requested):

costin-zaharia-sonarsource (approved):
LGTM!

PR #5313, opened 2022-01-24
UTs: Unify VerifierBuilder.CodeFixPaths

pavel-mikula-sonarsource:
Batch version is never used without a fixed version. So we can unify the methods.


costin-zaharia-sonarsource (approved):
LGTM!

PR #5317, opened 2022-01-25
S5122: Refactor tests and cleanup

mary-georgiou-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/PermissiveCorsTest.cs
```diff
@@ -35,27 +36,33 @@ namespace SonarAnalyzer.UnitTest.Rules
     [TestClass]
     public class PermissiveCorsTest
     {
+        private readonly VerifierBuilder builder = new VerifierBuilder()
```
costin-zaharia-sonarsource, line 39: - There is one more occurrence to change at line 84.
- I would move `.AddReferences(AdditionalReferences)` here since it's used everywhere.

costin-zaharia-sonarsource (changes requested):

costin-zaharia-sonarsource (approved):
LGTM!

PR #5319, opened 2022-01-25
S2255: Refactor tests and add tests with .NET core latest version.

mary-georgiou-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/UsingCookiesTest.cs
```diff
@@ -36,45 +37,76 @@ namespace SonarAnalyzer.UnitTest.Rules
...
             FrameworkMetadataReference.SystemWeb;
 
 #else
         [TestMethod]
         public void UsingCookies_CS_NetCore() =>
-            OldVerifier.VerifyAnalyzer(
-                @"TestCases\Hotspots\UsingCookies_NetCore.cs",
-                new CS.UsingCookies(AnalyzerConfiguration.AlwaysEnabled),
-                GetAdditionalReferencesForNetCore(Constants.DotNetCore220Version));
+            builderCS
```
costin-zaharia-sonarsource, line 69: I think we should let only this test with AspNet Core 2 and remove the other tests with Asp.Net Core 2.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/UsingCookiesTest.cs
```diff
@@ -36,45 +37,76 @@ namespace SonarAnalyzer.UnitTest.Rules
...
+                .AddPaths("UsingCookies_NetCore.cs")
+                .AddReferences(GetAdditionalReferencesForNetCore(Constants.DotNetCore220Version))
```
costin-zaharia-sonarsource, line 71: I think we should rename this method to `GetAspNetCoreReferences` since they are related to Asp.Net Core.

costin-zaharia-sonarsource (changes requested):

costin-zaharia-sonarsource (approved):
LGTM!

PR #5320, opened 2022-01-25
Cleanup around S5693

costin-zaharia-sonarsource:


analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/RequestsWithExcessiveLengthTest.cs
```diff
@@ -34,56 +34,56 @@ namespace SonarAnalyzer.UnitTest.Rules
     [TestClass]
     public class RequestsWithExcessiveLengthTest
     {
+        private readonly VerifierBuilder builderCS = new VerifierBuilder()
+                                                     .AddAnalyzer(() => new CS.RequestsWithExcessiveLength(AnalyzerConfiguration.AlwaysEnabled))
+                                                     .WithBasePath(@"Hotspots")
+                                                     .AddReferences(GetAdditionalReferences());
```
mary-georgiou-sonarsource, line 40: nitpick:  Shouldn't the alignment be like this (not sure, but I'm a bit confused):
```suggestion    
{
        private readonly VerifierBuilder builderCS = new VerifierBuilder()
            .AddAnalyzer(() => new CS.RequestsWithExcessiveLength(AnalyzerConfiguration.AlwaysEnabled))
            .WithBasePath(@"Hotspots")
            .AddReferences(GetAdditionalReferences());
```

mary-georgiou-sonarsource (approved):
LGTM - nitpick comment on coding style.

analyzers/tests/SonarAnalyzer.UnitTest/Rules/Hotspots/RequestsWithExcessiveLengthTest.cs
costin-zaharia-sonarsource: me neither :) 

for me, if we move the calls there they would be to far from the invocation target. I don't have a strong opinion about this.

mary-georgiou-sonarsource: I'll add an extra note in the discussion about chained invocations for this. :)

PR #5321, opened 2022-01-25
Improve old Symbolic Execution memory consumption

pavel-mikula-sonarsource:
We can help GC to collect unused nodes sooner


costin-zaharia-sonarsource (approved):
Nice find. They were added to a list without any need.

PR #5322, opened 2022-01-25
Refactor ProgramState to init-only setters

pavel-mikula-sonarsource:
This approach will help to reduce duplications when more items are tracked.

It should prevent duplications like this:
https://github.com/SonarSource/sonar-dotnet/blob/c0d92c7faa43e0e0e112d3f667a78fc64e51bddd/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Sonar/ProgramState.cs#L268-L341

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/Helpers/IsExternalInit.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Helpers/IsExternalInit.cs
@@ -22,7 +22,10 @@ using System.ComponentModel;
 
 namespace System.Runtime.CompilerServices
 {
-    // This empty class needs to exist when C# 9 init-only setters are used in project targeting net48.
+    // This empty class needs to exist when C# 9 init-only setters are used in project targeting .NET Framework.
+    // https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.isexternalinit?view=net-6.0
+    // It is used only by compiler to track metadata. It does not affect MSIL, CLR nor runtime.
+    // https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/init#metadata-encoding
     [EditorBrowsable(EditorBrowsableState.Never)]
     internal static class IsExternalInit { }
 }
--- a/analyzers/src/SonarAnalyzer.Common/Helpers/IsExternalInit.cs
+++ b/analyzers/src/SonarAnalyzer.Common/Helpers/IsExternalInit.cs
@@ -18,8 +18,6 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-#if NETFRAMEWORK
-
 using System.ComponentModel;
 
 namespace System.Runtime.CompilerServices
@@ -28,5 +26,3 @@ namespace System.Runtime.CompilerServices
     [EditorBrowsable(EditorBrowsableState.Never)]
     internal static class IsExternalInit { }
 }
-
-#endif
--- a/analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
+++ b/analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
@@ -26,7 +26,7 @@ using StyleCop.Analyzers.Lightup;
 
 namespace SonarAnalyzer.SymbolicExecution.Roslyn
 {
-    public sealed class ProgramState
+    public sealed record ProgramState
     {
         public static readonly ProgramState Empty = new();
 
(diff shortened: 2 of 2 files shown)


analyzers/src/SonarAnalyzer.Common/Helpers/IsExternalInit.cs
```diff
@@ -18,8 +18,6 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-#if NETFRAMEWORK
```
costin-zaharia-sonarsource, line 21: Why do we need this change? This hack was supposed to be needed only for .Net Framework.

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
```diff
@@ -28,31 +28,37 @@ namespace SonarAnalyzer.SymbolicExecution.Roslyn
 {
     public sealed class ProgramState
     {
-        public static readonly ProgramState Empty = new(ImmutableDictionary<IOperation, SymbolicValue>.Empty, ImmutableDictionary<ISymbol, SymbolicValue>.Empty);
+        public static readonly ProgramState Empty = new();
```
costin-zaharia-sonarsource, line 31: I would do a couple of changes to this approach:
- make this class a `Record`
- drop de constructor with parameters (`ProgramState`) - it's not actually needed
- update `SetOperationValue` body to `this with { OperationValue = OperationValue.SetItem(operation, value) };`
- update `SetSymbolValue` body to `this with { SymbolValue = SymbolValue.SetItem(symbol, value) };`

costin-zaharia-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Common/Helpers/IsExternalInit.cs
pavel-mikula-sonarsource: The file was moved from UT project targeting 
* `net6.0` - works out of the box, #if made sense there
* `net48` - needs hack
to the `SonarAnalyzer.Common` project targeting
* net46 - needs hack
* netstandard2.0 - needs hack

So I had to remove the conditional compilation. It does no harm to the UT project.

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
pavel-mikula-sonarsource: That's a nice idea! I'll do the same for the `VerifierBuilder`.

analyzers/src/SonarAnalyzer.Common/Helpers/IsExternalInit.cs
costin-zaharia-sonarsource: Very good point. This hack was previously part of the UTs where we could mock the missing part with low risk. 

On the other hand, I think it's not right to do this for production code since the analyzer will run also on PC with only .Net Framework where the syntax is not officially supported. We might have unexpected issues due to this.

I think it would be better to let the file in the UTs and find another approach to this problem.

costin-zaharia-sonarsource (changes requested):
Please see my comment related to `IsExternalInit`

analyzers/src/SonarAnalyzer.Common/Helpers/IsExternalInit.cs
pavel-mikula-sonarsource: Based on [the docs](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.isexternalinit?view=net-6.0), this class is used only by a compiler to track metadata. It is not used during runtime and by-design should not change program behavior. There are also several blog posts suggesting describing this solution and a nugget package that adds this file.
I don't see this as a risk.

costin-zaharia-sonarsource: If we want to go with this approach we should document the solution with at least a link to the blog posts and do proper testing on a machine that has only .Net Framework installed. An idea would be to do a manual build of the plugin, install it on a VM and try to analyze a reasonable big project to see if there are any errors or differences in the issues we raise.

pavel-mikula-sonarsource: Only S2222 uses this `ProgramState` and it will not detect anything on any known project.
The init-only setter generates a classic setter in the MSIL form. So there is no change for CLR.
I'll add links to the class.

pavel-mikula-sonarsource: I've added this link that describes the behavior as part of C# 9 specification:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/init#metadata-encoding

costin-zaharia-sonarsource (approved):
LGTM!

Issue #5306, opened 2022-01-24
Support branching in Symbolic Execution


pavel-mikula-sonarsource:
Add proper branching support in the new Symbolic Execution engine

PR #5323, opened 2022-01-25
Limit Symbolic Execution MaxSteps

pavel-mikula-sonarsource:
Related to #5306

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
+++ b/analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
@@ -30,6 +30,8 @@ namespace SonarAnalyzer.SymbolicExecution.Roslyn
 {
     internal class RoslynSymbolicExecution
     {
+        internal const int MaxStepCount = 2000;
+
         private readonly ControlFlowGraph cfg;
         private readonly SymbolicCheck[] checks;
         private readonly Queue<ExplodedNode> queue = new();
@@ -48,9 +50,14 @@ namespace SonarAnalyzer.SymbolicExecution.Roslyn
         public void Execute()
         {
             // ToDo: Forbid running twice
+            var steps = 0;
             queue.Enqueue(new ExplodedNode(cfg.EntryBlock, ProgramState.Empty));
             while (queue.Any())
             {
+                if (steps++ > MaxStepCount)
+                {
+                    return;
+                }
                 var current = queue.Dequeue();
                 var successors = current.Operation == null ? ProcessBranching(current) : ProcessOperation(current);
                 foreach (var node in successors)
--- a/analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.Branching.cs
+++ b/analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/RoslynSymbolicExecutionTest.Branching.cs
@@ -139,7 +139,6 @@ else
             validator.ValidateExecutionNotCompleted();
         }
 
-
         [TestMethod]
         public void EndNotifications_MultipleBranches()
         {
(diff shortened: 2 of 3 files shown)


analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
```diff
@@ -30,6 +30,8 @@ namespace SonarAnalyzer.SymbolicExecution.Roslyn
 {
     internal class RoslynSymbolicExecution
     {
+        internal const int MaxStepCount = 2000;
```
costin-zaharia-sonarsource, line 33: Is now the right time to do this? Shouldn't we add this check only after we measure the performance once the engine is mature enough?

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
```diff
@@ -48,9 +50,14 @@ public RoslynSymbolicExecution(ControlFlowGraph cfg, SymbolicCheck[] checks)
         public void Execute()
         {
             // ToDo: Forbid running twice
+            var steps = 0;
             queue.Enqueue(new ExplodedNode(cfg.EntryBlock, ProgramState.Empty));
             while (queue.Any())
```
costin-zaharia-sonarsource, line 55: Could we move the condition here `while (queue.Any() && steps++ <= MaxStepCount)`? I find it easier to understand.

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
```diff
@@ -48,9 +50,14 @@ public RoslynSymbolicExecution(ControlFlowGraph cfg, SymbolicCheck[] checks)
...
             {
+                if (steps++ > MaxStepCount)
+                {
+                    return;
```
costin-zaharia-sonarsource, line 59: I think we should break the loop and not return. The execution was completed (earlier than planned but it's still completed) and we should send a message that this has happened.

costin-zaharia-sonarsource (changes requested):

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
pavel-mikula-sonarsource: We need to add safety limitations as soon as possible to avoid unexpected performance explosions. Once the engine is mature enough, we can measure it and adjust this to a better value.

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
pavel-mikula-sonarsource: Unfortunately no, because we need to avoid `NotifyExecutionCompleted` call.

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs
pavel-mikula-sonarsource: That could raise FPs that we need to avoid. As rules depending on `NotifyExecutionCompleted` (like the S2222, as a lucky example) needs to visit the sanitizing call.

costin-zaharia-sonarsource: We also have rules which are fine with partial execution (e.g. NullDereference). Can't we return a result that lets the rules know how the execution is completed?

pavel-mikula-sonarsource: Those rules will report their partial results on the fly and should not be affected. And this will actually be an improvement with the new engine, as the old engine also returns and doesn't reach the `OnExplorationEnded` part.

https://github.com/SonarSource/sonar-dotnet/blob/773f29259126499169fb5954bda5b7a44b1ce8d3/analyzers/src/SonarAnalyzer.CSharp/SymbolicExecution/Sonar/AbstractExplodedGraph.cs#L87-L91

costin-zaharia-sonarsource (approved):
LGTM! 

There is one small code smell you might want to fix.

PR #5331, opened 2022-01-26
Update Ember-MM log4net dependency

pavel-mikula-sonarsource:
Update Integration Test dependency to avoid CVE-2018-1285 being raised. 
Our analyzer is not at risk.


andrei-epure-sonarsource (approved):
LGTM

PR #5335, opened 2022-01-28
Add VS2022 to NuGet install scripts

### LINKED ISSUE #5334: Unable to install or update NuGet package in Visual Studio 2022
Trying to install or update the SonarAnalyzer.VisualBasic NuGet package fails in Visual Studio 2022. The error message displayed is "This package can only be installed on Visual Studio 2015 Update 3 or later." I assume the same error would occur for the SonarAnalyzer.CSharp package.

I can update the package by directly editing the "packages.config" file.

The problem seems to be [these lines in install.ps1](https://github.com/SonarSource/sonar-dotnet/blob/3a67c9104315465eaa204442ad58cca02e67f27c/analyzers/packaging/tools-vbnet/install.ps1#L5-L7):

```powershell
if ('14.0', '15.0', '16.0' -notcontains $project.DTE.Version) {
    $invalidVsVersion = $true
}
```

Visual Studio 2022 is version 17!

pavel-mikula-sonarsource:
Fixes #5334

This is a hotfix, we should discuss the purpose and strategy around this piece.

I've tested the change manually by updating cached plugin.

### DIFF
--- a/analyzers/packaging/tools-cs/install.ps1
+++ b/analyzers/packaging/tools-cs/install.ps1
@@ -2,7 +2,7 @@
 
 $invalidVsVersion = $false
 
-if ('14.0', '15.0', '16.0' -notcontains $project.DTE.Version) {
+if ('14.0', '15.0', '16.0', '17.0' -notcontains $project.DTE.Version) {
     $invalidVsVersion = $true
 }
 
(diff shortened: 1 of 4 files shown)


costin-zaharia-sonarsource (approved):
LGTM!

PR #5339, opened 2022-01-28
Add ExplodedNode.ToString() for debugging SE

pavel-mikula-sonarsource:
Related to #5306

### DIFF
--- a/analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ExplodedNode.cs
+++ b/analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ExplodedNode.cs
@@ -62,5 +62,10 @@ namespace SonarAnalyzer.SymbolicExecution.Roslyn
             other is not null
             && other.programPointHash == programPointHash
             && other.State.Equals(State);
+
+        public override string ToString() =>
+            Operation is null
+                ? $"Block #{Block.Ordinal}, Branching{Environment.NewLine}{State}"
+                : $"Block #{Block.Ordinal}, Operation #{index}, {Operation.Instance.Serialize()}{Environment.NewLine}{State}";
     }
 }
--- a/analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
+++ b/analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
@@ -69,30 +69,28 @@ namespace SonarAnalyzer.SymbolicExecution.Roslyn
             && other.OperationValue.DictionaryEquals(OperationValue)
             && other.SymbolValue.DictionaryEquals(SymbolValue);
 
-        public override string ToString()
+        public override string ToString() =>
+            Equals(Empty) ? "Empty" : SerializeSymbols() + SerializeOperations();
+
+        private string SerializeSymbols() =>
+            Serialize(SymbolValue, "Symbols", x => x.ToString(), x => x?.ToString() ?? "<null>");
+
+        private string SerializeOperations() =>
+            Serialize(OperationValue, "Operations", x => x.Serialize(), x => x?.ToString() ?? "<null>");
+
+        private static string Serialize<TKey, TValue>(ImmutableDictionary<TKey, TValue> dictionary, string title, Func<TKey, string> serializeKey, Func<TValue, string> serializeValue)
         {
-            if (Equals(Empty))
+            if (dictionary.IsEmpty)
             {
-                return "Empty";
+                return null;
             }
             else
             {
                 var sb = new StringBuilder();
-                if (SymbolValue.Any())
+                sb.Append(title).AppendLine(":");
+                foreach (var kvp in dictionary.Select(x => new KeyValuePair<string, string>(serializeKey(x.Key), serializeValue(x.Value))).OrderBy(x => x.Key))
                 {
-                    sb.AppendLine("Symbols:");
-                    foreach (var kvp in SymbolValue.OrderBy(x => x.Key.ToString()))
-                    {
-                        sb.Append(kvp.Key).Append(": ").AppendLine(kvp.Value?.ToString() ?? "<null>");
-                    }
-                }
-                if (OperationValue.Any())
-                {
-                    sb.AppendLine("Operations:");
-                    foreach (var kvp in OperationValue.Select(x => new KeyValuePair<string, string>(x.Key.Serialize(), x.Value?.ToString() ?? "<null>")).OrderBy(x => x.Key))
-                    {
-                        sb.Append(kvp.Key).Append(": ").AppendLine(kvp.Value);
-                    }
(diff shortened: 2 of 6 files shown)


analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ExplodedNode.cs
```diff
@@ -62,5 +62,10 @@ private ExplodedNode(BasicBlock block, IOperationWrapperSonar[] operations, int
             other is not null
             && other.programPointHash == programPointHash
             && other.State.Equals(State);
+
+        public override string ToString() =>
```
Corniel, line 66: For debugging, newlines are not particular useful. Consider to (also) have a [DebuggerDisplay] attribute that has a single line representation.

pavel-mikula-sonarsource: Hi @Corniel, thanks for the suggestion. We've considered the `[DebuggerDisplay]` option and went for the `ToString`. While the `ExplodedNode` itself doesn't really need the new line, the `ProgramState` does.

The intended usage is to add `Console.WriteLine(current.ToString())` locally here, as the flows can be often tricky to follow and understand:
https://github.com/SonarSource/sonar-dotnet/blob/3f1f4cdbebba4d87acd571579389bb7d8324a9ed/analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/RoslynSymbolicExecution.cs#L50-L62

Corniel: You could have a debugger display to, that does `ToString().Replace(Environement.NewLine, "; "). For Console.WriteLine, new lines indeed have added value.

pavel-mikula-sonarsource: We'll see if that's needed later. For now, I'm a bit afraid that the state will be too long to read in a single line anyway based on the experience with the old engine.

pavel-mikula-sonarsource: But I like the `Replace` idea if we'll need taht

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
```diff
@@ -66,5 +68,34 @@ private ProgramState()
...
+                if (OperationValue.Any())
+                {
+                    sb.AppendLine("Operations:");
+                    foreach (var kvp in OperationValue.Select(x => new KeyValuePair<string, string>(x.Key.Serialize(), x.Value?.ToString() ?? "<null>")).OrderBy(x => x.Key))
+                    {
+                        sb.Append(kvp.Key).Append(": ").AppendLine(kvp.Value);
+                    }
+                }
+                return sb.ToString();
+            }
```
costin-zaharia-sonarsource, line 98: I think we should extract 2 methods: GetSymbols, GetOperations and then transform this full method into an expression.

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/ExplodedNodeTest.cs
```diff
@@ -131,5 +131,23 @@ public void GetHashCode_ReturnsSameForEquivalent()
             basic.GetHashCode().Should().NotBe(differentLocation.GetHashCode());
             basic.GetHashCode().Should().NotBe(differentState.GetHashCode());
         }
+
+        [TestMethod]
+        public void ToString_SerializeOperationAndState()
+        {
+            var cfg = TestHelper.CompileCfgBodyCS("var a = true;");
+            var state = ProgramState.Empty.SetSymbolValue(cfg.Blocks[1].Operations[0].Children.First().TrackedSymbol(), new(new()));
```
costin-zaharia-sonarsource, line 139: `new(new())` -> hard to figure out what's happening here without looking at the method signature.

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/ProgramStateTest.cs
```diff
@@ -315,6 +316,88 @@ public void GetHashCode_ReturnsSameForEquivalent()
...
+a: SV_1
+Sample.Main(): SV_2: Second
+");
+        }
+
+        [TestMethod]
+        public void ToString_WithOperations()
+        {
+            var assignment = TestHelper.CompileCfgBodyCS("var a = true;").Blocks[1].Operations[0];
+            var counter = new SymbolicValueCounter();
```
costin-zaharia-sonarsource, line 355: We use a counter in all the tests. We could make it a field and use a TestInitialize method to construct it before each test.

costin-zaharia-sonarsource (changes requested):

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/ExplodedNodeTest.cs
pavel-mikula-sonarsource: That's a kind of build-in for target typed new feature. As we're inside UT, it is just a boring scaffolding.

analyzers/tests/SonarAnalyzer.UnitTest/SymbolicExecution/Roslyn/ProgramStateTest.cs
pavel-mikula-sonarsource: I'm considering refactoring these UTs to use `TestInitialize` to generate some `context` field that would provide `Operation1`, `Operation2`, `Operation3`, `Value1`, `Value2`, `Value3`, `Symbol1`, `Symbol2` and `Symbol3` as those repeat in most of the cases. The same type could be then used in `ProgramStateTest`. In another PR

analyzers/src/SonarAnalyzer.Common/SymbolicExecution/Roslyn/ProgramState.cs
```diff
@@ -66,5 +68,32 @@ private ProgramState()
             other is not null
             && other.OperationValue.DictionaryEquals(OperationValue)
             && other.SymbolValue.DictionaryEquals(SymbolValue);
+
+        public override string ToString() =>
+            Equals(Empty) ? "Empty" : SerializeSymbols() + SerializeOperations();
+
+        private string SerializeSymbols() =>
+            Serialize(SymbolValue, "Symbols", x => x.ToString(), x => x?.ToString() ?? "<null>");
```
costin-zaharia-sonarsource, line 76: This last part `x => x?.ToString() ?? "<null>"`, can be included in the method.

costin-zaharia-sonarsource (approved):
LGTM! Just a small comment to address.

